Issue #2121, opened 2020-10-28
Slack transport with very large error messages errors out


chrismaree:
**Description**

If a very large message is sent to the slack transport then the slack API will error out and return 400. This causes the transport to freeze and fail silently. 

**Reproduction**

Send a very large (over 4000 char) string over the slack transport from a bot.

**Resolution**
To fix the transport should better deal with large messages by truncating what is sent and not erroring out.

PR #2118, opened 2020-10-27
improve(serverless-hub): Update spoke error log

nicholaspai:
**Motivation**

#2112

**Summary**

This PR refactors how the hub deals with error messages by making two main changes:
1) the hub now detects if the error thrown occurred in the hub process (an "internal hub error") and throws accordingly.
2) if the error was created in a spoke process then the hub will print the associated error message with the spoke process that generated it.

**Details**

**Issue(s)**

Fixes #2112

PR #2124, opened 2020-10-29
improve(slack-transport) remove upper bound on slack message length

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2121

Previously the slack transport was limited to sending slack messages no larger than 3500 chars due to a limitation in the slack webhook API. If a message was sent to the API larger than this limit then the transport would fail silently. Normally, this was never an issue as the usage of the Winston logger than would send slack messages is conscious of the size as we are always consistent to not send massive error messages. 

However, this is not always going to be a fair assumption that messages are less than this limit. For example, https://github.com/UMAprotocol/protocol/pull/2118 could produce a slack message that exceeds this limit if the number of bots running is fairly high and they all error out at the same time.

**Summary**

This PR removes the previously used `winston-slack-webhook-transport` and replaces it with our own implementation. This implementation considers the size of the message being sent to slack, provided to the Winston transport, and splits up the log into multiple slack messages accordingly. 

**Details**
There are three main logic branches that this PR introduces:
- If a message is less than 3000 chars in total (considering the header, markdown formatting, and all other associated details) then it is sent to slack in one API call.
- A slack message comprises multiple blocks. In the slack transport, each block would be defined as a key-value pair component in the log sent to Winston. For example, a message generated as `Logger.error({at:"example",message:"some string",some-random-key:"this is an example key"}) would be turned into 3 blocks, for each key within the object. If each block's length exceeds 3000 chars then the new transport will redact the middle part of the block message.
- If the overall set of blocks within a message exceeds 3000 chars then the transport will split the log up into multiple messages to send to the slack API.

**Testing**
This implementation was tested against the three different kinds of messages described above and the output was validated in slack. For example

1) simple case base message:
```
Logger.error({at:"test",message:"testmessage"})
```

2) long message block where redaction occurs:
```
Logger.error({at:"test",message:"testmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessage"})
```

3) multiple block, each has no redaction but collectively gets split over multiple messages:
```
Logger.error({at:"test",message:"testmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagees",message2:"tmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessage"})
``` 

4) multiple blocks where each gets redacted and spread over multiple messages
```
Logger.error({at:"test",message:"testmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagees",message2:"tmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessagetestmessageestmessagetestmessagetestmessagetessage"})
```

**Issue(s)**

close https://github.com/UMAprotocol/protocol/issues/2121

### DIFF
--- a/packages/financial-templates-lib/src/logger/SlackTransport.js
+++ b/packages/financial-templates-lib/src/logger/SlackTransport.js
@@ -21,8 +21,8 @@
 //      liquidationResult: logResult});
 //    In this log the liquidation and txnConfig are objects. these are spread as nested bullet points in the slack message.
 //    The amount is a string value. This is shown as a bullet point item.
-
-const SlackHook = require("winston-slack-webhook-transport");
+const Transport = require("winston-transport");
+const axios = require("axios").default;
 const { createEtherscanLinkMarkdown } = require("@uma/common");
 
 function slackFormatter(info) {
@@ -160,6 +160,71 @@ function slackFormatter(info) {
         }
       ]
     };
+  }
+}
+
+class SlackHook extends Transport {
+  constructor(opts) {
+    super(opts);
+    opts = opts || {};
+    this.name = opts.name || "slackWebhook";
+    this.level = opts.level || undefined;
+    this.webhookUrl = opts.webhookUrl;
+    this.formatter = opts.formatter || undefined;
+    this.mrkdwn = opts.mrkdwn || false;
+
+    this.axiosInstance = axios.create({
+      proxy: opts.proxy || undefined
+    });
+  }
+
+  async log(info, callback) {
+    let payload = {
+      mrkdwn: this.mrkdwn
+    };
+
+    let layout = this.formatter(info);
+    payload.text = layout.text || undefined;
+    payload.attachments = layout.attachments || undefined;
+    payload.blocks = layout.blocks || undefined;
+    let errorThrown = false;
+    // If the overall payload is less than 3000 chars then we can send it all in one go to the slack API.
+    if (JSON.stringify(payload).length < 3000) {
+      let response = await this.axiosInstance.post(this.webhookUrl, payload);
+      if (response.status != 200) errorThrown = true;
+    } else {
+      // If it's more than 3000 chars then we need to split the message sent to slack API into multiple calls.
+      let messageIndex = 0;
+      let processedBlocks = [[]];
+      for (let block of payload.blocks) {
+        if (JSON.stringify(block).length > 3000) {
+          // If the block (one single part of a message) is larger than 3000 chars then we must redact part of the message.
+          const stringifiedBlock = JSON.stringify(block);
+          const redactedBlock =
+            stringifiedBlock.substr(0, 1000) +
+            "- MESSAGE REDACTED DUE TO LENGTH -" +
+            stringifiedBlock.substr(stringifiedBlock.length - 1000, stringifiedBlock.length);
+          block = JSON.parse(redactedBlock);
+        }
+        if (JSON.stringify([...processedBlocks[messageIndex], block]).length > 3000) {
+          // If the set blocks is larger than 3000 then we must increment the message index, to enable sending the set
+          // of messages over multiple calls to the slack API. The amounts to splitting up one Winston log into multiple
+          // slack messages with no single slack message exceeding the 3000 char limit.
+          messageIndex += 1;
+        }
+        if (!processedBlocks[messageIndex]) processedBlocks[messageIndex] = [];
+        processedBlocks[messageIndex].push(block);
+      }
+      // Iterate over each message to send and generate a axios call for each message.
+      for (const processedBlock of processedBlocks) {
+        payload.blocks = processedBlock;
+        let response = await this.axiosInstance.post(this.webhookUrl, payload);
+        if (response.status != 200) errorThrown = true;
+      }
+    }
+    callback();
+    if (!errorThrown) this.emit("logged", info);
+    else this.emit("error", errorThrown);
   }
 }
 
--- a/packages/financial-templates-lib/src/logger/SlackTransport.js
+++ b/packages/financial-templates-lib/src/logger/SlackTransport.js
@@ -202,7 +202,7 @@ class SlackHook extends Transport {
           const stringifiedBlock = JSON.stringify(block);
           const redactedBlock =
             stringifiedBlock.substr(0, 1000) +
-            "- MESSAGE REDACTED DUE TO LENGTH -" +
+            "-MESSAGE REDACTED DUE TO LENGTH-" +
             stringifiedBlock.substr(stringifiedBlock.length - 1000, stringifiedBlock.length);
           block = JSON.parse(redactedBlock);
         }
--- a/packages/financial-templates-lib/src/logger/SlackTransport.js
+++ b/packages/financial-templates-lib/src/logger/SlackTransport.js
@@ -201,9 +201,9 @@ class SlackHook extends Transport {
           // If the block (one single part of a message) is larger than 3000 chars then we must redact part of the message.
           const stringifiedBlock = JSON.stringify(block);
           const redactedBlock =
-            stringifiedBlock.substr(0, 1000) +
+            stringifiedBlock.substr(0, 1400) +
             "-MESSAGE REDACTED DUE TO LENGTH-" +
-            stringifiedBlock.substr(stringifiedBlock.length - 1000, stringifiedBlock.length);
+            stringifiedBlock.substr(stringifiedBlock.length - 1400, stringifiedBlock.length);
           block = JSON.parse(redactedBlock);
         }
         if (JSON.stringify([...processedBlocks[messageIndex], block]).length > 3000) {
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -15,7 +15,6 @@
     "node-fetch": "^2.6.0",
     "node-pagerduty": "^1.2.0",
     "winston": "^3.2.1",
-    "winston-slack-webhook-transport": "^1.2.1",
     "winston-transport": "^4.3.0"
   },
   "devDependencies": {


packages/financial-templates-lib/src/logger/SlackTransport.js
```diff
@@ -163,6 +163,71 @@ function slackFormatter(info) {
...
+    } else {
+      // If it's more than 3000 chars then we need to split the message sent to slack API into multiple calls.
+      let messageIndex = 0;
+      let processedBlocks = [[]];
+      for (let block of payload.blocks) {
+        if (JSON.stringify(block).length > 3000) {
+          // If the block (one single part of a message) is larger than 3000 chars then we must redact part of the message.
+          const stringifiedBlock = JSON.stringify(block);
+          const redactedBlock =
+            stringifiedBlock.substr(0, 1400) +
```
nicholaspai, line 204: What's the 1400 magic number?

nicholaspai (approved):
Really like this solution so that we don't have to worry about truncating messages in the bot loggers.

packages/financial-templates-lib/src/logger/SlackTransport.js
chrismaree: The upper bound for a message is 3000 charts. I set this number to 1400 chars on each end of the truncation to leave ~100 chars free in the block to ensure that it doesn't overflow.

packages/financial-templates-lib/src/logger/SlackTransport.js
```diff
@@ -163,6 +163,71 @@ function slackFormatter(info) {
   }
 }
 
+class SlackHook extends Transport {
```
mrice32, line 166: Can you briefly explain what this is replacing? I'm assuming we were previously using a built-in slack transport, but we had a custom slack formatter, but now we have both a custom formatter and custom transport?

mrice32 (approved):
Looks great!!

packages/financial-templates-lib/src/logger/SlackTransport.js
chrismaree: Previously we were using the [winston-slack-transport](https://github.com/TheAppleFreak/winston-slack-webhook-transport#readme) but this had limitations on the size of slack messages that could get sent over the transport. The implementation would init a new instance of their [SlackTransport](https://github.com/TheAppleFreak/winston-slack-webhook-transport/blob/a6d3bc82692cd7737593de1cd12b2ad44dc4f703/slackHook.js#L6). 

After this PR we implement a more similar pattern to what we do in the [`PagerDudty`](https://github.com/UMAprotocol/protocol/blob/ad3bac05ef38a17ebf2be370e4e61a1452f5291e/packages/financial-templates-lib/src/logger/PagerDutyTransport.js#L10) transport which is our own custom transport. The standard pattern for these transports is to extend the Transport class from Winston and then implement a `log` function. We could merge the formatter into this transport but it felt cleaner to leave it as two separate functions. 

I might take another stab at refactoring all the transports in this directory to be more consistent in how they are exported but for now, I think it's sufficient.

PR #2127, opened 2020-10-29
feat(dev-mining) update price feed to cover two discrete timestamps

chrismaree:
**Motivation**

Previously the Coingecko price feed was limited to only read x number of days in the past. This makes results unpredictable and impossible to reproduce if re-run at a later time. This PR updates the price feed to use a discrete start and end timestamp.

**Summary**

Updated price feed, its usage, and associated tests.

**Details**

Data sets were also updated accordingly.


packages/affiliates/test/datasets/set1.js
```diff
@@ -6,9 +6,9 @@ const empContracts = ["0xaBBee9fC7a882499162323EEB7BF6614193312e3", "0x3605Ec11B
 const syntheticTokens = ["0xF06DdacF71e2992E2122A1a0168C6967aFdf63ce", "0xD16c79c8A39D44B2F3eB45D2019cd6A42B03E2A9"];
 const syntheticTokenDecimals = [18, 18];
 
-const startingTimestamp = moment("9/23/2020 23:00:00", "MM/DD/YYYY  HH:mm z").valueOf(); // utc timestamp
+const startingTimestamp = moment("2020-09-23 23:00:00", "YYYY-MM-DD  HH:mm Z").valueOf(); // utc timestamp
```
chrismaree, line 9: changed this to match the other time standards through the repo.

packages/affiliates/test/models.js
```diff
@@ -74,8 +74,10 @@ describe("History", function() {
 describe("Prices", function() {
   let prices, seed;
   const token = "0xD16c79c8A39D44B2F3eB45D2019cd6A42B03E2A9";
+  const startingTimestamp = moment("2020-09-23 23:00:00", "YYYY-MM-DD  HH:mm Z").valueOf();
+  const endingTimestamp = moment("2020-10-05 23:00:00", "YYYY-MM-DD  HH:mm Z").valueOf();
   it("init", async function() {
-    seed = await Coingecko().chart(token, "usd", "10");
+    seed = await Coingecko().chart(token, "usd", startingTimestamp, endingTimestamp);
```
daywiss, line 80: i forot about this test, but its making a real call out to coingecko in CI. seems fine to me but what are your thoughts on that?

daywiss (approved):

packages/affiliates/test/models.js
chrismaree: I think it's fine for now. we should refactor it to use the sample set though in the future.

nicholaspai: We should probably mock the API call instead of making the real one. We've had previous tests break that used to call out to Intrinio API if you recall.

nicholaspai (approved):
100% agreed that its better to remove uncertainty and be able to determistically replicate a "timestamp's" data. LGTM

PR #2129, opened 2020-10-29
feat(affiliates): add emp and erc20 interfaces for getting informatio…

### LINKED ISSUE #2123: add a new library module to gather state from emps: collateral/token addresses and collateral/token decimals
## Why
The script will require this information in order to evaluate each emp contract. User could technically supply this data, but since we can get it from Web3 it would simplify UX. We should modularize this and add it as a library. 

## How
This could be implemented as a library which takes in web3 and has a function like `getEMPState(address)` which returns something like 
```
{
  collateral:{
    address, decimals
  },
  tokens:{
    address, decimals
  },
  ...etc
}
```

daywiss:
…n for calculations from web3

Signed-off-by: David Adams <david@adams.software>

**Motivation**

#2123 

**Summary**

Adds a small wrapper around web3 to allow grabbing emp state per address. Also adds truffle tests, so seperates test folder into
`/datasets` `/mocha` `/truffle` and adds yarn scripts for testing each.

**Issue(s)**

Fixes #2123

### DIFF
--- a/packages/affiliates/libs/contracts.js
+++ b/packages/affiliates/libs/contracts.js
@@ -1,7 +1,8 @@
 const ethers = require("ethers");
 const assert = require("assert");
+const { getAbi } = require("@uma/core");
 function DecodeLog(abi, meta = {}) {
-  assert(abi, "requries abi");
+  assert(abi, "requires abi");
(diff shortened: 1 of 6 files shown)


nicholaspai (approved):
+1 for including tests

PR #2138, opened 2020-11-02
feat(affiliates): add emp and erc20 interfaces for getting informatio…

### LINKED ISSUE #2123: add a new library module to gather state from emps: collateral/token addresses and collateral/token decimals

daywiss:
…n for calculations from web3

Signed-off-by: David Adams <david@adams.software>

**Motivation**

#2123 

**Summary**

Adds a small wrapper around web3 to allow grabbing emp state per address. Also adds truffle tests, so seperates test folder into
`/datasets` `/mocha` `/truffle` and adds yarn scripts for testing each.

**Issue(s)**

Fixes #2123

### DIFF
--- a/packages/affiliates/libs/contracts.js
+++ b/packages/affiliates/libs/contracts.js
@@ -1,7 +1,8 @@
 const ethers = require("ethers");
 const assert = require("assert");
+const { getAbi } = require("@uma/core");
 function DecodeLog(abi, meta = {}) {
-  assert(abi, "requries abi");
+  assert(abi, "requires abi");
   const iface = new ethers.utils.Interface(abi);
   return (log, props = {}) => {
     return {
@@ -12,7 +13,7 @@ function DecodeLog(abi, meta = {}) {
   };
 }
 function DecodeTransaction(abi, meta = {}) {
-  assert(abi, "requries abi");
+  assert(abi, "requires abi");
   const iface = new ethers.utils.Interface(abi);
   return (transaction, props = {}) => {
     return {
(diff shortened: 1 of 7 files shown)


nicholaspai (approved):
LGTM

packages/affiliates/libs/contracts.js
```diff
@@ -27,11 +28,76 @@ function decodeAttribution(
...
+  }
+  return {
+    decimals
+  };
+}
+
+// Wrapper for some basic emp functionality.
+// Currently we just need token and collateral info
+// Lookup by emp address
+function Emp({ abi = getAbi("ExpiringMultiParty"), web3 } = {}) {
```
mrice32, line 53: I'm assuming this class is used for many different EMPs at the same time? That's the reason for not parameterizing it with an address up-front?

Even so, this might be a little simpler to follow if emp instances were 1:1 with this class, just so you wouldn't have to pass address into each function. Not strongly opinionated (as I haven't seen the code that uses it), but it's something to consider.

packages/affiliates/libs/contracts.js
```diff
@@ -27,11 +28,76 @@ function decodeAttribution(
...
+    return {
+      address: tokenAddress,
+      decimals: await erc20.decimals(tokenAddress)
+    };
+  }
+  async function collateralInfo(address) {
+    const tokenAddress = await collateralCurrency(address);
+    return {
+      address: tokenAddress,
+      decimals: await erc20.decimals(tokenAddress)
```
mrice32, line 78: nit: since we have a lot of addresses floating around and method names that refer to other contracts, it may be a little more readable to call the `address` you pass into these methods `empAddress` just to differentiate it from other possible addresses.

mrice32 (approved):
Looks great! Two minor questions/comments

packages/affiliates/libs/contracts.js
```diff
@@ -1,7 +1,8 @@
 const ethers = require("ethers");
```
chrismaree, line 1: If we using web3.js and `getAbi` from `@uma/core` then I feel that we should not use `ethers` at all. I don't like muddling using the one library for somethings and the other for other things.

packages/affiliates/libs/contracts.js
```diff
@@ -27,11 +28,76 @@ function decodeAttribution(
...
+    contract.options.address = address;
+    return contract.methods.decimals().call();
+  }
+  return {
+    decimals
+  };
+}
+
+// Wrapper for some basic emp functionality.
+// Currently we just need token and collateral info
```
chrismaree, line 51: nit: you can make these comments longer. our styling guide prescribes 120 char line length so this does not need to be 3 lines

chrismaree (approved):
Looks good to me! just have one comment on using both `ethers` and `web3`.

packages/affiliates/libs/contracts.js
daywiss: yes thats right, i guess it could be designed either way, but its just instantiated once and used to gather info on several emps in a loop

packages/affiliates/libs/contracts.js
daywiss: yes theres a task to remove ethers, id like to do that as a seperate issue https://github.com/UMAprotocol/protocol/projects/6#card-48326707

PR #2139, opened 2020-11-02
feat(affiliates): add in entrypoint to run deployer rewards app

### LINKED ISSUE #2130: add cli script as entrypoint to call rewards calculator
We should be able to run the devmining as a command line script with minimal parameters. See the deployerRewards for the type of parameters it needs. This might be best split into 2 parts:
- create file for parsing cli params
- create a file to gather context such as token addresses, decimals from web, and then instanciate and call the deployer rewards library

daywiss:
**Motivation**

#2130 

**Summary**

This adds the "app" a cli for querying rewards ratios for users. Not really testable outside of a full integration test unfortunately, but the logic is fairly simple and it works with the example.config.js.

**Details**

- This does not write to a file, but outputs to console. You can easily output to file by piping using `>` or `>>` to the filename you would like. 
- this outputs the configuration input that generated the results as well as the reward results as 2 objects
- this relies on a configuration file rather than env vars to query the data. This is for easier creation/readability for user, and ease of parsing and standardization in the code.
- web3 must be configured with an env var for its provider address

**Output**
This took about 5-10 minutes to produce from around 1 month of data (9/20 - 10/19).   There is a def a path for optimization here with some careful thought. Output is showing a single emp deployer and is currently incorrect. Will need to fix this in a future PR.
image

**Issue(s)**

Fixes #2130


packages/affiliates/apps/DeployerRewards.js
```diff
@@ -0,0 +1,73 @@
...
+    config,
+    result
+  };
+}
+
+// Run the app and log output to console to allow user to pipe to file
+// This can be changed easily if need be.
+Start(App)
+  .then(console.log)
+  .catch(console.error);
```
mrice32, line 73: I'm a little confused around the architecture of building an "App" function and then passing it to another function that calls into it with additional args and returns a promise.

Is there a reason this architecture is necessary? It seems that we're just using it to pull in a user-provided config.

Could we, instead, just have a `config.js` file that offers a function called `getConfig()` to retrieve the config? Then we could just call that function in the top level function, right?

I could be misunderstanding the motivation, though.

packages/affiliates/libs/start.js
```diff
@@ -0,0 +1,19 @@
...
+const Web3 = require("web3");
+const Path = require("path");
+const [, , configPath] = process.argv;
+
+assert(configPath, "Requires relative path to config params");
+const config = require(Path.join(process.cwd(), configPath));
+assert(process.env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL");
+
+module.exports = async App => {
+  const web3 = new Web3(new Web3.providers.HttpProvider(process.env.CUSTOM_NODE_URL));
```
mrice32, line 17: nit: can we use `getWeb3` from https://github.com/UMAprotocol/protocol/blob/master/packages/common/src/ProviderUtils.js?

packages/affiliates/libs/start.js
```diff
@@ -0,0 +1,19 @@
...
+// example: see how its used in apps/DeployerRewards
+require("dotenv").config();
+const assert = require("assert");
+const Web3 = require("web3");
+const Path = require("path");
+const [, , configPath] = process.argv;
+
+assert(configPath, "Requires relative path to config params");
+const config = require(Path.join(process.cwd(), configPath));
+assert(process.env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL");
```
mrice32, line 14: nit: should we perform this assert logic inside the function so that this file won't error out if transitively included by something that doesn't end up calling this start function?

mrice32 (commented):
Looks good -- a few minor comments

packages/affiliates/apps/DeployerRewards.js
```diff
@@ -0,0 +1,73 @@
...
+
+  const rewards = DeployerRewards({
+    queries,
+    coingecko,
+    empAbi,
+    empCreatorAbi
+  });
+
+  // get emp info
+  const { tokensToPrice, tokenDecimals } = await Promise.reduce(
```
nicholaspai, line 41: This is a really nice chunk of code to look at, love me some `reduce`'s

packages/affiliates/libs/start.js
nicholaspai: Agreed, since `web3` is in the App's constructor this makes sense

packages/affiliates/apps/DeployerRewards.js
daywiss: you are right, i think its clearer the way you are suggesting. The main point is just to create a common way to parse configs and set up dependencies into the script.

packages/affiliates/libs/start.js
daywiss: I think i was using getWeb3 but found it confusing to reason about when doing testing without truffle and/or switching networks. I don't actually remember the exact reason. @chrismaree suggested I copy what he does in another script, though i am open to whatever you think is the best here.

mrice32: What issues did you have with it? Could we improve that function rather than using a custom solution? IIRC, the reason @chrismaree used this method was that we hadn't written `getWeb3()` yet.

The point of `getWeb3()` is to:
1. Mirror the way truffle resolved networks when not using truffle (i.e. still using our truffle config).
2. Detect when running within truffle/truffle test and just fall back to the truffle web3 in those cases.

chrismaree: I agree with @mrice32 that we should ideally use `getWeb3` everywhere. It's already configured in the `affiliates` script [here](https://github.com/UMAprotocol/protocol/blob/4bf7ae5006390c8030d62b887e6a71311d8871c4/packages/affiliates/libs/affiliates.js#L9). I think ideally the main entry point (be it this `start.js` or a set of unit tests) should inject their respective web3 object configured for the appropriate network.

packages/affiliates/config.example.js
```diff
@@ -0,0 +1,24 @@
...
+
+// emps to look at
+const empWhitelist = ["0xaBBee9fC7a882499162323EEB7BF6614193312e3", "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56"];
+
+// start time and endtime in ms timestamp
+const startTime = moment("9/20/2020 23:00:00", "MM/DD/YYYY  HH:mm z").valueOf(); // utc timestamp
+const endTime = moment("10/19/2020 23:00:00", "MM/DD/YYYY HH:mm z").valueOf();
+
+// 100 rewards to split across all emp creators
+const totalRewards = (100n * 10n ** 18n).toString();
```
chrismaree, line 13: should this be scaled like this to 1e18? I think the downstream scripts assume that it's not 18 decimal scaled.

packages/affiliates/config.example.js
```diff
@@ -0,0 +1,24 @@
+// This is an example file for how to run deployer rewards.  Use this as template to adjust parameters for your own run
+// example: node apps/DeployerRewards ./config.example.js >> output.json
+const moment = require("moment");
+
+// emps to look at
+const empWhitelist = ["0xaBBee9fC7a882499162323EEB7BF6614193312e3", "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56"];
+
+// start time and endtime in ms timestamp
+const startTime = moment("9/20/2020 23:00:00", "MM/DD/YYYY  HH:mm z").valueOf(); // utc timestamp
```
chrismaree, line 9: moment date-time formatting is different in other places in the repo. I think we should try stick to `"YYYY-MM-DD  HH:mm Z"`

packages/affiliates/config.example.js
```diff
@@ -0,0 +1,24 @@
...
+
+// start time and endtime in ms timestamp
+const startTime = moment("2020-9-20 23:00:00", "YYYY/MM/DD  HH:mm z").valueOf(); // utc timestamp
+const endTime = moment("2020-10-19 23:00:00", "YYYY/MM/DD HH:mm z").valueOf();
+
+// 10000 rewards to split across all emp creators
+const totalRewards = 10000n.toString();
+
+// network 1 = mainnet, 42 = kovan, will default to 1 if not specified
+const network = 1;
```
nicholaspai, line 16: The `empWhitelist` hard coded addresses would not work with kovan right?

nicholaspai (approved):

packages/affiliates/config.example.js
daywiss: yea the idea is that the user should create their own config file with the right values and the network they want to test.

PR #2146, opened 2020-11-03
feat(scripts): update UMIP collateral script

mrice32:
**Motivation**

Often we need to propose collateral currencies that do not have a standard number of decimals.

We also sometimes need to propose multiple collateral currencies in the same UMIP.

**Summary**

Updates the script to allow both of these use cases.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/collateral-umip/1_Propose.js
+++ b/packages/core/scripts/collateral-umip/1_Propose.js
@@ -2,7 +2,7 @@
 // fork of the main net or can be run directly on the main net to execute the upgrade transactions.
 // To run this on the localhost first fork main net into Ganache with the proposerWallet unlocked as follows:
 // ganache-cli --fork https://mainnet.infura.io/v3/5f56f0a4c8844c96a430fbd3d7993e39 --unlock 0x2bAaA41d155ad8a4126184950B31F50A1513cE25 --unlock 0x7a3a1c2de64f20eb5e916f40d11b01c441b2a8dc --port 9545
-// Then execute the script as: yarn truffle exec ./scripts/collateral-umip/1_Propose.js --network mainnet-fork --collateral 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 --fee 0.1 from core
+// Then execute the script as: yarn truffle exec ./scripts/collateral-umip/1_Propose.js --network mainnet-fork --collateral 0xc02aaa39b223fe8d0a0e5c4f27ead9083c756cc2 --fee 0.1 --collateral 0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 --fee 400 from core
 
 const AddressWhitelist = artifacts.require("AddressWhitelist");
 const Store = artifacts.require("Store");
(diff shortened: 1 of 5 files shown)


nicholaspai (approved):

packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -2,16 +2,37 @@
...
+  if (decimalsArg) {
+    console.log(`Using user input decimals: ${decimalsArg} for collateral ${collateralAddress}`);
+    return decimalsArg;
+  } else {
+    try {
+      const decimals = (await collateral.decimals()).toString();
+      console.log(`Using decimals returned by contract: ${decimals} for collateral ${collateralAddress}`);
+      return decimals;
+    } catch (error) {
+      throw "Must provide --decimals if token has no decimals function.";
```
daywiss, line 31: i like the intention of this function, and this is really just nit picking but this will end up obscuring the error for everything but the contract having no decimals function. May want to pass through the actual error message by appending error.message. like throw "Try providing --decimals to prevent the following error: " + error.message. 

Also id say wrap that in a `new Error()`

packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -20,40 +41,74 @@ async function runExport() {
...
-  await governor.propose(
-    [
-      {
-        to: store.address,
-        value: 0,
-        data: addFinalFeeToStoreTx
-      },
-      {
+  const argObjects = _.zipWith(collaterals, fees, decimals, (collateral, fee, numDecimalsArg) => {
+    return { collateral, fee, numDecimalsArg };
```
daywiss, line 53: this is cool, so it looks like you added ability to pass multiple collaterals in. could you show how to do this in the comment for this file? i think i could probably guess the right way but it prob should be documented

packages/core/scripts/collateral-umip/3_Verify.js
```diff
@@ -2,27 +2,60 @@
...
 const AddressWhitelist = artifacts.require("AddressWhitelist");
 const Store = artifacts.require("Store");
+const ERC20 = artifacts.require("ERC20");
+
+const { parseUnits } = require("@ethersproject/units");
+const _ = require("lodash");
 
 const argv = require("minimist")(process.argv.slice(), { string: ["collateral", "fee"] });
 
+async function getDecimals(collateralAddress, decimalsArg) {
```
daywiss, line 18: not really a suggestion for this PR but we probably could use some organizational thing where we can avoid code duplication. i usually end up throwing stuff like this in a utils.js file. not a blocker by curious on your thoughts here.

daywiss (commented):
nice improvements

packages/core/scripts/collateral-umip/1_Propose.js
mrice32: This should already be up there -- notice that the collateral and fees args are provided twice in the example. Is that explanatory enough or is there more context I could add?

packages/core/scripts/collateral-umip/3_Verify.js
mrice32: yeah, this small scale duplication is difficult to prevent since it would, in this case require adding an extra file to hold a function without a lot of context. Adding a utils.js file for good measure :).

mrice32: Done.

packages/core/scripts/collateral-umip/1_Propose.js
mrice32: Added a bit more context about how to apply the args below the example.

packages/core/scripts/collateral-umip/1_Propose.js
mrice32: +1. Done.

mrice32:
@daywiss responded to your comments and pushed changes. PTAL.

chrismaree (approved):
This is pretty useful and cool. I'm not a fan of having to add `lodash` and `ethersproject/units` to accommodate this functionality but it's not the end of the world & will be useful in other contexts.

Issue #2061, opened 2020-10-05
Build FundingRateStore


nicholaspai:
**Why**

Funding Rate Store must return a current funding rate for a specified `fundingRateIdentifier`, and might also provide logic for updating the funding rate in a priceless manner.

**How**

Proposed interface:
- `getFundingRateForIdentifier(bytes id)`: Returns the latest funding rate for a `fundingRateIdentifier`. If this contract needs to request a price to the DVM (i.e. not a MockFundingRateStore), then the `fundingRateIdentifier` must be whitelisted.

Proposed construction params:
- `timerAddress`: Address of testable timing contract.
- `finderAddress`: Points to DVM-related contracts. This contract will need to request "prices" for funding rates.
- `proposeLiveness`: amount of time until a proposed funding rate is eligible to be published as the current funding rate
- `proposerRewardRate`: the percent per unit time that rewards accumulate for an eventual funding rate reporter.

Proposed additional methods:
- `proposeFundingRate`: enqueues a pending funding rate and resets the `proposeLivenessTimer`, fails if there is already a pending funding rate. **This should fail if the proposed funding rate is the same as the current funding rate.**
- `publishFundingRate`: if pending funding rate liveness has expired, then replaces the current funding rate with the pending funding rate, deletes the pending funding rate, and pays the original proposer a reward. If there is no pending funding rate that has expired, but there is a settled disputed funding rate, then sets the result of the dispute as the current funding rate. 
- `disputeFundingRate`: triggers a DVM request to resolve whether a pending funding rate was valid. Removes pending funding rate from queue. Disputer must post a bond.
- `calculateFundingRate`: Helper method for DVM voters, can pass in some data to compute what the store would calculate as the funding rate based on the data.

**Open questions**
- How does the store attain a treasury of reserves from which it can reward proposers? 
- How do you determine the amount of rewards to pay successful funding rate proposers? 
    - Should the rewards accrue over time (since the last update)? 
    - Should the rewards be proportional to the delta between the proposed funding rate and the current funding rate?
- What happens if store doesn't have "enough" rewards to pay a proposer?
- Upgradeability: is it is as simple as deploying a new Store and changing the pointer in the Finder?

**Difficulty Score [1-10]**

This is a measure of the amount of developer time this task will require to be completed. Note: this should include design and planning time in addition to implementation time. The scale is not linear:

- 8: ~1 week

mrice32:
[...]

PR #2147, opened 2020-11-03
feat(funding-rate-store) Add dispute proposal logic

### LINKED ISSUE #2037: Funding Rate Store: Add funding rate dispute functionality
**Why**

Project Evergreen.

**Difficulty Score [1-10]**

- 6: ~1 day

nicholaspai:

**Motivation**

#2037
Ignoring proposal/dispute bonds, this PR adds the ability to dispute a pending proposal (and trigger a price request) and settle a dispute. The diff is relatively large because the FRS must now connect to DVM system contracts like the Oracle and the Store.

**Summary**

This is the updated funding rate proposal flow:
0. There is 1 "current funding rate" per identifier and by default it is 0.
1. There is 1 "pending funding rate" per identifier and by default there is none.
2. Anyone can "propose" a pending funding rate IFF there is not already an existing pending proposal. The proposer must post a bond that includes a final fee .
3. A pending funding rate does not overwrite the current funding rate until a liveness period passes from the time of proposal
4. Once the liveness expires for a pending funding rate, a pending rate will overwrite the current funding rate. The proposer will receive a rebate for the final fee  bond they posted.
5. During a pending proposal's liveness period, anyone can "dispute" the pending proposal. The disputer must post a bond that includes a final fee.
6. A "dispute" deletes the pending funding rate and triggers a DVM price request. This means that there can now be another funding rate proposed.
7. Once the DVM has resolved a dispute, the disputer or proposer (whoever wins) can claim their dispute rewards (including a rebate for their final fee bond)
8. A resolved dispute updates the current funding rate to the resolved price value. The resolved dispute has no affect on a pending funding rate if it exists.

**Details**

**Issue(s)**

Fixes #2037


packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +90,177 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+        delete fundingRateRecords[identifier].proposal;
+    }
+
+    /**
+     * @notice Settle a disputed funding rate. The winner of the dispute, either the disputer or the proposer,
+     * will receive a reward plus their final fee bond. This method will also overwrite the current funding rate
+     * with the resolved funding rate returned by the Oracle. Pending funding rates are unaffected by this method.
+     * @dev This will revert if there is no price available for the disputed funding rate.
+     */
+    function settleDispute(bytes32 identifier, uint256 proposalTime) external {
```
mrice32, line 186: @nicholaspai, I like this a lot.

There is no `msg.sender` in this method. That's super cool. Something we could do in the perp is have the liquidation withdrawal work the same way -- where all the payouts happen on one call without caring who's calling. That would actually reduce some of the weird edge cases and state we have to maintain.

Thoughts?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +90,177 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+            identifier,
+            fundingRateRecord.proposal.rate.rawValue,
+            fundingRateRecord.proposal.time,
+            fundingRateRecord.proposal.proposer,
+            msg.sender
+        );
+
+        // Delete pending proposal and copy into dispute records.
+        fundingRateRecord.proposal.disputer = msg.sender;
+        fundingRateDisputes[identifier][fundingRateRecord.proposal.time] = fundingRateRecord;
```
mrice32, line 176: Might be a weird question, but what if someone makes a proposal, disputes it, and then makes a second proposal and disputes it in the same txn? Would that break anything?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +90,177 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+            collateralCurrency.safeTransfer(
+                fundingRateDispute.proposal.proposer,
+                fundingRateDispute.proposal.finalFee.rawValue
+            );
         }
 
-        fundingRateRecord.proposal = Proposal({ rate: rate, time: getCurrentTime(), proposer: msg.sender });
+        // Update current rate to settlement rate.
+        fundingRateRecords[identifier].rate = settlementRate;
+        emit PublishedRate(identifier, settlementRate.rawValue, proposalTime, proposer);
```
mrice32, line 220: What if a proposal has come in since this dispute started?

mrice32 (commented):
Very good stuff! A few comments and questions

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -1,15 +1,41 @@
...
+import "../../../oracle/interfaces/FinderInterface.sol";
+import "../../../oracle/implementation/Constants.sol";
+
 import "../interfaces/FundingRateStoreInterface.sol";
 import "../../../common/implementation/Testable.sol";
+import "../../../common/implementation/FixedPoint.sol";
 
 
 contract FundingRateStore is FundingRateStoreInterface, Testable {
+    using SafeMath for int256;
```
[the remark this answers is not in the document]
nicholaspai: So just remove this you think?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: I was thinking the same thing as we write this. Basically the one who profits the most is incentivized probably to call this, which distributes all of the rewards. I don't actually know why we chose to do it in 3 separate payments before. Maybe we implement this in the PerpetualLiquidarable contract?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Proposer would post a final fee bond, disputer would post a final fee bond. 

Second proposal would post a final fee bond. 

Now assume that the second proposal expires before the dispute settles, then the second proposal will get published. Once the dispute settles it will again publish a new rate. 

I think we talked about this and I think i haven't implemented it to the spec, but should we have the settled dispute publish a new rate IFF there have been no published proposals since the dispute started?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: It just behaves like a normal proposal. Should we not allow settled disputes to update the funding rate if a normal proposal updates during the dispute?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: +1 yeah, we should add to liquidatable. Probably will reduce the bytecode tbh.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Well wouldn't the proposal that went through while this was being resolved be for a _later_ timestamp, and therefore, be more up to date?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Yeah, I see what you're saying, and that's definitely something worth discussing (I think we're covering a similar thing in the other thread). I'm talking about something a little lower level: won't the second dispute overwrite the first dispute in this data structure (same keys)? Would that cause any bad things to happen? Would the worst case just be that the person doing this would get their money locked up indefinitely? Are there any weird (positive EV) attack vectors that this overwriting would allow?

nicholaspai: Ah. The first dispute deletes the pending proposal. So the second dispute should not be possible. Do you see something diff?

mrice32: Yeah, the situation I'm talking about is the following steps in a single transaction:
1. Create proposal.
2. Dispute proposal.
3. Create proposal.
4. Dispute proposal.

Since they all have the same timestamp, would the second dispute overwrite the 1st? What are the consequences, if any, of that happening?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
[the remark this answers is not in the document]
nicholaspai: I'm going to correct this to `SafeMath for uint256` because we will need SafeMath when doing math on determining the last publish time. More on that in another comment

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: All in the same block I'm assuming? I think you're right the second dispute would overwrite the first dispute, and this one user would pay 4 final fees to earn back 1. The final fees from (1) and (2) would effectively be lost.

This makes me think of another bad outcome:
- Block 1:
    - User 1 creates a proposal
- Block 2:
    - User 2 disputes User 1's proposal
    - User 2 creates a proposal
    - User 2 disputes User 2's proposal

User 2 pays 3 final bonds to win back 1, but User 1 loses their funds. This would be uneconomical, but not good. WDYT?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Yep I agreed, and this is what we wrote up in the [spec](https://github.com/UMAprotocol/protocol/issues/2061): 

> publishFundingRate: if pending funding rate liveness has expired, then replaces the current funding rate with the pending funding rate, deletes the pending funding rate, and pays the original proposer a reward. If there is no pending funding rate that has expired, but there is a settled disputed funding rate, then sets the result of the dispute as the current funding rate.

I implemented this in this [commit](https://github.com/UMAprotocol/protocol/pull/2147/commits/9d9406265c4932eae0d3b655dfbbcea55df1a987)

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
         }
     }
 
+    /**
+     * @notice Returns the timestamp at which the current funding rate was published.
+     * @dev The "current funding rate" is defined as that returned by `getFundingRateForIdentifier(id)`
```
chrismaree, line 103: nit: missing param natspec

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+     */
+    function getLastPublishTimeForIdentifier(bytes32 identifier) external view returns (uint256) {
+        return _getLastPublishTimeForIdentifier(identifier);
+    }
+
+    /**
+     * @notice Propose a new funding rate for an identifier. A side effect of this method is that it will
+     * overwrite the current funding rate with a pending funding rate if its liveness has expired. If this update
+     * occurs, then this method will also pay the proposer their reward for successfully updating the current rate.
+     * @dev This will revert if there is already a pending funding rate for the identifier.
```
chrismaree, line 113: nit: missing param natspec

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+        });
+        emit ProposedRate(identifier, rate.rawValue, currentTime, msg.sender);
+
+        // Pull final fee bond from proposer.
+        collateralCurrency.safeTransferFrom(msg.sender, address(this), finalFeeBond.rawValue);
+    }
+
+    /**
+     * @notice Dispute a pending funding rate. This will delete the pending funding rate, meaning that a
+     * proposer can now proposer another rate with a fresh liveness.
```
chrismaree, line 165: nit: missing param natspec

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+     * @dev This will revert if there is no pending funding rate for the identifier.
+     * @dev Caller must approve this this contract to spend `finalFeeBond` amount of collateral, which they can
+     * receive back if their dispute is successful.
+     */
+    function dispute(bytes32 identifier) external {
+        FundingRateRecord storage fundingRateRecord = _getFundingRateRecord(identifier);
+        ProposalState proposalState = _getProposalState(fundingRateRecord.proposal);
+
+        require(proposalState == ProposalState.Pending, "Existing proposal not pending.");
```
chrismaree, line 174: nit
```suggestion
        require(proposalState == ProposalState.Pending, "Existing proposal not pending");
```

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+        delete fundingRateRecords[identifier].proposal;
+    }
+
+    /**
+     * @notice Settle a disputed funding rate. The winner of the dispute, either the disputer or the proposer,
+     * will receive a reward plus their final fee bond. This method will also overwrite the current funding rate
+     * with the resolved funding rate returned by the Oracle. Pending funding rates are unaffected by this method.
+     * @dev This will revert if there is no price available for the disputed funding rate.
+     */
+    function settleDispute(bytes32 identifier, uint256 proposalTime) external {
```
chrismaree, line 204: nit: missing param natspec

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
chrismaree: I dont see any issue with this as the disputer always ends up losing more final fees than what they get back, making this strictly uneconomical. It does spam the DVM but multiple queries of the price at the same timestamp for the same identifier will not be separate votes so the DVM does not care too much.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +98,193 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+            ? fundingRateDispute.proposal.disputer
+            : fundingRateDispute.proposal.proposer;
+
+        if (disputeSucceeded) {
+            // If dispute succeeds:
+            // - pay disputer the dispute reward + final fee rebate
+
+            collateralCurrency.safeTransfer(
+                fundingRateDispute.proposal.disputer,
+                fundingRateDispute.proposal.finalFee.rawValue
```
chrismaree, line 224: how is this sending the `finalFee + dispute reward`? I'm guessing this will come in later within the `fundingRateRecords` structure.

packages/core/test/financial-templates/funding-rate-store/FundingRateStore.js
```diff
@@ -1,74 +1,397 @@
...
+        proposeTime.add(toBN(liveness)).toString()
+      );
     });
 
     it("New proposal allowed", async function() {
-      await fundingRateStore.propose(identifier, { rawValue: toWei("-0.01") }, { from: account1 });
+      await fundingRateStore.propose(identifier, { rawValue: toWei("-0.01") }, { from: proposer });
 
       // Double check that existing value still persists even after a fresh proposal.
       assert.equal((await fundingRateStore.getFundingRateForIdentifier(identifier)).rawValue.toString(), toWei("0.01"));
```
chrismaree, line 397: these tests look good! Do you not want to also add a test that advances time past liveness and ensures that the `getFundingRateForIdentifier` returns the new funding rate?

chrismaree (approved):
This is great! nice clean code & tests. Good work @nicholaspai

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Done. FYI I only put natspec comments on the external methods to help reviewers but we should still do a PR cleaning up comments in a subsequent PR

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: There is no dispute reward yet as you guess. We would add it in here.

packages/core/test/financial-templates/funding-rate-store/FundingRateStore.js
nicholaspai: Technically the previous test tests for this already since it doesn't call `propose()` again. It only calls the view only method `getFundingRateForIdentifier` which returns the "updated" funding rate even though the `fundingRateRecords` struct hasn't been updated

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Created #2149 to track this

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
daywiss: interesting... this is really a question of if you should uniquely identify each dispute vs using timestamp as id.
perhaps you can require that `fundingRateDisputes[identifier][getCurrentTime()]` doesnt exist when adding a new proposal

daywiss (approved):

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: @daywiss That's a good idea. I agree with you the question is do we care enough to prevent this edge case? This one liner of `require(!fundingRateDisputes[identifier][getCurrentTime()])` in the proposal method would only serve to protect against single transactions containing multiple create+dispute proposal transactions.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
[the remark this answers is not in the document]
mrice32: SGTM

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Nice +1

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
daywiss: yea im not sure if it matters or not!

mrice32: It seems bad to have any scenario where storage is unwittingly overwritten and collateral gets locked up, although I agree that it doesn't seem to cause issues for anyone other than the caller.

Any reason not to disallow it out of principle/conservatism (maybe there's an attack vector we haven't thought of, like griefing a legit proposer?) the way David has suggested?

nicholaspai: Nope, I'll prevent it

nicholaspai: Using @daywiss 's recommended logic I've implemented this check along with a unit test in this [commit](https://github.com/UMAprotocol/protocol/pull/2147/commits/e29b8284d014d42e781bcfd268cd9b53a11e10e2). I think I've described and tested the issue here but not 100% sure.

Also @mrice32 I think your attack vector described below can be reduced to only steps 1,2,3 since the main risk is that the original proposer's bond is lost, and the second dispute doesn't really add anything to the "attack"

> 1. Create proposal.
> 2. Dispute proposal.
> 3. Create proposal.
> 4. Dispute proposal.

mrice32: @nicholaspai But in that case, wouldn't that original dispute end up getting resolved and the original proposal getting their bond back (if they were right)?

Either way, SGTM!

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -38,27 +107,210 @@ contract FundingRateStore is FundingRateStoreInterface, Testable {
...
+            );
+        }
+
+        // Update current rate to settlement rate if there has not been a published funding rate since the dispute
+        // began.
+        if (_getLastPublishTimeForIdentifier(identifier) <= fundingRateDispute.proposal.time) {
+            FundingRateRecord storage fundingRateRecord = _getFundingRateRecord(identifier);
+            uint256 currentTime = getCurrentTime();
+            fundingRateRecord.rate = settlementRate;
+            fundingRateRecord.publishTime = currentTime;
```
mrice32, line 268: Just curious -- why do we need publiishTime?

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: So we can check if there has been a newer update since the dispute started. If so, then the resolved price isn't published as the current rate

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Yeah isn't that by design?

packages/core/test/financial-templates/funding-rate-store/FundingRateStore.js
nicholaspai: @chrismaree WDYT?

chrismaree: fair enough. SGTM!

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Yep, all I'm saying is that you don't get the overwrite if you remove step 4, so I think no money is lost and everything works as expected. I think you need step 4 to have the contract do something outside of the spec.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Hmmm, but then it seems like we may be comparing apples and oranges, right?

Are we comparing the _proposal time_ to the _publish time_? Shouldn't we either compare proposal time to proposal time or publish time to publish time?

mrice32: I would think just comparing the _proposal time_ would make the most sense.

nicholaspai: Yeah that's more reasonable I agree

nicholaspai: But we still need a way to determine the time that the **current proposal** was proposed. So I'm thinking we just change `publishTime`'s name to `proposeTime` in the `FundingRateRecord` struct

mrice32: Good point. Yeah, that makes sense to me.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -230,10 +223,13 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
     function settleDispute(bytes32 identifier, uint256 proposalTime) external nonReentrant() {
         FundingRateRecord storage fundingRateDispute = _getFundingRateDispute(identifier, proposalTime);
 
+        // This should be an invariant, remove after testing
+        require(proposalTime == fundingRateDispute.proposal.time, "mismatching propose times");
```
nicholaspai: I removed this in the next PR

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Implemented in this [commit](https://github.com/UMAprotocol/protocol/pull/2147/commits/be1ae8b2dd53363033f7eb799703a41dd87e0df9)

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Ah yes you're right

nicholaspai: So do you agree that by blocking step 3, which is what David's suggestion does and what I implemented, prevents step 4 from being an issue?

nicholaspai: I don't really see a way to block step 4 explicitly unless we create some sort of way to store multiple disputes per <proposal-time, identifier>, which I think is just a bit more complex.

mrice32 (approved):
LGTM!! Very awesome stuff

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: As discussed offline, 100% agree with your conclusion here @nicholaspai. Blocking 3 seems to be the simplest solution.

PR #2148, opened 2020-11-03
feat(pricefeed): adds a new price feed for the GASETH-1M identifier

smb2796:
**Motivation**

This PR adds the ability for liquidation and dispute bots to get monthly median gas prices. This is done by querying Google BigQuery and using the query response as a price feed.

**Summary**

To run this price feed locally, you need to set a `GOOGLE_APPLICATIONS_CREDENTIALS` environment variable with GCP admin or BigQuery permissions.

EMP_ADDRESS also needs to be set to the uGAS Kovan contract address: `0xE57DcDf729E851d68a1B8b71E84BA43B1071456b` 

1. `@google-cloud/bigquery` was added to interact with bigquery in node
2. [Moment.js](https://www.npmjs.com/package/moment) package was added to change dates into acceptable UTC BigQuery format.
3. The option to use a BigQuery price feed, when using the GASETH-1M identifier, was added to `CreatePriceFeed.js`.
4. `BigQueryPriceFeed.js` was added to return the current or historical price for GASETH-1M.
5. `update()` uses [(current time - one month), (current time - five minutes)] UTC as the time interval to parameterize the BQ query with. `getHistoricalPrice()` receives a liquidation time and uses the liquidation time to construct its time interval.
6. `update()` and `getHistoricalPrice()` use `runQuery` to create a BQ api call.

**Details**

**Issue(s)**

No associated issue.

### DIFF
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -4,6 +4,7 @@
   "description": "Arbitrage automation and libraries for UMA financial templates",
   "dependencies": {
     "@ethersproject/bignumber": "^5.0.5",
+    "@google-cloud/bigquery": "^5.3.0",
     "@google-cloud/logging-winston": "^3.0.6",
     "@google-cloud/trace-agent": "^4.2.5",
     "@uma/common": "^1.0.1",
@@ -11,7 +12,9 @@
     "@uniswap/sdk": "^2.0.5",
     "bluebird": "^3.7.2",
     "dotenv": "^6.2.0",
+    "highland": "^2.13.5",
     "minimist": "^1.2.0",
+    "moment": "^2.29.1",
     "node-fetch": "^2.6.0",
     "node-pagerduty": "^1.2.0",
     "winston": "^3.2.1",
--- a/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
@@ -19,7 +19,7 @@ class BigQueryPriceFeed extends PriceFeedInterface {
    *      this number of seconds has passed, it will be a no-op.
    * @param {Number} decimals Number of decimals to use to convert price to wei.
    */
-  constructor(logger, web3, lookback, getTime, minTimeBetweenUpdates, decimals = 0) {
+  constructor(logger, web3, lookback, getTime, minTimeBetweenUpdates, decimals = 18) {
     super();
     this.logger = logger;
     this.web3 = web3;
@@ -30,7 +30,7 @@ class BigQueryPriceFeed extends PriceFeedInterface {
 
     this.convertDecimals = number => {
       // Converts price result to wei and returns price conversion to correct decimals as a big number.
-      return this.toBN(parseFixed(number.toString(), decimals).toString());
+      return this.toBN(parseFixed(number.toString(), decimals - 18).toString());
     };
   }
 
--- a/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
@@ -57,17 +57,13 @@ class BigQueryPriceFeed extends PriceFeedInterface {
     // Create the query with the needed time interval.
     const query = this.createQuery(t2, formattedCurrentTime);
 
-    // Submit async call to BigQuery.
+    // Submit async call to BigQuery and check the response.
     let priceResponse;
-    await this.runQuery(query)
-      .then(res => {
-        console.log(res[0].gas_price);
-        priceResponse = res[0].gas_price;
-      })
-      .catch(console.log);
-
-    // Check response.
-    if (!priceResponse) {
+    try {
+      priceResponse = await this.runQuery(query);
+      priceResponse = priceResponse[0].gas_price;
+      console.log(priceResponse);
+    } catch (error) {
       throw new Error(`🚨Could not parse price result from bigquery: ${priceResponse}`);
     }
 
@@ -118,17 +114,13 @@ class BigQueryPriceFeed extends PriceFeedInterface {
     // Create the query with the needed time interval.
     const query = this.createQuery(t2, formattedCurrentTime);
 
-    // Submit async call to BigQuery.
+    // Submit async call to BigQuery and check response.
     let priceResponse;
-    await this.runQuery(query)
-      .then(res => {
-        console.log(res[0].gas_price);
-        priceResponse = res[0].gas_price;
-      })
-      .catch(console.log);
-
-    // Check responses.
-    if (!priceResponse) {
+    try {
+      priceResponse = await this.runQuery(query);
+      priceResponse = priceResponse[0].gas_price;
+      console.log(priceResponse);
+    } catch (error) {
       throw new Error(`🚨Could not parse price result from bigquery: ${priceResponse}`);
     }
 
--- a/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
@@ -5,8 +5,6 @@ const highland = require("highland");
 const { parseFixed } = require("@ethersproject/bignumber");
 const { createQuery } = require("../queries/GasEthQuery");
 
-const client = new BigQuery();
-
 // An implementation of PriceFeedInterface that uses BigQuery to retrieve prices.
 class BigQueryPriceFeed extends PriceFeedInterface {
   /**
@@ -28,11 +26,17 @@ class BigQueryPriceFeed extends PriceFeedInterface {
     this.minTimeBetweenUpdates = minTimeBetweenUpdates;
     this.toBN = this.web3.utils.toBN;
     this.dateConversionString = "YYYY-MM-DD HH:mm:ss";
+    this.decimals = decimals;
+    this.client = new BigQuery();
 
-    this.convertDecimals = number => {
-      // Converts price result to wei and returns price conversion to correct decimals as a big number.
-      return this.toBN(parseFixed(number.toString(), decimals - 18).toString());
-    };
+    // this.convertDecimals = number => {
+    //   // Converts price result to wei and returns price conversion to correct decimals as a big number.
+    //   return this.toBN(parseFixed(number.toString(), decimals - 18).toString());
+    // };
+  }
+
+  convertDecimals(number) {
+    return this.toBN(parseFixed(number.toString(), this.decimals - 18).toString());
   }
 
   getCurrentPrice() {
@@ -134,7 +138,7 @@ class BigQueryPriceFeed extends PriceFeedInterface {
   }
   async runQuery(query) {
     // returns a node read stream
-    const stream = await client.createQueryStream({ query });
+    const stream = await this.client.createQueryStream({ query });
     // highland wraps a stream and adds utilities simlar to lodash
     // https://caolan.github.io/highland/
     return (
--- a/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
@@ -3,6 +3,7 @@ const { BigQuery } = require("@google-cloud/bigquery");
 const moment = require("moment");
 const highland = require("highland");
 const { parseFixed } = require("@ethersproject/bignumber");
+const { createQuery } = require("../queries/GasEthQuery");
 
 const client = new BigQuery();
 
@@ -55,7 +56,7 @@ class BigQueryPriceFeed extends PriceFeedInterface {
       .format("YYYY-MM-DD HH:mm:ss");
 
     // Create the query with the needed time interval.
-    const query = this.createQuery(t2, formattedCurrentTime);
+    const query = createQuery(t2, formattedCurrentTime);
 
     // Submit async call to BigQuery and check the response.
     let priceResponse;
@@ -112,7 +113,7 @@ class BigQueryPriceFeed extends PriceFeedInterface {
     });
 
     // Create the query with the needed time interval.
-    const query = this.createQuery(t2, formattedCurrentTime);
+    const query = createQuery(t2, formattedCurrentTime);
 
     // Submit async call to BigQuery and check response.
     let priceResponse;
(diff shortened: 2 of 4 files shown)


chrismaree (commented):
This is pretty awesome! I really like that we can run big queries directly from the bots 🤩. 

I had a few nits on styling and syntax usage. 

Also, there are no tests for this. What should the testing strategy be? It might be useful to be able to refactor out the `query` logic that gets returned from bigquery into a separate class and then Mock the response, in the same way, we do with the other price feeds. Then, we can write unit tests against this implementation. What do you think?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,218 @@
...
+   */
+  constructor(logger, web3, lookback, getTime, minTimeBetweenUpdates, decimals = 0) {
+    super();
+    this.logger = logger;
+    this.web3 = web3;
+    this.lookback = lookback;
+    this.getTime = getTime;
+    this.minTimeBetweenUpdates = minTimeBetweenUpdates;
+    this.toBN = this.web3.utils.toBN;
+
```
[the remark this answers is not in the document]
chrismaree: Actually, this comment is not relevant and the way you've done it here matches to the `CryptoWatchPriceFeed.js` implementation as the `decimals` is not needed anywhere else within the price feed. You can disregard my above comment.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,218 @@
...
+   * @notice Constructs the CryptoWatchPriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Function} getTime Returns the current time.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Number} decimals Number of decimals to use to convert price to wei.
+   */
+  constructor(logger, web3, lookback, getTime, minTimeBetweenUpdates, decimals = 0) {
```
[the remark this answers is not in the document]
smb2796: Yes, this was my intention. The value is already returned in Wei from BigQuery, so it is kept in the same notation.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,218 @@
...
+const moment = require("moment");
+const highland = require("highland");
+const { parseFixed } = require("@ethersproject/bignumber");
+
+const client = new BigQuery();
+
+// An implementation of PriceFeedInterface that uses CryptoWatch to retrieve prices.
+class BigQueryPriceFeed extends PriceFeedInterface {
+  /**
+   * @notice Constructs the CryptoWatchPriceFeed.
```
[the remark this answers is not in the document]
smb2796: Thanks for pointing this out. I had some outdated comments in other places as well, so those were fixed.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,218 @@
...
+    // Using current time minus 5 minutes as end-bound of query. This is because the BQ dataset lags ~5 minutes
+    // Using moment package to changeto UTC and create acceptable format for BQ
+    let currentTime = new Date();
+    currentTime = currentTime - 300000;
+    const formattedCurrentTime = moment(currentTime)
+      .utc()
+      .format("YYYY-MM-DD HH:mm:ss");
+
+    // 2592000000 is a month time interval
+    let t2 = new Date(currentTime - 2592000000);
```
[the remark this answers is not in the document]
smb2796: I agree that using moment data manipulation is much cleaner. I changed the code to use `subtract()`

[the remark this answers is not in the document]
smb2796: And yes it does force the use of the 1-month price identifier. Since this was the pid that we are using for our contract, I focused on only the 1m, but happy to try and generalize this.

packages/financial-templates-lib/package.json
```diff
@@ -4,14 +4,17 @@
...
     "@google-cloud/logging-winston": "^3.0.6",
     "@google-cloud/trace-agent": "^4.2.5",
     "@uma/common": "^1.0.1",
     "@uma/core": "^1.1.0",
     "@uniswap/sdk": "^2.0.5",
     "bluebird": "^3.7.2",
     "dotenv": "^6.2.0",
+    "highland": "^2.13.5",
     "minimist": "^1.2.0",
+    "moment": "^2.29.1",
```
nicholaspai, line 17: `moment` is also imported in `core`, should we move this to the main package.json in `protocol`?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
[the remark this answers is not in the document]
nicholaspai: Isn't this inconsistent with how we define a default `decimals = 18` throughout the repo? Example in [`ExpiringMultiPartyClient.js`](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js#L16)

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,219 @@
...
+
+  async getHistoricalPrice(time) {
+    if (this.lastUpdateTime === undefined) {
+      return undefined;
+    }
+
+    // Using moment package to changeto UTC and create acceptable format for BQ.
+    const formattedCurrentTime = moment(time)
+      .utc()
+      .format("YYYY-MM-DD HH:mm:ss");
```
nicholaspai, line 48: Since you use this format string several times in this file I'd define a file constant so you minimize hard-coded values

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,219 @@
...
+
+    // 2592000000 is a month time interval
```
nicholaspai, line 50: Do you mean 2592000000 = 30 * 24 * 60 * 60 seconds?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,219 @@
...
+    let t2 = new Date(time);
```
nicholaspai, line 51: What is `t2`? Could this be given a more descriptive name?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,219 @@
...
+    const query = this.createQuery(t2, formattedCurrentTime);
+
+    // Submit async call to BigQuery.
+    let priceResponse;
+    await this.runQuery(query)
+      .then(res => {
+        console.log(res[0].gas_price);
+        priceResponse = res[0].gas_price;
+      })
+      .catch(console.log);
```
nicholaspai, line 67: Since you throw an error if `priceResponse` is not filled out, because `runQuery` failed, why not just throw the below error in this `catch`?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,218 @@
...
+      lastUpdateTimestamp: this.lastUpdateTime
+    });
+
+    // Create the query with the needed time interval
+    const query = this.createQuery(t2, formattedCurrentTime);
+
+    // Submit async call to BigQuery
+    let priceResponse;
+    await this.runQuery(query)
+      .then(res => {
```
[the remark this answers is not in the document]
nicholaspai: Agreed. I'd wrap the `runQuery` call inside of a `try...catch` and then throw the below error in the `catch`

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,219 @@
...
+        .collect()
+        // emit the stream as a promise when the stream ends
+        // this is the start of a data pipeline so you can imagine
+        // this could also "pipe" into some other processing pipeline or write to a file
+        .toPromise(Promise)
+    );
+  }
+  // This is a helper method to create a GASETH BQ query with time arguments.
+  createQuery(t2, formattedCurrentTime) {
+    const query = `
```
nicholaspai, line 159: Should we place queries in an external Query.js file? This seems like something we might want to tweak in the future so it might be easier to find in another file

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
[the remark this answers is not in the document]
smb2796: I based it off of the example in `CryptoWatchPriceFeed.js`. Which part is inconsistent? How it is defined or defining the amount of decimals as 0?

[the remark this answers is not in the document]
nicholaspai: Well I see `CryptoWatchPriceFeed.js` has a default set of 18, seen [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L34). Typically when talking about decimal precision in web3, "decimals" refers to the number of wei that 1 unit of the asset is worth. For example, ETH has a precision of 18 meaning that 1e18 wei are equal to 1 ETH.

So, the CryptoWatchPriceFeed returns prices using 18 decimals of precision, meaning that if it wants to report a price of "0.5" it actually returns "0.5e18". The actualy CryptoWatch API would return the price as "0.5" so the `CryptoWatch` class will convert 0.5 -> 0.5e18.

So, if BigQuery returns values as wei, then maybe you should can use `(decimals - 18)` as your precision. For example, BQ returns "9e18" and the user wants prices using 8 decimals of precision, then you convert that into:
```
bq_price = 9e18
decimals = 8
converted_price = bq_price ^ (decimals - 18)
converted_price = bq_price ^ (8 - 18)
converted_price = 9e18 ^ (-10)
converted_price = 9e8
```

By setting `decimals=18` this formula would just return the same price that BQ does.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
mrice32: +1

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: Agreed - renamed to `earlierTimeBound`

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: Yes, definitely confusing. Changed to use the `moment` calculation so I updated the comment.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: Good point - set a dateConversionString to "YYYY-MM-DD HH:mm:ss"

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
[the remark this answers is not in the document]
smb2796: Updated to use a `try...catch`

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: Good call - moved the error to the `catch`.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: Created a separate file in a new folder `../queries/GasEthQuery.js`.

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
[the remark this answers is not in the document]
smb2796: I think this makes sense: 

`this.convertDecimals = number => {
      // Converts price result to wei and returns price conversion to correct decimals as a big number.
      return this.toBN(parseFixed(number.toString(), decimals - 18).toString());
    };`

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,155 @@
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const { BigQuery } = require("@google-cloud/bigquery");
+const moment = require("moment");
+const highland = require("highland");
+const { parseFixed } = require("@ethersproject/bignumber");
+const { createQuery } = require("../queries/GasEthQuery");
+
+const client = new BigQuery();
```
mrice32, line 8: nit: why not initialize this in the constructor?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,155 @@
...
+    this.lookback = lookback;
+    this.getTime = getTime;
+    this.minTimeBetweenUpdates = minTimeBetweenUpdates;
+    this.toBN = this.web3.utils.toBN;
+    this.dateConversionString = "YYYY-MM-DD HH:mm:ss";
+
+    this.convertDecimals = number => {
+      // Converts price result to wei and returns price conversion to correct decimals as a big number.
+      return this.toBN(parseFixed(number.toString(), decimals - 18).toString());
+    };
```
mrice32, line 35: nit: can this just be a normal class function that references `this.decimals`?

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
```diff
@@ -0,0 +1,155 @@
...
+const { BigQuery } = require("@google-cloud/bigquery");
+const moment = require("moment");
+const highland = require("highland");
+const { parseFixed } = require("@ethersproject/bignumber");
+const { createQuery } = require("../queries/GasEthQuery");
+
+const client = new BigQuery();
+
+// An implementation of PriceFeedInterface that uses BigQuery to retrieve prices.
+class BigQueryPriceFeed extends PriceFeedInterface {
```
mrice32, line 11: High-level comment: do we have a strategy for testing this in any way? Not necessarily just concerned about the query itself, but also making sure that this class is bug-free.

mrice32 (commented):
Looks great! a few comments

packages/financial-templates-lib/src/price-feed/BigQueryPriceFeed.js
smb2796: I'm working on writing a test similar to the tests for the other price feeds. Should be in shortly!

smb2796:
Closing for now due to this no longer being necessary for the uGAS bots.

PR #2150, opened 2020-11-04
feat(developer-mining): calculate contract TVL using bot's pricefeed and collateral value

### LINKED ISSUE #2125: convert emp value calculation to use collateral price / gcr
The current implementation of the devming calculator requires that the synthetic token is listed on Coingecko to calculate the total value of token debt within a financial contract. This places an unnecessary burden on contract creators.

This burden can be mitigated by rather calculating the token debt as a function of collateral within a EMP & the GCR to work out the total USD amount of debt within a contract. This modification would require that the collateral type is listed on coingecko, rather than the synthetic itself.

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2125

Updating devmining TVL computation to use use a combination of collateral price and synthetic token debt price.

**Summary**

Before this PR, EMP TVL was computed by using a coingecko price feed for each synthetic. This placed an undue requirement on developers and created unreasonable assumptions.

This PR changes this by enabling the computation of TVL without the need for a synthetic token coingecko price feed. This is done through the use of the liquidator/disputer bots price feeds to compute the value of each synthetic in the collateral currency in conjunction with a coingecko price feed to find collateral prices. In conjunction, these two inputs can be used to find the USD amount of each synthetic with the only requirements being:
1) collateral currency is listed on coin-gecko
2) the bots `CreatePriceFeed.js` has a default price feed for the emp

The second requirement here is a bit restrictive as it assumes that the bots always have a default config for each and every emp that we want to support for devmining. I think that this is acceptable as each partner that we support through devmining will also have a bot run by us in the liquidator of last resort program. Ideally, these default configs should be refactored out to be placed within a separate config file, as documented in this issue https://github.com/UMAprotocol/protocol/issues/2154

**Details**

This PR adds in a few additional data files as well. Specifically, the previous data files for coingecko price feeds for each synthetic were removed and replaced with price data for collateral. Additionally, the PR adds in price histories for synthetics, denominated in the collateral currency.

This PR also changes the sample `set1.js` file by adding in an additional EMP which uses PERL as collateral.

**Issue(s)**

close #2125

**How to review this PR**
Skip over the data sets and look at the code changes :)


packages/affiliates/libs/coingecko.js
```diff
@@ -7,22 +7,24 @@ module.exports = (host = "https://api.coingecko.com") => {
     return result.data;
   }
 
-  // fetch historic prices for a `contract` denominated in `currency` between timestamp `from` and `to`. Note timestamps
+  // Fetch historic prices for a `contract` denominated in `currency` between timestamp `from` and `to`. Note timestamps
   // are assumed to be js timestamps and are converted to unixtimestamps by dividing by 1000.
-  function chart(contract, currency, from, to) {
+  async function getHistoricContractPrices(contract, currency, from, to) {
```
chrismaree, line 12: This function name was changed to be more descriptive as well as changing the return type. now the function only returns the prices. Previously it would return exchange volume + total market cap which are not useful for any downstream scripts.

packages/affiliates/libs/synthPrices.js
```diff
@@ -0,0 +1,23 @@
+const { createReferencePriceFeedForEmp, Networker } = require("@uma/financial-templates-lib");
```
chrismaree, line 1: this is the new price feed for developer mining. it returns the value of a synthetic, denominated in collateral currency, for a given `empAddress` between `from` and `to`.

packages/affiliates/test/datasets/set1.js
```diff
@@ -2,9 +2,22 @@
 // use this in scripts that start with "write-" to generate a dateset
 const moment = require("moment");
 const empCreator = "0x9A077D4fCf7B26a0514Baa4cff0B481e9c35CE87";
-const empContracts = ["0xaBBee9fC7a882499162323EEB7BF6614193312e3", "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56"];
-const syntheticTokens = ["0xF06DdacF71e2992E2122A1a0168C6967aFdf63ce", "0xD16c79c8A39D44B2F3eB45D2019cd6A42B03E2A9"];
-const syntheticTokenDecimals = [18, 18];
+
+const empContracts = [
```
chrismaree, line 6: I added an extra EMP here to test the script.

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -97,12 +119,17 @@ describe("DeployerRewards", function() {
...
+      collateralTokens: collateralTokens,
+      collateralTokenDecimals: collateralTokenDecimals,
+      syntheticTokenDecimals: syntheticTokenDecimals
     });
-    // WIP
 
-    assert.equal(Object.keys(result).length, 1); // there should only be 1 deployer
-    assert.equal(Number(Object.values(result)[0]), Number(devRewardsToDistribute)); // the total rewards distributed should equal the number specified
+    assert.equal(Object.keys(result).length, 2); // There should be 2 deplorers for the 3 EMPs.
+    assert.equal(
```
chrismaree, line 128: there are now two contract deployers for the 3 emps. check the sum adds up as expected.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -295,6 +296,13 @@ const defaultConfigs = {
       { type: "cryptowatch", exchange: "binance", pair: "btcusdt" },
       { type: "cryptowatch", exchange: "bitstamp", pair: "btcusd" }
     ]
+  },
+  USDPERL: {
```
chrismaree, line 300: Adds the USDPERL price feed to enable the identifier to be queried to get historic prices. For now, this is simply in this file but in future, these configs should be refactored out to a separate file. I created an issue to track this: https://github.com/UMAprotocol/protocol/issues/2154

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -31,7 +32,8 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     getTime,
     minTimeBetweenUpdates,
     invertPrice,
-    decimals = 18
+    decimals = 18,
+    ohlcPeriod = 60 // One minute is CryptoWatch's most granular option.
```
chrismaree, line 36: Changed the price feed to make this configurable. by default still uses 1 min resolution but can be tuned to accommodate any different duration. for example, developer mining does not need more precision than once per hour(even then that's very granular).

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -117,6 +118,18 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     return returnPrice;
   }
 
+  getHistoricalPricePeriods() {
```
chrismaree, line 121: this method simply exposes data that was previously internal to the price feed. it's useful for things like developer mining to get all historic data.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -156,7 +169,8 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     // See https://docs.cryptowat.ch/rest-api/markets/ohlc for how this url is constructed.
     const ohlcUrl = [
       `https://api.cryptowat.ch/markets/${this.exchange}/${this.pair}/ohlc`,
-      `?after=${earliestHistoricalTimestamp}`,
+      `?before=${currentTime}`,
+      `&after=${earliestHistoricalTimestamp}`,
```
chrismaree, line 173: I changed the restful query that gets sent to crypto watch to include the `before` field. Previously the query would return all data points up to the present. Now, we can define the upper time limit on prices returned.

For the operation of the normal bots, this will make no difference as the `before` is the `currentTime`.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -169,14 +183,10 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     // 3. Check responses.
     if (!priceResponse || !priceResponse.result || !priceResponse.result.price) {
-      // TODO: Slack transport might not be able to handle long messages such as the stringified `ohlcResponse`,
```
chrismaree, line 172: Unrelated to PR but removed as addressed in PR: https://github.com/UMAprotocol/protocol/pull/2124

packages/affiliates/test/mocha/models.js
```diff
@@ -71,15 +72,46 @@ describe("History", function() {
...
+
+describe("Synthetic prices", function() {
+  let prices, seed;
+  const empAddress = "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56";
+  const startingTimestamp = moment("2020-09-23 23:00:00", "YYYY-MM-DD  HH:mm Z").valueOf();
+  const endingTimestamp = moment("2020-10-05 23:00:00", "YYYY-MM-DD  HH:mm Z").valueOf();
+  it("init", async function() {
+    this.timeout(100000);
+    const Web3 = require("web3");
+    const web3 = new Web3(new Web3.providers.HttpProvider(process.env.CUSTOM_NODE_URL));
```
chrismaree, line 112: I'm not sure what the best way is to have this in the tests? The `SyntheticPrices` constructor requires a web3 instance connected to mainnet. Thoughts?

packages/affiliates/libs/synthPrices.js
nicholaspai: I really like that we're re-using the financial-templates-lib SDK here

nicholaspai (approved):
I think its great that you're re-using the price feed classes from `financial-templates-lib`. Its also clever to make the assumption that we can calculate the historical synthetic prices if and only if the liquidator bot has access to the same synthetic price. 

This assumption is pretty strong because we shouldn't be supporting any financial contracts that don't have bots monitoring them.

packages/affiliates/apps/DeployerRewards.js
```diff
@@ -31,35 +32,41 @@ async function App(config) {
...
     async (result, address) => {
       // switch this to tokenInfo if you want to base prices off tokens
-      const info = await emp.collateralInfo(address);
-      result.tokensToPrice.push(info.address);
-      result.tokenDecimals.push(info.decimals);
+      const collateralInfo = await emp.collateralInfo(address);
+      result.collateralTokens.push(collateralInfo.address);
+      result.collateralTokenDecimals.push(collateralInfo.decimals);
+
+      const syntheticInfo = await emp.collateralInfo(address);
```
daywiss, line 54: this should probably be emp.tokenInfo

packages/affiliates/libs/coingecko.js
daywiss: could probably just add a new function rather than removing the chart function, ie compose them by having getHistoricalPrices call chart, and then checking for .prices and returning that. This is mainly to not break compatibility if some thing is relying on chart to exist. Not a big deal though at this stage.

packages/affiliates/libs/synthPrices.js
```diff
@@ -0,0 +1,23 @@
...
+const winston = require("winston");
+
+module.exports = ({ web3 } = {}) => {
+  // Fetch historic synthetic prices for a given `empAddress` between timestamps `from` and `to.
+  // Note timestamps are assumed to be js timestamps and are converted to unixtimestamps by dividing by 1000.
+  async function getHistoricSynthPrices(empAddress, from, to) {
+    from = from / 1000;
+    to = to / 1000;
+    const priceFeed = await createReferencePriceFeedForEmp(
+      winston.createLogger({ transports: [new winston.transports.Console()] }),
```
daywiss, line 11: we might want to move instantiation of this into the app or even at the "config" level and pass it down

packages/affiliates/test/mocha/affiliates.js
daywiss: awesome yea im glad you did this

packages/affiliates/test/mocha/models.js
daywiss: yea this kind of sucks. Either you create a truffle test, which I think one of the previous prs adds test/truffle and then mock everything out to a local chain, or we can add it to e2e tests which dont get run in CI at the moment and just run locally. Imo its probably good enough to just run it locally to verify.  Not sure of what else we can do here without refactoring the underlying module.

packages/affiliates/apps/DeployerRewards.js
chrismaree: ye, you 100% right. Update updated accordingly

packages/affiliates/libs/synthPrices.js
chrismaree: For sure. That makes sense. for now, I didn't want to couple anything else to Winston as I'm really keen to refactor this price feed lib away from being dependent on Winston and using the emit pattern you introduced a few PRs ago. I felt that this would be the cleanest way for now.

packages/affiliates/test/mocha/models.js
chrismaree: I've addresses this by leveraging the `getWeb3` function to generate a web3 instance. I also modified the `package.json` to run the `mocks tests` with a `--network mainnet_mnemonic` such that the `getWeb3` will connect to main net by default. This test file (`models.js`) pulls main net info for this test as well as `Coingecko().getHistoricContractPrices` to validate the price feed data so its fine that it's connecting to main net.

daywiss (approved):

daywiss (approved):
looks good

PR #2152, opened 2020-11-04
refactor(perp-liquidatable) Withdraw dispute/liquidation rewards in one transaction instead of multiple

### LINKED ISSUE #2149: Liquidatable: Dispurse dispute rewards in one transaction
**Why**

Currently the sponsor, liquidator, and disputer must withdraw their own rewards individually. We should just dispurse them on the first `withdrawLiquidatorRewards` call and anyone can call this. This might also reduce bytecode by eliminating local variables.

**Resolution**

This will impact unit tests and more importantly it might change the ABI for the following methods and events:
- `withdrawLiquidation(uint256 liquidationId, address sponsor) returns returns (FixedPoint.Unsigned memory amountWithdrawn)`
- `event LiquidationWithdrawn(
        address indexed caller,
        uint256 withdrawalAmount,
        Status indexed liquidationStatus,
        uint256 settlementPrice
    )`

On the first point, `withdrawLiquidation` should return a different amount, perhaps either the total amount withdrawn, or a tuple containing the three different reward payouts. The event `LiquidationWithdrawn` should also be modified to include more detailed breakdown of the payouts.

This will mean that bot code that uses the updated ABI's will break.

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:
**Motivation**

#2149 

**Summary**

Anyone can call `withdrawLiquidation` now, although it will revert if there are no rewards to disperse. It will pay out rewards owed to the liquidator, sponsor, and disputer.

The `LiquidationWithdrawn` event is also changed to include more details about all of the payments. Previously it was only emitted to display data about the caller's withdrawal.

This reduces the contract bytecode by ~400 bytes because it gets rid of an if-else branch, previously determining whether the caller was the sponsor, liquidator, or disputer.

**Issue(s)**

Fixes #2149

### DIFF
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
@@ -240,6 +240,7 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         PositionData storage positionToLiquidate = _getPositionData(sponsor);
 
         tokensLiquidated = FixedPoint.min(maxTokensToLiquidate, positionToLiquidate.tokensOutstanding);
+        require(tokensLiquidated.isGreaterThan(0), "Liquidating 0 tokens");
 
         // Starting values for the Position being liquidated. If withdrawal request amount is > position's collateral,
         // then set this to 0, otherwise set it to (startCollateral - withdrawal request amount).
@@ -485,13 +486,6 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
 
             collateralCurrency.safeTransfer(liquidation.liquidator, rewards.paidToLiquidator.rawValue);
         }
-
-        // If the raw amounts of collateral to pay out are 0, which is possible if the liquidation is for 0 collateral,
-        // then revert because there are no rewards to pay.
-        require(
-            (rewards.payToLiquidator.add(rewards.payToSponsor).add(rewards.payToDisputer)).isGreaterThan(0),
-            "Rewards equal to 0"
-        );
 
         emit LiquidationWithdrawn(
             msg.sender,
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
@@ -406,6 +406,7 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
      * This method will revert if rewards have already been dispersed.
      * @param liquidationId uniquely identifies the sponsor's liquidation.
      * @param sponsor address of the sponsor associated with the liquidation.
+     * @return data about rewards paid out.
      */
     function withdrawLiquidation(uint256 liquidationId, address sponsor)
         public
@@ -413,6 +414,7 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         fees()
         updateFundingRate()
         nonReentrant()
+        returns (RewardsData memory)
     {
         LiquidationData storage liquidation = _getLiquidationData(sponsor, liquidationId);
 
@@ -434,6 +436,9 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
 
         // There are three main outcome states: either the dispute succeeded, failed or was not updated.
         // Based on the state, different parties of a liquidation receive different amounts.
+        // After assigning rewards based on the liquidation status, decrease the total collateral held in this contract
+        // by the amount to pay each party. The actual amounts withdrawn might differ if _removeCollateral causes
+        // precision loss.
         RewardsData memory rewards;
         if (liquidation.state == Status.DisputeSucceeded) {
             // If the dispute is successful then all three users should receive rewards:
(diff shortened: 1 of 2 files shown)


daywiss (approved):
i like this idea, nice change and a win on byte size

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -72,6 +72,19 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         FixedPoint.Unsigned disputerDisputeRewardPct;
     }
 
+    // This struct is used in the `withdrawLiquidation` method that disperses liquidation and dispute rewards.
+    // `payToX` stores the total collateral to withdraw from the contract to pay X. This value might differ
+    // from `paidToX` due to precision loss between accounting for the `rawCollateral` versus the
+    // fee-adjusted collateral. These variables are stored within a struct to avoid the stack too deep error.
+    struct RewardsData {
```
chrismaree, line 79: Would it make sense to add the `RewardsData` within the `LiquidationData` Struct? this way it will be persistent after the fact and can be seen after the liquidation is complete?

chrismaree: In fact that doesn't work nor make sense as we delete the liquidation data once the withdrawal is finished.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -426,70 +433,66 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
...
-        amountWithdrawn = _removeCollateral(rawLiquidationCollateral, withdrawalAmount);
+        require(
+            (rewards.payToLiquidator.add(rewards.payToSponsor).add(rewards.payToDisputer)).isGreaterThan(0),
+            "Nothing to withdraw"
+        );
 
-        emit LiquidationWithdrawn(msg.sender, amountWithdrawn.rawValue, liquidation.state, settlementPrice.rawValue);
+        // Decrease the total collateral held in liquidatable by the amount to pay each party. The actual amounts
+        // withdrawn might differ if _removeCollateral causes precision loss.
+        if (rewards.payToLiquidator.isGreaterThan(0)) {
```
chrismaree, line 472: Seeing there are only three branches that payment can occur above (`DisputeSucceeded`, `DisputeFailed ` or `PreDispute`) would it make sense to remove these if statements and simply put the logic to decrement the total collateral and preform the token transfers within the respective if/else branches above? If you did this you'd be able to remove 3 if statements. WDYT?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -426,70 +433,66 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
...
+        emit LiquidationWithdrawn(
+            msg.sender,
+            rewards.paidToLiquidator.rawValue,
+            rewards.paidToDisputer.rawValue,
+            rewards.paidToSponsor.rawValue,
+            liquidation.state,
+            settlementPrice.rawValue
+        );
 
-        return amountWithdrawn;
```
chrismaree, line 492: do we want to return anything? it could be useful to return 3 uints for the amount sent to the liquidator, disputer and sponsor. WDYT?

chrismaree: However, this will increase bytecode a bit and I'm not sure how useful it is. Just thinking from a composability perspective: if you were connecting the EMP to another contract that preforms liquidations you'd want to be able to see these numbers returned from the function call.

chrismaree (approved):
I really like this change! Doing all the token transfers in the same call makes a lot of sense & is far cleaner. The incentives still work fine as anyone who would get a payout is still incentivized to do so.

I had a few comments/questions on the refactor but other than these I think this is good to go and a big improvement 👍

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: Yeah this struct is only initialized and modified within the `withdrawLiquidation` function. I originally wanted to just use `payToX` and `paidToX` as temporary variables within the function but due to scoping errors I decided to smash these into one struct.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: You're totally right I'll try that

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: > do we want to return anything? it could be useful to return 3 uints for the amount sent to the liquidator, disputer and sponsor. WDYT?

It could yeah. Currently the bots use this return value to check if they are getting > 0 rewards so maybe we return the 3 numbers. It doesn't use much bytecode, ~200 bytes

nicholaspai: Addressed in this [commit](https://github.com/UMAprotocol/protocol/pull/2152/commits/ba30b32ddad538de6e07dadfff09e7ace7e8989f)

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: Addressed in this [commit](https://github.com/UMAprotocol/protocol/pull/2152/commits/ba30b32ddad538de6e07dadfff09e7ace7e8989f)

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -426,70 +435,77 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
...
-        require(withdrawalAmount.isGreaterThan(0), "Invalid withdrawal amount");
 
-        // Decrease the total collateral held in liquidatable by the amount withdrawn.
-        amountWithdrawn = _removeCollateral(rawLiquidationCollateral, withdrawalAmount);
+        // If the raw amounts of collateral to pay out are 0, which is possible if the liquidation is for 0 collateral,
+        // then revert because there are no rewards to pay.
+        require(
+            (rewards.payToLiquidator.add(rewards.payToSponsor).add(rewards.payToDisputer)).isGreaterThan(0),
+            "Rewards equal to 0"
+        );
```
mrice32, line 494: This may be necessary, but just playing devil's advocate, what downside does this guard against?

mrice32 (approved):
Looks awesome! Very very cool stuff. I almost suggested a restructuring of the if-else to do all the collateral removal and transfers at the end, but because of the rounding stuff, that would probably be much more complex.

mrice32:
> Looks awesome! Very very cool stuff. I almost suggested a restructuring of the if-else to do all the collateral removal and transfers at the end, but because of the rounding stuff, that would probably be much more complex.

@nicholaspai side note: I think this abi change will definitely break the bots (albeit in a small way).

nicholaspai:
> > Looks awesome! Very very cool stuff. I almost suggested a restructuring of the if-else to do all the collateral removal and transfers at the end, but because of the rounding stuff, that would probably be much more complex.
> 
> @nicholaspai side note: I think this abi change will definitely break the bots (albeit in a small way).

Totally will. Should we (a) save the legacy ABI somewhere or (b) pin the bots to an older commit hash?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: It prevents the one edge case where liquidated collateral is 0 so rewards are 0 paid out. Regardless the liquidation will be deleted afterwards so maybe we can just remove this. The liquidator has no incentive to send out such a liquidation, in fact we can block liquidations for 0 collateral

nicholaspai:
> > Looks awesome! Very very cool stuff. I almost suggested a restructuring of the if-else to do all the collateral removal and transfers at the end, but because of the rounding stuff, that would probably be much more complex.
> 
> @nicholaspai side note: I think this abi change will definitely break the bots (albeit in a small way).

Actually this won't affect the EMP ABI at all because its only affecting the Perp. This is why I held off on editing the EMP

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: @mrice32 This [test](https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/financial-templates/perpetual-multiparty/PerpetualLiquidatable.js#L1662) in particular show cases a situation where the `require` statement is hit.

I decided to block liquidations for 0 tokens in this [commit](https://github.com/UMAprotocol/protocol/pull/2152/commits/b6b774a3bf7ae516f72cfa61d8cdd9e49aebaf2e) because we explicitly block 0 collateral transactions for `deposit`, `withdraw`, and 0 token `create`'s, PTAL!

mrice32:
> > > Looks awesome! Very very cool stuff. I almost suggested a restructuring of the if-else to do all the collateral removal and transfers at the end, but because of the rounding stuff, that would probably be much more complex.
> > 
> > 
> > @nicholaspai side note: I think this abi change will definitely break the bots (albeit in a small way).
> 
> Actually this won't affect the EMP ABI at all because its only affecting the Perp. This is why I held off on editing the EMP

Oh right! 🤦 . Thanks for pointing this out @nicholaspai!!

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
mrice32: Just to be totally clear here, once the liquidation is deleted, it will no longer get here because the withdrawable modifier will revert it, right? So this is _only_ guarding against 0-value liquidations, right?

Not super opinionated either way. Totally up to you. Might be worth removing if we get desperate for bytecode space :).

SGTM!

mrice32 (approved):
LGTM! This is very very cool. Either way you decide to go on https://github.com/UMAprotocol/protocol/pull/2152#discussion_r518854593 is good with me!!

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: > Just to be totally clear here, once the liquidation is deleted, it will no longer get here because the withdrawable modifier will revert it, right? So this is only guarding against 0-value liquidations, right?

Correct. This require statement was preventing the 1 `withdrawLiquidation()` call that could go through if someone liquidated 0 collateral.

nicholaspai: FWIW I did end up removing this require statement and also prevented the edge case by blocking 0 token l iqudiations

PR #2153, opened 2020-11-04
feat(affiliates): improve dataset generation and mocking libraries

### LINKED ISSUE #2140: improve dataset generation and mocking to be a more generalized utility
Currently we can generate data sets from bigquery for testing through setting configuration file with parameters. This can be improved through creating a utility to both generate the files and read them out through a mock api. This can benefit usability in testing as well as reduce the file sizes when committing them to repo.

daywiss:
**Motivation**

#2140

**Summary**

This adds a library file called `datasets` which is a collection of modules to generate data set files from bigquery/coingecko and save them to disk in an organized way. This can then be read out through the "mocks" which reconstruct the data from the file. 
Some effort has been put into reducing the filesize for the block list as it will be stored as comma separated values.

This adds a new app called "CreateDataSet" which takes in a configuration file found in createdataset.example.js. 
End to end tests have also been added which are ignored by CI to test the dataset functionality. This can be expanded on in the future as we add more apps/scripts.

**Issue(s)**

Fixes #2140

### DIFF
--- a/packages/affiliates/package.json
+++ b/packages/affiliates/package.json
@@ -18,6 +18,7 @@
   "scripts": {
     "test": "yarn test-mocha && yarn test-truffle",
     "test-mocha": "mocha test/mocha/*.js",
-    "test-truffle": "truffle test test/truffle/*.js --network=test"
+    "test-truffle": "truffle test test/truffle/*.js --network=test",
+    "test-e2e": "mocha test/e2e/*.js"
   }
 }
--- a/packages/affiliates/apps/CreateDataSet.js
+++ b/packages/affiliates/apps/CreateDataSet.js
@@ -1,0 +1,21 @@
+// saves a dataset to a folder. Users a config file, see example in createdataset.example.js
+// outputs a path for you to load the mocks from.
+// node apps/CreateDataSet.js ./createdataset.example.js
+const { Dataset } = require("../libs/datasets");
+const Config = require("../libs/config");
+const { BigQuery } = require("@google-cloud/bigquery");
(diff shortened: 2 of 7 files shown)


packages/affiliates/package.json
```diff
@@ -18,6 +18,7 @@
   "scripts": {
     "test": "yarn test-mocha && yarn test-truffle",
```
chrismaree, line 19: Q: should this also have `yarn test-e2e` or we only want to run them separately to unit tests?

chrismaree (commented):
This is really awesome and super useful! Will the PRs that modify the sample sets come in separately?

daywiss:
> This is really awesome and super useful! Will the PRs that modify the sample sets come in separately?

yes i will make a new pr to update the current tests as well as fix any remaining bugs with the calculation

packages/affiliates/package.json
daywiss: currently needs  to not run in CI since it requires BQ auth. Its more meant right now to run locally until we have something in place that doesn't mind connecting to production sources in CI.

mrice32: I think this makes sense. @daywiss and I discussed this, and certain tests that hit apis particularly hard or use gcloud resources directly seem like they may be better suited for nightly runs on master in ci rather than running on every PR. Thoughts?

mrice32 (approved):
Looks great!

One note: I think the data file size is less important than the _ease of manually viewing, auditing, or changing_ the files. So formatting isn't a huge concern IMO as long as the formatting is human readable.

Issue #2038, opened 2020-09-30
Funding Rate Store: Add mechanism for accruing rewards for funding rate proposers


mrice32:
**Why**

Project Evergreen.

**Difficulty Score [1-10]**

- 6: ~1 day

nicholaspai:
Closed by #2157

Issue #2039, opened 2020-09-30
Funding Rate Store: Add Mechanism for distributing rewards to funding rate proposers


mrice32:
**Why**

Project Evergreen.

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:
### Motivation
It is undecided currently how rewards should work in the Funding Rate Store. The Store can use rewards and penalties to incentivize "proposers" to report the correct funding rate. We need to first determine what the cost of an incorrect funding rate is.

### Proposal Bond
Let's determine the size of a funding rate proposal bond. This is the "penalty" that a dishonest proposer forfeits if they propose an incorrect funding rate. Whenever a funding rate is incorrect, an error occurs in which the contract's reported total debt outstanding deviates from its correct value. Seen from another light, this "error" is equivalent to a someone's potential profit.

In summary: The funding rates reported by the Store are used to **secure the value in the PerpetualContract**. **Therefore, the cost to corrupt the funding rate should exceed the potential profit from corruption of the funding rate.** What affects the potential profit from corruption?

- This `PfC` or error  compounds over time because it is applied as a % of the contract's total debt `D`. For example, imagine the funding rate is off by `X`% for `S` seconds, then after `S` seconds, `D` displays an error of `(D * X)^S`
- Every unit time's `PfC` is proportional to the magnitude between the latest funding rate (i.e. whatever `store.getCurrentFundingRate()` returns) and the true funding rate. If the latest funding rate is 1%/second, and true funding rate is 5%/second, then the `PfC` of `X` is 4%/second.

Let's make some assumptions and put conservative numbers on the `PfC`:
- First of all, let's make an assumption about the maximum absolute `rate` that can be charged. Let's say that the `rate` cannot be higher/lower than the equivalent of 360%/year. In other words, if the maximum `rate` is set and unchanged, then the contract will pay out all of its collateral in 100 days. 
- How long can a funding rate be incorrect for? Assuming a proposer liveness period of 2 hours, a funding rate can only be incorrect for 2 hours. Let's say the current funding rate is incorrect, and an honest proposer submits a new funding rate. Right before the 2 hours expire, a dishonest person (who is incentivized to keep the funding rate unchanged) disputes the proposed funding rate and pays a dispute bond. This deletes the pending funding rate, and the honest proposer can again submit a new funding rate. **Clearly the value of the dispute bond needs to be higher than the profit earned from 2 hours of an incorrect funding rate.**
[...]

PR #2157, opened 2020-11-05
feat(perpetual) Extend FeePayer to include a way to pay FundingRateStore

nicholaspai:
**Motivation**
#2038 

## FundingRatePayer as an extension of FeePayer
Adds a `payFundingRateFee` method to the `FeePayer` contract so that the Perpetual can pay a continuous fee to the FundingRateStore in the exact same manner it pays a regular fee to the Store. 
- The `FundingRateStore` is modified to have public `computeFundingRateFee` and `payFundingRateFee` methods for the FundingRatePayer to call.
- The major difference between the FundingRateStore and the (Oracle)Store is that the OracleStore allows an admin to change its regular/late fee post-deployment. Currently, you can only set these fees in the FundingRateStore upon deployment, but we can similarly permission Setters for these state variables. 
- Also, currently an admin can withdraw funds from the Store, I thought this might be useful for the Perpetual to implement a method `withdrawExcess` that could pull back funds from the FundingRateStore above an excess threshold.

## Mapping Funding-Rates to Perpetual-Contract-Addresses instead of Funding-Rate-Identifiers
Changes the key for `fundingRateRecords` in the `FundingRateStore` from `bytes32 identifier` to `address perpetualContract`. This is done because funding rate proposers need to post bonds denominated in the same units that fees are paid to the Store from the Perpetual. 
- Assumption 1: We can’t force the Store to collect fees from perpetuals in any one currency, because different perpetuals will have different collateral currencies. So we need to assume that FRS has a treasury of fees denominated in several different collateral currencies. 
- In order to align a proposer's costs with their rewards and the security of financial contracts, their costs and rewards need to be in the same collateral currency that the Store taxes the Perpetual in. Otherwise it is possible that the Store taxes Perpetual in unit X and pays a proposer in unit X, but the proposer posted a bond in Y. The proposer's cost can clearly be very different than the theory implies if this denomination mismatch occurs.

So, in conclusion we need to ENFORCE that (1) the proposer's bond, (2) the proposer's reward, and (3) the taxes that the financial contract whose funding rate the proposer is securing are all paid in the same currency.  We can do this by changing the `fundingRateRecord` key from the `fundingRateIdentifier` to the `perpetualContract` itself. 

This would mean that a proposer must state up front which perpetual contract it is proposing a funding rate for. This takes advantage of the fact that each perpetual is mapped to only ONE funding rate identifier, so we can derive the identifier from the contract. Secondly, each perpetual is mapped to only ONE collateral currency, so we know which unit that the proposer is going to pay a bond in. The disputer also now must reference a funding rate using this perpetual's contract, so the disputer also is forced to pay a bond in this currency. 

Finally, it is trivial for the Perpetual contract to query the Store for a funding rate, because it just uses `address(this)` as the key to `getFundingRate()` instead of `fundingRateIdentifier`.
 
**Issue(s)**

Fixes ##2038


packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -26,6 +27,12 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
      *        STORE DATA STRUCTURES         *
      ****************************************/
 
+    enum Roles { Owner, Withdrawer }
+
+    // TODO: Remove this hardcoded `collateralCurrency` and dynamically reward proposers
+    // in the same currency that the financial contract client's collateral is denominated in.
+    // i.e. USDBTC Perp is collateralized by tBTC, it should pay fees to this contract in tBTC,
+    // and it should reward proposers in tBTC.
```
nicholaspai, line 35: See #2039

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -39,9 +39,6 @@ abstract contract FundingRateApplier is Testable, Lockable {
     // Last time the `cumulativeFundingRateMultiplier` was updated.
     uint256 public lastUpdateTime;
 
-    // Identifier in funding rate store to query for.
-    bytes32 public fundingRateIdentifier;
```
nicholaspai, line 43: This is no longer used in this contract to fetch a funding rate, instead the `address(this)` is the new key.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -0,0 +1,121 @@
...
+     * @notice Pays continuous fees (as a % of the collateral pool) to the FundingRateStore contract.
+     * @dev These must be paid periodically for the life of the contract. If the contract has not paid its fee
+     * in a week or more then a late penalty is applied which is sent to the caller. If the amount of
+     * fees owed are greater than the pfc, then this will pay as much as possible from the available collateral.
+     * An event is only fired if the fees charged are greater than 0.
+     * @return totalPaid Amount of collateral that the contract paid (sum of the amount paid to the Store and caller).
+     * This returns 0 and exit early if there is no pfc, fees were already paid during the current block, or the fee rate is 0.
+     */
+    function payFundingRateFees() public nonReentrant() returns (FixedPoint.Unsigned memory totalPaid) {
+        FundingRateStoreInterface fundingRateStore = FundingRateStoreInterface(
```
nicholaspai, line 71: This is mostly copied logic from `payRegularFees`, so I should consider refactoring shared logic

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -48,51 +49,69 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
         require(_proposalLiveness > 0, "Proposal liveness is 0");
         proposalLiveness = _proposalLiveness;
-        collateralCurrency = IERC20(_collateralAddress);
         finder = FinderInterface(_finderAddress);
+
+        // Continuous fees at or over 100% don't make sense.
+        require(_fixedFundingRateFeePerSecondPerPfc.isLessThan(1), "Fee must be < 100% per second.");
+        require(_weeklyDelayFeePerSecondPerPfc.isLessThan(1), "weekly delay fee must be < 100%");
+
+        // TODO: Should we make these fee rates modifiable (by some admin) post deployment?
```
nicholaspai, line 94: Open questions about funding rate fees:
- Should we have a late penalty for this? I included because we want to enforce that fees are continuously paid, mostly taking inspiration from the Regular fee logic
- Should we allow these fee rates to be changeable post-deployment? If so, by whom?
- Should we allow a way for the Perpetual to withdraw fees?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -101,33 +120,97 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+    )
+        external
+        override
+        view
+        returns (FixedPoint.Unsigned memory fundingRateFee, FixedPoint.Unsigned memory latePenalty)
+    {
+        uint256 timeDiff = endTime.sub(startTime);
+
+        // Multiply by the unscaled `timeDiff` first, to get more accurate results.
+        fundingRateFee = pfc.mul(timeDiff).mul(fixedFundingRateFeePerSecondPerPfc);
```
nicholaspai, line 158: We should consider refactoring `computeFundingRateFee` and `computeRegularFee` into a library as these are mostly stateless methods

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualInterface.sol
```diff
@@ -0,0 +1,14 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+
+/**
+ * @notice External methods that the FundingRateStore needs access to.
+ */
+interface PerpetualInterface {
```
nicholaspai, line 10: New interface for the Perpetual which is now expected to be called by the FundingRateStore to determine the `collateralCurrency` and `fundingRateIdentifier` of the calling contract

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -770,7 +771,103 @@ contract("PerpetualPositionManager", function(accounts) {
     assert(await didContractThrow(positionManager.repay({ rawValue: "1" }, { from: sponsor })));
   });
 
-  it("Basic fees", async function() {
+  it("Basic funding rate fees", async function() {
```
nicholaspai, line 774: Pretty much copied (but a lot is stripped out that explicitly tests `_removeCollateral` and `_addCollateral`) from the regular fee test

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -0,0 +1,121 @@
...
+        if (collateralPool.isEqual(0)) {
+            return totalPaid;
+        }
+
+        // Exit early if fees were already paid during this block.
+        if (lastPaymentTime == time) {
+            return totalPaid;
+        }
+
+        (FixedPoint.Unsigned memory fundingRateFee, FixedPoint.Unsigned memory latePenalty) = fundingRateStore
```
daywiss, line 87: Why is there a late penalty?

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -0,0 +1,121 @@
...
+
+        _adjustCumulativeFeeMultiplier(totalPaid, collateralPool);
+
+        if (fundingRateFee.isGreaterThan(0)) {
+            collateralCurrency.safeIncreaseAllowance(address(fundingRateStore), fundingRateFee.rawValue);
+            fundingRateStore.payFundingRateFees(fundingRateFee);
+        }
+
+        if (latePenalty.isGreaterThan(0)) {
+            collateralCurrency.safeTransfer(msg.sender, latePenalty.rawValue);
```
daywiss, line 117: who is the sender here? if it is an EOA why would they ever call this function

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Same reason as it exists for the regular fee I was thinking. If a contract stops paying for whatever reason you impose a fee

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: This is paid in the fees() modifier. But yeah you're right no EOA should be incentivized to call it. It mirrors the FeePayer interface

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: +1 very nice

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
mrice32: Yeah.

Just out of curiosity, I wonder if it would be simpler just to fork the FeePayer and have it do two payments instead of 1. Not sure if that would be simpler though.

What are your thoughts?

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
mrice32: Wait, what? The contract is _paying_ the EOA, right? So the late fee _is_ what incentivizes the EOA to pay IIUC.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -0,0 +1,121 @@
...
+    event FundingRateFeesPaid(uint256 indexed fundingRateFee, uint256 indexed lateFee);
+
+    /****************************************
+     *              MODIFIERS               *
+     ****************************************/
+
+    // modifier that calls continuous fee paying methods.
+    // TODO: Prettier doesn't allow modifiers to be overridable?
+    // prettier-ignore
+    modifier fees override {
```
mrice32, line 36: cool, didn't realize you could override modifiers.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -10,13 +10,15 @@ import "../../../oracle/interfaces/OracleInterface.sol";
...
+import "../../perpetual-multiparty/PerpetualInterface.sol";
 import "../interfaces/FundingRateStoreInterface.sol";
 import "../../../common/implementation/Testable.sol";
 import "../../../common/implementation/Lockable.sol";
+import "../../../common/implementation/Withdrawable.sol";
 import "../../../common/implementation/FixedPoint.sol";
 
 
-contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
+contract FundingRateStore is FundingRateStoreInterface, Withdrawable, Testable, Lockable {
```
mrice32, line 21: Why make this withdrawable?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: 1. Seems reasonable, but don't have a strong opinion.
2. Maybe a similarly relevant question is: where should the fees be set -- in the funding rate store as a global value for all contracts? Or should they be set per-contract as a deployment param? In the former case, it seems like it would make sense to make it changeable by the DVM since it's a system-wide param. In the latter case, it seems more like a contract param that you might want to be immutable. I would lean towards the latter, but curious to hear others' thoughts.
3. The only case I can think of where this might be necessary would be a shutdown of some type that terminates the contract or a funding rate store migration. I could see us sending the fees back to the contract in those cases, but it sounds complicated to handle that. I wonder if there's a simpler way to deal with one or both of those cases. Any ideas?

mrice32 (commented):
Very nice! A few comments and responses to your open questions

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: I don't think that would simpler, I actually benefit from using the internal cumulative fee multiplier modifying functions and its cleaner to just extend the `fees()` modifier instead of adding another one

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: 2.
> "it's a system-wide param"
My only counter to this is that this is not a DVM system param, it is a "uLabs" or Risk labs system param. I also lean towards the latter and make it immutable per each contract. Therefore the change to make is to store `_fixedFundingRateFeePerSecondPerPfc` and  `_weeklyDelayFeePerSecondPerPfc` in the Perpetual and not the Store

3. Agreed that we don't want to have a situation where we decide to shutdown the Perpetual/Store and have collateral frozen in the Store. We could relinquish these on an emergency shutdown?

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Ah you are right @mrice32 , everyone is incentivized to call this function if `latePenalty` is ever > gas fees

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Yeah neither does prettier apparently

mrice32: Lol

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: No reason at the current implementation, but as discussed in [this comment](https://github.com/UMAprotocol/protocol/pull/2157#discussion_r519061259) we might want to allow the Perpetual to withdraw some fees in an emergency shutdown event. The question though is who should withdraw, and how much.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Looks like they resolved this [issue](https://github.com/prettier-solidity/prettier-plugin-solidity/issues/279) and I upgraded `prettier-plugin-solidity` as a result.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -12,7 +12,6 @@ import "../../oracle/interfaces/StoreInterface.sol";
 import "../../oracle/interfaces/FinderInterface.sol";
 import "../../oracle/implementation/Constants.sol";
 
-
```
nicholaspai, line 15: FYI this new line was removed by updated prettier-plugin-solidity

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -189,7 +188,7 @@ abstract contract FeePayer is Testable, Lockable {
         store.payOracleFeesErc20(address(collateralCurrency), amount);
     }
 
-    function _pfc() internal virtual view returns (FixedPoint.Unsigned memory);
+    function _pfc() internal view virtual returns (FixedPoint.Unsigned memory);
```
nicholaspai, line 191: Also switched ordering by prettier-plugin-solidity

chrismaree (approved):
This PR looks good, if the treasury implementation is what we want to go for. 

However, I'm still not 100% convinced that this design makes the most sense as it introduces a fair amount of seemingly unnecessary complexity & additional token transfers. Granted, I should have raised this in your original issue but wanted to bring it up again here just to discuss it further.

Fundamentally this design works by the `FundingRateStore` imposing a fee on the `FundingRatePayer` to build up a treasury to incentivize people to provide funding rates to the `FundingRateStore` for the given perp. This means that funds follow as follows:

perp contract -> `FundingRateStore` treasury -> funding rate proposer.

As you discuss in detail in the original issue for the [FundingRateStore](https://github.com/UMAprotocol/protocol/issues/2039#issuecomment-717380646) the rewards paid out to proposes is calculated as a function of the total treasury.

The part that I'm questioning is: why do we need to build up a treasury at all in the `FundingRateStore`? If we can define an equation that computes the rewards for a proposed funding rate as a function of time elapsed from the last proposal, the total collateral within a perp & potentially the absolute error between the old and new funding rates then could we not simply pull funds from the perp contract directly to the funding rate proposes wallet without needing to build up a treasury? 

The advantage of this is that we only ever transfer the _exact_ amount of collateral needed to pay the incentives to the proposer, rather than charging a fixed rate to the `FundingRateStore`'s treasury which is then sent to the proposer. This removes the situations where the Fee imposed by the FundingRateStore on the perp is either: 1) too high and the funding rate store needs to return funds to the perp of 2) too low and incentives become misaligned resulting in slow funding rates being pushed on-chain.

The alternative that I am proposing would remove the need for the FundingRateStore to charging a fee and would enable the FundingRateStore to simply pull funds from the perp to pay proposer rewards.

The disadvantages of this are that the perp does not know upfront the cost associated with the funding rate store. In your design, the perp sponsors know that the fund rate store will charge x% per y period (assuming no late fee). In the alternative proposal, this is less explicitly defined as the fees could change depending on the frequency of funding rate requests. However, this difference is not too pronounced if the rewards paid to proposers scales with the time elapsed between previously proposed funding rates.

I'm curious to hear what you guys think about this alternative design.

nicholaspai:
> [Coverage Status](https://coveralls.io/builds/34788587)
> 
> Coverage decreased (-0.2%) to 93.533% when pulling **[07b9d90](https://github.com/UMAprotocol/protocol/commit/07b9d90bff06e4aa329b9430215d558fedda5063) on npai/2038** into **[e7bcc2d](https://github.com/UMAprotocol/protocol/commit/e7bcc2d065bb13043569f51eb6443d60ae2c7966) on master**.

I will fix this, need to add tests to MockFundingRateStore, FundingRateStore, and PerpetualPositionManager

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualInterface.sol
nicholaspai: > Also, would it make sense to put this in an interfaces directory?

I can do this yeah but this would be the only `interface` in this directory so I held off for now

nicholaspai: > Are there other methods we'd want to expose here? we currently don't have an interface for the ExpiringMultiParty so exposing an interface like this is a slightly different pattern.

This interface is only used (currently) by the FundingRateStore, because the store uses `address perpetual` as function arguments, and then uses that to look up `perpetual.collateralCurrency()` or `perpetual.fundingRateIdentifier`. So this interface is specifically only useful to the Store and I kept its interface minimal (and referenced this in the comment)

nicholaspai:
> The disadvantages of this are that the perp does not know upfront the cost associated with the funding rate store.

This is my main concern. To implement the design that you propose, my first thought would be to remove the taxed collateral (i.e. the collateral that is currently being transferred to the FRS) from the Perpetual's `pfc()` and just allow the FRS to withdraw this collateral to reward proposers. However, we still run into the same problem--how does the Perpetual withdraw this now-separated collateral if it gets shutdown (for whatever reason)?

Instead, it could, as you suggest, NOT remove the collateral from the PfC but then it would not know how much collateral at any point in time could be removed by the FRS.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
daywiss: i see i misunderstood what was going where with safeTransfer.

nicholaspai:
@chrismaree I changed the `FundingRatePayer` implementation to provide an external `payFundingRateFees()` method that is callable only by the `FundingRateStore`. This makes the Perpetual assume that the FundingRateStore calls it correctly.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualInterface.sol
```diff
@@ -0,0 +1,16 @@
...
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../../common/implementation/FixedPoint.sol";
+
+/**
+ * @notice External methods that the FundingRateStore needs access to.
+ */
+interface PerpetualInterface {
+    function payFundingRateFees(FixedPoint.Unsigned memory amount) external;
```
mrice32, line 11: nit: should this method be called `retrieve` or `withdraw` rather than `pay` since the caller is pulling the funds?

mrice32 (commented):
I like the simplicity, but I feel as if the previous implementation was a bit more specified:
1. Where is the rate/fee set and by whom?
2. What's the formula for computing this rate/fee? Does it depend on the instantaneous pfc?

Overall, I like this idea -- it does move some of the complexity out of the perp. We don't need to implement everything right now, but I want to make sure that the formula we want to implement makes sense when being computed inside the FRS.

nicholaspai:
> I like the simplicity, but I feel as if the previous implementation was a bit more specified:
> 
> 1. Where is the rate/fee set and by whom?
> 2. What's the formula for computing this rate/fee? Does it depend on the instantaneous pfc?
> 
> Overall, I like this idea -- it does move some of the complexity out of the perp. We don't need to implement everything right now, but I want to make sure that the formula we want to implement makes sense when being computed inside the FRS.

1. Rate will be set in the FRS, probably set immutably on construction. 
2. I'm thinking the rate is either a % of the Perpetual's total `pfc`, or it is a %perSecond that accrues since the last collection time. Either way it depends on instantaneous `pfc` is that a problem? What do you think?

mrice32:
@chrismaree I think there are a few other things to consider:
- Pull method has lower gas costs because no payments need to be made on every "touch" of the contract.
- Push method is more accurate because it extracts fees before any change to pfc (if fees are computed based on pfc).
- In the pull method, if I can trigger the payment by proposing, I can withdraw my funds from the perp before triggering the payment so that I don't have to pay. This probably isn't a huge deal for sufficiently large and diverse contracts.
- Pull method (depending on the formula) can lock up if the contract doesn't have enough funds to pay for the proposal. In the push method, IIUC, this would be guaranteed to default to paying out less to the treasury resulting in a lower incentive rather than reverting.

mrice32:
> > I like the simplicity, but I feel as if the previous implementation was a bit more specified:
> > 
> > 1. Where is the rate/fee set and by whom?
> > 2. What's the formula for computing this rate/fee? Does it depend on the instantaneous pfc?
> > 
> > Overall, I like this idea -- it does move some of the complexity out of the perp. We don't need to implement everything right now, but I want to make sure that the formula we want to implement makes sense when being computed inside the FRS.
> 
> 1. Rate will be set in the FRS, probably set immutably on construction.
> 2. I'm thinking the rate is either a % of the Perpetual's total `pfc`, or it is a %perSecond that accrues since the last collection time. Either way it depends on instantaneous `pfc` is that a problem? What do you think?

1. So all contracts have the same rate for all time? i.e. this is a global "system" parameter?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualInterface.sol
nicholaspai: Yes I like `withdraw`

nicholaspai:
> > > I like the simplicity, but I feel as if the previous implementation was a bit more specified:
> > > 
> > > 1. Where is the rate/fee set and by whom?
> > > 2. What's the formula for computing this rate/fee? Does it depend on the instantaneous pfc?
> > > 
> > > Overall, I like this idea -- it does move some of the complexity out of the perp. We don't need to implement everything right now, but I want to make sure that the formula we want to implement makes sense when being computed inside the FRS.
> > 
> > 
> > 
> > 1. Rate will be set in the FRS, probably set immutably on construction.
> > 2. I'm thinking the rate is either a % of the Perpetual's total `pfc`, or it is a %perSecond that accrues since the last collection time. Either way it depends on instantaneous `pfc` is that a problem? What do you think?
> 
> 1. So all contracts have the same global rate for all time? i.e. this is a global, immutable (without redeploying the FRS) "system" parameter?

Yes, or slightly more nuanced--one fixed rate for each perpetual address. Or, we could allow the DVM (i.e. someone with the owner role) to modify this post-deployment

nicholaspai:
@mrice32 @chrismaree @daywiss To keep the conversation moving and following feedback, I propose the following plan to reward proposers using Perpetual PfC taxes:
- `FundingRateStore` stores a `reward-rate-per-second` for each perpetual address
- The reward-rate-per-second defaults to some global default constant value (i.e. `GLOBAL_DEFAULT=0.0001`) but can be reset post-deployment by the Owner of the `FundingRateStore`.
- Let's assume that the Owner will initially be set to the Governor so that the DVM can ultimately change reward rates
- Following a successfully published funding rate, the proposer can claim the rewards accrued since the last publish time: `seconds_elapsed * reward-rate-per_second * PfC`

The only equal alternative I would propose is that instead of a fixed-rate-per-second rewards rate, there is a fixed-rate. This would not rely on the `seconds_elapsed` since the last publish time and instead would reward a % of the `PfC`

mrice32:
> @mrice32 @chrismaree @daywiss To keep the conversation moving and following feedback, I propose the following plan to reward proposers using Perpetual PfC taxes:
> 
> * `FundingRateStore` stores a `reward-rate-per-second` for each perpetual address
> * The reward-rate-per-second defaults to some global default constant value (i.e. `GLOBAL_DEFAULT=0.0001`) but can be reset post-deployment by the Owner of the `FundingRateStore`.
> * Let's assume that the Owner will initially be set to the Governor so that the DVM can ultimately change reward rates
> * Following a successfully published funding rate, the proposer can claim the rewards accrued since the last publish time: `seconds_elapsed * reward-rate-per_second * PfC`
> 
> The only equal alternative I would propose is that instead of a fixed-rate-per-second rewards rate, there is a fixed-rate. This would not rely on the `seconds_elapsed` since the last publish time and instead would reward a % of the `PfC`

Absent other constraints (like pull vs push), would we like this parameter to be immutably set on deployment of the perp? It seems like that was the (rough idea before)[https://github.com/UMAprotocol/protocol/pull/2157#discussion_r519061259]? Happy to revisit that, though.

nicholaspai:
> > @mrice32 @chrismaree @daywiss To keep the conversation moving and following feedback, I propose the following plan to reward proposers using Perpetual PfC taxes:
> 
> > 
> 
> > * `FundingRateStore` stores a `reward-rate-per-second` for each perpetual address
> 
> > * The reward-rate-per-second defaults to some global default constant value (i.e. `GLOBAL_DEFAULT=0.0001`) but can be reset post-deployment by the Owner of the `FundingRateStore`.
> 
> > * Let's assume that the Owner will initially be set to the Governor so that the DVM can ultimately change reward rates
> 
> > * Following a successfully published funding rate, the proposer can claim the rewards accrued since the last publish time: `seconds_elapsed * reward-rate-per_second * PfC`
> 
> > 
> 
> > The only equal alternative I would propose is that instead of a fixed-rate-per-second rewards rate, there is a fixed-rate. This would not rely on the `seconds_elapsed` since the last publish time and instead would reward a % of the `PfC`
> 
> 
> 
> Absent other constraints (like pull vs push), would we like this parameter to be immutably set on deployment of the perp? It seems like that was the [rough idea before](https://github.com/UMAprotocol/protocol/pull/2157#discussion_r519061259)? Happy to revisit that, though.

Sure let's do that for now. Regardless I'm going to leave all the store rewards logic for the next PR which should close out the core store logic

chrismaree:
> @chrismaree I changed the `FundingRatePayer` implementation to provide an external `payFundingRateFees()` method that is callable only by the `FundingRateStore`. This makes the Perpetual assume that the FundingRateStore calls it correctly.

@nicholaspai  This looks great! I really like how you've been able to pull this logic out and make the functions pull-based. far cleaner IMOP.

> @chrismaree I think there are a few other things to consider:
> 
> * Pull method has lower gas costs because no payments need to be made on every "touch" of the contract.
> * Push method is more accurate because it extracts fees before any change to pfc (if fees are computed based on pfc).
> * In the pull method, if I can trigger the payment by proposing, I can withdraw my funds from the perp before triggering the payment so that I don't have to pay. This probably isn't a huge deal for sufficiently large and diverse contracts.
> * Pull method (depending on the formula) can lock up if the contract doesn't have enough funds to pay for the proposal. In the push method, IIUC, this would be guaranteed to default to paying out less to the treasury resulting in a lower incentive rather than reverting.


@mrice32 I agree with all your points. I just wanted to touch of point 3 as I don't really think this is an issue unless: 1) the sponsor performing this action is the vast majority of the pool or 2) the rewards (and therefore fees) for funding rates are extremely high.


> @mrice32 @chrismaree @daywiss To keep the conversation moving and following feedback, I propose the following plan to reward proposers using Perpetual PfC taxes:
> 
> * `FundingRateStore` stores a `reward-rate-per-second` for each perpetual address
> * The reward-rate-per-second defaults to some global default constant value (i.e. `GLOBAL_DEFAULT=0.0001`) but can be reset post-deployment by the Owner of the `FundingRateStore`.
> * Let's assume that the Owner will initially be set to the Governor so that the DVM can ultimately change reward rates
> * Following a successfully published funding rate, the proposer can claim the rewards accrued since the last publish time: `seconds_elapsed * reward-rate-per_second * PfC`
> 
> The only equal alternative I would propose is that instead of a fixed-rate-per-second rewards rate, there is a fixed-rate. This would not rely on the `seconds_elapsed` since the last publish time and instead would reward a % of the `PfC`

I don't think that this equation should be exclusively time-based. In my mind, the most robust design is to reward proposes based on: a) the error in the funding rate that they correct b) the time elapsed from the previous reward and c) proportional to the total contract `PfC`. i.e. the larger the difference between the previous funding rate and the "correct" funding rate that they provide, the larger the reward paid. The longer the duration from the previous funding rate until the next funding rate (stale rates) the larger the reward. The larger the `PfC` the larger the reward. It could be something like this. Then, the funding rate store pulls the exact amount of the associated reward depending on the above equation only when the payment needs to be made at the conclusion of a successful funding rate proposal.

Overall this LGTM and I think we can now move onto specifying the reward equation more specifically and implementing the associated logic in the Store.

chrismaree (approved):

PR #2158, opened 2020-11-06
feat(scripts) Add script to return non-expiry, non-admin price requests

nicholaspai:

**Motivation**

We might want to display this data on an UMA client.

**Summary**

Queries all `PriceRequestAdded` events and successively filters out non-Admin and non-Expiry requests.

**Details**

Example run:

```sh
yarn truffle exec ./packages/core/scripts/local/HistoricalPriceRequests.js --network mainnet_mnemonic
yarn run v1.22.10
$ /Users/nicholaspai/UMA/protocol/node_modules/.bin/truffle exec ./packages/core/scripts/local/HistoricalPriceRequests.js --network mainnet_mnemonic
Using network 'mainnet_mnemonic'.

There are currently 0 unresolved non-admin price requests.
There have been 3 non-expiry price requests.
✨  Done in 19.04s.
```

### DIFF
--- a/packages/core/scripts/local/HistoricalPriceRequests.js
+++ b/packages/core/scripts/local/HistoricalPriceRequests.js
@@ -1,6 +1,7 @@
 // Usage: From protocol/, run `yarn truffle exec ./packages/core/scripts/local/HistoricalPriceRequests.js --network mainnet_mnemonic`
 const Voting = artifacts.require("Voting");
 const ExpiringMultiParty = artifacts.require("ExpiringMultiParty");
+const Governor = artifacts.require("Governor");
 const { hexToUtf8 } = web3.utils;
 const { getTransactionReceipt } = web3.eth;
 
@@ -9,7 +10,9 @@ const { isAdminRequest } = require("@uma/common");
 async function run() {
   const votingLegacy = await Voting.at("0x9921810C710E7c3f7A7C6831e30929f19537a545");
   const voting = await Voting.deployed();
+  const governor = await Governor.deployed();
 
+  // There have been 2 DVM's deployed on Mainnet to receive price requests so we need to query events from both.
   let priceRequests = await voting.contract.getPastEvents("PriceRequestAdded", {
     fromBlock: 0,
     toBlock: "latest"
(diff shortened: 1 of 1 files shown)


packages/core/scripts/local/HistoricalPriceRequests.js
```diff
@@ -0,0 +1,114 @@
...
+
+  // Get all non admin votes.
+  const nonAdminReqs = resolvedReqs.filter(req => {
+    return !isAdminRequest(hexToUtf8(req.req.returnValues.identifier));
+  });
+
+  // To determine if a price request was a non-expiry price request, we can check if a ContractExpired
+  // event was emitted in the same block # as the PriceRequestAdded event.
+  let getEventsPromises = [];
+  for (let i in nonAdminReqs) {
```
mrice32, line 71: nit: I think this could probably be a little simpler if you passed an async function to `.map` since it would allow you to use async await throughout rather than chaining `.then`s:
```
const getEventsPromises = nonAdminReqs.map(async request => {
  const receipt = await getTransactionReceipt(request.req.transactionHash);
  const emp = await ExpiringMultiParty.at(receipt.to);
  ...
}
```

packages/core/scripts/local/HistoricalPriceRequests.js
```diff
@@ -0,0 +1,114 @@
...
+    const promise = new Promise(resolve => {
+      // To trigger a price request, some accounts called a financial contract that triggered a price request. So
+      // we'll grab the contract that was called.
+      getTransactionReceipt(nonAdminReqs[i].req.transactionHash).then(txn => {
+        const empAddress = txn.to;
+
+        ExpiringMultiParty.at(empAddress).then(emp => {
+          emp.contract
+            .getPastEvents("ContractExpired", {
```
mrice32, line 80: lol, this is v clever -- very nice.

mrice32 (approved):
Looks awesome!

packages/core/scripts/local/HistoricalPriceRequests.js
nicholaspai: interesting this doesn't speed up anything but prob easier to read

nicholaspai: Actually speeded up by 2 seconds because map is faster than pushing into an array of promises

mrice32 (approved):
LGTM!!

Issue #2154, opened 2020-11-05
Refactor CreatePriceFeed.js's default price feed configs to a seperate file


chrismaree:
**Why**

`CreatePriceFeed.js` reads in the identifier from an EMP and returns a given price feed config. Right now there are a few price feeds that are hardcoded in this file. While this is fine for now it will not scale well once we have more than a couple of identifiers. This should be refactored out to be in a separate file containing the configs.

**How**

Refactor `packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js` to remove default configs.

**Resolution**

`packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js` is refactored and tests still pass as expected.

**Difficulty Score [1-10]**
- 3: ~1 hour

PR #2162, opened 2020-11-09
feat(price-feed) refactor price feeds configs to a seperate file

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2154

**Summary**

Refactors out the default configs from the `CreatePriceFeed.js` file. 

**Details**

This refactor will be important when we launch developer mining as each emp receiving developer mining needs to have a registered price feed config within this file.

**Issue(s)**

close 2154


daywiss:
What is the idea here? I remember you mentioning this to me before but i forget

daywiss (approved):

mrice32 (approved):

chrismaree:
> What is the idea here? I remember you mentioning this to me before but i forget

The main idea is that each partner in developer mining & liquidator of last resort program **must** have their contract identifier registered in this file for the liquidator/disputer bots to run and for us to be able to get synthetic prices denominated in collateral for payouts. Therefore, the list of default price feeds that we support are going to grow over time as we bring on more and more identifiers (this could happen quickly in the next few weeks as well). 

I wanted to refactor this to a separate file to make the addition process easier for third parties. In an ideal world, someone could simply submit a PR to the repo to add their identifier config to this file & then we support their identifier in the bots and developer mining.

PR #2163, opened 2020-11-09
feat(solidity-licences) Added licenses to all Solidity contracts

### LINKED ISSUE #2143: Add SPDX license to all contract files
**Why**

It is best practice to include an SDPDX license in contract files. Without this, every time you compile the repo contracts you are hit with a warning saying they should be included.

**How**

Add the licence to all `.sol` files.

**Resolution**

All contracts are updated and truffle passes compilation with no warnings.

**Difficulty Score [1-10]**

- 3: ~1 hour

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2143

**Summary**

To be compliant with the Solidity styling guide, all Solidity files should specify their SPDX license at the top in a comment. This PR adds this. The list of possible licenses can be found [here](https://spdx.org/licenses/). The `AGPL-3.0-only` tag best matches the license of this repo.

**Question**
The other license option would be `AGPL-3.0-or-later`. Would we prefer that one or is the `AGPL-3.0-only` fine?

**Issue(s)**

close #2143

### DIFF
--- a/packages/core/contracts/Migrations.sol
+++ b/packages/core/contracts/Migrations.sol
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.6.0;
 
 
--- a/packages/core/contracts/common/implementation/AddressWhitelist.sol
+++ b/packages/core/contracts/common/implementation/AddressWhitelist.sol
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.6.0;
 
 import "@openzeppelin/contracts/access/Ownable.sol";
--- a/packages/core/contracts/common/implementation/ExpandedERC20.sol
+++ b/packages/core/contracts/common/implementation/ExpandedERC20.sol
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.6.0;
 
 import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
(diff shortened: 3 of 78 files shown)


nicholaspai (approved):
> Would we prefer that one or is the AGPL-3.0-only fine?

Fine with 3.0-only for now so we don't unintentionally accept any future license changes that would be unsavory.

Also, OOC this does not contribute any bytecode right?

daywiss:
Is there any reason to choose this one over any of the others? I skimmed it and it seems fine, but less permissive than what i commonly see like MIT or BSD licenses. I am not really opinionated here though so i am ok with it.

daywiss (approved):

mrice32 (approved):

chrismaree:
> Is there any reason to choose this one over any of the others? I skimmed it and it seems fine, but less permissive than what i commonly see like MIT or BSD licenses. I am not really opinionated here though so i am ok with it.

I chose this license because it is the same one the repo has.

mrice32:
> Is there any reason to choose this one over any of the others? I skimmed it and it seems fine, but less permissive than what i commonly see like MIT or BSD licenses. I am not really opinionated here though so i am ok with it.

We chose this a while back for the entire repo. GPLV3 is the license that linux uses. The TLDR is that if you modify the code and distribute the compiled modification to users, you need to open source your code. The "A" in AGPL is a clause that requires that even if you don't distribute the compiled code, allowing users to interact with it "over a network" still requires that you open source it.

I'm not a lawyer, but you're right that it's less permissive than some others.

PR #2168, opened 2020-11-10
feat(funding-rate-store) Compute and pay rewards and charge bonds

### LINKED ISSUE #2039: Funding Rate Store: Add Mechanism for distributing rewards to funding rate proposers

nicholaspai:

**Motivation**

#2039 

**Summary**

## Proposal rewards
1. Reward rates are stored for each perpetual address
2. Reward rates can be changed by the perpetual contract itself, so these rates are set once in the Perpetual contract's constructor
3. Proposal rewards are proportional to: time elapsed since last proposal, difference between current and proposed rates, and PfC
4. Reward %'s to pay to successful proposer are stored in the proposal struct and are later withdrawn from the Perpetual contract as needed. Note that if the `PfC` changes between time of proposal and time proposal expires, the reward will change.

## Proposal bonds
1. Proposer must stake a proposal bond
2. Disputer must also stake a proposal bond. Winner of dispute receives back their bond + loser's bond. Importantly, no proposal rewards are paid out on a dispute, ensuring that self-propose-dispute cycles are zero-sum guaranteed losses.
3. Currently, proposal bond %, charged as a % of the Perpetual's `PfC`, is stored as a contract-wide constant, initialized on construction. This [thread](https://github.com/UMAprotocol/protocol/issues/2039) indicates it should be somewhere > 0.04%, so I've set it to 0.08% in the migration scripts as a placeholder.

Additional changes:
- Every public method in `FundingRateStore` now attempts to publish pending proposals and pay out proposer rewards. This is motivated mostly to make sure that proposal rewards are paid out ASAP, to minimize chance that proposers can delay their withdrawals to maximize rewards.

**Issue(s)**

Fixes #2039


packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -11,14 +11,15 @@ import "../../common/implementation/Testable.sol";
...
+import "./FinancialContractInterface.sol";
 
 /**
  * @title FeePayer contract.
  * @notice Provides fee payment functionality for the ExpiringMultiParty contract.
  * contract is abstract as each derived contract that inherits `FeePayer` must implement `pfc()`.
  */
 
-abstract contract FeePayer is Testable, Lockable {
+abstract contract FeePayer is FinancialContractInterface, Testable, Lockable {
```
nicholaspai, line 22: `FundingRateStore` needs access to `pfc()`, so I added an interface it can access instead of importing this entire `FeePayer` contract

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -72,14 +72,18 @@ abstract contract FundingRateApplier is Testable, Lockable {
     /**
      * @notice Constructs the FundingRateApplier contract. Called by child contracts.
      * @param _finderAddress Finder used to discover financial-product-related contracts.
+     * @param _fundingRateRewardRate Reward rate to pay FundingRateStore.
      */
-    constructor(address _finderAddress) public {
+    constructor(address _finderAddress, FixedPoint.Unsigned memory _fundingRateRewardRate) public {
```
nicholaspai, line 77: 1 new contructor param, used to call `setRewardRate()` on the FundingRateStore for this perpetual address

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -58,7 +58,7 @@ abstract contract FundingRatePayer is FeePayer, PerpetualInterface {
 
     /**
      * @notice Sends `amount` fees to the FundingRateStore contract and debits the raw collateral accordingly.
-     * @dev Callable only by the FundingRateStore.
+     * @dev Callable only by the FundingRateStore. This should never revert.
```
nicholaspai, line 61: I made changes in this method so that it doesn't revert, we don't want the FundingRateStore to call this method and have it revert

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -49,21 +52,53 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
     );
-    event PublishedRate(address indexed perpetual, int256 rate, uint256 proposalTime, address indexed proposer);
     event FinalFeesPaid(address indexed collateralCurrency, uint256 indexed amount);
 
+    /****************************************
+     *                MODIFIERS             *
+     ****************************************/
+
+    // Pubishes any pending proposals whose liveness has passed, pays out rewards for such proposals
+    modifier publishAndWithdrawProposal(address perpetual) {
```
nicholaspai, line 97: This internal method, and modifier added to every external contract call, is the bulk of this contract's changes. After discussions with @mrice32, we decided that there should be an external method callable by anyone to withdraw proposal rewards.

Additionally, every method ultimately was checking whether a pending proposal should be published, so I decided to move this into an internal method and have every method call it before running their core logic.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -301,6 +417,34 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
         store.payOracleFeesErc20(address(collateralCurrency), amount);
     }
 
+    function _calculateProposalRewardPct(
```
nicholaspai, line 420: This method returns reward amounts as:
- `reward % per second` * `seconds elapsed` * `% change in proposed funding rate from current rate`

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -49,28 +55,64 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
     constructor(
         uint256 _proposalLiveness,
         address _finderAddress,
-        address _timerAddress
+        address _timerAddress,
+        FixedPoint.Unsigned memory _proposalBondPct
     ) public Testable(_timerAddress) {
         require(_proposalLiveness > 0, "Proposal liveness is 0");
         proposalLiveness = _proposalLiveness;
+        proposalBondPct = _proposalBondPct;
```
nicholaspai, line 115: TODO: we could make `proposalBondPct` unique for each `fundingRateRecord`

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -238,43 +283,139 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+        );
 
         IERC20 collateralCurrency = IERC20(PerpetualInterface(perpetual).getCollateralCurrency());
+        // TOOD: Decide whether loser of dispute should lose entire bond or partial
         if (disputeSucceeded) {
             // If dispute succeeds:
-            // - pay disputer the dispute reward + final fee rebate
+            // - Disputer earns back their bonds: dispute bond + final fee bond
+            // - Disputer earns as reward: the proposal bond
+            FixedPoint.Unsigned memory disputerRebate = fundingRateDispute.proposal.proposalBond.add(
```
nicholaspai, line 300: Question: Should we store an additional `disputeBond` in the `fundingRateRecord`? Currently I do NOT because the `disputeBond == proposalBond`.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -84,71 +126,48 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
 
     /**
      * @notice Returns the timestamp at which the current funding rate was proposed.
-     * @dev The "current funding rate" is defined as that returned by `getFundingRateForIdentifier(id)`
      * @param perpetual Perpetual to retrieve propose time for.
      * @return propose timestamp.
      */
-    function getProposeTimeForContract(address perpetual) external view nonReentrantView() returns (uint256) {
-        return _getLatestProposeTimeForContract(perpetual);
+    function getProposeTimeForContract(address perpetual)
```
nicholaspai, line 143: This method is only used for testing, am open to removing it. The main feature it provides is that the caller can get an up to date proposeTime after any pending proposals are published

packages/core/contracts/financial-templates/funding-rate-store/test/MockFundingRateStore.sol
```diff
@@ -19,9 +19,10 @@ contract MockFundingRateStore is FundingRateStoreInterface, Testable {
         // e.g. fundingRate = 0.01 means 1% of token amount charged per second.
         FixedPoint.Signed fundingRate;
         uint256 timestamp; // Time the verified funding rate became available.
+        FixedPoint.Unsigned rewardRate;
     }
 
-    mapping(address => FundingRate[]) private fundingRates;
```
nicholaspai, line 24: We were only using the last index of this FundingRate array for this contract, and the `FundingRateStore` does NOT store an array of FundingRates for each perp.

packages/core/test/financial-templates/funding-rate-store/FundingRateStore.js
```diff
@@ -74,12 +74,328 @@ contract("FundingRateStore", function(accounts) {
...
   it("Liveness check", async function() {
-    assert(await didContractThrow(FundingRateStore.new(0, finder.address, timer.address)));
+    assert(await didContractThrow(FundingRateStore.new(0, finder.address, timer.address, { rawValue: "0" })));
   });
 
-  it("Initial Funding Rate and Propose Time of 0", async function() {
-    assert.equal((await fundingRateStore.getFundingRateForContract(mockPerpetual.address)).rawValue.toString(), "0");
-    assert.equal((await fundingRateStore.getFundingRateForContract(mockPerpetual.address)).toString(), "0");
+  it("Initial Funding Rate, Reward Rate, and Propose Time of 0", async function() {
+    const getFundingRate = fundingRateStore.getFundingRateForContract;
```
nicholaspai, line 81: These methods have to be called via the `.call()` pattern now because they are not `view` methods anymore becuse they all attempt to publish any pending proposals

packages/core/contracts/financial-templates/common/FinancialContractInterface.sol
```diff
@@ -0,0 +1,11 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "../../common/implementation/FixedPoint.sol";
+
+/**
+ * @notice External methods that any financial contract must implement.
+ */
+interface FinancialContractInterface {
```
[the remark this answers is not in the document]
nicholaspai: +1 will implement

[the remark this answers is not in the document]
nicholaspai: OOC why isn't thi salready in AdminstrateeInterface?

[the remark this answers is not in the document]
nicholaspai: What about just adding `pfc()` to the PerpetualContractInterface versus adding `pfc` to AdministrateeInterface?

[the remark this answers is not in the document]
nicholaspai: Actually ^ is problematic because the ExpiringMultiParty would have to implement `withdrawFundingRateFees` because `FeePayer` would have to be a `PerpetualInterface`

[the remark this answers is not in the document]
nicholaspai: Added `pfc` to `AdministrateeInterface` in this [PR](https://github.com/UMAprotocol/protocol/pull/2168/commits/6c5738091a74d24c92724d4a7d470ebf07e5e6a8)

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -22,7 +21,7 @@ import "../common/FeePayer.sol";
  * on a price feed. On construction, deploys a new ERC20, managed by this contract, that is the synthetic token.
  */
 
-contract PricelessPositionManager is FeePayer, AdministrateeInterface {
```
mrice32, line 25: nit: why was this removed? Shouldn't this administratee interface apply to all financial contracts? (this is what allows them to be emergency shutdown, etc, by the DVM)

mrice32 (approved):
Looks great! One minor comment

packages/core/contracts/financial-templates/common/FinancialContractInterface.sol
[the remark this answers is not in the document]
mrice32: LGTM! I actually liked your naming here. Maybe not in this PR, but we might consider changing the AdministrateeInterface to FinancialContractInterface, as I think that name is a little clearer.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
nicholaspai: FeePayer is now AdminstrateeInterface so PricelessPositionManager inherits it

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -84,71 +128,48 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
      */
     function getFundingRateForContract(address perpetual)
```
nicholaspai, line 129: Should we make this method `view` only, and therefore not disperse rewards?

Pro:
- Every perpetual contract method calls `getFundingRateForContract` so it would reduce Perpetual function gas 

Cons:
- Perpetual contract method calls would no longer cause proposal rewards to get dispersed, instead only `propose`, `withdrawProposalReward`, `dispute`, `settleDispute` on the FundingRateStore would disperse rewards. This opens up room for a proposer to "time" their proposal rewards to when the Perpetual's `pfc` is highest (recall that proposal reward is a stored % of the current `pfc`)

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -238,43 +285,142 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+        FixedPoint.Signed memory currentRate
+    ) external view returns (FixedPoint.Unsigned memory reward) {
+        return _calculateProposalRewardPct(perpetual, startTime, endTime, proposedRate, currentRate);
+    }
+
     /****************************************
      *         INTERNAL FUNCTIONS           *
      ****************************************/
 
+    // TODO: Do we need to add a reentrancy guard to this method because it makes external calls?
```
chrismaree, line 383: yes. I think it should have the guard. could leave this for a separate PR where we go through again double checking all reentrancy points.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -238,43 +285,142 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+        fundingRateRecord.rate = fundingRateRecord.proposal.rate;
+        fundingRateRecord.proposeTime = fundingRateRecord.proposal.time;
+
+        // Calculate reward for proposer using saved reward rate and current perpetual PfC.
+        FixedPoint.Unsigned memory pfc = AdministrateeInterface(perpetual).pfc();
+        FixedPoint.Unsigned memory rewardRate = fundingRateRecord.proposal.rewardRate;
+        FixedPoint.Unsigned memory reward = rewardRate.mul(pfc);
+
+        // Pull reward payment from Perpetual, and transfer (payment + final fee bond) to proposer.
+        // TODO: Should we add a try-catch around these external calls incase the `perpetual.withdrawFundingRateFees`
```
chrismaree, line 402: What would the catch state do? It would make sense to do this perhaps as this function is applied as a modifier in a few places and so it could cause other logic to revert & that might be a broken state. For example, if this function is reverting due to the perpetual being unable to withdraw funds for whatever reason it could be possible that the funding rate store's pending funding rate can not be disputed.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
chrismaree: pros of keeping it like this is it's simpler. cons are that we are bound to keeping these bonds equal. there could be a scenario where we want them to be different but I think for now it's fine to assume that they are equal.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -84,71 +128,48 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+     * @notice Propose a new funding rate for a perpetual.
      * @dev This will revert if there is already a pending funding rate for the perpetual.
-     * @dev Caller must approve this this contract to spend `finalFeeBond` amount of collateral, which they can
+     * @dev Caller must approve this contract to spend `finalFeeBond` + `proposalBond` amount of collateral, which they can
      * receive back once their funding rate is published.
      * @param perpetual Perpetual contract to propose funding rate for.
      * @param rate Proposed rate.
      */
-    function propose(address perpetual, FixedPoint.Signed memory rate) external nonReentrant() {
+    function propose(address perpetual, FixedPoint.Signed memory rate)
```
chrismaree, line 163: should you add a chack to see if the proposal rate != current funding rate? based off how you compute rewards the reward would be 0 but it should perhaps still be blocked. wdyt?

packages/core/contracts/financial-templates/funding-rate-store/test/MockFundingRateStore.sol
chrismaree: are we sure we want this to be the behavior? does the levered token not perhaps need this info? the downsides for storing past funding rates are simply slightly higher gas, no?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
chrismaree: nvm. I see this is exactly how you deal with it in the tests here: https://github.com/UMAprotocol/protocol/pull/2168/files#r521478112

I'm ok with it being not viewed as we can still compute the output of the function

packages/core/test/financial-templates/funding-rate-store/FundingRateStore.js
```diff
@@ -74,12 +74,359 @@ contract("FundingRateStore", function(accounts) {
...
+      postBalanceDisputer = await collateralCurrency.balanceOf(disputer);
+
+      // Disputer gets proposer's bond.
+      assert.equal(postBalanceProposer.sub(preBalanceProposer).toString(), "0");
+      assert.equal(postBalanceDisputer.sub(preBalanceDisputer).toString(), toWei("1.25"));
+
+      // Store and Perpetual balance does not change post propose-dispute process
+      assert.equal(postBalanceStore.toString(), preBalanceStore.toString());
+      assert.equal(postBalancePerpetual.toString(), preBalancePerpetual.toString());
+    });
```
chrismaree, line 429: +1 these tests are really good & thorough. good job

chrismaree (approved):
On the initial pass, I think this is excellent! your code is clean, well commented, and tested. I really like the pull technique you used for extracting the fees from the perp & the associated calculations thereof. good work @nicholaspai !

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Yeah I think if we want to make them different in the future we can just add the param to the struct

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: I am going to remove it, as its not part of any interface

packages/core/contracts/financial-templates/funding-rate-store/test/MockFundingRateStore.sol
nicholaspai: Can't we store funding rates just as effectively via events? Also, for the levered token I am pretty sure we can implement it by only knowing the last funding rate.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: I'm going to leave it as non-view for now, but we should revisit it and highlight to OZ because `getFundingRateForContract` does get called on every single Perpetual method so it contributes non-trivially to gas

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -18,7 +19,7 @@ import "../../oracle/implementation/Constants.sol";
  * contract is abstract as each derived contract that inherits `FeePayer` must implement `pfc()`.
  */
 
-abstract contract FeePayer is Testable, Lockable {
+abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
```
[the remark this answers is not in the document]
nicholaspai: Yep exactly, because `AdministrateeInterface` now includes a `pfc` method which was already in FeePayer

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -49,28 +55,66 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
-    event ProposedRate(address indexed perpetual, int256 rate, uint256 proposalTime, address indexed proposer);
+    event ChangedRewardRate(address indexed perpetual, uint256 rewardRate);
+    event ProposedRate(
+        address indexed perpetual,
+        int256 rate,
+        uint256 indexed proposalTime,
+        address indexed proposer,
+        uint256 rewardPct,
+        uint256 proposalBond
+        // TODO: Should we add final fee bond to this event?
```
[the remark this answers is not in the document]
nicholaspai: done

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: I added the guard. In other contracts we don't add this guard to non-external methods so we should think more about this in another PR

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Yeah I think the catch state should maybe just emit an event, but it should make sure that the Perpetual.withdrawFees doesn't throw for any reason

nicholaspai: I added a catch statement + tests to make sure a reverting `withdrawFundingRateFees` doesn't freeze the contract

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Yes let's catch that

nicholaspai:
Two important changes I just made thanks to comments, [commit](https://github.com/UMAprotocol/protocol/pull/2168/commits/9010680207c9e987bd41877f7cf2c498c3e4b36c):

1. Internal method `_publishRateAndWithdrawRewards()` now adds a try-catch around the `Perpetual.withdrawFundingRateFees()` external call which ensures that the Store will not lock up if the Perpetual's method reverts for any reason. This will effectively ignore the revert and not pay the proposer their reward (but they'll get back their bonds). Tests are added as well.

2. Detailed comments above the `FundingRateStore` contract detailing all cash flows, hopefully this is helpful.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: SGTM

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Yeah, agree with @chrismaree that if you just need it for testing, then we can create a test contract to expose the info. Prefer to remove anything that we don't see a use for in a production context.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: I don't think internal methods should have guards since they are only called by external methods that _do_ have guards. So they are effectively guarded. If you want to emphasize that this method is protected from re-entrancy, it might be worth putting a comment on the method that callers are expected to be guarded.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -238,43 +308,150 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+        if (_getProposalState(fundingRateRecord.proposal) != ProposalState.Expired) {
+            return;
+        }
+
+        // Publish rate and proposal time.
+        fundingRateRecord.rate = fundingRateRecord.proposal.rate;
+        fundingRateRecord.proposeTime = fundingRateRecord.proposal.time;
+
+        // Calculate reward for proposer using saved reward rate and current perpetual PfC.
+        FixedPoint.Unsigned memory pfc = AdministrateeInterface(perpetual).pfc();
```
chrismaree, line 419: The use of pfc in this equation like this makes this implementation and every function in the funding rate store associated with rewards vulnerable to a flash loan attack! Specifically, because the reward is a function of pfc, a flash swap can be used to increase the perp pfc during the withdrawal tx and to extract an unfair reward for proposing a funding rate. 

This attack would look as follows:
1) Propose a valid funding rate.
2) wait until liveness period has passed.
3) execute a transaction that does the following internal steps:
3.a) flash swap perp collateral currency.
3.b) create a small position, well above the GCR, collateralized with the flashswap'ed collateral. The larger the relative size of the flashswap to the total collateral in the contract the more profitable the attack.
3.c) withdraw your reward from the proposal. During this, the calculation of pfc is biased upwards from your collateral created during 3.b.
3.d) redeem position created in 3.b. This will return all flashborrowed collateral
3.e) repay flashloan
4) profit

In a discussion with @nicholaspai there is a solution to this by enforcing that the last addition to contract pfc was done in a separate block to `_publishRateAndWithdrawRewards` being called. This solution is discussed in issue https://github.com/UMAprotocol/protocol/issues/2183 in more detail. While this is a breaking issue that **must** be addressed before this contract is audited, it should be left for a separate PR due to the implementation complexity in this proposal.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -72,20 +72,18 @@ abstract contract FundingRatePayer is FeePayer, PerpetualInterface {
...
         FixedPoint.Unsigned memory collateralPool = _pfc();
 
-        // If contract has no PfC, then cannot pay any fees.
-        require(!collateralPool.isEqual(0), "PfC is 0");
-        // The fee must be non-zero.
-        require(!amount.isEqual(0), "Funding fee is 0");
-        // The fee must be < available collateral.
-        require(collateralPool.isGreaterThan(amount), "Funding fee is more than PfC");
+        // If amount to pay or collateral pool is 0, then return early.
+        if (!amount.isEqual(0) && !collateralPool.isEqual(0)) {
```
daywiss, line 76: small nit here but why not return early like the comment suggests?

packages/core/contracts/financial-templates/funding-rate-store/interfaces/FundingRateStoreInterface.sol
```diff
@@ -16,5 +16,11 @@ interface FundingRateStoreInterface {
      * @return FixedPoint.Signed representing the funding rate for the given contract. 0.01 would represent a funding
      * rate of 1% per second. -0.01 would represent a negative funding rate of -1% per second.
      */
-    function getFundingRateForContract(address perpetual) external view returns (FixedPoint.Signed memory);
+    function getFundingRateForContract(address perpetual) external returns (FixedPoint.Signed memory);
```
daywiss, line 19: noticed the `view` keyword is gone, how do you know when to use it or not?

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Ah confusing comment I guess, if the condition is met (i.e. amount to pay or collateral pool is 0) it skips to the end of the if statement and basically does nothing.

daywiss (approved):
Looks good to me, I am looking more here to try to follow along than to critique

packages/core/contracts/financial-templates/funding-rate-store/interfaces/FundingRateStoreInterface.sol
nicholaspai: `view` means that the function does not modify contract state. Now that this method has the `publishAndWithdrawProposal` modifier, it can now change the `fundingRecords` state variable and transfer collateral tokens.

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
mrice32: Thoughts on flipping it to work like this:
```js
if (amount.isEqual(0) || collateralPool.isEqual(0)) {
  return;
}

FixedPoint.Unsigned memory amountToPay = (amount.isGreaterThan(collateralPool) ? collateralPool : amount);
...
```

IMO, that makes the non-payment case a little more obvious and it makes the conditional in the if statement a little simpler.

packages/core/contracts/financial-templates/test/MockPerpetual.sol
```diff
@@ -26,7 +32,31 @@ contract MockPerpetual is PerpetualInterface {
...
+    function toggleRevertWithdraw() external {
+        revertWithdraw = !revertWithdraw;
+    }
+
     function withdrawFundingRateFees(FixedPoint.Unsigned memory amount) external override {
+        if (revertWithdraw) {
+            require(false, "set to always reverts");
+        } else {
+            collateralCurrency.safeTransfer(msg.sender, amount.rawValue);
+        }
```
mrice32, line 48:
```suggestion
        require(!revertWithdraw, "set to always reverts");
        collateralCurrency.safeTransfer(msg.sender, amount.rawValue);
```

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: Great point -- I hadn't considered these sorts of attacks. +1. However, I don't _think_ this is a vulnerability (I think there's a minor flaw in your analysis). Will explain in the issue.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: I removed .

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Good point, I'm not gonna add a comment however because there are a lot of internal methods in the Perp system that should also have that same warning, so let's just say its implied for now

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: sure

nicholaspai: Done

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
mrice32: SGTM!

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: As discussed IRL this is not a vulnerability upon further thought as the attacker would not gain marginal net profit

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
```diff
@@ -72,20 +72,20 @@ abstract contract FundingRatePayer is FeePayer, PerpetualInterface {
         );
         FixedPoint.Unsigned memory collateralPool = _pfc();
 
-        // If contract has no PfC, then cannot pay any fees.
-        require(!collateralPool.isEqual(0), "PfC is 0");
-        // The fee must be non-zero.
-        require(!amount.isEqual(0), "Funding fee is 0");
-        // The fee must be < available collateral.
-        require(collateralPool.isGreaterThan(amount), "Funding fee is more than PfC");
+        // If amount to pay or collateral pool is 0, then return early.
```
mrice32, line 75: One minor note I just realized... we don't pay regular fees before paying out funding rate fees. Does this have any weird implications? Previously, the assumption was that you pay regular fees before doing anything to change the pfc, but if we were to do that in this case, the calculation in the FRS would be slightly inaccurate since the regular fees would change the pfc before calling this method.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -239,43 +309,150 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+
+        // Pull reward payment from Perpetual, and transfer (payment + final fee bond + proposal bond) to proposer.
+        // Note: Proposer always receives rebates for their bonds, but may not receive their proposer reward if the
+        // perpetual fails to send fees.
+        IERC20 collateralCurrency = IERC20(PerpetualInterface(perpetual).getCollateralCurrency());
+        FixedPoint.Unsigned memory amountToPay = fundingRateRecord.proposal.finalFee.add(
+            fundingRateRecord.proposal.proposalBond
+        );
+        try PerpetualInterface(perpetual).withdrawFundingRateFees(reward)  {
+            // Only transfer rewards if withdrawal from perpetual succeeded .
```
mrice32, line 432:
```suggestion
            // Only transfer rewards if withdrawal from perpetual succeeded.
```

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -302,6 +479,34 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+
+        FixedPoint.Unsigned memory rewardRate = _getFundingRateRecord(perpetual).rewardRatePerSecond;
+
+        // First compute the reward for the time elapsed.
+        reward = rewardRate.mul(timeDiff);
+
+        // Next scale the reward based on the absolute difference % between the current and proposed rates.
+        FixedPoint.Signed memory diff = (
+            currentRate.isGreaterThan(proposedRate) ? currentRate.sub(proposedRate) : proposedRate.sub(currentRate)
+        );
```
mrice32, line 499: Why are you trying to make the result positive up here if you just handle negatives below anyway (when you compute the absolute value)?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -302,6 +479,34 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+        FixedPoint.Signed memory currentRate
+    ) private view returns (FixedPoint.Unsigned memory reward) {
+        uint256 timeDiff = endTime.sub(startTime);
+
+        FixedPoint.Unsigned memory rewardRate = _getFundingRateRecord(perpetual).rewardRatePerSecond;
+
+        // First compute the reward for the time elapsed.
+        reward = rewardRate.mul(timeDiff);
+
+        // Next scale the reward based on the absolute difference % between the current and proposed rates.
```
mrice32, line 496: Can you write a one-line equation that shows the computation being done below in plain math?

Something like

```
// reward = (proposedRate - currentRate) / currentRate ....
```

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -302,6 +479,34 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+            currentRate.isGreaterThan(proposedRate) ? currentRate.sub(proposedRate) : proposedRate.sub(currentRate)
+        );
+        FixedPoint.Signed memory diffPercent = (currentRate.isEqual(0) ? diff : diff.div(currentRate));
+        FixedPoint.Unsigned memory absDiffPercent = (
+            diffPercent.isLessThan(FixedPoint.fromUnscaledInt(0))
+                ? FixedPoint.fromSigned(diffPercent.mul(FixedPoint.fromUnscaledInt(-1)))
+                : FixedPoint.fromSigned(diffPercent)
+        );
+        // TODO: Should we set ceiling on this
+        reward = reward.mul(absDiffPercent.add(1));
```
mrice32, line 507: I think we should either consider capping this or using a different functional form (maybe not in this PR) -- what's to stop a funding rate that's very close to 0 causing the _next_ reward to be so large that it bankrupts the contract?

mrice32 (approved):
Looks great! I have some lingering concerns around how we're computing payments, but those can be addressed in a follow-up PR if need be!

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: Great point. What if withdrawFundingRateFees(amount) only withdrew the regular-fee-adjusted amount of funding rate fees to the store, and returned the adjusted amount to the caller. And then the Store used this adjusted amount to pay out rewards.

nicholaspai: This would still allow the proposers reward to be a ”% of the PfC” but comments and perhaps variable names should be changed to reflect that this is the "fee adjusted Pfc"

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Agreed this is why I think it makes sense to set a cap, but I'm curious what you have in mind for other functions that could achieve the same effect of "scaling" the proposers reward proportional to the magnitude of the update

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -17,6 +18,39 @@ import "../../../common/implementation/Testable.sol";
 import "../../../common/implementation/Lockable.sol";
 import "../../../common/implementation/FixedPoint.sol";
 
+/**
```
chrismaree, line 21: +1 these comments are great.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: good point fixed

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: `reward = reward * (1 + (proposed - current) / current)`

nicholaspai: Or if `current = 0` then `reward = reward * (1 + (proposed - current))`

nicholaspai: I will add this to the comments

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: For now I've set a 200% ceiling on the diff %

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
mrice32: Thinking more about this, I think both ways of doing this (the way you have it now vs changing it) are equally "correct."

The way you have it now, the regular fees and the proposal fees are both assessed on the "pre-transaction" collateral. That makes the percentages effectively additive. Although it may seem like the pfc we're using is "stale," I think there's a way of looking at it where we're just choosing not to "compound" these deductions. This implies that the calculations are done simultaneously without being ordered. I think that might be the more sensible choice here.

To pay the fees first and then apply the proposal reward percentage would be effectively making the reward a percentage of the post-fee pfc, which means that we've effectively "compounding" the deductions. This would result in a slightly smaller payment. This implies an ordering on the calculations whereby the regular fees are paid first and then the proposal deduction is applied in a later step. This also seems fairly sensible.

I think the only thing we _must_ do is ensure that regular fees are paid during `withdrawFundingRateFees`.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -98,7 +102,7 @@ abstract contract FundingRateApplier is Testable, Lockable {
         return FundingRateStoreInterface(finder.getImplementationAddress("FundingRateStore"));
     }
 
-    function _getLatestFundingRate() internal view returns (FixedPoint.Signed memory) {
+    function _getLatestFundingRate() internal returns (FixedPoint.Signed memory) {
```
mrice32, line 105: Can this cause a re-entrancy lockup?

1. This calls out to the FRS. (re-entrancy locked)
2. The FRS tries to distribute rewards.
3. The FRS calls into this contract to withdraw the fees. (re-entrancy guard tries to lock, but fails since it's already locked)

packages/core/contracts/financial-templates/common/FundingRatePayer.sol
nicholaspai: I agree with your analysis, for implementation simplcity let's add the `fees()` modifier to `withdrawFundingRateFees`

mrice32 (approved):
Very awesome work here! +1

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: FYI we need to change the `withdrawFundingRateRewards` interface to return the actual amount of rewards withdrawn, in case the regular fees reduce the `PfC` such that the caller cannot withdraw their full amount.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Yeah good catch... this causes a reentrancy whenever a Perpetual contract interaction occurs while a pending funding rate proposal has expired.

1. The Perpetual contract asks the FRS for `getFundingRateForContract()`
2. The FRS then sees a pending proposal has expired and attempts to disperse proposer rewards via `perpetual.withdrawFundingRateFees()`. This re-enters the perpetual contract.

I propose that we convert `getFundingRateForContract()` into a view method that does NOT attempt to disperse proposer rewards.

nicholaspai:
Thanks to feedback from @mrice32 I've pushed some important changes:

- `getFundingRateForContract` is now a view-only method on the FundingRateStore interface. This is done so that the PerpetualContract can call this method without fear of re-entrancy. Previously, this method also attempted to withdraw funding rate rewards from the Perpetual contract but this would cause re-entrancy if `getFundingRateForContract` was triggered by a Perpetual contract interaction. This means that the only way to disperse funding rate proposal rewards is to call the FundingRateStore directly.

- `withdrawFundingRateFees` now pays DVM regular fees via the `fees()` modifier. This enforces that any collateral pulled out of the contract follows any pending regular fees owed. One side effect of this is that it is now possible that the desired withdrawal amount gets reduced by taxed regular fees, so `withdrawFundingRateFees` now returns the actual amount of collateral withdrawn post-regular-fees.

[Commit link](https://github.com/UMAprotocol/protocol/pull/2168/commits/c29d2d0767fa4af0c256b2d100adcde1f6370084)

mrice32 (approved):
LGTM!

PR #2172, opened 2020-11-10
feat(affiliates): add extra data to deployer reward results

### LINKED ISSUE #2171: Add additional data about emp weights to deployer reward output
Currently the output of deployer rewards only shows data based on how much each deployer earns. We may want to give additional data about how much each EMP earned as well as a way to weight further incentive rewards plans down stream, ie "siloing" rewards earned by an emp for liquidity mining for that emp.

daywiss:
**Motivation**

#2171

**Summary**

- adds extra data to allow for "siloing" of rewards per emp
- also adds start / end block timestamp and numbers

image

**Issue(s)**

Fixes #2171

### DIFF
--- a/packages/affiliates/apps/CreateDataSet.js
+++ b/packages/affiliates/apps/CreateDataSet.js
@@ -22,4 +22,5 @@ async function Run(config) {
 const config = Config();
 Run(config)
   .then(console.log)
-  .catch(console.error);
+  .catch(console.error)
+  .finally(() => process.exit());
--- a/packages/affiliates/libs/affiliates.js
+++ b/packages/affiliates/libs/affiliates.js
@@ -108,11 +108,14 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
 
     balanceHistories = new Map(balanceHistories);
     empDeployers = new Map(empDeployers);
+    let startBlock, endBlock;
 
     const rewardsPerBlock = toBN(toWei(totalRewards)).div(toBN(blocks.length));
     const payoutPerSnapshot = rewardsPerBlock.mul(toBN(snapshotSteps));
     const valuePerSnapshot = blocks.reduce((result, block, index) => {
       if (index % snapshotSteps !== 0) return result;
+      if (startBlock == null) startBlock = block;
+      endBlock = block;
       const { timestamp } = block;
       const valueByEmp = empWhitelist.reduce((result, empAddress, empIndex) => {
         try {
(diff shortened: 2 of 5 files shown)


packages/affiliates/libs/affiliates.js
```diff
@@ -144,32 +147,49 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
...
+          // calculate per emp rewards
+          if (empPayouts[emp] == null) empPayouts[emp] = toBN("0");
+          empPayouts[emp] = empPayouts[emp].add(rewards);
+        });
 
-      return result;
-    }, {});
+        return { deployerPayouts, empPayouts };
+      },
+      { deployerPayouts: {}, empPayouts: {} }
```
daywiss, line 176: this just adds a counters for deployer and emppayouts and seperate objects. None of the math has changed

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -93,12 +95,25 @@ describe("DeployerRewards", function() {
...
+    assert.equal(Object.keys(result.empPayouts).length, 3); // There should be 3 emps
+
+    assert.isBelow(
+      // compare floats with an epsilon
+      Math.abs(
+        Object.values(result.deployerPayouts).reduce((total, value) => {
+          return Number(total) + Number(value);
+        }, 0) - Number(devRewardsToDistribute)
+      ),
+      epsilon
```
daywiss, line 108: do float check against epsilon because sometimes the math ends up being off by like .000000000001

packages/affiliates/apps/CreateDataSet.js
```diff
@@ -22,4 +22,5 @@ async function Run(config) {
 const config = Config();
 Run(config)
   .then(console.log)
-  .catch(console.error);
+  .catch(console.error)
+  .finally(() => process.exit());
```
daywiss, line 26: for some reason process does not exit. I think web3 or bigquery is holding a connection open but dont know how to close them. So this forces exit.

packages/affiliates/libs/affiliates.js
```diff
@@ -144,32 +147,49 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
...
+          const contribution =
+            totalValueLocked.toString() != "0"
+              ? toBN(value) // eslint-disable-line indent
+                  .mul(toBN(toWei("1"))) // eslint-disable-line indent
+                  .div(totalValueLocked) // eslint-disable-line indent
+              : toBN("0"); // eslint-disable-line indent
+          const rewards = contribution.mul(payoutPerSnapshot).div(toBN(toWei("1")));
+
+          // calculate deployer rewards
+          if (deployerPayouts[deployer] == null) deployerPayouts[deployer] = toBN("0");
```
nicholaspai, line 166: I notice you check if objects are explicitly equal to `null`, why not just shorten the if statements to `if(!object)` instead? Also you should use javascript's triple equal signs

nicholaspai (approved):
Really like the idea of siloing data by EMP, LGTM

packages/affiliates/libs/affiliates.js
daywiss: Yea that `if(!object)` would be equivalent, I can consider changing to that, I just do it this way out of habit. 
I use `==` rather than `===` in this case because I want to check for null or undefined (falsey values). Not the best because it will also catch 0s. I  agree in general though `===` is preferred.

packages/affiliates/test/mocha/affiliates.js
mrice32: Do you know why this error occurs? Is it because individual payouts are rounded down by BN?

mrice32 (approved):
LGTM!

packages/affiliates/test/mocha/affiliates.js
daywiss: I dont know exactly but i think it has to do with floating point math. Theres a division in there and numbers end up being like .9 repeated rather than going evenly into 1.

Issue #2170, opened 2020-11-10
Test run deployer rewards for a weeks worth of data


daywiss:
Do an e2e test and see output for a weeks worth of rewards as if we are doing a real run.

start: 2020-11-02 23:00:00 UTC  (11180530) 
end: 2020-11-10 23:00:00 UTC (11226178)

empContracts: 
- 0xaBBee9fC7a882499162323EEB7BF6614193312e3  uUSDrBTC-DEC
- 0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56  uUSDrWETH-DEC
- 0x306B19502c833C1522Fbc36C9dd7531Eda35862B pxUSD-OCT2020

daywiss:
output from these timestamps for 50k rewards:
![image](https://user-images.githubusercontent.com/4429761/98757339-53098280-239a-11eb-99aa-8194be223a00.png)
Block numbers are slightly off, will probably need to look into that

PR #2174, opened 2020-11-11
fix(affiliates): when getting balances use all time logs instead of logs from start time

### LINKED ISSUE #2173: EMP balances need different start time than specified reward period
This was an oversight, but in order to calculate balances correctly, we need all events from the start of the emp contract. Currently the script uses the start time provided by the user, which may be well after EMP was deployed. 

Script needs to correctly deduce start of emp, and use that as the start time for events to calculate user balances even if reward period starts well after that.

daywiss:
**Motivation**
#2173 

**Summary**
Previously when running script for some time frames, it would fail with incorrect or negative balances. This is because we need to generate balances from start of each emp deployment. 

This fix allows us to run deployer rewards for any time period that does not encompass the emp creation time. This brute forces the query from bigquery, it has to search through all time to get these logs and costs about 25 cents per contract.

- gets balances and emp deployer events for all time vs respecting user start time. We need balances to generate from all time, but user may specify a start time that does not meet this requirement.
- fixes bug when handling events that call settle expired. apparently users without positions can call this.

**Issue(s)**

Fixes #2173, #2170

### DIFF
--- a/packages/affiliates/libs/affiliates.js
+++ b/packages/affiliates/libs/affiliates.js
@@ -16,11 +16,11 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
   assert(empAbi, "requires empAbi");
   assert(coingecko, "requires coingecko api");
   assert(synthPrices, "requires synthPrices api");
-  async function getBalanceHistory(address) {
+  async function getBalanceHistory(address, start, end) {
     // stream is a bit more optimal than waiting for entire query to return as array
     // We need all logs from beginning of time. This could be optimized by deducing or supplying
     // the specific emp start time to narrow down the query.
-    const stream = await queries.streamAllLogsByContract(address);
+    const stream = await queries.streamLogsByContract(address, start, end);
     const decode = DecodeLog(empAbi);
     const balancesHistory = EmpBalancesHistory();
     await highland(stream)
@@ -219,7 +219,9 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
       getBlocks(startTime, endTime),
       // these are block events, and they ignore start/end time as we need all events from start of each contract
       getEmpDeployerHistory(empCreatorAddress),
-      getAllBalanceHistories(empWhitelist)
+      // TODO: optimize this to get start time from the results of getEmpDeployerHistory. Will be different for
+      // each emp, so getAllBalanceHistories won't be able to be used.
+      getAllBalanceHistories(empWhitelist, 1, endTime)
     ]);
 
     return calculateRewards({
--- a/packages/affiliates/libs/affiliates.js
+++ b/packages/affiliates/libs/affiliates.js
@@ -16,9 +16,11 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
   assert(empAbi, "requires empAbi");
   assert(coingecko, "requires coingecko api");
   assert(synthPrices, "requires synthPrices api");
-  async function getBalanceHistory(address, start, end) {
+  async function getBalanceHistory(address) {
     // stream is a bit more optimal than waiting for entire query to return as array
-    const stream = await queries.streamLogsByContract(address, start, end);
+    // We need all logs from beginning of time. This could be optimized by deducing or supplying
+    // the specific emp start time to narrow down the query.
+    const stream = await queries.streamAllLogsByContract(address);
     const decode = DecodeLog(empAbi);
     const balancesHistory = EmpBalancesHistory();
     await highland(stream)
@@ -66,14 +68,19 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
   }
 
   // returns array of tuples [emp address, deployer address]
-  async function getEmpDeployerHistory(address, start, end) {
-    // this query is relatively small
-    const logs = await queries.getLogsByContract(address, start, end);
+  async function getEmpDeployerHistory(address) {
+    // this query is relatively small but expensive, gets all logs from begginning of time
+    const logs = await queries.getAllLogsByContract(address);
     const decode = DecodeLog(empCreatorAbi);
-    return logs.map(decode).reduce((result, log) => {
-      result.push([log.args.expiringMultiPartyAddress, log.args.deployerAddress]);
-      return result;
-    }, []);
+    return logs
+      .map(log => decode(log, log))
+      .reduce((result, log) => {
+        result.push([
+          log.args.expiringMultiPartyAddress,
+          { deployer: log.args.deployerAddress, timestamp: log.block_timestamp, number: log.block_number }
+        ]);
+        return result;
+      }, []);
   }
 
   // Calculates the value of x `tokens` in USD based on `collateralPrice` in USD, `syntheticPrice` in collateral considering
@@ -153,7 +160,7 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
           return result.add(toBN(value));
         }, toBN("0"));
         valueByEmp.forEach(([emp, value]) => {
-          const deployer = empDeployers.get(emp);
+          const { deployer } = empDeployers.get(emp);
           const contribution =
             totalValueLocked.toString() != "0"
               ? toBN(value) // eslint-disable-line indent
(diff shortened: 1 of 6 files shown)


packages/affiliates/libs/affiliates.js
```diff
@@ -16,9 +16,11 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
   assert(empAbi, "requires empAbi");
   assert(coingecko, "requires coingecko api");
   assert(synthPrices, "requires synthPrices api");
-  async function getBalanceHistory(address, start, end) {
+  async function getBalanceHistory(address) {
     // stream is a bit more optimal than waiting for entire query to return as array
-    const stream = await queries.streamLogsByContract(address, start, end);
+    // We need all logs from beginning of time. This could be optimized by deducing or supplying
+    // the specific emp start time to narrow down the query.
+    const stream = await queries.streamAllLogsByContract(address);
```
daywiss, line 23: This is the main logical change here, events to generate balances are queried for all time ( to get all contract events) vs previously used start and end time.

packages/affiliates/libs/affiliates.js
```diff
@@ -66,14 +68,19 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
...
-    return logs.map(decode).reduce((result, log) => {
-      result.push([log.args.expiringMultiPartyAddress, log.args.deployerAddress]);
-      return result;
-    }, []);
+    return logs
+      .map(log => decode(log, log))
+      .reduce((result, log) => {
+        result.push([
+          log.args.expiringMultiPartyAddress,
+          { deployer: log.args.deployerAddress, timestamp: log.block_timestamp, number: log.block_number }
```
daywiss, line 80: this just adds additional meta data as we processs the deployer logs. Could be used to optimize balance queries by getting deployment time for each emp contract. Currently though its not used.

packages/affiliates/libs/affiliates.js
```diff
@@ -203,17 +209,19 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
...
-    const syntheticTokenPrices = await Promise.map(
-      empWhitelist,
-      async address => await getSyntheticPriceHistory(address, startTime, endTime)
-    );
-    const blocks = await getBlocks(startTime, endTime);
-    const balanceHistories = await getAllBalanceHistories(empWhitelist, startTime, endTime);
-    const empDeployers = await getEmpDeployerHistory(empCreatorAddress, startTime, endTime);
+    // query all required data ahead of calcuation
+    const [collateralTokenPrices, syntheticTokenPrices, blocks, empDeployers, balanceHistories] = await Promise.all([
+      Promise.map(
```
daywiss, line 214: small optimization to wrap these awaits in promise.all

packages/affiliates/libs/bigquery.js
```diff
@@ -8,7 +8,21 @@ module.exports = ({ client } = {}) => {
 
   // types of queries we are interested in
   const queries = {
-    // logs aka events by contract
+    // this is an expensive call (50gb at time of writing), but is necessary to get all logs without knowing
+    // the deployment time of a contract.
+    allLogsByContract(contract, selection = ["*"]) {
```
daywiss, line 13: added another bigquery function to get logs without a date range, vs logsByContract which you specify start and end time

packages/affiliates/libs/datasets.js
```diff
@@ -133,6 +133,17 @@ function Dataset(basePath, { queries, coingecko, synthPrices }) {
         .on("close", res);
     });
   }
+  function allLogs({ contract, select }, path) {
```
daywiss, line 136: update dataset generation as well to use all time event logs

packages/affiliates/libs/datasets.js
```diff
@@ -257,6 +271,14 @@ function MockQueries(basePath) {
...
       .toPromise(Promise);
   }
+  function getAllLogsByContract(address) {
+    return streamLogsByContract(address)
+      .collect()
+      .toPromise(Promise);
+  }
+  function streamAllLogsByContract(address) {
+    return streamLogsByContract(address);
+  }
```
daywiss, line 281: adds in mock functions to mimick the bigquery library

packages/affiliates/libs/processors.js
```diff
@@ -147,9 +147,23 @@ function EmpBalances(handlers = {}, { collateral, tokens } = {}) {
...
+          // ignore balance below 0 error
+        }
+      }
+      if (tokens.has(caller)) {
+        try {
+          tokens.sub(caller, tokensBurned.toString());
+        } catch (err) {
+          // ignore balance below 0 error
+        }
+      }
```
daywiss, line 166: convoluted logic, but when emps expire anyone without a position can settle. This was causing balances to go negative and throw errors. So now we just swallow the error, though maybe this could be done better.

packages/affiliates/libs/affiliates.js
nicholaspai: Is there any reason why you can't manually set an end time, even if you must now default start time to deployment time?

daywiss: Could definitely be done that way, and I would consider that as an improvement/optimization to this. Really just wanted to get a fix in quickly without too much additional complexity though, because getting the start time of each emp was a bit ugly.

[made an issue for your suggestion](https://github.com/UMAprotocol/protocol/projects/6#card-49245181)

nicholaspai (approved):
This makes sense, but why can't you specify end date? As long as you default start date to deployment time I think it should be OK.

packages/affiliates/libs/affiliates.js
nicholaspai: Got it, yeah you can always add it later if you need it.

daywiss:
> This makes sense, but why can't you specify end date? As long as you default start date to deployment time I think it should be OK.

I can specify end date, i just didn't want to add a weird function where it was not taking a start date but taking an end date for the query. But i can imitate this by reusing the function which takes start/end and specify start time as something small. I will make that change.

packages/affiliates/libs/processors.js
mrice32: Hmmm, interesting case. Have you thought about querying only blocks _up to_ expiry? Have we decided whether to include expired contracts in developer mining?

packages/affiliates/scripts/example-balances.js
```diff
@@ -1,26 +1,27 @@
...
-const start = moment("2020-9-20", "YYYY-MM-DD").valueOf();
-const end = moment("2020-10-10", "YYYY-MM-DD").valueOf();
+const { getAbi } = require("@uma/core");
+const { EmpBalancesHistory, EmpBalances } = require("../libs/processors");
+
+const abi = getAbi("ExpiringMultiParty");
+// sanity check various contracts
+// const empContract = "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56";
+// const empContract = "0xaBBee9fC7a882499162323EEB7BF6614193312e3"
+// const empContract =   "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56"
```
mrice32, line 12: nit: did you mean to leave this comment here? If so, maybe we should label the different addresses for clarity?

packages/affiliates/libs/affiliates.js
```diff
@@ -66,14 +68,19 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
   }
 
   // returns array of tuples [emp address, deployer address]
-  async function getEmpDeployerHistory(address, start, end) {
-    // this query is relatively small
-    const logs = await queries.getLogsByContract(address, start, end);
+  async function getEmpDeployerHistory(address) {
+    // this query is relatively small but expensive, gets all logs from begginning of time
+    const logs = await queries.getAllLogsByContract(address);
```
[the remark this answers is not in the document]
daywiss: about 25 cents per contract, 50+ gb.

packages/affiliates/libs/processors.js
daywiss: > Have you thought about querying only blocks up to expiry?

I have not.  Would be a bit more complicated to handle that though.

> Have we decided whether to include expired contracts in developer mining?

Theres been no decision around that, the calculation currently has no notion of an expired contract so it will continue to weight them for rewards.  That is something we should probably figure out though!

chrismaree (approved):

PR #2178, opened 2020-11-11
fix(affiliates): fix bigquery bug where timestamps computed incorrectly

### LINKED ISSUE #2177: Block numbers differ from liquidity mining when querying for dates
look into issue where block numbers differ from liquidity mining script when using the same timestamp. Dev mining uses bigquery while LM uses a search in web3.

daywiss:

**Motivation**

#2177 

**Summary**
This ends up being more accurate than the liquidity mining script, since big query does a linear search over all blocks.

- found bug with bigquery timestamp calculation. Now correctly converts MS timestamp to bigquery representation
- adds a script to compare results with the liquidity mining block timestamp finder
- regenerated dataset with fixes

**Issue(s)**

Fixes #2177


packages/affiliates/libs/bigquery.js
```diff
@@ -12,17 +11,15 @@ module.exports = ({ client } = {}) => {
...
       // require an array of values
       selection = lodash.castArray(selection);
       return `
         SELECT ${selection.join(", ")}
         FROM
           bigquery-public-data.crypto_ethereum.logs
         WHERE
-          block_timestamp >= TIMESTAMP('${start}')
-          AND block_timestamp < TIMESTAMP('${end}')
+          block_timestamp >= TIMESTAMP_MILLIS(${start})
```
mrice32, line 21: Can you briefly explain what the issue was? Is this a BQ native function? What are the inputs and outputs?

mrice32 (approved):
Looks great!

packages/affiliates/libs/bigquery.js
daywiss: Yes after running the emp rewards script with the same date as liquidity mining, there were really wildy different block numbers returned from each script for start/end blocks. They are both doing block number lookups from time with different techniques. It turns out that the bigquery time was being interpreted as AM, while the LM script was properly looking up by PM.  This was an error in how we were formatting the date/time before submitting it to bigquery. The fix was to remove timestamp conversion to a string format, and use BQ TIMESTAMP_MILLIS to interpret within BQ instead.

packages/affiliates/libs/bigquery.js
```diff
@@ -12,17 +11,15 @@ module.exports = ({ client } = {}) => {
     logsByContract(contract, start, end = Date.now(), selection = ["*"]) {
       assert(contract, "requires contract");
       assert(start, "requires start");
-      start = moment(start).format("YYYY-MM-DD hh:mm:ss");
```
chrismaree, line 15: I suspect the issue was in how this is formatted. in the LM scripts it's defined as:

```
argv.dateTime, "YYYY-MM-DD  HH:mm Z"
```

see: https://github.com/UMAprotocol/protocol/blob/a0f038ca961df8e00050a57542228da89294575e/packages/core/scripts/liquidity-mining/FindBlockAtTimeStamp.js#L13

In any case I'm happy with doing it with tmiestamps as well as you've done in your PR

chrismaree (approved):
looks good to me! doing it as a js timestamp is better as is more consistent and avoids any strange time zoning or conversion issues.

daywiss:
this is going to depend on #2174 going through. It handles some new events this currently will throw on during testing

packages/affiliates/libs/bigquery.js
mrice32: I see. +1

packages/affiliates/libs/bigquery.js
mrice32: Agreed. SGTM

PR #2181, opened 2020-11-11
feat(core): add legacy versioning

### LINKED ISSUE #2180: Allow users to access legacy abis/addresses via the core package
**Why**

Offchain infra may need to be able to support multiple versions of financial contracts simultaneously.

**How**

We should be able to alias old `@uma/core` versions using aliases in the package.json file:

```
"@uma/core-1-1-0": "npm:@uma/core@1.1.0"
```

**Resolution**

What is the deliverable to consider this task done?

**Difficulty Score [1-10]**

- 5: ~4 hours

mrice32:
**Motivation**

#2180.

**Summary**

This just adds an arg to all core index.js functions to take a version string.

**Issue(s)**

Fixes #2180.

### DIFF
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -3,7 +3,17 @@ const fs = require("fs");
 const path = require("path");
 const { getWeb3 } = require("@uma/common");
 
+/**
+ * @notice Gets the directory for version of core specified by an input version string.
+ * @param {String} version Version string in the form of x.y.z.
+ */
 function getDirectoryForVersion(version) {
+  // Note: this establishes a convention where any previous core version that is pulled in here must be aliased in the
+  // package.json file as follows:
+  // "@uma/core-x-y-z": "npm:@uma/core@x.y.z"
+  // This forces yarn to pull the package @uma/core with version x.y.z into a local package name/folder with name
+  // @uma/core-x-y-z.
+  // To reiterate: any version passed in here must be listed in the package.json.
   const packageName = `@uma/core-${version.split(".").join("-")}`;
   return path.dirname(require.resolve(`${packageName}/package.json`));
 }
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -98,7 +98,7 @@ function getTruffleContractTest(contractName, web3, version = "latest") {
  * @param {String} [version] version identifier x.y.z for the contract. Defaults to "latest".
  */
 function getAddressTest(contractName, networkId, version = "latest") {
-  const truffleContract = getTruffleContractTest(contractName, version);
+  const truffleContract = getTruffleContractTest(contractName, undefined, version);
 
   if (truffleContract.networks[networkId]) {
     return truffleContract.networks[networkId].address;
--- a/packages/core/index.js
+++ b/packages/core/index.js
@@ -3,12 +3,19 @@ const fs = require("fs");
 const path = require("path");
 const { getWeb3 } = require("@uma/common");
 
+function getDirectoryForVersion(version) {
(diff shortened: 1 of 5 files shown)


packages/core/index.js
```diff
@@ -3,12 +3,19 @@ const fs = require("fs");
 const path = require("path");
 const { getWeb3 } = require("@uma/common");
 
+function getDirectoryForVersion(version) {
+  const packageName = `@uma/core-${version.split(".").join("-")}`;
```
nicholaspai, line 7: Does this mean we would create a `core-1` directory containing older contracts...for example?

nicholaspai (approved):

packages/core/index.js
mrice32: I can add a comment on this because it's confusing. It basically resolves the directory containing the npm package `@uma/core-x-y-z`, where the version is `x.y.z`. So, this establishes a convention that we alias relevant core versions in the package.json as `@uma/core-x-y-z` as I've done in this PR with `1.1.0`. Does that make sense?

nicholaspai: ah very cool

packages/core/index.js
```diff
@@ -3,12 +3,19 @@ const fs = require("fs");
...
 /**
  * @notice Gets the truffle artifact for an UMA contract.
  * @param {String} contractName Name of the UMA contract whose artifact object will be returned.
+ * @param {String} [version] version identifier x.y.z for the contract. Defaults to "latest".
  */
-function getArtifact(contractName) {
-  const filePath = path.join(__dirname, "build", "contracts", `${contractName}.json`);
+function getArtifact(contractName, version = "latest") {
+  const contractDirectory = version === "latest" ? __dirname : getDirectoryForVersion(version);
+  const filePath = path.join(contractDirectory, "build", "contracts", `${contractName}.json`);
```
chrismaree, line 18: One nice artifact of using the old build directories (and this technique in general) is that we still have access to old contract bytecode meaning we could use this same technique to deploy old contracts (even ones that we don't have the `.sol` for in the `contracts` directory.

packages/core/index.js
```diff
@@ -3,12 +3,19 @@ const fs = require("fs");
 const path = require("path");
 const { getWeb3 } = require("@uma/common");
 
+function getDirectoryForVersion(version) {
```
chrismaree, line 6: how does this function handle if there is no `package.json` for the associated version specified? this function is called by `getArtifact` which is called in a few places such as `getAbi`. what if I put a version that is not defined in the base `package.json`?

chrismaree: Nvm I see that you test this in the mocha tests.

chrismaree (approved):

packages/core/index.js
mrice32: Just for posterity, if the version is bad OR the file name is bad, it just throws.

packages/core/index.js
mrice32: Yep! This might be really nice for testing backcompat.

daywiss (approved):

packages/core/index.js
mrice32: Ahhh, this is a great point. However, I don't think this index.js file will work at all in frontends right now due to the `path` import. I think the abis have to be pulled in manually regardless on the frontend (which you can do by importing the `.json` files directly). I think it makes sense to just leave this be and then figure out a separate method for the frontends down the road as you suggest. +1. Thanks for making this point.

PR #2182, opened 2020-11-12
improve: adds price identifier BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20 to whitelist in IdentifierWhitelist

apeir99n:
**Motivation**

This PR adds price identifier to get prices for BTC Quarterly Futures from https://cryptowat.ch/charts/BINANCE:BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20.

### DIFF
--- a/packages/core/config/identifiers.json
+++ b/packages/core/config/identifiers.json
@@ -1,6 +1,6 @@
 {
   "BTC/USD": {},
-  "BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20": {}
+  "BTC-FUTURE-QUARTERLY/USD": {}
   "USD/BTC": {},
   "ETH/USD": {},
   "CMC Total Market Cap": {},
--- a/packages/core/config/identifiers.json
+++ b/packages/core/config/identifiers.json
@@ -1,5 +1,6 @@
 {
   "BTC/USD": {},
+  "BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20": {}
   "USD/BTC": {},
   "ETH/USD": {},
   "CMC Total Market Cap": {},


packages/core/config/identifiers.json
```diff
@@ -1,5 +1,6 @@
 {
   "BTC/USD": {},
+  "BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20": {}
```
nicholaspai, line 3: Out of curiosity are you adding this to aid your testing? I ask because `BTC-USD-FUTURE-INVERSE-QUARTERLY-25DEC20` is not approved on the Mainnet `IdentifierWhitelist` yet. If you just need it for testing locally, then you can add to this file, migrate the contracts, and it will be whitelisted via the migration scripts. If you want this identifier whitelisted on the Kovan `IdentifierWhitelist` listed in `core/networks/42.json` then I can add it for you.

apeir99n: I've already tested it locally. Now want to test it on Kovan. Would be great if you can add it to Kovan IdentifierWhitelist

nicholaspai: I'm getting this error

```
ERROR: Error: invalid bytes32 value (arg="identifier", coderType="bytes32", value="0x4254432d5553442d4655545552452d494e56455253452d515541525445524c592d32354445433230")
```

Generally the identifier should not have any dates in it, so I'd suggest you just use the `USDBTC` price since it is (1) the inverse of BTCUSD and (2)  you can make it "quarterly" by setting its expiration.

mrice32: The identifiers are stored in `bytes32`, so they can't be more than ~32 characters long.

mrice32: But we're happy to whitelist on kovan! Let us know an alternative name for the identifier and we'll get that figured out for you!

apeir99n: ok. changed identifier name.

apeir99n: I just have one question: this price identifier is then used in CryptoWatchPriceFeed.js to get price from CryptoWatch, right?  What type of CryptoWatchPriceFeed.pair ? It should be bytes32 too or not?

nicholaspai: The bot would just grab the BTCUSD price on Cryptowatch. This is identical to how contracts with the BTCUSD or USDBTC identifirs would work, which is why I think you should just use that identifier when testing. If you were to deploy your product to mainnet we'd recommend that you use those identifiers anyways.

mrice32:
Closing due to inactivity. Happy to reopen if you feel we should reconsider!

Issue #2160, opened 2020-11-06
Refactor `_getOraclePrice` and `_getCollateralRequirement` into external libraries, allow contract factories to have customizable constructor params


nicholaspai:
**Why**

Allows EMP deployers to modify how oracle price and collateral requirements are calculated, s.t. different types of financial contracts besides vanilla futures contracts can be created. 

Currently the EMP always uses the reported DVM price as its settlement price (relevant on settlements at expiry and disputes) and hardcodes its collateral.

But what if you wanted to create the following Expiring product:
- The contract requires sponsors to deposit 1 WETH in collateral and allows them to mint (and distribute) tokens that pay out that 1 WETH if, at expiry, ETHUSD is below a set strike. If ETHUSD is above that strike, the contract pays out a given dollar amount of WETH.
- Example: expiry is DEC 31. Strike is $400. Each token is backed by 1 WETH
- If ETHUSD < $400 at expiry, token is redeemed for 1 ETH.
- If ETHUSD >= $400 at expiry, token is redeemed for $400 worth of ETH, as determine by the oracle price returned at that time.

## `_getOraclePrice`
You could modify the existing EMP by changing the `_getOraclePrice()` method to:
```
oraclePrice = _getOraclePrice(expirationTimestamp);
// if price request is made before expiry, return 1. This means that we can keep
// the contract 100% collateralized with 1 WETH pre-expiry, and that disputes 
// pre-expiry are illogical (aka token is always backed by 1 WETH pre-expiry)
if (requestedTime < expirationTimestamp) {
     return 1;
}
if (oraclePrice < strike) {
  expiryPrice = 1;   // token expires to be worth 1 WETH
}
else {
  // token expires to be worth strike $ worth of WETH
  // eg if ETHUSD is $500 and strike is $400, token is
  // redeemable for 400/500 = 0.8 WETH
  expiryPrice = strike / _getOraclePrice(expirationTimestamp);
}
```

## `_getCollateralRequirement`
Currently EMP disputes enforce the following `requiredCollateral`:
```
 FixedPoint.Unsigned memory requiredCollateral = tokenRedemptionValue.mul(collateralRequirement);
```

But what's interesting about this contract is that the collateral required should decrease as ETHUSD goes up
```
ETHUSD = $350, you get 1 WETH
ETHUSD = $375, you get 1 WETH
ETHUSD = $400, you get 1 WETH
ETHUSD = $425, you get 0.941 WETH (worth $400)
[...]

PR #2185, opened 2020-11-16
feat(emp): financial product library to apply price transformation

chrismaree:
**Motivation**

issue #2160

Add the ability to arbitrarily override financial product logic, such as applying a transformation function to the oracle price to enable custom contracts to be built on the EMP, such as a structured node.

**Summary**

a) Adds a `FinancialProductLibrary` abstract contract that can be used to override EMP (or perpetual) prices returned from the DVM or the collateral requirement of a position based on the specific financial product.

b) Adds unit tests to `PricelessPositionManager.js` to showcase basic price transformation integration using a simple test `FinancialProductLibrary` that linearly scales input prices.

c) Adds an example `FinancialProductLibrary` called `StructureNoteFinancialProductLibrary` which adds the logic to create a structured note on top of the EMP. This file is meant to showcase how this pattern is intended on being used. Associated unit tests are also added.

d) updates associated contracts to pass down the `financialProductLibraryAddressAddress`.

**Details**
The implementation of the `FinancialProductLibrary` was designed to be as simple as possible by simply taking in the `oraclePrice` or `collateralRequirement` and returning a transformed price/collaterlRequirement. No additional params are provided & these functions are both `view`. Based on the implementation of each custom `FinancialProductLibrary`, the library can then pull additional info from the EMP to compute the transformation function. This makes the `FinancialProductLibrary`'s stateful (as they can store info like strike prices) and keeps all additional logic outside of the domain of the emp and withing the `FinancialProductLibrary`. An example of this is the setting of strike prices in the structured node library. 

One important consideration of this design pattern is that it places a LOT of trust in the financial product library. a malicious deployer could do _many_ different attack vectors by manipulating the oracle settlement price of the collateral requirement to steal the contract collateral. The pattern in this PR provides no assurances to the EMP that the library implemented is safe & implements the correct logic. 

One solution to this is to create a `FinancialProductLibraryFinder` that contains only approved & vetted financial products for EMPs to use. I'm curious what others think about this pattern.

This PR implements the `transformPrice` part of the `FinancialProductLibrary`. A second PR will address the `transformCollateralRequirement` once this pattern is approved by engineering.

**Issue(s)**

first PR of issue #2160


packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,81 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "../../../common/implementation/Testable.sol";
+
+import "./FinancialProductLibrary.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+
+interface ExpiringMultiParty {
```
nicholaspai, line 10: Should this contract be in a test folder?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,81 @@
...
+    constructor(address _timerAddress) public Testable(_timerAddress) {}
+
+    /**
+     * @notice Enables the deployer of the library to set the strike price for an associated financial product.
+     * @param financialProduct address of the financial product.
+     * @param strikePrice the strike price for the structured note to be applied to the financial product.
+     * @dev Note: a) A strike price can not be 0. b) A strike price can only be set once to prevent the deployer from
+     * changing the strike after the fact. c)  financialProduct must exposes an expirationTimestamp method.
+     */
+    function setFinancialProductStrike(address financialProduct, FixedPoint.Unsigned memory strikePrice)
```
nicholaspai, line 36: nit: would `emp` be a better name than `financialProduct` since we are assuming that these are all EMP's?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,81 @@
...
+        public
+        onlyOwner
```
nicholaspai, line 38: Who would be this owner?

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -667,6 +676,22 @@ contract PricelessPositionManager is FeePayer, AdministrateeInterface {
         return _getFeeAdjustedCollateral(rawTotalPositionCollateral);
     }
 
+    /**
+     * @notice Accessor method to calculate a transformed price using the provided finanicalProductLibrary specified
+     * during contract deployment. If no library was provided then no modification to the price is done.
+     * @param price input price to be transformed.
```
nicholaspai, line 682: Is it fair to add a `@dev` note that this method should never revert?

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -667,6 +676,22 @@ contract PricelessPositionManager is FeePayer, AdministrateeInterface {
...
+     * @return transformedPrice price with the transformation function applied to it.
+     */
+
+    function transformPrice(FixedPoint.Unsigned memory price) public view returns (FixedPoint.Unsigned memory) {
```
nicholaspai, line 686: Why can't this be an internal method and removed from the interface?

packages/core/contracts/financial-templates/test/ExpiringMultiPartyMock.sol
```diff
@@ -0,0 +1,31 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "../../common/implementation/FixedPoint.sol";
+
+
+interface FinancialProductLibrary {
```
nicholaspai, line 7: Maybe I'm missing something obvious but why aren't you using the actual `financial-templates/common/financial-product-libraries/FinancialProductLibrary.sol` contract here?

nicholaspai (approved):
Pattern makes sense to me and is well implemented with a clear example, I'd like to not add to the EMP's interface if possible though.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
mrice32: +1

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -667,6 +676,22 @@ contract PricelessPositionManager is FeePayer, AdministrateeInterface {
...
+        if (address(financialProductLibrary) == address(0)) return price;
+        try financialProductLibrary.transformPrice(price) returns (FixedPoint.Unsigned memory transformedPrice) {
```
mrice32, line 688: One question that I don't have a strong opinion on: should we _require_ that the user provides a valid financial product library and remove this branching logic (even if they just use a default one that does no transform)?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
chrismaree: I don't think so. It's not a test. It's an actual Financial product library that implements a desired kind of product. This is why I put it in this new directory `financial-product-libraries`. I invasion we'll have a collection of different libraries in this directory to do a bunch of custom overrides, this is just the first for this.

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
chrismaree: It _could_ be but that's somewhat oppinionated. It's possible to implement other kinds of contracts. Also, the wording through this contract refers to is as a "Financial Product" rather than "EMP". I'm happy to change it through if you prefer. WDYT?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
chrismaree: The owner is the deployer of the Library who is the same address who deploys the EMP. The deployer is the only one permissioned to set the strikes, just like the deployer is the only one permissioned to configure the emp at deployment time. I will add a comment to this to make it clearer.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
chrismaree: done

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
chrismaree: I left this as a public method as it's pretty useful to be able to transform the price as a caller. For example, the front end of `emp-tools` would need to transform the price based on this method to "see" what the current price is, based off `emp-tools` belief of the current price.

We could leave this as internal but then calling this method becomes a bit of a pain. you'd need to then need to instantiate an instance of the library and then call this method from the context of the EMP to get the transformed price.

packages/core/contracts/financial-templates/test/ExpiringMultiPartyMock.sol
chrismaree: No real reason. I've updated to use that import.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
chrismaree: We could do this, especially if we have a "default" public one that simply returns the price with no transformation logic applied. However, this amounts to the exact same thing that this PR implements, but adds a higher barrier as now you need to know this address. Also, this would require us to validate the library at emp deployment.

Curious to hear others thoughts @nicholaspai

mrice32 (commented):
Flushing some more comments. Very very cool stuff!!

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,82 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "../../../common/implementation/Testable.sol";
+
+import "./FinancialProductLibrary.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+
```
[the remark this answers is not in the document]
chrismaree: fair enough & done.

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,82 @@
...
+    function transformPrice(FixedPoint.Unsigned memory oraclePrice)
+        public
+        override
+        view
+        returns (FixedPoint.Unsigned memory)
+    {
+        FixedPoint.Unsigned memory strike = financialProductStrikes[msg.sender];
+        require(strike.isGreaterThan(0), "Caller has no strike");
+        // If price request is made before expiry, return 1. Thus we can keep the contract 100% collateralized with 1
+        // WETH pre-expiry, and that disputes pre-expiry are illogical (token is always backed by 1 WETH pre-expiry).
```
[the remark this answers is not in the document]
chrismaree: can do. The only reason I commented on it this as throughout the comments I referred to a WETH example. will refine the comment.

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,82 @@
...
+    function getStrikeForFinancialProduct(address financialProduct) public view returns (FixedPoint.Unsigned memory) {
+        return financialProductStrikes[financialProduct];
+    }
+
+    /**
+     * @notice Returns a transformed price by applying the structured note payout structure.
+     * @param oraclePrice price from the oracle to be transformed.
+     * @return transformedPrice the input oracle price with the price transformation logic applied to it.
+     */
+    function transformPrice(FixedPoint.Unsigned memory oraclePrice)
```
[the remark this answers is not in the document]
chrismaree: This is a good point as this would apply the settlement strike transformation to their liquidation price, which is not correct. rather, the transformation function should not consider `currentTime` at all and simply look at the `requestTime` of the `oraclePrice`. In fact this makes the library independent of any notion of local time & we can remove `Testable` from it.

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * @title Structured Note Financial Product Library
+ * @notice Adds custom price transformation logic to modify the behavour of the expiring multi party contract.  The
+ * contract holds say 1 WETH in collateral and pays out that 1 WETH if, at expiry, ETHUSD is below a set strike. If
+ * ETHUSD is above that strike, the contract pays out a given dollar amount of ETH.
+ * Example: expiry is DEC 31. Strike is $400. Each token is backed by 1 WETH
+ * If ETHUSD < $400 at expiry, token is redeemed for 1 ETH.
+ * If ETHUSD >= $400 at expiry, token is redeemed for $400 worth of ETH, as determine by the DVM.
+ */
+contract StructuredNoteFinancialProductLibrary is FinancialProductLibrary, Testable, Ownable {
```
mrice32, line 24: nit: can we remove testable?

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
nicholaspai: Leave it be for now! You're right its a bit opinionated

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
nicholaspai: Gotcha

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
nicholaspai: Ah got it. Maybe add a comment that a monitoring party tracking the oracle price would also need access to this `transformPrice` method to derive the "price" that the FP sees.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
nicholaspai: I'd side with Chris' current implementation and just assume zero-address = no transformation precisely because we don't want users to have to know which address the default library is deployed at. 

The only downside of this is the 1 extra line that performs the `if (x == address(0) ...` check, I think.

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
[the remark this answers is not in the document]
nicholaspai: Nice +1

packages/core/buidler.config.js
```diff
@@ -17,7 +17,7 @@ const { getLongVersion } = require("@nomiclabs/buidler-etherscan/dist/solc/SolcV
 usePlugin("@nomiclabs/buidler-truffle5");
 usePlugin("solidity-coverage");
 usePlugin("@nomiclabs/buidler-etherscan");
-usePlugin("buidler-gas-reporter");
+// usePlugin("buidler-gas-reporter");
```
nicholaspai, line 20: Did you mean to leave this here?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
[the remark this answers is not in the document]
nicholaspai: I like ExpiringContractInterface because its interface for now is just an expirationTimestamp

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -0,0 +1,83 @@
...
+     * @dev Note: a) Only the owner (deployer) of this library can set new strike prices b) A strike price can not be 0.
+     * b) A strike price can only be set once to prevent the deployer from changing the strike after the fact.
+     * c)  financialProduct must exposes an expirationTimestamp method.
+     */
+    function setFinancialProductStrike(address financialProduct, FixedPoint.Unsigned memory strikePrice)
+        public
+        onlyOwner
+    {
+        require(strikePrice.isGreaterThan(0), "Cant set 0 strike");
+        require(financialProductStrikes[financialProduct].isEqual(0), "Strike already set");
```
nicholaspai, line 42: What happens if a strike is actually meant to be 0?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
nicholaspai: I believe we can because it does not use currentTime at all.

packages/core/contracts/financial-templates/test/ExpiringMultiPartyMock.sol
```diff
@@ -0,0 +1,38 @@
...
+        financialProductLibrary = FinancialProductLibrary(_financialProductLibraryAddress);
+    }
+
+    function transformPrice(FixedPoint.Unsigned memory price, uint256 requestTime)
+        public
+        view
+        returns (FixedPoint.Unsigned memory)
+    {
+        if (address(financialProductLibrary) == address(0)) return price;
+        try financialProductLibrary.transformPrice(price, getCurrentTime()) returns (
```
nicholaspai, line 30: Should this be `requestTime` instead of `getCurrentTime()`?

packages/core/test/financial-templates/common/financial-product-libraries/StructuredNoteFinancialProductLibrary.js
```diff
@@ -0,0 +1,106 @@
...
+  it("Can not re-set the strike for a given financial product", async () => {
+    assert(
+      await didContractThrow(
+        structuredNoteFPL.setFinancialProductStrike(expiringMultiParty.address, {
+          rawValue: strikePrice.toString()
+        })
+      )
+    );
+  });
+  it("Can not set strike price for invalid financial product", async () => {
```
nicholaspai, line 46: This only fails for the ZERO-ADDRESS right? Or does it fail for any contract that doesn't implement `expirationTimestamp`? If the latter, then can you add a test case for a non-zero-address, non-EMP address?

nicholaspai (approved):
Looks great, LGTM! Can you run a quick test on how much bytecode this adds to the EMP and Perp and how much bytecode room remains?

packages/core/buidler.config.js
chrismaree: nope. thanks for catching this

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
chrismaree: ye, that's a good point. I think for now it's fine as this library is meant to fill the specific structured note use case for eth-related products & right now the emp does not support any negative prices so it would be illogical to have a 0 strike. If a product needs 0 strikes then the whole design needs to be reconsidered.

packages/core/contracts/financial-templates/test/ExpiringMultiPartyMock.sol
chrismaree: ye, it should be. have updated.

packages/core/test/financial-templates/common/financial-product-libraries/StructuredNoteFinancialProductLibrary.js
chrismaree: it's the latter and I've added a test

chrismaree:
> Looks great, LGTM! Can you run a quick test on how much bytecode this adds to the EMP and Perp and how much bytecode room remains?

Before this PR we had 977 bytes left in the `ExpiringMultiPartyLib`. After this PR we have 585. Overall this logic added ~ 392 bytes. There is another 2 PRs still to be added to this library; one to transform collateralization ratio and one to transform price identifiers. It's going to be close to get it in under the limit!

package.json
```diff
@@ -45,7 +45,7 @@
     "lint-staged": "^10.1.3",
     "lodash.startcase": "^4.4.0",
     "prettier": "1.19.1",
-    "prettier-plugin-solidity": "^1.0.0-alpha.59",
+    "prettier-plugin-solidity": "1.0.0-alpha.59",
```
chrismaree, line 48: I added this to lock the prettier version as the latest version introduced a lot of changes. this should come in as a separate PR, as done here https://github.com/UMAprotocol/protocol/pull/2209

Issue #2164, opened 2020-11-09
Refactor serverless-orchestration into a seperate package and add tests to CI 


chrismaree:
**Why**

Right now the `serverless-orchestration` directory within `packages/core/scripts/serverless-orchestration/` is nestled away within the core's scripts folder. This should be pulled out into a separate package a the `packages/` level. Also, The unit tests within this directory are currently not running in CI. This should be changed.

**How**

Move the scrips & associated directory & update tests to ensure they pass in CI.

**Resolution**

These scripts are within their own package & are tested in CI.

**Difficulty Score [1-10]**
- 5: ~4 hours

PR #2186, opened 2020-11-16
improve(serverless): refactor scripts into their own package & run in CI

chrismaree:
**Motivation**

Previously the serverless scripts were stored in the core scripts directory. This PR refractors this to a separate package.

issue: #2164

**Summary**

All server less scripts and tests for the hub and spoke are moved to their own package.

**Details**

Note: this PR will break production as configs in gcp will need to be updated in accordance with this release.

**Issue(s)**

close: #2164


packages/core/package.json
```diff
@@ -31,7 +31,6 @@
     "moment": "^2.24.0",
     "node-fetch": "^2.3.0",
     "solidity-coverage": "^0.7.2",
-    "supertest": "^4.0.2",
```
nicholaspai, line 34: What was this?

nicholaspai (approved):

drumenbassik (approved):

packages/core/package.json
chrismaree: This was the package I was using to test serverless. as this serverless is now in it's own package this is no longer needed in core :)

mrice32 (approved):
Looks great! Do you need to add a truffle config to this directory?

chrismaree:
> Looks great! Do you need to add a truffle config to this directory?

I didn't think so but looking at how we've configured other packages that re-use the common truffle config we should

packages/serverless-orchestration/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@uma/serverless-orchestration",
+  "version": "1.1.0",
+  "description": "Serverless orchestration scripts to run a parallel instances of scripts or bots",
+  "dependencies": {
+    "@truffle/contract": "^4.2.20",
+    "@uma/common": "^1.0.1",
+    "@uma/financial-templates-lib": "^1.1.0"
+  },
+  "devDependencies": {
```
mrice32, line 10: nit: add `@uma/core` as a dev dependency for the tests?

mrice32 (approved):
LGTM!

PR #2191, opened 2020-11-17
feat(emp): financial product library to apply collateralization ratio transformation

### LINKED ISSUE #2160: Refactor `_getOraclePrice` and `_getCollateralRequirement` into external libraries, allow contract factories to have customizable constructor params

chrismaree:
**Motivation**

Issue #2160

Add the ability to transform contract collateral requirement based on an external library.

Follows on from PR https://github.com/UMAprotocol/protocol/pull/2185 which applied similar logic to the settlement price.

**Summary**

This PR introduces a function `transformCollateralRequirement` to the `FinancialProductLibrary` which enables the collateral requirement of a contract to be transformed by an external library. This is useful for contracts that might want to shift their collateral requirement as a function of price. An example of this is a structured note product with a defined strike price. If the price is above the strike then and additional collateral is redundant (and inefficient). 

**Details**

`transformCollateralRequirement` is applied to the `Liquidatable` contract by adding a `try`/`catch` wrapped method to transform the collateral requirement of the contract, if the library has the provided method. `Liquidatable` uses the `collateralRequirement` in exactly one place: computation of a position's correct collateralization at dispute settlement this can be seen here:
https://github.com/UMAprotocol/protocol/blob/bccaefe1ee4e187af03a91c3e84512d1d5353ba6/packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol#L530

This PR modifies this logic at dispute settlement time by changing the `_settle` function to use the transformed collateral requirement. 

**Issue(s)**

close #2160

**Bytecode impact**
Before this PR `ExpiringMultiPartyLib` had `595` remaining byte within the size limit. After this PR it has `350` bytes. Cutting it fine but still within the limits :)


packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
```diff
@@ -504,6 +504,28 @@ contract Liquidatable is PricelessPositionManager {
...
     }
 
+    /**
+     * @notice Accessor method to calculate a transformed collateral requirement using the finanical product library
+      specified during contract deployment. If no library was provided then no modification to the collateral requirement is done.
+     * @param price input price used as an input to transform the collateral requirement.
+     * @return transformedCollateralRequirement collateral requirement with transformation applied to it.
+     * @dev This method should never revert.
+     */
+    function transformCollateralRequirement(FixedPoint.Unsigned memory price)
```
nicholaspai, line 514: Are you purposefully making this public?

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
```diff
@@ -526,8 +548,11 @@ contract Liquidatable is PricelessPositionManager {
         FixedPoint.Unsigned memory tokenRedemptionValue =
             liquidation.tokensOutstanding.mul(liquidation.settlementPrice);
 
-        // The required collateral is the value of the tokens in underlying * required collateral ratio.
-        FixedPoint.Unsigned memory requiredCollateral = tokenRedemptionValue.mul(collateralRequirement);
+        // The required collateral is the value of the tokens in underlying * required collateral ratio. The Transform
+        // Collateral requirement method applies a from the financial Product library to change the scaled the collateral
+        // requirement based on the settlement price. If no library was specified when deploying the emp then this makes no change.
+        FixedPoint.Unsigned memory requiredCollateral =
```
nicholaspai, line 554: What if the `sponsorCollateral < requiredCollateral` at liquidation time, but then at dispute time, the oracle price is high enough such that `requiredCollateral` decreases and `sponsorCollateral > requiredCollateral`, so the liquidator gets disputed successfully to their surprise.

This doesn't seem like a bug from the contract's point of view but is bad liquidator/disputer ux.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
```diff
@@ -504,6 +504,28 @@ contract Liquidatable is PricelessPositionManager {
...
+     * @return transformedCollateralRequirement collateral requirement with transformation applied to it.
+     * @dev This method should never revert.
+     */
+    function transformCollateralRequirement(FixedPoint.Unsigned memory price)
+        public
+        view
+        returns (FixedPoint.Unsigned memory)
+    {
+        if (address(financialProductLibrary) == address(0)) return collateralRequirement;
+        try financialProductLibrary.transformCollateralRequirement(price, collateralRequirement) returns (
```
nicholaspai, line 520: Have you considered adding a test that tests what happens when `financialProductLibrary. transformCollateralRequirement` reverts?  I did something similar when testing what would happen when the perpetual's `withdrawFundingRateFees()` method reverts [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/test/MockPerpetual.sol#L48)

nicholaspai (commented):
This PR + the transformPrice PR are going to make the EMP very powerful, had some comments about some edge cases

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
chrismaree: yes. it's a view function so no risk and it is needed for frontends to compute the current collateral requirement based on a given input price.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
chrismaree: I have not tested such but can add a mock and a test to do it. will report back once it's added.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
chrismaree: Hhhmm. I'm not sure I agree. This is different from the case in the previous library with the comment that @mrice32 brought up https://github.com/UMAprotocol/protocol/pull/2185#discussion_r525366736 that identified that for the transformation of price, you need to know the timestamp of the price request. This differs because we already know the `liquidation.settlementPrice` at the liquidation timestamp, as resolved by the DVM. The price requested to the DVM and the resolution there of are all around the liquidation timestamp. 

Therefore, we are always performing the transformation considering the timestamp _of the liquidation_. What this means is that the bot that ran the liquidation/dispute (or if this was liquidation/disputed from the front end) would first need to call `transformCollateralRequirement` to get their current value of the collateral requirement when computing if a position is correctly collateralized.

Does this make sense?

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
chrismaree: I've added tests for reverting library to liquidatable & retroactively to the priceless position manager to verify this try/catch logic works as expected.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
mrice32: Yeah, @chrismaree, that makes sense to me.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
```diff
@@ -504,6 +504,28 @@ contract Liquidatable is PricelessPositionManager {
...
+     * @param price input price used as an input to transform the collateral requirement.
+     * @return transformedCollateralRequirement collateral requirement with transformation applied to it.
+     * @dev This method should never revert.
+     */
+    function transformCollateralRequirement(FixedPoint.Unsigned memory price)
+        public
+        view
+        returns (FixedPoint.Unsigned memory)
+    {
+        if (address(financialProductLibrary) == address(0)) return collateralRequirement;
```
mrice32, line 519: nit: if we're already using a try-catch here, would it make sense to _not_ do this explicit check and just let the try-catch handle the fallback?

packages/core/contracts/financial-templates/common/financial-product-libraries/StructureNoteFinancialProductLibrary.sol
```diff
@@ -75,4 +75,31 @@ contract StructuredNoteFinancialProductLibrary is FinancialProductLibrary, Ownab
...
+        // If the price is less than the strike than the original collateral requirement is used.
+        if (oraclePrice.isLessThan(strike)) {
+            return collateralRequirement;
+        } else {
+            // If the price is more than the strike then the collateral requirement is scaled by the strike. For example
+            // a strike of $400 and a CR of 1.2 would yield:
+            // ETHUSD = $350, payout is 1 WETH. CR is multiplied by 1. resulting CR = 1.2
+            // ETHUSD = $400, payout is 1 WETH. CR is multiplied by 1. resulting CR = 1.2
+            // ETHUSD = $425, payout is 0.941 WETH (worth $400). CR is multiplied by 0.941. resulting CR = 1.1292
+            // ETHUSD = $500, payout is 0.8 WETH (worth $400). CR multiplied by 0.8. resulting CR = 0.96
```
mrice32, line 101: very cool

mrice32 (approved):
LGTM! minor comments

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
mrice32: I also think the try-catch is subtly really cool here because it allows people to pass in financial product libraries that only implement the methods that they actually want to override.

chrismaree: this was intentional :) I initially did it the way you propose, (and the same is true with the price transformation) but I get an error as follows:
```
     Error: Transaction reverted: function call to a non-contract account
```

This occurs if the `financialProductLibrary` is set to an EOA, such as the address(0). This check would ensure that it would not revert at this point.

However, you did make me realize that I did not have a test for setting the `financialProductLibrary` to another EOA address! I tested this and indeed the same revert happens. 

I've updated the function to first check if the library is set to a contract. If it is not, then it does nothing. If it is, then it enters into the try-catch logic block.

I've also updated the previous library for price transformations to use the same logic.

PTAL.

packages/core/contracts/financial-templates/expiring-multiparty/Liquidatable.sol
nicholaspai: yeah makes sense

nicholaspai (approved):
This was pretty easy to review because of the comments and thought out tests! Left some comments but LGTM after

PR #2192, opened 2020-11-17
feat(affiliates): add 2020-11-09 emp deployer payouts calculation

daywiss:
**Motivation**

We need to calculate deployer rewards for 2020-11-09 23:00:00 through 2020-11-16 23:00:00

**Summary**
This fixes some small issues and creates an output file for emp deployer rewards calculation.

**Details**

- creates a file folder called payouts/deployers/2020-11-09_2020-11-16.json which has script results
- allows balances to go negative to allow correct valuation of total locked in EMP. Before fix an expired contract was receiving significantly more rewards.
- adds new tests for historical balances, and found issue where snapshots needed to be "finalized" before processing. This just adds in 1 more snapshot for the last block processed.
- adds additional metadata to reward output to map empaddress -> deployer address for informational purposes
- update: major bug found in synthetic value calculation, that is now fixed and rewards regenerated from initial PR

image

**Reproducing Output** 
**Prerequisites**
- You will need GCP setup and Bigquery available to your user. 

To run this script you will need to create a js or json file with the config parameters. These are saved in the output, which you can copy into a file:
```js
{
    "empWhitelist": [
      "0xaBBee9fC7a882499162323EEB7BF6614193312e3",
      "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56",
      "0x306B19502c833C1522Fbc36C9dd7531Eda35862B",
      "0x3a93E863cb3adc5910E6cea4d51f132E8666654F"
    ],
    "startTime": 1604962800000,
    "endTime": 1605567600000,
    "totalRewards": "50000",
    "network": 1
}
```
once you have this file, in the affiliate root directory you can run
`node app/DeployerRewards ./your-config.json --network=mainnet_mnemonic`
This may take several minutes to run, but it will output to console once complete. To save to file append
` >> output.json` to the previous call.


packages/affiliates/libs/models.js
```diff
@@ -186,10 +186,21 @@ function History() {
...
     return result;
   }
+  // get exact block number
+  function get(blockNumber) {
+    const result = history.find(x => x.blockNumber == blockNumber);
+    assert(result, "Not found in history: " + blockNumber);
+    return result;
+  }
+  function has(blockNumber) {
+    !!history.find(x => x.blockNumber == blockNumber);
```
nicholaspai, line 196: Why the double !?

packages/affiliates/libs/models.js
```diff
@@ -186,10 +186,21 @@ function History() {
...
+  }
   function length() {
     return history.length;
   }
   return {
+    get,
+    has,
```
nicholaspai, line 203: I like these simple method names

packages/affiliates/libs/processors.js
```diff
@@ -150,19 +168,15 @@ function EmpBalances(handlers = {}, { collateral, tokens } = {}) {
...
-      if (collateral.has(caller)) {
-        try {
-          collateral.sub(caller, collateralReturned.toString());
-        } catch (err) {
-          // ignore balance below 0 error
-        }
+      try {
+        collateral.sub(caller, collateralReturned.toString());
+      } catch (err) {
+        // ignore balance below 0 error
```
nicholaspai, line 174: Is this error thrown by the `collateral.sub()` method? Isn't this a BN class? I'm wondering if it'd be better to change that method to just not throw on negative numbers

packages/affiliates/payouts/deployers/2020-11-09_2020-11-16.json
```diff
@@ -0,0 +1,32 @@
...
+      "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56",
+      "0x306B19502c833C1522Fbc36C9dd7531Eda35862B",
+      "0x3a93E863cb3adc5910E6cea4d51f132E8666654F"
+    ],
+    "startTime": 1604962800000,
+    "endTime": 1605567600000,
+    "totalRewards": "50000",
+    "network": 1
+  },
+  "startBlock": { "timestamp": 1604962804000, "number": 11226178 },
```
chrismaree, line 14: do we want to convert these from js timestamps to unix timestamps? (1000x smaller).

packages/affiliates/payouts/deployers/2020-11-09_2020-11-16.json
```diff
@@ -0,0 +1,32 @@
...
+  "endBlock": { "timestamp": 1605567593000, "number": 11271760 },
```
chrismaree, line 15: is this correct? should the timestamp not be `1605564000` for `11/16/2020 @ 10:00pm (UTC)`

chrismaree (approved):
Looks awesome to me! numbers seem right. your payouts file is super clean, I like it.

packages/affiliates/libs/models.js
daywiss: "has" technically should return a boolean, so that just converts truthy or falsey into a true or false

daywiss: there should be a return there though actually

packages/affiliates/libs/processors.js
daywiss: when instantiating the balances class, i have an option to not throw on negative balances, which is set to true. I can probably get rid of this try catch completely now.

packages/affiliates/payouts/deployers/2020-11-09_2020-11-16.json
daywiss: i think if we made the change for this part of the code it would be easy, but it would be inconsistent with the config timestamp format above it.

if we are to do this, id like to keep all timestamps consistent S vs MS. would not be a hard change necessarily, but tedious. would require updating big query queries and any timestamp conversions with moment as well as potentially example scripts and config files to reflect seconds instead of ms. 

if you are OK with the inconsistent output then i can make the change to just that. WDYT?

packages/affiliates/libs/processors.js
daywiss: oh btw its not a BN class, its a custom model "Balances" defined in models.js.

packages/affiliates/payouts/deployers/2020-11-09_2020-11-16.json
daywiss: i thought it went to 11pm - 11pm utc. can rerun, but please confirm that. if it goes from 11pm - 10pm we will lose an hour between payouts.

mrice32 (approved):
Looks great conditional on resolving @chrismaree's and @nicholaspai's comments above! Minor note: in the future, I would suggest splitting up any code changes from committing payouts just to keep the purpose of the PR succinct.

packages/affiliates/libs/models.js
mrice32: Side note: `!!(expression)` does the same thing as `Boolean(expression)`.

packages/affiliates/payouts/deployers/2020-11-09_2020-11-16.json
chrismaree: Sorry, you are correct on the time range here! It should run from 1ppm to 11pm UTC over the week.

The timestamp you've got runs until 11/16/2020 @ 10:59pm (UTC). So we effectively 11 to 11 returns 11 to 10.99999. This is slightly different to the LM which uses the `block+1` for each subsequent week. In either case, this is interchangeable by 15 seconds on either side so no need to change anything.

packages/affiliates/libs/models.js
daywiss: its inspired by JS Map syntax, get, set, has, delete

PR #2196, opened 2020-11-18
feat(funding-rate-store) Permission `getFundingRate` method, make funding rate structs private

### LINKED ISSUE #2040: Funding Rate Store: Use Registry to permission funding rate request/access
**Why**

Project Evergreen.

**Difficulty Score [1-10]**

- 4: ~2 hours

nicholaspai:

**Motivation**

#2040

**Summary**

`fundingRateRecords` and `fundingRateDisputes` mappings are now private so that funding rate data is only available via the `getFundingRateForContract()` method permissioned call.

Proposers still are expected to be interest in their a perpetual contract's base reward rate (per second) and its last update time, which they can use to project their rewards. So 1 view method is added:
- `getRewardRateForContract(perpetualAddress, proposedRate)` --> returns the projected reward rate for `perpetual` using the time elapsed since the last proposal and the magnitude of change between the `proposedRate` and the current rate.

**Issue(s)**

Fixes #2040

### DIFF
--- a/packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
+++ b/packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
@@ -10,6 +10,8 @@ import "../../../oracle/interfaces/StoreInterface.sol";
 import "../../../oracle/interfaces/OracleInterface.sol";
 import "../../../oracle/interfaces/FinderInterface.sol";
 import "../../../oracle/interfaces/AdministrateeInterface.sol";
+import "../../../oracle/interfaces/IdentifierWhitelistInterface.sol";
+import "../../../oracle/interfaces/RegistryInterface.sol";
 import "../../../oracle/implementation/Constants.sol";
 
 import "../../perpetual-multiparty/PerpetualInterface.sol";
@@ -85,8 +87,10 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
 
     enum ProposalState { None, Pending, Expired }
 
-    mapping(address => FundingRateRecord) public fundingRateRecords;
-    mapping(address => mapping(uint256 => FundingRateRecord)) public fundingRateDisputes;
+    mapping(address => FundingRateRecord) private fundingRateRecords;
+    // TODO: Is there any reason to make `fundingRateDisputes` public? Could users use the dispute struct
+    // to get funding rate data "for free"? If not, then I see no harm in making it public.
+    mapping(address => mapping(uint256 => FundingRateRecord)) private fundingRateDisputes;
 
     uint256 public proposalLiveness;
 
@@ -136,9 +140,16 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
      *                MODIFIERS             *
      ****************************************/
 
-    // Pubishes any pending proposals whose liveness has passed, pays out rewards for such proposals
+    // Pubishes any pending proposals whose liveness has passed, pays out rewards for such proposals.
     modifier publishAndWithdrawProposal(address perpetual) {
         _publishRateAndWithdrawRewards(perpetual);
+        _;
+    }
+
+    // Function callable only by contracts registered with the DVM.
+    modifier onlyRegisteredContract() {
+        RegistryInterface registry = RegistryInterface(finder.getImplementationAddress(OracleInterfaces.Registry));
+        require(registry.isContractRegistered(msg.sender), "Caller must be registered");
         _;
     }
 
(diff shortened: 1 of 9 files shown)


packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -179,6 +191,41 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+     * of change between the proposed and current funding rate.
+     * @param perpetual perpetual contract whose base reward rate the caller is querying.
+     * @return FixedPoint.Unsigned representing the base reward rate for the given contract. 0.01 would represent a
+     * reward rate of 1% per second.
+     */
+    function getRewardRateForContract(address perpetual)
+        external
+        view
+        nonReentrantView()
+        returns (FixedPoint.Unsigned memory)
```
mrice32, line 223: Just curious: what do you think the advantages/disadvantages would be for combining these two methods into one that returns the projected reward? (not suggesting that's what we should do -- just want to hear your thoughts)

mrice32 (commented):
Looks awesome! One minor comment.

nicholaspai:
@UMAprotocol/eng Do we also need a similar method for the `cumulativeFeeMultiplier`? Currently this has not been needed with the EMP because regular fees are 0, but wouldn't sponsors need to know their up to date fee-adjusted collateral?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Interesting. That could be a good idea if we anticipate that `getRewardRate` and `getProposalTime` would only be used to calculate potential rewards. Do you think there'd be other reasons to call them?

mrice32: Yeah, not sure. I can't think of any other use case.

mrice32:
> @UMAprotocol/eng Do we also need a similar method for the `cumulativeFeeMultiplier`? Currently this has not been needed with the EMP because regular fees are 0, but wouldn't sponsors need to know their up to date fee-adjusted collateral?

Isn't that what `getCollateral()` does?

nicholaspai:
> > @UMAprotocol/eng Do we also need a similar method for the `cumulativeFeeMultiplier`? Currently this has not been needed with the EMP because regular fees are 0, but wouldn't sponsors need to know their up to date fee-adjusted collateral?
> 
> Isn't that what `getCollateral()` does?

Not quite though `getCollateral` uses the current `cumulativeFeeMultiplier` but it doesn't actually update the `cumulativeFeeMultiplier`. This would return a slightly higher value if the last feeMultiplier was updated >= 1 second ago, and in the mean-time regular fees are still accruing but just haven't been reflected on the feeMultiplier.

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: For the sake of reducing interface size, I'll combine these for now into one method that returns projected rewards.

mrice32 (approved):
LGTM!

mrice32:
> > > @UMAprotocol/eng Do we also need a similar method for the `cumulativeFeeMultiplier`? Currently this has not been needed with the EMP because regular fees are 0, but wouldn't sponsors need to know their up to date fee-adjusted collateral?
> > 
> > 
> > Isn't that what `getCollateral()` does?
> 
> Not quite though `getCollateral` uses the current `cumulativeFeeMultiplier` but it doesn't actually update the `cumulativeFeeMultiplier`. This would return a slightly higher value if the last feeMultiplier was updated >= 1 second ago, and in the mean-time regular fees are still accruing but just haven't been reflected on the feeMultiplier.

I see. Yeah, good point! That could be nice, although, it seems annoying to compute.

nicholaspai:
> > > > @UMAprotocol/eng Do we also need a similar method for the `cumulativeFeeMultiplier`? Currently this has not been needed with the EMP because regular fees are 0, but wouldn't sponsors need to know their up to date fee-adjusted collateral?
> > > 
> > > 
> > > Isn't that what `getCollateral()` does?
> > 
> > 
> > Not quite though `getCollateral` uses the current `cumulativeFeeMultiplier` but it doesn't actually update the `cumulativeFeeMultiplier`. This would return a slightly higher value if the last feeMultiplier was updated >= 1 second ago, and in the mean-time regular fees are still accruing but just haven't been reflected on the feeMultiplier.
> 
> I see. Yeah, good point! That could be nice, although, it seems annoying to compute.

Yeah it's not a clean refactor but without the method, sponsor positions will read collateral incorrectly via `getCollateral()` once regular fees are > 0. For now I'll leave it but add in a comment to check this out later

PR #2199, opened 2020-11-18
feat(funding-rate-applier): Don't update multiplier post-emergency shutdown

### LINKED ISSUE #2198: Add view method to `FundingRateApplier` so that caller can view updated funding rate multiplier
**Why**

Liquidator, disputer, and sponsors will need access to the contract's updated cumulative funding rate multiplier, which would be a function of the current value of `perpetual.cumulativeFundingRateMultiplier()` and the time elapsed since the last multiplier update (emitted on a `NewFundingRate` Perpetual event).

The multiplier is updated on every contract interaction, but these actors will need a more up to date multiplier value. Currently the only way to get a live value is to call `perpetual.applyFundingRate()` and then call `perpetual.cumulativeFundingRateMultiplier()`. This is not ideal because `applyFundingRate()` is a state-changing transaction that costs gas.

I propose adding a `getCumulativeFundingRateMultiplier()` method that returns what the multiplier would be if someone were to successively call `applyFundingRate()` and `cumulativeFundingRateMultiplier()`

**Difficulty Score [1-10]**

- 4: ~2 hours (Potentially longer if this takes the Perpetual above the contract bytecode limit)

nicholaspai:
**Motivation**

#2198

**Summary**

Replaces `cumulativeFundingRateMultiplier` public view method with `getCumulativeFundingRateMultiplier` which is a `view` method that returns what the updated funding rate multiplier would be, even if the contract state has not been updated. Post emergency shutdown, this method should just return `cumulativeFundingRateMultiplier` because the multiplier should no longer be updated post shutdown.

In the course of working on this feature, I stumbled upon the  following bug:
- `cumulativeFundingRateMultiplier` should not update post-emergency shutdown. It updates via the `updateFundingRate` modifier which is applied to two methods that can be called post-shutdown: `dispute()` and `withdrawLiquidationRewards()`. This means that it is possible that the funding rate multiplier can be updated post-shutdown by these methods which would modify the emergency shutdown's token redemption value.

The solution I implemented is to move `emergencyShutdownTimestamp` into the `FundingRateApplier` contract so that `_applyFundingRate()` can exit early if the contract is in a shutdown state. This effectively freezes the `cumulativeFundingRateMultiplier` post-shutdown.

**Issue(s)**

Fixes #2198


packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -1068,8 +1068,7 @@ contract("PricelessPositionManager", function(accounts) {
     assert.equal((await pricelessPositionManager.getCollateral(sponsor)).toString(), toWei("98.99"));
 
     // Test that regular fees accrue after an emergency shutdown is triggered.
-    const shutdown = await financialContractsAdmin.callEmergencyShutdown(pricelessPositionManager.address);
-    truffleAssert.eventNotEmitted(shutdown, "EmergencyShutdown");
```
nicholaspai, line 1072: Side note: Anyone know why this `truffleAssert.eventNotEmitted` was added? Seems kind of weird to check for an event not emitting. I think the intention is to make sure that the emergency shutdown executes but this test should just assume it works.

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -1046,8 +1051,7 @@ contract("PerpetualPositionManager", function(accounts) {
     assert.equal((await positionManager.getCollateral(sponsor)).toString(), toWei("98.99"));
 
     // Test that regular fees accrue after an emergency shutdown is triggered.
-    const shutdown = await financialContractsAdmin.callEmergencyShutdown(positionManager.address);
-    truffleAssert.eventNotEmitted(shutdown, "EmergencyShutdown");
```
nicholaspai, line 1050: Same idea as [above](https://github.com/UMAprotocol/protocol/pull/2199/files#r526434510)

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -95,7 +128,7 @@ abstract contract FundingRateApplier is Testable, Lockable {
         view
         returns (FixedPoint.Unsigned memory tokenDebt)
     {
-        return rawTokenDebt.mul(cumulativeFundingRateMultiplier);
+        return rawTokenDebt.mul(getCumulativeFundingRateMultiplier());
```
mrice32, line 131: Does this mean that this method is now expected to be called without pre-updating the funding rate multiplier? Not necessarily a problem, but IIUC, where we use this, we would basically be running the _calculateEffectiveFundingRate function twice within the same transaction, right? Since we update the funding rate multiplier in a modifier that's called before the main logic begins?

nicholaspai: `getCumulativeFundingRateMultiplier()` is a helper method so that sponsors can more easily track the value of their debt in real time. The alternative is to call `applyFundingRate()` followed by `cumulativeFundingRateMultiplier()`. 

Moreover, yes, where wherever `_getFundingRateAppliedTokenDebt()` is used the computation will be done twice, but the state transition that stores a new value in `cumulativeFundingRateMultiplier()` is only done once. Look how `_applyEffectiveFundingRate()` differs from `getCumulativeFundingRateMultiplier()`.

This is my design at least, let me know if you think the implementation is diff.

nicholaspai: Thinking more about this. The main benefit of this feature is that it enhances funding-multiplier liveness: sponsors can get their up to date adjusted token debt value without having to send a transaction. However this would be a bit inconsistent with the fee-multiplier liveness, which we have no plans to offer the same view-only method that returns live data. I'm curious what the rest of the team thinks. 

@daywiss @chrismaree

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -87,6 +90,36 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+     * @return newCumulativeFundingRateMultiplier The current funding rate multiplier.
+     */
+    function getCumulativeFundingRateMultiplier()
+        public
+        view
+        returns (FixedPoint.Unsigned memory newCumulativeFundingRateMultiplier)
+    {
+        // If contract is emergency shutdown, return latest multiplier from shutdown time.
+        if (emergencyShutdownTimestamp != 0) {
+            newCumulativeFundingRateMultiplier = cumulativeFundingRateMultiplier;
```
daywiss, line 110: this must be a solidity thing but nothing is being returned. how does this syntax work?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
daywiss: I dont think i really understand!

nicholaspai: Multiplier is supposed to reflect latest funding rate (% per second) which affects the value of sponsor debt. 

Multiplier is updated on every contract call. However, in between calls the multiplier itself hasn't changed but it's "intrinsic value" has changed because seconds continue to tick. 

The question is, is this "live" multiplier useful for sponsors, proposers, etc?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: `newCumulativeFundingRateMultiplier` is declared in the function name declaration in the `returns()` block

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -112,6 +145,11 @@ abstract contract FundingRateApplier is Testable, Lockable {
     // Note: 1 is set as the neutral rate because there are no negative numbers in FixedPoint, so we decide to treat
     // values < 1 as "negative".
     function _applyEffectiveFundingRate() internal {
+        // If contract is emergency shutdown, then the funding rate multiplier should no longer change.
```
nicholaspai, line 148: Reference for comment

nicholaspai:
@mrice32 I think I'm gonna remove the method that provides an up-to-date funding rate multiplier (including any pending funding rates that haven't been applied via `_applyFundingRate()) because the `cumulativeFeeMultiplier` and the `recordParams` introduced in #2208 don't have the same liveness properties. 

But I do think we need to address the bug [here](https://github.com/UMAprotocol/protocol/pull/2199/files#r528221245) which freezes the fundingRateMultiplier post-emergency shutdown, which applies specifically to the `dispute` and `withdrawLiquidationRewards` methods (which can be called post-shutdown)

mrice32 (approved):
LGTM!

mrice32:
> @mrice32 I think I'm gonna remove the method that provides an up-to-date funding rate multiplier (including any pending funding rates that haven't been applied via `_applyFundingRate()) because the `cumulativeFeeMultiplier`and the`recordParams` introduced in #2208 don't have the same liveness properties.
> 
> But I do think we need to address the bug [here](https://github.com/UMAprotocol/protocol/pull/2199/files#r528221245) which freezes the fundingRateMultiplier post-emergency shutdown, which applies specifically to the `dispute` and `withdrawLiquidationRewards` methods (which can be called post-shutdown)

I think this totally makes sense. Agreed 100%

Issue #2202, opened 2020-11-19
WDF Liquidation Strategy and bot fix


nicholaspai:
**Description**

False positive tests on Liquidatable and liquidator-bot made this bug hard to find:
- Liquidatable should only allow withdrawal livenesses to be reset if the withdrawal has NOT expired yet.

PR #2203, opened 2020-11-19
bug(liquidatable) Fix liveness-reset feature and replace false-positive tests

### LINKED ISSUE #2022: Improve serverless logging flakeyness by waiting for slow transports
**Why**

Some transports can be slow, such as the slack transport. in the case where a process is yielding (due to an error or end of loop) sometimes these slow transports are not yet finished reporting their messages and the bot can exit before logging. this can result in lots not propagating to the respective transport.

**How**

Add the `waitForLogger()` method to the hub and spoke.

**Resolution**

Logs are consistently delivered.

**Difficulty Score [1-10]**

- 3: ~1 hour

nicholaspai:
**Motivation**

#2202 

**Issue(s)**

Fixes #2022


packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -1077,7 +1077,7 @@ contract("PricelessPositionManager", function(accounts) {
     const feesOwed = (
       await store.computeRegularFee(startTime.addn(1), startTime.addn(102), { rawValue: pfc.toString() })
     ).regularFee;
-    assert(Number(pfc.toString()) < Number(feesOwed.toString()));
+    assert.isTrue(Number(pfc.toString()) < Number(feesOwed.toString()));
```
nicholaspai, line 1080: This came from looking through all financial-templates unit tests and removing all `asserts()` that are not asserting `didContractThrow`

mrice32 (approved):
Looks great! Thanks for finding and fixing this everywhere!

packages/liquidator/test/Liquidator.js
```diff
@@ -1206,12 +1204,12 @@ contract("Liquidator.js", function(accounts) {
           // show position has been extended
           assert.equal(
             sponsor2Positions.withdrawalRequestPassTimestamp.toNumber(),
-            Number(sponsor2Liquidations[0].liquidationTime) + Number(liquidationLiveness)
+            Number(sponsor2Liquidations[1].liquidationTime) + Number(withdrawLiveness)
```
nicholaspai, line 1207: This was the false positive test: we were checking the withdrawalRequestPassTimestamp of the wrong liquidation struct.

At this point in the test, the sponsor's `withdrawalRequestPassTimestamp`  has been reset to the latest liquidation time (`sponsor2Liquidations[1].liquidationTime`) plus the `withdrawalLiveness`.

Previously, this test was "correct" because the liquidation was done pre-withdrawal-expiry, meaning that the `withdrawalRequestPassTimestamp` was not actually extended. Therefore, the sponsor's `withdrawalRequestPassTimestamp` should remain equal to the original liquidation time (`sponsor2Liquidations[0].liquidationTime`) + the original liveness.*

*Original liquidation done on line 1167

packages/core/test/financial-templates/expiring-multiparty/Liquidatable.js
```diff
@@ -581,16 +581,16 @@ contract("Liquidatable", function(accounts) {
...
 
-      assert(
+      assert.equal(
         expectedTimestamp,
         (await liquidationContract.positions(sponsor)).withdrawalRequestPassTimestamp.toString()
       );
 
       // Advance time by half of the liveness duration.
-      await liquidationContract.setCurrentTime(startingTime.add(liquidationLiveness.divn(2)).toString());
+      await liquidationContract.setCurrentTime(startingTime.add(withdrawalLiveness.divn(2)).toString());
```
nicholaspai, line 593: This is one part of the false positive test: in this file, `withdrawalLiveness = 3600` but non-intuitively `liquidationLiveness = 3600 * 3`. So, advancing time forward by half of the liquidation liveness actually **expires the withdrawal** which due to the contract bug allows the `withdrawalRequestPassTimestamp` to reset!

packages/core/test/financial-templates/expiring-multiparty/Liquidatable.js
```diff
@@ -604,13 +604,13 @@ contract("Liquidatable", function(accounts) {
       // After the liquidation the liveness timer on the withdrawl request should be re-set to the current time +
       // the liquidation liveness. This opens the position up to having a subsequent liquidation, if need be.
       const liquidation1Time = await liquidationContract.getCurrentTime();
-      assert(
```
nicholaspai, line 607: This is the second part of the false positive. No matter what you throw into this assert, it returns true.

packages/core/test/financial-templates/expiring-multiparty/Liquidatable.js
daywiss: unbelievable

daywiss (approved):

PR #2204, opened 2020-11-19
ci(slither): fix slither ci job

mrice32:
**Motivation**

Fix the slither ci job, now that the executable is updated in the docker image.

**Summary**

try-catches are not handled correctly by slither -- see https://github.com/crytic/slither/issues/511, so that check needed to be disabled. This also re-enables the locked ether check and set slither to not analyze test files.

**Issue(s)**

N/A

### DIFF
--- a/ci/run_slither.sh
+++ b/ci/run_slither.sh
@@ -14,7 +14,7 @@ run_slither() {
 
     # print out slither version for debugging
     slither --version
-    slither --exclude=divide-before-multiply,unused-return,timestamp,naming-convention,pragma,solc-version,external-function,reentrancy-benign,reentrancy-no-eth,arbitrary-send,incorrect-equality,reentrancy-events,assembly,uninitialized-local --filter-paths="@openzeppelin|WETH9.sol|ReentrancyAttack.sol|ReentrancyMock.sol|test" $1
+    slither --exclude=divide-before-multiply,unused-return,timestamp,naming-convention,pragma,solc-version,external-function,reentrancy-benign,reentrancy-no-eth,arbitrary-send,incorrect-equality,reentrancy-events,assembly,uninitialized-local --filter-paths="@openzeppelin|WETH9.sol|test" $1
 }
 
 run_slither $PROTOCOL_DIR/packages/core
--- a/ci/run_slither.sh
+++ b/ci/run_slither.sh
@@ -14,7 +14,7 @@ run_slither() {
 
     # print out slither version for debugging
     slither --version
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
LGTM! Excited to re-add this. Questions:
-Why re-add `locked-ether` test now?
-Why weren't we filtering out `test` files before?

mrice32:
> -Why re-add `locked-ether` test now?

Because it only broke due to one of the test contracts, so filtering out tests allowed us to re-add it.

> -Why weren't we filtering out `test` files before?

I think it was just an oversight!

mrice32:
> > -Why re-add `locked-ether` test now?
> 
> Because it only broke due to one of the test contracts, so filtering out tests allowed us to re-add it.
> 
> > -Why weren't we filtering out `test` files before?
> 
> I think it was just an oversight!

Actually @nicholaspai let me refine that. I think we thought 🤷 what's the harm in running static analysis over the test contracts. However, I think removing them will decrease our false-positive rate (which is still pretty high).

nicholaspai:
> > > -Why re-add `locked-ether` test now?
> > 
> > 
> > Because it only broke due to one of the test contracts, so filtering out tests allowed us to re-add it.
> > > -Why weren't we filtering out `test` files before?
> > 
> > 
> > I think it was just an oversight!
> 
> Actually @nicholaspai let me refine that. I think we thought 🤷 what's the harm in running static analysis over the test contracts. However, I think removing them will decrease our false-positive rate (which is still pretty high).

Do you mean false-positive rate on the prod contracts? Do you have examples? I agree though i'd rather enable more tests even if it means we need to ignore the test contracts

mrice32:
> > > > -Why re-add `locked-ether` test now?
> > > 
> > > 
> > > Because it only broke due to one of the test contracts, so filtering out tests allowed us to re-add it.
> > > > -Why weren't we filtering out `test` files before?
> > > 
> > > 
> > > I think it was just an oversight!
> > 
> > 
> > Actually @nicholaspai let me refine that. I think we thought 🤷 what's the harm in running static analysis over the test contracts. However, I think removing them will decrease our false-positive rate (which is still pretty high).
> 
> Do you mean false-positive rate on the prod contracts? Do you have examples? I agree though i'd rather enable more tests even if it means we need to ignore the test contracts

Yeah, basically, the less code we test, the less likely it is that we'll hit an edge case where _we_ are happy with the code, but we have to disable a slither detector because it thinks something's wrong. And fewer of those edge cases means we get to enable more detectors. If you enable some of the disabled detectors, you'll see some existing false-positives. The locked ether was one that we were able to re-enable because it only happened in a test contract.

PR #2208, opened 2020-11-20
improve(funding-rate-store) allow reward rate, proposal bond, and update timelock to be modified by Owner

### LINKED ISSUE #2207: Enable updates to funding rate reward rates during liveness of Perpetual
**Why**

Currently the reward rate is set once on Perpetual deployment and not touched afterwards. This is a key parameter that tunes how incentivized proposers will be to propose accurate funding rates. We should not expect the Perpetual deployed to set this correctly upon deployment and have it be an effective value for the lifespan of the contract. 

**How**

Solution 1: Permissioned account can change the reward rate. Downside of this approach is centralization, which can be mitigated with a time lock that delays how quickly the param can be changed. Upside is that this is easy to implement. 

Solution 2: Allow DVM to set the rate. Downside of this is coupling DVM administration with financial contract administration. Setting rates would create additional and possibly frequent price requests, and UMA holders shouldn't necessarily have to worry about that. Upside is that this decentralizes who can set the reward rate and is also easy to implement. 

Solution 3: Re-use propose-dispute logic to allow anyone to propose a new reward rate. A value can be proposed, and is set after a liveness period passes. Anyone else can dispute the param which would likely create a price request. Downside is that this also couples DVM administration with financial contract administration. This would also be more complex to implement and require a way to create new price identifiers for reward rate params.* Upside is that this decentralizes who can set the reward rate and also would create fewer price requests than solution 2. 

*This would actually be an interesting application for "variadic DVM identifiers" where we could re-use the perpetual's fundingRateIdentifier but somehow specify to the DVM that we are asking for a reward rate and not a funding rate.

**Resolution**

Decide on a design and then implement with tests. 

**Difficulty Score [1-10]**

- 6: ~1 day

nicholaspai:

**Motivation**

#2207 

**Summary**

- `proposalBondPct` is now unique to each `perpetual`.
- New struct: `recordParams` which maps a `proposalBondPct`, `rewardRatePerSecond` and `updateLiveness` to each `perpetual`. These can be upgraded via `setRecordParamsForContract`.
- Some permissioned account is now able to call `setRecordParams` post-deployment to set `recordParams` for a specific perpetual contract

**Details**

**Issue(s)**

Fixes #2207


packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -422,23 +446,55 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
-     * @dev Callable only by the Perpetual contract.
+     * @notice Set new parameters for a specific `perpetual` contract. New params go into effect
+     * after a liveness period passes.
+     * @dev Callable only by owner. Calling this while there is already a pending RecordParams struct
+     * will overwrite it.
      */
-    function setRewardRate(address perpetual, FixedPoint.Unsigned memory rewardRate)
+    function setRecordParams(address perpetual, RecordParams memory newRecordParams)
+        external
+        onlyOwner()
```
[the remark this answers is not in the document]
nicholaspai: yep. the deployer of the perp would be the factory, what we could do is use MultiRole to be more flexible. Right now the deployer of the perp doesn't have any permissions anywhere else in the contract so I didn't think that was the natural owner

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -33,6 +33,24 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
      *  PRICELESS POSITION DATA STRUCTURES  *
      ****************************************/
 
+    // Define the contract's constructor parameters as a struct to enable more variables to be specified.
+    // This is required to enable more params, over and above Solidity's limits.
+    struct PMConstructorParams {
```
nicholaspai, line 38: I've had to do this because `coverage` adds constructor params (I think) and throws a Stack too deep error on compilation. But `test` works fine

nicholaspai: I ended up removing the requirement to add funding-rate-record-params to the Perp constructor params, but left this change here because we might want it in future if we add more constructor params

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -134,40 +149,28 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
...
-     * @param _finderAddress UMA protocol Finder used to discover other protocol contracts.
-     * @param _priceIdentifier registered in the DVM for the synthetic.
-     * @param _fundingRateIdentifier Unique identifier for DVM price feed ticker for child financial contract.
-     * @param _fundingRateRewardRate Reward rate to pay FundingRateStore to use funding rate data.
-     * @param _minSponsorTokens minimum amount of collateral that must exist at any time in a position.
-     * @param _timerAddress Contract that stores the current time in a testing environment. Set to 0x0 for production.
-     * @param _excessTokenBeneficiary Beneficiary to send all excess token balances that accrue in the contract.
+     * @param params struct to define input parameters for construction of PositionManager.
      */
-    constructor(
```
chrismaree, line 148: this is the first major interface deviation between the emp and the perp as now we feed in a struct. (clearly, the params are different between the contracts but at least the structure is consistent).

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -458,6 +507,26 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+
+        // If liveness has passed, publish new reward rate.
+        if (_pendingRewardProposalPassed(perpetual)) {
+            recordParam.current = recordParam.pending;
+
+            // Reset reward rate proposal state.
+            delete recordParam.pending;
+            recordParam.pendingPassedTimestamp = 0;
+
+            emit ChangedRecordParams(
```
chrismaree, line 521: +1

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -79,14 +78,32 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
+    struct RecordParamsForContract {
+        // Upgradeable params specific to this record.
+        RecordParams current;
+        RecordParams pending;
+        uint256 pendingPassedTimestamp;
     }
 
     enum ProposalState { None, Pending, Expired }
 
+    mapping(address => RecordParamsForContract) public recordParams;
```
chrismaree, line 106: can you add some more comments on these data structures, perhaps above them just to help with explaining exactly what they are used for like we have in other contracts?

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -1899,20 +1878,12 @@ contract("PerpetualPositionManager", function(accounts) {
...
-      finder.address, // _finderAddress
-      priceFeedIdentifier, // _priceFeedIdentifier
-      fundingRateFeedIdentifier, // _fundingRateFeedIdentifier
-      { rawValue: fundingRateRewardRate }, // _fundingRateRewardRate
-      { rawValue: minSponsorTokens }, // _minSponsorTokens
-      timer.address, // _timerAddress
-      beneficiary, // _excessTokenBeneficiary
-      { from: contractDeployer }
-    );
+    constructorParams = {
```
chrismaree, line 1881: so much cleaner like this

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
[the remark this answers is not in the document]
chrismaree: so who is responsible for setting this & how does multi-access work on this? Is this a DVM parameter or some external DVM logic? If it was a MultiRole who would be included?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -270,7 +310,10 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
         );
 
         // Compute proposal bond.
-        FixedPoint.Unsigned memory proposalBond = proposalBondPct.mul(AdministrateeInterface(perpetual).pfc());
+        RecordParamsForContract storage recordParam = recordParams[perpetual];
+        FixedPoint.Unsigned memory proposalBond = recordParam.current.proposerBondPct.mul(
+            AdministrateeInterface(perpetual).pfc()
```
chrismaree, line 315: do you want to add an accessor method to read the `pfc` from a given `perpetual`?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: yeah I can remove this struct for now

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: definitely

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: like an internal `getPfc(perpetual)` method?

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: done

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: done

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
chrismaree: yes

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
```diff
@@ -78,27 +80,65 @@ contract FundingRateStore is FundingRateStoreInterface, Testable, Lockable {
...
-        FixedPoint.Unsigned rewardRatePerSecond; // Percentage of 1 E.g., .1 is 10%.
+    }
+
+    // For readability, wrap the specific configuration settings in RecordParamsForContract in this struct.
+    struct RecordParams {
+        // Liveness period for an update to value in RecordParams to become official.
+        uint256 paramUpdateLiveness;
+        // Reward rate paid to successful proposers. Percentage of 1 E.g., .1 is 10%.
+        FixedPoint.Unsigned rewardRatePerSecond;
+        // Bond % (of given contract's PfC) that must be staked by proposers. Percentage of 1, e.g. 0.0005 is 0.05%
```
chrismaree, line 98:
```suggestion
        // Bond % (of given contract's PfC) that must be staked by proposers. Percentage of 1, e.g. 0.0005 is 0.05%.
```

chrismaree (approved):

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
nicholaspai: Made the change because it improves readability, small hit to bytecode (<100 bytes)

packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol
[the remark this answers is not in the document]
nicholaspai: If we used MultiRole, then there'd still be some owner who had permission to add new roles, this would presumably be the FRS' deployer. If it was multirole, I imagine, each Perp's deployer would be added for that specific perpetual, but it gets tricky. 

It might be easier if this were all configurable on the Perpetual contract's side.

PR #2209, opened 2020-11-21
chore(tools): upgrade prettier-solidity

fvictorio:
**Motivation**

We released a beta of prettier-solidity this week and I'm checking which projects are actively using our plugin.

**Summary**

Upgrade prettier-solidity to its first beta version.

Also: I'm compiling a list of good projects using prettier-solidity, would you be ok if we include UMA on the list?


nicholaspai:
@fvictorio:

> Also: I'm compiling a list of good projects using prettier-solidity, would you be ok if we include UMA on the list?
Yes go for it!

Would you mind sharing or any breaking changes? Assuming our `lint` CI task (which runs prettier on the contracts) doesn't break then this LGTM

nicholaspai:
@fvictorio looks like `lint` CI task broke because the files haven't been linted with `beta.1`, do you mind if I commit to your PR?

fvictorio:
> looks like `lint` CI task broke because the files haven't been linted with `beta.1`

Oh, sorry, I thought that I had done that :sweat_smile: Of course, feel free to commit.

> Would you mind sharing or any breaking changes? Assuming our `lint` CI task (which runs prettier on the contracts) doesn't break then LGTM

It's mostly bug fixes and some changes in formatting. One change that seems to affect several places in the UMA contracts is that long assignement statements now are broken like this if they don't fit in one line:

```solidity
string s =
  "this is a very very very very very long line";
```

Which is similar to what prettier-js does. I think this is fine in general, the problem is that when you have chains of method calls, this is not what you want; you want your chain to be broken. I opened an issue (https://github.com/prettier-solidity/prettier-plugin-solidity/issues/399) to improve that; with any luck, I will have that implemented next week.

chrismaree (approved):
Thanks for doing this! In all the places where there were changes, it looks clean and more consistent. I like how long functions now wrap between lines.

Regarding your question on listing UMA on a list of projects that use prettier-solidity, this sounds good to me but best to get sign-off from @mrice32 on this one.

mrice32 (approved):
LGTM!

mrice32:
> Thanks for doing this! In all the places where there were changes, it looks clean and more consistent. I like how long functions now wrap between lines.
> 
> Regarding your question on listing UMA on a list of projects that use prettier-solidity, this sounds good to me but best to get sign-off from @mrice32 on this one.

@fvictorio Feel free to add us to the list!!

fvictorio:
Awesome, thank you!

PR #2212, opened 2020-11-23
feat(optimistic-oracle): add optimistic oracle

### LINKED ISSUE #2161: Build an optimistic settlement contract
**Why**

Currently the EMP and perpetual request the DVM directly for prices to handle dispute and expiry settlements. This is theoretically not required. You could imagine the perpetual optimistically sending a settlement price to an external contract. This external contract would provide an escalation layer that would allow a disputer to dispute the settlement price.

For example, say I hold 1 USDETH expiring token and I believe that the expiration settlement price is 0.002 on December 31. Instead of forcing the contract to ask the DVM for the USDETH price at Dec 31, and waiting 48 hours for resolution, why can't I just "optimistically" settle my token at 0.002, and if someone disagrees with me then they can dispute me. This dispute would roll up to the DVM, but if I don't get disputed, then anybody else can use that same 0.002 price to settle their Dec 31 expiring tokens.

**How**

From the EMP's point of view, it needs to swap out the Oracle with the Optimistic Settlement contract. The optimistic settlement also needs to correctly pay any regular fees, but more thought needs to be done here.

### LINKED ISSUE #2226: Connect Perpetual to ConfigStore and remove config-setting from Perpetual
**Why**

Currently the Perpetual stores configs such as reward % and bond %, this should be housed (and upgradeable) in an external contract. 

**How**
The ConfigStore address will have to be passed into the Perpetual.

- 4: ~2 hours

mrice32:
**Motivation**

The goal of this PR is to generalize the funding rate store into an optimistic oracle that integrates directly with the perp to provide funding rates as well as serving other contracts.

**Summary**

This adds the optimistic oracle and modifies the perpetual to use it instead of the funding rate store.

**Issue(s)**
Fixes #2161
Fixes #2226


packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,357 @@
...
+        defaultLiveness = _liveness;
+    }
+
+    function requestPrice(
+        bytes32 identifier,
+        uint256 timestamp,
+        IERC20 currency,
+        uint256 reward
+    ) external nonReentrant() returns (uint256 totalBond) {
+        bytes32 id = _getId(msg.sender, identifier, timestamp);
```
chrismaree, line 110: nit: put this below the requires as none of them need it.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,357 @@
...
+            finalFee: finalFee,
+            bond: finalFee,
+            customLiveness: 0
+        });
+
+        if (reward > 0) {
+            currency.safeTransferFrom(msg.sender, address(this), reward);
+        }
+
+        emit RequestPrice(msg.sender, identifier, timestamp);
```
chrismaree, line 134: this event should perhaps contain some more info. useful would include the `currency`, `reward` and `finalFee`.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
chrismaree: in fact, this is more of a stylistic thing than anything else. there are a few cases in this contract where you could put the require's above any local variable decleration.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,357 @@
...
+        request.currency.safeTransferFrom(msg.sender, address(this), request.bond.add(request.finalFee));
+        _getOracle().requestPrice(identifier, timestamp);
+        StoreInterface store = _getStore();
+        request.currency.safeIncreaseAllowance(address(store), request.finalFee);
+        _getStore().payOracleFeesErc20(address(request.currency), FixedPoint.Unsigned(request.finalFee));
+
+        // Compute refund.
+        uint256 refund = 0;
+        if (request.reward != 0 && request.refundOnDispute) {
+            refund = request.reward;
```
chrismaree, line 219: do you not need to transfer this refund in the `_settle` function or send it back here?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -31,31 +33,39 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
-    // address.
-    FinderInterface private finder;
-
-    // Last time the `cumulativeFundingRateMultiplier` was updated.
-    uint256 public lastUpdateTime;
+    struct FundingRate {
+        // Current funding rate value.
+        FixedPoint.Signed rate;
+        // Identifier to retrieve the funding rate.
+        bytes32 identifier;
```
nicholaspai, line 40: Would this `identifier` ever != `fundingRateIdentifier`?

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,357 @@
...
+    function disputePrice(
+        address requester,
+        bytes32 identifier,
+        uint256 timestamp
+    ) external {
+        // Note: re-entrancy guard is done in the inner call.
+        disputePriceFor(msg.sender, requester, identifier, timestamp);
+    }
+
+    function getPrice(bytes32 identifier, uint256 timestamp) external nonReentrant() returns (int256) {
```
chrismaree, line 239: this method will revert if the State is not `Expired` or `Resolved` due to the `_settle` function. This is similar behavior to the `Voting` contract's `GetPrice` method. 

Do you want to add a `hasPrice` function to enable a non-reverting function to check if the `OptimisticOracle` has a price?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -557,25 +562,6 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
         return;
     }
 
-    /**
-     * @notice Drains any excess balance of the provided ERC20 token to a pre-selected beneficiary.
-     * @dev This will drain down to the amount of tracked collateral and drain the full balance of any other token.
-     * @param token address of the ERC20 token whose excess balance should be drained.
-     */
-    function trimExcess(IERC20 token) external nonReentrant() returns (FixedPoint.Unsigned memory amount) {
```
chrismaree, line 565: were you able to remove this because of `gulp`?

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
chrismaree: with that said, the right answer could be to simply call `getState` and validate if the request is within a state that will return a price & not revert when calling `getPrice`.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,357 @@
...
+
+import "../../../oracle/interfaces/StoreInterface.sol";
+import "../../../oracle/interfaces/OracleInterface.sol";
+import "../../../oracle/interfaces/FinderInterface.sol";
+import "../../../oracle/interfaces/IdentifierWhitelistInterface.sol";
+import "../../../oracle/implementation/Constants.sol";
+
+import "../../../common/implementation/Testable.sol";
+import "../../../common/implementation/Lockable.sol";
+import "../../../common/implementation/FixedPoint.sol";
```
chrismaree, line 17: The only place you use `FixedPoint` is `payOracleFeesErc20`. Would it make sense to replace the use of `int` in this contract with a `FixedPoint.Signed`, seeing we have this functionality now?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
chrismaree: nvm. the gulp function acts to sync up the `cumulativeFeeMultiplier`, and does not transfer anything. can you explain why you removed this function?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -75,16 +85,56 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+    }
+
+    function proposeNewRate(FixedPoint.Signed memory rate, uint256 timestamp)
+        external
+        nonReentrant()
+        updateFundingRate()
+    {
+        require(fundingRate.proposalTime == 0, "Proposal in progress");
+
+        // Timestamp must be after the last funding rate update time, within the last 30 minutes, and it cannot be more than 90 seconds ahead of the block timestamp.
```
chrismaree, line 111: nit: line length.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -75,16 +85,56 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
 
-        lastUpdateTime = getCurrentTime();
+        // Set up optmistic oracle.
+        bytes32 identifier = fundingRate.identifier;
+        uint256 bondToPay = optimisticOracle.requestPrice(identifier, timestamp, collateralCurrency, reward.rawValue);
+        optimisticOracle.setRefundOnDispute(identifier, timestamp);
 
-        // Seed the initial funding rate in the cumulativeFundingRateMultiplier 1.
-        cumulativeFundingRateMultiplier = FixedPoint.fromUnscaledUint(1);
+        // TODO: set custom bond amount?
```
chrismaree, line 132: this would be set in accordance with the previous bonds, as a function of PFC, right? cool that you can leverage that functionality right here to scale the bond size depending on the individual proposal in question.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Nope, just moved the funding rate identifier to this struct.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: I was actually thinking that there's no reason to use FixedPoint in this class since it doesn't do any arithmetic. I figured it would be simpler just to stick with primitives. Thoughts?

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Good point. This was leftover from some previous churn, definitely agree that it should go below the requires.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Agreed, will add.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Ha, oops! Great catch.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Not super opinionated here. Was trying to reduce the number of methods offered initially, but ultimately ended up adding `getState` and `getRequest` to provide users with more info where needed. I was initially thinking you would just call `getPrice` inside a try-catch instead of using `hasPrice`, but would be okay adding a `hasPrice` too. Thoughts?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Done.

mrice32:
@chrismaree @nicholaspai I think this is getting close to a final version. I've added a unit test suite for the optimistic oracle. I'm still working through the perp tests.

A few minor notes:
- I simplified the repay function to save bytecode. I think the only material difference between this version and the previous one is that the new version is probably less gas efficient and it requires the user to have approved the contract to pull collateral.
- Also for bytecode reasons, I dropped the gulp AND the trim excess functions. I will try to slip the gulp back in if I can because I think it'll ultimately be the cheaper one.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -225,8 +227,8 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         external
         fees()
         notEmergencyShutdown()
-        updateFundingRate()
         nonReentrant()
+        updateFundingRate()
```
nicholaspai, line 231: Why do this? This might actually throw if its after `nonReentrant`

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -424,30 +426,8 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
...
-        // Update the totalTokensOutstanding after redemption.
-        totalTokensOutstanding = totalTokensOutstanding.sub(numTokens);
-
-        emit Repay(msg.sender, numTokens.rawValue, newTokenCount.rawValue);
-
-        // Transfer the tokens back from the sponsor and burn them.
-        tokenCurrency.safeTransferFrom(msg.sender, address(this), numTokens.rawValue);
-        tokenCurrency.burn(numTokens.rawValue);
+    function repay(FixedPoint.Unsigned memory numTokens) public {
+        deposit(redeem(numTokens));
```
nicholaspai, line 430: this is genius

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,53 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+
+    function proposeNewRate(FixedPoint.Signed memory rate, uint256 timestamp)
+        external
+        nonReentrant()
+        updateFundingRate()
+    {
+        require(fundingRate.proposalTime == 0, "Proposal in progress");
+
+        // Timestamp must be after the last funding rate update time, within the last 30 minutes, and it cannot be more
+        // than 90 seconds ahead of the block timestamp.
```
nicholaspai, line 119: This comment isn't consistent with the inequality, either the inequality should be `timestamp <= currentTime.add(90)` or the comment should read "it cannot be 90 seconds or more ahead of the block timestamp"

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
mrice32: Yeah, this is probably not the right way to fix this, but the updateFundingRate is _not_ currently protected by a reentrancy guard because it isn't a public method. So the quick fix was to put it inside the re-entrancy guard modifier whenever used. But this means we have an inconsistent setup where some modifiers are protected on their own and some need to be placed after the re-entrancy guard.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
mrice32: I initially removed it because I was relying on periodically calling gulp to deal with refunds. Now that the refunds are gone, it was just removed to save bytes. Will attempt to re-add. I think we should _either_ have gulp or trimExcess. My preference would be gulp as I think it has a smaller impact on bytecode and is more consistent with the emp design.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: This is being done now!

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
mrice32: What do you think?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Ah, I see. Fixed.

mrice32:
Update: added a new suite of funding rate applier unit tests. The only tests that haven't been updated are the core perp tests, which shouldn't be super coupled to the funding rate logic.

@chrismaree @nicholaspai PTAL whenever you get a chance.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,60 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+        updateFundingRate()
+        returns (FixedPoint.Unsigned memory totalBond)
+    {
+        require(fundingRate.proposalTime == 0, "Proposal in progress");
+
+        // Timestamp must be after the last funding rate update time, within the last 30 minutes, and it cannot be more
+        // than 90 seconds ahead of the block timestamp.
+        uint256 currentTime = getCurrentTime();
+        uint256 updateTime = fundingRate.updateTime;
+        require(
```
chrismaree, line 123: these times seem pretty arbitary. do we want to be able to tune them? what if we want to change them?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,60 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+        bytes32 identifier = fundingRate.identifier;
+        optimisticOracle.requestPrice(identifier, timestamp, collateralCurrency, 0);
+        totalBond = FixedPoint.Unsigned(
+            optimisticOracle.setBond(identifier, timestamp, _pfc().mul(fundingRate.bondPercentage).rawValue)
+        );
+
+        // Pull bond from caller and send to optimistic oracle.
+        if (totalBond.isGreaterThan(0)) {
+            collateralCurrency.safeTransferFrom(msg.sender, address(this), totalBond.rawValue);
+            collateralCurrency.safeIncreaseAllowance(address(optimisticOracle), totalBond.rawValue);
```
chrismaree, line 143: is there a good reason to not set a large approval at the start of contract deployment to remove the need for this here? clearly, if the `optimisticOracle` gets compromised then it could drain the perp, which is a good enough reason in and of its self but. I thought it worthwhile to discuss.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -31,34 +34,43 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
 
-    // Tracks the cumulative funding payments that have been paid to the sponsors.
-    // The multiplier starts at 1, and is updated by computing cumulativeFundingRateMultiplier * (1 + effectivePayment).
-    // Put another way, the cumulativeFeeMultiplier is (1 + effectivePayment1) * (1 + effectivePayment2) ...
-    // For example:
-    // The cumulativeFundingRateMultiplier should start at 1.
-    // If a 1% funding payment is paid to sponsors, the multiplier should update to 1.01.
-    // If another 1% fee is charged, the multiplier should be 1.01^2 (1.0201).
-    FixedPoint.Unsigned public cumulativeFundingRateMultiplier;
+    uint256 emergencyShutdownTimestamp;
```
nicholaspai, line 64: did you intend to leave out the `public` key word?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -31,34 +34,43 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
 
     /****************************************
      *                EVENTS                *
      ****************************************/
 
     event NewFundingRate(
```
nicholaspai, line 70: This should probably be renamed more accurately to `NewFundingRateMultiplier` wdyt

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,60 @@ abstract contract FundingRateApplier is Testable, Lockable {
      * @notice Constructs the FundingRateApplier contract. Called by child contracts.
      * @param _finderAddress Finder used to discover financial-product-related contracts.
      */
-    constructor(address _finderAddress) public {
-        finder = FinderInterface(_finderAddress);
+    constructor(
+        FixedPoint.Unsigned memory _fundingRateBondPercentage,
```
nicholaspai, line 93: It will be nice to read `_fundingRateBondPercentage` and `_fundingRateRewardRate` from the ConfigStore

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: We can add them to the ConfigStore?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,60 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+            timestamp > updateTime && timestamp >= currentTime.sub(30 minutes) && timestamp <= currentTime.add(90),
+            "Invalid proposal time"
+        );
+
+        // Set the proposal time in order to allow this contract to track this request.
+        fundingRate.proposalTime = timestamp;
+
+        OptimisticOracle optimisticOracle = _getOptimisticOracle();
+
+        // Set up optmistic oracle.
```
nicholaspai, line 133:
```suggestion
        // Set up optimistic oracle.
```

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -31,34 +34,43 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
-    // address.
-    FinderInterface private finder;
-
-    // Last time the `cumulativeFundingRateMultiplier` was updated.
-    uint256 public lastUpdateTime;
+    struct FundingRate {
+        // Current funding rate value.
+        FixedPoint.Signed rate;
+        // Identifier to retrieve the funding rate.
+        bytes32 identifier;
```
nicholaspai, line 41: Can we move the `fundingRateIdentifier` state var to this contract? I know right now its in the `FeePayer` and removing it from there would break tests, but once we deprecate the FRS can we add it here?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Can we move the fundingRateIdentifier state var to this contract? I know right now its in the FeePayer and removing it from there would break tests, but once we deprecate the FRS can we add it here?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -94,15 +153,65 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+                // If the request is more recent than the last update then we should update the funding rate.
+                if (timestamp >= lastUpdateTime) {
+                    // Update funding rates
+                    fundingRate.rate = FixedPoint.Signed(price);
+                    fundingRate.updateTime = timestamp;
+
+                    // If there was no dispute, send a reward.
+                    if (request.disputer == address(0)) {
+                        FixedPoint.Unsigned memory reward =
+                            _pfc().mul(fundingRate.rewardRate).mul(timestamp.sub(lastUpdateTime));
```
nicholaspai, line 199: Thoughts on a future PR: When this `rewardRate` is read from the ConfigStore, do you think we should snapshot it (as a %) at propose time? Or should we just leave it here, which seems simpler, and assume that the ConfigStore is always correct.

nicholaspai, line 199: Thinking through weird scenarios, is there ever a situation where `timestamp.sub(lastUpdateTime) < 0`?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree: for sure, good idea.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -94,15 +153,65 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
 
-    function _getFundingRateStore() internal view returns (FundingRateStoreInterface) {
-        return FundingRateStoreInterface(finder.getImplementationAddress("FundingRateStore"));
+    function _getOptimisticOracle() internal view returns (OptimisticOracle) {
+        return OptimisticOracle(finder.getImplementationAddress(OracleInterfaces.OptimisticOracle));
     }
 
-    function _getLatestFundingRate() internal view returns (FixedPoint.Signed memory) {
-        return _getFundingRateStore().getFundingRateForContract(address(this));
+    // function priceSettled(
```
chrismaree, line 163: nit: was this meant to be removed?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -117,27 +226,27 @@ abstract contract FundingRateApplier is Testable, Lockable {
         }
 
         uint256 currentTime = getCurrentTime();
-        uint256 paymentPeriod = currentTime.sub(lastUpdateTime);
+        uint256 paymentPeriod = currentTime.sub(fundingRate.applicationTime);
```
nicholaspai, line 229: Where do you set `applicationTime`, I assume it should be in `_getLatestFundingRate`?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: I think its fine as long as you leave a comment because I could see this easily confusing us in the future and leading to strange test breakage

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,368 @@
...
+            request.resolvedPrice = _getOracle().getPrice(identifier, timestamp);
+            bool disputeSuccess = request.resolvedPrice != request.proposedPrice;
+            uint256 payout = request.bond.mul(2).add(request.finalFee).add(request.reward);
+            request.currency.safeTransfer(disputeSuccess ? request.disputer : request.proposer, payout);
+        } else {
+            revert("_settle: not settleable");
+        }
+
+        // Event.
+        emit Settle(requester, request.proposer, request.disputer, identifier, timestamp, request.resolvedPrice);
```
chrismaree, line 300: might be useful to add payout to this

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: i.e. "placing this modifier after the reentrancy-guard modifier protects its internal methods against reentrancy"

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree, line 163: guessing you can remove this?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -77,13 +89,60 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+        );
+
+        // Set the proposal time in order to allow this contract to track this request.
+        fundingRate.proposalTime = timestamp;
+
+        OptimisticOracle optimisticOracle = _getOptimisticOracle();
+
+        // Set up optmistic oracle.
+        bytes32 identifier = fundingRate.identifier;
+        optimisticOracle.requestPrice(identifier, timestamp, collateralCurrency, 0);
```
nicholaspai, line 135: Why do you set reward to 0 here, and then you pay out the reward in `_getLatestFundingRate()`? Could we just remove the rewards param from this `optimisticOracle` method?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,368 @@
...
+            customLiveness: 0
+        });
+
+        if (reward > 0) {
+            currency.safeTransferFrom(msg.sender, address(this), reward);
+        }
+
+        emit RequestPrice(msg.sender, identifier, timestamp);
+
+        return finalFee.mul(2);
```
nicholaspai, line 135: I don't see this return val being used in the FundingRateApplier

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,368 @@
...
+        uint256 timestamp,
+        uint256 bond
+    ) external nonReentrant() returns (uint256 totalBond) {
+        require(getState(msg.sender, identifier, timestamp) == State.Requested, "setBond: Requested");
+        Request storage request = _getRequest(msg.sender, identifier, timestamp);
+        request.bond = bond;
+        return bond.add(request.finalFee);
+    }
+
+    function setRefundOnDispute(bytes32 identifier, uint256 timestamp) external nonReentrant() {
```
nicholaspai, line 149: This also seems deprecated, I'm going to stop commenting on this contract now since it seems like there are still local changes to be pushed!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,368 @@
...
+    function disputePrice(
+        address requester,
+        bytes32 identifier,
+        uint256 timestamp
+    ) external {
+        // Note: re-entrancy guard is done in the inner call.
+        disputePriceFor(msg.sender, requester, identifier, timestamp);
+    }
+
+    function getPrice(bytes32 identifier, uint256 timestamp) external nonReentrant() returns (int256) {
```
nicholaspai, line 250: Would we make this an `onlyRegisteredContract` method?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Okay, so I'll leave these hardcoded for now, but we'll consider adding them to ConfigStore in a follow-up?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: As discussed IRL, the optimistic oracle is meant to be more general, so even though we don't use the rewards, it's likely that others might want to.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: It's a great question. The reason I chose to do it here is that we're pulling the optimistic oracle address from the finder, so if it were upgraded for some reason, that big initial approval would no longer apply, which would make this revert (and lock up indefinitely I think). So, we'd effectively need to at least check and conditionally set the approval here if we wanted to do it that way. I figured it was simpler just to do the approval on each transaction rather than checking past approval. 

Thoughts?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yep, I'll remove.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yeah, I see what you're saying. A proposer could technically have the rug pulled out from under them if the percentage changes between propose time and expiration time. I, personally, think the simpler solution is probably better because: 
1. The minimum timelock time is 1 day and this uses the default optimistic oracle liveness of 2 hours. So the proposer would be able to know this change was coming before making the proposal.
2. There is already some degree of trust put in the owner of the config store.

Thoughts?

mrice32: I had the exact same question -- that's why I put this inside the outer `if` that checks that condition, so this whole update would be skipped in that case.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: It gets set at the end of this method (it basically just replaced lastUpdateTime).

This did make me realize that we don't initialize it in the constructor to the current time, which causes the first funding rate to be over a very long period. This didn't really have much of an impact because the funding rate starts at 0, but it's probably good to set it just to avoid any strange edge cases. So I'm adding a line to initialize it as it was being done before.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
mrice32: Okay, will do. I'll consider fixing this in a more elegant way in a follow-up, but for now, I'll add a comment since I agree that the subtlety is confusing.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Yeah, the funding rate applier changes the bond in a subsequent call, and the return value from that call is stored (since it would override this value anyway). However, I think returning the bond is generally useful for callers. In fact, I think I'm going to add it to propose and dispute too for good measure.

Thoughts?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: This is a really good point.

I _don't_ think we do. Would be great to hear your thoughts here: the "master plan" is for this optimistic oracle to effectively allow callers to be trial users of the DVM (this is something @hlambur has been pushing for to improve UX). If the caller is paying their fees fairly then the optimistic oracle should work fine. If the caller is not paying fees, the DVM voters can choose to respond with an invalid value (INT_MIN, I was thinking). This is something we don't have to 100% figure out now since a the logic is basically off-chain, but it would be nice to develop this in a way that's compatible with the DVM trial idea.

What are your thoughts?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Will do!

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yeah, 100% agreed.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Agreed. Will change.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Nope, nice catch -- this was a typo when I was resolving merge conflicts.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Merging this in soon.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Sure

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: I agree with @mrice32 its more expensive gas-wise but it is more safe in that no transaction should fail

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Agreed, but since this contributes negatively to proposer UX can you add a comment about the fact that reward rate could change?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: > This did make me realize that we don't initialize it in the constructor to the current time, which causes the first funding rate to be over a very long period. This didn't really have much of an impact because the funding rate starts at 0, but it's probably good to set it just to avoid any strange edge cases. So I'm adding a line to initialize it as it was being done before.

Yes actually I do this exact same logic with the FundingRateStore and attempt to set it to the current time as soon as possible [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/funding-rate-store/implementation/FundingRateStore.sol#L190).

nicholaspai: > It gets set at the end of this method (it basically just replaced lastUpdateTime). Do you think it should be updated elsewhere?

My original question came up because I couldn't find the `fundingRate.applicationTime = currentTime` line but I see iw not at the end.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: Yeah that's reasonable

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yep!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Done.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: I think it should be `onlyRegisteredContract` until we figure out the economic consequences of opening up DVM state to everyone. I think this gets into parasitic DVM usage.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -14,14 +14,18 @@ import "../../oracle/interfaces/StoreInterface.sol";
 import "../../oracle/interfaces/FinderInterface.sol";
 import "../../oracle/implementation/Constants.sol";
 
-import "../funding-rate-store/interfaces/FundingRateStoreInterface.sol";
+// TODO: point this at an interface instead.
```
nicholaspai, line 17: Are you gonna do this in a follow up PR?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -31,34 +35,41 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+        uint256 applicationTime;
+        // The time for the active (if it exists) funding rate proposal. 0 otherwise.
+        uint256 proposalTime;
+    }
 
-    // Last time the `cumulativeFundingRateMultiplier` was updated.
-    uint256 public lastUpdateTime;
+    FundingRate public fundingRate;
 
-    // Timestamp used in case of emergency shutdown.
```
nicholaspai, line 42: Why remove this comment?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualCreator.sol
```diff
@@ -35,6 +35,8 @@ contract PerpetualCreator is ContractCreator, Testable, Lockable {
         address collateralAddress;
         bytes32 priceFeedIdentifier;
         bytes32 fundingRateIdentifier;
+        FixedPoint.Unsigned fundingRateRewardRate;
```
nicholaspai, line 38: Do we need to pass these in anymore to the Perpetual constructor?

mrice32: Ah, nice catch.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -309,7 +305,7 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
     /**
      * @notice Cancels a pending withdrawal request.
      */
-    function cancelWithdrawal() external notEmergencyShutdown() updateFundingRate() nonReentrant() {
```
nicholaspai, line 312: Why are you removing `updateFundingRateHere()`? I think it's actually ok not to update the fee/funding multipliers because this method doesn't modify sponsor positions, but curious if this was a mistake?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -248,7 +246,6 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
         public
         notEmergencyShutdown()
         noPendingWithdrawal(msg.sender)
-        updateFundingRate()
```
nicholaspai, line 251: Same question as for `cancelWithdrawal`, are you just making this consistent with it not changing the fee-multiplier?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,385 @@
...
+            customLiveness: 0
+        });
+
+        if (reward > 0) {
+            currency.safeTransferFrom(msg.sender, address(this), reward);
+        }
+
+        emit RequestPrice(msg.sender, identifier, timestamp, address(currency), reward, finalFee);
+
+        return finalFee.mul(2);
```
nicholaspai, line 143: Why is this final fee * 2? Can you add a comment

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
mrice32: Nope, this was intended. I was just matching the regular fees.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
mrice32: Yep, exactly.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Probably a merge mistake :)

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yeah, I think we should leave this for a follow-up since it doesn't change any core logic.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: My only worry with doing that is having it audited with that restriction and then needing another audit before we can open it up.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: will do.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Done.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: Hm I'm just not sure this would be caught by a security audit and it doesn't seem to increase the "code" attack vector. It just might increase the economic attack vector, I think.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Sorry, I'm a little confused. The other one exists in the fundingRateFeePayer, but that entire contract will be deprecated after this PR as it's no longer in the inheritance tree for the perp.

So I was thinking it makes sense to just leave that contract be and remove it along with the FRS in a later PR. Thoughts?

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Done.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Done.

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
mrice32: Done.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Done.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualCreator.sol
mrice32: Done.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Done.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Agreed we will remove `FundingRatePayer` contract, so in that process are you saying we'll move the `fundingRateIdentifier` into this contract?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,389 @@
...
+
+        if (reward > 0) {
+            currency.safeTransferFrom(msg.sender, address(this), reward);
+        }
+
+        emit RequestPrice(msg.sender, identifier, timestamp, address(currency), reward, finalFee);
+
+        // The total bond is the final fee + the default bond, which is set to the final fee. So the total bond is
+        // 2x the final fee.
+        return finalFee.mul(2);
```
nicholaspai, line 145: Can the `default bond` get overwritten anywhere? or is that set on this contract deployment

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -68,22 +81,88 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
+
+        // Seed the cumulative multiplier as 1, from which it will be scaled as funding rates are applied over time.
+        fundingRate.cumulativeMultiplier = FixedPoint.fromUnscaledUint(1);
         emergencyShutdownTimestamp = 0;
-        lastUpdateTime = getCurrentTime();
 
-        // Seed the initial funding rate in the cumulativeFundingRateMultiplier 1.
-        cumulativeFundingRateMultiplier = FixedPoint.fromUnscaledUint(1);
+        fundingRate.identifier = _fundingRateIdentifier;
+        configStore = ConfigStoreInterface(_configStoreAddress);
```
nicholaspai, line 118: Should we do any checking that _configStoreAddress is a valid contract?

nicholaspai (approved):
Thanks for reviewing all my comments, left a few more!

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree: Prehaps a good idea to verify it here in the constructor. If it's not the perp is pretty broken

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree: Ye, good enough reason! Other option could be large approval at deployment then check approval size before doing the transfer and increase if need be. 

This is somewhat complex for no reason so I'm happy to keep it how you've got it here

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -94,15 +173,56 @@ abstract contract FundingRateApplier is Testable, Lockable {
...
     }
 
-    function _getLatestFundingRate() internal view returns (FixedPoint.Signed memory) {
-        return _getFundingRateStore().getFundingRateForContract(address(this));
+    function _getConfig() internal view returns (ConfigStoreInterface.ConfigSettings memory) {
+        return configStore.getCurrentConfig();
+    }
+
+    function _getLatestFundingRate() internal returns (FixedPoint.Signed memory) {
+        uint256 timestamp = fundingRate.proposalTime;
```
chrismaree, line 188: Nit: this variable could be better named prehaps. When you are comparing timestamp later on it's not clear that it's the funding rate proposal time by just looking at the variable name

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Or do you want to just pass it here by construction, looks like no other contract in the Perp system needs access to the `fundingRateIdentifier` so it might be cleaner to just store the identifier at the state level. WDYT

packages/core/contracts/financial-templates/funding-rate-store/implementation/OptimisticOracle.sol
nicholaspai: A `hasPrice`, as opposed to a `getState`, reduces burden on the calling contract but trades off with less information. It would also match the Voting.sol interface which might be useful

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Yeah there are no fallback values or anything if the ConfigStore reverts on `getConfigSettings`

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Good point. There are a lot of timestamps floating around. More specific names would be helpful.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: The default bond is the final fee. That defaulting logic is set on contract deployment. However, a contract that requests a price can customize the bond by calling `setBond()`. Do you think this should be otherwise changeable?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: How should we check that? By doing a test `getConfigSettings` call or doing an `isContract` check?

My only minor objection to doing checks in the contract is that we decided not to do similar checks for the synthetic token to save bytecode.

nicholaspai: > My only minor objection to doing checks in the contract is that we decided not to do similar checks for the synthetic token to save bytecode.

This is a good point, and it would be consistent to maintain a "buyer beware" stance here with this method (i.e. don't interact with a Perp that uses a bad ConfigStore).

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: Ah I see this is a big confusing. So the actual proposal bond is set in `propose()` by this line: `totalBond = request.bond.add(request.finalFee)`.

So basically this return value, defaulted and hardcoded to ` finalFee.mul(2)` is indicating to the caller that this is the default bond value. I think comments on this function specifically about the return value (e.g. "this function returns the initial proposal bond for this request, which can be changed by the requested of the price") would clear this up.

This makes sense to me now.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Done.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Yeah, like the synthetic token, if you launch this contract through the contract creator (and that's the only way it's approved), the config store is guaranteed to be valid.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Gotcha. +1 -- will add clearer comments!

mrice32: Done.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -102,6 +102,9 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
 
         // Exit early if there is no collateral from which to pay fees.
         if (collateralPool.isEqual(0)) {
+            // Note: set the lastPaymentTime in this case so the contract is credited for paying during periods when it
+            // has no locked collateral.
+            lastPaymentTime = time;
```
mrice32, line 107: Note: this was a minor bug that existed in our existing implementation. Fees are not "credited" while the pfc == 0. I think this timestamp should be updated even if we return early cc @chrismaree @nicholaspai.

mrice32 (commented):
@nicholaspai @chrismaree 
The tests are updated. I think this is ready to merge once y'all are happy.

packages/core/contracts/financial-templates/common/FeePayer.sol
nicholaspai: Ah, so IIUC without this change, once its PfC > 0, the contract would be charged retrospectively for all of the time it spent at PfC=0.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -523,17 +494,13 @@ contract PerpetualPositionManager is FundingRatePayer, FundingRateApplier {
...
-    }
-
     /**
      * @notice Premature contract settlement under emergency circumstances.
      * @dev Only the governor can call this function as they are permissioned within the `FinancialContractAdmin`.
      * Upon emergency shutdown, the contract settlement time is set to the shutdown time. This enables withdrawal
      * to occur via the `settleEmergencyShutdown` function.
      */
-    function emergencyShutdown() external override notEmergencyShutdown() updateFundingRate() nonReentrant() {
+    function emergencyShutdown() external override notEmergencyShutdown() fees() nonReentrant() {
```
nicholaspai, line 503: Wait, `emergencyShutdown` was not paying regular fees before, but now it is?

mrice32: Yeah, I think it should be, right?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -0,0 +1,389 @@
...
+        uint256 bond;
+        uint256 customLiveness;
+    }
+
+    mapping(bytes32 => Request) public requests;
+
+    FinderInterface public finder;
+    uint256 public defaultLiveness;
+
+    constructor(
```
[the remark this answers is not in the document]
mrice32: Ahhh, no I will add it before merging. Meant to add it last night. Sorry about that. Thanks for pointing it out.

packages/core/contracts/financial-templates/common/FeePayer.sol
mrice32: Yep, exactly.

chrismaree: good catch!

chrismaree (approved):
This PR is a beast to review but also super exciting! I can't wait to see the Optimistic oracle in production. Also, it came together unbelievably quickly, which was awesome to watch.

It looks good to me, once you've added nat spec to your new contracts.

Great work @mrice32!

mrice32:
> It looks good to me, once you've added nat spec to your new contracts.

@chrismaree added natspec. Will merge once ci passes.

chrismaree:
> > It looks good to me, once you've added nat spec to your new contracts.
> 
> @chrismaree added natspec. Will merge once ci passes.

image

PR #2213, opened 2020-11-23
bug(liquidatable): snapshot funding-rate adjusted token debt at liquidation, not settlement, time

nicholaspai:
**Motivation**

#2211

**Summary**

**Details**

**Issue(s)**

Fixes #2211

PR #2220, opened 2020-11-24
bug(perp-liquidatable): don't use adjusted debt when comparing liquidationPrice against CR

nicholaspai:

**Motivation**

Fix this bug introduced in #2213 which I merged accidentally prematurely. The liquidator should specify the sponsor's current collateral-token ratio, which is not based on any adjusted value. It is simply the `rawCollateral : rawTokens`.


mrice32 (approved):
LGTM!

chrismaree (approved):

mrice32 (approved):
Looks great! +1

Thanks for putting a comment in the tests that imply that some of the language here is a bit misleading.

PR #2221, opened 2020-11-25
feat(emp): financial product library to apply price identifier transformation

chrismaree:
**Motivation**

Issue #2160 & https://github.com/UMAprotocol/protocol/issues/2231

Add the ability to transform contract price identifier based on an external library.

Follows on from PR https://github.com/UMAprotocol/protocol/pull/2185 which applied similar logic to the settlement price and PR https://github.com/UMAprotocol/protocol/pull/2191 which applies similar logic to the collateral requirement.

**Summary**

This PR introduces a function `transformPriceIdentifier` to the `FinancialProductLibrary` which enables the price identifier a contract to be transformed by an external library. This is useful for contracts that might want to use a different identifier pre and post expiration. An example of this is the uGas synth that requires a TWAP pre-expiration for liquidations and a long-running medium price post expiration for settlement.

**Details**

`transformPriceIdentifier` is applied to the `PricelessPositionManager` contract by adding a `try`/`catch` wrapped method to transform the price identifer requirement of the contract, if the library has the provided method. `PricelessPositionManager` uses the `priceIdentifier` in three places:
a) https://github.com/UMAprotocol/protocol/blob/2c7c11dbd107bb5f3093300d5d099ac3435ccceb/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L788
b) https://github.com/UMAprotocol/protocol/blob/2c7c11dbd107bb5f3093300d5d099ac3435ccceb/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L795
c) https://github.com/UMAprotocol/protocol/blob/2c7c11dbd107bb5f3093300d5d099ac3435ccceb/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L796

This PR modifies this logic at by applying the transformation before using the identifier in these methods.

`Liquidatable` also uses the `priceIdentifier` indirectly by calling `_getOraclePrice`. As a result, the changes to `PricelessPositionManager` sufficiently change `Liquidatable` to be compatible with this library. 

Unit tests on both `PricelessPositionManager` & `Liquidatable` were added to verify the functionality.

close 2231


packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -939,251 +939,614 @@ contract("PricelessPositionManager", function(accounts) {
     // No excess collateral after all have settled.
     await expectNoExcessCollateralToTrim();
   });
+  describe("Financial Product Library", () => {
```
chrismaree, line 942: introducing this extra describe added an extra indent through this set of tests and totally broke the git diff.

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
```diff
@@ -0,0 +1,59 @@
...
+
+    /**
+     * @notice Enables the deployer of the library to set the transformed identifier for an associated financial product.
+     * @param financialProduct address of the financial product.
+     * @param transformedIdentifier the identifier for the financial product to be used if the contract is post expiration.
+     * @dev Note: a) Only the owner (deployer) of this library can set identifier transformations b) The identifier Cant
+     * be set to blank. c) A transformed price can only be set once to prevent the deployer from changing it after the fact.
+     * d)  financialProduct must exposes an expirationTimestamp method.
+     */
+    function setFinancialProductTransformedIdentifier(address financialProduct, bytes32 transformedIdentifier)
```
nicholaspai, line 24: Should this be `nonReentrant`?

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -699,6 +700,24 @@ contract PricelessPositionManager is FeePayer {
...
     }
 
+    /**
+     * @notice Accessor method to compute a transformed price identifier using the finanicalProductLibrary specified
+     * at contract deployment. If no library was provided then no modification to the identifier is done.
+     * @param requestTime timestamp the identifier is to be used at.
+     * @return transformedPrice price with the transformation function applied to it.
+     * @dev This method should never revert.
+     */
+    function transformPriceIdentifier(uint256 requestTime) public view returns (bytes32) {
```
nicholaspai, line 710: These should also be nonReentrant b/c they call an external contract?

chrismaree: We could add that but this is a view function so don't really see any issues with not having it. 

I'll add it for consistency

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
chrismaree: It is permissioned to be only called by the owner of the library and so reentrancy to is not really a risk here.

nicholaspai: Not sure I agree, the owner can be malicious and use a reentrancy attack after enough collateral has been deposited in the system. Also I think its better to just be safe and make every public method in the Perp system reentrancy-guarded

nicholaspai (commented):
This is a really cool feature and I have two concerns:
- You assume that the `transformPriceIdentifier` can only be set once, but what if the `financialProductLibrary` is upgraded in the Finder? This could be problematic for a live EMP that had transformed its identifier but would now see its identifier reset to its original. Perhaps its better to snapshot the `transformPriceIdentifier` into the EMP state right after its initially transformed, similar to how the `expirationPrice` is snapshotted, protecting us against a DVM upgrade (which happened IRL!).
- I am a bit uncomfortable with public methods calling external contracts not being reentrancy-guarded, even if they are permissioned methods. Can we trust `financialProductLibrary` owners that much?

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
nicholaspai: We guard view functions with `nonReentrantView` all throughout the EMP

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
```diff
@@ -0,0 +1,59 @@
...
+        public
+        onlyOwner
+    {
+        require(transformedIdentifier != "", "Cant set to empty transformation");
+        require(financialProductTransformedIdentifiers[financialProduct] == "", "Transformation already set");
+        require(ExpiringContractInterface(financialProduct).expirationTimestamp() != 0, "Invalid EMP contract");
+        financialProductTransformedIdentifiers[financialProduct] = transformedIdentifier;
```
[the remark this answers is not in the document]
mrice32: We could, but IIUC, the transaction will fail later for the same reason, so this doesn't really change the outcome. One advantage of _not_ checking for this is that this contract doesn't require the finder or any other DVM contracts to work.

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * @notice Returns a transformed price identifier if the contract is pre-expiration and no transformation if post.
+     * @param identifier input price identifier to be transformed.
+     * @param requestTime timestamp the identifier is to be used at.
+     * @return transformedPriceIdentifier the input price identifier with the transformation logic applied to it.
+     */
+    function transformPriceIdentifier(bytes32 identifier, uint256 requestTime) public view override returns (bytes32) {
+        require(financialProductTransformedIdentifiers[msg.sender] != "", "Caller has no transformation");
+        // If the request time is before contract expiration then return the transformed identifier. Else, return the
+        // transformed price identifier.
```
mrice32, line 52: Just curious -- why is the pre-expiration one the "transformed" one and the expiry one the "non-transformed" one?

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
chrismaree: no problem. I will add it 👍.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
chrismaree: More than happy to add this. As discussed offline, this will require a small refactor to use an internal method that has no guards & then applying the guards on a public accessor. I will save this for a separate PR. I've created an issue to track it here: https://github.com/UMAprotocol/protocol/issues/2232

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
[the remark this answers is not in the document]
chrismaree: I agree with @mrice32. Also, this library is really just meant as an example for now and I suspect it might require small changes potentially for a real product using this kind of transformation logic so better to keep it as simple as possible for now.

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
```diff
@@ -0,0 +1,59 @@
...
+     * @param transformedIdentifier the identifier for the financial product to be used if the contract is post expiration.
+     * @dev Note: a) Only the owner (deployer) of this library can set identifier transformations b) The identifier Cant
+     * be set to blank. c) A transformed price can only be set once to prevent the deployer from changing it after the fact.
+     * d)  financialProduct must exposes an expirationTimestamp method.
+     */
+    function setFinancialProductTransformedIdentifier(address financialProduct, bytes32 transformedIdentifier)
+        public
+        onlyOwner
+    {
+        require(transformedIdentifier != "", "Cant set to empty transformation");
```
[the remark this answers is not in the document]
chrismaree: The `FinancialProductLibrary` is not registered in the finder actually and is expected to be deployed alongside an EMP. It is up to the financial contract deployer to deploy the library and provide the address to the EMP. 

For now, it has no migrations and no connection with the creator either. However, this is a good point and I believe the creator might need to be updated as well to accommodate the constructor param. I have added an issue to track this here https://github.com/UMAprotocol/protocol/issues/2233

packages/core/contracts/financial-templates/common/financial-product-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.sol
chrismaree: My thinking was that in most scenarios you're payout as a user will be the non-transformed (or the parameterized) identifier. The "transformed" in this scenario is used only in case of liquidation and so is rarer than the non-transformed. 

For example, in the case of uGas it settles to the non-transformed price which is the long-running median and in the case of liquidation(with is less common) only then the identifier is transformed. You can think of it almost like being able to set separate identifiers for liquidations to that used at settlement.

mrice32: Yeah, that makes sense.

nicholaspai (approved):
LGTM thanks for responding to my comemnts

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
mrice32: SGTM

mrice32 (approved):

PR #2222, opened 2020-11-25
improve(position-manager): check for no pending withdrawal in create() can be in modifier not in function body

nicholaspai:


nicholaspai:
Closing because this decision was actually made on purpose, look at the comment above `_positionHasNoPendingWithdrawal`:

```
// Note: This checks whether an already existing position has a pending withdrawal. This cannot be used on the
    // `create` method because it is possible that `create` is called on a new position (i.e. one without any collateral
    // or tokens outstanding) which would fail the `onlyCollateralizedPosition` modifier on `_getPositionData`.
```

PR #2223, opened 2020-11-25
feat(perpetual): Create ConfigStore to house mutable perpetual params (upgradeable after timelock)

### LINKED ISSUE #2219: Build ConfigurationStore to enable time-locked changes to perpetual state variables
**Why**

Certain variables in perpetual contract should be upgradeable by an admin (after passing a timelock) such as:
- funding rate proposer reward rate
- proposer bond %
- timelock liveness

**Details**
- Store should have an owner that can change, possibly by timelock? This is an open question and worth leaving as a comment.
- Store address should be passed into Perpetual on construction

- 4: ~2 hours

nicholaspai:
**Motivation**

#2219 

**Summary**

Builds a new `ConfigStore` on each `createPerpetual()` call. `createPerpetual()` now takes an additional `ConfigSettings` param for default config setting values.

**Details**

Follow up PR will add the `ConfigStore` address to the Perpetual's constructor params and read from the `ConfigStoreInterface` settings such as reward rate, proposer bond pct, etc.

**Issue(s)**

Fixes #2219

### DIFF
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
@@ -147,7 +147,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
     function _validateConfig(ConfigSettings memory config) internal pure {
         // Make sure timelockLiveness is not too long, otherwise contract can might not be able to fix itself
         // before a vulnerability drains its collateral. e.g. 604800 = 7 days.
-        require(config.timelockLiveness <= 604800, "Invalid timelockLiveness");
+        require(config.timelockLiveness <= 7 days, "Invalid timelockLiveness");
 
         // Upper limits for the reward and bond rates are estimated based on offline discussions,
         // and it is expected that these hard-coded limits can change in future deployments.
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
@@ -101,7 +101,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
         pendingConfig = newConfig;
 
         // Use current config's liveness period to timelock this proposal.
-        pendingPassedTimestamp = getCurrentTime() + currentConfig.timelockLiveness;
+        pendingPassedTimestamp = getCurrentTime().add(currentConfig.timelockLiveness);
 
         emit ProposedNewConfigSettings(
             msg.sender,
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
@@ -26,7 +26,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
     // All of the configuration settings available for querying by a perpetual.
     struct ConfigSettings {
         // Liveness period (in seconds) for an update to currentConfig to become official.
-        uint256 updateLiveness;
+        uint256 timelockLiveness;
         // Reward rate paid to successful proposers. Percentage of 1 E.g., .1 is 10%.
         FixedPoint.Unsigned rewardRatePerSecond;
         // Bond % (of given contract's PfC) that must be staked by proposers. Percentage of 1, e.g. 0.0005 is 0.05%.
@@ -49,10 +49,10 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
         address indexed proposer,
         uint256 rewardRate,
         uint256 proposerBond,
-        uint256 updateLiveness,
+        uint256 timelockLiveness,
         uint256 proposalPassedTimestamp
     );
-    event ChangedNewConfigSettings(uint256 rewardRate, uint256 proposerBond, uint256 updateLiveness);
+    event ChangedNewConfigSettings(uint256 rewardRate, uint256 proposerBond, uint256 timelockLiveness);
 
     /****************************************
      *                MODIFIERS             *
@@ -90,7 +90,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
     /**
      * @notice Propose new configuration settings. New settings go into effect
      * after a liveness period passes.
-     * @param newConfig Configuration settings to publish after `currentConfig.updateLiveness` passes from now.
+     * @param newConfig Configuration settings to publish after `currentConfig.timelockLiveness` passes from now.
      * @dev Callable only by owner. Calling this while there is already a pending proposal
      * will overwrite the pending proposal.
      */
@@ -101,13 +101,13 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
         pendingConfig = newConfig;
 
         // Use current config's liveness period to timelock this proposal.
-        pendingPassedTimestamp = getCurrentTime() + currentConfig.updateLiveness;
+        pendingPassedTimestamp = getCurrentTime() + currentConfig.timelockLiveness;
 
         emit ProposedNewConfigSettings(
             msg.sender,
             newConfig.rewardRatePerSecond.rawValue,
             newConfig.proposerBondPct.rawValue,
-            newConfig.updateLiveness,
+            newConfig.timelockLiveness,
             pendingPassedTimestamp
         );
     }
@@ -129,7 +129,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
             emit ChangedNewConfigSettings(
                 currentConfig.rewardRatePerSecond.rawValue,
                 currentConfig.proposerBondPct.rawValue,
-                currentConfig.updateLiveness
+                currentConfig.timelockLiveness
             );
         }
     }
@@ -145,9 +145,9 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
 
     // Use this method to constrain values with which you can set ConfigSettings.
     function _validateConfig(ConfigSettings memory config) internal pure {
-        // Make sure updateLiveness is not too long, otherwise contract can might not be able to fix itself
+        // Make sure timelockLiveness is not too long, otherwise contract can might not be able to fix itself
         // before a vulnerability drains its collateral. e.g. 604800 = 7 days.
-        require(config.updateLiveness <= 604800, "Invalid paramUpdateLiveness");
+        require(config.timelockLiveness <= 604800, "Invalid timelockLiveness");
 
         // Upper limits for the reward and bond rates are estimated based on offline discussions,
         // and it is expected that these hard-coded limits can change in future deployments.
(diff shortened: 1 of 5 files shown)


packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    /**
+     * @notice Propose new configuration settings. New settings go into effect
+     * after a liveness period passes.
+     * @dev Callable only by owner. Calling this while there is already a pending proposal
+     * will overwrite the pending proposal.
+     */
+    function proposeNewConfig(ConfigSettings memory newConfig) external onlyOwner() nonReentrant() updateConfig() {
+        // Alternatively, we can revert if there is already a pending proposal.
+        // require(pendingPassedTimestamp == 0, "Pending proposal exists");
```
chrismaree, line 92: should this be commented out?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,148 @@
...
+
+    function _pendingProposalPassed() internal view returns (bool) {
+        return (pendingPassedTimestamp != 0 && pendingPassedTimestamp <= getCurrentTime());
+    }
+
+    // Use this method to constrain values with which you can set ConfigSettings.
+    function _validateConfig(ConfigSettings memory config) internal pure {
+        // Make sure updateLiveness is not too long, otherwise contract can might not be able to fix itself
+        // before a vulnerability drains its collateral.
+        require(config.updateLiveness <= 604800, "Invalid paramUpdateLiveness");
```
chrismaree, line 146: nothing else we need to verify?

chrismaree (approved):
Nice and simple. how much bytecode does it safe?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: If we enable this, it would mean that you cannot overwrite a proposal once it's in progress. IMO, I think it's better to allow overwrites as long as they restart the timer. WDYT?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,148 @@
...
+        // Alternatively, we can revert if there is already a pending proposal.
+        // require(pendingPassedTimestamp == 0, "Pending proposal exists");
+
+        _validateConfig(newConfig);
+
+        // Warning: This overwrites a pending proposal!
+        pendingConfig = newConfig;
+
+        // Use current config's liveness period to timelock this proposal.
+        pendingPassedTimestamp = getCurrentTime() + currentConfig.updateLiveness;
```
mrice32, line 100: Is this liveness only used for this? Or is this meant to control liveness periods inside the contract as well?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32, line 146: nit: maybe add a variable or a comment noting how much time this is?

mrice32: Should we also verify that the reward rate and proposer bond are < 100%?

mrice32 (commented):

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: Sure, I will remove this code but I wanted to put the comment here to draw your attention to it while reviewing. It was an alternate implementation but @mrice32 I do agree with your intuition.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: We could prob set upper limits for `rewardRate` and `proposerBond` but its a bit hard to determine at which level these become impractical.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: Yes its only meant to control the timelock for this ConfigStore contract

nicholaspai: I can rename it to `configStoreTimelock` for more accuracy

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: ah right I meant to. 7 days

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: Yeah, I think `timelockLiveness` or `configStoreTimelock` would make that more obvious. +1

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: Yeah, agreed, we might want reward rate to be even lower. Do you have a good idea of where that threshold should be? Maybe we should come up with an annualized rate that we think is the max.. 100% per year maybe? Thoughts?

I think 100% makes sense for the bond. I could imagine someone wanting to put it over 1% but no real use case for 100%.

mrice32 (approved):
Looks awesome!

I agree with your intuition that we may want to make the check on reward rate somewhere way below 1% per second. Up to you whether to impose that here or reconsider in a follow-up.

Once https://github.com/UMAprotocol/protocol/pull/2223/files#r530692149 and https://github.com/UMAprotocol/protocol/pull/2223/files#r530690931 are done, LGTM!

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: So the reward % is a bit of a "feel" paramater and we should fully expect to tune it as the contract is live in order to balance the responsiveness of proposers to actual changes in the funding rate. I think 100% a year is a reasonable limit for now (and we can adjust in subsequent perpetuals). 

Regarding the proposer bond, @hlambur and I did some back of the envelope math in [this thread](https://github.com/UMAprotocol/protocol/issues/2039#issuecomment-719734383) and concluded that a 0.04% bond would be equivalent to assuming that the reported funding rate is off from the true funding rate by 200%/year. So I think we should set a limit here << 0.1% per bond.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: Done with `timelockLiveness`

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: Sounds good! Thanks for bringing it up -- hadn't considered overwrites.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,165 @@
...
+     * will overwrite the pending proposal.
+     */
+    function proposeNewConfig(ConfigSettings memory newConfig) external onlyOwner() nonReentrant() updateConfig() {
+        _validateConfig(newConfig);
+
+        // Warning: This overwrites a pending proposal!
+        pendingConfig = newConfig;
+
+        // Use current config's liveness period to timelock this proposal.
+        pendingPassedTimestamp = getCurrentTime() + currentConfig.timelockLiveness;
```
mrice32, line 104:
```suggestion
        pendingPassedTimestamp = getCurrentTime().add(currentConfig.timelockLiveness);
```

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,165 @@
...
+
+    function _pendingProposalPassed() internal view returns (bool) {
+        return (pendingPassedTimestamp != 0 && pendingPassedTimestamp <= getCurrentTime());
+    }
+
+    // Use this method to constrain values with which you can set ConfigSettings.
+    function _validateConfig(ConfigSettings memory config) internal pure {
+        // Make sure timelockLiveness is not too long, otherwise contract can might not be able to fix itself
+        // before a vulnerability drains its collateral. e.g. 604800 = 7 days.
+        require(config.timelockLiveness <= 604800, "Invalid timelockLiveness");
```
mrice32, line 150: nit: would it be cleaner to just do this?
```suggestion
        require(config.timelockLiveness <= 7 days, "Invalid timelockLiveness");
```

mrice32 (approved):
LGTM! Two unimportant nits.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,165 @@
...
+    }
+
+    function _pendingProposalPassed() internal view returns (bool) {
+        return (pendingPassedTimestamp != 0 && pendingPassedTimestamp <= getCurrentTime());
+    }
+
+    // Use this method to constrain values with which you can set ConfigSettings.
+    function _validateConfig(ConfigSettings memory config) internal pure {
+        // Make sure timelockLiveness is not too long, otherwise contract can might not be able to fix itself
+        // before a vulnerability drains its collateral. e.g. 604800 = 7 days.
```
mrice32, line 149: Oh! Is there a reason we don't want to set a lower bound on the timelock? Seems like it should be > a day or so...? Compound has [two days](https://etherscan.io/address/0x6d903f6003cca6255d85cca4d3b5e5146dc33925#code) as their absolute min, but I feel like 1 day is kinda a natural value...?

Thoughts?

nicholaspai: Yeah that's reasonable! done

mrice32: Did you forget to push?

nicholaspai: just pushed

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -0,0 +1,165 @@
...
+     ****************************************/
+
+    event ProposedNewConfigSettings(
+        address indexed proposer,
+        uint256 rewardRate,
+        uint256 proposerBond,
+        uint256 timelockLiveness,
+        uint256 proposalPassedTimestamp
+    );
+    event ChangedNewConfigSettings(uint256 rewardRate, uint256 proposerBond, uint256 timelockLiveness);
```
mrice32, line 55:
```suggestion
    event ChangedConfigSettings(uint256 rewardRate, uint256 proposerBond, uint256 timelockLiveness);
```

mrice32 (approved):
LGTM! Side note: this contract is super clean. very nice

PR #2224, opened 2020-11-25
ci: fix double compilation target issue

mrice32:
**Motivation**

ci is failing due to multiple solidity versions being compiled twice.

**Summary**

Truffle is finding absolute paths in our previously published artifacts and looking for them in ci, which causes it to find a second set of contracts that have a different (newer) version. To keep it from doing this, we basically just move the protocol dir in the test job to disrupt these absolute paths.

**Issue(s)**
N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -111,7 +111,12 @@ jobs:
           key: protocol-completed-build-{{ .Environment.CIRCLE_SHA1 }}
       - run:
           name: Run tests
-          command: yarn run test
+          command: |
+            cd ..
+            mkdir truffle_workaround
+            mv protocol truffle_workaround/
+            cd truffle_workaround/protocol
+            yarn run test
   coverage:
     docker:
       - image: circleci/node:12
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -111,7 +111,12 @@ jobs:
           key: protocol-completed-build-{{ .Environment.CIRCLE_SHA1 }}
       - run:
           name: Run tests
-          command: yarn run test
+          command: |
+            cd ..
+            mkdir truffle_workaround
+            mv protocol truffle_workaround/
+            cd truffle_workaround/protocol
+            yarn run test
```
nicholaspai, line 119: Should we `rm -rf truffle_workaround` somewhere in ci or does circleci handle cleanup

mrice32: circleci handles the cleanup as far as I'm aware. We don't cache the results of this job for any other job to use, so I think we should be fine just moving stuff around without cleaning up.

mrice32:
@nicholaspai added a comment about why this is necessary and the implications. PTAL.

nicholaspai (approved):
LGTM thanks for the comment it will help spare us confusion down the line

PR #2229, opened 2020-11-27
feat(DVM) Add arbitrary data to price requests

chrismaree:
**Motivation**

issue: https://github.com/UMAprotocol/protocol/issues/2234

This PR adds the ability for the DVM to accept any arbitrary data as part of a price request. 

**Summary**

**Details**

This is done by accepting an `bytes ancillaryData` into the dvm price requests.

**Issue(s)**

Fixes 2234


packages/common/src/EncryptionHelper.js
```diff
@@ -18,6 +18,7 @@ function computeVoteHash(request) {
     { t: "int", v: request.salt },
     { t: "address", v: request.account },
     { t: "uint", v: request.time },
+    { t: "bytes", v: "0x" },
```
chrismaree, line 21: This change makes all downstream usages of `EncryptionHelper` work be default; with nothing other than this blank ancillary data.

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -22,7 +24,14 @@ import "@openzeppelin/contracts/cryptography/ECDSA.sol";
...
  */
-contract Voting is Testable, Ownable, OracleInterface, VotingInterface {
+contract Voting is
+    Testable,
+    Ownable,
+    OracleInterface,
+    OracleAncillaryInterface,
+    VotingInterface,
+    VotingAncillaryInterface
+{
```
chrismaree, line 34: the `Voting` contract now considers the two new interfaces that define the Ancillary voting data.

packages/cli/src/admin.js
```diff
@@ -27,7 +27,8 @@ async function decodeGovernorProposal(artifacts, id) {
 
 async function decodeAllActiveGovernorProposals(artifacts, web3) {
   const Voting = artifacts.require("Voting");
-  const voting = await Voting.deployed();
```
chrismaree, line 30: This is basically the ripple change throughout the repo to force Truffle to use the overloaded methods (without the ancillary data prop). 

To remove this we need to make all downstream tests and infrastructure consider _only_ the ancillary methods and then we can desperate the overloaded method. This will enable us to roll back these test changes.

packages/core/contracts/oracle/implementation/VoteTiming.sol
```diff
@@ -32,7 +32,7 @@ library VoteTiming {
      * @return roundId defined as a function of the currentTime and `phaseLength` from `data`.
      */
     function computeCurrentRoundId(Data storage data, uint256 currentTime) internal view returns (uint256) {
-        uint256 roundLength = data.phaseLength.mul(uint256(VotingInterface.Phase.NUM_PHASES_PLACEHOLDER));
```
chrismaree, line 35: Unfortunately, some state had to move to the VotingAncillaryInterface to get the inheritance tree to compile correctly with truffle.

packages/core/contracts/oracle/interfaces/VotingInterface.sol
```diff
@@ -62,7 +59,23 @@ abstract contract VotingInterface {
      * commitments that can fit in one transaction.
      * @param commits array of structs that encapsulate an `identifier`, `time`, `hash` and optional `encryptedVote`.
      */
-    function batchCommit(Commitment[] calldata commits) external virtual;
+    function batchCommit(Commitment[] memory commits) public virtual;
```
chrismaree, line 62: this needed to become public to enable calling internally for the override method.

packages/core/scripts/umip-15/2_VoteSimulate.js
```diff
@@ -34,7 +35,7 @@ async function runExport() {
    ***********************************/
 
   console.log("0. SETUP PHASE");
-  const voting = await Voting.deployed();
```
chrismaree, line 37: this was not added to the umip upgrade scripts clearly but is added to the vote related ones.

packages/common/src/EncryptionHelper.js
chrismaree: For now, we will simply not support ancillary data in the voter dapp.

packages/common/src/EncryptionHelper.js
```diff
@@ -18,6 +18,19 @@ function computeVoteHash(request) {
...
+    { t: "uint", v: request.roundId },
+    { t: "bytes32", v: request.identifier }
+  );
+}
+
+function computeVoteHashAncillary(request) {
```
chrismaree, line 27: however, I did add a method and tests to verify the behavior works as expected.

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -444,38 +550,73 @@ contract Voting is Testable, Ownable, OracleInterface, VotingInterface {
...
+
     /**
      * @notice Reveal multiple votes in a single transaction.
      * Look at `project-root/common/Constants.js` for the tested maximum number of reveals.
      * that can fit in one transaction.
      * @dev For more information on reveals, review the comment for `revealVote`.
      * @param reveals array of the Reveal struct which contains an identifier, time, price and salt.
      */
-    function batchReveal(Reveal[] calldata reveals) external override {
+    function batchReveal(RevealAncillary[] memory reveals) public override {
```
mrice32, line 590: Wow, I forgot about all of these methods.

packages/common/src/EncryptionHelper.js
mrice32: +1

packages/common/src/EncryptionHelper.js
mrice32: SGTM

packages/common/src/EncryptionHelper.js
```diff
@@ -18,6 +18,19 @@ function computeVoteHash(request) {
     { t: "int", v: request.salt },
     { t: "address", v: request.account },
     { t: "uint", v: request.time },
+    { t: "bytes", v: "0x" },
```
mrice32, line 21: Do the following hashes match?

```
sha3(uint(5), uint(6));
sha3(uint(5), byes(0), uint(6));
```

If _not_, we need to be careful about releasing the common package as it will likely break compatibility with the current voting contract, right?

mrice32: Update: they _do_ seem to match (although someone else should verify :)). Very cool. In that case, is the "0x" element here basically just documentation to show that we're not leaving the value out?

mrice32 (approved):
@chrismaree, this is awesome! Big +1. Contracts/prod code all look great! I left a few pretty minor comments -- more as just mental notes for us to think about down the road.

Will move on to the tests, etc.

packages/core/test/oracle/Voting.js
```diff
@@ -1793,4 +1805,111 @@ contract("Voting", function(accounts) {
...
+    assert(await didContractThrow(voting.getPrice(identifier2, time2, ancillaryData2, { from: registeredContract })));
+
+    // Move to the voting round.
+    await moveToNextRound(voting);
+    const roundId = (await voting.getCurrentRoundId()).toString();
+
+    // Ancillary data should be correctly preserved and accessible to voters.
+    const priceRequests = await voting.getPendingRequests();
+
+    assert.equal(priceRequests.length, 2);
```
chrismaree, line 1841: -

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -45,6 +54,7 @@ contract Voting is Testable, Ownable, OracleInterface, VotingInterface {
         // The index in the `pendingPriceRequests` that references this PriceRequest. A value of UINT_MAX means that
         // this PriceRequest is resolved and has been cleaned up from `pendingPriceRequests`.
         uint256 index;
+        bytes ancillaryData;
```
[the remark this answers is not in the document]
chrismaree: I agree. ideally, emit the `ancillaryData` and just refer to a hash of it on-chain. This is a complexity trade-off that I think we leave for later though.

packages/common/src/EncryptionHelper.js
chrismaree: yup, it shows we are leaving it out. it also mimics _exactly_ how the hash is generated on-chain in `revealVote(identifier, time, price, "", salt);` with the blank string for the bytes type.

mrice32: It's really really nice that a 0 length data is perfectly backwards compatible when doing tight packing 👌

packages/core/contracts/oracle/implementation/Voting.sol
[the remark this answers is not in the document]
mrice32: Agreed.

packages/core/test/oracle/Voting.js
```diff
@@ -1793,4 +1808,302 @@ contract("Voting", function(accounts) {
     // After retrieval, the length should be decreased back to 0 since the element added in this test is now deleted.
     assert.equal((await votingTest.getPendingPriceRequestsArray()).length, 0);
   });
+  it("Votes can correctly handel arbitrary ancillary data", async function() {
```
mrice32, line 1811:
```suggestion
  it("Votes can correctly handle arbitrary ancillary data", async function() {
```

packages/core/test/oracle/Voting.js
```diff
@@ -1793,4 +1808,302 @@ contract("Voting", function(accounts) {
...
+    await supportedIdentifiers.addSupportedIdentifier(identifier2);
+
+    // Instantiate a voting interface that supports ancillary data.
+    voting = await Voting.at(voting.address);
+
+    // Store the number of price requests to verify the right number are enqued.
+    const priceRequestLengthBefore = (await voting.getPendingRequests()).length;
+
+    // Requests should not be added to the current voting round.
+    await voting.methods["requestPrice(bytes32,uint256,bytes)"](identifier1, time1, ancillaryData1, {
```
mrice32, line 1956: nit: I thought we weren't using this syntax bc it didn't work?

packages/core/test/oracle/Voting.js
```diff
@@ -1793,4 +1808,302 @@ contract("Voting", function(accounts) {
...
     assert.equal((await votingTest.getPendingPriceRequestsArray()).length, 0);
   });
+  it("Votes can correctly handel arbitrary ancillary data", async function() {
+    const identifier1 = web3.utils.utf8ToHex("request-retrieval1");
+    const time1 = "1000";
+    const ancillaryData1 = web3.utils.utf8ToHex("some-random-extra-data"); // ancillary data should be able to store any extra dat
+
+    const identifier2 = web3.utils.utf8ToHex("request-retrieval2");
+    const time2 = "2000";
+    const ancillaryData2 = web3.utils.utf8ToHex(`callerAddress:${account4}`); // ancillary data should be able to store addresses
```
mrice32, line 1818: nit: I think it would make this test slightly better if we made the identifier and timestamp match between both requests and only have the data differ. Then, this test would verify that the DVM correctly differentiates.

Thoughts? I think it would only require us to set identifier2 and time2 to identifier1 and time1 on these lines.

mrice32 (approved):
Tests look awesome. A few very minor comments/nits

packages/core/test/oracle/Voting.js
chrismaree: good call. Have updated the tests accordingly and added a comment to explain why.

mrice32: +1 -- thanks @chrismaree!

PR #2242, opened 2020-11-30
feat(emp): replace liquidation and settlement logic from DVM to optimistic oracle

chrismaree:
**Motivation**

This PR points the EMP at the OptimisticOracle for settlement rather than the DVM directly. This will allow settlement to be much faster (potentially 2 hours).

**Summary**

Some notes:
- The payout from the contract's perspective is the same, as it now offers a reward for resolving the request that is equal to the final fee.
- This adds an interface for the OptimisticOracle so other contracts don't need to import the entire contract.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
+++ b/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
@@ -794,13 +794,20 @@ contract PricelessPositionManager is FeePayer {
     // Requests a price for transformed `priceIdentifier` at `requestedTime` from the Oracle.
     function _requestOraclePriceExpiration(uint256 requestedTime) internal {
         OptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
+
+        // Increase token allowance to enable the optimistic oracle reward transfer.
+        FixedPoint.Unsigned memory reward = _computeFinalFees();
+        collateralCurrency.safeIncreaseAllowance(address(optimisticOracle), reward.rawValue);
         optimisticOracle.requestPrice(
             _transformPriceIdentifier(requestedTime),
             requestedTime,
             _getAncillaryData(),
             collateralCurrency,
-            0
+            reward.rawValue // Reward is equal to the final fee
         );
+
+        // Apply haircut to all sponsors by decrementing the cumlativeFeeMultiplier by the amount lost from the final fee.
+        _adjustCumulativeFeeMultiplier(reward, _pfc());
     }
 
     // Fetches a resolved Oracle price from the Oracle. Reverts if the Oracle hasn't resolved for this request.
--- a/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
+++ b/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
@@ -11,7 +11,7 @@ import "../../common/implementation/FixedPoint.sol";
 import "../../common/interfaces/ExpandedIERC20.sol";
 import "../../common/interfaces/IERC20Standard.sol";
 
-import "../../oracle/interfaces/OracleInterface.sol";
+import "../../oracle/interfaces/OptimisticOracleInterface.sol";
 import "../../oracle/interfaces/IdentifierWhitelistInterface.sol";
 import "../../oracle/implementation/Constants.sol";
 
@@ -458,7 +458,7 @@ contract PricelessPositionManager is FeePayer {
 
         // Transfer tokens into the contract from caller and mint corresponding synthetic tokens to the caller's address.
         collateralCurrency.safeTransferFrom(msg.sender, address(this), collateralAmount.rawValue);
-        require(tokenCurrency.mint(msg.sender, numTokens.rawValue), "Minting synthetic tokens failed");
+        require(tokenCurrency.mint(msg.sender, numTokens.rawValue));
     }
 
     /**
@@ -602,7 +602,7 @@ contract PricelessPositionManager is FeePayer {
      * `emergencyShutdown` as the governor who would call the function would also receive the fees.
      */
     function emergencyShutdown() external override onlyPreExpiration() onlyOpenState() nonReentrant() {
-        require(msg.sender == _getFinancialContractsAdminAddress(), "Caller not Governor");
+        require(msg.sender == _getFinancialContractsAdminAddress());
 
         contractState = ContractState.ExpiredPriceRequested;
         // Expiratory time now becomes the current time (emergency shutdown time).
@@ -705,6 +705,12 @@ contract PricelessPositionManager is FeePayer {
         return _transformPriceIdentifier(requestTime);
     }
 
+    function _getAncillaryData() public view returns (bytes memory) {
+        // Note: when ancillary data is passed to the optimistic oracle, it should be tagged with the token address
+        // whose funding rate it's trying to get.
+        return abi.encodePacked(address(tokenCurrency));
+    }
+
     /****************************************
      *          INTERNAL FUNCTIONS          *
      ****************************************/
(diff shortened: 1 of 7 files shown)


packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -790,16 +796,31 @@ contract PricelessPositionManager is FeePayer {
...
+            requestedTime,
+            _getAncillaryData(),
+            collateralCurrency,
+            0
+        );
     }
 
     // Fetches a resolved Oracle price from the Oracle. Reverts if the Oracle hasn't resolved for this request.
-    function _getOraclePrice(uint256 requestedTime) internal view returns (FixedPoint.Unsigned memory) {
+    function _getOraclePrice(uint256 requestedTime) internal returns (FixedPoint.Unsigned memory) {
```
chrismaree, line 810: this `view` modifier needed to be removed as the optimistic oracle's `hasPrice` is state modifying.

packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -90,33 +97,70 @@ contract("PricelessPositionManager", function(accounts) {
     assert.notEqual(beneficiaryCollateralBalance.toString(), "0");
   };
 
-  before(async function() {
+  const proposeAndSettleOptimisticOraclePrice = async (priceFeedIdentifier, requestTime, price) => {
```
chrismaree, line 100: This is the basic method that I used to replace most instances of `await mockOracle.pushPrice`.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -313,7 +316,8 @@ contract OptimisticOracle is Testable, Lockable {
         emit ProposePrice(requester, proposer, identifier, timestamp, ancillaryData, proposedPrice);
 
         // Callback.
-        try OptimisticRequester(requester).priceProposed(identifier, timestamp, ancillaryData) {} catch {}
+        if (address(requester).isContract())
```
mrice32, line 319: nit: do you need to add this to the settlement callback too?

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -178,8 +179,8 @@ contract PricelessPositionManager is FeePayer {
         address _excessTokenBeneficiary,
         address _financialProductLibraryAddress
     ) public FeePayer(_collateralAddress, _finderAddress, _timerAddress) nonReentrant() {
-        require(_expirationTimestamp > getCurrentTime(), "Invalid expiration in future");
```
chrismaree, line 181: removed to save bytecode. these checks in the constructor are not useful.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -256,7 +255,7 @@ contract PricelessPositionManager is FeePayer {
      */
     function cancelTransferPosition() external onlyPreExpiration() nonReentrant() {
         PositionData storage positionData = _getPositionData(msg.sender);
-        require(positionData.transferPositionRequestPassTimestamp != 0, "No pending transfer");
```
chrismaree, line 259: equally, we don't really ever have people transferring positions.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -529,7 +528,7 @@ contract PricelessPositionManager is FeePayer {
 
         // Get the current settlement price and store it. If it is not resolved will revert.
         if (contractState != ContractState.ExpiredPriceReceived) {
-            expiryPrice = _getOraclePrice(expirationTimestamp);
```
chrismaree, line 532: to accommodate the two kinds of oracle requests we can now request an `Expiration` oracle request or a `Liquidation` oracle request.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -784,18 +782,64 @@ contract PricelessPositionManager is FeePayer {
...
+
+        // Increase token allowance to enable the optimistic oracle reward transfer.
+        FixedPoint.Unsigned memory reward = _computeFinalFees();
+        collateralCurrency.safeIncreaseAllowance(address(optimisticOracle), reward.rawValue);
+        optimisticOracle.requestPrice(
+            _transformPriceIdentifier(requestedTime),
+            requestedTime,
+            _getAncillaryData(),
+            collateralCurrency,
+            reward.rawValue // Reward is equal to the final fee
```
nicholaspai, line 805: Why do you set reward equal to final fee? Is this just an arbitrary default choice or is there a reason behind this.

packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
```diff
@@ -0,0 +1,245 @@
...
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @title Financial contract facing Oracle interface.
+ * @dev Interface used by financial contracts to interact with the Oracle. Voters will use a different interface.
+ */
+abstract contract OptimisticOracleInterface {
```
nicholaspai, line 11: Should the OptimisticOracle inherit this interface?

packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -2111,8 +2203,25 @@ contract("PricelessPositionManager", function(accounts) {
     assert.equal(collateralPaid, toWei("60"));
 
     // Push a different price to the new oracle to ensure the contract still uses the old price.
```
nicholaspai, line 2205: Is this test done also for liquidation+dispute price requests? i.e the Dispute resolves to a certain price, and then OO is switched out in the finder, but the Perp should still settle the dispute and pay out rewards using the originally resolved dispute price.

nicholaspai (approved):
Looks great, have two main comments
- Should OO inherit OOInterface
- Should the PerpLiquidatable test resolved dispute settlements when the OO is swapped out in the finder post-dispute but pre-dispute-settlement

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
mrice32: I think this is just an arbitrary default choice -- it also matches what the contract was paying out previously to the DVM.

packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
mrice32: Probably! Will update.

packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
mrice32: Good point. Will make sure this is tested.

mrice32: Wait, sorry, I don't think we _can_ test this. Because the settlement for liquidation and disputes happens at the same time as the price retrieval. So you pull the price and immediately send the payouts. There is no in-between time for you to swap out the oracle. Does that make sense?

packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
mrice32: Done.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
nicholaspai: Cool SGTM. As long as no changes to economics I agree

packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
nicholaspai: Yup, I thought that might be the case. Nice

PR #2244, opened 2020-11-30
feat(perp): allow deployer to scale token value

### LINKED ISSUE #2243: Allow deployers to scale the perp token value
**Why**

Deployers probably don't want the initial token value to match the index.

**How**

Likely by scaling the initial FRM by the input.

**Resolution**

Deployers can pass in an initial scaling.

**Difficulty Score [1-10]**

- 4: ~2 hours

mrice32:
**Motivation**

To aid UX, cTokens should generally not start at the value of the index because they will likely deviate from the index substantially over time. To allow the deployer flexibility around this initial value, the deployer can effectively scale the token value at deployment time.

Note: this also removes the unused PerpetualInterface.sol.

**Summary**

The scaling is easily implemented by just changing the initial value of the funding rate multiplier.

**Issue(s)**

Fixes #2243


packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualInterface.sol
```diff
@@ -1,16 +0,0 @@
-pragma solidity ^0.6.0;
-pragma experimental ABIEncoderV2;
-
-import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "../../common/implementation/FixedPoint.sol";
-
-/**
- * @notice External methods that the FundingRateStore needs access to.
- */
-interface PerpetualInterface {
```
nicholaspai, line 10: +1 removing this as it was only used by FRS

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -106,14 +106,23 @@ abstract contract FundingRateApplier is FeePayer {
...
     ) public FeePayer(_collateralAddress, _finderAddress, _timerAddress) {
         uint256 currentTime = getCurrentTime();
         fundingRate.updateTime = currentTime;
         fundingRate.applicationTime = currentTime;
 
-        // Seed the cumulative multiplier as 1, from which it will be scaled as funding rates are applied over time.
-        fundingRate.cumulativeMultiplier = FixedPoint.fromUnscaledUint(1);
+        // Seed the cumulative multiplier with the token scaling, from which it will be scaled as funding rates are
+        // applied over time.
+        FixedPoint.Unsigned memory minScaling = FixedPoint.Unsigned(1e8); // 1e-10
```
nicholaspai, line 118: OOC why validate this within the perp as opposed to in the PerpCreator?

nicholaspai (approved):
LGTM, but have questions about where you choose to validate the `_tokenScaling` param

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: No super strong reason. We could move it out for the same reasons we moved out the validation of other parameters, like the synthetic token. Thoughts?

nicholaspai: Bytecode definitely is a reason to move it out, but I was thinking more for consistency. The creator does some validation on Perp constructor params while it seems that our stance on the Perp itself is not to validate any params. Thoughts?

mrice32: Agreed, consistency is another good reason! We'll also be able to add back the require message!

mrice32: Will update

mrice32: Done.

nicholaspai (approved):

PR #2251, opened 2020-12-01
improve(umip-dry-run-script): Fix ABI's in mainnet fork script

nicholaspai:

**Motivation**

Mainnet fork tests are designed to test behavior of already-deployed contracts after potential future transactions. Therefore, these test scripts will break if the ABI's in `artifacts` diverge from the deployed contract ABI's, which is exactly what is happening now with the `Voting` contracts. This problem will exist until an updated Voting contract is deployed.

**Details**

~Added comments to scripts to warn users about running scripts with new ABI's. An alternative solution to this problem is to create a "legacy" contract interface that can be used with deployed ABI's.~

Use `getAbi` and `getAddress` to fix contracts to match deployed ABI's.


mrice32:
> Signed-off-by: Nick Pai [npai.nyc@gmail.com](mailto:npai.nyc@gmail.com)
> 
> **Motivation**
> 
> Mainnet fork tests are designed to test behavior of already-deployed contracts after potential future transactions. Therefore, these test scripts will break if the ABI's in `artifacts` diverge from the deployed contract ABI's, which is exactly what is happening now with the `Voting` contracts. This problem will exist until an updated Voting contract is deployed.
> 
> **Details**
> 
> Added comments to scripts to warn users about running scripts with new ABI's. An alternative solution to this problem is to create a "legacy" contract interface that can be used with deployed ABI's.

We should be able to do this today using [this function](https://github.com/UMAprotocol/protocol/blob/1631ef7ad29aaeba756ef3b9a01c667e1343df85/packages/core/index.js#L70), I think. Will try to write up a PR to do this.

nicholaspai:
> > Signed-off-by: Nick Pai [npai.nyc@gmail.com](mailto:npai.nyc@gmail.com)
> > **Motivation**
> > Mainnet fork tests are designed to test behavior of already-deployed contracts after potential future transactions. Therefore, these test scripts will break if the ABI's in `artifacts` diverge from the deployed contract ABI's, which is exactly what is happening now with the `Voting` contracts. This problem will exist until an updated Voting contract is deployed.
> > **Details**
> > Added comments to scripts to warn users about running scripts with new ABI's. An alternative solution to this problem is to create a "legacy" contract interface that can be used with deployed ABI's.
> 
> We should be able to do this today using [this function](https://github.com/UMAprotocol/protocol/blob/1631ef7ad29aaeba756ef3b9a01c667e1343df85/packages/core/index.js#L70), I think. Will try to write up a PR to do this.

Will move this to draft status until you do!

nicholaspai:
> > > Signed-off-by: Nick Pai [npai.nyc@gmail.com](mailto:npai.nyc@gmail.com)
> > > **Motivation**
> > > Mainnet fork tests are designed to test behavior of already-deployed contracts after potential future transactions. Therefore, these test scripts will break if the ABI's in `artifacts` diverge from the deployed contract ABI's, which is exactly what is happening now with the `Voting` contracts. This problem will exist until an updated Voting contract is deployed.
> > > **Details**
> > > Added comments to scripts to warn users about running scripts with new ABI's. An alternative solution to this problem is to create a "legacy" contract interface that can be used with deployed ABI's.
> > 
> > 
> > We should be able to do this today using [this function](https://github.com/UMAprotocol/protocol/blob/1631ef7ad29aaeba756ef3b9a01c667e1343df85/packages/core/index.js#L70), I think. Will try to write up a PR to do this.
> 
> Will move this to draft status until you do!

Actually I made the change, pushing now

packages/core/scripts/identifier-umip/1_Propose.js
```diff
@@ -28,21 +35,26 @@ async function runExport() {
...
     return {
-      to: identifierWhitelist.address,
+      to: identifierWhitelist.options.address,
       value: 0,
       data: addIdentifierTx
     };
   });
 
+  // TODO: Calling this via the web3.eth.Contract method fails, possibly because `transactions` array is in an
+  // unparseable format for governor.methods.propose() to read.
```
nicholaspai, line 57: @mrice32 any idea how to fix this? Transactions is in the shape:

```
[{ to: ..., value: ..., data: ... }]
```

packages/core/scripts/umip-3/2_VoteSimulate.js
```diff
@@ -19,9 +19,10 @@ const argv = require("minimist")(process.argv.slice(), { boolean: ["revert"] });
 // Address which holds a lot of UMA tokens to mock a majority vote
 const foundationWallet = "0x7a3A1c2De64f20EB5e916F40D11B01C441b2A8Dc";
 
-const Voting = artifacts.require("Voting");
-const VotingInterfaceTesting = artifacts.require("VotingInterfaceTesting");
```
nicholaspai, line 23: @chrismaree is this OK to remove since we fix Voting to a certain ABI now? Is there any reason to use the ancillary data in this script?

chrismaree: nope, that seems like a reasonable change.

chrismaree (approved):
this looks good! you updated the UMIP-3 script, as well as the proposal scripts. UMIP-15 & UMIP-8 should probs be updated as well if we are changing it in this way.

nicholaspai:
> this looks good! you updated the UMIP-3 script, as well as the proposal scripts. UMIP-15 & UMIP-8 should probs be updated as well if we are changing it in this way.

You ok if we do this in a subsequent PR? Once this PR is merged we can go through and change them.

mrice32:
> > this looks good! you updated the UMIP-3 script, as well as the proposal scripts. UMIP-15 & UMIP-8 should probs be updated as well if we are changing it in this way.
> 
> 
> 
> You ok if we do this in a subsequent PR? Once this PR is merged we can go through and change them.

Seems reasonable to me.

mrice32 (approved):
LGTM! Long term, we might consider moving all these scripts to a different package to make controlling abi versions easier.

PR #2253, opened 2020-12-01
improve(core): add 1.2.0 release to previous core versions

mrice32:
**Motivation**

Make the 1.2.0 abis available to those who want to access previous versions.

**Summary**

It just requires us to add the alias to the package.json.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -5,7 +5,8 @@
   "dependencies": {
     "@truffle/contract": "^4.2.20",
     "@uma/common": "^1.1.0",
-    "@uma/core-1-1-0": "npm:@uma/core@1.1.0"
+    "@uma/core-1-1-0": "npm:@uma/core@1.1.0",
+    "@uma/core-1-2-0": "npm:@uma/core@1.2.0"
   },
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -5,7 +5,8 @@
   "dependencies": {
     "@truffle/contract": "^4.2.20",
     "@uma/common": "^1.1.0",
-    "@uma/core-1-1-0": "npm:@uma/core@1.1.0"
+    "@uma/core-1-1-0": "npm:@uma/core@1.1.0",
+    "@uma/core-1-2-0": "npm:@uma/core@1.2.0"
```
nicholaspai, line 9: Is it easy to tell which commit hash differentiates 1.1 from 1.2? and 1.2 from latest?

mrice32: You can look at the repo tags to tell -- that's probably the only way to tell right now.

mrice32: https://github.com/UMAprotocol/protocol/releases

nicholaspai (approved):

PR #2256, opened 2020-12-02
bug(emp-creator): add financialProductLib address to deployment params

### LINKED ISSUE #2233: Add financial product library address parameter to ExpiringMultiPartyCreator

nicholaspai:

**Motivation**

#2233 

**Details**

Allows deployer to specify a FPLib when deploying a new EMP. Previous behavior would default all EMP's to not using a FPLib, which isn't a major bug but doesn't take advantage of the FPLib contract param.

**Issue(s)**

Fixes #2233

### DIFF
--- a/packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
+++ b/packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
@@ -15,6 +15,7 @@ const ExpiringMultiParty = artifacts.require("ExpiringMultiParty");
 const IdentifierWhitelist = artifacts.require("IdentifierWhitelist");
 const AddressWhitelist = artifacts.require("AddressWhitelist");
 const Store = artifacts.require("Store");
+const StructuredNoteFinancialProductLibrary = artifacts.require("StructuredNoteFinancialProductLibrary");
 
 contract("ExpiringMultiPartyCreator", function(accounts) {
   let contractCreator = accounts[0];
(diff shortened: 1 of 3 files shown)


packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
```diff
@@ -293,4 +294,27 @@ contract("ExpiringMultiPartyCreator", function(accounts) {
...
+      return ev.expiringMultiPartyAddress != 0 && ev.deployerAddress == contractCreator;
+    });
+    let expiringMultiParty = await ExpiringMultiParty.at(expiringMultiPartyAddress);
+    let linkedFPLib = await StructuredNoteFinancialProductLibrary.at(
+      await expiringMultiParty.financialProductLibrary()
+    );
+
+    // FPLib address is saved correctly.
+    assert.equal(linkedFPLib.address, structuredNoteFPL.address);
+  });
```
nicholaspai, line 319: @UMAprotocol/eng Should any other tests be done here to test that the EMP is "correct linked" to the FPLib?

chrismaree (approved):
Looks good! Thanks for fixing this

packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
chrismaree: Can try calling a transformation on a deployed emp that uses the transformation such as one of the public view functions? Could also check that the deployed emp has the library set correctly to the address you created

chrismaree: You've done the second test actually already

daywiss (approved):

packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
mrice32: IMO, the unit tests for the emp should cover whether the address returned by the public financialProductLibrary() method match the transformations done in the contract, and that sort of test shouldn't be necessary for the scope of this contract. The creator tests should just be sure that the address is injected properly.

But more tests wouldn't necessarily hurt :).

mrice32 (approved):
LGTM! Thanks for adding this!

nicholaspai:
@chrismaree @mrice32 
> IMO, the unit tests for the emp should cover whether the address returned by the public financialProductLibrary() method match the transformations done in the contract, and that sort of test shouldn't be necessary for the scope of this contract. The creator tests should just be sure that the address is injected properly.

I  agree with this, because the `financial-product-library` tests should validate that if the FPLib is properly loaded, the transformations work. We don't want this creator test to be opinionated about what a correct "transformation" looks like.

packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js
nicholaspai: In response to first suggestion see [here](https://github.com/UMAprotocol/protocol/pull/2256#issuecomment-737513983)!

PR #2257, opened 2020-12-02
feat(affiliates): add coingecko synthetic price fallback and calculator

### LINKED ISSUE #2254: Update Dev mining to accommodate non-standard price feed inputs (like uGAS)
**Why**

To accommodate new entries to the developer mining program we have to expand how we calculate total value for an emp. Currently we use a combination of cryptowatch for synthetic/collateral price and coingecko for collateral/usd price. For uGas in particular this wont work since its not available on cryptowatch, but the synthetic is available on coingecko. We want to add a fallback to coingecko for when synthetic price cant be pulled from cryptowatch. 

**How**

Needs to be some thought around how to switch between price sources to calculate value, but most changes should be confined to the affiliates.js file. 

**Resolution**

See that we can add the new ugas emp to developer mining.

**Difficulty Score [1-10]**
- 6: ~1 day

daywiss:

**Motivation**

#2254 

**Summary**

This adds a fallback function for getting synthetic prices. Changes the api internally, each synthetic price now gets a "calcValue" function with it, to convert it to usd.  If synthetic price feed fails it will fallback to coingecko feed. Tested with uGas and did a basic sanity check vs renbtc dec contract.

**Issue(s)**

Fixes #2254

### DIFF
--- a/packages/affiliates/libs/affiliates.js
+++ b/packages/affiliates/libs/affiliates.js
@@ -51,9 +51,21 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
     );
   }
   // Gets the value of collateral currency in USD.
-  async function getCollateralPriceHistory(address, currency = "usd", start, end) {
+  async function getCoingeckoPriceHistory(address, currency = "usd", start, end) {
     const result = await coingecko.getHistoricContractPrices(address, currency, start, end);
     return Prices(result);
+  }
+
+  // the fallback requires a different value calculation, so this is probably the cleanest way to "switch"
+  // between value by and value by usd. The function to calculate value is passed along with the data.
+  async function getSyntheticPriceHistoryWithFallback(address,start,end, syntheticAddress){
+    try{
+      // this will throw if the feed isnt available (among other errors probably)
+      return [await getSyntheticPriceHistory(address,start,end), calculateValue]
+    }catch(err){
+      // this will need to do a lookup based on synthetic token address not emp
+      return [await getCoingeckoPriceHistory(syntheticAddress,'usd',start,end), calculateValueFromUsd]
+    }
   }
 
   // Gets the value of synthetics in collateral currency.
@@ -61,6 +73,7 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
     const result = await synthPrices.getHistoricSynthPrices(address, start, end);
     return Prices(result);
   }
+
   async function getBlocks(start, end) {
     const blocks = await queries.getBlocks(start, end, ["timestamp", "number"]);
     return blocks.map(block => {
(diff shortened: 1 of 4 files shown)


packages/affiliates/libs/affiliates.js
```diff
@@ -51,16 +51,29 @@ const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrice
...
   }
 
+  // the fallback requires a different value calculation, so this is probably the cleanest way to "switch"
+  // between value by and value by usd. The function to calculate value is passed along with the data.
+  async function getSyntheticPriceHistoryWithFallback(address,start,end, syntheticAddress){
+    try{
+      // this will throw if the feed isnt available (among other errors probably)
+      return [await getSyntheticPriceHistory(address,start,end), calculateValue]
+    }catch(err){
+      // this will need to do a lookup based on synthetic token address not emp
```
nicholaspai, line 66: Is coingecko less reliable than what `getSyntheticPriceHistory` returns? Just want to check that the fallback is a worst option than the default

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -51,10 +55,10 @@ describe("DeployerRewards", function() {
       assert.ok(history.history.length());
     });
   });
-  it("getCollateralPriceHistory", async function() {
+  it("getCoingeckoPriceHistory", async function() {
     this.timeout(10000);
     const [, address] = collateralTokens;
-    const result = await affiliates.utils.getCollateralPriceHistory(address, "usd", startingTimestamp, endingTimestamp);
+    const result = await affiliates.utils.getCoingeckoPriceHistory(address, "usd", startingTimestamp, endingTimestamp);
```
nicholaspai, line 61: Is this external API call potentially problematic to put into our CI?

nicholaspai (approved):
Fallbacks are great in general! LGTM

packages/affiliates/libs/affiliates.js
daywiss: its really more to offer more options for getting the value of the synthetic. originally we use cryptowatch which returns the synthetic value compared to its collateral, but ugas does not have this feed. In this case it has a coingecko feed with returns synthetic value compared to usd. So now we can basically offer support for a wider variety of synthetic types.

packages/affiliates/test/mocha/affiliates.js
daywiss: I dont think this was problematic, though it does reach outside, so it technically could be. i think the problem was with the getSyntheticPrice which was causing an issue since it relied on web3 mainnet and abis.

packages/affiliates/libs/affiliates.js
chrismaree: yup. Just to add a bit more context the `getSyntheticPriceHistory` and cryptowatch feeds use the `financial-templates-lib`'s price feed which includes the `getHistoricalPricePeriods` which gives you all data over a large window. This method is not (yet) implemented in the uniswap feed. We _could_ add this method to the uniswap feed but this would not really solve the underlying problem that we cant support feeds that do not expose this `getHistoricalPricePeriods` method publically. Therefore, by adding the coingecko feed, we can support a much wider range; any product that is listed on coingecko.

chrismaree (approved):
This is awesome. Thanks for adding this @daywiss . Nice and simple and solves a problem that we _will_ have again with other synths in Dev mining.

PR #2260, opened 2020-12-03
improve(perpetual): Modify ConfigStore constraints and params

### LINKED ISSUE #2259: Perpetual should constrain funding rate range read from OO and not constrain proposal bond
**Why**

The PfC for a dishonest funding rate proposer is theoretically unbounded if the range of possible funding rates read by the Perpetual, from the OptimisticOracle, is unbounded. This is assuming that the funding rate proposer's [approximate attack vector is to dispute proposed funding rates for 74 hours](https://github.com/UMAprotocol/protocol/issues/2039#issuecomment-719734383) (2 hours of proposal liveness + 72 hours maximum DVM voting cycle), during which time there is a funding rate error between the current and the true rate.

Additionally, we [currently set an upper bound](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol#L154) for the funding rate proposer’s bond in the Perpetual, which is linked to a maximum magnitude of funding rate range 200%/year. The original motivation to limit this bond was to enhance proposer UX and trust that the `ConfigStore` would not change their costs unfairly. However, we already protect the proposer with a timelock that any `ConfigStore` updates must pass.

It is inconsistent and even dangerous to (1) limit the size of the proposer bond and (2) not limit the funding rate range. This limits the potential cost to dishonest proposers and sets no limit on their profit potential. The more conservative implementation is to (1) not limit the size of the proposer bond and (2) limit the funding rate range. Proposers would still be protected by the `ConfigStore`'s timelock, while the Perpetual's PfC would be bounded according to the funding rate range.

**How**

We can constrain the funding rate that the [`FundingRateApplier`](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/FundingRateApplier.sol#L224) queries from the `OptimisticOracle`.

A 600% funding rate range is equal to a 5% PfC: 600%/year / 360 days / 24 hours * 74 hours max attack time. 600% range is equivalent to a range of +300%/-300% for funding rates.

- 3: ~1 hour

nicholaspai:

**Motivation**

#2259 This enhances the Perpetual's economic security by allowing the Perpetual contract deployer to set funding rate bounds via timelock.

Changes:
- Adds config params to limit the range of possible funding rates to propose. This bounds a dishonest proposer's PfC
- Remove bounds from proposer bound, which can be changed to limit a proposer's PfC based on the funding rate.
- Remove future time limit from proposer timestamp, OptimisticOracles must be in the past starting with #2267

**Issue(s)**

Fixes #2259

### DIFF
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -167,7 +167,7 @@ abstract contract FundingRateApplier is FeePayer {
         // Set up optimistic oracle.
         bytes32 identifier = fundingRate.identifier;
         bytes memory ancillaryData = _getAncillaryData();
-        // Note: requestPrice will revert if `timestamp` is less than the current block timestamp
+        // Note: requestPrice will revert if `timestamp` is less than the current block timestamp.
         optimisticOracle.requestPrice(identifier, timestamp, ancillaryData, collateralCurrency, 0);
         totalBond = FixedPoint.Unsigned(
             optimisticOracle.setBond(
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -149,6 +149,7 @@ abstract contract FundingRateApplier is FeePayer {
         returns (FixedPoint.Unsigned memory totalBond)
     {
         require(fundingRate.proposalTime == 0, "Proposal in progress");
+        _validateFundingRate(rate);
 
         // Timestamp must be after the last funding rate update time, within the last 30 minutes, and it cannot be more
         // than 90 seconds ahead of the block timestamp.
@@ -257,6 +258,26 @@ abstract contract FundingRateApplier is FeePayer {
             }
         }
         return fundingRate.rate;
+    }
+
+    function _validateFundingRate(FixedPoint.Signed memory rate) internal pure {
+        // Constraining range of funding rates limits the PfC for any dishonest proposer and enhances the perpetual's
+        // security. A 600% funding rate range is equal to a 5% PfC for a proposer who can deter honest proposers
+        // for 74 hours: 600%/year / 360 days / 24 hours * 74 hours max attack time = ~ 5%.
+        // Note: 600% range is equivalent to a funding rate bounds of +300%/-300%. Imagine that the market is very
+        // volatile currently and that the "true" funding rate for the next 74 hours is -300%, but a dishonest
+        // proposer successfully proposes a rate of +300% (after a two hour liveness) and disputes honest
+        // proposers for the next 72 hours. This results in a funding rate error of 600% for 74 hours, until the DVM
+        // can set the funding rate back to its correct value.
+        FixedPoint.Signed memory MAX_FUNDING_RATE_PER_SECOND =
+            FixedPoint.fromUnscaledInt(300).div(360).div(24).div(60).div(60);
+        FixedPoint.Signed memory MIN_FUNDING_RATE_PER_SECOND =
+            FixedPoint.fromUnscaledInt(-300).div(360).div(24).div(60).div(60);
+
+        require(
+            rate.isLessThanOrEqual(MAX_FUNDING_RATE_PER_SECOND) &&
+                rate.isGreaterThanOrEqual(MIN_FUNDING_RATE_PER_SECOND)
+        );
     }
 
     // Fetches a funding rate from the Store, determines the period over which to compute an effective fee,
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -151,14 +151,11 @@ abstract contract FundingRateApplier is FeePayer {
         require(fundingRate.proposalTime == 0, "Proposal in progress");
         _validateFundingRate(rate);
 
-        // Timestamp must be after the last funding rate update time, within the last 30 minutes, and it cannot be more
-        // than 90 seconds ahead of the block timestamp.
+        // Timestamp must be after the last funding rate update time, within the last 30 minutes.
         uint256 currentTime = getCurrentTime();
         uint256 updateTime = fundingRate.updateTime;
         require(
-            timestamp > updateTime &&
-                timestamp >= currentTime.sub(_getConfig().proposalTimePastLimit) &&
-                timestamp <= currentTime.add(_getConfig().proposalTimeFutureLimit),
+            timestamp > updateTime && timestamp >= currentTime.sub(_getConfig().proposalTimePastLimit),
             "Invalid proposal time"
         );
 
@@ -170,6 +167,7 @@ abstract contract FundingRateApplier is FeePayer {
         // Set up optimistic oracle.
         bytes32 identifier = fundingRate.identifier;
         bytes memory ancillaryData = _getAncillaryData();
+        // Note: requestPrice will revert if `timestamp` is less than the current block timestamp
         optimisticOracle.requestPrice(identifier, timestamp, ancillaryData, collateralCurrency, 0);
         totalBond = FixedPoint.Unsigned(
             optimisticOracle.setBond(
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -262,17 +262,19 @@ abstract contract FundingRateApplier is FeePayer {
 
     function _validateFundingRate(FixedPoint.Signed memory rate) internal pure {
         // Constraining range of funding rates limits the PfC for any dishonest proposer and enhances the perpetual's
-        // security. A 600% funding rate range is equal to a 5% PfC for a proposer who can deter honest proposers
-        // for 74 hours: 600%/year / 360 days / 24 hours * 74 hours max attack time = ~ 5%.
-        // Note: 600% range is equivalent to a funding rate bounds of +300%/-300%. Imagine that the market is very
-        // volatile currently and that the "true" funding rate for the next 74 hours is -300%, but a dishonest
-        // proposer successfully proposes a rate of +300% (after a two hour liveness) and disputes honest
-        // proposers for the next 72 hours. This results in a funding rate error of 600% for 74 hours, until the DVM
+        // security. A 1000% funding rate range is equal to a 8.6% PfC for a proposer who can deter honest proposers
+        // for 74 hours: 1000%/year / 360 days / 24 hours * 74 hours max attack time = ~ 8.6%.
+        // Note: 1000% range is equivalent to a funding rate bounds of +500%/-500%. Imagine that the market is very
+        // volatile currently and that the "true" funding rate for the next 74 hours is -500%, but a dishonest
+        // proposer successfully proposes a rate of +500% (after a two hour liveness) and disputes honest
+        // proposers for the next 72 hours. This results in a funding rate error of 1000% for 74 hours, until the DVM
         // can set the funding rate back to its correct value.
+
+        // Approximate range is corresponding to +/- 500% is [0.000016, -0.000016].
         FixedPoint.Signed memory MAX_FUNDING_RATE_PER_SECOND =
-            FixedPoint.fromUnscaledInt(300).div(360).div(24).div(60).div(60);
+            FixedPoint.fromUnscaledInt(500).div(360).div(24).div(60).div(60);
         FixedPoint.Signed memory MIN_FUNDING_RATE_PER_SECOND =
-            FixedPoint.fromUnscaledInt(-300).div(360).div(24).div(60).div(60);
+            FixedPoint.fromUnscaledInt(-500).div(360).div(24).div(60).div(60);
 
         require(
             rate.isLessThanOrEqual(MAX_FUNDING_RATE_PER_SECOND) &&
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -260,17 +260,16 @@ abstract contract FundingRateApplier is FeePayer {
         return fundingRate.rate;
     }
 
+    // Constraining the range of funding rates limits the PfC for any dishonest proposer and enhances the
+    // perpetual's security. For example, let's examine the case where the max and min funding rates
+    // are equivalent to +/- 500%/year. This 1000% funding rate range allows a 8.6% profit from corruption for a
+    // proposer who can deter honest proposers for 74 hours:
+    // 1000%/year / 360 days / 24 hours * 74 hours max attack time = ~ 8.6%.
+    // How would attack work? Imagine that the market is very volatile currently and that the "true" funding
+    // rate for the next 74 hours is -500%, but a dishonest proposer successfully proposes a rate of +500%
+    // (after a two hour liveness) and disputes honest proposers for the next 72 hours. This results in a funding
+    // rate error of 1000% for 74 hours, until the DVM can set the funding rate back to its correct value.
     function _validateFundingRate(FixedPoint.Signed memory rate) internal view {
-        // Constraining the range of funding rates limits the PfC for any dishonest proposer and enhances the
-        // perpetual's security. For example, let's examine the case where the max and min funding rates
-        // are equivalent to +/- 500%/year. This 1000% funding rate range allows a 8.6% profit from corruption for a
-        // proposer who can deter honest proposers for 74 hours:
-        // 1000%/year / 360 days / 24 hours * 74 hours max attack time = ~ 8.6%.
-        // How would attack work? Imagine that the market is very volatile currently and that the "true" funding
-        // rate for the next 74 hours is -500%, but a dishonest proposer successfully proposes a rate of +500%
-        // (after a two hour liveness) and disputes honest proposers for the next 72 hours. This results in a funding
-        // rate error of 1000% for 74 hours, until the DVM can set the funding rate back to its correct value.
-
         require(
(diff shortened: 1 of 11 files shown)


packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -259,6 +260,26 @@ abstract contract FundingRateApplier is FeePayer {
...
+        // Constraining range of funding rates limits the PfC for any dishonest proposer and enhances the perpetual's
+        // security. A 600% funding rate range is equal to a 5% PfC for a proposer who can deter honest proposers
+        // for 74 hours: 600%/year / 360 days / 24 hours * 74 hours max attack time = ~ 5%.
+        // Note: 600% range is equivalent to a funding rate bounds of +300%/-300%. Imagine that the market is very
+        // volatile currently and that the "true" funding rate for the next 74 hours is -300%, but a dishonest
+        // proposer successfully proposes a rate of +300% (after a two hour liveness) and disputes honest
+        // proposers for the next 72 hours. This results in a funding rate error of 600% for 74 hours, until the DVM
+        // can set the funding rate back to its correct value.
+        FixedPoint.Signed memory MAX_FUNDING_RATE_PER_SECOND =
+            FixedPoint.fromUnscaledInt(300).div(360).div(24).div(60).div(60);
```
nicholaspai, line 273: Would it be more readable if I approximated these values as `1e5` and `-1e5`?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -241,7 +241,7 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         PositionData storage positionToLiquidate = _getPositionData(sponsor);
 
         tokensLiquidated = FixedPoint.min(maxTokensToLiquidate, positionToLiquidate.tokensOutstanding);
-        require(tokensLiquidated.isGreaterThan(0), "Liquidating 0 tokens");
```
nicholaspai, line 244: I needed to claw back some bytecode so I took some liberties with which require messages I removed. Please take a hard look at these and let me know what you think.

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -815,14 +815,29 @@ contract("PerpetualPositionManager", function(accounts) {
     await setFundingRateAndAdvanceTime("0");
     await positionManager.applyFundingRate();
 
-    // Clock has been advanced during the proposal by 10k seconds. The fee rate is one one-thousandth of a percent.
+    // Clock has been advanced during the proposal by 10k seconds. The reward rate is one one-thousandth of a percent.
```
nicholaspai, line 818: This test was a bit unclear, its title is "Basic funding rate fees", which reference the fact that sponsors' locked collateral is taxed to pay funding rate rewards.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -259,6 +260,28 @@ abstract contract FundingRateApplier is FeePayer {
...
+        // for 74 hours: 1000%/year / 360 days / 24 hours * 74 hours max attack time = ~ 8.6%.
+        // Note: 1000% range is equivalent to a funding rate bounds of +500%/-500%. Imagine that the market is very
+        // volatile currently and that the "true" funding rate for the next 74 hours is -500%, but a dishonest
+        // proposer successfully proposes a rate of +500% (after a two hour liveness) and disputes honest
+        // proposers for the next 72 hours. This results in a funding rate error of 1000% for 74 hours, until the DVM
+        // can set the funding rate back to its correct value.
+
+        // Approximate range is corresponding to +/- 500% is [0.000016, -0.000016].
+        FixedPoint.Signed memory MAX_FUNDING_RATE_PER_SECOND =
+            FixedPoint.fromUnscaledInt(500).div(360).div(24).div(60).div(60);
```
daywiss, line 275: i wonder if the 500 number should be set as a contract constructor? could this also save cost for the caller if computation was done in constructor and saved.

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -198,7 +198,7 @@ contract("FundingRateApplier", function(accounts) {
   });
 
   it("Proposal time checks", async () => {
-    const newRate = { rawValue: toWei("-0.0001") };
+    const newRate = { rawValue: toWei("-0.000001") };
```
daywiss, line 201: what happened here?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
daywiss: this is sad but what choice do we have!

packages/core/test/financial-templates/common/FundingRateApplier.js
nicholaspai: The funding rate range is now constrained to [0.000001, -0.000001] so this rate had to be changed.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Oh this is a good idea, we could allow the deployer to update this limit?

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -1051,76 +1074,75 @@ contract("PerpetualPositionManager", function(accounts) {
 
     assert.equal((await positionManager.fundingRate()).cumulativeMultiplier.toString(), toWei("1.05"));
 
-    // Set the funding rate to a negative funding rate of -0.00001 in the store and apply it for 10k seconds. New funding rate
-    // should be 1.05 * (1 - -0.00001 * 10k) = 0.945
-    await setFundingRateAndAdvanceTime(toWei("-0.00001"));
+    // Set the funding rate to a negative funding rate of -0.000001 in the store and apply it for 10k seconds. New funding rate
```
nicholaspai, line 1077: The following changes came about originally because I set a bound on the funding rate on Perpetual construction, but now that the bound itself can be configured, to no bound for example, the changes are not technically neccessary. However, I'm going to keep them because they are more readable than before.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
daywiss: interesting... i think since its a magic number id lean towards yes. But if this value could be shown to be optimal through most use cases id say its ok to hard code.

nicholaspai: I ended up allowing the ConfigStore to update these max/min values. When we deploy our first Perp we will prob set them closely to the previously hard coded values

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -259,6 +260,23 @@ abstract contract FundingRateApplier is FeePayer {
         return fundingRate.rate;
     }
 
+    function _validateFundingRate(FixedPoint.Signed memory rate) internal view {
+        // Constraining the range of funding rates limits the PfC for any dishonest proposer and enhances the
```
chrismaree, line 264: nit: do you want to put this comment above the function name?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -280,8 +279,7 @@ contract PerpetualPositionManager is FundingRateApplier {
         PositionData storage positionData = _getPositionData(msg.sender);
         require(
```
chrismaree, line 280: I feel that this one might get hit more frequently than some of the others.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -385,7 +383,7 @@ contract PerpetualPositionManager is FundingRateApplier {
         returns (FixedPoint.Unsigned memory amountWithdrawn)
     {
         PositionData storage positionData = _getPositionData(msg.sender);
-        require(numTokens.isLessThanOrEqual(positionData.tokensOutstanding), "Invalid token amount");
```
chrismaree, line 388: :(

chrismaree (approved):
Looks good to me! clean, well commented and overall a good PR.

daywiss (approved):

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: done

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
nicholaspai: I think these reverts whenever the user passes in 0 tokens don't need require messages because users are less likley to do this in the firs tplace

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: agreed ill re-add i have 500 bytes to work with

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: This one I removed for the same reason as the `require(collateralAmount.isGreaterThan(0), "Invalid collateral amount")` which is that these requires should get hit less frequently because honest acting users won't try to input 0 tokens or a number thats clearly over their balance/debt outstanding.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -160,17 +168,24 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
...
-        require(config.proposerBondPct.isLessThan(maxProposerBond), "Invalid proposerBondPct");
-
         // Reward rate should be less than 100% a year => 100% / 360 days / 24 hours / 60 mins / 60 secs
         // = 0.0000033
         FixedPoint.Unsigned memory maxRewardRatePerSecond = FixedPoint.fromUnscaledUint(33).div(1e7);
         require(config.rewardRatePerSecond.isLessThan(maxRewardRatePerSecond), "Invalid rewardRatePerSecond");
+
+        // We don't set a limit on the proposer bond because it is a defense against dishonest proposers. If a proposer
+        // were to successfully propose a very high or low funding rate, then their PfC would be very high. The proposer
+        // could theoretically keep their "evil" funding rate alive for 74 hours by continuously disputing honest
```
mrice32, line 184: nit: a proposer can theoretically keep disputing indefinitely and a new funding rate will never come in. There is some logic we can add in so that the DVM settlement actually pushes the new funding rate into the contract but it was stripped for bytecode.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -153,24 +161,33 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
...
-        // Proposer bond of 0.04% is based on a maximum expected funding rate error of 200%/year.
-        FixedPoint.Unsigned memory maxProposerBond = FixedPoint.fromUnscaledUint(4).div(1e4);
-        require(config.proposerBondPct.isLessThan(maxProposerBond), "Invalid proposerBondPct");
-
         // Reward rate should be less than 100% a year => 100% / 360 days / 24 hours / 60 mins / 60 secs
         // = 0.0000033
         FixedPoint.Unsigned memory maxRewardRatePerSecond = FixedPoint.fromUnscaledUint(33).div(1e7);
         require(config.rewardRatePerSecond.isLessThan(maxRewardRatePerSecond), "Invalid rewardRatePerSecond");
+
+        // We don't set a limit on the proposer bond because it is a defense against dishonest proposers. If a proposer
```
mrice32, line 182: I think we might want to mention the downside of not limiting this -- namely that the config store owner can set it arbitrarily high and preclude a new funding rate from ever coming in. For that reason, it _could_ (not saying it necessarily should) make sense to hard limit this in the perp to 100% since the pfc should never be more than 100% of the collateral in the contract, right?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: Each dispute can effectively delay a funding rate for 2 hours.

nicholaspai: Oh shoot I thought the DVM settlement pushed the funding rate into the contract? That was a core assumption in calculating our PfC, any reason not to add this?

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: done

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: Added this comment, we resolved this offline FYI. Our PfC calculation found [here](https://github.com/UMAprotocol/protocol/issues/2039#issuecomment-719734383) actually calculates the proposal bond per proposal liveness window (i.e. 2 hours)

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -153,24 +161,34 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
 
     // Use this method to constrain values with which you can set ConfigSettings.
     function _validateConfig(ConfigStoreInterface.ConfigSettings memory config) internal pure {
-        // Its hard to ascertain what are reasonable limits, into the future and the past, for proposal timestamps.
-        // So, we do not set any limits.
+        // We don't set limits on proposal timestamps because there are already natural limits:
+        // - Future: price requests to the OptimisticOracle must be in the past (plus some buffer allowing for the time
```
nicholaspai, line 165: @mrice32 The OO doesn't currently enforce requests to be in the past, I can wait to. merge this until that PR is done.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: +1 -- thanks for checking up on this.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
mrice32: Sounds good. Will put up a PR momentarily.

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -153,24 +161,34 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
 
     // Use this method to constrain values with which you can set ConfigSettings.
     function _validateConfig(ConfigStoreInterface.ConfigSettings memory config) internal pure {
-        // Its hard to ascertain what are reasonable limits, into the future and the past, for proposal timestamps.
-        // So, we do not set any limits.
+        // We don't set limits on proposal timestamps because there are already natural limits:
```
mrice32, line 164: Should we just remove the future limit from the config now that the OO enforces a future limit, and this limit would never be hit?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -169,6 +167,7 @@ abstract contract FundingRateApplier is FeePayer {
         // Set up optimistic oracle.
         bytes32 identifier = fundingRate.identifier;
         bytes memory ancillaryData = _getAncillaryData();
+        // Note: requestPrice will revert if `timestamp` is less than the current block timestamp
```
mrice32, line 170:
```suggestion
        // Note: requestPrice will revert if `timestamp` is less than the current block timestamp.
```

mrice32 (approved):

packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
nicholaspai: Done!

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -214,10 +223,8 @@ contract("FundingRateApplier", function(accounts) {
     // Time must be within the past and future bounds around the current time.
     assert(await didContractThrow(fundingRateApplier.proposeNewRate(newRate, currentTime - proposalTimePastLimit - 1)));
     await fundingRateApplier.proposeNewRate.call(newRate, currentTime - proposalTimePastLimit);
-    assert(
-      await didContractThrow(fundingRateApplier.proposeNewRate(newRate, currentTime + proposalTimeFutureLimit + 1))
-    );
-    await fundingRateApplier.proposeNewRate(newRate, currentTime + proposalTimeFutureLimit);
+    assert(await didContractThrow(fundingRateApplier.proposeNewRate(newRate, currentTime + 1)));
```
nicholaspai, line 226: This test will fail until #2267 is merged

PR #2262, opened 2020-12-03
improve(perpetual) move proposal time bounds into config store

### LINKED ISSUE #2228: Go through PerpetualContract and add hard-coded values to ConfigStore
**Why**

[this comment should have a follow up PR](https://github.com/UMAprotocol/protocol/pull/2212#discussion_r530849290)

- 3: ~1 hour

nicholaspai:

**Motivation**

#2228

**Summary**

TODO: Determine if we should set reasonable limits on these time bounds.

This adds 100 bytes of bytecode to `PerpetualLib`.

**Issue(s)**

Fixes #2228


chrismaree (approved):
Nice! I think pulling as much as we can into the config is great. so this is a good change.

daywiss (approved):
nice @nicholaspai, i think this makes a lot of sense

PR #2263, opened 2020-12-03
improve(affiliates): add some basic test cases to validate payout logic

### LINKED ISSUE #2175: Add some synthetic test cases to verify final output
Affiliates package currently uses real data stored in a dataset for testing, but the output results cant easily be verified. We should add some synthetic inputs just so we can calculate results by hand to validate in a test.

- mock up a small synthetic dataset into a test
- hand calculate correct output to compare
- improve tests for historical models and processors

daywiss:

**Motivation**

#2175 

**Summary**

This adds some very basic tests based on easy to reason about scenarios
- scenario 1: 2 emps and only 1 has had a balance: 100% of rewards go to emp with balance
- scenario 2: bunch of emps, and 2 have equal balance: 50% of rewards get split between emps with balance

There is some extra code here around mocking libraries which did not end up getting used. It was too complicated to mock on chain transaction data with custom values. 

There were a bunch of nasty merge issues in the tests/mocha/affiliates. I will leave a comment on the only part that changed.

**Issue(s)**

Fixes #2175

### DIFF
--- a/packages/affiliates/libs/mocks.js
+++ b/packages/affiliates/libs/mocks.js
@@ -1,0 +1,70 @@
+// Mocks that allow you to inject in your datasets as arrays/objects for testing. This is different from
+// dataset mocks which load mocked data from a file. This allows you to pass in custom data.
+// This was going to be used for simple testing but its too difficult to mock transaction data.
+const highland = require("highland");
+const assert = require("assert");
+
+// Mocking the big query library.
+// Pass in blocks as an array of blocks
+// Pass in logs as logs[contractAddress] = [{...log data}]
+function Queries({ blocks = [], logs = {} }) {
+  function streamAllLogsByContract(address) {
+    return highland(getAllLogsByContract(address));
+  }
+  function getAllLogsByContract(address) {
+    assert(logs[address], "no logs for address: " + address);
+    return logs[address];
+  }
+  function streamLogsByContract(address, start, end) {
+    return highland(getLogsByContract(address, start, end));
+  }
+  function getLogsByContract(address, start, end) {
+    assert(logs[address], "no logs for address: " + address);
+    return logs[address].filter(x => {
+      return x.block_timestamp >= start && x.block_timestamp <= end;
+    });
+  }
+  function streamBlocks(start, end) {
+    return highland(getBlocks(start, end));
+  }
+  function getBlocks(start, end) {
+    return blocks.filter(x => {
+      return x.timestamp >= start && x.timestamp <= end;
+    });
+  }
+  return {
+    streamLogsByContract,
+    streamAllLogsByContract,
+    streamBlocks,
+    getBlocks,
+    getLogsByContract,
+    getAllLogsByContract
+  };
+}
+function Coingecko({ prices = {} }) {
+  return {
+    getHistoricContractPrices(address, currency, start, end) {
+      assert(prices[address], "no prices found for address: " + address);
+      return prices[address].filter(x => {
+        return x[0] >= start && x[0] <= end;
+      });
+    }
+  };
+}
+
+function SynthPrices({ prices = {} }) {
+  return {
+    getHistoricSynthPrices(address, start, end) {
+      assert(prices[address], "no prices found for address: " + address);
+      return prices[address].filter(x => {
+        return x[0] >= start && x[0] <= end;
+      });
+    }
+  };
+}
+
+module.exports = {
+  Coingecko,
+  SynthPrices,
+  Queries
+};
--- a/packages/affiliates/libs/models.js
+++ b/packages/affiliates/libs/models.js
@@ -171,7 +171,7 @@ function History() {
   const history = [];
   // Used internally, but will insert a block into cache sorted by timestamp ascending
   function insert(data) {
-    assert(data.blockNumber === 0 || data.blockNumber > 0, "requires blockNumber");
+    assert(typeof data.blockNumber === "number" && data.blockNumber >= 0, "requires blockNumber");
     const index = lodash.sortedIndexBy(history, data, "blockNumber");
     history.splice(index, 0, data);
     return data;
--- a/packages/affiliates/libs/models.js
+++ b/packages/affiliates/libs/models.js
@@ -171,13 +171,14 @@ function History() {
   const history = [];
   // Used internally, but will insert a block into cache sorted by timestamp ascending
   function insert(data) {
-    assert(data.blockNumber, "requires blockNumber");
+    assert(data.blockNumber === 0 || data.blockNumber > 0, "requires blockNumber");
     const index = lodash.sortedIndexBy(history, data, "blockNumber");
     history.splice(index, 0, data);
     return data;
   }
   function lookup(blockNumber) {
     const index = lodash.sortedIndexBy(history, { blockNumber }, "blockNumber");
+    assert(history.length, "history is empty");
     if (history[index] && history[index].blockNumber === blockNumber) return history[index];
     const result = history[index - 1];
     assert(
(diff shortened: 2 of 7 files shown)


packages/affiliates/test/mocha/affiliates.js
```diff
@@ -21,129 +32,249 @@ const devRewardsToDistribute = "50000";
...
-    const queries = Queries(datasetPath);
-    const coingecko = Coingecko(datasetPath);
-    const synthPrices = SynthPrices(datasetPath);
-    affiliates = DeployerRewards({
-      queries,
-      empAbi,
-      empCreatorAbi,
-      coingecko,
-      synthPrices
+  describe("CalculateRewards Simple Data", function() {
```
daywiss, line 36: this is the new section of tests that were added. The rest of the diffs were from merge issues and shouldnt have changed, though its hard to see.

packages/affiliates/libs/mocks.js
```diff
@@ -0,0 +1,70 @@
+// Mocks that allow you to inject in your datasets as arrays/objects for testing. This is different from
```
nicholaspai, line 1: +1 these are super useful for testing

packages/affiliates/libs/models.js
```diff
@@ -171,13 +171,14 @@ function History() {
...
   function insert(data) {
-    assert(data.blockNumber, "requires blockNumber");
+    assert(data.blockNumber === 0 || data.blockNumber > 0, "requires blockNumber");
     const index = lodash.sortedIndexBy(history, data, "blockNumber");
     history.splice(index, 0, data);
     return data;
   }
   function lookup(blockNumber) {
     const index = lodash.sortedIndexBy(history, { blockNumber }, "blockNumber");
+    assert(history.length, "history is empty");
```
nicholaspai, line 181: What if `history === undefined`, will this unexpectedly throw an error?

nicholaspai (commented):
Can you point to the tests where there are multiple EMP's but only 1 has balance?

Also, do you test for situations where there are multiple EMP's, both have balance at a start-date/block, but then 1 expires and sees all of its collateral drained/withdrawn?

packages/affiliates/libs/models.js
daywiss: history is defined as a const and given an empty array as the default value in construction so it should be ok

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -21,129 +32,249 @@ const devRewardsToDistribute = "50000";
...
-  it("getAllBalanceHistory", async function() {
-    this.timeout(10000);
-    const result = await affiliates.utils.getAllBalanceHistories(empContracts, startingTimestamp, endingTimestamp);
-    assert.equal(result.length, empContracts.length);
-    result.forEach(([address, history]) => {
-      assert.ok(address);
-      assert.ok(history);
-      assert.ok(history.history.length());
+    it("should work with basic test cases", function() {
+      // add some balance histories to the emp contracts. this is really what ends up adjusting the distribution.
```
daywiss, line 85: i should probably name these tests better but here @nicholaspai 

Below we add a single position to the balance history at block 0. The datastructure passed in is what a the decoded `PositionCreated` event looks like in the emp.

daywiss:
> Also, do you test for situations where there are multiple EMP's, both have balance at a start-date/block, but then 1 expires and sees all of its collateral drained/withdrawn?

That would be a really good test. We have a real dataset that does something close to this which has been run and it actually uncovered at least 1 issue which has been fixed. I'll try to work on a synthetic approximation of this.

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -21,129 +32,249 @@ const devRewardsToDistribute = "50000";
...
-    assert(diff > 0 ? diff : -diff < epsilon);
-  });
-  it("calculateValueFromUsd", async function() {
-    // these are all stable coins so they should roughly be around 1 dollar
-    // epsilon is high because variations could be nearly a dollar in any direction
-    const target = 10n ** 18n;
-    const syntheticPrice = 26.358177384415466;
-    let result = affiliates.utils.calculateValueFromUsd(target, 0, syntheticPrice, 18, 18).toString();
-    assert.equal(result, toWei(syntheticPrice.toFixed(18)));
+    it("should work with basic test cases", function() {
```
mrice32, line 97: nit: are these two test cases named the same thing?

packages/affiliates/libs/models.js
```diff
@@ -171,13 +171,14 @@ function History() {
   const history = [];
   // Used internally, but will insert a block into cache sorted by timestamp ascending
   function insert(data) {
-    assert(data.blockNumber, "requires blockNumber");
+    assert(data.blockNumber === 0 || data.blockNumber > 0, "requires blockNumber");
```
mrice32, line 174: What is this testing? Not entirely self-documenting since it _feels_ like this could be reduced to blockNumber >= 0? Is it trying to ensure the type is a number too? Or that it's also not undefined? Might be easier to test that in a specific ||, like:
```
assert(typeof(data.blockNumber) === "number" && data.blockNumber >= 0, "requires blockNumber");
```

Or a comment would suffice -- I just find the == 0 || > 0 expression a little confusing.

packages/affiliates/test/mocha/affiliates.js
daywiss: I have an update coming to better name these test cases, it was an oversight

packages/affiliates/libs/models.js
daywiss: The goal was to allow 0 and above but not undefined or null, it was the best i could come up with, but yours i think i agree is easier to understand! Will update.

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -21,129 +32,284 @@ const devRewardsToDistribute = "50000";
...
+        args: ["bb", "2", "1"],
+        blockTimestamp: 0
+      });
+      balanceHistories[1][1].finalize();
+
+      const result = affiliates.utils.calculateRewards(params);
+      assert.equal(result.empPayouts["a"], totalRewards / 2);
+      assert.equal(result.empPayouts["b"], totalRewards / 2);
+    });
+    it("should work with an emp which had balance and expired", function() {
```
nicholaspai, line 120: These titles are very readable now, thanks!

nicholaspai (approved):
LGTM! Thanks for changing the test titles

packages/affiliates/test/mocha/affiliates.js
```diff
@@ -21,129 +32,284 @@ const devRewardsToDistribute = "50000";
...
+        name: "PositionCreated",
+        // creating a position for address "aa" with 2 collateral 1 synthetic
+        args: ["aa", "2", "1"],
+        blockTimestamp: 0
+      });
+      balanceHistories[0][1].finalize();
+      const result = affiliates.utils.calculateRewards(params);
+      assert.equal(result.empPayouts["a"], totalRewards);
+    });
+    it("should split rewards equally between equally funded emps", function() {
```
mrice32, line 97: Could be nice to have a test that verified that it's splitting based on synthetics and not collateral -- i.e. make the ratio of synths different from the ratio of collaterals.

packages/affiliates/test/mocha/affiliates.js
mrice32, line 120: Do we have an open issue on whether this is the desired behavior? I know we've discussed this a little, but it seems _bad_ to reward people for expired synths. Not saying we should change immediately, but could you open an issue to discuss?

mrice32 (approved):
LGTM!

packages/affiliates/test/mocha/affiliates.js
daywiss: #2265

PR #2264, opened 2020-12-04
fix(deploy-script): add version parameter to deployment script

mrice32:
**Motivation**

Deployment script needs to use older abis to interact with mainnet contracts.

**Summary**

Moved the deployment script to using a custom abi version that the user can input.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/local/DeployEMP.js
+++ b/packages/core/scripts/local/DeployEMP.js
@@ -19,25 +19,29 @@
  * - `./PushPriceEMP.js`: "resolves" a pending mock oracle price request with a price.
  *
  *
- * Example: $(npm bin)/truffle exec ./scripts/local/DeployEMP.js --network test --test true --identifier ETH/BTC
+ * Example: $(npm bin)/truffle exec ./scripts/local/DeployEMP.js --network test --test true --identifier ETH/BTC --cversion 1.1.0
  */
 const { toWei, utf8ToHex, hexToUtf8 } = web3.utils;
 const { interfaceName } = require("@uma/common");
+const { getAbi, getTruffleContract } = require("../../index");
+const argv = require("minimist")(process.argv.slice(), {
+  boolean: ["test"],
+  string: ["identifier", "collateral", "cversion"]
+});
+const abiVersion = argv.cversion || "1.1.0"; // Default to most recent mainnet deployment, 1.1.0.
 
 // Deployed contract ABI's and addresses we need to fetch.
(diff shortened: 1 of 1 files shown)


packages/core/scripts/local/DeployEMP.js
```diff
@@ -115,16 +111,16 @@ const deployEMP = async callback => {
...
     await expiringMultiPartyCreator.createExpiringMultiParty(constructorParams, { from: deployer });
-    emp = await ExpiringMultiParty.at(_emp);
+    const emp = await ExpiringMultiParty.at(_emp);
 
     const empConstructorParams = {
       ...constructorParams,
-      finderAddress: Finder.address,
-      tokenFactoryAddress: TokenFactory.address,
+      finderAddress: finder.address,
+      tokenFactoryAddress: tokenFactory.address,
```
mrice32, line 119: Note: when we initialize using `getTruffleContract`, we have to use `Finder.deployed()` rather than `Finder.address` because truffle needs an async operation to detect the network.

packages/core/scripts/local/DeployEMP.js
```diff
@@ -19,34 +19,29 @@
...
-const { getAbi } = require("../../index");
-
-// Contracts we need to interact with.
-let collateralToken;
-let emp;
-let mockOracle;
-let identifierWhitelist;
-let collateralTokenWhitelist;
-let expiringMultiPartyCreator;
-let store;
```
mrice32, line 49: Note: just for cleanliness, I got rid of these globals since they aren't really necessary since they are _only_ used in a single scope.

packages/core/scripts/local/DeployEMP.js
```diff
@@ -19,34 +19,29 @@
...
  *
- * Example: $(npm bin)/truffle exec ./scripts/local/DeployEMP.js --network test --test true --identifier ETH/BTC
+ * Example: $(npm bin)/truffle exec ./scripts/local/DeployEMP.js --network test --test true --identifier ETH/BTC --cversion 1.1.0
  */
 const { toWei, utf8ToHex, hexToUtf8 } = web3.utils;
 const { interfaceName } = require("@uma/common");
+const { getAbi, getTruffleContract } = require("../../index");
+const argv = require("minimist")(process.argv.slice(), {
+  boolean: ["test"],
+  string: ["identifier", "collateral", "cversion"]
```
mrice32, line 29: Used `cversion` (as in contract version) because I didn't want outer layers of node/yarn processes to get confused by the arg `version`.

nicholaspai (approved):
Nicely done LGTM.

PR #2274, opened 2020-12-08
improve(voter-gas-rebate): Rebate 4 covering + improvements to the script

nicholaspai:
**Motivation**

- Adds helpful summary data to `Rebate.json` output file including: # of voters, unique identifier-timestamp combinations covered in votes for a given script run, total rebate amount of tokens
- I ran the script retrospectively on the previous rebates for the current DVM (note that `Rebate_1.json` was for a deprecated `Voting` contract so I ignored it for simplicity) and added in the summary data to the output JSONs
- Adds the feature for the script to filter on block # instead of UNIX timestamp
- Adds Rebate 4 covering the blocks from the end of Rebate 3 until now


chrismaree (approved):

mrice32 (approved):
LGTM!

PR #2276, opened 2020-12-09
improve(liquidator): enable arbitrary decimal combinations between collateral, synthetic and price feed in liquidator bot

chrismaree:
**Motivation**

issue #2351 has a fair amount of detail on the problem space.

The current UMA ecosystem of contracts(EMP & perp) contains a range of different decimal configurations between a) synthetic, b) collateral & c) price feed. The bots should be updated to accommodate any combination of the decimal convention.

Below the different possible configurations are listed:

1) The synthetic, collateral, and price feed decimals all match up. This is the simplest and most common configuration and occurs when the collateral type has 18 decimals (such as WETH or DAI). In this case nothing fancy needs to happen.
2) The synthetic is 18 decimals, collateral is 8 decimals and the price feed is 8 decimals. This is the case with old uUSDrBTC contracts and BTC denominated contracts expiring as late as March.
3) The synthetic is 8 decimals, collateral is 8 and the price feed is 18 decimals. This is the latest form of decimal delimitation, with the latest EMP & Perp conforming to this standard. This format was chosen to make voting easier as price feeds always have the same number of decimals (18) and the collateral matches synthetic to accommodate scaling within the contract.

The complexity of this issue is to be able to accommodate all 3 types of delimitation at the same time and to enable the bots to dynamically handle each type.

**Summary**

To accommodate different decimals combinations this PR introduces the following pattern: price feeds are parameterized with a `decimals` prop that defines how the DVM would respond to a given price for a given identifier. Then, the decimals for the collateral and synthetic are considered when computing position collateralization.

**Details**

To see this convention, consider the calculation to define if a position is correctly collateralized, as used within the EMPClient in the `_isUnderCollateralized` method. This explanation goes into a bit of detail as this pattern will need to be re-applied through the stack and it is important that @UMAprotocol/eng is happy with the pattern set out here. In the case of an undercollateraled position the following inequality is evaluated:

image

If all terms within this expression are scaled by 1e18 then there is no complexity as the decimals cancel through the expression. However, we need to be able to consider _any_ combination. In particular:

- `numTokens` is scaled by synthetic decimals (`sD`)
- `tokenRedemptionValue` is scaled by the price feed decimals (`trvD`)
- `amountCollateral` is scaled by the collateral decimals (`cD`)
- `collateralRequirement` is always scaled by `10^18`

considering the decimals of each term within the expression this equation becomes:
image
(note this equation wont work in practice as the scaling through is not normalized correctly through a correction factor; rather, it aims to illustrate the problem space).

**The solution presented in this PR addresses this problem by normalizing the number of decimals through the expression such that all terms use `10^18` as their scaling unit.** In doing so each term (`numTokens`, `amountCollateral` etc) are considered "base units" with no scaling applied and all terms within the equation are _consistantly scaled_.

This normalization is done using the existing https://github.com/UMAprotocol/protocol/blob/00ff20195ae5dcc6c0f6c4e237206fbbbc60b069/packages/common/src/FormattingUtils.js#L135 method within the `FormattingUtils` common helper. This method works by converting a base number, delimited in `fromDecimals` to a unitless number, delimited in `toDecimals`. For example 100 USDC, which has 6 decimal points (represented as `100*10^6` in Ethereum) would be converted to `100*10^18` with this function, thereby transforming the `fromDecimals` of 6 to the `toDecimals` of 18.

Using this pattern, all input numbers are normalized and then the evaluation can be performed directly on the expression without considering the underlying decimals as all terms are equally scaled.

**Tests update**
To correctly validate the behavior of the changes applied in this PR, a fair amount of changes were needed to be made to the tests. Bot unit tests run a set of 3 different decimal configurations to accommodate the three combinations identified at the top of the PR. Specifically, these are encompassed in the following config:
```
[
  {
    tokenName: "Wrapped Ether",
    tokenSymbol: "WETH",
    collateralDecimals: 18,
    syntheticDecimals: 18,
    priceFeedDecimals: 18
  },
  {
    tokenName: "Legacy Wrapped Bitcoin",
    tokenSymbol: "BTC",
    collateralDecimals: 8,
    syntheticDecimals: 18,
    priceFeedDecimals: 8
  },
  {
    tokenName: "Wrapped Bitcoin",
    tokenSymbol: "BTC",
    collateralDecimals: 8,
    syntheticDecimals: 8,
    priceFeedDecimals: 18
  }
```
Then, each test defines a set of 3 conversion functions `convertCollateral`, `convertSynthetic` & `convertPrice` which scale inputs depending on a the particular test. These are used when creating positions, defining prices, or any operation that deals with collateral, synthetics or price respectively.

**Issue(s)**

First PR addressing issue #2351. Next PRs will apply the same changes to the monitor and disputer.


packages/common/src/FormattingUtils.js
```diff
@@ -124,9 +124,9 @@ function addSign(number) {
   }
 }
 
-// Take an amount based on fromDecimals and convert it to an amount based on toDecimals
```
chrismaree, line 127: Updated the comment here to be clearer.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,19 +1,32 @@
 // A thick client for getting information about an ExpiringMultiParty. Used to get sponsor information, outstanding
 // positions, undisputed Liquidations, expired liquidations, disputed liquidations.
 
-const { ConvertDecimals, parseFixed, LiquidationStatesEnum } = require("@uma/common");
+const { ConvertDecimals, LiquidationStatesEnum } = require("@uma/common");
 const Promise = require("bluebird");
-
 class ExpiringMultiPartyClient {
```
chrismaree, line 6: The core logic for dealing with decimals now takes place within this `ExpiringMultiPartyClient`. Previously, the price feed was scaled according to decimals of the collateral type to accommodate BTC collateral currencies. After this PR the Client takes in all three decimals and uses them within the computation to calculate position collateralization.

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -39,13 +40,20 @@ class BalancerPriceFeed extends PriceFeedInterface {
...
+
   getLastUpdateTime() {
     return this.lastUpdateTime;
   }
+
   getCurrentPrice() {
     // current price can be undefined, will throw for any other errors
     return this.priceHistory.currentPrice();
   }
+
```
chrismaree, line 52: This method was added to the price feed (and others) to enable the `index.js` entry point to query the number of decimals within a price feed.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -49,6 +49,7 @@ const defaultConfigs = {
     lookback: 7200,
     invertPrice: true,
     minTimeBetweenUpdates: 60,
+    priceFeedDecimals: 8,
```
chrismaree, line 52: For `USDBTC` contracts that are currently deployed we are using an 8 decimal price feed.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -39,7 +62,9 @@ contract("ExpiringMultiPartyClient.js", function(accounts) {
       let identifierWhitelist;
       let store;
       let identifier;
-      let convert;
+      let convertCollateral;
```
chrismaree, line 65: This is the primary change made to the tests for the `ExpiringMultiPartyClient`: introduction of `convertCollateral` `convertSynthetic` and `convertPrice` which are used accordingly within the tests to validate that the EMP client works with all 3 combinations of decimals.

packages/liquidator/index.js
```diff
@@ -117,12 +122,6 @@ async function run({
       withdrawLiveness
     };
 
-    // Price feed must use same # of decimals as collateral currency.
-    let customPricefeedConfig = {
-      ...priceFeedConfig,
-      decimals: collateralCurrencyDecimals
```
chrismaree, line 123: This previously worked by scaling the price feed by the number of decimals within the collateral currency. Now, the price feed has it's own scaling along with the collateral and decimals.

packages/liquidator/index.js
```diff
@@ -132,26 +131,37 @@ async function run({
...
 
     // Create the ExpiringMultiPartyClient to query on-chain information, GasEstimator to get latest gas prices and an
     // instance of Liquidator to preform liquidations.
-    const empClient = new ExpiringMultiPartyClient(logger, getAbi("ExpiringMultiParty"), web3, empAddress);
+    const empClient = new ExpiringMultiPartyClient(
+      logger,
+      getAbi("ExpiringMultiParty"),
+      web3,
+      empAddress,
+      collateralTokenDecimals,
```
chrismaree, line 160: notice that the EMP client now takes in all 3 respective decimals. Note that if `priceFeed.getPriceFeedDecimals()` this call will error out.

packages/liquidator/test/Liquidator.js
```diff
@@ -16,7 +16,7 @@ const { Liquidator } = require("../src/liquidator.js");
 const {
   ExpiringMultiPartyClient,
   GasEstimator,
-  PriceFeedMockScaled: PriceFeedMock,
```
chrismaree, line 19: no longer need to use the scaled mock.

packages/liquidator/test/Liquidator.js
```diff
@@ -33,16 +33,39 @@ const SyntheticToken = artifacts.require("SyntheticToken");
 const Timer = artifacts.require("Timer");
 const Store = artifacts.require("Store");
 
+// Run the tests against 3 different kinds of token/synth decimal combinations:
```
chrismaree, line 36: As with the emp client we accommodate all 3 types of combinations within the tests.

packages/liquidator/test/index.js
```diff
@@ -152,8 +152,10 @@ contract("index.js", function(accounts) {
       priceFeedConfig: defaultPriceFeedConfig
     });
 
-    // First log should include # of decimals
-    assert.isTrue(spyLogIncludes(spy, 0, "8"));
+    // Third log, which prints the decimal info, should include # of decimals for the price feed, collateral and synthetic
+    assert.isTrue(spyLogIncludes(spy, 3, '"collateralDecimals":8'));
```
chrismaree, line 156: validate that the index correctly pulls the appropriate decimals from the three sources.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -220,20 +233,25 @@ class ExpiringMultiPartyClient {
...
   }
-  _isUnderCollateralized(numTokens, amountCollateral, trv) {
+  // The formula for an undercollateralized position is: (numTokens * trv) * collateralRequirement > amountCollateral.
+  // This equation assumes the decimal points across the inputs are normalized to the same basis. However, this wont always
+  // be the case and so we need to consider arbitrary decimals coming into the equation. When considering decimals of
+  // each variable within the as collateral (cD), synthetic (sD), CR (1e18), trv (trvD) this equation becomes:
+  // numTokens * 10^sD * trv * 10^trvD * collateralRequirement * 10^18 > amountCollateral * 10^cD.
+  // To accommodate these different decimal points we can normalize each term to the 10^18 basis and then apply a "correction"
+  // factor due to the additional scalling from multiplying basis numbers.
+  _isUnderCollateralized(numTokens, amountCollateral, tokenRedemptionValue) {
```
chrismaree, line 243: This is the primary logic change within the EMP client.

daywiss (approved):
This looks good. Just for my understanding, is the idea that all the conversion logic gets moved into the emp client, and the liquidator can just treat everything as if it was 18 dec?

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -38,12 +51,12 @@ class ExpiringMultiPartyClient {
...
-    const Convert = (decimals = 18) => number => this.toBN(parseFixed(number.toString(), decimals).toString());
-
-    // currently not implemented
-    this.convertSynthetic = Convert(syntheticDecimals);
-    this.convertCollateral = Convert(collateralDecimals);
-    this.convertCollateralToSynthetic = ConvertDecimals(collateralDecimals, syntheticDecimals, this.web3);
+    // Define a set of normalization functions. These Convert a number delimited with given base number of decimals to a
+    // number delimited with a given number of decimals (18). For example, consider normalizeCollateralDecimals. 100 BTC
+    // is 100*10^8. This function would return 100*10^18, thereby converting collateral decimals to 18 decimal places.
+    this.normalizeCollateralDecimals = ConvertDecimals(collateralDecimals, 18, this.web3);
```
mrice32, line 57: nit: Not sure if this should be changed here or in another PR, but I don't think `ConvertDecimals` needs to take in web3, it should just be able to import it, right?

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -220,20 +233,25 @@ class ExpiringMultiPartyClient {
...
+  // be the case and so we need to consider arbitrary decimals coming into the equation. When considering decimals of
+  // each variable within the as collateral (cD), synthetic (sD), CR (1e18), trv (trvD) this equation becomes:
+  // numTokens * 10^sD * trv * 10^trvD * collateralRequirement * 10^18 > amountCollateral * 10^cD.
+  // To accommodate these different decimal points we can normalize each term to the 10^18 basis and then apply a "correction"
+  // factor due to the additional scalling from multiplying basis numbers.
+  _isUnderCollateralized(numTokens, amountCollateral, tokenRedemptionValue) {
+    // Normalize the inputs. Now all terms are 18 decimal delimited and no extra conversion is needed.
+    const normalizedNumTokens = this.normalizeSyntheticDecimals(numTokens);
+    const normalizedAmountCollateral = this.normalizeCollateralDecimals(amountCollateral);
+    const normalizedTokenRedemptionValue = this.normalizePriceFeedDecimals(tokenRedemptionValue);
```
mrice32, line 247: +1 -- this is a very clear way to perform this math.

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -72,6 +72,16 @@ class MedianizerPriceFeed extends PriceFeedInterface {
     return Math.max(...lastUpdateTimes);
   }
 
+  // Gets the decimals of the medianized price feeds. Errors out if any price feed had a different number of decimals.
+  getPriceFeedDecimals() {
+    const priceFeedDecimals = this.priceFeeds.map(priceFeed => priceFeed.getPriceFeedDecimals());
+    if (priceFeedDecimals.every((val, i, arr) => val !== arr[0])) {
```
mrice32, line 78: Should this be `any` rather than `every`?

Also might be slightly simpler/more readable to leave out the latter two args and use `priceFeedDecimals[0]` (or pull that value into a named variable before the call), but I could see arguments for doing it the way you have it too.

mrice32 (approved):
Looks great -- a few very minor comments.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
chrismaree: ye, you right. it does not need it. I'd rather leave this for a separate PR as it will ripple into other changes and this PR is already pretty big.

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
chrismaree: I think we want `.every` as this method [tests](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every) whether all elements in the array pass the test implemented by the provided function. We want to be sure that _all_ decimals within the `MedianizedPrice` feed match their precision else it will break the downstream computation. This syntax is saying that `every` position within the array should equal the 0th element within the array. How would you want to structure it with `.any`?

Regarding your second point, I agree that will help with readability. I've changed it accordingly.

mrice32: Sorry, I should've been clearer -- I think the reason it should be `any` is because you're checking for the error case, not the success case in this _if_. I would agree with everything you said above if you were checking for === in the condition, but you are looking for "any" that do _not_ match your desired pattern if I'm reading the code correctly.

mrice32: To frame it differently, i think in its current form, this if will _always_ return false and therefore never hit the `throw` because you're effectively doing `if (priceFeedDecimals[0] !== priceFeedDecimals[0] && ... )` which will always short-circuit and return false.

chrismaree: you are 100% right, sorry for the confusion. I've updated it accordingly and added and added a test to validate the behavior with different feeds.

packages/common/src/FormattingUtils.js
nicholaspai: Do we need to handle the case where `fromDecimals > toDecimals`?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: We should do the same for BCHNBTC right?

packages/liquidator/index.js
```diff
@@ -132,26 +131,37 @@ async function run({
...
+      syntheticTokenDecimals,
+      priceFeed.getPriceFeedDecimals()
```
nicholaspai, line 162: So currently we hardcode that every type of `PriceFeed` returns 18 for `getPriceFeedDecimals`, but imagine that we want to change this to 8 (for some reason) for one EMP.

Wouldn't it be easier to make this change if set the decimals to 8 within this `index.js` entrypoint, and simultaneously passed it to the `PriceFeed` and the `ExpiringMultiPartyClient`?

My main point is that if we want to change the `PriceFeed`'s decimals to something other than 18, then how can we dynamically change it for any EMP config.

nicholaspai (commented):
Looks good but I am not sure that setting the `pricefeed decimals` WITHIN the `PriceFeed` classes makes the most sense, I still think it should be configured at the same level where an EMP's config is detected and read (i.e. within the `index.js` entrypoint)

mrice32 (approved):
LGTM!

packages/liquidator/index.js
chrismaree: So the way this works is that the price feed itself contains a `priceFeedDecimals` that defines how to scale the price feed. this can be parameterized within the json or is added to the default config by re-using the method from common.

This is auto-detected based on the identifier or is parameterizable by the user.

i.e this _is_ set within the `index.js` file entry point and is fed downstream to all associated modules and clients.

Does that make sense?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -92,4 +94,10 @@ const defaultConfigs = {
   }
 };
 
+// Pull in the number of decimals for each identifier from the common getPrecisionForIdentifier. This is used within the
```
chrismaree, line 97: @nicholaspai see here. I add the `priceFeedDecimals` from your default scaling from common for all price feeds. In the future, we will deprecate this when we move to all 18 decimal feeds but for now it keeps one single source of truth for all scaling between the bots and the voter dapp.

packages/liquidator/index.js
nicholaspai: Yep makes sense

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -1,5 +1,5 @@
 const { PriceFeedInterface } = require("./PriceFeedInterface");
-const { parseFixed } = require("@ethersproject/bignumber");
+const { parseFixed } = require("@uma/common");
```
nicholaspai, line 2: Good catch, didn't realize we exported this from `common`

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -514,6 +514,51 @@ contract("CreatePriceFeed.js", function(accounts) {
...
+    };
+
+    let emp = await ExpiringMultiParty.new(constructorParams);
+
+    // Should create a valid price feed with no config.
+    const priceFeed = await createReferencePriceFeedForEmp(logger, web3, networker, getTime, emp.address, {
+      minTimeBetweenUpdates: 5
+    });
+
+    assert.isTrue(priceFeed != null);
```
nicholaspai, line 556:
```suggestion
    assert.isTrue(priceFeed !== null);
```

nicholaspai: I think you can also use `assert.exists`

nicholaspai (approved):
I left a couple more comments but LGTM after you resolve

packages/common/src/FormattingUtils.js
chrismaree: The util does this already :)

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
chrismaree: done

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
chrismaree: I think we get this for free from our linter. If you try and make invalid JSON within this file husky will reject it. If you get it past husky then CI will reject it on the linter.

See the screenshot below which shows how husky detects this:
image

nicholaspai: Oh for real! Great

PR #2278, opened 2020-12-09
feat(fee-payer): Add gulp method to distribute excess collateral to all sponsors

### LINKED ISSUE #2277: Re-add way for Perpetual to remove non-PfC collateral
**Why**

`trimExcess()` was removed in #2212  to save bytecode, this is important functionality that prevents collateral from being locked forever that is sent to the contract.

**How**

Less important bytecode will have to be reduced elsewhere.

- 5: ~2 hours

nicholaspai:
**Motivation**

#2277 

**Details**

I removed several require messages to make up for increased bytecode, contract's have < 50 bytes remaining in this PR. The contract reverts in these require statements for reasons that are relatively obvious to the user, such as depositing 0 tokens, or withdrawing more than their collateral amount.

The Perpetual still has no way of removing non-collateral tokens sent to the contract FYI.

**Issue(s)**

Fixes #2277

### DIFF
--- a/packages/core/contracts/financial-templates/common/FeePayer.sol
+++ b/packages/core/contracts/financial-templates/common/FeePayer.sol
@@ -158,6 +158,15 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
         return _pfc();
     }
 
+    /**
+     * @notice Removes excess collateral balance not counted in the PfC by distributing it out pro-rata to all sponsors.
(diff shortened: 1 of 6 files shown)


mrice32 (approved):
Looks good!

mrice32 (commented):
Really sucks we have to keep getting rid of require messages :/. At some point we might consider stripping them out at the compilation level to get rid of the OZ library ones, too.

PR #2280, opened 2020-12-09
fix(funding-rate-applier): reset proposal time to 0 if proposal doesn't exist

### LINKED ISSUE #2279: FundingRateApplier should "forget" about proposals that the Oracle does not know about
**Description**
If the OptimisticOracle (OO) is upgraded in the `Finder` while a price request is unresolved, then the FundingRateApplier (FRA) will forever experience reverts on `getPrice()` and have no way to know that it should allow for new funding rate proposals. The current funding rate will therefore be frozen forever.

How this problem arises:
1. Normally when the FundingRateApplier (FRA) proposes a price (i.e. a user calls `proposeNewRate`) to the OptimisticOracle (OO), it first requests a price of the OO and then proposes a price for that price request, which sets its own address as the `proposer` of the request. Finally, it stores the current time as the `proposalTime`. 
2. When `proposalTime !== 0`, the FRA will attempt to call the OO's `getPrice` method, which will revert until the FRA's proposal to the OO passes liveness. If `getPrice` reverts, the FRA will do an additional check to see if the OO's proposal was disputed. If the proposal was disputed OR the proposal passed liveness, then the FRA will reset its `proposalTime` to 0.
- When `proposalTime == 0`, new proposals can be made.

If the OO is upgraded in the Finder after step (1) but before its proposal passes liveness or gets disputed, then the the new OO will always revert on `getPrice()` AND it will not show that the proposal was disputed. Therefore, the FRA will never reset `proposalTime` to 0 and the current funding rate will not be updateable.

Solution:
- Add a check in the FRA for if the OO actually knows about the pending proposal. A simple check is if the proposal's `proposer` is the FRA's address. If the OO was upgraded while the price request was still pending, then the new OO would not have a proposal entry for this FRA address!
- This check should reset the proposalTime to 0.

**Reproduction**

0. Propose a funding rate to the FRA
1. Create a new OO and register it in the Finder
2. Try to propose a new funding rate to the FRA.

(2) will fail because the FRA will not be able reset the proposal time to 0, and moreover the OO will have no way to resolve this price request because it will not have been requested on the upgraded OO,

nicholaspai:

**Motivation**

#2279 

**Issue(s)**

Fixes #2279


packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -196,11 +196,32 @@ contract("FundingRateApplier", function(accounts) {
     assert.equal(test4[1].rawValue, toWei("0"));
   });
 
-  it("Initial 0 funding rate", async () => {
+  it("Initial funding rate struct is correct", async () => {
```
nicholaspai, line 199: This test is important because the `updateTime` and `applicationTime` are set on construction to the deployment time, we should explicitly check for this.

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -196,11 +196,32 @@ contract("FundingRateApplier", function(accounts) {
...
+  });
+  it("Calling applyFundingRate without a pending proposal does not change multiplier", async () => {
     await fundingRateApplier.setCurrentTime(startTime + 1000);
-    await fundingRateApplier.applyFundingRate();
+    const receipt = await fundingRateApplier.applyFundingRate();
     assert.equal((await fundingRateApplier.fundingRate()).cumulativeMultiplier.rawValue.toString(), toWei("1"));
     assert.equal((await fundingRateApplier.fundingRate()).rate.rawValue.toString(), "0");
+
+    // A NewFundingRateMultiplier is emitted.
+    truffleAssert.eventEmitted(receipt, "NewFundingRateMultiplier", ev => {
```
nicholaspai, line 215: This test is important because it makes it obvious that `NewFundingRateMultiplier` is emitted even if the funding rate is unchanged. Open question: Feature or bug?

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -238,10 +259,6 @@ contract("FundingRateApplier", function(accounts) {
     });
 
     it("Two proposals cannot coexist", async () => {
-      // Move time forward a small amount to not expire the previous one, but to give room for a new proposal.
```
nicholaspai, line 241: This was an unnecessary time jump

packages/core/test/financial-templates/common/FundingRateApplier.js
mrice32: IMO, this event is really just saying "FundingRateApplied", so it's kinda a weird name (even though it is technically changing the funding rate multiplier).

It gets emitted effectively every time the contract is touched. Does that mean it's not useful and a bad use of users' gas? I could see that argument.

Another question is whether we should have an event for when the funding rate is updated after the getPrice method returns. Something like "FundingRateUpdated" that just gives the new funding rate?

IMO, we should weigh these concerns against the cost of making changes this late.

Thoughts?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -184,7 +184,7 @@ contract PerpetualPositionManager is FundingRateApplier {
         fees()
         nonReentrant()
     {
-        require(collateralAmount.isGreaterThan(0), "Invalid collateral amount");
+        require(collateralAmount.isGreaterThan(0));
```
mrice32, line 187: Are these the same require messages that were removed in #2278? Will that mean that once these get merged on top of one another, we'll still be over the size limit?

mrice32 (commented):
Looks awesome! A few minor comments and nits.

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: Yes so I merged #2278 and now this PR is 150 bytes over.

packages/core/test/financial-templates/common/FundingRateApplier.js
nicholaspai: I don't think its a bug as much as an artifact of bad naming. I do think we should deprecate `NewFundingRateMultiplier` in favor of a `FundingRateUpdated` event and will address in a follow up PR: #2282

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -248,9 +248,15 @@ abstract contract FundingRateApplier is FeePayer {
...
-                // Stop tracking if in dispute to allow other proposals to come in.
+                // Stop tracking and allow other proposals to come in if:
+                // - The requester address is empty, indicating that the Oracle does not know about
+                //   this funding rate request. This is possible if the Oracle is replaced while the price request
+                //   is still pending.
+                // - The request has been disputed.
                 if (
                     optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData).disputer !=
+                    address(0) ||
+                    optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData).proposer ==
```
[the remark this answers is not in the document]
nicholaspai: Done

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -217,18 +217,17 @@ abstract contract FundingRateApplier is FeePayer {
         uint256 proposalTime = fundingRate.proposalTime;
         if (proposalTime != 0) {
             // Attempt to update the funding rate.
-            OptimisticOracle optimisticOracle = _getOptimisticOracle();
+            OptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
             bytes32 identifier = fundingRate.identifier;
             bytes memory ancillaryData = _getAncillaryData();
 
             // Try to get the price from the optimistic oracle.
+            OptimisticOracleInterface.Request memory request =
```
nicholaspai, line 225: @mrice32 We can refactor this `getRequest` call above this try-catch since its called no matter what with the same params. This saves 100 bytes of bytecode

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -248,11 +247,12 @@ abstract contract FundingRateApplier is FeePayer {
                 // Set proposal time to 0 since this proposal has now been resolved.
                 fundingRate.proposalTime = 0;
             } catch {
-                // Stop tracking if in dispute to allow other proposals to come in.
-                if (
-                    optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData).disputer !=
-                    address(0)
-                ) {
+                // Stop tracking and allow other proposals to come in if:
```
chrismaree, line 250: +1

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -15,7 +15,7 @@ import "../../oracle/interfaces/FinderInterface.sol";
 import "../../oracle/implementation/Constants.sol";
 
 // TODO: point this at an interface instead.
-import "../../oracle/implementation/OptimisticOracle.sol";
+import "../../oracle/interfaces/OptimisticOracleInterface.sol";
```
chrismaree, line 18: did this change save any bytecode?

chrismaree (approved):

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: 100 bytes I believe!

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -217,18 +217,17 @@ abstract contract FundingRateApplier is FeePayer {
...
             // Attempt to update the funding rate.
-            OptimisticOracle optimisticOracle = _getOptimisticOracle();
+            OptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
             bytes32 identifier = fundingRate.identifier;
             bytes memory ancillaryData = _getAncillaryData();
 
             // Try to get the price from the optimistic oracle.
+            OptimisticOracleInterface.Request memory request =
+                optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData);
             try optimisticOracle.getPrice(identifier, proposalTime, ancillaryData) returns (int256 price) {
```
daywiss, line 227: its worth having a comment here i think, the logic in getPrice is really complicated to trace. Basically it will throw if there is no price request, or the price request response is being disputed. This causes an error in settlement as its unable to resolve the proposal and throws.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -248,11 +247,12 @@ abstract contract FundingRateApplier is FeePayer {
...
-                if (
-                    optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData).disputer !=
-                    address(0)
-                ) {
+                // Stop tracking and allow other proposals to come in if:
+                // - The requester address is empty, indicating that the Oracle does not know about
+                //   this funding rate request. This is possible if the Oracle is replaced while the price request
+                //   is still pending.
+                // - The request has been disputed.
+                if (request.disputer != address(0) || request.proposer == address(0)) {
```
daywiss, line 255: what are your thoughts on comparing this to `oo.getState` where `state != invalid or state != Settled` or something to that effect

daywiss (commented):
I happened to be going through this code yesterday so this is an opportunity to ask some questions :)

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -217,18 +217,17 @@ abstract contract FundingRateApplier is FeePayer {
         uint256 proposalTime = fundingRate.proposalTime;
         if (proposalTime != 0) {
```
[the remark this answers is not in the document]
nicholaspai: Do you think your refactor is more readable?

Regarding bytecode size you're right, it costs 40 bytes more to duplicate the return call

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -248,11 +247,12 @@ abstract contract FundingRateApplier is FeePayer {
...
-                // Stop tracking if in dispute to allow other proposals to come in.
-                if (
-                    optimisticOracle.getRequest(address(this), identifier, proposalTime, ancillaryData).disputer !=
-                    address(0)
-                ) {
+                // Stop tracking and allow other proposals to come in if:
+                // - The requester address is empty, indicating that the Oracle does not know about
+                //   this funding rate request. This is possible if the Oracle is replaced while the price request
+                //   is still pending.
+                // - The request has been disputed.
```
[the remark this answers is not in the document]
nicholaspai: Sorry can you write an example of this? I'm not sure what you're asking

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: This is an interesting idea. I think comparing the state would be more readable, but would add more bytes because we'd still have to make the `getRequest` call to determine who the original proposer was (to potentially pay rewards to). WDYT?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: [Comments in this commit](https://github.com/UMAprotocol/protocol/pull/2280/commits/6d62763a2773e9ef5e62c3481ff530d9fa5867f3)! Thanks @daywiss feedback about the readability is really useful.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
daywiss: If someone proposes a new funding rate at a certain timestamp and it gets disputed, a new funding rate cannot be submitted for that same timestamp. Im not 100% sure if this is true but is this a problem?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
daywiss: yea that is really interesting. I do think its more readable to return early when you can, but there are other important dimensions to optimize (gas and bytes), which is hard to make the call on. I would say leave it as is

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
nicholaspai: Proposer 1 proposes a funding rate at time 1
Disputer disputes proposal at time 2, a price request is submitted to the OO for time 1
FundingRateApplier now allows new price requests to come in for any time >= 2**

** There is a configuration option to allow for some "lookback buffer" that allows price requests up to 30 mins in the past

I don't think this is a problem though

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
daywiss: yea again easier to read but less optimized! This really should be directed at the OO interface to include current state on the request. I think leave it as is.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
mrice32: @daywiss I agree that returning early is more readable, but I would personally prefer the 40 bytes :)

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
mrice32: @daywiss, I think this is a good point. Arguably, in a perfect world, the contract would just take the OO price at face value and ignore the proposal in this case.

Practically, the only exploit I see is where someone "fills up" the proposal space with requests by sending many requests and disputes, each for a different second to block new funding rates. However, I think there are cheaper ways to do this (namely by submitting bad funding rates and letting someone else dispute you). Arguably, though, if the final fee were low enough (or 0), this could be a problematic exploit. However, I'd argue that the sheer amount of gas to send `3600` requests every hour seems prohibitively expensive.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: @nicholaspai I worry about edge cases here because `getPrice` changes the state of the request, which could change the result of this call. This may be a non-issue, but it might be worth thinking through the implications.

mrice32 (approved):
Looks good! Before merging, could you take a minute to think through [this comment](https://github.com/UMAprotocol/protocol/pull/2280/files#r541003274)?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Hm this is a really good point. It's not an issue right now but I could imagine upgrading the OO in the future and failing to take this state modification into account, because `getPrice` is confusingly NOT a read only call. I moved the `getRequest` call into both the body of the try and the catch

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
nicholaspai: Same, sadly.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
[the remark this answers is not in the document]
nicholaspai: Yeah I think the final fee is meant to deter this type of attack.

nicholaspai:
> Looks good! Before merging, could you take a minute to think through [this comment](https://github.com/UMAprotocol/protocol/pull/2280/files#r541003274)?

I'm not sure if the comment that you are referring to linked properly but are you referring to [this one](https://github.com/UMAprotocol/protocol/pull/2280#discussion_r541003274)?

mrice32:
> > Looks good! Before merging, could you take a minute to think through [this comment](https://github.com/UMAprotocol/protocol/pull/2280/files#r541003274)?
> 
> I'm not sure if the comment that you are referring to linked properly but are you referring to [this one](https://github.com/UMAprotocol/protocol/pull/2280#discussion_r541003274)?

Yep, I was.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
mrice32: Sounds good!

mrice32 (approved):
LGTM!

PR #2283, opened 2020-12-09
feat(perp-position-manager): Add trim excess 

### LINKED ISSUE #2281: Re-add way for Perpetual to remove non-PfC non-collateral tokens from balance
**Why**

Related to #2277 but handles the case where the token balance is not the same as the collateral currency of the contract. This method should be like `trimExcess` in the EMP and just send tokens to some beneficiary address.

**How**

This will r

**Resolution**

What is the deliverable to consider this task done?

**Difficulty Score [1-10]**

- 4: ~2 hours

nicholaspai:

**Motivation**

#2281

This PR ensures that no ERC20 tokens that are not included in the `PfC` will be stuck within the Perp. `trimExcess` can also be used to drain excess collateral arising from precision loss. 

Moreover, this adds a note to `gulp()` that it will revert if `PfC` is 0. I don't think we need to specially handle this case because no other methods are dependent on `_gulp()` succeeding.

**Issue(s)**

Fixes #2281

### DIFF
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
@@ -74,6 +74,9 @@ contract PerpetualPositionManager is FundingRateApplier {
     // Expiry price pulled from the DVM in the case of an emergency shutdown.
     FixedPoint.Unsigned public emergencyShutdownPrice;
 
+    // The excessTokenBeneficiary of any excess tokens added to the contract.
+    address public excessTokenBeneficiary;
+
     /****************************************
      *                EVENTS                *
      ****************************************/
@@ -134,6 +137,7 @@ contract PerpetualPositionManager is FundingRateApplier {
      * @param _fundingRateIdentifier Unique identifier for DVM price feed ticker for child financial contract.
      * @param _minSponsorTokens minimum amount of collateral that must exist at any time in a position.
      * @param _tokenScaling initial scaling to apply to the token value (i.e. scales the tracking index).
+     * @param _excessTokenBeneficiary Beneficiary to which all excess token balances that accrue in the contract can be sent.
      * @param _timerAddress Contract that stores the current time in a testing environment. Set to 0x0 for production.
      */
     constructor(
@@ -146,6 +150,7 @@ contract PerpetualPositionManager is FundingRateApplier {
         FixedPoint.Unsigned memory _minSponsorTokens,
         address _configStoreAddress,
         FixedPoint.Unsigned memory _tokenScaling,
+        address _excessTokenBeneficiary,
         address _timerAddress
     )
         public
@@ -164,6 +169,7 @@ contract PerpetualPositionManager is FundingRateApplier {
         tokenCurrency = ExpandedIERC20(_tokenAddress);
         minSponsorTokens = _minSponsorTokens;
         priceIdentifier = _priceIdentifier;
+        excessTokenBeneficiary = _excessTokenBeneficiary;
     }
 
     /****************************************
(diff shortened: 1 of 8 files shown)


packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -524,6 +530,25 @@ contract PerpetualPositionManager is FundingRateApplier {
...
+     * @dev This will drain down to the amount of tracked collateral and drain the full balance of any other token.
+     * @param token address of the ERC20 token whose excess balance should be drained.
+     */
+    function trimExcess(IERC20 token) external fees() nonReentrant() returns (FixedPoint.Unsigned memory amount) {
+        FixedPoint.Unsigned memory balance = FixedPoint.Unsigned(token.balanceOf(address(this)));
+
+        if (address(token) == address(collateralCurrency)) {
+            // If it is the collateral currency, send only the amount that the contract is not tracking.
+            // Note: this could be due to rounding error or balance-changing tokens, like aTokens.
+            amount = balance.sub(_pfc());
```
daywiss, line 544: is it the case that pfc is essentially all the collateral held by sponsors?

daywiss (approved):

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
nicholaspai: Correct

chrismaree (approved):
Looks good to me! Having both this and gulp is nice; I'm of the opinion that we really only need gulp but no real downsides in including this.

mrice32 (commented):
High-level comment: it's not clear to me that this addition is worthwhile if we have gulp, especially since this means that the beneficiary could ultimately compete with sponsors to call this before gulp is called. Is the primary argument that it can siphon non-collateral tokens? Is that use case worth the additional constructor arg and the complexity of giving a single address certain privileges? Have we seen a demonstrated need for siphoning off non-collateral tokens in the past?

Thoughts @nicholaspai @daywiss @chrismaree?

daywiss:
You are right @mrice32 , there's a race between gulp and trim excess when dealing with collateral. I doubt there is a precedent for people sending non collateral to contracts, which would be the motivation for trim excess. It seems though like a relatively benign, low risk addition. I guess it makes me feel better if for some reason if there was a reason for us to recover funds due to someones error, but im not sure that is a strong enough reason to include it.

nicholaspai:
Yeah I agree with @daywiss that trimExcess can be used as an insurance policy in case anyone does send non collateral tokens to it. I also wanted to be consistent with the EMP's interface which does include this method. I propose either removing from both the Perp and EMP or having it in both (I.e merging this PR). Thoughts @mrice32 ?

mrice32:
> Yeah I agree with @daywiss that trimExcess can be used as an insurance policy in case anyone does send non collateral tokens to it. I also wanted to be consistent with the EMP's interface which does include this method. I propose either removing from both the Perp and EMP or having it in both (I.e merging this PR). Thoughts @mrice32 ?

This is a really good point (unifying the two interfaces where possible). Personally, my preference would be to remove trim excess from both contracts, leaving only gulp.

nicholaspai:
> > Yeah I agree with @daywiss that trimExcess can be used as an insurance policy in case anyone does send non collateral tokens to it. I also wanted to be consistent with the EMP's interface which does include this method. I propose either removing from both the Perp and EMP or having it in both (I.e merging this PR). Thoughts @mrice32 ?
> 
> This is a really good point (unifying the two interfaces where possible). Personally, my preference would be to remove trim excess from both contracts, leaving only gulp, but curious to hear everyone else's thoughts. Since adding it, I've always felt like trimExcess was a somewhat inelegant way to solve the problem. gulp seems like a solution that's more consistent with our general philosophy.

`gulp` is totally better for `collateral` units but we should just be on the same page that we are OK never having access to non-collateral tokens sent to the contract.

nicholaspai:
Closing in favor of #2293 which removes `trimExcess` from the EMP instead of adding it to the Perp.

PR #2284, opened 2020-12-09
improve(funding-rate-applier): Replace NewFundingRateMultiplier event with FundingRateUpdated

### LINKED ISSUE #2282: Remove NewFundingRateMultiplier event and replace with FundingRateUpdated
**Why**

The original aim of `NewFundingRataeMultiplier` was to allow indexers to know when the funding rate was updated, not the multiplier. The multiplier can be queried anytime as an external read-only method.

- 2: ~30 minutes

nicholaspai:

**Motivation**

#2282

**Issue(s)**

Fixes #2282

### DIFF
--- a/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
+++ b/packages/core/contracts/financial-templates/common/FundingRateApplier.sol
@@ -68,14 +68,7 @@ abstract contract FundingRateApplier is FeePayer {
      *                EVENTS                *
      ****************************************/
 
-    event NewFundingRateMultiplier(
-        uint256 indexed newMultiplier,
-        uint256 lastApplicationTime,
-        uint256 applicationTime,
-        uint256 indexed paymentPeriod,
-        int256 indexed latestFundingRate,
-        int256 periodRate
-    );
+    event FundingRateUpdated(int256 newFundingRate, uint256 indexed updateTime);
 
     /****************************************
      *              MODIFIERS               *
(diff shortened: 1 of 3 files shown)


packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -68,14 +68,7 @@ abstract contract FundingRateApplier is FeePayer {
...
 
-    event NewFundingRateMultiplier(
-        uint256 indexed newMultiplier,
-        uint256 lastApplicationTime,
-        uint256 applicationTime,
-        uint256 indexed paymentPeriod,
-        int256 indexed latestFundingRate,
-        int256 periodRate
-    );
+    event FundingRateUpdated(int256 newFundingRate, uint256 indexed updateTime);
```
daywiss, line 71: I wonder if we should also have an event which emits the new cumulative multiplier. I cant really forsee how that would be useful though. I imagine if some offchain infra was using events to recreate historical contract state maybe this one event is enough to do that. thoughts?

daywiss (approved):

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Yeah I think its sufficient to just read the cumulative multiplier on-chain. I don't think there's a need to see historical multipliers, but you should always know the current multiplier before modifying a position.

mrice32 (approved):
LGTM!

PR #2285, opened 2020-12-10
feat(liquidator,disputer,monitor,financial-templates-lib) move to using hardhat for testing

### LINKED ISSUE #2292: Upgrade packages to use hardhat for testing where posible
**Why**

Truffle is extremely slow in testing when compared to hardhat. As of the latest versions of hardhat, much of the previous issues we were experiencing have been resolved. Ideally, we should be able to move many of these tests over to hard hat with minimal code changes required to drastically speed up CI.

**How**

Update packages and tests.

**Resolution**

Packages and tests all run with hardhat, where possible. not all packages will be supported, and that's fine for now.

**Difficulty Score [1-10]**

- 7: ~3 days

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2292

The primary goals are to speed up CI and running tests locally.

**Summary**

This PR rippled through the following packages:
- liquidator: updated tests & package.json
- disputer: updated tests & package.json
- monitor: updated tests & package.json
- financial-templates-lib: updated tests & package.json
- core: removed buidler config and moved to use a `common` implementation
- common: added common `HardhatConfig.js` which is re-used by the other packages.

Additionally, this PR upgrades buidler to hardhat. This change in particular was pretty painful as there was a few breaking changes in a number of the plugins we use (looking at you `@nomiclabs/hardhat-etherscan`).

**Details**

The design & implementation follows closely to the `TruffleConfig.js` wherein a common implementation is placed within the `common` package under a `HardhatConfig.js` file which provides a `getHardhatConfig` method which is re-used within all sub-packages to get the hardhat config.

**Speed up results**
Running tests locally is MUCH faster. For example, previously running the `liquidator` tests locally would take ~30 mins to complete on my 2019 i9 Macbook pro. this is simply running `yarn truffle test ./test/Liquidator.js --network test` from the liquidator package. Now, running the same set of tests, takes only 3 mins. This should help hugely with local development experience and adding new features in the bots.

Running tests in CI is measurably faster. Previously the `test` job would take on average 50 mins over the last 145 runs. See the screenshot below of the Insights into circleCI duration for this test run:
image

After this PR the `test` job takes on average 21 mins on the 3 tests that I've run. 

image

Note that this is not the same 10x speedup as seeing in local env for a few main reasons:
- there are still a few packages with truffle test (`affiliates` & `serverless-orchestration` in particular).
- I think that hardhat is re-compiling the contracts within each package run. This would add about 2 mins per package in CI.
- Truffle also looks to be re-compiling contracts which also adds a few mins.
- We are not (yet) taking advantage of concurrency in unit tests. This we should be able to do pretty easily as hard hat runs an in-process EVM & so we can have isolated test instances between packages.

Lastly, this PR speed up build time by about 5x vs the previous build time in the `build` job.

**Still to be done on the journey to fast CI**
- Update the remaining packages to use hardhat. This is left for a separate PR as this is a bit complex for some of the packages.
- Make coverage optional in the CI pipeline!!! Coverage takes ~40 mins in CI. if we removed it from being a strict requirement from merging PRs we'd still be ensuring tests pass (with the tests job) but would be able to merge in much faster.

**Issue(s)**

close #2292


packages/core/contracts/oracle/implementation/test/OptimisticRequesterTest.sol
```diff
@@ -1,4 +1,7 @@
 // SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
```
chrismaree, line 3: was missing version and this threw off hardhat

packages/core/package.json
```diff
@@ -54,12 +49,12 @@
...
   "scripts": {
-    "test": "yarn mocha-test && yarn buidler-test",
-    "buidler-test": "node --max-old-space-size=1024 $(yarn bin buidler) test",
+    "test": "yarn mocha-test && yarn hardhat-test",
+    "hardhat-test": "node --max-old-space-size=1024 $(yarn bin hardhat) test",
     "mocha-test": "mocha mocha-test --recursive",
     "load-addresses": "yarn run apply-registry",
     "clean": "rm -rf build",
-    "build": "yarn truffle compile && yarn load-addresses",
+    "build": "yarn truffle compile && yarn load-addresses && yarn hardhat compile",
```
chrismaree, line 57: this additional `yarn hardhat compile` might not be required but I just added it for completeness.

packages/core/package.json
```diff
@@ -11,15 +11,10 @@
   "devDependencies": {
     "@awaitjs/express": "^0.3.0",
     "@ethersproject/units": "^5.0.3",
-    "@nomiclabs/buidler": "^1.3.8",
```
chrismaree, line 14: these packages are removed from core as they are not part of common and as core has a dependency on common it gets all the required hardhat packages for free.

packages/disputer/hardhat.config.js
```diff
@@ -0,0 +1,17 @@
+const { getHardhatConfig } = require("@UMA/common");
```
chrismaree, line 1: this shows how easily each package can pull its hard hat config from common. all the common configs and tasks come along for free and we simply override the relevant components for this package.

packages/disputer/test/Disputer.js
```diff
@@ -8,7 +8,7 @@ const winston = require("winston");
 const sinon = require("sinon");
 const { parseFixed } = require("@ethersproject/bignumber");
 
-const { toWei, toBN } = web3.utils;
+const { toWei, toBN, utf8ToHex } = web3.utils;
```
chrismaree, line 11: this `utf8ToHex` was made to the tests that I had to update for consistency.

packages/disputer/index.js
```diff
@@ -52,8 +52,13 @@ async function run({
     // Setup web3 accounts and network
     const [accounts, networkId] = await Promise.all([web3.eth.getAccounts(), web3.eth.net.getId()]);
 
-    // Setup contract instances. NOTE that getAddress("Voting", networkId) will resolve to null in tests.
-    const voting = new web3.eth.Contract(getAbi("Voting"), getAddress("Voting", networkId));
+    // Setup contract instances. NOTE getAddress("Voting", networkId) will error if run in tests. Catch & set it to null.
```
chrismaree, line 55: Previously voting would simply return `null` when running the `index.js` unit tests for the monitor, liquidator and disputer. However, when run with hardhat, this throws an error that breaks the tests. The easiest solution is to just try & catch the response of this call.

packages/disputer/test/Disputer.js
```diff
@@ -90,18 +92,25 @@ contract("Disputer.js", function(accounts) {
         await collateralToken.mint(sponsor3, convert("100000"), { from: contractCreator });
         await collateralToken.mint(liquidator, convert("100000"), { from: contractCreator });
         await collateralToken.mint(disputeBot, convert("100000"), { from: contractCreator });
+
+        // Create identifier whitelist and register the price tracking ticker with it.
+        identifierWhitelist = await IdentifierWhitelist.new();
```
chrismaree, line 97: This basically shows the main changes that the tests required: removing all `.deployed` syntax and replacing it with actual deployed contract instances for each instance.

daywiss: looks like for the most part the api is the same, you just have to instantiate with the new function?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -300,6 +300,16 @@ async function createReferencePriceFeedForEmp(logger, web3, networker, getTime,
     if (!combinedConfig) {
       throw new Error("createReferencePriceFeedForEmp: No default config was found and no user config was provided.");
     }
```
chrismaree, line 302: This change is made to be compatible with how hardhat treats time within it's test network. Each time a block is mined hard hat's internal clock of time moves forward. This cant be disabled. This breaks a bunch of bot logic and price feeds as they rely on the internal time of the bot and relies on that being close to the passage of block time. To provide compatibility, this override is added to enable use the block time as the bot's clock.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -81,7 +81,11 @@ class UniswapPriceFeed extends PriceFeedInterface {
     });
 
     // Search backwards through the array and grab block timestamps for everything in our lookback window.
-    const currentTime = this.getTime();
+    // Get Time can either be a synchronous OR asynchronous function depending on how the UniswapPriceFeed is setup.
```
chrismaree, line 84: This enables the bots clock within this feed to use EITHER the `getTime` synchronous or `getTime` asynchronous methods to accommodate the override in the `CreatePriceFeed.js`.

packages/common/src/HardhatConfig.js
```diff
@@ -0,0 +1,234 @@
+function getHardhatConfig(configOverrides) {
```
chrismaree, line 1: This file is mostly the same as the previous buidler config. Main changes are updating from buidler to hardhat and all the associated import changes thereof.

packages/monitors/index.js
```diff
@@ -56,6 +56,7 @@ async function run({
   tokenPriceFeedConfig,
   medianizerPriceFeedConfig
 }) {
+  console.log("top");
```
daywiss, line 59: stray comment

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -43,8 +43,9 @@ spoke.post("/", async (req, res) => {
             : JSON.stringify(req.body.environmentVariables[key]);
       });
     }
-
+    console.log("EXEC");
     const execResponse = await _execShellCommand(req.body.serverlessCommand, processedEnvironmentVariables);
+    console.log("After exec");
```
daywiss, line 48: some strays maybe here

mrice32:
👏 very excited to review!

packages/common/src/HardhatConfig.js
daywiss, line 1: :open_mouth: :+1:  you are a hardhat expert now

daywiss (approved):
im not in a great position to critique this, but well done! i can probably tackle updating the affiliate test update once this is merged. made an issue [here](https://github.com/UMAprotocol/protocol/projects/6#card-51497164)

packages/serverless-orchestration/src/ServerlessSpoke.js
chrismaree: good catch! have removed.

nicholaspai (commented):
Drive by:

- Why is `HardHatConfig.js` in the `common/` folder whereas `buidlerConfig.js` was in the `core/` folder before?

packages/disputer/index.js
mrice32, line 55: Hmmm, this seems bad that we're not testing our bots' ability to grab our contracts from the artifacts.... I don't see a straightforward solution, though, other than to not rely on addresses in the artifacts at all (which seems like a nonstarter). Any ideas?

We could have node tests that run against infura that specifically test this, but `try-catching` this in the local tests doesn't seem to be scalable for infra that more heavily relies on global contracts.

One really nasty way of doing this is to do some sort of in-memory mocking or modification of the buidler artifacts. I think this is something we'll need a concrete solution for -- deploying local versions of global contracts at the start of the test and accessing them through artifacts will be a must-have I suspect.

mrice32 (commented):
Looks great! A few minor questions

packages/disputer/test/Disputer.js
chrismaree: yup. using the `new` keyword is deploying the contract here rather than relying on the migrations with the `.deployed` keyword.

packages/common/package.json
```diff
@@ -26,16 +26,23 @@
     "@ethersproject/bignumber": "^5.0.5",
     "@google-cloud/kms": "^0.3.0",
     "@google-cloud/storage": "^2.4.2",
+    "@nomiclabs/hardhat-etherscan": "^2.1.0",
+    "@nomiclabs/hardhat-truffle5": "^2.0.0",
+    "@nomiclabs/hardhat-web3": "^2.0.0",
+    "@openzeppelin/contracts": "3.0.0",
```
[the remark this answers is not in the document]
chrismaree: dont think it is, good catch. have removed it.

packages/core/contracts/oracle/implementation/test/OptimisticRequesterTest.sol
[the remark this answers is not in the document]
chrismaree: ye, I dont think it is.

packages/core/test/scripts/DecodeTransactionData.js
```diff
@@ -2,14 +2,16 @@ const DecodeTransactionData = require("../../scripts/DecodeTransactionData");
...
 
 const Registry = artifacts.require("Registry");
-const Voting = artifacts.require("Voting");
 const VotingInterfaceTesting = artifacts.require("VotingInterfaceTesting");
 
 contract("scripts/DecodeTransactionData.js", function() {
+  let registry;
+  before(async function() {
+    registry = await Registry.new();
+  });
```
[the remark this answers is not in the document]
chrismaree: I dont think it has to be! I have reverted it and the tests pass just as they should seeing this is within `core`.

packages/disputer/index.js
chrismaree: yup! I agree it's really not ideal. I was not super happy about this solution.

One solution is to modify the `getAddress` to return `null` as it does with truffle tests so this does not error out. We can then work on fixing this method to work correctly with hardhat by correctly deploying the voting contract within the tests and ensuring that this artifact is pulled in as expected.

The thing about this specific `voting` instance is that it is not tested within these `index.js` tests. these tests are testing the main entry point to ensure that the looping, retry and instantiation logic works. The actual testing of the `disputer.js` & `liquidator.js` tests are not affected by this as the voting contract's address is provided as a param in these tests.

chrismaree: I’ve found a workaround for this. Hardhat’s internal structure is not the same as buidler and so we can't simply pull the address from the global return `global.artifacts._provisioner._deploymentAddresses[contractName];` the way we've been doing it before within the core package. However, a really simple workaround is to just set the addresses we need for this method within the global space within the tests and then they are accessible.

What I’ve done looks like this:
1. within the unit tests deploy the `Voting` contract.
2. within unit tests set `global.hardhatTestingAddresses["Voting"] = voting.address`
3. Change the `getAddressTest` to try to pull this address return `global.hardhatTestingAddresses[contractName]` before erroring out. 

This lets us remove the null logic checks within the bots, which is great.

One small pain point with this is that the bots (liquidator, disputer and monitor) are dependent on `"@uma/core": "npm:@uma/core@1.2.0"`. To test this logic I removed the `npm:@uma/core` part of the dependency so I could use my local package. However, this wont work with the PR as we still need the bots to be dependent on this old package until we’ve updated the bots to work with the new contracts. What we can do therefore is create a new core release that has the old contract interfaces & methods from `1.2.0` but includes the new logic for `hardhatTestingAddresses` within the `getAddressTest` method. This will, however, require a separate PR which includes this 1.2.1 release.

mrice32 (approved):
LGTM! Very awesome stuff!

PR #2287, opened 2020-12-10
fix(voting): Limit length of ancillary data

### LINKED ISSUE #2240: Consider limiting the length of ancillary data
**Why**

Avoid any attacks where ancillary data can be set so long that certain lifecycle steps in a DVM or optimistic oracle request cannot be completed without running over the block gas limit.

**How**

A simple require should suffice.

**Resolution**

A hard limit on the number of bytes is imposed.

**Difficulty Score [1-10]**

- 3: ~1 hour

nicholaspai:

**Motivation**

#2240

**Issue(s)**

Fixes #2240


packages/core/test/oracle/Voting.js
```diff
@@ -1932,6 +1933,89 @@ contract("Voting", function(accounts) {
       price2.toString()
     );
   });
+  it("Stress testing the size of ancillary data", async function() {
+    let identifier = web3.utils.utf8ToHex("stress-test");
+    let time = "1000";
+    const DATA_LIMIT = 8192;
+    let ancillaryData = web3.utils.bytesToHex(crypto.randomBytes(DATA_LIMIT));
```
nicholaspai, line 1940: Does anyone know a better way to create random hex data?

daywiss: thats a pretty good way but i remember seeing this: https://web3js.readthedocs.io/en/v1.2.11/web3-utils.html#randomhex

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -255,6 +256,14 @@ contract Voting is
         require(time <= blockTime, "Can only request in past");
         require(_getIdentifierWhitelist().isIdentifierSupported(identifier), "Unsupported identifier request");
 
+        // As of December 2020, the current Ethereum gas limit is 12.5 million. This function's gas primarily
+        // comes from computing a Keccak-256 hash in _encodePriceRequest and writing a new PriceRequest to
+        // storage. We have empirically determined an ancillary data limit of 8192 bytes that keeps this function
+        // well within the gas limit at ~8 million gas. To learn more about the gas limit and EVM opcode costs go here:
+        // - https://etherscan.io/chart/gaslimit
+        // - https://github.com/djrtwo/evm-opcode-gas-costs
+        require(ancillaryData.length <= 8192, "Invalid ancillary data");
```
daywiss, line 265: we might want to expose somewhere the limit as a public constant so that this check can be done before tx is sent

packages/core/test/oracle/Voting.js
nicholaspai: Oh this is a great find! thanks

packages/core/contracts/oracle/implementation/Voting.sol
nicholaspai: Good idea

nicholaspai: Done

packages/core/test/oracle/Voting.js
nicholaspai: Done

mrice32: Yeah, we use the web3 method in a few older tests to generate random addresses :). My guess is that the web3 method is basically a simple wrapper over what you had originally haha.

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -254,6 +264,7 @@ contract Voting is
         uint256 blockTime = getCurrentTime();
         require(time <= blockTime, "Can only request in past");
         require(_getIdentifierWhitelist().isIdentifierSupported(identifier), "Unsupported identifier request");
+        require(ancillaryData.length <= ancillaryBytesLimit, "Invalid ancillary data");
```
mrice32, line 267: Should we add the same require to the OO?

mrice32 (commented):
Looks great!

daywiss (approved):

packages/core/contracts/oracle/implementation/Voting.sol
nicholaspai: yeah its a bit redundant bytecode-wise but it makes it more readable. Do you think a comment in the OO is not enough?

nicholaspai: Ah yes good catch, done!

nicholaspai:
@daywiss @mrice32 I added a check for the ancillary data in the OO's `requestPrice` method as well and added a test. @chrismaree

mrice32 (approved):
LGTM!

PR #2290, opened 2020-12-11
feat(affiliates)!: add ability to set deployer reward address. 

### LINKED ISSUE #2271: affilates: add custom payout addresses to override deployer address
Currently rewards for dev mining / deployer rewards are paid out to the address which deployed the emp. This can be deduced through event data on chain. It is a pretty strong assumption to pay rewards automatically to that address and most likely should not be the default behavior. We should require that participants provide us a payout address and add functionality to override the deployer address with their preferred payout address.

daywiss:
BREAKING CHANGE: this changes config format and requires user to provider deployer for each emp address

**Motivation**
#2271 

**Summary**

Changes configuration arguments to require a reward address to accompany emp address. This deprecates `getEmpDeployerHistory` which has been moved as a helper utility in `libs/contracts`. Errors will be thrown to alert the user if the config is incorrect. 

Other small updates
- addresses will be converted to eth checksum address so bigquery does not fail
- minimum lookback date is hard coded to JAN 1 2020, so that queries dont have to go back to time 0 when finding events.

**Details**
Previously in order to generate deployer rewards, the empWhitelist would be specified like this: 
image

Now they must be specified like this:
image

**Issue(s)**

Fixes #2271


packages/affiliates/libs/affiliates.js
```diff
@@ -5,17 +5,21 @@ const highland = require("highland");
...
 
-const DeployerRewards = ({ queries, empCreatorAbi, empAbi, coingecko, synthPrices }) => {
+const DeployerRewards = ({ queries, empAbi, coingecko, synthPrices, firstEmpDate }) => {
   assert(queries, "requires queries class");
-  assert(empCreatorAbi, "requires creator abi");
   assert(empAbi, "requires empAbi");
   assert(coingecko, "requires coingecko api");
   assert(synthPrices, "requires synthPrices api");
+
+  // use firstEmpDate as a history cutoff when querying for events. We can safely say no emps were deployeed before Jan of 2020.
```
nicholaspai, line 20: +1

packages/affiliates/libs/affiliates.js
```diff
@@ -5,17 +5,21 @@ const highland = require("highland");
...
+  firstEmpDate = firstEmpDate || moment("2020-01-01", "YYYY-MM-DD").valueOf();
```
nicholaspai, line 21: Do you think there's any reason we should look into allowing the user to configure the `firstEmpDate` or is it fine just hardcoding Jan 1?

nicholaspai (approved):
This is well implemented, my main lingering question is why you are making users specify the deployer reward address now? It does seem more flexible, but is this motivated by performance or something else?

daywiss:
@nicholaspai This is motivated by the mario.cash group, as they have a payout address different from deployer address, so we needed to add some kind of functionality to specify this in rewards script ( or otherwise override it manually).  In general I think we shouldn't assume participants want rewards paid to the deployer address, and there's an opportunity during the whitelisting process where we can have them be explicit about where rewards go.  Other nice side effects of this is that it simplifies the code and queries a lot.

We can continue to discover deployer addresses if we wanted to maintain the old way of rewarding, it would just need to be done as a separate query and provided in the payout config file.

packages/affiliates/libs/affiliates.js
daywiss: Yea thats a good question. I can imagine if you know the start date of the set of EMPs you are rewarding, you could optimize it a lot by setting the start date manually. Will that ever be done in practice though probably not.

nicholaspai:
> @nicholaspai This is motivated by the mario.cash group, as they have a payout address different from deployer address, so we needed to add some kind of functionality to specify this in rewards script ( or otherwise override it manually). In general I think we shouldn't assume participants want rewards paid to the deployer address, and there's an opportunity during the whitelisting process where we can have them be explicit about where rewards go. Other nice side effects of this is that it simplifies the code and queries a lot.
> 
> We can continue to discover deployer addresses if we wanted to maintain the old way of rewarding, it would just need to be done as a separate query and provided in the payout config file.

I agree that the simplicity is awesome. I also think that by setting the fromBlock to this value, we get a huge improvement over the default fromBlock of 0. I actually don't think further increases of the `fromBlock` would even be that much of an improvement. LGTM

packages/affiliates/README.md
```diff
@@ -0,0 +1,15 @@
+# Token Distributions
```
mrice32, line 1: nit: should you link these to particular directories? Either by specifying the names of the dirs inline or by adding a link?

packages/affiliates/apps/DeployerRewards.js
```diff
@@ -36,16 +33,22 @@ async function App(config) {
...
+    rewards.utils.validateEmpInput(empInput);
+    // convert to standard eth checksum address otherwise lookups through BQ or web3 will fail
+    return empInput.map(rewards.utils.toChecksumAddress);
+  });
+
   // get emp info
   const { collateralTokens, collateralTokenDecimals, syntheticTokenDecimals, syntheticTokens } = await Promise.reduce(
     empWhitelist,
-    async (result, address) => {
+    async (result, [address]) => {
```
mrice32, line 51: nit: since this is now dealing with two addresses, it might be clearer to change this variable name to `empAddress`.

packages/affiliates/libs/contracts.js
```diff
@@ -92,10 +92,29 @@ function Emp({ abi = getAbi("ExpiringMultiParty"), web3 } = {}) {
   };
 }
 
+// returns array of tuples [emp address, deployer address]
+// This acted as a discovery function to assign rewards to an address in lieu of explicitly being given one.
+// It is now deprecated we are making a requirement that EMP Deployers provide us an address rewards should go to.
```
mrice32, line 97: Can we just delete this now? Is it still being used?

mrice32 (approved):
Looks great! A few minor nits

packages/affiliates/README.md
chrismaree: ye, there is a lot of stuff we could add to these readmes. We should think of a standard set of information that each package readme should contain and what the best way is to present the info.

chrismaree (approved):
Looks good to me! Totally a fair restriction to expect all dev mining recipients to add a payment address as a requirement.

PR #2293, opened 2020-12-14
improve(fee-payer) Remove trimExcess from EMP and add gulp edge case unit tests

nicholaspai:
**Motivation**

We closed #2283 after deciding that `gulp()` addresses the original problem that `trimExcess()` was supposed to solve, that of frozen collateral currency in the contract. With `gulp()` added, `trimExcess()` only addresses the additional concern of non-collateral currency frozen in the contract, which we deem unnecessary to address. 

For the sake of consistency between EMP and Perp interfaces, this PR removes `trimExcess` from the EMP and adds unit tests highlighting some "edge" cases for `gulp()` functionality.

Additionally, I have highlighted a "flash gulp" known vulnerability that the team has chosen to ignore in [this document](https://docs.google.com/document/d/19-BbHWfbdCRyDCxPTGr8fSla5zKWmqUbmQetkiVB4i4/edit).

### DIFF
--- a/packages/core/contracts/financial-templates/common/FeePayer.sol
+++ b/packages/core/contracts/financial-templates/common/FeePayer.sol
@@ -162,6 +162,7 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
      * @notice Removes excess collateral balance not counted in the PfC by distributing it out pro-rata to all sponsors.
      * @dev Multiplying the `cumulativeFeeMultiplier` by the ratio of non-PfC-collateral :: PfC-collateral effectively
      * pays all sponsors a pro-rata portion of the excess collateral.
+     * @dev This will revert if PfC is 0.
      */
     function gulp() external nonReentrant() {
         _gulp();
--- a/packages/core/contracts/financial-templates/common/FeePayer.sol
+++ b/packages/core/contracts/financial-templates/common/FeePayer.sol
@@ -162,7 +162,7 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
      * @notice Removes excess collateral balance not counted in the PfC by distributing it out pro-rata to all sponsors.
      * @dev Multiplying the `cumulativeFeeMultiplier` by the ratio of non-PfC-collateral :: PfC-collateral effectively
      * pays all sponsors a pro-rata portion of the excess collateral.
-     * @dev This will revert if PfC is 0.
+     * @dev This will revert if PfC is 0 and this contract's collateral balance > 0.
      */
     function gulp() external nonReentrant() {
         _gulp();
--- a/packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
+++ b/packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
@@ -1694,6 +1694,39 @@ contract("PerpetualPositionManager", function(accounts) {
     assert(await didContractThrow(positionManager.redeem({ rawValue: "16" }, { from: sponsor })));
   });
 
+  it("Gulp edge cases", async function() {
+    // Gulp does not revert if PfC and collateral balance are both 0
+    await positionManager.gulp();
+
+    // Send 1 wei of excess collateral to the contract.
+    await collateral.transfer(positionManager.address, "1", { from: sponsor });
+
+    // Gulp reverts if PfC is 0 but collateral balance is > 0.
+    assert(await didContractThrow(positionManager.gulp()));
+
+    // Create a position to gulp.
+    await collateral.approve(positionManager.address, toWei("100000"), { from: sponsor });
+    await tokenCurrency.approve(positionManager.address, toWei("100000"), { from: sponsor });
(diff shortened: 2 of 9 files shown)


packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -1694,6 +1694,39 @@ contract("PerpetualPositionManager", function(accounts) {
...
 
+  it("Gulp edge cases", async function() {
+    // Gulp does not revert if PfC and collateral balance are both 0
+    await positionManager.gulp();
+
+    // Send 1 wei of excess collateral to the contract.
+    await collateral.transfer(positionManager.address, "1", { from: sponsor });
+
+    // Gulp reverts if PfC is 0 but collateral balance is > 0.
+    assert(await didContractThrow(positionManager.gulp()));
```
daywiss, line 1705: Just curious what was the behavior before you reverted with 0 pfc? Also this should probably be duplicated for the EMP, i didnt see this check happening.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -162,6 +162,7 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
      * @notice Removes excess collateral balance not counted in the PfC by distributing it out pro-rata to all sponsors.
      * @dev Multiplying the `cumulativeFeeMultiplier` by the ratio of non-PfC-collateral :: PfC-collateral effectively
      * pays all sponsors a pro-rata portion of the excess collateral.
+     * @dev This will revert if PfC is 0.
```
daywiss, line 165: i did not see the code that does this revert

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
```diff
@@ -1694,6 +1694,39 @@ contract("PerpetualPositionManager", function(accounts) {
...
+
+    // Create a position to gulp.
+    await collateral.approve(positionManager.address, toWei("100000"), { from: sponsor });
+    await tokenCurrency.approve(positionManager.address, toWei("100000"), { from: sponsor });
+    await positionManager.create({ rawValue: toWei("10") }, { rawValue: "20" }, { from: sponsor });
+
+    // Gulp does not do anything if the intermediate calculation (collateral balance / PfC) has precision loss.
```
daywiss, line 1712: Nice test, this is really interesting. If I understand right the multiplier wont change unless the excess collateral is significant enough in size? Its possible that there could be some dust left in the contract in that case, not that it would be a problem i dont think.

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
nicholaspai: I test this [here](https://github.com/UMAprotocol/protocol/pull/2293/files#diff-b18cec4947547731c8f1cec2b5b452e46e1967d11f8d2879033e94162d913be5R2378). What do you mean by "the behavior before you reverted with 0 pfc"? When PfC and collateral balance are both 0, there is no revert.

packages/core/contracts/financial-templates/common/FeePayer.sol
nicholaspai: Its subtle but this [line](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/FeePayer.sol#L207) can revert because of a divide by 0 where PfC = 0 but balance > 0

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
nicholaspai: >  If I understand right the multiplier wont change unless the excess collateral is significant enough in size? Its possible that there could be some dust left in the contract in that case, not that it would be a problem i dont think.

Correct, if the calculation (collateral balance / pfc) has dust truncates to 1e18, then the multiplier will not change. I think this is no worst than the other situations where dust can accumulate. What's nice is that as the dust accumulates, `gulp()` can be used to absorb it once it reaches a certain threshold. Without `gulp`, the dust would be frozen collateral locked in the contract and not part of PfC.

packages/core/test/financial-templates/perpetual-multiparty/PerpetualPositionManager.js
daywiss: Ok thanks, yea I thought you explicitly added a new revert condition and was testing it here, but based on your comments it seems you are just adding a new test case. So disregard that question.

daywiss (approved):

chrismaree (approved):
Looks good to me! how much bytecode do we have left? do you want to (in a separate PR) perhaps add back some of our lost require messages?

nicholaspai:
> Looks good to me! how much bytecode do we have left? do you want to (in a separate PR) perhaps add back some of our lost require messages?

There are 215 bytes remaining in the Perp and 482 in the EMP. I think we should add back require messages after the audit review since we might have to add code then

mrice32 (approved):
LGTM!

PR #2300, opened 2020-12-15
feat(proposer): add index.js script for optimistic oracle infrastructure package

### LINKED ISSUE #2294: Create skeleton structure for Optimistic Oracle proposer bot
**Why**

This bot should respond to any Optimistic Oracle price request in a timely and accurate manner. Without this bot, the OO is an insecure price feed.

**How**

- Main for loop:
    - Check for pending price requests
        - If no proposal, then propose a price
    - Check for proposed price requests
        - Attempt to settle the proposal        
        - If cannot settle, then check if a dispute is possible
    
- Other logic:
    - Needs to ensure it has enough collateral to stake a proposal or dispute bond
    - For any given price identifier, needs to know to which precision the price should be resolved to. For example, it is possible that price identifiers such as USDETH are supposed to be resolved to 5 decimal points and funding rate identifiers are supposed to be resolved to 4 decimal points.

**Resolution**

Create an `index.js` main script within a new package that runs without error. This should use as little code as possible and the goal is just to get a structure started for implementing this bot.

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:

**Motivation**

#2294 

Adds a `optimistic-oracle` package which will contain a runner and strategies used by the OO bots. 

This PR adds a skeleton of the `index.js` code that can be used to run both the OO proposer and disputer strategies. The proposer+disputer strategies, which I think can be combined into one strategy, will need to read data from an OO client which I'll develop in the next PR.

I think the OO client can live in the `financial-templates-lib/src/clients` directory, and we should consider renaming `financial-templates-lib` to `uma-sdk` or something along those lines.

** Open questions **

Meta level: what would be a good name for this new package? Some other naming ideas:
- `oo-keeper`
- `optimistic-oracle-keeper`

**Issue(s)**

Fixes #2294


packages/proposer/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
+See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
```
nicholaspai, line 4: @mrice32 should I add anything to this?

packages/proposer/index.js
```diff
@@ -0,0 +1,128 @@
...
+ * @param {Number} pollingDelay The amount of seconds to wait between iterations. If set to 0 then running in serverless
+ *     mode which will exit after the loop.
+ * @param {Number} errorRetries The number of times the polling loop will re-try before throwing if an error occurs.
+ * @param {Number} errorRetriesTimeout The amount of milliseconds to wait between re-try iterations on failed loops.
+ * @param {Object} config General configuration params passed to setupFunc and pollingFunc.
+ * @return None or throws an Error.
+ */
+async function run({ logger, setupFunc, pollingFunc, pollingDelay, errorRetries, errorRetriesTimeout, config }) {
+  try {
+    // TODO: Should we set default values for the required fields in `config` and use the
```
nicholaspai, line 23: @UMAprotocol/eng Thoughts on defining default values for the required params in this config object that the runner script itself needs, such as `name` which is used in logging.

chrismaree: I don't think that should be added here. Adding a `name` to logs, to inform the bot that reported a message in GCP or in slack is not done line this. Rather, we use the `bot-identifier` tag that is appended to the environment variables of the bot running and it extracted at the transport layer. You can see it within the Winston Logger here: https://github.com/UMAprotocol/protocol/blob/829e2c51f4ce193a58593f8f2847c73627d3f605/packages/financial-templates-lib/src/logger/Logger.js#L52

Therefore, I feel that it should be removed and we should stick to the convention where the `at` within the Winston logs defines the literal file where the log is produced and we leave the `bot-identifier` up to pulling the `name` prop from the env.

If we want to change this I think it should be done consistently through the bot stack in a separate PR.

chrismaree (approved):
This looks good to me as an initial stub. It's basically the same as what we've got for the other bots, which is a good pattern to follow IMHO. I just have one nit on the name param but other than this it looks good to me.

packages/proposer/index.js
nicholaspai: Ah yes I forgot about `bot-identifier`. So should I be replacing `name` with `bot-identifier` in this file?

nicholaspai:
> This looks good to me as an initial stub. It's basically the same as what we've got for the other bots, which is a good pattern to follow IMHO. I just have one nit on the name param but other than this it looks good to me.

Just wanted to check again, do you think that this runner can be a generalized runner that the liquidators, disputers, and future proposer bots all can use to execute their strategies?

packages/proposer/index.js
nicholaspai: @chrismaree the main difference in this file is that where previously the "at" was "Liquidator#index", now there is no "Liquidator" to speak of. We could either try to use the `bot-identifier` value or use a more generic value like "Runner#index".

For example we could use:
```
const name = process.env.BOT_IDENTIFIER ? process.env.BOT_IDENTIFIER : "Runner"
```

Or maybe "Runner" would just be more accurate

nicholaspai:
@UMAprotocol/eng I changed the PR description and initial stub to make it more clear that this runner should be only be used for the OO proposer and disputer strategies. PTAL

chrismaree (approved):
LGTM!

daywiss:
> Just wanted to check again, do you think that this runner can be a generalized runner that the liquidators, disputers, and future proposer bots all can use to execute their strategies?

I could see this happening, how much effort it would be im not sure, but probably a lot. Id design it by refining the API such that the business logic could be a single async call, called with common parameters. The injected parameters would consist of readable state about the world (all contracts, prices, etc) and a way to write state ( local state cache and transactions). Your common infrastructure would essentially manage the process of updating world state, calling the "business logic" function and handling any state mutations or transactions which come out of that.  Anyone programming a new bot would just learn the API and provide an async function with the logic to do what it needs.

daywiss (approved):

packages/optimistic-oracle/CHANGELOG.md
```diff
@@ -0,0 +1,4 @@
+# Change Log
+
+All notable changes to this project will be documented in this file.
```
nicholaspai, line 3: @mrice32 is this a reasonable starting change log for this package?

mrice32: Yeah, that works. If you want, you can add a line for the initial version -- not sure what we've done for the other packages.

packages/optimistic-oracle/index.js
```diff
@@ -0,0 +1,129 @@
...
+    // TODO:
+    // - Miscellaneous setup
+    // - Instantiate an OO client
+    // - Run proposer and/or disputer strategy
+
+    // Create a execution loop that will run indefinitely (or yield early if in serverless mode)
+    for (;;) {
+      await retry(
+        async () => {
+          return true;
```
mrice32, line 50: What does the Boolean signify/do?

mrice32 (approved):
Looks great -- a few minor questions/nits.

packages/optimistic-oracle/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "@uma/proposer",
+  "version": "1.2.0",
```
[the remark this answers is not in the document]
nicholaspai: I didn't realize this change was made, do you think we should start at 1.0?

packages/optimistic-oracle/index.js
nicholaspai: It's a placeholder for now I'll make that more clear

mrice32: Sorry, a bit more context: in the other bots, this function doesn't return anything. Any reason to return a Boolean over just `return;`?

packages/optimistic-oracle/CHANGELOG.md
nicholaspai: done

packages/optimistic-oracle/package.json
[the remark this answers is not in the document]
nicholaspai: Started at 1.0.0

packages/optimistic-oracle/index.js
nicholaspai: Good point I'll just return nothing

mrice32 (approved):
LGTM!

Issue #2006, opened 2020-09-18
Reorganize umips scripts directory in core


chrismaree:
**Why**

Right now the core's scripts directory is a bit all over the place. We refer to some UMIP upgrades as `UMIP-3`, & `UMIP-8` but others are `creator-umip`. Additionally, some UMIPs use the same script (partially the `2_VoteSimulate.js` is common. This should be cleaned up.

**How**

Move all UMIP scripts to an UMIPs directory within scripts. Re-name the all directories to follow the same convention. Pull vote simulate out into a separate `common` directory and re-use this logic within the UMIP directories.

**Resolution**

UMIPs are better organized.

**Difficulty Score [1-10]**

- 3: ~1 hour

daywiss:
> is there a pattern we want to consider to enable us to re-use these scripts more effectively? right now we have a bunch of sim links but there might be a better layout for this.

I think I've worked the least with these scripts but here are some thoughts:

I do think using these symlinks in general is annoying, it causes confusion on the dev and causes confusion in the tooling ( at least with my setup).  Rather than symlinks, I'm pretty sure you can just create a js file which its only job is to `require` the other file. Not much better, but at least symlinks are gone. 

The next lowest lift here would be to pull out the vote simulate file to work across all the scripts, like what you are suggesting chris. There's a copy of this file in umip-15 and umip-3 which is pretty confusing. Recent changes to the umip-3 file in #2302 brings it very close to the file in umip-15.  If its possible to reconcile differences and make this one file work for all the scripts that would be nice, and then pull it out as a shared dependency.

Next would be to refactor the scripts in such a way that the functionality is separate from the umip configuration. These scripts seem like they were originally intended to be re-written and almost hard coded for each new umip, but in the process a lot of common themes were discovered and previous work was reused. Ideally we would want to abstract the scripts in such a way where they could be named based on what they do rather than what umip they did work for, and provide configuration variables to run for a specific umip.  Not clear to me how much work this is and if effort to payoff is worth it though.

PR #2302, opened 2020-12-15
feat(core): update vote simulate script to run on N proposols rather than just 1

### LINKED ISSUE #2301: umip vote simulator script umip-3/2_VoteSimulate.js fails when more than 1 proposal is proposed
**Description**

Recently when verifying UMIP transactions for a recent vote, there were 6 new identifiers added in 3 transactions. When running the vote simulator script linked to core/scripts/umip-3/2_VoteSimulate.js this would fail with an error related to only expecting a single pending proposal. This should be updated so that all pending proposals get voted on.

**Reproduction**

1. see the comments on how to fork mainnet in core/scripts/umip-3/2_VoteSimulate.js 
2. append @11459780 to the end of the infura link, signifying the block number to fork from, and run the fork
2.  Run the script specified in the script comment
3. See an error 

**Expected Behavior**
Forking from this block at a point where there are 4 pending votes, the script to try to vote on all pending votes in the affirmative and verify they all pass.

daywiss:

**Motivation**

#2301 

**Summary**

This fixes vote simulate script to vote on all active proposals rather than being hardcoded to work with just one. Most of logic copied from `core/scripts/umip-15/2_VoteSimulate.js`.   This was tested in production by forking mainnet at a time where we have several proposed votes.

Fixes #2301

### DIFF
--- a/packages/core/scripts/umip-3/2_VoteSimulate.js
+++ b/packages/core/scripts/umip-3/2_VoteSimulate.js
@@ -46,12 +46,12 @@ async function runExport() {
   const secondsPerDay = 86400;
   const accounts = await web3.eth.getAccounts();
 
-  // These two assertions must be true for the script to be starting in the right state. This assumes the script
-  // is run within the same round and phase as running the `Propose.js` script.
-  assert((await governor.numProposals()).toNumber() > 1);
+  const numProposals = (await governor.numProposals()).toNumber();
+
+  assert(numProposals >= 1);
 
   console.log(
-    "1 pending proposal. No pending requests.\nCurrent timestamp:",
+    `${numProposals} pending proposal. No pending requests.\nCurrent timestamp:`,
     currentTime,
     "Voting phase:",
     votingPhase,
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
Thank you for finally tackling this debt! LGTM

chrismaree (approved):
LGTM! thanks for updating this script. I see you added it to the `umip-3` script. is there a pattern we want to consider to enable us to re-use these scripts more effectively? right now we have a bunch of sim links but there might be a better layout for this. There is actually a tasks for this here: https://github.com/UMAprotocol/protocol/issues/2006 but I'm curious to hear what you think having just updated these scripts.

daywiss:
> LGTM! thanks for updating this script. I see you added it to the umip-3 script. is there a pattern we want to consider to enable us to re-use these scripts more effectively? right now we have a bunch of sim links but there might be a better layout for this. There is actually a tasks for this here: #2006 but I'm curious to hear what you think having just updated these scripts.

Ill jump into that issue to respond

PR #2303, opened 2020-12-16
fix(dec-2020-audit): [L05] Event missing emit keyword

chrismaree:
**Problem:**
In the createPerpetual function of the PerpetualCreator contract, the CreatedConfigStore event is emitted without the emit keyword.

Since Solidity v0.4.21, the emit keyword has been recommended to distinguish events from other function calls.

**Solution in this PR**
The `PerpetualCreator` is updated to follow the correct specification by using the `emit` keyword.


nicholaspai (approved):
I didn't even realize that you could NOT use the `emit` keyword. LGTM!

PR #2304, opened 2020-12-16
fix(dec-20-audit): [L07] Incorrect error message

chrismaree:
**Problem:**
The Liquidatable constructor and the PerpetualLiquidatable constructor both revert if the collateral requirement is not greater than 100%, but the error message states “CR is more than 100%”.

**Solution in this PR:**
This PR fixes this by making the error messages more descriptive.


nicholaspai (approved):
LGTM nice fix

mrice32 (approved):

PR #2306, opened 2020-12-16
fix(dec-2020-audit): [L10] Functions not failing early

chrismaree:
**Problem:**
The `createPerpetual` function of the PerpetualCreator contract will run most of its logic before eventually reverting when the `params.collateralAddress` is not approved. Similarly, the withdraw function of the PerpetualPositionManager contract reverts when `collateralAmount` is equal to zero, but only after already getting the positionData using the nontrivial _getPositionData.

**Solution in this PR:**
This PR moves the required checks to the top of the function to follow the Checks-Effects-Interactions pattern. 

Note that the audit recommends checking that an identifier is registered within the `IdentifierWhitelist` to preserve gas within the `PerpetualCreator` in the case of a reverting call. This change was chosen to _not_ be implemented as this function is not called very often (only by perpetual deployers) so reverting calls will be far less common than other public methods. Additionally, to enable the `PerpetualCreator` to be able to verify if an identifier is registered the `PerpetualCreator` would need to import two additional interfaces (`OracleInterfaces` & `IdentifierWhitelistInterface`) as well as implement additional methods and logic to repeat an existing check that occurs within the constructor of the `PerpetualPositionManager`. This additional complexity & code is not warranted to save a small amount of gas for a perpetual deployer.

### DIFF
--- a/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
+++ b/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
@@ -309,8 +309,8 @@ contract PricelessPositionManager is FeePayer {
         nonReentrant()
         returns (FixedPoint.Unsigned memory amountWithdrawn)
     {
-        PositionData storage positionData = _getPositionData(msg.sender);
         require(collateralAmount.isGreaterThan(0));
+        PositionData storage positionData = _getPositionData(msg.sender);
 
         // Decrement the sponsor's collateral and global collateral amounts. Check the GCR between decrement to ensure
         // position remains above the GCR within the witdrawl. If this is not the case the caller must submit a request.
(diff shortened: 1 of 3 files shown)


packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -309,8 +309,8 @@ contract PricelessPositionManager is FeePayer {
         nonReentrant()
         returns (FixedPoint.Unsigned memory amountWithdrawn)
     {
-        PositionData storage positionData = _getPositionData(msg.sender);
```
chrismaree, line 312: this change was not identified in the audit but was made for consistency.

nicholaspai (approved):
LGTM but I don't see this comment in the audit:

```
Note that the audit recommends checking that an identifier is registered within the IdentifierWhitelist to preserve gas within the PerpetualCreator in the case of a reverting call.
```

Can you provide more details about this suggestion?

chrismaree:
> LGTM but I don't see this comment in the audit:
> 
> ```
> Note that the audit recommends checking that an identifier is registered within the IdentifierWhitelist to preserve gas within the PerpetualCreator in the case of a reverting call.
> ```
> 
> Can you provide more details about this suggestion?

Ye, I chose to not implement this. Please see the PR description for my rationale and let me know if you agree with me.

nicholaspai (approved):
LGTM!

PR #2308, opened 2020-12-16
fix(dec-2020-audit): [L08] Misleading comments

chrismaree:
**Problem**
The following comments are not accurate:

- In the constructor of the ConfigStore contract, the comment says “New settings go into effect after a liveness period passes.” This could be confusing because the constructor directly sets the initial configuration, and the liveness period only applies to the following configuration changes.
- In the _updateConfig function of the same contract, the comment says “If liveness has passed, publish new reward rate.” but the proposerBondPct and timelockLiveness might also have changed.
- In the setBond functions of the OptimisticOracleInterface and the OptimisticOracle contract, the comments say “Requests a new price.”, which is not what these functions do.
- In the applyFundingRate function of the FundingRateApplier contract, the comment says “This method takes 4 distinct actions”, but then it only describes 3 actions.
- In the transformPriceIdentifier function of the FinancialProductLibrary contract, the comment refers to a transformedCollateralRequirement return value, but what is returned is the priceIdentifier.
- In the fromUnscaledUint function of the FixedPoint library, the comment says “b=5 gets stored internally as 5**18“, but it is stored as 5*(10**18).
- In the transformPriceIdentifier function of the PreExpirationIdentifierTransformationFinancialProductLibrary contract, the comment says “Else, return the transformed price identifier.”, but what is returned is the untransformed price identifier.
- In the repay function of the PerpetualPositionManager contract, the comment says “numTokens is the number of tokens to be burnt for a commensurate amount of collateral” but the collateral is not refunded.

**Solution in this PR:**
Each issue is addressed and the comments are updated. A few additional packing and styling changes were also made to the comments for consistency.


packages/core/contracts/financial-templates/perpetual-multiparty/ConfigStore.sol
```diff
@@ -94,8 +93,7 @@ contract ConfigStore is ConfigStoreInterface, Testable, Lockable, Ownable {
     }
 
     /**
-     * @notice Propose new configuration settings. New settings go into effect
-     * after a liveness period passes.
+     * @notice Propose new configuration settings. New settings go into effect after a liveness period passes.
```
nicholaspai, line 96: Did you miss the comment:

```
In the _updateConfig function of the same contract, the comment says “If liveness has passed, publish new reward rate.” but the proposerBondPct and timelockLiveness might also have changed.
```

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -419,7 +419,7 @@ contract PerpetualPositionManager is FundingRateApplier {
      * This is done by a sponsor to increase position CR. Resulting size is bounded by minSponsorTokens.
      * @dev Can only be called by token sponsor. This contract must be approved to spend `numTokens` of `tokenCurrency`.
      * @dev This contract must have the Burner role for the `tokenCurrency`.
-     * @param numTokens is the number of tokens to be burnt for a commensurate amount of collateral.
+     * @param numTokens is the number of tokens to be burnt from the sponsors debt position.
```
nicholaspai, line 422:
```suggestion
     * @param numTokens is the number of tokens to be burnt from the sponsor's debt position.
```

nicholaspai (approved):
Thanks for addressing my comment, lgtm

PR #2309, opened 2020-12-16
fix(dec-20-audit): [L02] Complex Repay Function

### LINKED ISSUE #2307: Add repay logic to EMP, as it is implemented in the perp
**Why**

Right now the Perpetual has an implementation of the simple `repay` method that enables a token sponsor to decrease their token debt by calling redeem followed by deposit. You can find the implementation here: https://github.com/UMAprotocol/protocol/blob/631504f6beabf7e405aa6504a9399dc2d178742b/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol#L424

This logic should also be added to the EMP.

**How**

Update the EMP to include this `repay` function.

**Resolution**

Contracts updated & associated tests are added.

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:

**Motivation**

#2307

The `repay` function of the `PerpetualPositionManager` contract simply redeems the tokens and then deposits the retrieved collateral. This should have no net effect on the collateral token. This produces clean code but introduces unnecessary fragilities and complexities:

- the deposit function pulls the collateral, so the user needs to have previously granted the relevant collateral token allowance to the contract before either having the collateral or knowing the amount of the deposit. Confusingly, they need to do this for an operation that ostensibly has nothing to do with the collateral token.
- the matching transfers correctly result in the same collateral balances, but they unexpectedly emit opposing Transfer events
- the transfer may be unexpectedly effected by idiosyncrasies of the collateral token, which could include transfer fees or recipient blacklists (like USDC)
- the repay operation unexpectedly emits Redeem and Deposit events

Consider directly implementing the repay function to avoid unnecessarily involving the collateral token and the sponsor’s collateral balance.

**Summary**

Directly implements `repay` and adds it to the ExpiringMultiParty as well. Updates unit tests + adds one check that the caller has set their token allowance.

**Issue(s)**

Fixes #2307

### DIFF
--- a/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
+++ b/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
@@ -420,7 +420,7 @@ contract PerpetualPositionManager is FundingRateApplier {
      * This is done by a sponsor to increase position CR. Resulting size is bounded by minSponsorTokens.
      * @dev Can only be called by token sponsor. This contract must be approved to spend `numTokens` of `tokenCurrency`.
      * @dev This contract must have the Burner role for the `tokenCurrency`.
-     * @param numTokens is the number of tokens to be burnt for a commensurate amount of collateral.
+     * @param numTokens is the number of tokens to be burnt from the sponsor's debt position.
      */
     function repay(FixedPoint.Unsigned memory numTokens)
         public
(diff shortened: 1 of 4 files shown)


packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol
```diff
@@ -732,7 +751,7 @@ contract PerpetualPositionManager is FundingRateApplier {
     // `create` method because it is possible that `create` is called on a new position (i.e. one without any collateral
     // or tokens outstanding) which would fail the `onlyCollateralizedPosition` modifier on `_getPositionData`.
     function _positionHasNoPendingWithdrawal(address sponsor) internal view {
-        require(_getPositionData(sponsor).withdrawalRequestPassTimestamp == 0, "Pending withdrawal");
+        require(_getPositionData(sponsor).withdrawalRequestPassTimestamp == 0);
```
chrismaree, line 754: rip to this require message. I feel like this one would be used somewhat frequently.

chrismaree (approved):
Looks good to me! While I really liked the simple `deposit(redeem(numTokens));` logic that we had before OZ does have a great point that it is unnecessarily complex under the hood, emits illogical events confusing to the user, and requires seemingly unnecessary token allowances.

mrice32 (approved):

PR #2310, opened 2020-12-16
fix(dec-20-audit): [L04] Emergency shutdown defined at the wrong level

nicholaspai:
**Motivation**

The emergency shutdown functionality is defined in the `PerpetualPositionManager` contract, but it is also used in its parent, the `FundingRateApplier` contract. Instead of defining the common functionality at a common base level of inheritance, it is defined at the bottom level and accessed in an unclear way by the parent. Since this parent contract has no access to the emergency shutdown functions and modifiers, it duplicates the code.

Consider extracting the emergency shutdown functionality to a separate contract that gets inherited by the first parent contract that requires it.

**Summary**

Moved the `emergencyShutdownTimestamp` along with timestamp-related modifiers into a parent contract that the `FundingRateApplier` inherits. 

This PR still leaves the `emergencyShutdownPrice` and the actual logic of how an emergency shutdown affects sponsor positions in the `PositionManager`.

**Details**

@UMAprotocol/eng Please take a hard look at the new inheritance tree and comment on if it makes the code easier to reason about. This audit fix is intended to improve readability,


packages/core/contracts/financial-templates/common/EmergencyShutdownable.sol
```diff
@@ -0,0 +1,71 @@
...
+    constructor() public {
+        emergencyShutdownTimestamp = 0;
+    }
+
+    /**
+     * @notice Premature contract settlement under emergency circumstances.
+     */
+    function emergencyShutdown() external virtual override notEmergencyShutdown() {
+        emit EmergencyShutdown(msg.sender, emergencyShutdownTimestamp);
+    }
```
mrice32, line 56: Two questions:
1. Is the expectation that this is overridden to include any necessary settlement logic?
2. Should this default implementation set the emergency shutdown timestamp/

mrice32 (commented):
Looks great! I think this new inheritance tree makes sense as long as it's clear what the responsibility of inheriting contracts is.

packages/core/contracts/financial-templates/common/EmergencyShutdownable.sol
nicholaspai: 1. Yes this should be overridden, so maybe its clearer if we don't implement it for now.
2. The reason why it doesn't set the timestamp is because the contract would need to inherit `Testable` in order to call `getCurrentTime()` but this would throw inheritance errors in `FundingRateApplier` which inherits `Testable` from both FeePayer and this contract.


I think the cleanest solution now is to remove the `emergencyShutdown()` method from this contract.

packages/core/contracts/financial-templates/common/EmergencyShutdownable.sol
```diff
@@ -0,0 +1,60 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+
+import "@openzeppelin/contracts/math/SafeMath.sol";
+
+/**
+ * @title EmergencyShutdownable contract.
+ * @notice Any contract that inherits this contract will have an emergency shutdown timestamp state variable.
```
nicholaspai, line 9: @UMAprotocol/eng I made the comments more clear explaining that this contract's purpose is to (1) store an emergency shutdown timestamp and (2) provide modifiers that check if the timestamp has been set. All settlement logic that takes place post-shutdown should be implemented by the child.

chrismaree (approved):
This looks good to me. it's a clean pattern. 

Do you want to add this to other places where emergency shutdown is possible, such as the EMP?

packages/core/contracts/financial-templates/common/EmergencyShutdownable.sol
mrice32: Seems reasonable. I don't see a super strong argument either way. Other options:
1. Remove Testable from FeePayer since it will be inherited through here anyway.
2. Declare this method, but don't implement it, which would _require_ derived classes to implement it.

nicholaspai:
> This looks good to me. it's a clean pattern.
> 
> Do you want to add this to other places where emergency shutdown is possible, such as the EMP?

The difference with the EMP is that it does not have an `emergencyShutdownTimestamp`, it just expires the contract prematurely, and therefore there are no `is/notEmergencyShutdown` modifiers.

packages/core/contracts/financial-templates/common/EmergencyShutdownable.sol
nicholaspai: 1. Then the ExpiringMultiParty won't be `Testable` because the EMP does not use this contract
2. I think this is reasonable but I couldn't think of a good enough reason to do it. It seems like the main point of this contract is to just provide the `emergencyShutdownTimestamp` and modifiers that check its state, but it leave all of the logic surrounding the timestamp to the child.

chrismaree:
> > This looks good to me. it's a clean pattern.
> > Do you want to add this to other places where emergency shutdown is possible, such as the EMP?
> 
> The difference with the EMP is that it does not have an `emergencyShutdownTimestamp`, it just expires the contract prematurely, and therefore there are no `is/notEmergencyShutdown` modifiers.

fair enough!

PR #2311, opened 2020-12-16
fix(dec-20-audit): [M01] Lack of event emission after sensitive actions

nicholaspai:

**Motivation**

The `_getLatestFundingRate` function of the `FundingRateApplier` contract does not emit relevant events after executing the sensitive actions of setting the fundingRate, updateTime and proposalTime, and transferring the rewards.

Consider emitting events after sensitive changes take place, to facilitate tracking and notify off-chain clients following the contract’s activity.

**Details**

#2284 added the `FundingRateUpdated` event, and this PR adds more information to the event emitted about the reward payment.

I've chosen to NOT add the `proposalTime` to the event because this event will only emit if the `updateTime` is set to the `proposalTime` after a successful proposal. We might want to consider adding the timestamp at which the proposal was published, but I don't think this  would be useful to anyone.


packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -228,18 +228,18 @@ abstract contract FundingRateApplier is FeePayer {
...
-                            _pfc().mul(_getConfig().rewardRatePerSecond).mul(proposalTime.sub(lastUpdateTime));
+                        reward = _pfc().mul(_getConfig().rewardRatePerSecond).mul(proposalTime.sub(lastUpdateTime));
                         if (reward.isGreaterThan(0)) {
                             _adjustCumulativeFeeMultiplier(reward, _pfc());
                             collateralCurrency.safeTransfer(request.proposer, reward.rawValue);
                         }
                     }
 
-                    emit FundingRateUpdated(fundingRate.rate.rawValue, fundingRate.updateTime);
+                    emit FundingRateUpdated(fundingRate.rate.rawValue, fundingRate.updateTime, reward.rawValue);
```
nicholaspai, line 242: As highlighted in the PR description, do we think there's any reason to also include the `currentTime` in this event? i.e. the timestamp at which the proposal was published?

chrismaree: Emitted logs _do_ include timestamps but we've historically included time for sensitive things in other places (such as the DVM) so I don't really see any downside in including it. However, I think the issue from OZ also wants the `proposalTime` within the event, which is currently not there.

mrice32: Agreed. Proposal time might be useful to identify the request.

mrice32 (approved):
LGTM once proposal time is added!

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: The thing is, this event only fires once a new proposal is published. So the `updateTime` is always set to the `proposalTime`. So the `proposalTime` is always emitted, but its just equivalent to the `updateTime` now.

mrice32: Ahhh good point.

mrice32 (approved):

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree: in that case, then it's good then. perhaps add a comment to explain that?

chrismaree (approved):
LGTM. Please just add a comment to make it clear that proposal time is = update time

nicholaspai:
> LGTM. Please just add a comment to make it clear that proposal time is = update time

Done, let me know if you think this comment is OK:

```
// This event will only be emitted after the fundingRate struct's "updateTime" has been set
// to the latest proposal's proposalTime, indicating that the proposal has been published.
// So, it suffices to just emit fundingRate.updateTime here.
```

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: Done ptal

daywiss:
@nicholaspai am i missing something here, it seems like the issue they have was with missing events, but i dont really understand if they are saying the existing events don't send relevant data or that we need to add additional events to some functions?

nicholaspai:
> @nicholaspai am i missing something here, it seems like the issue they have was with missing events, but i dont really understand if they are saying the existing events don't send relevant data or that we need to add additional events to some functions?

@daywiss The audit review was made before the `FundingRateUpdated` event was added so at the time there was no event that emitted information about a changing funding rate.

This [PR](https://github.com/UMAprotocol/protocol/commit/472ffeb332fa96e15a43bcaea525006ee4818375) added in the events they are ultimately looking for.

daywiss (approved):

PR #2313, opened 2020-12-16
fix(dec-20-audit): [L03] Current config not update

nicholaspai:

**Motivation**

In the `getCurrentConfig` of the `ConfigStore` contract, when the pending proposal has passed the `pendingConfig` is returned but the `currentConfig` is not updated. Since `currentConfig` is a private variable this has minimal implications, but it causes a confusing internal state for the contract, which would emit the `ChangedConfigSettings` later than the moment where it starts to be used.

Consider using the `updateConfig` modifier to the `getCurrentConfig` function.

**Summary**

Adds `updateConfig()` modifier to `getCurrentConfig`, which only needs to return the current config now, and updates tests


packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -200,7 +200,7 @@ abstract contract FundingRateApplier is FeePayer {
         return OptimisticOracleInterface(finder.getImplementationAddress(OracleInterfaces.OptimisticOracle));
     }
 
-    function _getConfig() internal view returns (ConfigStoreInterface.ConfigSettings memory) {
+    function _getConfig() internal returns (ConfigStoreInterface.ConfigSettings memory) {
```
chrismaree, line 203: why is this no longer view? is this not separate to the point of this issue?

chrismaree (commented):
Can you explain why other functions `_getConfig` & `_validateFundingRate` have their view modifies removed?

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: It's no longer view because this method calls `getCurrentConfig()` which is not a view method anymore

mrice32 (commented):
Hmmmmmmmm, this is an interesting issue. Doesn't the fix that's suggested cause us to have another `get` method that modifies state, which they recommend against in M02? It seems that L03's suggested fix contradicts M02... we might want to ask for clarification.

nicholaspai:
> Hmmmmmmmm, this is an interesting issue. Doesn't the fix that's suggested cause us to have another `get` method that modifies state, which they recommend against in M02? It seems that L03's suggested fix contradicts M02... we might want to ask for clarification.

Yeah it does and I think the main issue is that the naming of these methods as "getX" is confusing since they also modify state

nicholaspai:
> Hmmmmmmmm, this is an interesting issue. Doesn't the fix that's suggested cause us to have another `get` method that modifies state, which they recommend against in M02? It seems that L03's suggested fix contradicts M02... we might want to ask for clarification.

After double checking with the reviewers their main issue is:

> there's something unusual about getCurrentConfig returning a particular value and then later emitting the ChangedConfigSettings event to reflect that value

And I think this is very valid. The event should be emitted the first time that the new state is available.

mrice32:
> > Hmmmmmmmm, this is an interesting issue. Doesn't the fix that's suggested cause us to have another `get` method that modifies state, which they recommend against in M02? It seems that L03's suggested fix contradicts M02... we might want to ask for clarification.
> 
> 
> 
> After double checking with the reviewers their main issue is:
> 
> 
> 
> > there's something unusual about getCurrentConfig returning a particular value and then later emitting the ChangedConfigSettings event to reflect that value
> 
> 
> 
> And I think this is very valid. The event should be emitted the first time that the new state is available.

Do you thing we should change the name of the getCurrentConfig method to something that implies it's state-modifying?

nicholaspai:
> > > Hmmmmmmmm, this is an interesting issue. Doesn't the fix that's suggested cause us to have another `get` method that modifies state, which they recommend against in M02? It seems that L03's suggested fix contradicts M02... we might want to ask for clarification.
> > 
> > 
> > After double checking with the reviewers their main issue is:
> > > there's something unusual about getCurrentConfig returning a particular value and then later emitting the ChangedConfigSettings event to reflect that value
> > 
> > 
> > And I think this is very valid. The event should be emitted the first time that the new state is available.
> 
> Do you thing we should change the name of the getCurrentConfig method to something that implies it's state-modifying? @nicholaspai

Yes I think we should, I'll commit something now.

mrice32 (approved):

PR #2314, opened 2020-12-16
fix(dec-20-audit): [L11] Propose or Dispute a price from the zero address

daywiss:

**Motivation**
The Optimistic oracle allows anyone to propose a price or dispute a proposal on behalf of the zero address. Since the state of the incentive mechanism is partially determined by whether the proposer or disputer addresses are non-zero, this won’t update the state machine, but it will set variables and emit events prematurely. For example, a client may incorrectly react to a DisputePrice event even though the dispute is not recognized by the oracle. Consider restricting the proposePriceFor and disputePriceFor function parameters to non-zero participants.

**Summary**

Adds 2 require statements in the appropriate functions and adds a tests to ensure calls fail when passing 0x0 address.

### DIFF
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -270,6 +270,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         bytes memory ancillaryData,
         int256 proposedPrice
     ) public override nonReentrant() returns (uint256 totalBond) {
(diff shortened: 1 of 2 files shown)


nicholaspai (commented):
Drive by: can you change the PR's title and the Motivation section of the description to copy exactly the audit's title and comments? Look at the other audit PR's for example.

nicholaspai (approved):
LGTM nicely done

mrice32 (approved):
This was a nice audit catch and good change. +1!

chrismaree (approved):

PR #2315, opened 2020-12-17
fix(dec-2020-audit): [N09] Unnecessary imports

chrismaree:
**Problem:**

In the PerpetualPositionManager.sol file, consider removing the import statement for FeePayer.sol, as it is never used in the PerpetualPositionManager contract.

Also, in the FundingRateApplier.sol file, consider removing the import statements for StoreInterface.sol and FinderInterface.sol, as they are never used in the FundingRateApplier contract.

**Solution in this PR**
These redundant imports were removed.


nicholaspai (approved):
LGTM

mrice32 (approved):

mrice32 (approved):
Would be really nice to have a linter for this!

PR #2316, opened 2020-12-17
fix(dec-2020-audit): [N10] Unnecessary type cast

chrismaree:
**Problem identified in audit:**

In lines 102 and 104 of PerpetualCreator.sol the derivative variable is cast to the address type. Since it is defined as address type, the casts are unnecessary.

**Solution in this PR**
The redundant type casting was removed. This change was also made to the EMP creator, which was not audited but the change was made for consistency.

### DIFF
--- a/packages/core/contracts/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.sol
+++ b/packages/core/contracts/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.sol
@@ -87,9 +87,9 @@ contract ExpiringMultiPartyCreator is ContractCreator, Testable, Lockable {
 
         _registerContract(new address[](0), address(derivative));
 
-        emit CreatedExpiringMultiParty(address(derivative), msg.sender);
+        emit CreatedExpiringMultiParty(derivative, msg.sender);
 
-        return address(derivative);
+        return derivative;
     }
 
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):
Good catch LGTM

packages/core/contracts/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.sol
```diff
@@ -87,9 +87,9 @@ contract ExpiringMultiPartyCreator is ContractCreator, Testable, Lockable {
 
         _registerContract(new address[](0), address(derivative));
```
mrice32, line 88: Should we remove the cast on this line as well?

packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualCreator.sol
```diff
@@ -99,9 +99,9 @@ contract PerpetualCreator is ContractCreator, Testable, Lockable {
 
         _registerContract(new address[](0), address(derivative));
```
mrice32, line 100: Same as above.

mrice32 (commented):
This one is interesting -- why was this here in the first place?

PR #2318, opened 2020-12-17
fix(dec-2020-audit): [M02] Functions with unexpected side-effects

chrismaree:
**Problem Identified in the audit**
Some functions have side-effects. For example, the _getLatestFundingRate function of the FundingRateApplier contract might also update the funding rate and send rewards. The getPrice function of the OptimisticOracle contract might also settle a price request. These side-effect actions are not clear in the name of the functions and are thus unexpected, which could lead to mistakes when the code is modified by new developers not experienced in all the implementation details of the project.

**Solution in this PR**
This function splits this `_getLatestFundingRate` into separate getters and setters by defining a new function `_updateFundingRate` which modify the funding rate state and then the `_getLatestFundingRate` returns the latest funding rate. In conjunction, these functions perform the same logic as the previous `_getLatestFundingRate` but in a two-step process to make it clear which internal method is modifying state and which is simply a getter method.

Additionally, this PR updates the `getPrice` method in the OO to `settleAndGetPrice` to make it clear that this is a state modifying function.


packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -251,6 +249,9 @@ abstract contract FundingRateApplier is FeePayer {
                 }
             }
         }
+    }
+
+    function _getLatestFundingRate() internal view returns (FixedPoint.Signed memory) {
```
nicholaspai, line 254: Can this function be removed, and just replace any `_getLatestFundingRate` calls with `fundingRate.rate`?

nicholaspai (commented):
WDYT we should do about the comment "The getPrice function of the OptimisticOracle contract might also settle a price request"?

On one hand, the `Voting.sol` contract has the same non-intuitive naming of the method `getPrice()` which modifies state, so keeping the function in the OO as is would be consistent. On the other hand, we could rename `getPrice()` to `settleAndGetPrice()` to be more clear.

mrice32:
> the `Voting.sol` contract has the same non-intuitive naming of the method `getPrice()` which modifies state

`Voting.sol`'s `getPrice` is a view method -- it determines the resolution without changing state. So making the OO's `getPrice` a `view` method would be more consistent I think.

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
chrismaree: sure. I've removed this function as it's only used once.

chrismaree:
> > the `Voting.sol` contract has the same non-intuitive naming of the method `getPrice()` which modifies state
> 
> `Voting.sol`'s `getPrice` is a view method:
> 
> https://github.com/UMAprotocol/protocol/blob/0f53b4962a19e20db97d8276a7454ac3c40e7dab/packages/core/contracts/oracle/implementation/Voting.sol#L331
> 
> It determines the resolution without changing state. So making the OO's `getPrice` a `view` method would be more consistent I think.
 
We can make this change but how do we `settle` a price request? This would require that each price request also involves calling the `settle` function https://github.com/UMAprotocol/protocol/blob/9954a8731153dcb65fb70a6a045481fa64611326/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L426

which seems to me to be unnecessary complexity in the pursuit of consistency.

nicholaspai:
> > > the `Voting.sol` contract has the same non-intuitive naming of the method `getPrice()` which modifies state
> > 
> > 
> > `Voting.sol`'s `getPrice` is a view method:
> > https://github.com/UMAprotocol/protocol/blob/0f53b4962a19e20db97d8276a7454ac3c40e7dab/packages/core/contracts/oracle/implementation/Voting.sol#L331
> > 
> > It determines the resolution without changing state. So making the OO's `getPrice` a `view` method would be more consistent I think.
> 
> We can make this change but how do we `settle` a price request? This would require that each price request also involves calling the `settle` function
> 
> https://github.com/UMAprotocol/protocol/blob/9954a8731153dcb65fb70a6a045481fa64611326/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L426
> 
> which seems to me to be unnecessary complexity in the pursuit of consistency.

As discussed offline, I think we have two equally good choices:
1) To be consistent with only naming `view` methods as `getX`, rename `getPrice` to `settleAndGetPrice`.
2) Remove `_settle` logic from `getPrice` and force users to call `settle` before `getPrice`

nicholaspai (approved):
LGTM this change is hopefully going to clear up future confusion between the Voting.sol's `getPrice` method and this contract's similarly named `settleAndGetPrice`

chrismaree:
@mrice32 PTAL and let me know if you happy with this PR.

mrice32 (approved):
LGTM!

PR #2319, opened 2020-12-17
fix(dec-20-audit): [L06] Inconsistent hasPrice determination

daywiss:

**Motivation**

[L06] Inconsistent hasPrice determination
The hasPrice function of the OptimisticOracle contract returns true if the request is either Settled or Resolved. Although the internal state of the contract does not record the price in the Resolved state, this makes sense because retrieving the price will settle the request. This means that from a user’s point of view, a Resolved price request can be considered to have a price. However, the same reasoning suggests that an Expired price request can also be considered to have a price.

Consider adding another condition to the hasPrice function so it returns true for Expired price requests.

**Summary**

This adds an additional state check to resolve hasPrice to true if price proposal is also expired. Adds in several tests to make sure hasPrice works as intended in various scenarios, including the scenario indicated above where price proposal is expired but not settled.

### DIFF
--- a/packages/core/test/oracle/OptimisticOracle.js
+++ b/packages/core/test/oracle/OptimisticOracle.js
@@ -154,6 +154,70 @@ contract("OptimisticOracle", function(accounts) {
     await verifyBalanceSum(optimisticOracle.address, reward);
   });
 
+  describe("hasPrice", function() {
+    beforeEach(async function() {
+      await collateral.transfer(optimisticRequester.address, reward);
+      await optimisticRequester.requestPrice(identifier, requestTime, "0x", collateral.address, "0");
+    });
+    it("Should return false when no price was ever proposed", async function() {
+      const result = await optimisticOracle.hasPrice(optimisticRequester.address, identifier, requestTime, "0x");
+      assert.equal(result, false);
+    });
+    it("Should return false when price is proposed but not past liveness", async function() {
+      await collateral.approve(optimisticOracle.address, totalDefaultBond, { from: proposer });
+      await optimisticOracle.proposePrice(optimisticRequester.address, identifier, requestTime, "0x", correctPrice, {
+        from: proposer
+      });
+      const result = await optimisticOracle.hasPrice(optimisticRequester.address, identifier, requestTime, "0x");
+      assert.equal(result, false);
+    });
+    it("Should return false when price is proposed and disputed", async function() {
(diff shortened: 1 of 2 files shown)


packages/core/test/oracle/OptimisticOracle.js
```diff
@@ -154,6 +154,72 @@ contract("OptimisticOracle", function(accounts) {
...
+
+      await collateral.approve(optimisticOracle.address, totalDefaultBond, { from: disputer });
+      await optimisticOracle.disputePrice(optimisticRequester.address, identifier, requestTime, "0x", {
+        from: disputer
+      });
+
+      const result = await optimisticOracle.hasPrice(optimisticRequester.address, identifier, requestTime, "0x");
+      assert.equal(result, false);
+    });
+    it("Should return true when price is proposed and past liveness but not settled", async function() {
```
nicholaspai, line 197: Isn't this a duplicate test of the one beginning on line 174?

packages/core/test/oracle/OptimisticOracle.js
```diff
@@ -154,6 +154,72 @@ contract("OptimisticOracle", function(accounts) {
...
+      await collateral.approve(optimisticOracle.address, totalDefaultBond, { from: proposer });
+      await optimisticOracle.proposePrice(optimisticRequester.address, identifier, requestTime, "0x", correctPrice, {
+        from: proposer
+      });
+      await timer.setCurrentTime((await timer.getCurrentTime()).add(await optimisticOracle.defaultLiveness()));
+      const result = await optimisticOracle.hasPrice(optimisticRequester.address, identifier, requestTime, "0x");
+      assert.equal(result, true);
+    });
+    it("Should return true when price is proposed, disputed and resolved by dvm", async function() {
```
nicholaspai, line 206: Do you think you should also test the path where the price is proposed, past liveness, and settled?

nicholaspai (commented):
I'm surprised we didn't have tests for `hasPrice` before, thanks for adding! I left some nits on your tests

mrice32 (approved):
LGTM once @nicholaspai's comments are resolved!

chrismaree (approved):

packages/core/test/oracle/OptimisticOracle.js
daywiss: could be, i will check it out

daywiss: yes you are right, thanks good catch

PR #2321, opened 2020-12-17
feat(affiliates): add readmes and examples for dapp mining tagging

### LINKED ISSUE #2320: Document and show transaction tagging for dapp mining.
This is the part of the codebase which developers must integrate with to participate in Dapp mining, so we need to get out a specification early and show how to do it. 

- Add documentation about the specification for tagged transactions 
- show example code on how to encode with
  - ethers 
  - web3

daywiss:
**Motivation**

#2320 

**Summary**
Basic readme for context around dapp mining, and examples of how to tag transactions in web3 and ethers.

**Issue(s)**

Fixes #2320

### DIFF
--- a/packages/affiliates/examples/dapp-mining/README.md
+++ b/packages/affiliates/examples/dapp-mining/README.md
@@ -2,29 +2,9 @@
 
 This contains examples of how to tag transactions with both Web3 and Ethers libraries.
 
-## Background
-
-"Dapp Mining" is the name of a new program being run by Risk Labs. The goal is to incentivize third party developers
-to build Dapps (Decentralized Applications) on top of the UMA infrastructure. Through these Dapp interfaces
-we can measure the amount of synthetic tokens minted by "tagging" certain transaction on chain with a developers
-address or unique identifier. The incentives come in the form of UMA token airdrops each week to developers who bring in liquidity
-into UMA EMP (Expiring Multi Party) contracts. The exact amounts and values of UMA airdroped may vary based on
-many factors, but essentially the more liquidity provided through a Dapp, the higher the proportion of weekly
-rewards earned.
-
-## How It Works
-
-EMP contracts are designed to allow participants to "mint" synthetic tokens by depositing some form of collateral.
-This action can be attributed in a process we are calling "tagging". By tagging a transaction, offchain infrastructure
-can track the referrer for a minting action to a particular Dapp developer. This tagging will allow offchain infrastructure to estimate the proportion
-of synthetic token liquidty created through each interface and reward the developers proportionally. Every week
-a quantity of UMA will be airdropped to the tagged addresses. For more detail on tagging see the examples.
+These examples show how to tag your transactions when submitting a "create" request to the EMP.
+The examples will output the data fields on the transaction you would need to sign. Use it as a starting
+point to integrate the feature into your own front end.
 
 **[Web3 Example](./web3-tagging.js)**
 **[Ethers Example](./ethers-tagging.js)**
-
-## Joining The Program
-
-Participation in this Dapp Mining program requires you first whitelist your payout address with the Uma Project.
-Once white listed your address can be added to transactions that originate from your Dapp. This program will
-not necessarily be available for all EMP contracts, more detail will be provided closer to launch.
--- a/packages/affiliates/examples/dapp-mining/README.md
+++ b/packages/affiliates/examples/dapp-mining/README.md
@@ -1,0 +1,30 @@
+# Dapp Mining Examples
+
+This contains examples of how to tag transactions with both Web3 and Ethers libraries.
+
+## Background
+
+"Dapp Mining" is the name of a new program being run by Risk Labs. The goal is to incentivize third party developers
+to build Dapps (Decentralized Applications) on top of the UMA infrastructure. Through these Dapp interfaces
+we can measure the amount of synthetic tokens minted by "tagging" certain transaction on chain with a developers
+address or unique identifier. The incentives come in the form of UMA token airdrops each week to developers who bring in liquidity
+into UMA EMP (Expiring Multi Party) contracts. The exact amounts and values of UMA airdroped may vary based on
+many factors, but essentially the more liquidity provided through a Dapp, the higher the proportion of weekly
+rewards earned.
+
+## How It Works
+
+EMP contracts are designed to allow participants to "mint" synthetic tokens by depositing some form of collateral.
+This action can be attributed in a process we are calling "tagging". By tagging a transaction, offchain infrastructure
+can track the referrer for a minting action to a particular Dapp developer. This tagging will allow offchain infrastructure to estimate the proportion
+of synthetic token liquidty created through each interface and reward the developers proportionally. Every week
+a quantity of UMA will be airdropped to the tagged addresses. For more detail on tagging see the examples.
+
+**[Web3 Example](./web3-tagging.js)**
+**[Ethers Example](./ethers-tagging.js)**
+
+## Joining The Program
+
+Participation in this Dapp Mining program requires you first whitelist your payout address with the Uma Project.
+Once white listed your address can be added to transactions that originate from your Dapp. This program will
+not necessarily be available for all EMP contracts, more detail will be provided closer to launch.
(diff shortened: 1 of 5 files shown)


packages/affiliates/examples/dapp-mining/web3-tagging.js
```diff
@@ -0,0 +1,59 @@
...
+  const encodedData = emp.methods
+    .create({ rawValue: web3.utils.toWei(collateralToSend) }, { rawValue: web3.utils.toWei(tokensToCreate) })
+    .encodeABI();
+  return encodedData;
+}
+
+// This takes encoded data field and just appends the tag
+function tagData(data, tag) {
+  // convert tag to hex and remove the 0x prefix
+  return data.concat(web3.utils.toHex(tag).slice(2));
```
daywiss, line 30: Would like to maybe hash the tag so that we can use any kind of string, obscure address and produces fixed length output. What do you guys think of this? Currently this just converts a string to hex.

nicholaspai: What would be the hash secret that we could use to decode it?

packages/affiliates/examples/dapp-mining/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+## Background
+
+"Dapp Mining" is the name of a new program being run by Risk Labs. The goal is to incentivize third party developers
+to build Dapps (Decentralized Applications) on top of the UMA infrastructure. Through these Dapp interfaces
+we can measure the amount of synthetic tokens minted by "tagging" certain transaction on chain with a developers
+address or unique identifier. The incentives come in the form of UMA token airdrops each week to developers who bring in liquidity
+into UMA EMP (Expiring Multi Party) contracts. The exact amounts and values of UMA airdroped may vary based on
+many factors, but essentially the more liquidity provided through a Dapp, the higher the proportion of weekly
+rewards earned.
```
smb2796, line 13: Might be good to highlight the fact that this is also an example implementation for third party developer miners who want to launch their own dapp mining programs. 

Also would be valuable to note that these rewards are intended to be a piece of the $UMA received for Developer mining by each contract. Maybe something along the lines of:

```suggestion
rewards earned. Typically the size of Dapp mining rewards will be a percentage of the developer mining rewards received by the contract that the Dapp mints for.

This reference implementation is also intended to be used by developer miners to create their own Dapp mining programs. With the same tagging mechanism, developer miners could payout a portion of their rewards to third party application developers or affiliate marketers.
```

packages/affiliates/examples/dapp-mining/README.md
```diff
@@ -0,0 +1,30 @@
...
+can track the referrer for a minting action to a particular Dapp developer. This tagging will allow offchain infrastructure to estimate the proportion
+of synthetic token liquidty created through each interface and reward the developers proportionally. Every week
+a quantity of UMA will be airdropped to the tagged addresses. For more detail on tagging see the examples.
+
+**[Web3 Example](./web3-tagging.js)**
+**[Ethers Example](./ethers-tagging.js)**
+
+## Joining The Program
+
+Participation in this Dapp Mining program requires you first whitelist your payout address with the Uma Project.
```
chrismaree, line 28:
```suggestion
Participation in this Dapp Mining program requires you first whitelist your payout address with the UMA Project.
```

packages/affiliates/examples/dapp-mining/README.md
```diff
@@ -0,0 +1,30 @@
...
+Once white listed your address can be added to transactions that originate from your Dapp. This program will
```
chrismaree, line 29: should we include a link to the onboarding process?

packages/affiliates/examples/dapp-mining/ethers-tagging.js
```diff
@@ -0,0 +1,57 @@
...
+}
+
+// Example of how you would construct the transaction with data field to be signed by private key
+function makeTransaction(data) {
+  return {
+    from: tag,
+    to: emp,
+    value: 0,
+    data,
+    gas: 1000000
```
chrismaree, line 38: is this required?

packages/affiliates/examples/dapp-mining/ethers-tagging.js
```diff
@@ -0,0 +1,57 @@
...
+function tagData(data, tag) {
+  return ethers.utils.hexConcat([data, tag]);
+}
+
+// Example of how you would construct the transaction with data field to be signed by private key
+function makeTransaction(data) {
+  return {
+    from: tag,
+    to: emp,
+    value: 0,
```
chrismaree, line 36: is this also required?

packages/affiliates/examples/dapp-mining/ethers-tagging.js
```diff
@@ -0,0 +1,57 @@
...
+  console.log("Running example with the following parameters:");
+  console.table({
+    emp,
+    tag,
+    tokens,
+    collateral
+  });
+  const data = createData(emp, collateral, tokens);
+  const taggedData = tagData(data, tag);
+  const transaction = makeTransaction(taggedData);
```
chrismaree, line 52: do you want to send the transaction?

chrismaree (commented):
Looks good! had a few nits and questions about required data but other than that I think this is great!

packages/affiliates/examples/dapp-mining/web3-tagging.js
daywiss: i just mean use `web3.utils.keccak256(string)`. Since we have a whitelist of the addresses we can still match on the other end. its kind of nice since then we have a fixed length string and can hash arbitrary data.

packages/affiliates/examples/dapp-mining/README.md
daywiss: yea good point. I dont know exactly how to organize all these docs but I will add this in the next commit. may also want to link to various articles/tweets for more info.

packages/affiliates/examples/dapp-mining/ethers-tagging.js
daywiss: Not required technically, but i just wanted to show a sample transaction output to be signed.

packages/affiliates/examples/dapp-mining/ethers-tagging.js
daywiss: I didnt want to actually send transaction in these examples otherwise to run it user would need to specify provider url, working emp address, etc. Just want to illustrate how to get 90% of the way there. But it might be useful to show how you would sign and send this, without actually doing it.

I think if i can abstract these into libraries id add some tests where the actual sending could be shown safely.

packages/affiliates/examples/dapp-mining/ethers-tagging.js
daywiss: i think i will remove both props you are suggesting, probably added automatically when sent

packages/affiliates/examples/dapp-mining/ethers-tagging.js
nicholaspai: I think its worth showing how to send it FWIW, we should make this as seamless as possible for the user to plug and play

chrismaree: ye, I agree. this is not really a standard tx anymore so showing how to send it on-chain is a good idea.

packages/affiliates/examples/dapp-mining/web3-tagging.js
daywiss: im going to just leave this as is for now

packages/affiliates/examples/dapp-mining/ethers-tagging.js
daywiss: I've added some inline comments for showing how to send the transactions in both ethers and web3. I think this is probably the bare minimum to get the point across without having the script more complicated to run. Maybe i feel like its a lot to ask the user to run a local test node in order to run the script, but if you guys think that is an ok to assume i can make the script submit the TX as well

chrismaree: sounds good to me!

chrismaree (approved):
LGTM! this is agreat readme. thanks for implementing all the nits :)

PR #2324, opened 2020-12-19
improve(core): add patched emp addresses

mrice32:
**Motivation**

Add newly deployed patch addresses that patch a previous bug in the EMP template (note: will not match current ABIs, but neither do the current addresses :)). 

**Summary**

Just adds new kovan and mainnet addresses.

**Issue(s)**

N/A


packages/core/networks/1.json
```diff
@@ -53,10 +53,10 @@
   },
   {
     "contractName": "ExpiringMultiPartyLib",
-    "address": "0x38015ddB8b34c84934Cff058F571349Cc7d4139d"
+    "address": "0x0246FBF444cAe32867b410464664f8F02e1822C7"
```
nicholaspai, line 56: Think you forgot the Kovan addresses

nicholaspai: Done.

nicholaspai (approved):
LGTM

chrismaree (approved):

PR #2327, opened 2020-12-21
fix(dec-20-audit): [N07] TODOs in code

chrismaree:
**Problem identified in audit:**
There are “TODO” comments in the code base that should be tracked in the project’s issues backlog. See for example:

1) Line 17 of FundingRateApplier.sol.
2) Line 427 of PerpetualLiquidatable.sol.
3) Line 532 of PerpetualPositionManager.sol.

**Solution presented in this PR**
Each todo has been handled separately.

1) This TODO has been fixed in a previous PR that was added post-audit commit hash. See the current [FundingRateApplier](https://github.com/UMAprotocol/protocol/blob/7649d459b499519b8e87461ce286d617d00feeae/packages/core/contracts/financial-templates/common/FundingRateApplier.sol#L14) that correctly uses the OO interface.
2) This TODO Notes that we should consider funding rate changes over the duration of a liquidation. There is not anything to do here as the funding rate at liquidation time is captured when the liquidation is created with the `_getFundingRateAppliedTokenDebt` called on line [310](https://github.com/UMAprotocol/protocol/blob/9ce936bc0e96027099cc687fd6a766f0af1e4353/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol#L310). As a result, when calculating RTV on line [431](https://github.com/UMAprotocol/protocol/blob/9ce936bc0e96027099cc687fd6a766f0af1e4353/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol#L431) we are already using the funding rate adjusted token debt, based on the funding rate at liquidation time. at this exact moment, the debt is paid off by the liquidator and so we must use the calculated funding rate at this point. The commenting in `PerpetualLiquidatable` has been updated to reflect this behavior.
3) The final TODO refers to `getCollateral` not considering any pending regular fees that are yet to be taken from the contract. To correct this behavior the `getCollateral` and `totalPositionCollateral` methods have been updated to consider pending fees through the introduction of the `_getPendingRegularFeeAdjustedCollateral` internal method. This method fetches any pending fees from the store and then pro-rata applies it to the requested collateral.

	Note that this change keeps the interfaces the same as before and keeps the methods as view functions. As a result, this will not impact any off-chain infra as we can simply call the methods with the `.call` notation to get the equivalent request.


packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol
```diff
@@ -423,9 +420,9 @@ contract PerpetualLiquidatable is PerpetualPositionManager {
         _settle(liquidationId, sponsor);
 
         // Calculate rewards as a function of the TRV.
-        // Note: all payouts are scaled by the unit collateral value so all payouts are charged the fees pro rata.
-        // TODO: Do we also need to apply some sort of funding rate adjustment to account for multiplier changes
-        // since liquidation time?
+        // Note1: all payouts are scaled by the unit collateral value so all payouts are charged the fees pro rata.
```
chrismaree, line 423: todo2 comment update

packages/core/test/financial-templates/expiring-multiparty/PricelessPositionManager.js
```diff
@@ -1746,9 +1746,19 @@ contract("PricelessPositionManager", function(accounts) {
...
 
     // Move time in the contract forward by 1 second to capture a 1% fee.
+
     const startTime = await pricelessPositionManager.getCurrentTime();
     await pricelessPositionManager.setCurrentTime(startTime.addn(1));
 
+    // getCollateral for a given sponsor should correctly reflect the pending regular fee that has not yet been paid.
+    // As no function calls have been made after incrementing time, the fees are still in a "pending" state.
+    // Sponsor has a position with 1e18 collateral in it. After a 1% fee is applied they should have 0.99e18.
+    assert.equal((await pricelessPositionManager.getCollateral(sponsor)).toString(), toWei("0.99"));
```
chrismaree, line 1756: this test verifies that a pending regular fee is applied to the collateral and total position collateral.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -148,6 +145,27 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
         return totalPaid;
     }
 
+    function getOutstandingRegularFees(uint256 time)
```
chrismaree, line 148: this view method is pulled out & re-used in the `payRegularFees` and `_getPendingRegularFeeAdjustedCollateral` functions.

packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol
```diff
@@ -652,33 +651,24 @@ contract PricelessPositionManager is FeePayer {
...
-    function getCollateral(address sponsor)
-        external
-        view
-        nonReentrantView()
-        returns (FixedPoint.Unsigned memory collateralAmount)
-    {
+    function getCollateral(address sponsor) external view nonReentrantView() returns (FixedPoint.Unsigned memory) {
         // Note: do a direct access to avoid the validity check.
-        return _getFeeAdjustedCollateral(positions[sponsor].rawCollateral);
+        return _getPendingRegularFeeAdjustedCollateral(_getFeeAdjustedCollateral(positions[sponsor].rawCollateral));
```
chrismaree, line 661: here the internal `_getPendingRegularFeeAdjustedCollateral` method is applied. Note that it NOT applied anywhere else as all other functions include the `fees()` modifier which means that pending regular fees are already accounted for.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -96,31 +96,28 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
...
         (FixedPoint.Unsigned memory regularFee, FixedPoint.Unsigned memory latePenalty) =
-            store.computeRegularFee(lastPaymentTime, time, collateralPool);
+            getOutstandingRegularFees(time);
         lastPaymentTime = time;
 
         totalPaid = regularFee.add(latePenalty);
         if (totalPaid.isEqual(0)) {
             return totalPaid;
         }
+        StoreInterface store = _getStore();
```
chrismaree, line 120: this is moved lower down as if the function returns early (pays no fees) there is no need to get the store interface sooner.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -138,6 +134,7 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
         _adjustCumulativeFeeMultiplier(totalPaid, collateralPool);
 
         if (regularFee.isGreaterThan(0)) {
+            StoreInterface store = _getStore();
```
chrismaree, line 137: this is moved to here has the store is not needed sooner in this method. save a small bit of gas on calls.

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -148,6 +145,27 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
...
+
+        // Exit early if there is no collateral from which to pay fees.
+        if (collateralPool.isEqual(0)) {
+            return (regularFee, latePenalty);
+        }
+
+        // Exit early if fees were already paid during this block.
+        if (lastPaymentTime == time) {
+            return (regularFee, latePenalty);
+        }
```
mrice32, line 164: nit: I think we end up doing these checks twice in the payRegularFees flow. Could we remove them from that function and only do them here?

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -148,6 +145,27 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
...
+
```
mrice32, line 165: nit: don't we also need to move the block of code that starts with:
```
if (totalPaid.isGreaterThan(collateralPool)) {
```

into this function in case the fees are > pfc?

packages/core/contracts/financial-templates/common/FeePayer.sol
chrismaree: I had done this to save gas on tx so that the `payRegualrFees` will exit early without calling the `getOutstandingRegularFees` if these null states are true. However, it is code repetition and I'm not actually sure it saves that much as in the worse case when it goes past the null checks it will do the checks twice.

I've refactored this to only check them within the `getOutstandingRegularFees` function.

packages/core/contracts/financial-templates/common/FeePayer.sol
chrismaree: Good point. this check & the associated logic needs to be put into the `getOutstandingRegularFees` method.

I have refactored this pretty drastically to accommodate this and clean up this implementation. PTAL

packages/core/contracts/financial-templates/common/FeePayer.sol
```diff
@@ -100,31 +100,74 @@ abstract contract FeePayer is AdministrateeInterface, Testable, Lockable {
...
-            // has no locked collateral.
-            lastPaymentTime = time;
-            return totalPaid;
+            return (regularFee, latePenalty, totalPaid);
         }
 
         // Exit early if fees were already paid during this block.
         if (lastPaymentTime == time) {
-            return totalPaid;
+            return (regularFee, latePenalty, totalPaid);
```
mrice32, line 163: Could this and the above if statement be combined to simplify:
```
// Exit early if there is no collateral or if fees were already paid during this block.
if (collateralPool.isEqual(0) || lastPaymentTime == time) {
   return (regularFee, latePenalty, totalPaid);
}
```

mrice32 (approved):
Looks awesome -- one minor nit.

packages/core/contracts/financial-templates/common/FeePayer.sol
chrismaree: yup can do & done!

mrice32 (approved):

PR #2329, opened 2020-12-21
fix(dec-20-audit): [H01] Bond penalty may not apply

mrice32:
**Motivation**

The optimistic oracle is designed so that in the event of a dispute, the incorrect party pays the bond penalty to the vindicated party, as determined by the DVM. However, if the proposer and disputer are the same entity, this transfer has no effect. The only remaining deterrent is the DVM fee. Since the bond size is specifically chosen to dissuade attackers from submitting the wrong price and delaying resolution, the ability to nullify the bond penalty undermines the economic reasoning. Moreover, if the reward exceeds the DVM fee, the attacker may actually be positively rewarded for delaying the resolution.

This attack does not apply to contracts deployed from the Perpetual Multiparty template, because they disregard disputed price requests. Nevertheless, it does limit the simplicity and applicability of the optimistic oracle. Consider burning some or all of the bond penalty and ensuring the reward is not high enough to compensate.

**Summary**

As suggested above, half of the bond is paid to the store in order to "burn" it. This makes the loss for someone who intentionally delays a request by proposing and disputing themselves proportional to the bond.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -345,15 +345,25 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         request.disputer = disputer;
 
         uint256 finalFee = request.finalFee;
-        totalBond = request.bond.add(finalFee);
+        uint256 bond = request.bond;
+        totalBond = bond.add(finalFee);
         if (totalBond > 0) {
             request.currency.safeTransferFrom(msg.sender, address(this), totalBond);
         }
 
         StoreInterface store = _getStore();
         if (finalFee > 0) {
-            request.currency.safeIncreaseAllowance(address(store), finalFee);
-            _getStore().payOracleFeesErc20(address(request.currency), FixedPoint.Unsigned(finalFee));
+            // Along with the final fee, "burn" half of the loser's bond to ensure that a larger bond always makes it
+            // proportionally more expensive to delay the resolution even if the proposer and disputer are the same
+            // party.
+            // burnedBond = ceil(bond / 2)
+            // Warning: this math _only_ works with a divisor of 2. This is why this number is not made into a variable/constant.
+            uint256 burnedBond = bond.div(2).add(bond % 2);
+
+            // Pay the final fee and the burned bond to the store.
+            uint256 totalFee = finalFee.add(burnedBond);
+            request.currency.safeIncreaseAllowance(address(store), totalFee);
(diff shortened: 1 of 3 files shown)


packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -549,7 +559,17 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
                 _stampAncillaryData(ancillaryData, requester)
             );
             bool disputeSuccess = request.resolvedPrice != request.proposedPrice;
-            payout = request.bond.mul(2).add(request.finalFee).add(request.reward);
+            uint256 bond = request.bond;
+
+            // Unburned portion of the loser's bond = floor(bond / 2)
+            uint256 unburnedBond = bond.div(2);
```
nicholaspai, line 565: Is there any way to link `unburnedBond` and `burnedBond` via state variables? It's not very readable how `burnedBond + unburnedBond = bond`.

However, I don't really like the idea of storing `burnedBond` on-chain because it contributes bytecode and gas to store a variable that isn't that important to the user, so I understand why you did it this way.

packages/core/test/oracle/OptimisticOracle.js
```diff
@@ -243,6 +251,49 @@ contract("OptimisticOracle", function(accounts) {
       await verifyBalanceSum(optimisticOracle.address, reward, totalCustomBond);
     });
 
+    it("Burned bond rounding", async function() {
```
nicholaspai, line 254: +1 for testing rounding

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Maybe we could use a helper method to standardize the computation?

nicholaspai: Yeah, we could return `burnedBond` + `unburnedBond` + an assertion that their sum equals `bond` which should help reduce some of the testing code

mrice32: I added a function to compute the burned bond and compute unburned bond by doing bond - burned bond, which should guarantee burned bond + unburned bond = bond.

chrismaree (approved):
LGTM!

nicholaspai (approved):
Thanks for adding the internal method I think this makes it more readable and easier to change in the future

PR #2330, opened 2020-12-22
feat(optimistic-oracle-keeper): Add OptimisticOracleClient stub code

### LINKED ISSUE #2322: Create MVP features of OptimisticOracleClient
**Why**

Optimistic Oracle keeper bots will need be able to query all pending requests and proposals from the Optimistic Oracle. This issue follows the pattern of the liquidator bot and how it uses the ExpiringMultiPartyClient to query for undercollateralized positions.

**How**

The MVP features would include an `update()` method designed to run on each OO-keeper bot iteration, and it should call the following functions:
- `getPriceRequests`
- `getPriceProposals`

**Resolution**

This client should be constructed by the OO-keeper main `index.js` file and should have its own set of unit tests.

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:
**Motivation**

#2322 

**Summary**

The main decision I want to highlight is that the the client is NOT placed within the `financial-templates-lib` package because it needs access to the `latest` contract ABI's, and the `financial-templates-lib` package uses `@uma/core@1.2`, which does not include the OO.

If there is a better way to engineer around this please LMK!

I've written a tiny amount of code to read OO events but I want to save that for a subsequent PR.

**Issue(s)**

Fixes #2322


mrice32:
@nicholaspai Can we grab the latest core version and just specify the version string in the getX() calls?

nicholaspai:
> @nicholaspai Can we grab the latest core version and just specify the version string in the getX() calls?

I think so but we first need to change the `ExpiringMultiPartyClient` unit tests to explicitly use the 1.2 contracts. I'll do a PR for that first

nicholaspai:
@mrice32 I know this is a draft PR but I'm trying to test that moving `OptimisticClient` into `financial-templates-lib` works. So far I've merged #2344 into this branch but I can't seem to call `getTruffleContract(OptimisticOracle, web3, latest)` in this [file](https://github.com/UMAprotocol/protocol/blob/14d121dbed5a1bec220721160450e2d5dc7f45ac/packages/financial-templates-lib/test/clients/OptimisticOracleClient.js#L11)

The error thrown is shortly after compiling:

```
Compilation finished successfully
Error HH700: Artifact for contract "OptimisticRequesterTest" not found.
```

I'm running this command to test: `yarn hardhat test ./test/clients/OptimisticOracleClient.js`

Any clue?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -0,0 +1,118 @@
...
+
+    // Fetch request states in parallel batches, 20 at a time, to be safe and not overload the web3 node.
+    const WEB3_CALLS_BATCH_SIZE = 150;
+
+    const [requestStates] = await Promise.all([
+      Promise.map(
+        requestEvents,
+        request =>
+          // Since we're making an async web3 call here, consider calling
+          // requests(web3.utils.soliditySha3(requester, id, timestamp, acData)) instead
```
nicholaspai, line 72: This is logic I'll tackle in the next PR: efficiently grabbing all OO price requests and categorizing them into proposals, disputes, requests, etc.

packages/optimistic-oracle/index.js
```diff
@@ -21,33 +21,36 @@ const { getWeb3 } = require("@uma/common");
  */
 async function run({ logger, web3, pollingDelay, errorRetries, errorRetriesTimeout }) {
   try {
-    // Load unlocked web3 accounts and get the networkId.
     const [networkId] = await Promise.all([web3.eth.net.getId()]);
-
-    // Setup contract instances. NOTE that getAddress("OptimisticOracle", networkId) will resolve to null in tests.
-    const oracle = new web3.eth.Contract(getAbi("OptimisticOracle"), getAddress("OptimisticOracle", networkId));
```
nicholaspai, line 28: We might need to construct this contract in this runner in the future, but I envision just passing the address + ABI to the Client and letting the client construct the contract object there.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -0,0 +1,118 @@
...
+
+  // Returns the last update timestamp.
+  getLastUpdateTime() {
+    return this.lastUpdateTimestamp;
+  }
+
+  async update() {
+    // Fetch contract state variables in parallel.
+    const [requestEvents, currentTime] = await Promise.all([
+      this.oracle.getPastEvents("RequestPrice", { fromBlock: 0 }), // TODO: Change this `fromBlock` to something > 0
```
mrice32, line 60: Can we do this now that #2360 has been merged?

Another thought: maybe we should set a configurable lookback limit for requests that defaults to a few days and query for all request "updates"  in that time period as opposed to just request events (i.e. query for disputes, proposals, etc). The idea being that we wouldn't expect a bot to be down for days, and a request with no activity (by us or others) over the past 4 days or so, probably warrants no action. Thoughts?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -0,0 +1,118 @@
...
+          // requests(web3.utils.soliditySha3(requester, id, timestamp, acData)) instead
+          // so we can grab more data including that relevant for proposals.
+          this.oracle.methods
+            .getState(
+              request.returnValues.requester,
+              request.returnValues.identifier,
+              request.returnValues.timestamp,
+              request.returnValues.ancillaryData ? request.returnValues.ancillaryData : "0x"
+            )
+            .call(),
```
mrice32, line 81: Hmmm, I'm curious about how to make this scalable. I know we're trying to rethink how much we're relying on big event queries, but this could very quickly blow through a lot of infura quota if I'm reading this correctly, since it seems that each run would query the state of every past request (unless you're planning to pre-filter the events in some way?). Thoughts?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -0,0 +1,118 @@
...
+    this.hexToUtf8 = this.web3.utils.hexToUtf8;
+  }
+
+  // Returns an array of Price Request objects that have no proposals yet.
+  getAllPriceRequests() {
+    return this.priceRequests;
+  }
+
+  // Returns an array of Price Request objects that someone has proposed a price for.
+  getAllPriceProposals() {
```
mrice32, line 40: Will the bot try to settle its own proposals? If so, should we add a similar method for "settleable" requests?

mrice32 (commented):
This looks awesome! Very nice -- a few high-level questions.

packages/core/index.js
```diff
@@ -107,9 +107,12 @@ function getAddressTest(contractName, networkId, version = "latest") {
     global.artifacts._provisioner._deploymentAddresses[contractName] &&
     artifacts._provisioner._networkConfig.chainId === networkId
   ) {
-    // In the hardhat case, there is no networks object, so we fall back to hardhat's global list of deployed addresses as long as hardhat's network id matches the one passed in.
+    // In the production hardhat case, there is no networks object, so we fall back to hardhat's global list of deployed addresses as long as hardhat's network id matches the one passed in.
     // Note: this is a bit hacky because it depends on internal hardhat details.
     return global.artifacts._provisioner._deploymentAddresses[contractName];
+  } else if (global.hardhatTestingAddresses[contractName]) {
+    // If running tests in hardhat, check if there is a testing address set.
+    return global.hardhatTestingAddresses[contractName];
```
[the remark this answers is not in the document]
chrismaree: nope. all good! it should have been added here before I believe.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
chrismaree: I belove this PR is merged.

packages/optimistic-oracle/test/index.js
```diff
@@ -43,10 +64,9 @@ contract("index.js", function() {
...
     }
 
-    // Logger should emit two logs: one indicating the start of the runner, and one about the early termination of the
-    // of the loop because the pollingDelay is set to 0.
-    assert.strictEqual(spy.getCalls().length, 2);
+    // The first log should indicate that the OO-Keeper runner started successfully
+    // and auto detected the OO's deployed address.
     assert.isTrue(spyLogIncludes(spy, 0, "OO keeper started"));
-    assert.isTrue(spyLogIncludes(spy, 1, "End of serverless execution loop - terminating process"));
+    assert.isTrue(spyLogIncludes(spy, 0, optimisticOracle.address));
```
chrismaree, line 70: why did you make this change? Surely you still want to double-check it exits at the end of the loop?

packages/optimistic-oracle/package.json
```diff
@@ -27,7 +30,7 @@
   ],
   "bin": "index.js",
   "scripts": {
-    "test": "truffle test",
+    "test": "hardhat test",
```
chrismaree, line 33: nice

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Hm yeah we need a caching solution once a bot has been running for a while. I think on startup the default START_BLOCK can be the current time, but then once its going it can store price request data in its state. WDYT

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Yeah that's an option, future PR work as I just want to get some resolution about how we query price requests in general.

packages/optimistic-oracle/test/index.js
nicholaspai: Yes good point I'll change

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: We have to think about what states should we no longer track price requests, for example once a PriceRequest is `REQUESTED` and responded to, then it should no longer be checked in the `propose()` method, but once it is `PROPOSED` and either disputed/not-disputed, then it should no longer be checked in the `dispute()` method.

packages/optimistic-oracle/test/index.js
nicholaspai: Done

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Done

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
mrice32: Sounds good! Just wanted to raise the issue.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
mrice32: As discussed IRL, IMO, I think it makes sense to look for any requests that have had "activity" within the last `n` days. We could get more specific, depending on the type of activity.

For instance, we could consider any request that:
- Has been requested within the last 1 day (arbitrary -- this could be as low as 2 hours IMO). I don't think there's any danger in "forgetting" about requested prices, since it will show up with a proposed event if someone proposes a price.
- Has been proposed within the last 2 hours. Even though the liveness can differ from the default, we generally expect that our bots will dispute bad proposals within 2 hours, so the only danger in a proposal being "forgotten" after 2 hours is if this bot is down for the entire 2 hours. Otherwise, it would dispute it.
- Has been disputed within the past 4 days (only if the bot intends to settle disputes).

The simplest solution, though, is probably just to search over the past 4 days or so (or 1 week for a more round number) for all of these event types.

nicholaspai: So it seems like one simple solution would be to parameterize a `lookback` that we can use to filter any requests that occurred (or had some other action) earlier than.

mrice32: Yeah, I think a parameterizable lookback with a sensible default is the right way to go. +1

chrismaree (approved):
This client is a great starting point. Clean well commented code. I'm glad you've split this task up the way you have into a few small steps along the way and testing each step as you go. This follows our current pattern of dealing with bots & client combinations.

image

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Done PTAL, I ended up replacing the `getFromBlock` (@pemulis 's implementation) with the block derived from the `lookbackWindow` parameter.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -0,0 +1,135 @@
...
+    const [averageBlockTime, currentBlock] = await Promise.all([
+      averageBlockTimeSeconds(),
+      this.web3.eth.getBlock("latest")
+    ]);
+    const lookbackBlocks = Math.ceil(this.lookback / averageBlockTime);
+    const earliestBlockToQuery = Math.max(currentBlock.number - lookbackBlocks, 0);
+
+    // Fetch contract state variables in parallel.
+    const [requestEvents, currentTime] = await Promise.all([
+      this.oracle.getPastEvents("RequestPrice", { fromBlock: earliestBlockToQuery }),
```
mrice32, line 77: Starting a new thread so this is more clear, I think we should consider any requests that have been requested, proposed, or disputed in this time range, since you could have a request that has no reward and therefore sits without a proposal for a long time. Once someone decides to propose a price for that request, we should consider it for disputes, etc. Thoughts?

mrice32 (commented):
Looks great! One minor comment about the lookback.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: I agree, and I was going to iterate on this on the next PR. Here's the problem how I see it:
- Every unique Proposal and Dispute event is also associated with a RequestPrice

So the way I was going to query for Proposals & Disputes was: query for all RequestPrices then call `getState()` on each one to sort it.

On the other hand, we could run event queries for Disputes and Proposals. I'm not super sure yet which strategy results in fewer web3 calls but was going to tackle this in the next issue.

mrice32: Tacking in the next PR SGTM. The issue I see with the way it's being done now is just that a request can sit for weeks (if it's unrewarded, for instance) and then get a proposal that requires input from our bot (or else could return a bad price). This is really only the case if we force our bot to only make proposals for requests that are rewarded. Otherwise, we might expect that our bot ensure no requests sit for very long.

mrice32 (approved):
LGTM!

PR #2335, opened 2020-12-22
feat(serverless-hub): add retry hub logic for rejected calls

chrismaree:
**Motivation**

https://github.com/UMAprotocol/protocol/issues/2340

Sometimes the serverless infrastructure within GCP errors out due to an internal server error or a dropped network connection between the hub and spoke. This tends to materialize through a "rejected" response between the hub and spoke. This can be difficult to debug and is a pretty non-descript error.

To decrease the number of false alerts generated, this PR adds re-try logic for rejected spoke calls.

**Summary**

After the main promise.all is resolved within the hub, the script checks if any of the calls contained a rejected call. If they did then they are added to a separate set of calls that are subsequently made from the hub.

**Details**

How spoke responses are handled is refined into a separate internal function `_processSpokeResponse` which is re-used between the initial spoke processing logic and the re-tried logic.

**Issue(s)**

close: #2340


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -99,50 +101,57 @@ hub.post("/", async (req, res) => {
...
+      _processSpokeResponse(Object.keys(configObject)[index], result, validOutputs, errorOutputs);
+    });
+    if (retriedOutputs.length > 0) {
+      logger.debug({
+        at: "ServerlessHub",
+        message: "One or more spoke calls were rejected - Retrying",
+        retriedOutputs
+      });
+    }
+    // Re-try the rejected outputs in a separate se
```
nicholaspai, line 125: "separate se" wdym here

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -99,50 +101,57 @@ hub.post("/", async (req, res) => {
...
+    let rejectedRetryPromiseArray = [];
+    retriedOutputs.forEach(botName => {
+      rejectedRetryPromiseArray.push(_executeServerlessSpoke(spokeUrl, botConfigs[botName]));
+    });
+    const rejectedRetryResults = await Promise.allSettled(rejectedRetryPromiseArray);
+    rejectedRetryResults.forEach((result, index) => {
+      _processSpokeResponse(Object.keys(configObject)[index], result, validOutputs, errorOutputs);
     });
+
+    // If there ar
```
nicholaspai, line 135: incomplete comment

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -99,50 +101,57 @@ hub.post("/", async (req, res) => {
...
-              result.reason.response.data.execResponse),
-          botIdentifier: Object.keys(configObject)[index]
-        };
-      } else {
-        validOutputs[Object.keys(configObject)[index]] = {
-          status: result.status,
-          execResponse: result.value && result.value.execResponse,
-          botIdentifier: Object.keys(configObject)[index]
-        };
+      if (result.status == "rejected") {
```
nicholaspai, line 110: So IIUC this is the main addition, where you catch rejected spoke outputs and push them into an array to retry?

chrismaree: That's right. if the spoke call is rejected then we generate a separate array that is executed below. It's only the rejected execution types that have this re-try logic. Other errors are caught as before and logged.

nicholaspai (approved):
LGTM!

PR #2339, opened 2020-12-23
feat(designated-voting)!: update designated voting to use ancillary data

mrice32:
**Motivation**

Designated voting should allow users to provide ancillary data.

**Summary**

Designated voting interface was updated to include ancillary data. This will require existing users to upgrade.

**Issue(s)**

N/A


chrismaree (approved):
Looks good to me! Adding in the Ancillary interface is simple enough & the changes are not too drastic.

PR #2342, opened 2020-12-23
improve(admin-proposal-script): Make Admin Proposal script more idiot-proof

nicholaspai:

**Motivation**

Force user to specify `gasPrice` and `nonce` so that they are less likely to submit duplicate Admin proposals or send transactions with an impractically low gas price.

Additionally, force the user to explicitly run the script in "production mode" since it is intended to be ran on mainnet. Default execution only calls `estimateGas` on the Governor contract transaction.


chrismaree (commented):
This is a pretty good solution but I don't think this logic should sit within this script as it will mean that _each_ script will need to be updated to accommodate custom gas. What about identifier umip scripts or factory scripts? 

IMHO this change should sit within the common's [TruffleConfig](ttps://github.com/UMAprotocol/protocol/blob/master/packages/common/src/TruffleConfig.js) where you can basically add the same logic to update the default gas which is set here:
https://github.com/UMAprotocol/protocol/blob/9df66ce98dad66ed110240af46f95e8c9dbc7153/packages/common/src/TruffleConfig.js#L35

Then, this `--gas` flag will work with _any_ script within the repo.

one downside with adding it to the truffleconfig is you wont be able to specify the `--nonce` flag, but I think this is fine.

nicholaspai:
> This is a pretty good solution but I don't think this logic should sit within this script as it will mean that _each_ script will need to be updated to accommodate custom gas. What about other UMIP scripts?
> 
> IMHO this change should sit within the common's TruffleConfig where you can basically add the same logic to update the default gas which is set here:
> https://github.com/UMAprotocol/protocol/blob/9df66ce98dad66ed110240af46f95e8c9dbc7153/packages/common/src/TruffleConfig.js#L35
> 
> Then, this `--gas` flag will work with _any_ script within the repo.
> 
> One downside with adding it to the truffleconfig is you wont be able to specify the `--nonce` flag, but I think this is fine. I'm curious what @mrice32 thinks about how we could override the nonce? I suspect this particular functionality will need to be applied per script.

I'm on board with adding this flag (and possible the nonce) flag to the truffle config, and I'll make the change in this script to explicitly check that those flags are set so that transactions that will take a really long time to get mined will not get sent.

mrice32:
> > This is a pretty good solution but I don't think this logic should sit within this script as it will mean that _each_ script will need to be updated to accommodate custom gas. What about other UMIP scripts?
> > IMHO this change should sit within the common's TruffleConfig where you can basically add the same logic to update the default gas which is set here:
> > https://github.com/UMAprotocol/protocol/blob/9df66ce98dad66ed110240af46f95e8c9dbc7153/packages/common/src/TruffleConfig.js#L35
> > 
> > Then, this `--gas` flag will work with _any_ script within the repo.
> > One downside with adding it to the truffleconfig is you wont be able to specify the `--nonce` flag, but I think this is fine. I'm curious what @mrice32 thinks about how we could override the nonce? I suspect this particular functionality will need to be applied per script.
> 
> I'm on board with adding this flag (and possible the nonce) flag to the truffle config.

@chrismaree @nicholaspai one idea is that we add a gas price flag to the truffle config and then we also add script(s) to speedup and/or cancel an existing transaction at a particular nonce. That's probably all you'd likely want to do, right?

nicholaspai:
> > > This is a pretty good solution but I don't think this logic should sit within this script as it will mean that _each_ script will need to be updated to accommodate custom gas. What about other UMIP scripts?
> > > IMHO this change should sit within the common's TruffleConfig where you can basically add the same logic to update the default gas which is set here:
> > > https://github.com/UMAprotocol/protocol/blob/9df66ce98dad66ed110240af46f95e8c9dbc7153/packages/common/src/TruffleConfig.js#L35
> > > 
> > > Then, this `--gas` flag will work with _any_ script within the repo.
> > > One downside with adding it to the truffleconfig is you wont be able to specify the `--nonce` flag, but I think this is fine. I'm curious what @mrice32 thinks about how we could override the nonce? I suspect this particular functionality will need to be applied per script.
> > 
> > 
> > I'm on board with adding this flag (and possible the nonce) flag to the truffle config.
> 
> @chrismaree @nicholaspai one idea is that we add a gas price flag to the truffle config and then we also add script(s) to speedup and/or cancel an existing transaction at a particular nonce. That's probably all you'd likely want to do, right?

Solves the same problem I had yeah. So far I've added the gas price flag to truffle, I'll add a script to cancel a transaction at a particular nonce (by default cancel the latest one) and speed up one (this one might be more challenging if we want to make it generalizable)

mrice32:
> > > > This is a pretty good solution but I don't think this logic should sit within this script as it will mean that _each_ script will need to be updated to accommodate custom gas. What about other UMIP scripts?
> > > > IMHO this change should sit within the common's TruffleConfig where you can basically add the same logic to update the default gas which is set here:
> > > > https://github.com/UMAprotocol/protocol/blob/9df66ce98dad66ed110240af46f95e8c9dbc7153/packages/common/src/TruffleConfig.js#L35
> > > > 
> > > > Then, this `--gas` flag will work with _any_ script within the repo.
> > > > One downside with adding it to the truffleconfig is you wont be able to specify the `--nonce` flag, but I think this is fine. I'm curious what @mrice32 thinks about how we could override the nonce? I suspect this particular functionality will need to be applied per script.
> > > 
> > > 
> > > I'm on board with adding this flag (and possible the nonce) flag to the truffle config.
> > 
> > 
> > @chrismaree @nicholaspai one idea is that we add a gas price flag to the truffle config and then we also add script(s) to speedup and/or cancel an existing transaction at a particular nonce. That's probably all you'd likely want to do, right?
> 
> Solves the same problem I had yeah. So far I've added the gas price flag to truffle, I'll add a script to cancel a transaction at a particular nonce (by default cancel the latest one) and speed up one (this one might be more challenging if we want to make it generalizable)

Agreed. Speed up may be tricky since you could have two pending transactions with the same nonce or something.

packages/core/scripts/mainnet/CancelPendingTransaction.js
```diff
@@ -0,0 +1,61 @@
...
+     *********************************/
+    const signingAccount = (await web3.eth.getAccounts())[0];
+    console.group(`Signing account: ${signingAccount}`);
+    // By default, try to cancel the latest nonce.
+    const transactionCount = await web3.eth.getTransactionCount(signingAccount);
+    console.log(`- Next nonce to be used: ${transactionCount + 1}`);
+    const nonce = argv.nonce ? argv.nonce : transactionCount + 1;
+
+    if (!argv.gasPrice) {
+      throw new Error("Please specify a --gasPrice (denominated in gWe), for example: --gasPrice 150");
```
chrismaree, line 27:
```suggestion
      throw new Error("Please specify a --gasPrice (denominated in gwei), for example: --gasPrice 150");
```

packages/core/scripts/mainnet/ProposeAdmin.js
```diff
@@ -0,0 +1,116 @@
+// This script proposes a contrived admin proposal to the DVM. Specifically, its a proposal to whitelist
```
chrismaree, line 1: +1 for the new `mainnet` directory. Are there other scripts we want to move here?

chrismaree (approved):
LGTM!

packages/core/scripts/mainnet/ProposeAdmin.js
nicholaspai: Probably yeah but let's move them lazily

PR #2344, opened 2020-12-23
improve(financial-templates-lib): Update @uma/core version to latest and fix tests and clients to correct versions

### LINKED ISSUE #2343: Refactor `financial-templates-lib` package to use latest @uma/core version
**Why**

We want to be able to add clients to this package that work with contracts in the latest version of @uma/core, such as an OptimisticOracleClient. 

**How**

All of the existing client tests will need to explicitly ask for contract ABI versions instead of grabbing them from `artifacts`.

- 3: ~1 hour

nicholaspai:

**Motivation**

#2343

**Issue(s)**

Fixes #2343


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -5,10 +5,11 @@ const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
 const { BalancerPriceFeed } = require("./BalancerPriceFeed");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
+const { getTruffleContract } = require("@uma/core");
 
-const Uniswap = require("@uma/core/build/contracts/Uniswap.json");
-const ExpiringMultiParty = require("@uma/core/build/contracts/ExpiringMultiParty.json");
-const Balancer = require("@uma/core/build/contracts/Balancer.json");
+const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", "1.2.0");
```
mrice32, line 10: The second argument to `getTruffleContract` is web3 (you can also set it to null/undefined and it will try to grab it from global.web3).

I think that's why it's not respecting your versions strings.

mrice32 (commented):
Looks awesome! Two minor comments.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
nicholaspai: ahhhh let me try that!

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -5,10 +5,11 @@ const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
...
+const Uniswap = getTruffleContract("Uniswap", "latest");
+const Balancer = getTruffleContract("Balancer", "latest");
```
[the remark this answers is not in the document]
nicholaspai: I actually think it's a bit more readable to make the `latest` version explicit

nicholaspai:
@chrismaree this PR seems to be breaking the `serverless-orchestration` tests

nicholaspai:
I don't think importing `"@uma/core": "^1.2.2"` gives the `financial-templates-lib` package access to the latest contracts in `core`. I say this because I tried changing the version string for `ExpiringMultiParty` in `ExpiringMultiPartyEventClient.js` to `latest` to see if the tests would fail, but they did not. Additionally, I tried merging this branch into #2330 to no luck.

mrice32:
> I don't think importing `"@uma/core": "^1.2.2"` gives the `financial-templates-lib` package access to the latest contracts in `core`. I say this because I tried changing the version string for `ExpiringMultiParty` in `ExpiringMultiPartyEventClient.js` to `latest` to see if the tests would fail, but they did not. Additionally, I tried merging this branch into #2330 to no luck. 
> 
> 
> 
> It looks like the contract imported by `global.artifacts.require(...)` is not the latest version, and as a test I tried inserting a `global.artifacts.require("OptimisticOracle")` [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/index.js#L89) which threw an error.
> 
> 
> 
> @chrismaree @mrice32 any ideas?

Just to triple check, you reran yarn every time you changed the package.json, right?

nicholaspai:
> > I don't think importing `"@uma/core": "^1.2.2"` gives the `financial-templates-lib` package access to the latest contracts in `core`. I say this because I tried changing the version string for `ExpiringMultiParty` in `ExpiringMultiPartyEventClient.js` to `latest` to see if the tests would fail, but they did not. Additionally, I tried merging this branch into #2330 to no luck.
> > It looks like the contract imported by `global.artifacts.require(...)` is not the latest version, and as a test I tried inserting a `global.artifacts.require("OptimisticOracle")` [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/index.js#L89) which threw an error.
> > @chrismaree @mrice32 any ideas?
> 
> Just to triple check, you reran yarn every time you changed the package.json, right? @nicholaspai

Yep

packages/core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uma/core",
-  "version": "1.2.0",
+  "version": "1.2.2",
```
nicholaspai, line 3: h/t @mrice32 who pointed out that this version needs to be updated so that the import of `@uma/core^1.2.2` from `financial-templates-lib` resolves to the `core` directory and not from `financial-templates-lib/node_modules`

packages/financial-templates-lib/package.json
```diff
@@ -35,7 +35,7 @@
     "/src/**/*.js"
   ],
   "scripts": {
-    "test": "hardhat test ./test/clients/* ./test/helpers/* ./test/price-feed/* && truffle test ./test/truffle"
+    "test": "hardhat test ./test/clients/* ./test/helpers/* ./test/price-feed/* && truffle test ./test/truffle/*"
```
chrismaree, line 38: thanks for fixing this

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -3,17 +3,20 @@ const { parseFixed } = require("@ethersproject/bignumber");
 const winston = require("winston");
 
 const { interfaceName, MAX_UINT_VAL, ZERO_ADDRESS } = require("@uma/common");
+const { getTruffleContract } = require("@uma/core");
+
+const ABI_VERSION = "1.2.0";
```
chrismaree, line 8: is `ABI_VERSION` a correctly descriptive variable name? It's not _just_ the ABIs that are changing between versions; it's the whole contract + the bytecode & implementation details. Would `CONTRACT_VERSION` not be more descriptive?

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -3,17 +3,20 @@ const { parseFixed } = require("@ethersproject/bignumber");
...
 
-const ExpiringMultiParty = artifacts.require("ExpiringMultiParty");
-const Finder = artifacts.require("Finder");
-const IdentifierWhitelist = artifacts.require("IdentifierWhitelist");
-const MockOracle = artifacts.require("MockOracle");
-const Token = artifacts.require("ExpandedERC20");
-const SyntheticToken = artifacts.require("SyntheticToken");
-const Timer = artifacts.require("Timer");
-const Store = artifacts.require("Store");
+const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, ABI_VERSION);
```
chrismaree, line 12: Is the idea to eventually ripple this change through all the tests? I'm curious how we see this scaling. For example, if we make ABI or logic breaking changes between upgrades (as we've done with the most recent update to the EMP with withdraw liquidation logic) we would need to add some kind of branching logic within the tests to accommodate different sections based on the ABI version. This is with a _small_ change to the contracts. what happens when there is a larger difference between versions? 

I think this solution _might_ be ok for now but I'm hesitant to say this is the path we should go down for multi-version testing before we have a better strategy & plan for supporting these kinds of tests within the repo. As an example, It might be better to have separate test files for each version to make it clear what is being tested and to remove the need for any branching logic within the tests.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: hm yeah I think that'd be more accurate I'll make the change

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: I don't think I'm making a statement about how we should handle versioning within the EMPClient, in fact I'm hereby assuming that the EMPClient should only work with the 1.2 version. I think a subsequent change to the EMPClient should be made to change logic based on the versioning, but in a scalable smart way.

What do you think about placing ABI/CONTRACT_VERSION in the `ExpiringMultiPartyClient.js` file itself, exporting it as a variable, and importing that variable in the test? This way it is more clear what version the client is supposed to be working with?

chrismaree: ye, that could work! There will be situations wherein the client can support multiple versions; say for example we release 1.3.0 with the new contracts (when they are on main net). We could then define the EMPClient to support `1.2.x` & `1.3.x`, with some small internal branching logic to deal with both cases. In this scenario, you could pull the "supported" versions from the client within the test and then check that they overlap with those used within the unit tests. we might then want to have a `1.2.x` section of tests and a separate section that tests `1.3.x`.

What do you think? I'm also curious for input on this point from @UMAprotocol/eng as this is the first example of where we need to consider versioning into the future and how we want to interplay within tests.

packages/core/package.json
mrice32: This is merged into master, so you should be able to merge/rebase and this part of the diff should disappear.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
mrice32: I generally agree with @nicholaspai that this PR is just tweaking how the versions are accessed and making it more explicit. It's unopinionated wrt how multiple versions would be handled in the client/tests.

mrice32: @chrismaree i think the structure of the versioning tests depends on how different the versions are. For very similar versions, you could imagine wanting to run the exact same tests just with the version swapped. For more branch-y behavior, you might want one or two unique tests or even a separate test suite. Thoughts?

mrice32 (approved):
LGTM once the comment threads are resolved!

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: > We could then define the EMPClient to support 1.2.x & 1.3.x, with some small internal branching logic to deal with both cases.

I think we should build internal branching logic yes.

I also think eventually we'll want to export, from all clients, the version strings that they are compatible with (e.g. EMPClient compatible with <= 1.2.2 and OOClient compatible with >= 1.2.5), and their corresponding unit tests should pass for all eligible verrsions (tbd how unit tests for different versions will work).

Additionally, we'll want to enforce within the downstream packages, like `liquidator` and `disputer`, that they are importing the correct versions to work with the clients. For example, we don't want the liquidator to use `@uma/core: 1.2.7` with the EMPClient if the EMPClient is only compatible with <= 1.2.5. TBD how we enforce this and whether we can do it dynamically.

That being said, these are all tangential to this PR and IMO are not features that should block it. This PR is mostly just bumping the `financial-templates-lib` package's @uma/core version to latest so that I can put an OptimisticOracleClient in it.

chrismaree: That all sounds good & you are right this is unrelated to this PR. Did not mean to block it from going through! LGTM!

chrismaree (approved):

daywiss (approved):

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -5,12 +5,12 @@ const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
...
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
-
-const Uniswap = require("@uma/core/build/contracts/Uniswap.json");
-const ExpiringMultiParty = require("@uma/core/build/contracts/ExpiringMultiParty.json");
-const Balancer = require("@uma/core/build/contracts/Balancer.json");
+const { getTruffleContract } = require("@uma/core");
 
 async function createPriceFeed(logger, web3, networker, getTime, config) {
+  const Uniswap = getTruffleContract("Uniswap", web3, "latest");
+  const Balancer = getTruffleContract("Balancer", web3, "latest");
```
daywiss, line 12: small nit here, wont block pr with this, but the tests offer a way to change contract version, though its hard coded here. does this cause a problem if tests specify something other than latest? its also fine if latest is always to be assumed, that could be reasonable.

nicholaspai: > does this cause a problem if tests specify something other than latest?

Yes this does cause a problem if the tests specify a different version, this is why I think we [should explore exporting the "valid" versions from the client to the tests](https://github.com/UMAprotocol/protocol/pull/2344#discussion_r549693664), so that the tests are only testing the expected versions.  Do you think thats a good approach?

daywiss: im not sure actually... my gut instinct is i probably wouldn't want to do it that way, but i could be missing something since this is a broader convo. I feel like we should treat tests, software and contracts as part of the same versioned package. That way as contracts evolves, tests and supporting software can evolve with it without worry of breaking tests for some other version of the software.  Probably too little too late for something like this and you probably have better insight as to what a workable solution is at this point, so i would defer to your judgement here.

Ideally though it would be something that doesn't add increasing maintenance cost as the number of versions go up.

daywiss (approved):

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
nicholaspai: > I feel like we should treat tests, software and contracts as part of the same versioned package. That way as contracts evolves, tests and supporting software can evolve with it without worry of breaking tests for some other version of the software.

This I 100% agree with. I guess herein lies the problem:
- `liquidator` package imports `@uma/core:1.2.2` and uses that version import a contract ABI to create a new `ExpiringMultiPartyClient`
- `ExpiringMultiPartyClient` technically does not restrict which ABI version can be imported, but it will break if an ABI > `1.2.2` is imported. Accordingly, we explicitly set the contract ABI's to `1.2.2` in the `ExpiringMultiPartyClient` tests.

After writing this out, one solution would be to assert within the `ExpiringMultiPartyClient` that the inputted ABI is the correct version. I'm not sure there is a way to check an ABI "version" (in the @uma/core sense of the word), but I think within the client is the logical place to restrict versions.

Thoughts @UMAprotocol/eng ?

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: Hey @chrismaree @mrice32 can you check out this [comment](https://github.com/UMAprotocol/protocol/pull/2344#discussion_r550182007) related to this thread.

PR #2346, opened 2020-12-23
feat(financial-templates): adds a new price feed for the DefiPulseTotalTVL synthetic tokens

bryanjcampbell1:
**Title**

  feat(financial-templates): adds a new price feed for the DefiPulseTotalTVL synthetic tokens

**Motivation**

UMIP 24 introduces DefiPulseTotalTVL as a price id.  The price feed suggested in this PR would allow for the creation of liquidation and dispute bots for the DefiPulseTotalTVL price id.

**Summary**

Adds the DefiPulseTotalTVLPriceFeed.js file to the financial-templates repo

**Details**

This price feed was modified from, and closely follows, the CryptoWatchPriceFeed.  The cryptowatch api has been replaced with the [DeFi Pulse Data API.](https://docs.defipulse.com/)

**Issue(s)**

Fixes the current lack of a price feed for the DefiPulseTotalTVL price id.


packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const { parseFixed } = require("@ethersproject/bignumber");
+
+// An implementation of PriceFeedInterface that uses DefiPulse Data api to retrieve prices.
+class DEFI_PULSE_TOTAL_TVL_PriceFeed extends PriceFeedInterface {
```
chrismaree, line 5: nit: while the price feed is indeed called `DEFI_PULSE_TOTAL_TVL` This class should not be named like this. Ideally, it should be called `DefiPulsePriceFeed` I believe.

chrismaree, line 5: nit: while the price feed is indeed called DEFI_PULSE_TOTAL_TVL This class should not be named like this. Ideally, it should be called DefiPulsePriceFeed I believe.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
...
+  }
+
+  getHistoricalPrice(time, verbose = false) {
+    if (this.lastUpdateTime === undefined) {
+      return undefined;
+    }
+
+    let closestTime = {timestamp:0, value:0};
+
+    //go through all values and find time that that is the largest and still less than 'time'
```
chrismaree, line 71: nit: here (and in other places) it does not seem you've linted your code correctly. We normally have spaces after `//` in the comments. You should try running the linter with `yarn lint-fix` to address these formatting issues.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
...
+    getTime,
+    minTimeBetweenUpdates,
+    decimals = 18,
+
+  ) {
+    super();
+    this.logger = logger;
+    this.web3 = web3;
+
+    this.apiKey = apiKey;
```
chrismaree, line 32: you might want to verify that there is indeed an API key added in here. If there is not, then this price feed will error out. This differs from the `CryptoWatchPriceFeed` as in those feeds the API key is optional whereas here it is a requirement.

You can simply do this with an `assert(apiKey, "requires apiKey for Defi Pulse");` after importing `const assert = require("assert");`

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
...
+      let past = this.historicalPrices[i].timestamp;
+      let val = this.historicalPrices[i].value;
+
+      if( (past > closestTime.timestamp) && (past < time)){
+
+        closestTime.timestamp = past;
+        closestTime.value = val;
+
+      }
+
```
chrismaree, line 83: similar nit about formatting. Should be resolved once you run the linter.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
...
+      currentTime: currentTime,
+      lastUpdateTimestamp: this.lastUpdateTime
+    });
+
+    // 1. Construct URLs.
+    const priceUrl =
+    `https://data-api.defipulse.com/api/v1/defipulse/api/MarketData?api-key=${this.apiKey}`;
+
+    // 2. Send requests.
+    const [priceResponse] = await Promise.all([
```
chrismaree, line 124: nit: no point in using `Promise.all` synthax here seeing you are only sending out one async call.

chrismaree (commented):
This looks like a great starting point! I've left some nits on formatting and input validation.

You should try to write some unit tests the same way we have unit tests for the cryptoWatchPriceFeed which you can see here: https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js

Also, to get this feed 100% integrated into this repo we will need an integration made to the [CreatePriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js) file. You can see the similar setup in this file for CryptoWatch here: https://github.com/UMAprotocol/protocol/blob/898a6fef1aaa9474307b4b51a508dd3af4c68612/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js#L14

This will also contain some input validation using the `isMissingField` method, specifically for the `apiKey`.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
```diff
@@ -0,0 +1,164 @@
...
+    this.logger.debug({
+      at: "DEFI_PULSE_TOTAL_TVL_PriceFeed",
+      message: "Updating",
+      currentTime: currentTime,
+      lastUpdateTimestamp: this.lastUpdateTime
+    });
+
+    // 1. Construct URLs.
+    const priceUrl =
+    `https://data-api.defipulse.com/api/v1/defipulse/api/MarketData?api-key=${this.apiKey}`;
```
mrice32, line 121: Is there any way to get historical data from defi pulse? It seems as if this class is depending on caching historical data over time internally, but generally the bots expect to be able to immediately access historical data up to the lookback. Does DefiPulse offer any API for immediate historical data?

mrice32 (commented):
This looks awesome! One comment about caching of historical data. It would also be great to see a few tests added to verify the behavior of this class!

Thanks for working on this!

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
bryanjcampbell1: Using DefiPulseTotalPriceFeed so I can differentiate between it and the SUSHI/UNI TVL price feed in the future

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
bryanjcampbell1: Is there a config file with a preferred linter settings somewhere in the repo? Im getting Command "lint-fix" not found when running yarn lint-fix

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
bryanjcampbell1: There is a GetHistory endpoint. How should I go about using it? Call it from getCurrentPrice() and store it in state so it can be referenced in getHistoricalPrice(time)?

bryanjcampbell1: Modifying the CryptoWatchPriceFeed tests now. I looked more into the historical endpoint but the timestamps are a full day apart.  I imagine the bots need more precision than this.  Also, are there preferred ranges for  lookback period and minTimeBetweenUpdates? Defi pulse only updates their api hourly.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
chrismaree: did you install all the packages within the repo by running `yarn` in the root of the repo? it should work fine once you have done such.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -38,29 +39,30 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.decimals, // This defaults to 18 unless supplied by user
       config.ohlcPeriod // This defaults to 60 unless supplied by user
     );
-  } else if (config.type === "uniswap") {
```
chrismaree, line 41: why did you remove the existing `uniswap` feed? you should rather add your feed as another option rather than removing an existing feed that is used in production.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,163 @@
...
+
+    this.removeValuesOlderThanLookback = () => {
+
+        const currentTime = this.getTime();
+
+        this.historicalPrices = this.historicalPrices.filter(priceObj => priceObj.timestamp > (currentTime - lookback) );
+
+    }
+
+
```
chrismaree, line 58: this file still needs to be linted

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,151 @@
...
+    defiPulseTotalPriceFeed = new DefiPulseTotalPriceFeed(
+      dummyLogger,
+      web3,
+      apiKey,
+      lookback,
+      networker,
+      getTime,
+      minTimeBetweenUpdates
+    );
+
```
chrismaree, line 55: this file also needs to be linted.

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,151 @@
...
+require('dotenv');
+
+contract("DefiPulseTotalPriceFeed.js", function() {
+  let defiPulseTotalPriceFeed;
+  let mockTime = 1538018200;
+  let networker;
+
+  const apiKey = process.env.DEFIPULSE_API_KEY;
+
+  const lookback = 3600*24; //???????????????
```
chrismaree, line 13: what's up with this comment?

bryanjcampbell1: This was a note to myself that I was unclear about an appropriate lookback time.  The API only changes at hourly intervals so the bots would run less frequently than others like cryptowatch.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: I tried running yarn and then yarn lint-fix on the repo but got "lint-fix" not found. a) what linter should I be using? b) is there a config file for the linter somewhere in the repo?

packages/financial-templates-lib/src/price-feed/DefiPulseTotalTVLPriceFeed.js
bryanjcampbell1: Got it. Thanks

bryanjcampbell1:
I made significant changes to the way the price feed the historical record.  I am no longer building it by caching as I was able to get hourly values from the api.  I am a little confused about the unit test though. How should I structure the fake data to inject? Also, how do I run the test?  I tried calling 'truffle test' in root but just got a '0 passing' result.

chrismaree:
> I made significant changes to the way the price feed the historical record. I am no longer building it by caching as I was able to get hourly values from the api. I am a little confused about the unit test though. How should I structure the fake data to inject? Also, how do I run the test? I tried calling 'truffle test' in root but just got a '0 passing' result.

looks like you got the linter working, which is great!

To run your test file you should navigate to the `financial-templates-lib` directory then run:
```
yarn hardhat test ./test/price-feed/DefiPulseTotalPriceFeed.js 
```

seeing you also changed the create price feed you should also test:
```
yarn hardhat test ./test/price-feed/createPriceFeed.js 
```

I'll get your tests to run in CI now as well.

bryanjcampbell1:
Got the unit tests working! Do I need to worry about the git signoff errors? Running "git rebase HEAD~11 --signoff" returns "unknown option `signoff'"

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,144 @@
...
+    // Get tvlUSD for most most recent timestamp
+    let mostRecent = { timestamp: 0, tvlUSD: 0 };
+    for (let i = 0; i < response.length; i++) {
+      if (Number(response[i].timestamp) > mostRecent.timestamp) {
+        mostRecent.timestamp = Number(response[i].timestamp);
+        mostRecent.tvlUSD = Number(response[i].tvlUSD);
+      }
+    }
+
+    const newPrice = this.web3.utils.toWei((mostRecent.tvlUSD / 1000000000).toFixed(6), "ether");
```
mrice32, line 133: Can you provide a comment on what's happening here?

Also, do you think this should be a class function since you do the same operation in the getHistoricalPrice method?

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,144 @@
...
+      let past = this.historicalPrices[i].timestamp;
+      let val = this.historicalPrices[i].tvlUSD;
+
+      if (past > closestTime.timestamp && past < time) {
+        closestTime.timestamp = past;
+        closestTime.tvlUSD = val;
+      }
+    }
+
+    const historicalPrice = this.web3.utils.toWei((closestTime.tvlUSD / 1000000000).toFixed(6), "ether");
```
mrice32, line 64: Does this respect the `decimals` value that is passed in the constructor?

mrice32 (commented):
Looks awesome -- a few minor comments, but I think this is just about ready!!

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: I removed decimals from the constructor.  The UMIP does specify 3 decimals of precision though so I made that change.  What I want is that when tvlUSD =  24781234567, the token price to be $24.781 or 24781000000000000000 wei of the collateral token.

bryanjcampbell1: Am I understanding the price feed correctly? It should be returning a BN in wei correct?

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,141 @@
...
+      let past = this.historicalPrices[i].timestamp;
+      let val = this.historicalPrices[i].tvlUSD;
+
+      if (past > closestTime.timestamp && past < time) {
+        closestTime.timestamp = past;
+        closestTime.tvlUSD = val;
+      }
+    }
+
+    const historicalPrice = this.web3.utils.toWei((closestTime.tvlUSD / 1000000000).toFixed(3), "ether");
```
chrismaree, line 58: Seeing you do this computation two times (`this.web3.utils.toWei((closestTime.tvlUSD / 1000000000).toFixed(3), "ether");`) you could consider refactoring this into a helper method within this feed that is called in two places. This way you could add a more detailed comment around the umip definition and help with code re-use.

chrismaree: Also, you should consider adding `this.toWei` within the constructor, the same way you pulled out `this.toBN` on line 31. Then, you wont need to call `this.web3.utils.toWei(...` and can rather refer to `this.toWei`

chrismaree:
> Got the unit tests working! Do I need to worry about the git signoff errors? Running "git rebase HEAD~11 --signoff" returns "unknown option `signoff'"

you don't need to worry about the signoff part, we can set this from our side :)

Also looks like you need to lint again! normally we have a husky hook that should lint your code automatically whenever you commit. If you using something other than the console to commit it's possible this is not enabled.

I will re-run your tests through CI as well.

bryanjcampbell1:
What should my next steps be? I have run the cryptowatch bots on google cloud compute but I'm not sure how to configure to use my price feed.

mrice32:
@bryanjcampbell1 sorry that I'm just coming back to this PR. Side note/quick tip: once reviewers send a round of comments and you've responded and would like them to take another look, you can click the cycle button next to their names to request them again! This basically adds it back to each of our personal review queues so we see and come back to the PR. See our contributing guidelines for more info on our typical process: https://github.com/UMAprotocol/protocol/blob/master/CONTRIBUTING.md#prs. 

Screen Shot 2021-02-01 at 12 52 21 AM

mrice32:
@bryanjcampbell1 @chrismaree Actually, you _do_ need to signoff your commits. We need outside contributors to complete the DCO for legal reasons. It's how we show that you really agree to submit your code under the repo license and forgo your individual rights to the code.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,145 @@
...
+    return this.lastUpdateTime;
+  }
+
+  async update() {
+    const currentTime = this.getTime();
+
+    // Return early if the last call was too recent.
+    if (this.lastUpdateTime !== undefined && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+      this.logger.debug({
+        at: "DEFI_PULSE_TOTAL_TVL_PriceFeed",
```
chrismaree, line 76: nit: you should make this `at` the same as the file name. I know this feed was previous all upper case like this but that's no longer the case.
```suggestion
        at: "DefiPulseTotalPriceFeed",
```

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,145 @@
...
+        message: "Update skipped because the last one was too recent",
+        currentTime: currentTime,
+        lastUpdateTimestamp: this.lastUpdateTime,
+        timeRemainingUntilUpdate: this.lastUpdateTimes + this.minTimeBetweenUpdates - currentTime
+      });
+      return;
+    }
+
+    this.logger.debug({
+      at: "DEFI_PULSE_TOTAL_TVL_PriceFeed",
```
chrismaree, line 86:
```suggestion
      at: "DefiPulseTotalPriceFeed",
```

yarn.lock
```diff
@@ -4677,6 +4677,15 @@
     "@uma/common" "^1.1.0"
```
chrismaree, line 4677: not sure why there is a diff here?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -59,7 +60,32 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.twapLength,
       config.lookback,
       getTime,
-      config.invertPrice // Not checked in config because this parameter just defaults to false.
+      config.invertPrice, // Not checked in config because this parameter just defaults to false.
+      config.poolDecimals,
+      config.priceFeedDecimals // This defaults to 18 unless supplied by user
```
chrismaree, line 65: why is this showing up as a diff? I'm pretty sure this is what the actual `UniswapPriceFeed` looks like.

chrismaree:
This looks really good to me and super close to being done! 

note that you still need to add in an additional set of changes in the `DefaultPriceFeedConfigs` file: https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js

Also, what is the intended collateral for this synthetic? importantly, if it has non 18 decimals you will also need to add it to `packages/common/src/PriceIdentifierUtils.js` which can be found here: https://github.com/UMAprotocol/protocol/blob/master/packages/common/src/PriceIdentifierUtils.js

This ensures that the decimals in the price requests returned by the dvm is scalled according to the collateral type defined for the specific synthetic.

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,87 @@
...
+const winston = require("winston");
+var Web3 = require("web3");
+var web3 = new Web3("http://localhost:8545");
+
+contract("DefiPulseTotalPriceFeed.js", function() {
+  let defiPulseTotalPriceFeed;
+  let mockTime = 1611583300;
+  let networker;
+
+  const apiKey = process.env.DEFIPULSE_API_KEY;
```
chrismaree, line 13: unit tests should not rely on having this env set.

chrismaree:
I just tried to run your tests and they failed as follows:
image

Ideally, the tests should be independent of an API key and solely rely on injected data from networker. An option could be to set it to something within the unit tests so you get past the `assert(apiKey, "requires apiKey for Defi Pulse");` and then test with the mocked data from the networker.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: My guess is that there are some merge issues, where these changes accidentally got removed in the merge process, but then were re-added in this branch to get the tests to pass.

mrice32:
@bryanjcampbell1 I updated the title to match the title in your description, just FYI!

mrice32:
@bryanjcampbell1 Responding to this comment above (posting here because the comment is way up there and now outdated because you've made some updates :)): https://github.com/UMAprotocol/protocol/pull/2346#discussion_r564119135:

- @chrismaree is a decimals input value required or is it okay for the price feed to always use the same decimals value? I assume the latter?
- @smb2796 it appears as if [UMIP-24](https://github.com/UMAprotocol/UMIPs/blob/1bc93c0c032f0dbd440f6f42bf5f4d8b9eb150d0/UMIPs/umip-24.md) does not specify a collateral decimals value. Should we amend it to add one? @bryanjcampbell1 what collateral are y'all planning to use? (not necessary if you're planning to use a currency with 18 decimals).

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,145 @@
...
+    this.currentPrice = newPrice;
+    this.historicalPrices = response;
+  }
+
+  scaleResult(_tvlUSD) {
+    // As described in UMIP 24
+    // In an effort to make the token price affordable, the value of the token is the tvlUSD divided by 1 billion.
+    // We also cut off precision after 3 decimals to match the specified price step of .001
+
+    return this.toWei((_tvlUSD / 1000000000).toFixed(3), "ether");
```
mrice32, line 139: nit for readability:
```suggestion
    const billion = 1_000_000_000;
    return this.toWei((_tvlUSD / billion).toFixed(3), "ether");
```

mrice32 (commented):
Looks awesome!

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: Changed.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: Changed.

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: Changed to a test string

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
bryanjcampbell1: Changed.

bryanjcampbell1: Do most synths like this use USDC or TUSD? I had originally thought of using USDC (6 decimals).

bryanjcampbell1: Any tips for signing off on old commits? Just set up a GPG keypair. Should be good signing off from here on.

bryanjcampbell1:
What should I be using for my 'type' instead of 'medianizer'? I only have one source, the defipulse api, to get my value.

mrice32:
> What should I be using for my 'type' instead of 'medianizer'? I only have one source, the defipulse api, to get my value.

defipulsetotal would be your type.

smb2796:
> it appears as if UMIP-24 does not specify a collateral decimals value. Should we amend it to add one?

@mrice32 yes, if @bryanjcampbell1 wants to use USDC, we should amend the UMIP. @bryanjcampbell1 can you confirm, and I can amend?

bryanjcampbell1:
Can confirm you can amend the UMIP to include USDC.

bryanjcampbell1:
> > it appears as if UMIP-24 does not specify a collateral decimals value. Should we amend it to add one?
> 
> @mrice32 yes, if @bryanjcampbell1 wants to use USDC, we should amend the UMIP. @bryanjcampbell1 can you confirm, and I can amend?

Yes, I can confirm you can amend the UMIP to include USDC.

packages/common/src/PriceIdentifierUtils.js
```diff
@@ -6,7 +6,8 @@ const IDENTIFIER_BLACKLIST = {
 // Price identifiers that should resolve prices to non 18 decimal precision. Any identifiers
 // not on this list are assumed to resolve to 18 decimals.
 const IDENTIFIER_NON_18_PRECISION = {
-  USDBTC: 8
+  USDBTC: 8,
+  USDC: 6
```
mrice32, line 10:
```suggestion
  DEFI_PULSE_TOTAL_TVL: 6
```

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,146 @@
...
+    this.lookback = lookback;
+    this.networker = networker;
+    this.getTime = getTime;
+    this.minTimeBetweenUpdates = minTimeBetweenUpdates;
+
+    this.toWei = this.web3.utils.toWei;
+
+    this.historicalPrices = [];
+  }
+
```
mrice32, line 34: @chrismaree can correct me if I'm wrong, but I think you should make the `decimals` value (i.e. how many decimals you scale your output to) configurable with a constructor parameter.

You also need to add a `getPriceFeedDecimals()` method that returns the number of decimals that the price feed is outputting (which is a part of the PriceFeedInterface on master: https://github.com/UMAprotocol/protocol/blob/ec5908e93c72e547b94b2ccf675b3c5d94ad7b93/packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js#L37).

mrice32 (commented):
Can you rebase or merge master into your branch? Sorry that there are a lot of conflicts -- there's been a lot of activity in our repo lately :/. If you have any issues merging, I'm happy to help!

Once you merge, I think this is almost ready to ship! 🚀

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,146 @@
...
+    this.historicalPrices = response;
+  }
+
+  scaleResult(_tvlUSD) {
+    // As described in UMIP 24
+    // In an effort to make the token price affordable, the value of the token is the tvlUSD divided by 1 billion.
+    // We also cut off precision after 3 decimals to match the specified price step of .001
+
+    const billion = 1_000_000_000;
+    return this.toWei((_tvlUSD / billion).toFixed(3), "ether");
```
mrice32, line 140: @chrismaree can correct me if I'm wrong, but I think you should make the `decimals` value (i.e. how many decimals you scale your output to) configurable with a constructor parameter.

You also need to add a `getPriceFeedDecimals()` method that returns the number of decimals that the price feed is outputting (which is a part of the PriceFeedInterface on master: https://github.com/UMAprotocol/protocol/blob/ec5908e93c72e547b94b2ccf675b3c5d94ad7b93/packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js#L37).

chrismaree: yup! it should be configurable with a constructor parameter that then gets defined in the `CreatePriceFeed` in the same way that the other feeds have their decimals fed in. See the code below for an example:
https://github.com/UMAprotocol/protocol/blob/c91e11bad264509efd4ef98044e448e6e5b8b5f0/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js#L40

And @mrice32 is right about the interface for `getPriceFeedDecimals`

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -59,7 +60,32 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    );
+  } else if (config.type === "defipulsetotal") {
```
chrismaree, line 67: should the name of this feed not be `defipulsetvl` to be consistent?

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,87 @@
...
+    networker.getJsonReturns = [...validResponses];
+
+    await defiPulseTotalPriceFeed.update();
+
+    // Before period 1 should return null.
+    assert.equal(defiPulseTotalPriceFeed.getHistoricalPrice(1611572399), null);
+    assert.equal(defiPulseTotalPriceFeed.getHistoricalPrice(1611575900).toString(), web3.utils.toWei("24.78"));
+    assert.equal(defiPulseTotalPriceFeed.getHistoricalPrice(1611579500).toString(), web3.utils.toWei("23.50"));
+    assert.equal(defiPulseTotalPriceFeed.getHistoricalPrice(1611579500).toString(), web3.utils.toWei("23.50"));
+    assert.equal(defiPulseTotalPriceFeed.getHistoricalPrice(1611579601).toString(), web3.utils.toWei("22.25"));
```
mrice32, line 65:
```suggestion
    // Before period 1 should throw
    assert.isTrue(await defiPulseTotalPriceFeed.getHistoricalPrice(1611572399).catch(() => true));
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611575900)).toString(), web3.utils.toWei("24.78"));
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), web3.utils.toWei("23.50"));
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), web3.utils.toWei("23.50"));
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579601)).toString(), web3.utils.toWei("22.25"));
```

mrice32 (commented):
Looks awesome! There are a few minor updates to the price feed api that we've made recently: 1) getHistoricalPrice is now async and 2) getHistoricalPrice is supposed to throw when the price is not available instead of returning undefined. I have made suggestions to update these. Additionally, you'll need to add a configurable decimals parameter.

bryanjcampbell1:
I am a little confused on the decimals param. In the price feed does it refer to the decimals of the collateral token or the synth?

mrice32:
> I am a little confused on the decimals param. In the price feed does it refer to the decimals of the collateral token or the synth?

@chrismaree can probably give a better explanation than I can on what this param represents and how it's used.

chrismaree:
> I am a little confused on the decimals param. In the price feed does it refer to the decimals of the collateral token or the synth?

Ye, the decimals part is a bit confusing! There are three different kinds of decimals to consider with a synthetic token:
1) the decimals of the collateral currency
2) the decimals of the synthetic token (the minted ERC20)

.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,36 @@
+<component name="ProjectCodeStyleConfiguration">
```
mrice32, line 1: nit: there is still a top-level .idea folder at root.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,151 @@
...
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {String} apiKey DeFiPulse Data API key. Note: these API keys are rate-limited.
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the current time.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   */
+  constructor(logger, web3, apiKey, lookback, networker, getTime, minTimeBetweenUpdates, priceFeedDecimals) {
```
mrice32, line 17: Please see @chrismaree's comments around providing a priceFeedDecimals param that defaults to 18 and allows the user to output values with different numbers of decimals based on that param. See the cryptowatch price feed for an example of how this can work.

mrice32 (commented):
A few more comments! Also please see @chrismaree's answer to your question above!

bryanjcampbell1:
Could you recommend a way for me to signoff on my old commits? I tried following [this](https://gist.github.com/kwk/d70f20d17b18c4f3296d) but ended up overwriting signatures from Sean, Matt, ect. Spent at least 2 hours on this lol.

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,87 @@
...
+    );
+  });
+
+  it("No update", async function() {
+    assert.equal(defiPulseTotalPriceFeed.getCurrentPrice(), undefined);
+    assert.isTrue(await defiPulseTotalPriceFeed.getHistoricalPrice(1000).catch(() => true));
+    assert.equal(defiPulseTotalPriceFeed.getLastUpdateTime(), undefined);
+  });
+
+  it("Basic historical price", async function() {
```
mrice32, line 54: Can you add a test for non-18 decimals just to verify that that scaling function you added works correctly?

mrice32 (commented):
> Could you recommend a way for me to signoff on my old commits? I tried following this but ended up overwriting signatures from Sean, Matt, ect. Spent at least 2 hours on this lol.

Yep! Note: please back up your branch before you do this: `git checkout -b backup_branch`. Then switch back to the branch you've been using.

This is what I'd recommend as a simple way around these issues. Note: I'm using `upstream/master` as your local reference to our master branch. Just swap it out with whatever branch you typically merge into your branch to pull in upstream changes. Note: please make sure that whatever local reference you're using to refer to _our_ master branch has already been merged into your branch. If not, you'll get weird changes in the final diff.
- `git reset upstream/master` -- this destroys the git history on your branch and makes it match upstream/master, BUT it keeps all your changes as unstaged changes.
- `git add .` stages _all_ of the changes at once.
- `git commit -s -m "feat(financial-templates): adds a new price feed for the DefiPulseTotalTVL synthetic tokens"` creates a single signed commit on your branch that contains _all_ of your changes.
- `git push -f` force overwrites the remote branch on github.

Note: make _sure_ you backup everything in a place where you can retrieve it if something breaks in this process. Since you're wiping your git history on your branch, you could wind up accidentally wiping your changes, so be sure you have a way to start over if things go bad!

bryanjcampbell1:
> > Could you recommend a way for me to signoff on my old commits? I tried following this but ended up overwriting signatures from Sean, Matt, ect. Spent at least 2 hours on this lol.
> 
> Yep! Note: please back up your branch before you do this: `git checkout -b backup_branch`. Then switch back to the branch you've been using.
> 
> This is what I'd recommend as a simple way around these issues. Note: I'm using `upstream/master` as your local reference to our master branch. Just swap it out with whatever branch you typically merge into your branch to pull in upstream changes. Note: please make sure that whatever local reference you're using to refer to _our_ master branch (`upstream/master` in the example below) has already been merged into your branch. If not, you'll get weird changes in the final diff.
> 
> * `git reset upstream/master` -- this destroys the git history on your branch and makes it match upstream/master, BUT it keeps all your changes as unstaged changes.
> * `git add .` stages _all_ of the changes at once.
> * `git commit -s -m "feat(financial-templates): adds a new price feed for the DefiPulseTotalTVL synthetic tokens"` creates a single signed commit on your branch that contains _all_ of your changes.
> * `git push -f` force overwrites the remote branch on github.
> 
> Note: make _sure_ you backup everything in a place where you can retrieve it if something breaks in this process. Since you're wiping your git history on your branch, you could wind up accidentally wiping your changes, so be sure you have a way to start over if things go bad!

Thanks! This was a huge help!

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,152 @@
+const { PriceFeedInterface } = require("./PriceFeedInterface");
```
mrice32, line 1:
```suggestion
const { parseFixed } = require("@uma/common");
const { PriceFeedInterface } = require("./PriceFeedInterface");
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,91 @@
+const { DefiPulseTotalPriceFeed } = require("../../src/price-feed/DefiPulseTotalPriceFeed");
```
mrice32, line 1:
```suggestion
const { parseFixed } = require("@uma/common");
const { DefiPulseTotalPriceFeed } = require("../../src/price-feed/DefiPulseTotalPriceFeed");
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,91 @@
...
+var web3 = new Web3("http://localhost:8545");
+
+contract("DefiPulseTotalPriceFeed.js", function() {
+  let defiPulseTotalPriceFeed;
+  let mockTime = 1611583300;
+  let networker;
+
+  const apiKey = "test-api-key";
+
+  const lookback = 3600 * 24 * 7;
```
mrice32, line 14:
```suggestion
  const lookback = 3600 * 24 * 7;
  const decimals = 6;
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,91 @@
...
+
+    defiPulseTotalPriceFeed = new DefiPulseTotalPriceFeed(
+      dummyLogger,
+      web3,
+      apiKey,
+      lookback,
+      networker,
+      getTime,
+      minTimeBetweenUpdates,
+      6 // Add arbitrary decimal conversion and prove this works.
```
mrice32, line 44:
```suggestion
      decimals // Add arbitrary decimal conversion and prove this works.
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,91 @@
...
+    networker.getJsonReturns = [...validResponses];
+
+    await defiPulseTotalPriceFeed.update();
+
+    // Before period 1 should return null.
+    assert.isTrue(await defiPulseTotalPriceFeed.getHistoricalPrice(1611572399).catch(() => true));
+    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611575900)).toString(), web3.utils.toWei("24.78"));
+    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), web3.utils.toWei("23.50"));
+    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), web3.utils.toWei("23.50"));
+    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579601)).toString(), web3.utils.toWei("22.25"));
```
mrice32, line 69: Since the decimals are specified as 6, shouldn't these be coming out as 24.78 * 10^6? Right now, it looks like they're coming out as 24.78 * 10^18. See my comment above for how to make the price feed to do what you expect here.
```suggestion
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611575900)).toString(), parseFixed("24.78", decimals).toString());
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), parseFixed("23.50", decimals).toString());
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579500)).toString(), parseFixed("23.50", decimals).toString());
    assert.equal((await defiPulseTotalPriceFeed.getHistoricalPrice(1611579601)).toString(), parseFixed("22.25", decimals).toString());
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,91 @@
...
+  });
+
+  it("Basic current price", async function() {
+    // Inject data.
+    networker.getJsonReturns = [...validResponses];
+
+    await defiPulseTotalPriceFeed.update();
+
+    // Should return the current price in the data.
+    assert.equal(defiPulseTotalPriceFeed.getCurrentPrice().toString(), web3.utils.toWei("25.1"));
```
mrice32, line 79:
```suggestion
    assert.equal(defiPulseTotalPriceFeed.getCurrentPrice().toString(), parseFixed("25.1", decimals).toString());
```

mrice32 (commented):
This is great! I left a few suggestions to make price feed decimals match how it works in the other price feeds. Please take a look. I haven't tested the suggestions, but I think they should work. You'll probably need to re-lint the repo, though, because they probably aren't formatted correctly. You can do this by committing the suggestions, running `git pull`, then running `yarn lint-fix` from the root of the repo, and then committing and pushing any changes.

bryanjcampbell1:
Accepted changes and ran through linter!

mrice32 (approved):
I made a few small commits because I didn't realize you had manually made some of the changes I suggested -- then I had to undo them haha. But I think it looks good now!

LGTM assuming @chrismaree approves as well!

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,160 @@
...
+  /**
+   * @notice Constructs the DefiPulseTVL_ALLPriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {String} apiKey DeFiPulse Data API key. Note: these API keys are rate-limited.
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the current time.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
```
chrismaree, line 16: nit: missing `priceFeedDecimals` in your comments.

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const billion = 1000000000;
+
+    const precision = 3;
+    assert(
+      precision <= this.priceFeedDecimals,
+      `Precision of ${precision} is > priceFeedDecimals of ${this.priceFeedDecimals}. Cannot have more precision than decimals`
+    );
+    const decimalValue = (_tvlUSD / billion).toFixed(precision);
+    const fixedPointValue = parseFixed(decimalValue.toString(), this.priceFeedDecimals);
```
chrismaree, line 149: +1. this correctly scales by priceFeedDecimals.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -90,6 +91,29 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+
+    return new DefiPulseTotalPriceFeed(
+      logger,
+      web3,
+      config.apiKey,
+      config.lookback,
+      networker,
+      getTime,
+      config.minTimeBetweenUpdates,
+      config.decimals
```
chrismaree, line 115: nit: for consistency please keep this as
```suggestion
      config.priceFeedDecimals
```

packages/financial-templates-lib/test/price-feed/DefiPulseTotalPriceFeed.js
chrismaree: ye, please add a test for this :)

chrismaree (commented):
This is so close to being ready. I have 2 nits and questions on some small spots.

Also, please add a test to validate non-18 decimals works as expected. Once this is done I think you good to go!

bryanjcampbell1:
@chrismaree  Matt helped me modify the test for "Basic historical price" so that it would work for arbitrary decimals.  Please let me know if you are looking for something else!

chrismaree (approved):
Thanks for replying to all the changes and updating your PR accordingly! this is pretty awesome, I'm looking forward to seeing this on main net 🚀

chrismaree:
I'm running your test in CI again.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -93,6 +94,29 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.poolDecimals,
       config.priceFeedDecimals // This defaults to 18 unless supplied by user
     );
+  } else if (config.type === "defipulsetvl") {
+    const requiredFields = ["lookback", "minTimeBetweenUpdates"];
+
+    if (isMissingField(config, config.apiKey, requiredFields, logger)) {
```
nicholaspai, line 100: Why don't you add the `apiKey` to `requiredFields`?

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,161 @@
...
+    this.historicalPrices = [];
+  }
+
+  getCurrentPrice() {
+    return this.currentPrice;
+  }
+
+  async getHistoricalPrice(time) {
+    if (this.lastUpdateTime === undefined) {
+      throw new Error(`${this.uuid}: undefined lastUpdateTime`);
```
nicholaspai, line 44: I don't see that you've defined `this.uuid` in the constructor

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,161 @@
...
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+   */
+  constructor(logger, web3, apiKey, lookback, networker, getTime, minTimeBetweenUpdates, priceFeedDecimals = 18) {
+    super();
+    this.logger = logger;
+    this.web3 = web3;
+
+    this.apiKey = apiKey;
+    assert(apiKey, "requires apiKey for Defi Pulse");
```
nicholaspai, line 25: Instead of `asserting` it here, I recommend adding `apiKey` to the `requiredFields` array in `CreatePriceFeed`

nicholaspai (approved):
I left some comments that once resolved are good to merge

packages/financial-templates-lib/src/price-feed/DefiPulseTotalPriceFeed.js
```diff
@@ -0,0 +1,158 @@
...
+    this.historicalPrices = [];
+  }
+
+  getCurrentPrice() {
+    return this.currentPrice;
+  }
+
+  async getHistoricalPrice(time) {
+    if (this.lastUpdateTime === undefined) {
+      throw new Error("undefined lastUpdateTime");
```
nicholaspai, line 41: Can you follow [this `this.uuid` pattern](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js#L70) and add a `uuid` class param that makes the Error slightly more descriptive please? Sorry and thanks!

daywiss:
@bryanjcampbell1 looks like there is just one outstanding issue and this can go in!

PR #2354, opened 2020-12-29
improve(price-feed): Add BasketSpreadPriceFeed which computes prices for STABLESPREAD-esque identifiers

### LINKED ISSUE #2352: Add STABLESPREAD price feed for liquidation/dispute bots to use
**Why**

EMP contracts are live that use these identifiers, we should be running bots against them.

We should create a one-off stablespread price feed that can be used by liquidators against this STABLESPREAD type identifiers. However, future work must be done to generalize this price feed since it will make some big assumptions.

- 3: ~1 hour

nicholaspai:
**Motivation**

#2352 

This does NOT work for `ELASTIC_STABLESPREAD/USDC` because one of the constituent coins, BAC, only has a Uniswap price source listed [in this UMIP](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-30.md) and the `BasketSpreadPriceFeed` does not currently work when one of the constituent price feeds is not a medianizer price feed. I'm tracking this issue in #2380 and will complete it after this PR.

**Issue(s)**

Fixes #2352


packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -36,36 +37,41 @@ exports.BlockHistory = (getBlock, blocks = []) => {
   function getClosestBefore(timestamp) {
     const index = lodash.sortedIndexBy(blocks, { timestamp }, "timestamp");
     // need to check for an exact match in this case, otherwise go to the older block
-    if (blocks[index].timestamp === timestamp) return blocks[index];
+    if (blocks[index] && blocks[index].timestamp === timestamp) return blocks[index];
```
nicholaspai, line 40: When running the SyntheticPegMonitor, I noticed that an error would be thrown here because `index === blocks.length` which makes `blocks[index] === undefined` and `undefined.timestamp` throws an error. Therefore, I added a check to this if statement to make sure that `blocks[index]` exists.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -276,8 +330,14 @@ async function createReferencePriceFeedForEmp(logger, web3, networker, getTime,
...
-  // Infer lookback from liquidation liveness.
-  if (emp && defaultConfig) {
+  // Infer lookback from liquidation liveness if default config does not specify a lookback.
+  // - Explanation: Always setting the `lookback = emp.liquidationLiveness` can potentially introduce problems
+  // if the liquidation liveness is particularly long and the pricefeed to be constructed is a DEX.
+  // - For example: the Zelda Cash contract has a 172800 liveness period (48 hours) and it uses the
+  // BasketSpreadPriceFeed which depends on the BalancerPriceFeed. Updating a DEX pricefeed requires
+  // a web3 call for every block included in the lookback window. Therefore, we override some of the default config
+  // lookback values so that they do not make an impractically large amount of web3 requests.
+  if (emp && defaultConfig && !defaultConfig.lookback) {
```
nicholaspai, line 340: I added this condition so that we can override the default setting of `lookback = liquidationLiveness` for certain EMP's like the Zelda cash ones which have very large liquidationLiveness periods of 172800. This large lookback becomes a problem since one of the pricefeeds to update is the `BalancerPriceFeed` which makes a `web3.eth.getBlock` call for every block within the lookback window. Empirically, this throws web3 "too many requests"-esque errors until I override the lookback to 7200.

I think giving the bots 2 hours of lookback is sufficient since I suspect it unlikely that a liquidation or dispute would be missed for > 2 hours, and we can always manually remove the override if the bots are offline for a large amount of time.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -57,38 +52,90 @@ const defaultConfigs = {
...
     medianizedFeeds: [
       { type: "cryptowatch", exchange: "coinbase-pro", pair: "BCHBTC" },
       { type: "cryptowatch", exchange: "binance", pair: "BCHBTC" },
       { type: "cryptowatch", exchange: "huobi", pair: "BCHBTC" }
     ]
   },
+  STABLESPREAD: {
+    // This is alternatively known as "STABLESPREAD/ETH"
+    type: "basketspread",
+    lookback: 7200, // We override the default value of 172800, equal to the liquidationLiveness
```
nicholaspai, line 71: See [this comment](https://github.com/UMAprotocol/protocol/pull/2354/files#r550805900) for reasoning why I removed all `lookback` values in this file except for this pricefeed, which depends on a `BalancerPriceFeed` and is used by EMP's with a large lookback window.

Note that the `BCHNBTC` EMP also has a large lookback window of 172800, but it depends only on Cryptowatch Pricefeeds which handle larger lookbacks much better than the DEX-based pricefeeds.

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -27,8 +30,28 @@ class BalancerPriceFeed extends PriceFeedInterface {
     // Add a callback to get price, error can be thrown from web3 disconection or maybe something else
     // which affects the update call.
     this.priceHistory = PriceHistory(async number => {
-      return this.toBN(await this.contract.methods.getSpotPriceSansFee(this.tokenIn, this.tokenOut).call(number));
+      try {
+        return this.toBN(await this.contract.methods.getSpotPriceSansFee(this.tokenIn, this.tokenOut).call(number));
```
nicholaspai, line 34: I need to update this method to compute the TWAP instead of spot price, like the Uniswap price feed.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -76,33 +77,14 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       return null;
     }
 
-    // Loop over all the price feeds to medianize.
-    const priceFeeds = [];
-    for (const medianizedFeedConfig of config.medianizedFeeds) {
```
nicholaspai, line 81: This was refactored since we construct a MedianizerPriceFeed both here and for the `BasketSpreadPriceFeed`

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+  /**
+   * @notice Constructs new BasketSpreadPriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {List} baselinePriceFeeds The baseline list of priceFeeds to compute the average of. All elements must be of type PriceFeedInterface.
+   *      Must be an array of at least one element.
+   * @param {List} experimentalPriceFeed The baseline list of priceFeeds to compute the average of. All elements must be of type PriceFeedInterface.
+   *      Must be an array of at least one element.
+   * @param {Object} denominatorPriceFeed We divide the price spread between the baseline and experimental baskets by this denominator price
+   *      in order to "denominate" the basket spread price in a specified unit. For example, we might want to express the basekt spread in terms
```
chrismaree, line 18:
```suggestion
   *      in order to "denominate" the basket spread price in a specified unit. For example, we might want to express the basket spread in terms
```

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+    this.denominatorPriceFeed = denominatorPriceFeed;
+
+    // Helper modules.
+    this.web3 = web3;
+    this.toBN = this.web3.utils.toBN;
+    this.toWei = this.web3.utils.toWei;
+    this.logger = logger;
+
+    // The precision that the user wants to return prices in is determined by the denominator price feed.
+    this.decimals = denominatorPriceFeed.getPriceFeedDecimals();
```
chrismaree, line 39: nit: you should check that the decimals are consistent between your two price feeds.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+
+    // Scale `number` by 10**decimals.
+    this.convertDecimals = number => {
```
chrismaree, line 42: small request: can you please name this `convertPriceFeedDecimals` so that it's _Really_ clear that this is scaling by the number of `decimals`?

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
chrismaree: It's actually probs fine as you do this effectively in your `getPriceFeedDecimals` method later on.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -55,8 +55,9 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     this.convertPriceFeedDecimals = number => {
       // Converts price result to wei
-      // returns price conversion to correct priceFeedDecimals as a big number
-      return this.toBN(parseFixed(number.toString(), priceFeedDecimals).toString());
+      // returns price conversion to correct decimals as a big number.
+      // Note: Must ensure that `number` has no more decimal places than `priceFeedDecimals`.
+      return this.toBN(parseFixed(number.toString().substring(0, priceFeedDecimals), priceFeedDecimals).toString());
```
chrismaree, line 60: why are you doing this substring?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -5,7 +5,6 @@ const defaultConfigs = {
   "ETH/BTC": {
     type: "medianizer",
     pair: "ethbtc",
-    lookback: 7200,
```
chrismaree, line 8: did you make this a default? why is this not here anymore?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -59,38 +54,217 @@ const defaultConfigs = {
...
+        ]
+      },
+      {
+        type: "medianizer",
+        computeMean: true,
+        medianizedFeeds: [
+          { type: "cryptowatch", exchange: "bittrex", pair: "cusdusdt" }
+          // NOTE: The OKCoin exchange is not available on Cryptowatch for this pair,
+          // presumably because it has such low volume.
+          // { type: "cryptowatch", exchange: "okcoin" }
```
chrismaree, line 98: what are the implications for this? does this mean we are not actually meeting the UMIP spec for the feed?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -59,38 +54,217 @@ const defaultConfigs = {
...
+      medianizedFeeds: [
+        { type: "cryptowatch", exchange: "kraken", pair: "btcusd" },
+        { type: "cryptowatch", exchange: "bitstamp", pair: "btcusd" }
+      ]
+    }
+  },
   "GASETH-TWAP-1Mx1M": {
     type: "uniswap",
     uniswapAddress: "0x25fb29D865C1356F9e95D621F21366d3a5DB6BB0",
-    twapLength: 7200,
```
chrismaree, line 80: same question as before on this.

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -6,15 +6,18 @@ class MedianizerPriceFeed extends PriceFeedInterface {
    * @notice Constructs new MedianizerPriceFeed.
    * @param {List} priceFeeds a list of priceFeeds to medianize. All elements must be of type PriceFeedInterface.
    *      Must be an array of at least one element.
+   * @param {Boolean=false} computeMean Set this to true to return the mean over price feeds instead of the median.
+   *      Default behavior is to return median.
    */
-  constructor(priceFeeds) {
+  constructor(priceFeeds, computeMean = false) {
```
chrismaree, line 12: I think this is fine for now as it limits code churn. As discussed IRL we should consider refactoring this file to `CombinerPriceFeed` or `PriceFeedAggregator` as this is no really a **Median**izer any more :)

packages/financial-templates-lib/src/price-feed/PriceFeedInterface.js
```diff
@@ -32,6 +32,12 @@ class PriceFeedInterface {
     this._abstractFunctionCalled();
   }
 
+  // This returns the precision that prices are returned in. It is called by the Medianizer price feed to enforce that all
```
chrismaree, line 35: thanks for adding this.

packages/financial-templates-lib/test/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,466 @@
...
+
+    await basketSpreadPriceFeed.update();
+
+    // On the basket spread's update call, `priceFeeds[0]` should have been updated once for each
+    // Medianizer price feed that it is incorporated in. This is because the basket spread price feed
+    // updates its imported medianizer price feeds.
+    // Check work: 2x for baseline update, 2x experimental udpate, 1x for denominator = 5 total.
+    assert.equal(priceFeeds[0].updateCalled, 5);
+  });
+  describe("Computing basket spreads when the spread is within the range [0,2]", function() {
```
chrismaree, line 44: nit: do you want to add tests for trying out different scaled price feeds? This is important seeing we've got some non-18 decimal feeds in some of the baskets (BTC denomination, for example).

I'd say you can follow the pattern done in the liquidator for looping the tests with multiple decimals. Can use a simple config like this: `const configs = [{ priceFeedDecimals: 18 }, { priceFeedDecimals: 8 }];` as you only care about the decimals in the feed.

chrismaree: The only thing that really will change however is the decimals of your denomination currency I guess. In either case I think it should still be included for different base currencies (BTC & ETH)

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -79,6 +80,88 @@ contract("CreatePriceFeed.js", function(accounts) {
...
+      balancerTokenIn: accounts[1],
+      balancerTokenOut: accounts[2]
+    };
+
+    const validBasketSpreadFeed = await createPriceFeed(logger, web3, networker, getTime, config);
+    assert.isTrue(validBasketSpreadFeed instanceof BasketSpreadPriceFeed);
+
+    // Check that baseline and experimental pricefeeds are lists of medianizer price feeds with the correct
+    // constituent pricefeeds.
+    assert.isTrue(validBasketSpreadFeed.baselinePriceFeeds[0] instanceof MedianizerPriceFeed);
```
chrismaree, line 153: +1

chrismaree (approved):
This looks awesome! Good work on everything. I've left some comments/nits and questions. Once you've taken a look I think this is pretty much good to go.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Yeah I was copying the `MedianizerPriceFeed` pattern where it checks in that `getPriceFeedDecimals` as well. Do you think its better to check in construction or as it does currently?

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Done

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
nicholaspai: Because sometimes the price returned by the CW API has a lot of decimals, and if the # of decimals > `priceFeedDecimasl` this throws an error. 

For example say I want 6 decimal precision, but the CW API returns a price with 7 decimals. This will be an error. This is really only an issue with REST API price feeds, not Balancer/Uniswap for example which return smart contract data

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: `CreatePriceFeed` defaults the lookback to an EMP's `liquidationLiveness` so this wasn't actually doing anything

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
nicholaspai: Totally agreed

packages/financial-templates-lib/test/price-feed/BasketSpreadPriceFeed.js
nicholaspai: I actually do a test for custom price feeds for every single section of the test [here](https://github.com/UMAprotocol/protocol/pull/2354/files#diff-bfff0b1bdb2e054a81074606bef333f1339dc44757d032b3c88370e40fa2f712R219)

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
chrismaree: This is probs fine but it won't hurt to have it in the constructor as well. You could also just call the `getPriceFeedDecimals` to double-check it does not through from here if you wanted?

chrismaree: With that said, it is called from the index and seeing you have the validation here you should be good to go as this will error out in the index file.

daywiss (commented):
started reviewing but didnt finish, will need to come back and spend more time on it

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+// !!Note: This PriceFeed assumes that the baselinePriceFeeds, experimentalPriceFeed, and denominatorPriceFeed
+// are all returning prices in the same precision as `decimals`.
+class BasketSpreadPriceFeed extends PriceFeedInterface {
+  /**
+   * @notice Constructs new BasketSpreadPriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {List} baselinePriceFeeds The baseline list of priceFeeds to compute the average of. All elements must be of type PriceFeedInterface.
+   *      Must be an array of at least one element.
+   * @param {List} experimentalPriceFeed The baseline list of priceFeeds to compute the average of. All elements must be of type PriceFeedInterface.
```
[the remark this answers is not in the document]
nicholaspai: "baseline" and "experimental" are just two baskets of medianizers. For example, an Ethereum stablecoin basket could be the baseline and the non-ethereum stablecoin basket could be the experimental. Is there a better term you can think of?

The basket spread computes (and normalizes) the spread: experimental basket avg price - baseline basket avg price

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+      if (spreadValue.lt(this.toBN("0"))) {
+        spreadValue = this.toBN("0");
+      }
+      // Ensure symmetry
+      else if (spreadValue.gt(this.convertPriceFeedDecimals("2"))) {
+        spreadValue = this.convertPriceFeedDecimals("2");
+      }
+
+      // Divide by denominator pricefeed.
+      if (denominatorPrice) {
```
mrice32, line 102: nit: same comment as above

maybe prefer:
```js
if (!denominatorPrice) {
  return null;
}
```

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+
+  // Inputs are expected to be BNs.
+  _computeMean(inputs) {
+    let sum = this.toBN("0");
+
+    for (let priceBN of inputs) {
+      sum = sum.add(priceBN);
+    }
+
+    return sum.div(this.toBN(inputs.length));
```
mrice32, line 183:
```suggestion
    return sum.divn(inputs.length);
```

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+  constructor(web3, logger, baselinePriceFeeds, experimentalPriceFeeds, denominatorPriceFeed) {
+    super();
+
+    if (baselinePriceFeeds.length === 0 || experimentalPriceFeeds.length === 0) {
+      throw new Error("BasketSpreadPriceFeed cannot be constructed with empty baseline or experimental baskets.");
+    }
+
+    this.baselinePriceFeeds = baselinePriceFeeds;
+    this.experimentalPriceFeeds = experimentalPriceFeeds;
+    this.denominatorPriceFeed = denominatorPriceFeed;
```
mrice32, line 30: How might one parameterize this denominator feed for something like USDC?

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+    // the one that this.convertPriceFeedDecimals() uses.
+    if (baselineMean && experimentalMean) {
+      let spreadValue = experimentalMean.sub(baselineMean).add(this.convertPriceFeedDecimals("1"));
+      this.logger.debug({
+        at: "BasketSpreadPriceFeed",
+        message: "Basket spread value",
+        spreadValue: spreadValue.toString()
+      });
+
+      // TODO: Parameterize these lower and upper bounds, as well as allow for custom "spreadValue" formulas,
```
mrice32, line 89: +1 on this TODO -- this would be nice to add in the future.

mrice32 (commented):
Looks great! A few minor suggestions/comments

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+      at: "BasketSpreadPriceFeed",
+      message: "Average of baseline prices",
+      mean: baselineMean.toString()
+    });
+
+    // All calculations within this if statement will produce unexpected results if any of the
+    // experimental mean, baseline mean, or denominator price are NOT in the same precision as
+    // the one that this.convertPriceFeedDecimals() uses.
+    if (baselineMean && experimentalMean) {
+      let spreadValue = experimentalMean.sub(baselineMean).add(this.convertPriceFeedDecimals("1"));
```
mrice32, line 82: @nicholaspai note for the future, we may also want to parameterize this offset.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,189 @@
...
+    this.logger.debug({
+      at: "BasketSpreadPriceFeed",
+      message: "Average of baseline prices",
+      mean: baselineMean.toString()
+    });
+
+    // All calculations within this if statement will produce unexpected results if any of the
+    // experimental mean, baseline mean, or denominator price are NOT in the same precision as
+    // the one that this.convertPriceFeedDecimals() uses.
+    if (baselineMean && experimentalMean) {
```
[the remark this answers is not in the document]
nicholaspai: Makes sense, done

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Done

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Yeah so this is an open question. We could allow denominatorPriceFeed to be `null` in which case we just ignore the last computation step.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -138,6 +142,40 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
+  }
+  // Returns an array or "basket" of MedianizerPriceFeeds
+  async function _createBasketOfMedianizerPriceFeeds(medianizerConfigs) {
+    const basketOfMedianizers = [];
+    for (const _medianizerConfig of medianizerConfigs) {
+      const medianizer = await _createMedianizerPriceFeed(_medianizerConfig);
+      basketOfMedianizers.push(medianizer);
+    }
+    return basketOfMedianizers;
```
mrice32, line 177:
```suggestion
    return await Promise.all(medianizerConfigs.map(config => _createMedianizerPriceFeed(config)));
```

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Done PTAL

mrice32: +1 -- that seems simpler

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,175 @@
...
+  }
+
+  // Given lists of experimental and baseline prices, and a denominator price,
+  // return the spread price, which is:
+  // (avg(experimental) - avg(baseline) + 1) / denominator
+  _getSpreadFromBasketPrices(experimentalPrices, baselinePrices, denominatorPrice) {
+    // Compute experimental basket mean.
+    if (
+      experimentalPrices.length === 0 ||
+      experimentalPrices.some(element => element === undefined || element === null)
```
daywiss, line 62: Not sure what would cause an undefined element here, but you could use a lodash compact or filter(x=>x) to get rid of nulls in array to avoid the check

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -0,0 +1,175 @@
...
+      spreadValue: spreadValue.toString()
+    });
+
+    // Ensure non-negativity
+    if (spreadValue.lt(this.toBN("0"))) {
+      spreadValue = this.toBN("0");
+    }
+    // Ensure symmetry
+    else if (spreadValue.gt(this.convertPriceFeedDecimals("2"))) {
+      spreadValue = this.convertPriceFeedDecimals("2");
```
daywiss, line 104: What is the rationale for these bounds?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
daywiss: Ive found using toFixed(this.decimals) usually does the trick

daywiss (approved):
Nice work, looks great

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: As defined in UMIP 28: https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-28.md#implementation

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Hm but you'd also have to do the `length === 0` check prior to the filter

PR #2355, opened 2020-12-29
feat(affiliates): add dapp mining script entrypoint and config example

### LINKED ISSUE #2289: affiliates: add tagged transactions for reward attribution
This feature has been started but still needs work. 

**Shared Attributions Model**
Creates a pool of affiliates per user. Most likely needs some more tests around this.

**Attribution History Processor**
Handles incoming transactions and attributes them to affiliates. Need to integrate with the new dataset system for testing. 

Most likely also need to handle more events to track both collateral and synthetics added for users.

**Testing**
Need create a dataset based on real transactions with tags injected in.

Probably also want some simple synthetic tests just to sanity check math.

**App**
Needs an application to run query for a specific time period and set of EMPs and produce an output with reward distribution for each tag. This will probably require total rewards input, a whitelist of addresses (tags) as well as a whitelist of emps to look at.  This should normalize values in some way, similar to deployer rewards and output a distribution of rewards to each tagged address.

**Other Tasks**
- Need to expose a library function which helps third parties easily tag relevant transactions to their emp. 
- Need documentation on how to use this

**Resolution**
- see that we can run this on a dataset and get reasonable results
- see that a third party could figure out how to integrate this into their app

daywiss:
**Motivation**

#2289

**Summary**

Mainly adds dapp mining script, but also theres a few additional helpers functions and slight refactors. Tests will need to be added, but script has been tested on 3 different contracts with different behaviors, and numbers look right so far. 

**Details**

- main logic is in libs/affiliates/dappmining.js
- app entrypoint in apps/DappMiningRewards.js
- new bigqueries have been added, though not all used
- new utility for determining input length of function call given the abi has been added in libs/contracts.js

**Example Output**
image

**Issue(s)**

Fixes #2289


packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+        blockTimestamp: moment(log.block_timestamp.value).valueOf(),
+        ...log
+      });
+    });
+  }
+
+  function getAttributionStream(empAddress, start, end) {
+    // We get all attributions from the trace stream. This is very similar to transactions query but
+    // transactions do not contain transactions called from other contracts! Trace does.
+    const stream = queries.streamTracesByContract(empAddress, start, end);
```
chrismaree, line 32: so this is to enable sponsors who used proxies for interacting with their positions to still receive rewards, right? For example, if my sponsor account was behind a DSProxy?

Is this query much larger than the transactions call?

packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+  // returns the proportion of rewards given to each developer for a specific point in time
+  function calculateBlockReward(
+    { attributions, balances, whitelist },
+    // return all tags with a percent value for rewards
+    rewards = {}
+  ) {
+    // for every user with a token balance
+    balances.keys().forEach(userid => {
+      // get their stake percent relative to all balances in emp
+      // const balancePercent = balances.getPercent(userid)
```
chrismaree, line 62: nit: should this still be here?

packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+      const balance = balances.get(userid);
+      // console.log({balancePercent,userid})
```
chrismaree, line 64: same nit.

packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+      // go through all whitelisted tags
+      whitelist.forEach(tag => {
+        if (rewards[tag] == null) rewards[tag] = "0";
+        // get that tags attribution percent for that user relative to all other affiliates
+        const attributionPercent = attributions.calculateShare(userid, tag);
+        // multiply the balance percent and the attribution percent to get final attribution weight
+        // and add it to the rest of the weights.
+        rewards[tag] = (BigInt(rewards[tag]) + BigInt(balance) * BigInt(attributionPercent)).toString();
```
chrismaree, line 72: are we sure we want to use BigInt? We explicitly moved away from this in other parts of the affiliates package to conform to the rest of the repo? IMHO we should stick to only using BN.

chrismaree: Same comment applies throughout this PR.

packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+  // Broken out logic for consuming and processing attribution stream. This can now be tested
+  // independently. It essentially takes a stream of attribution transactions, reduces it to a single
+  // attribution sum, which weights all attributions across the block range specified. The final output
+  // is used to weight the final reward calculation.
+  function ProcessAttributionStream({ attributions, attributionsSum = SharedAttributions(), startBlock, endBlock }) {
+    assert(attributionsSum, "requires attributions sum");
+    assert(attributions, "requires attributions");
+    assert(startBlock >= 0, "requires startBlock >= 0");
+    assert(endBlock >= 0 && endBlock > startBlock, "requires endBlock >=0 and > startBlock");
+    return stream => {
```
chrismaree, line 140: some complex logic in here! code looks good though. well commented and clean +1

packages/affiliates/libs/affiliates/dappmining.js
```diff
@@ -0,0 +1,254 @@
...
+  // Main entrypoint to kick off reward calculations. Returns an object with rewards keyed by address and value.
+  async function getRewards({ empAddress, defaultAddress, startTime, endTime, whitelist, totalRewards, firstEmpDate }) {
+    assert(empAddress, "requires emp address for reward calculation");
+    assert(defaultAddress, "requires defaultAddress for untagged rewards");
+    assert(startTime >= 0, "requires startTime >= 0");
+    assert(endTime > 0 && endTime > startTime, "requires endTime >= 0 && > startTime");
+    assert(whitelist && whitelist.length, "requires an array of whitelisted payout addresses");
+    assert(totalRewards > 0, "requires totalRewards above 0 to be shared across dapps");
+
+    // use firstEmpDate as a history cutoff when querying for events. We can safely say no emps were deployeed before Jan of 2020.
```
chrismaree, line 197:
```suggestion
    // use firstEmpDate as a history cutoff when querying for events. We can safely say no emps were deployed before Jan of 2020.
```

packages/affiliates/libs/models.js
```diff
@@ -40,21 +40,26 @@ function Prices(prices = []) {
 
 // users show current ratio of affiliates tagged on thier deposits
 // keyed by user address, and show total attributed for each affiliate per user
+// addresses are converted to lower case because theres inconsistent address reporting from various sources.
```
chrismaree, line 43: this inconsistency is so painful within the web3 stack.

packages/affiliates/test/mocha/attribution-rewards.js
```diff
@@ -0,0 +1,21 @@
...
+const empAbi = getAbi("ExpiringMultiParty");
+const datasetPath = Path.join(__dirname, "../datasets/set1");
+
+describe("DappMining", function() {
+  it("should instanciate", function() {
+    const queries = Queries(datasetPath);
+    const result = DappMining({ queries, empAbi, web3 });
+    assert(result);
+  });
+  // TODO: add more tests and appropriate dataset
```
chrismaree, line 20: this feels like it's going to be a pretty epic task to add tests...

packages/affiliates/libs/contracts.js
```diff
@@ -25,14 +31,51 @@ function DecodeTransaction(abi, meta = {}) {
   };
 }
 
-function decodeAttribution(
-  data,
-  delimiter = "ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff000000000000000000000000"
-) {
-  assert(data, "requires data to decode");
-  return data.split(delimiter)[1];
+function GetInputLength(abi) {
```
chrismaree, line 34: nit: should this be capital `G`?

packages/affiliates/libs/contracts.js
```diff
@@ -25,14 +31,51 @@ function DecodeTransaction(abi, meta = {}) {
...
+    assert(find, "unable to find name in abi: " + name);
+    if (find.inputs == null || find.inputs.length == 0) return 0;
+    return find.inputs.reduce((length, input) => {
+      return length + componentsLength(input.components);
+    }, 32); // 4 bytes always added as function name header hash
+  };
 }
 
+const DecodeAttribution = (abi, defaultAddress) => {
+  const inputLength = GetInputLength(abi)("create") / 4 + 2;
```
chrismaree, line 71: can you add a comment on why you divide by 4 and add 2 to compute the `inputLength`? 
```suggestion
  const inputLength = GetInputLength(abi)("create") / 4 + 2; // convert bits to hex and + 2 for 0x
```

packages/affiliates/scripts/example-transactions.js
```diff
@@ -2,28 +2,34 @@ const { BigQuery } = require("@google-cloud/bigquery");
...
       // this filters based on matts scheme for matching attribution tag
       // but we could easily do any type of scheme with JS filtering
-      .filter(({ input }) => {
-        return input.match(/f{64}0{24}.{40}$/);
-      })
-      // from here you can map or reduce or whatever you need for down stream processing
-      // we are just going to "collect" stream into an array for display
+      .map(x => decode(x, x))
+      .filter(x => x.name == "create")
+      // .doto(x=>console.log(JSON.stringify(x,undefined,2)))
```
chrismaree, line 31: nit: should this still be here?

packages/affiliates/libs/contracts.js
```diff
@@ -25,14 +31,51 @@ function DecodeTransaction(abi, meta = {}) {
...
+  return createTransaction => {
+    assert(createTransaction.name == "create", "Only decodes create transactions");
+    if (createTransaction.input.length == inputLength) return defaultAddress;
```
chrismaree, line 74: This here is really the core logic of identifying a tagged tx. If the input length is not equal to the default length then the tx is tagged. I think you should add an extra comment or two here just to point out exactly what's happening here.

packages/affiliates/libs/contracts.js
```diff
@@ -25,14 +31,51 @@ function DecodeTransaction(abi, meta = {}) {
...
+    return "0x" + createTransaction.input.slice(inputLength);
```
chrismaree, line 75: What happens if the tagged data is not an address? say I appended some 100 bytes of rubbish to the end of my tx. it will return that as a `0x` appended string but will be invalid. Should we perhaps validate that this is indeed an address? Something like:
```
web3.utils.isAddress(address)
```
could be useful.

chrismaree (commented):
This is pretty awesome, great work @daywiss! 

Do you have a test data set that you produced from big query to test against, as I see you are no longer using the ones in `tagged-transactions`? Will this come in as a separate PR? Adding tests for this feels like a pretty big task!

I had some comments and questions through the PR but overall I think this is great and close to being ready to merge :)

packages/affiliates/libs/affiliates/dappmining.js
daywiss: I believe this will allow users to use proxies. I ran into an issue where some transactions were not being included when querying transactions. I didn't actually verify though they were being sent through a proxy. When I switched to trace, the missing transactions showed up.  I dont know how this affects tagging though, it may render it unusable anyway, might require a deeper look.

This is a much more expensive call, for instance Jan 1 2020 to Dec 2020: 
- traces: 693.5 GB vs transactions: 167.3 GB

packages/affiliates/libs/affiliates/dappmining.js
daywiss: I will update it to bignum

packages/affiliates/libs/contracts.js
daywiss: Id like to keep it capital since its more like a class which needs to be instanced with `abi` before it can be used.

packages/affiliates/libs/contracts.js
daywiss: This is a really good question. I think i might change this to not prepend the 0x, and let the caller be opinionated about the structure of the "tag", rather than put the opinion within this call. I want to leave open the possibility of having arbitrary tags, so i think better to leave this call as unopinionated as we can.  

That said though i will update the EmpAttributions class in processors to prepend 0x and check for an address.  For now if its an invalid address, I think we should tag it with the default.

daywiss:
> Do you have a test data set that you produced from big query to test against, as I see you are no longer using the ones in tagged-transactions? Will this come in as a separate PR? Adding tests for this feels like a pretty big task!

Thanks for review, good points. Yes testing will be a pain, ive tried to break out major logic to be more easily unit tested but it will still be a decent amount of work. Will probably want to generate new dataset as well for more e2e style test. I want to do this as a seperate PR since I imagine it will be a lot to review.

packages/affiliates/libs/affiliates/dappmining.js
mrice32: > I dont know how this affects tagging though, it may render it unusable anyway

What do you mean?

daywiss: Well I guess I'm not sure of how our tagging mechanism wo￼uld interact ￼with a proxy contract.

It's relying on knowing the Abi of the contract in order to decode the tag. Not sure that assumption holds when going through the proxy and following a "trace".

packages/affiliates/libs/contracts.js
chrismaree: sgtm!

chrismaree (approved):
This is pretty awesome! great work on this logic @daywiss !

packages/affiliates/dappmining-config.example.js
```diff
@@ -0,0 +1,26 @@
+const moment = require("moment");
+
+// Emp to process rewards for
+const empAddress = "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5";
```
nicholaspai, line 4: Can you leave a comment about which EMP this is?

packages/affiliates/dappmining-config.example.js
```diff
@@ -0,0 +1,26 @@
...
+
+// Start of payout period
+const startTime = moment("2020-12-10", "YYYY-MM-DD").valueOf();
+// End of payout period
+const endTime = moment("2020-12-28", "YYYY-MM-DD").valueOf();
+// Total amount of rewards to payout for period
+const totalRewards = 25000;
+
+// Default reward address for emp create events which are untagged
+const defaultAddress = "0xa0dfF1F09a0DCEda7641b7eB07c3BDbe66a06C35";
```
nicholaspai, line 14: Are these reward addresses sponsors or deployers? Might be worth clarifying that

packages/affiliates/scripts/example-dapp-mining-calculator.js
```diff
@@ -0,0 +1,31 @@
+const { BigQuery } = require("@google-cloud/bigquery");
+const { getWeb3 } = require("@uma/common");
+const moment = require("moment");
+const { getAbi } = require("@uma/core");
+const { DappMining } = require("../libs/affiliates");
+const Queries = require("../libs/bigquery");
+
+const web3 = getWeb3();
+
+const empAddress = "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5";
```
nicholaspai, line 10: Are these the same addresses from `affiliates/dappmining-config.example.js` and is there a way to link them programmatically so that you're not repeating hard coded vars?

packages/affiliates/scripts/example-transactions.js
```diff
@@ -2,28 +2,33 @@ const { BigQuery } = require("@google-cloud/bigquery");
 const highland = require("highland");
 const Queries = require("../libs/bigquery");
 const moment = require("moment");
+const { getAbi } = require("@uma/core");
+const { DecodeTransaction, GetInputLength } = require("../libs/contracts");
 
-const contract = "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56";
-const start = moment("2020-10-06", "YYYY-MM-DD").valueOf();
-const end = moment("2020-10-10", "YYYY-MM-DD").valueOf();
+// const contract = "0x3605Ec11BA7bD208501cbb24cd890bC58D2dbA56";
```
nicholaspai, line 8: Did you mean to leave these comments

packages/affiliates/test/mocha/attribution-rewards.js
nicholaspai: yea..... question is should we add tests before we merge? I think we should have tests before we actually execute this script in production

nicholaspai (commented):
Overall looks good, but there's a lot of complex logic so its a bit difficult to review thoroughly. A couple suggestions from me:
- Can you follow #2356 and eventually add some README's or a README to the `affiliates` package explaining highlevel what the different folders are used for
- I think its OK not to add tests for now

packages/affiliates/dappmining-config.example.js
daywiss: Yes will update. This is the dapp developer address to receive rewards by default when no tag is specified on transaction

packages/affiliates/scripts/example-dapp-mining-calculator.js
daywiss: yea this is a great idea. I am going to import the example config file for the parameters.

daywiss:
> Can you follow #2356 and eventually add some README's or a README to the affiliates package explaining highlevel what the different folders are used for

Cool, yes actually already added a bunch of readmes in a previous PR #2321. Will want to revisit and update these though eventually.

packages/affiliates/libs/affiliates/dappmining.js
daywiss: @mrice32 I did some digging on this, and yes the missing transactions go through a proxy. Heres an example: https://etherscan.io/tx/0x2f23b55607432961f4659065969fbb7094ab75b086acda6b6a1b73756ea93a11

It seems like as long as you correctly tag the data param for the proxy call, I think this will be seen in the trace transaction as a normal tag to the contract. One side effect of this is that the from_address originates at the proxy address and not the EOA, but that doesnt matter as far as any calculations go.  This is what the decoded transaction looks like: 
image

packages/affiliates/test/mocha/attribution-rewards.js
daywiss: will be working on this #2365

Issue #2351, opened 2020-12-28
Ability for price identifiers to have the same number of decimals as the collateral currency


chrismaree:
**Why**

The current UMA ecosystem of contracts(EMP & perp) contains a range of different decimal configurations between a) synthetic, b) price feed and c) collateral type. The bots should be updated to accommodate any combination of the decimal convention.

Below the different possible configurations are listed:
1) The synthetic, collateral, and price feed decimals all match up. This is the simplest and most common configuration and occurs when the collateral type has 18 decimals (such as WETH or DAI).
2) The synthetic is 18 decimals, collateral is 8 decimals and the price feed is 8 decimals. This is the case with old uUSDrBTC contracts _and_ BTC denominated contracts expiring as late as March.
3) The synthetic is 8 decimals, collateral is 8 and the price feed is 18 decimals. This is the latest form of decimal delimitation, with the latest EMP & Perp conforming to this standard. This format was imposed to make voting easier as price feeds always have the same number of decimals (18) and the collateral matches synthetic.

The complexity of this issue is to be able to accommodate all 3 types of delimitation at the same time.

A bot that can accommodate this should be able to deal with the following combinations:
```
const configs = [
  {
    tokenName: "Wrapped Ether",
    tokenSymbol: "WETH",
    collateralDecimals: 18,
    syntheticDecimals: 18,
    priceFeedDecimals: 18
  },
  {
    tokenName: "Legacy Wrapped Bitcoin Contract",
    tokenSymbol: "BTC",
    collateralDecimals: 8,
    syntheticDecimals: 18,
    priceFeedDecimals: 8
  },
  {
    tokenName: "Wrapped Bitcoin",
    tokenSymbol: "BTC",
    collateralDecimals: 8,
    syntheticDecimals: 8,
    priceFeedDecimals: 18
  }
];
```

**How**

Update the bot implementation to accommodate dynamic scalling.

For example, in the case of detecting an undercolltarealized position the formulation is: 
```
(numTokens * trv) * collateralRequirement > amountCollateral
```
 
This equation assumes the decimal points across the inputs are normalized to the same basis. However, this wont always be the case and so we need to consider arbitrary decimals coming into the equation. 

When considering decimals of each variable within the as collateral (`cD`), synthetic (`sD`), CR (`1e18`), trv (`trvD`) this equation becomes:
``` 
[...]

PR #2357, opened 2020-12-29
improve(disputer): enable arbitrary decimal combinations between collateral, synthetic and price feed in disputer bot

chrismaree:
**Motivation**

issue #2351 has a fair amount of detail on the problem space.

Follows very closely on from https://github.com/UMAprotocol/protocol/pull/2276 

**Summary**

The disputer re-uses the emp client in a very similar way to how the liquidator uses it. As a result, this PR simply updates the disputer to correctly implement the new client interfaces and updates the tests accordingly. Please see #2351 and https://github.com/UMAprotocol/protocol/pull/2276  for details on the original change made to the client.

**Issue(s)**

issue #2351


packages/common/src/FormattingUtils.js
```diff
@@ -124,9 +124,9 @@ function addSign(number) {
   }
 }
 
-// Take an amount based on fromDecimals and convert it to an amount based on toDecimals
-// For example amount = 100 usdt, its decimals are 6. You want to convert it to 18.
-// convertDecimals(6,18)(100000000)  => 1000000000000000000000.
+// Take an amount based on fromDecimals and convert it to an amount based on toDecimals. For example 100 usdt = 100e6,
+// with 6 decimals. If you wanted to convert this to a base 18 decimals you would get:
```
nicholaspai, line 128: It looks like this function catches both cases, where `fromDecimals > toDecimals` and `fromDecimals < toDecimals`, right?

packages/disputer/index.js
```diff
@@ -99,13 +98,22 @@ async function run({
...
     });
 
     // Client and dispute bot.
-    const empClient = new ExpiringMultiPartyClient(logger, getAbi("ExpiringMultiParty"), web3, empAddress);
+    const empClient = new ExpiringMultiPartyClient(
+      logger,
+      getAbi("ExpiringMultiParty"),
+      web3,
+      empAddress,
+      collateralTokenDecimals,
```
nicholaspai, line 112: Shouldn't the `collateralTokenDecimals == syntheticTokenDecimals` for all future  EMP's, if so then why pass both of them in here

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -49,6 +49,7 @@ const defaultConfigs = {
     lookback: 7200,
     invertPrice: true,
     minTimeBetweenUpdates: 60,
+    priceFeedDecimals: 8,
```
nicholaspai, line 52: Should we also add this exact same override to the BCHNBTC pricefeed?

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -72,6 +72,17 @@ class MedianizerPriceFeed extends PriceFeedInterface {
     return Math.max(...lastUpdateTimes);
   }
 
+  // Gets the decimals of the medianized price feeds. Errors out if any price feed had a different number of decimals.
```
nicholaspai, line 75: Good check

packages/liquidator/index.js
```diff
@@ -174,6 +184,15 @@ async function run({
       liquidatorConfig
     });
 
+    logger.debug({
+      at: "Liquidator#index",
+      message: "Liquidator initialized",
+      collateralDecimals: Number(collateralTokenDecimals),
+      syntheticDecimals: Number(syntheticTokenDecimals),
+      priceFeedDecimals: Number(priceFeed.getPriceFeedDecimals()),
```
nicholaspai, line 192: Is this the only case where a pricefeed's `getPriceFeedDecimals()` is used? If so, how do you anticipate this method can be used.

nicholaspai (commented):
Great start overall, have a question about the new `getPriceFeedDecimals()` interface method.

packages/disputer/index.js
chrismaree: The whole point of these changes is to make it both forwards and backward compatible. to accommodate this the CLient needs to know: a) collateralTokenDecimals. b) syntheticTokenDecimals and c) priceFeedDecimals.

packages/disputer/test/index.js
```diff
@@ -115,34 +115,45 @@ contract("index.js", function(accounts) {
...
-    collateralToken = await Token.new("DAI8", "DAI8", 8, { from: contractCreator });
-    constructorParams = {
-      ...constructorParams,
-      collateralAddress: collateralToken.address
-    };
-    emp = await ExpiringMultiParty.new(constructorParams);
+    collateralToken = await Token.new("BTC", "BTC", 8, { from: contractCreator });
+    syntheticToken = await SyntheticToken.new("Test Synthetic Token", "SYNTH", 18, { from: contractCreator });
+    // For this test we are using a lower decimal identifier, USDBTC. First we need to add it to the whitelist.
+    await identifierWhitelist.addSupportedIdentifier(padRight(utf8ToHex("USDBTC"), 64));
```
nicholaspai, line 128: Should we refactor `padRight(someHex, 64)` into a common util file?

chrismaree: ye, can do. should be in a separate PR though.

daywiss (approved):
looks good. It looks like this is mainly just updating the tests and there was very little that needed to be done in the actual disputer logic.

chrismaree:
> looks good. It looks like this is mainly just updating the tests and there was very little that needed to be done in the actual disputer logic.

yup. The disputer does not compute any logic itself so min changes needed there.

nicholaspai (approved):
LGTM good tests

PR #2360, opened 2020-12-29
feat(2291): set higher fromBlock for EMP client, and getFromBlock function for override

### LINKED ISSUE #2291: Performance Improvement: EMP Client should explicitly set a "fromBlock" when querying for past events
**Why**

The [`ExpiringMultiPartyClient`](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js#L120) queries for past EMP contract events on every `update()` call. It sets the default [`fromBlock` value when calling `web3.contract.getPastEvents`](https://web3js.readthedocs.io/en/v1.2.7/web3-eth-contract.html#getpastevents) to 0, which means that the entire Ethereum block history from 0 to `latest` is queried. This query is extremely slow and unnecessary since blocks prior to the EMP contract's deployment are irrelevant.

This query is moreover dangerous because the EMPClient's `update()` method is called by the liquidator and disputer's `update()` method and therefore it can cause timeout errors for the liquidator, disputer AND monitor bots. Additionally,  some web3 providers might block such lengthy lookbacks.

**How**

A logical `fromBlock` would be the EMP's deployment date. The complexity here is that the `deployment` date is not easily available on-chain. There are a few viable solutions with tradeoffs that I can think of currently:
- Add a `earliestQueryBlock` to the client constructor so that the ultimate user can prevent any event querying earlier than a certain block. This solution would require the most churn to the unit tests and production deployments because it would change the constructor's interface. Additionally, it is sub-optimal to add configuration settings, ideally we have the user have to worry about as few configuration options as possible. The upside is that this is easy to implement within the EMP client source code, and it would be [consistent with how the `ExpiringMultiPartyEvent` client limits query length](ExpiringMultiPartyClient).
- Outside of the core loop, fetch and store the relevant EMP's deployment block via a one-time `getPastEvents()` call using a `fromBlock = 0`. This call could take place in the [constructor](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js#L16) the contract to query events on would be the [`Registry` via the `NewContractRegistered` event](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/implementation/Registry.sol#L56). All officially supported EMP's are registered with the `Registry` so that they are allowed to make DVM requests, and [this event is emitted on the EMP's deployment](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.sol#L89). The block number associated with the `NewContractRegistered` event can be saved in the EMPClient's state and used as the `fromBlock` for all event queries in the core `update()` loop. The upside of this implementation is that the class's constructor interface can remain the same. The downside is that one event query starting from block 0 to `latest` must still be made, but at least this length of time is used only once instead of on every `update()` call.
- Hardcode a `fromBlock` that was prior to any EMP that was every deployed or will be deployed. For example, a safe starting block would be prior to Jan 1 2019 (i.e. before the v0 DVM was deployed). This is inspired by how we set the earliest event block in the `affiliates` package [here](https://github.com/UMAprotocol/protocol/blob/2bd1fc8e55db510c5143adbb019b8d15ea95c2a3/packages/affiliates/libs/affiliates.js#L21

**Resolution**

Implementation should be well tested by unit tests and the EMPClient should handle the case where no `NewContractRegistered` event is found. In this case, the class should use a fallback `fromBlock` value.

**Difficulty Score [1-10]**

- 1: <=5 minutes
- 2: ~30 minutes
- 3: ~1 hour
- 4: ~2 hours
- 5: ~4 hours
**- **6: ~1 day**
- 7: ~3 days
- 8: ~1 week
- 9: ~2 weeks
- 10: ~4 weeks

pemulis:
**Motivation**

Addresses timeout issues in the EMP client due to inefficient querying for the full Ethereum block history for EMP events. This PR changes the `fromBlock` to the first block where an EMP contract was deployed, since that is the first time in the block history where we could possibly find events. It also adds a `getFromBlock` function that allows for overriding the `fromBlock` with an environment variable. EMP client tests will work if the environment variable is set to `0`. This function could also be useful in scripts and bots, not just for testing.

**Summary**

I guess this was all outlined in the Motivation section above!

**Details**

**Issue(s)**

Fixes #2291

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,12 @@ one package, you can run:
 yarn workspace <package_name> test
 ```
 
+If you are testing the ExpiringMultiPartyClient, you will need to set an environment variable so that the test will work with Ganache.
+
+```
+export FROM_BLOCK=0
+```
+
 ### Running the linter 🧽
 
 To run the linter in autofix mode (it will attempt to fix any errors it finds), run:
--- a/README.md
+++ b/README.md
@@ -100,12 +100,6 @@ one package, you can run:
 yarn workspace <package_name> test
 ```
 
-If you are testing the ExpiringMultiPartyClient, you will need to set an environment variable so that the test will work with Ganache.
-
-```
-export FROM_BLOCK=0
-```
-
 ### Running the linter 🧽
 
 To run the linter in autofix mode (it will attempt to fix any errors it finds), run:
--- a/packages/common/src/Constants.js
+++ b/packages/common/src/Constants.js
@@ -31,5 +31,9 @@ module.exports = {
   ZERO_ADDRESS: "0x0000000000000000000000000000000000000000",
 
   // Names of different interfaces in the Finder.
-  interfaceName
+  interfaceName,
+
+  // Block number of first EMP created.
+  // https://etherscan.io/tx/0x741ccbf0f9655b0b71e3842d788d58770bd3eb80c8f5bdf4fdec7cd74a776ea3
+  UMA_FIRST_EMP_BLOCK: 10103723
 };
--- a/packages/common/src/TimeUtils.js
+++ b/packages/common/src/TimeUtils.js
@@ -1,4 +1,5 @@
 const { UMA_FIRST_EMP_BLOCK } = require("./Constants.js");
+require("dotenv").config();
 
 /**
  * @notice Return average block-time for a period.
@@ -17,7 +18,7 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
 };
 
 // Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
-async function getFromBlock() {
+async function getFromBlock(web3) {
   const networkType = await web3.eth.net.getNetworkType();
   if (process.env.FROM_BLOCK) {
     return Number(process.env.FROM_BLOCK);
--- a/packages/common/src/TimeUtils.js
+++ b/packages/common/src/TimeUtils.js
@@ -17,7 +17,7 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
 };
 
 // Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
-async function getFromBlock() {
+function getFromBlock() {
   if (process.env.FROM_BLOCK) {
     return process.env.FROM_BLOCK;
   } else {
--- a/packages/common/src/TimeUtils.js
+++ b/packages/common/src/TimeUtils.js
@@ -20,7 +20,7 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
 async function getFromBlock() {
   const networkType = await web3.eth.net.getNetworkType();
   if (process.env.FROM_BLOCK) {
-    return process.env.FROM_BLOCK;
+    return Number(process.env.FROM_BLOCK);
   } else if (networkType === "main") {
     return UMA_FIRST_EMP_BLOCK;
   } else {
--- a/packages/common/src/TimeUtils.js
+++ b/packages/common/src/TimeUtils.js
@@ -1,3 +1,5 @@
+const { UMA_FIRST_EMP_BLOCK } = require("./Constants.js");
+
 /**
  * @notice Return average block-time for a period.
  */
@@ -14,6 +16,16 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
   }
 };
 
+// Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
+async function getFromBlock() {
+  if (process.env.FROM_BLOCK) {
+    return process.env.FROM_BLOCK;
+  } else {
+    return UMA_FIRST_EMP_BLOCK;
+  }
+}
+
 module.exports = {
-  averageBlockTimeSeconds
+  averageBlockTimeSeconds,
+  getFromBlock
 };
--- a/packages/common/src/TimeUtils.js
+++ b/packages/common/src/TimeUtils.js
@@ -17,11 +17,14 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
 };
 
 // Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
-function getFromBlock() {
+async function getFromBlock() {
+  const networkType = await web3.eth.net.getNetworkType();
   if (process.env.FROM_BLOCK) {
     return process.env.FROM_BLOCK;
-  } else {
+  } else if (networkType === "main") {
     return UMA_FIRST_EMP_BLOCK;
+  } else {
+    return 0;
   }
 }
 
--- a/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
+++ b/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
@@ -118,9 +118,9 @@ class ExpiringMultiPartyClient {
     }
     // Fetch contract state variables in parallel.
     const [newSponsorEvents, endedSponsorEvents, liquidationCreatedEvents, currentTime] = await Promise.all([
-      this.emp.getPastEvents("NewSponsor", { fromBlock: await getFromBlock() }),
-      this.emp.getPastEvents("EndedSponsorPosition", { fromBlock: await getFromBlock() }),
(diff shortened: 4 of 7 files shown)


packages/common/src/TimeUtils.js
```diff
@@ -14,6 +16,16 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
   }
 };
 
+// Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
+async function getFromBlock() {
```
mrice32, line 20: nit: does this function need to be async?

packages/common/src/TimeUtils.js
```diff
@@ -14,6 +16,16 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
...
+    return process.env.FROM_BLOCK;
+  } else {
+    return UMA_FIRST_EMP_BLOCK;
+  }
+}
+
 module.exports = {
-  averageBlockTimeSeconds
+  averageBlockTimeSeconds,
+  getFromBlock
```
mrice32, line 30: Not strongly opinionated either way, but would this be better suited to sit inside financial-templates-lib under helpers or even the ExpiringMultiPartyClient, itself? I only ask because I think we should generally try to remove things from common unless they are common functionality used by multiple packages.

mrice32 (commented):
Looks great! One minor nit and a question.

packages/common/src/TimeUtils.js
pemulis: Nope! It was async from earlier when I was looking at getting the block number from the node, before changing tracks.

packages/common/src/TimeUtils.js
pemulis: I think `getFromBlock` would be useful in other packages when you want to set a custom `fromBlock` but don't want to hardcode it. Affiliate scripts and bots are possible examples.

packages/common/src/Constants.js
```diff
@@ -31,5 +31,9 @@ module.exports = {
   ZERO_ADDRESS: "0x0000000000000000000000000000000000000000",
 
   // Names of different interfaces in the Finder.
-  interfaceName
+  interfaceName,
+
+  // Block number of first EMP created.
+  // https://etherscan.io/tx/0x741ccbf0f9655b0b71e3842d788d58770bd3eb80c8f5bdf4fdec7cd74a776ea3
+  UMA_FIRST_EMP_BLOCK: 10103723
```
daywiss, line 38: Nice use of this constants file, I was not aware of it, could be useful but it might require being more generalized to work for different networks, not sure, and I don't have any suggestions so I think this is fine for now.

packages/common/src/TimeUtils.js
daywiss: I could see a use for this in the affiliates package, the one gripe I have though is having to set a new env var to the run tests outside of main net.

packages/common/src/TimeUtils.js
```diff
@@ -14,6 +16,16 @@ const averageBlockTimeSeconds = async (/* lookbackSeconds */) => {
   }
 };
 
+// Sets fromBlock to the value of an environment variable if one is set. This can be set to 0 to make tests work with Ganache, or any other value needed for a production script or bot.
+function getFromBlock() {
+  if (process.env.FROM_BLOCK) {
+    return process.env.FROM_BLOCK;
```
daywiss, line 22: This could be a problem since env vars are parsed as strings and someone is expecting a number. Also if this is set to "0" is it considered truthy?

packages/common/src/TimeUtils.js
daywiss: +1 for not making this async

daywiss (commented):
Looks good but i can't help but hope for a better way to override the mainnet number than using an env. It's like I'd rather pass through a network identifier to the function explicitly. Something like that would work well with affiliates, but I cant say that about the bots.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -118,9 +118,9 @@ class ExpiringMultiPartyClient {
     }
     // Fetch contract state variables in parallel.
     const [newSponsorEvents, endedSponsorEvents, liquidationCreatedEvents, currentTime] = await Promise.all([
-      this.emp.getPastEvents("NewSponsor", { fromBlock: 0 }),
-      this.emp.getPastEvents("EndedSponsorPosition", { fromBlock: 0 }),
-      this.emp.getPastEvents("LiquidationCreated", { fromBlock: 0 }),
+      this.emp.getPastEvents("NewSponsor", { fromBlock: await getFromBlock() }),
```
chrismaree, line 121: why is this `await getFromBlock()` when `getFromBlock` is not async and simply returns an environment var?

packages/common/src/TimeUtils.js
nicholaspai: yeah and we originally used this function in the affiliates package for calculating Gas fee rebates

README.md
```diff
@@ -100,6 +100,12 @@ one package, you can run:
 yarn workspace <package_name> test
 ```
 
+If you are testing the ExpiringMultiPartyClient, you will need to set an environment variable so that the test will work with Ganache.
+
+```
+export FROM_BLOCK=0
```
nicholaspai, line 106: This is a bit annoying to ask of devs but alas, I wish there was a way to start ganache at a certain block

packages/common/src/Constants.js
nicholaspai: Yeah actually @daywiss that's a great point. What if instead of making `UMA_FIRST_EMP_BLOCK` an integer, you make it a mapping between `networkID` and integer, that way you can auto detect the first block based on the dev's specified network

nicholaspai (commented):
Looks great, left some suggestions! Can't wait to propagate this performance savings across the monorepo :D

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
pemulis: Good catch! It shouldn't be anymore. This is an artifact from when I was looking at querying the current block of whatever chain (Ganache, mainnet, etc.) but is no longer necessary. Will fix that.

README.md
pemulis: same :\

packages/common/src/Constants.js
pemulis: Annoyingly, it looks like Ganache uses the current time as the default `networkId` instead of some consistent number.

```
-i or --networkId: Specify the network id ganache-cli will use to identify itself (defaults to the current time or the network id of the forked blockchain if configured)
```
[https://github.com/trufflesuite/ganache-cli](https://github.com/trufflesuite/ganache-cli)

This should still work to differentiate between mainnet and everything else, though. Could have it use the mainnet block number if the networkId is 1 and use block number 0 otherwise. Will take a look at this.

nicholaspai: > Could have it use the mainnet block number if the networkId is 1 and use block number 0 otherwise. Will take a look at this.

I think this is very reasonable. Other public networks like Kovan are OK to start at 0

nicholaspai: I think changing the integer to a mapping could also remove the slightly annoying ask to set the FROM_BLOCK env var (which should still be an optional setting but not required anymore)

pemulis: Ooh, even better: https://web3js.readthedocs.io/en/v1.2.0/web3-eth-net.html#getnetworktype

This method knows how to recognize `ropsten` and `morden` and could probably be taught to recognize `kovan` through an upstream PR.

packages/common/src/TimeUtils.js
pemulis: In earlier testing, when setting the `FROM_BLOCK` to zero in an environment variable this worked out okay. Think I should add an explicit conversion to a number, though?

daywiss: yea i think probably cast to number on return

daywiss (approved):
going to approve this assuming any outstanding comments get resolved!

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -117,10 +117,11 @@ class ExpiringMultiPartyClient {
       await this.initialSetup();
     }
     // Fetch contract state variables in parallel.
+    const fromBlock = await getFromBlock(this.web3);
```
nicholaspai, line 120: @UMAprotocol/eng Where else in the repo can we take advantage of this function and reduce web3 calls?

PR #2364, opened 2020-12-31
improve(voter-dapp): add hardcoded precision for new identifiers

nicholaspai:

**Motivation**

New identifiers must resolve to same precision as collateral:
- ELASTIC_STABLESPREAD/USDC, STABLESPREAD/USDC ==> 6 decimals, same as USDC
- STABLESPREAD/BTC, BCHNBTC ==> 8 decimals, same as renBTC

Once the latest EMP factory is deployed, we will be able to assume that all identifiers can resolve to 18 decimals, but currently the identifier precision must match the EMP's collateral precision.

### DIFF
--- a/packages/common/src/PriceIdentifierUtils.js
+++ b/packages/common/src/PriceIdentifierUtils.js
@@ -9,7 +9,8 @@ const IDENTIFIER_NON_18_PRECISION = {
   USDBTC: 8,
   "STABLESPREAD/USDC": 6,
   "STABLESPREAD/BTC": 8,
-  "ELASTIC_STABLESPREAD/USDC": 6
+  "ELASTIC_STABLESPREAD/USDC": 6,
+  BCHNBTC: 8
 };
 
 const getPrecisionForIdentifier = identifier => {
--- a/packages/common/src/PriceIdentifierUtils.js
+++ b/packages/common/src/PriceIdentifierUtils.js
@@ -6,7 +6,10 @@ const IDENTIFIER_BLACKLIST = {
 // Price identifiers that should resolve prices to non 18 decimal precision. Any identifiers
 // not on this list are assumed to resolve to 18 decimals.
 const IDENTIFIER_NON_18_PRECISION = {
-  USDBTC: 8
+  USDBTC: 8,
+  "STABLESPREAD/USDC": 6,
+  "STABLESPREAD/BTC": 8,
+  "ELASTIC_STABLESPREAD/USDC": 6
 };
 
 const getPrecisionForIdentifier = identifier => {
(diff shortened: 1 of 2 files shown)


packages/common/src/PriceIdentifierUtils.js
```diff
@@ -6,7 +6,11 @@ const IDENTIFIER_BLACKLIST = {
 // Price identifiers that should resolve prices to non 18 decimal precision. Any identifiers
 // not on this list are assumed to resolve to 18 decimals.
 const IDENTIFIER_NON_18_PRECISION = {
-  USDBTC: 8
+  USDBTC: 8,
+  "STABLESPREAD/USDC": 6,
+  "STABLESPREAD/BTC": 8,
+  "ELASTIC_STABLESPREAD/USDC": 6,
+  BCHNBTC: 8
```
smb2796, line 13:
```suggestion
  BCHNBTC: 8,
  ETHBTC: 8
```

nicholaspai: Hm this is tricky, because previously `ETHBTC` was used by an EMP that used DAI as collateral. Do you know that the next EMP to use this identifier will be collateralized by renBTC?

smb2796: Yeah this is tricky. There is interest in launching an ETH/BTC EMP collateralized by renBTC

nicholaspai: Hm let's just add the ETHBTC binding once the EMP is deployed then

smb2796: Good with this approach

mrice32: +1 -- we should let them know that they should propose an UMIP amendment.

mrice32 (approved):
LGTM!

chrismaree (approved):
Useful change! Note that we now have two definitions of similar information. There is also decimal and scaling definitions that are added to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js within https://github.com/UMAprotocol/protocol/pull/2276/files depending on the call of a price feed. This common should perhaps be refactored to be used within that set of defaults as well as within the voterdapp.

nicholaspai:
> Useful change! Note that we now have two definitions of similar information. There is also decimal and scaling definitions that are added to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js within https://github.com/UMAprotocol/protocol/pull/2276/files depending on the call of a price feed. This common should perhaps be refactored to be used within that set of defaults as well as within the voterdapp.

Great point. Notice that the precision bindings are defined in a `common` file: `packages/common/src/PriceIdentifierUtils.js`, why can't we grab these in `DefaultPriceFeed.js` or `CreatePriceFeed.js` from this same file?

PR #2368, opened 2021-01-02
improve(balancer-price-feed): Add TWAP

### LINKED ISSUE #2367: Add TWAP option to BalancerPriceFeed + other features
**Why**

BalancerPriceFeed currently returns prices by calling `getSpotPriceSansFee()` but this should be TWAP'd similar to how the UniswapPriceFeed implements it. DEX price feeds are inherently more manipulatable than REST API based price feeds and we should ensure that bots depending on these price feeds cannot be so easily exploited.

BalancerPriceFeed should also handle cases where the `bpool` contract returns prices in non-18 decimal precision and the user of the PriceFeed wants different precision.

- 5: ~4 hours

nicholaspai:

**Motivation**

#2367

**Summary**

In addition to addressing #2367:
- Adds additional refactoring to BalancerPriceFeed
- Changes BalancerPriceFeed tests to match that of UniswapPriceFeed to enforce consistent interfaces between DEX PriceFeeds, which should behave similarly to pricefeed clients.

**Issue(s)**

Fixes #2367


packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -1,13 +1,30 @@
...
+    logger,
+    web3,
+    getTime,
+    abi,
+    address,
+    tokenIn,
+    tokenOut,
+    lookback,
+    twapLength,
+    poolDecimals = 18,
```
chrismaree, line 21: This change (and the `decimals`) field below will need to change slightly following https://github.com/UMAprotocol/protocol/pull/2276/files getting merged. Specifically, the price feed decimals should return a price scaled in the same way that the DVM would respond to a price request. It's not a large change but does have a small impact on the interface and how you define the scaling within the feed.

chrismaree (commented):
this looks great! thanks for updating this. Left one comment on decimals which we can hopefully get merged in soon once https://github.com/UMAprotocol/protocol/pull/2276/files has been merged.

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
nicholaspai: Cool makes sense I'll hold off on merging until #2276 is merged

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -20,47 +36,140 @@ class BalancerPriceFeed extends PriceFeedInterface {
...
+      blocks = await this.blockHistory.update(this.lookback + this.twapLength, currentTime);
     }
+    // The priceHistory.update() method should strip out any blocks where the price is null
     await Promise.all(blocks.map(this.priceHistory.update));
+
+    this.lastUpdateTime = currentTime;
+  }
+  // If priceHistory only encompasses 1 block, which happens if the `lookback` window is 0,
+  // then this should return the last and only price.
+  _computeTwap(startTime, endTime) {
```
mrice32, line 140: I wonder if we should make this a utility function that takes in the start time, end time, and the list of price points just so we don't have to have it in two places. The only reason I suggest it here is that the logic is somewhat tricky, so I'd prefer to only have it written in one place to avoid discrepancies. Thoughts?

packages/financial-templates-lib/test/truffle/BalancerPriceFeed.js
```diff
@@ -0,0 +1,430 @@
+const { toWei, toBN } = web3.utils;
```
mrice32, line 1: I actually really like that this is a copy-paste of the uniswap tests, since it basically ensures that the behavior matches exactly.

I could imagine that there are some edge cases that are unique to each implementation that we want to test, but overall, I think it really makes sense to hit them with the same test cases.

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -36,36 +37,41 @@ exports.BlockHistory = (getBlock, blocks = []) => {
...
-    blockNumber = block.number;
-    if (has(blockNumber)) return fetchBetween(endTime, blockNumber - 1);
-    insert(block);
-    return fetchBetween(endTime, blockNumber - 1);
-  }
-
   // Main call to update cache, will take care of fetching all blocks, caching and pruning cache.
-  async function update(age, now, startBlock) {
-    assert(age >= 0, "requires age in seconds");
+  async function update(lookback, now) {
```
mrice32, line 45: Would like to get @daywiss's eyes on some of these deeper changes to these utils classes.

mrice32 (commented):
Overall looks great!! I really appreciate the focus on consistency with the uniswap feed as we're going to encounter a bunch of situations where they need to be used interchangeably. +1. This smashes a lot of good tech debt!

packages/financial-templates-lib/test/truffle/BalancerPriceFeed.js
nicholaspai: I'd like to eventually merge these two tests into a `TwapPriceFeed` generalized test that should work for any twap'd price feed.

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
nicholaspai: Yeah I like this I'll implement.

nicholaspai: Done PTAL!

daywiss (approved):
Small nits and questions but nothing that i would block for. LGTM

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -36,36 +37,41 @@ exports.BlockHistory = (getBlock, blocks = []) => {
...
+    // Note, we make an informed approximation about the block height that corresponds to the earliest timestamp,
+    // this allows us to query all block heights from this early number to the current number in parallel, instead of
+    // having to traverse backwards sequentially from the current number to this early number.
+    const latestBlock = await getBlock();
+    const latestBlockHeight = latestBlock.number;
+    // Add 50 block height buffer just to be safe, and if the result is negative then set it to 0. On a test network it
+    // is possible for the `earliestBlockHeight` to be negative.
+    const earliestBlockHeight = Math.max(
+      0,
+      latestBlockHeight - Math.floor(lookback / (await averageBlockTimeSeconds())) - 50
```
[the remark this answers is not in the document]
nicholaspai: Done thanks for the catch

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -135,36 +134,10 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
-      lastPrice = event.price;
-      lastTime = event.timestamp;
-    }
-
-    if (timeSum === 0) {
-      return null;
-    }
-
-    return priceSum.divn(timeSum);
+    const events = eventsIn.slice().map(e => {
```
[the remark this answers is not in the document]
nicholaspai: Done

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -158,3 +166,39 @@ exports.PriceHistory = (getPrice, prices = {}) => {
     list
   };
 };
+
+// Given a list of price events [timestamp, price] and a time window, returns the time-weighted
+// average price.
+exports.computeTWAP = (events, startTime, endTime, startingPriceSum) => {
```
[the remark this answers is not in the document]
nicholaspai: Ah good catch, added

PR #2369, opened 2021-01-04
improve(affiliates): add dapp mining tests, mocks and mock data 

### LINKED ISSUE #2365: Add dapp mining tests dataset for tests
Dapp Mining PR was already pretty large, in order to get a review going, tests were omitted and the logic was tested in "production" with several live contracts which have some tagged transactions already.

Actual tests should be added in, this includes basic unit tests and an E2E style tests with mock data from the live chain if possible.

daywiss:
**Motivation**
#2365

**Summary**

Adds a dataset generator specifically for dapp mining, and updates mocks to add "trace" calls.
Adds unit tests for most dapp mining utility functions and a full dataset test using data saved from mainnet contract which contains a handful of tagged and untagged create events.
Also does some minor refactoring with names to get dev mining code more in line with the correct naming.

**Issue(s)**

Fixes #2365


packages/affiliates/test/e2e/datasets.js
```diff
@@ -173,34 +173,7 @@ describe("Datasets", function() {
       assert(result.length);
     });
   });
-  describe("Synthprices Serializer", function() {
```
daywiss, line 176: I removed these tests mainly because they weren't working over the specified time range, and I deemed this ok, since this is more like an E2E test and is not part of CI.

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+  return {
+    blockNumber,
+    name: "create",
+    from_address: user,
+    args: [collateral, tokens],
+    input: encodeAttribution(encodeCallData("create", [collateral.toString()], [tokens.toString()]), tag)
+  };
+}
+
+describe("DappMining", function() {
```
daywiss, line 31: This starts the unit tests for the dapp mining logic. In here several functions are tested standalone and at the end one integration test using a downloaded (and unmodified) dataset with tagged and untagged transactions.

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+      attributions.attribute("c", "Dev2", "15");
+      attributions.attribute("d", "Dev1", "20");
+      attributions.attribute("d", "Dev2", "20");
+      let result = await utils.processRewardData({ attributions, balances, whitelist, totalRewards });
+      assert.equal(result["Dev1"], "50");
+      assert.equal(result["Dev2"], "50");
+    });
+  });
+  describe("dataset1", function() {
+    it("should run with dataset", async function() {
```
daywiss, line 222: this dataset is based on the  YD-ETH-MAR21 contract, and has some tagged and untagged transactions on mainnet which makes it a decent test.

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+      result = utils.calculatePercent(100, toWei("1"));
+      assert.equal(result, "100");
+    });
+    it("sums attributions", function() {
+      const attributions = SharedAttributions();
+      let sum;
+      sum = utils.sumAttributions(attributions);
+      assert(sum);
+      attributions.attribute("a", "a", "1");
+      attributions.attribute("a", "b", "2");
```
daywiss, line 51: these attribution calls are essentially saying, attribute creation event by sponsor "a" to developer "b" for 2 tokens. The call is this: `attribution(sponsor,developer,amount)`.

This ends up creating a structure where a user can have attributions by multiple developers (a and b) for different amounts.

packages/affiliates/libs/datasets.js
```diff
@@ -204,8 +215,26 @@ function Dataset(basePath, { queries, coingecko, synthPrices }) {
     return path;
   }
 
+  async function saveDappMining(name, config) {
```
daywiss, line 218: this saves dataset for dapp mining which omits some data that dev mining needs, such as coingecko and synth data.

packages/affiliates/libs/contracts.js
```diff
@@ -70,14 +70,29 @@ function GetInputLength(abi) {
...
 }
 
-const DecodeAttribution = abi => {
+// Given a transaction, decode the attribution tag from the function. By default only suppports create.
+const DecodeAttribution = (abi, name = "create") => {
   // convert bits to hex (div by 4) and add 2 for 0x
-  const inputLength = GetInputLength(abi)("create") / 4 + 2;
-  return createTransaction => {
-    assert(createTransaction.name == "create", "Only decodes create transactions");
+  const inputLength = GetInputLength(abi)(name) / 4 + 2;
```
mrice32, line 76: Somewhat unrelated, but took a brief look at your GetInputLength implementation. One idea that may be worth toying with is seeing if you can get web3 or ethers to do some of the heavy lifting for you. web3 has a simple interface for interacting with abis: https://web3js.readthedocs.io/en/v1.3.0/web3-eth-abi.html. Ethers has a more complex, but more complete, library for it: https://docs.ethers.io/v5/single-page/#/v5/api/utils/abi/.

One idea for a more generalized implementation to decode the attribution is to use the ethers `parseTransaction` [method](https://docs.ethers.io/v5/single-page/#/v5/api/utils/abi/interface/-%23-Interface--parsing). You could basically do two operations (may require a little testing):
1. Parse the transaction in full. Get the result (or catch the error).
2. Parse the transaction with the last 20 bytes removed. Get the result (or catch the error).

My guess is that you could use the errors and/or outputs to definitively determine whether the transaction was tagged without explicitly needing to calculate the data lengths yourself.

Not sure how necessary any of this is in the short term, just thought about it as I was reading through, and I wanted to mention the abi libs if you hadn't seen them.

packages/affiliates/test/e2e/datasets.js
```diff
@@ -213,10 +186,10 @@ describe("Datasets", function() {
         end: params.startingTimestamp + 1000 * 60 * 60 * 24 * 10
       };
       const ds = Dataset(basePath, { queries, coingecko, synthPrices });
-      path = await ds.save("test-set", config);
+      path = await ds.saveDevMining("dev-mining-test", config);
```
mrice32, line 189: nit: is this test saving data to disk? My understanding was that the on-disk data was generated and put into the repo rather than being generated during the e2e test... can you familiarize me with the architecture? Or is this a test for the infrastructure that saves the on-disk data?

packages/affiliates/test/e2e/datasets.js
```diff
@@ -231,9 +204,47 @@ describe("Datasets", function() {
...
+        empAddress: "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5",
+        startTime: 1607558400000,
+        endTime: 1609113600000
+      };
+      assert(path, "requires dataset path");
+      const query = mocks.Queries(path);
+      const blocks = await query.getBlocks(config.startTime, config.endTime);
+      assert(blocks.length);
+      const descBlocks = await query.getBlocksDescending(config.endTime, 2);
+      assert(descBlocks.length, 2);
```
mrice32, line 236: What is this assert doing? IIUC, this is asserting that the length is >= 0 and if it errors it prints an error message of 2?

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+      attributions.attribute("b", "b", "3");
+      sum = utils.sumAttributions(attributions, 10, sum);
+      assert.equal(sum.getAttribution("a", "a"), "10");
+      assert.equal(sum.getAttribution("a", "b"), "20");
+      assert.equal(sum.getAttribution("b", "b"), "30");
+    });
+    it("sums balances", function() {
+      const balances = Balances();
+      let sum;
+      sum = utils.sumBalances(balances);
```
mrice32, line 61: Can this not be:
```
let sum = utils.sumBalances(balances);
```
?

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+      assert(sum);
+      attributions.attribute("a", "a", "1");
+      attributions.attribute("a", "b", "2");
+      attributions.attribute("b", "b", "3");
+      sum = utils.sumAttributions(attributions, 10, sum);
+      assert.equal(sum.getAttribution("a", "a"), "10");
+      assert.equal(sum.getAttribution("a", "b"), "20");
+      assert.equal(sum.getAttribution("b", "b"), "30");
+    });
+    it("sums balances", function() {
```
mrice32, line 58: +1 these tests are super simple to read and understand.

mrice32 (commented):
Looks great! A few minor comments

packages/affiliates/test/e2e/datasets.js
daywiss: This does save data on disk, but the files get removed at the end of the test. This is one of the tests that does not run in CI and needs to be run manually. It is testing the actual dataset generation utility from end to end, so that involves it saving data to disk and verifying that it can be read back out. 

The datasets that are committed were generated by the utility being tested here, and are used in the other tests that calculate reward output.

packages/affiliates/test/e2e/datasets.js
daywiss: that was a typo, its supposed to be asserting that length is equal to 2!

chrismaree (commented):
This looks awesome! I did an initial review now. I want to take another as logs of stuff happening here. Have added some initial comments/questions/nits.

packages/affiliates/test/datasets/dapp-mining-set1/logs_0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5.txt
```diff
@@ -0,0 +1,94 @@
+{"log_index":204,"transaction_hash":"0x2f23b55607432961f4659065969fbb7094ab75b086acda6b6a1b73756ea93a11","transaction_index":131,"address":"0xe4256c47a3b27a969f25de8bef44eca5f2552bd5","data":"0x","topics":["0x4b82aa16e071a61de1a6b9aeec9edab0356331f8122c78683b469ac8e685dabc","0x000000000000000000000000049355e4380f8db88cb8a6ec0426b1a1a3560c67","0x00000000000000000000000000000000000000000000000003782dace9d90000","0x0000000000000000000000000000000000000000000000056bc75e2d63100000"],"block_timestamp":{"value":"2020-12-18T17:01:55.000Z"},"block_number":11478405,"block_hash":"0x891f28bd91de00c369c7425ff592d4351cad56e96aa1604a49ba68f0247ebd59"}
```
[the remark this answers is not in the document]
daywiss: It looks like json but its not a valid json file since the objects are not wrapped in an array or object. I didnt want to give the impression that you could import it into node without some pre-processing.

packages/affiliates/test/e2e/datasets.js
mrice32: Might be worth doing a brief audit of these tests to ensure we're not doing this anywhere else!

@chrismaree another instance where it might be useful to have a lint rule around assert. Maybe the right way to structure it is a lint rule that requires the second argument to assert to always be a string literal.... although there may be cases where a string literal is being tested against, so that's not a perfect solution...

packages/affiliates/libs/contracts.js
daywiss: @mrice32 Thanks for showing me this. One thing to note is that `parseTransaction` is currently used to parse the transaction data before tags are looked at. But I think what you are suggesting is that we slice off an address length chunk of data from the raw unparsed transaction string and see if it parses, if so, we have a tag. If it errors then parse the full intact string, and therefore no tag.

This approach works because it assumes your tags are address sized, which I wanted to avoid this assumption in case we ended up with some other kind of scheme for encoding tags (even variable length), but I think it could work with the current assumption of tag size and be quite a bit simpler to implement

The other thing is that you have pointed out an interesting observation, this works because data is encoded at the end of every transaction so you can assume the tag will always be at the end of the raw tx string. I did not realize that but its good to know.

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,230 @@
...
+      assert(sum);
+      balances.add("a", "1");
+      balances.add("b", "2");
+      balances.add("c", "3");
+      sum = utils.sumBalances(balances, 10, sum);
+      assert.equal(sum.get("a"), "10");
+      assert.equal(sum.get("b"), "20");
+      assert.equal(sum.get("c"), "30");
+    });
+    it("calculates block reward", function() {
```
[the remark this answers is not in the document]
daywiss: I added significantly more comments to the dap mining tests as well as an explainer to the top of the e2e tests for the dataset.

packages/affiliates/test/e2e/datasets.js
mrice32: Yeah, preferably in the readme if not already there.

packages/affiliates/test/e2e/datasets.js
```diff
@@ -173,50 +180,23 @@ describe("Datasets", function() {
...
-      this.timeout(10000);
+      this.timeout(100000);
       const config = {
         ...params,
         start: params.startingTimestamp,
         end: params.startingTimestamp + 1000 * 60 * 60 * 24 * 10
       };
       const ds = Dataset(basePath, { queries, coingecko, synthPrices });
-      path = await ds.save("test-set", config);
+      path = await ds.saveDevMining("dev-mining-test", config);
```
mrice32, line 196: nit: have you considered having these files saved to a temp dir so they have no chance of polluting the user's working directory (if something fails) and they will be automatically deleted by the OS? https://www.gnu.org/software/autogen/mktemp.html. May not be easy in js, but just wanted to toss out the suggestion!

packages/affiliates/libs/contracts.js
mrice32: Yep, that's what I'm suggesting.

It seems that we're looking at the problem in two different ways:
- I'm suggesting that we infer the details of the transaction data by assuming the structure of the tag.
- I think you're suggesting that we infer the details of the tag by assuming the structure of the transaction data.

I think those are both reasonable approaches. My only minor argument for the former is that we have many types of transactions in our system that vary substantially, but our tags are fairly simple. So my intuition is that assuming that the structure of the tags is stable and simple is safer than assuming that the structure of the transaction data is at least simple (i.e. no variable-length data or ABIV2 or variable length data complexities).

Either way, this LGTM! Just wanted to mention it :)

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -0,0 +1,265 @@
...
+    it("calculates percent", function() {
+      let result = utils.calculatePercent(100, toWei(".01"));
+      assert.equal(result, "1");
+      result = utils.calculatePercent(100, toWei("1"));
+      assert.equal(result, "100");
+    });
+    it("sums attributions", function() {
+      const attributions = SharedAttributions();
+      let sum;
+      sum = utils.sumAttributions(attributions);
```
mrice32, line 48: nit: can we combine these lines?

mrice32 (approved):
LGTM!

packages/affiliates/test/e2e/datasets.js
daywiss: thats a really good idea, but im not sure how to do that in an OS agnostic way, but maybe its good enough to run in linux and mac. Implementation would be a bit tricky as well, but i do like the idea since there is a chance this test could fail and the user would not know to clean it up.

packages/affiliates/test/mocha/dappmining.js
daywiss: yes sorry, must have had that in 2 places

packages/affiliates/test/e2e/datasets.js
daywiss: I didnt realize this but there is currently a README.md in the test/e2e folder that contains this: 

> These tests run the entire stack and reach out to external services or write to disk. The idea is that you can test as if you were in "production" as a last check to make sure everything works. These should be run locally, but maybe in the future they can be added to CI in some way.

packages/affiliates/test/e2e/datasets.js
daywiss: Looked through the tests and did not see any other violations

chrismaree (approved):
This is one beast of a PR! Some really good looking tests that seem to cover all cases we really care about. Nice refactors as well for more consistent naming and layout. Overall Looks good to me!

packages/affiliates/test/e2e/datasets.js
daywiss: I've updated readme to mention this folder exists, and how to clean it up

PR #2356, opened 2020-12-29
docs: add package readmes for half of remaining packages

mrice32:
**Motivation**

Packages don't have READMEs.

**Summary**

Adds READMEs for core, financial-templates-lib, common, disputer, and liquidator. Packages remaining without READMEs (cc @pemulis): reporter, monitors, and serverless-orchestration.

**Issue(s)**

N/A

PR #2372, opened 2021-01-04
docs: add readmes for monitor and reporter bots and serverless orchestration scripts

pemulis:
**Motivation**

We need more package readmes.

**Summary**

Adds readmes for `monitors`, `reporters`, and `serverless-orchestration`.

**Issue(s)**

N/A

### DIFF
--- a/packages/monitors/README.md
+++ b/packages/monitors/README.md
@@ -1,0 +1,17 @@
+# @uma/monitors
+
+## Installing the package
+
+## Importing the package
+
+## Monitors
+
+The four monitors available are:
+
+1. The `BalanceMonitor`
+
+1. The `ContractMonitor`
+
+1. The `CRMonitor`
+
+1. The `SyntheticPegMonitor`
--- a/packages/monitors/README.md
+++ b/packages/monitors/README.md
@@ -7,7 +7,7 @@ For more information about running a monitor bot, see the [docs](https://docs.um
 ## Installing the package
 
 ```bash
-yarn global add @uma/monitors
+yarn add @uma/monitors
 ```
 
 ## Running the monitors
--- a/packages/monitors/README.md
+++ b/packages/monitors/README.md
@@ -12,10 +12,10 @@ yarn add @uma/monitors
 
 ## Running the monitors
 
-The config below will start up a monitor bot that will: (1) send messages when new liquidations, alerts, or disputes occur and (2) fire if there is large volatility in the synthetic or price of the underlying. It wont report on any wallet or CR monitoring as no params have been defined.
+The config below will start up a monitor bot that will: (1) send messages when new liquidations, alerts, or disputes occur and (2) fire if there is large volatility in the synthetic or price of the underlying. It won't report on any wallet or CR monitoring as no params have been defined.
 
 ```bash
-EMP_ADDRESS=0x1234 CUSTOM_NODE_URL=your.node.url MNEMONIC="your mnemonic here" monitors --network mainnet_mnemonic
+EMP_ADDRESS=0x1234 CUSTOM_NODE_URL=your.node.url MNEMONIC="your mnemonic here" yarn monitors --network mainnet_mnemonic
 ```
 
 ## Monitors
--- a/packages/monitors/README.md
+++ b/packages/monitors/README.md
@@ -1,17 +1,31 @@
 # @uma/monitors
 
+This package contains four UMA monitor bot implementations: `BalanceMonitor`, `ContractMonitor`, `CRMonitor`, and `SyntheticPegMonitor`. The monitor bots are used to monitor the UMA ecosystem for key events. They have a number of monitor modules built into them that enable real time reporting on key events for a given EMP contract.
+
+For more information about running a monitor bot, see the [docs](https://docs.umaproject.org/developers/bots).
+
 ## Installing the package
 
-## Importing the package
+```bash
+yarn global add @uma/monitors
+```
+
+## Running the monitors
+
+The config below will start up a monitor bot that will: (1) send messages when new liquidations, alerts, or disputes occur and (2) fire if there is large volatility in the synthetic or price of the underlying. It wont report on any wallet or CR monitoring as no params have been defined.
+
+```bash
+EMP_ADDRESS=0x1234 CUSTOM_NODE_URL=your.node.url MNEMONIC="your mnemonic here" monitors --network mainnet_mnemonic
+```
 
 ## Monitors
 
 The four monitors available are:
 
-1. The `BalanceMonitor`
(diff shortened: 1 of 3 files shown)


packages/monitors/README.md
```diff
@@ -0,0 +1,31 @@
...
+```bash
+yarn global add @uma/monitors
+```
+
+## Running the monitors
+
+The config below will start up a monitor bot that will: (1) send messages when new liquidations, alerts, or disputes occur and (2) fire if there is large volatility in the synthetic or price of the underlying. It wont report on any wallet or CR monitoring as no params have been defined.
+
+```bash
+EMP_ADDRESS=0x1234 CUSTOM_NODE_URL=your.node.url MNEMONIC="your mnemonic here" monitors --network mainnet_mnemonic
```
nicholaspai, line 18: Have you tested that this is minimal viable config neccessary to run the bots? Also, shouldn't the execution command be `yarn monitors ...`

packages/monitors/README.md
```diff
@@ -0,0 +1,31 @@
...
+```
+
+## Monitors
+
+The four monitors available are:
+
+1. The `BalanceMonitor` takes a specified list of addresses to monitor and sends alerts if their collateral synthetic or ether balance drops below defined thresholds.
```
nicholaspai, line 25:
```suggestion
1. The `BalanceMonitor` takes a specified list of addresses to monitor and sends alerts if their collateral, synthetic or Ether balance drops below defined thresholds.
```

packages/monitors/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+1. The `ContractMonitor` sends alerts when liquidation, dispute, and dispute settlement events occur.
```
nicholaspai, line 27:
```suggestion
1. The `ContractMonitor` sends alerts when financial contract events occur, such as liquidations and disputes.
```

packages/serverless-orchestration/README.md
```diff
@@ -0,0 +1,15 @@
+# @uma/serverless-orchestration
+
```
nicholaspai, line 2:
```suggestion

This package contains scripts that enable bots to be run in a serverless fashion.
```

packages/monitors/README.md
pemulis: Running some tests this afternoon! Was using the examples from the main readme branch as a starting point.

mrice32: Agreed, do the monitors work with this few params?

@nicholaspai `monitors` alone works when you install the package globally as was done in the previous section.

@pemulis see the thread on my PR here: https://github.com/UMAprotocol/protocol/pull/2356#discussion_r550169440. @chrismaree suggested that we not install these globally, so I removed the `global` from `yarn global add liquidator` and I added `yarn` to the run command, making it `... yarn liquidator --network ...`.

pemulis: I also wasn't sure how to run the serverless orchestration or reporters, so those could use some fleshing out.

pemulis: Working from this section of the docs on the monitor bots, by the way:

https://docs.umaproject.org/developers/bot-param#minimum-viable-monitor-config

Hope to have a better handle on the bots from the review later this week.

packages/serverless-orchestration/README.md
```diff
@@ -0,0 +1,16 @@
...
+This package contains scripts that enable bots to be run in a serverless fashion.
+## Installing the package
+
+```bash
+yarn add @uma/serverless-orchestration
+```
```
mrice32, line 8: Note: this package is somewhat complex to run locally (or at least when using it to run the bots). Would be interested to hear @chrismaree's opinion as to what we should put in this readme and what is out of scope.

packages/reporters/README.md
```diff
@@ -0,0 +1,17 @@
+# @uma/reporters
+
+This package contains two UMA reporter bot implementations: `GlobalSummaryReporter` and `SponsorReporter`.
+
+## Installing the package
+
+```bash
+yarn add @uma/reporters
+```
```
mrice32, line 9: Not sure if we've run the monitors in a while. Were you able to do a local run?

packages/monitors/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+## Installing the package
+
+```bash
+yarn add @uma/monitors
+```
+
+## Running the monitors
+
+The config below will start up a monitor bot that will: (1) send messages when new liquidations, alerts, or disputes occur and (2) fire if there is large volatility in the synthetic or price of the underlying. It won't report on any wallet or CR monitoring as no params have been defined.
```
mrice32, line 15: It would be great to add a little documentation around how you can add a slack webhook or pagerduty info to get the monitor to alert you outside of the console. We don't have to go into great detail, but would be nice to mention the env variables (in a separate section).

mrice32 (commented):
Very awesome work here! I think there's still one or two sections to add in a few of these, but overall, this is very concise and well-explained. +1 -- thanks for putting all of this together!

packages/monitors/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```bash
+EMP_ADDRESS=0x1234 CUSTOM_NODE_URL=your.node.url MNEMONIC="your mnemonic here" yarn monitors --network mainnet_mnemonic
+```
+
+## Monitors
+
+The four monitors available are:
+
+1. The `BalanceMonitor` takes a specified list of addresses to monitor and sends alerts if their collateral, synthetic or Ether balance drops below defined thresholds.
```
chrismaree, line 25: nit: the data structures for this is a bit complex. we should perhaps show an example config that contains all the execution examples. WDYT?

Something like this:
```
 "walletsToMonitor": [
          {
            "name": "Monitored sponsor",
            "address": "0x...",
            "crAlert": 1.5
          }
        ],
```

The same goes for all other sections and monitors. it can be a bit of a pain to configure them so examples to copy-paste would be useful. You can make a redacted version of the `bot-configs` repo?

packages/reporters/README.md
chrismaree: I'm 90% sure they are actually currently broken. we should consider removing them from the repo. WDYT @nicholaspai ?

packages/serverless-orchestration/README.md
chrismaree: I think for now it should be out of scope. As a task for one of the following sprints, when looking into building out generalized bot infrastructure, we will refine this execution process and can improve the documentation then. 

For now, it's not really worth going into detail on how to run it for this reason I'd say.

chrismaree (commented):
Looks good to me! Left a few comments and nits. thanks for doing this tasks :)

PR #2379, opened 2021-01-06
feat(financial-templates-lib): add `DominationFinancePriceFeed` for BTCDOM, ALTDOM (UMIP-21)

### LINKED ISSUE #2378: Add price feed for BTCDOM and ALTDOM price identifiers
**Why**

There currently is no price feed support for the BTCDOM and ALTDOM price identifiers that were introduced in UMIP-21.

**How**

The Domination Finance Team in collaboration with CoinGecko has an API available for querying the current and historical prices of the two price identifiers (adhering to the UMIP rounding rules, etc.)

The following API's can be used:
 - Current Price:
   - https://api.domination.finance/api/v0/price/BTCDOM
   - https://api.domination.finance/api/v0/price/ALTDOM
 - Historical Price (minutely ticks for 72)
   - https://api.domination.finance/api/v0/price/BTCDOM/history
   - https://api.domination.finance/api/v0/price/ALTDOM/history

**Resolution**

 - Add price feed implementation and unit test cases for BTCDOM and ALTDOM

**Difficulty Score [1-10]**

- 3: ~1 hour

jaxrtech:
**Motivation**

This adds support for querying the current and 72 hour historic prices for the BTCDOM and ALTDOM price identifiers introduced in UMIP-21 via a new price feed. There previously was no support for these price identifiers.

**Summary**

 - Added test cases similar to the ones for the Cryptowatch price feed
 - Updated the `core/scripts/local/getMedianHistoricalPrice.js` script to support the BTCDOM, ALTDOM price identifiers (with appropriate UMIP rounding)

**Details**

Uses the following API URLs:
 - Current Price:
   - https://api.domination.finance/api/v0/price/BTCDOM
   - https://api.domination.finance/api/v0/price/ALTDOM
 - Historical Price (minutely ticks for 72)
   - https://api.domination.finance/api/v0/price/BTCDOM/history
   - https://api.domination.finance/api/v0/price/ALTDOM/history

**Issue(s)**

Fixes #2378


chrismaree (commented):
This feed is great! good job on updating all the associated places such as `CreatePriceFeed`, the `DefaultPriceFeedConfigs` and the `getMedianHistoricalPrice`. I left a comment on some nits/changes I'd like you to make for consistency.

To double-check something: the intended collateral for all domfin contracts is Dai, correct? Just double-checking that the scaling of the price feed is compatible with what you guys have deployed.

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
```diff
@@ -0,0 +1,240 @@
...
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Instance used for Web3 utilities and conversions.
+   * @param {String} pair Representation of the pair the price feed is tracking.
+   *    The string should be the representation used by the DomFi API to identify this pair.
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the Unix timestamp in seconds.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Number} decimals Number of decimals to use to convert price to wei.
```
chrismaree, line 17: This is great! Note it looks like you build your `DominationFinancePriceFeed` by mimicking the `CryptoWatchPriceFeed`, which is the perfect way to do this seeing it's got the right layout and structure.

There are a few small nits/changes I'd like you to make. After you copied  `CryptoWatchPriceFeed` we've made some pretty meaningful changes to how we deal with decimals within our feeds. In particular, feeds can be scaled to non-18 decimals to accommodate contracts with non-18 decimal collateral types. Note that the `CryptowatchPriceFeed` _could_ previously do this but we did make some changes to the flow. you can check it out here: https://github.com/UMAprotocol/protocol/pull/2276

To make your PR compatible with how we now deal with decimals could you please make the following changes:

- rename `decimals` to `priceFeedDecimals` throughout this PR.
- rename `convertDecimals` to `convertPriceFeedDecimals` throughout this PR.
- add a new method `priceFeedDecimals` that simply returns `return this.priceFeedDecimals`.

After you've made these changes you should be up to date with our latest structure. Because your feeds are 18 decimal scaled I don't think you really need to change anything else.

jaxrtech:
> This feed is great! good job on updating all the associated places such as `CreatePriceFeed`, the `DefaultPriceFeedConfigs` and the `getMedianHistoricalPrice`. I left a comment on some nits/changes I'd like you to make for consistency.
> 
> To double-check something: the intended collateral for all domfin contracts is Dai, correct? Just double-checking that the scaling of the price feed is compatible with what you guys have deployed.

Thanks @chrismaree for your time and feedback on this PR! I can get those comments resolved.

Yes, all the current DomFi contracts are using DAI as collateral (so the usual 18 decimals)

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
jaxrtech: Just to be on the same page with the "number of decimals" logic being used here -- so `priceFeedDecimals` is solely there to allow the client of the `*PriceFeed` classes the ability to shift the number of decimal places the BigNumber returns, correct? This doesn't have anything to do with the UMIP rounding rules, right? (At the moment, the API will perform the appropriate rounding according to the UMIP and returns those values as a string for simplicity)

I'll get those renamed and add the new `priceFeedDecimals` method 👍🏾

packages/core/scripts/local/getMedianHistoricalPrice.js
```diff
@@ -15,7 +15,9 @@ require("dotenv").config();
 
 const UMIP_PRECISION = {
```
nicholaspai, line 16: Its better actually to define these in [this file](https://github.com/UMAprotocol/protocol/blob/master/packages/common/src/PriceIdentifierUtils.js) because this is imported into the Liquidator/Disputer bots. 

This file `getMedianHistoricalPrice` should actually be updated to import from that file as well.

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
```diff
@@ -0,0 +1,240 @@
...
+      console.log(
+        '- This will return the historical prices as "data.rows". Each row contains: [unix_timestamp, price].'
+      );
+      console.log("- Note that you might need to invert the prices for certain identifiers.");
+      console.groupEnd();
+    }
+    return returnPrice;
+  }
+
+  getHistoricalPricePeriods() {
```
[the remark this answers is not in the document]
jaxrtech: Thanks! Just to be clear, at the moment the API only returns the minutely, 72 hour lookback. We internally have a separate system that handles arbitrary tick size and arbitrary lookback for charting purposes that isn't publicly documented (yet) since we've been still making adjustments here and there.

This would be a separate task/issue that we can tackle before UMA needs to run the numbers for the mining rewards, but if you know offhand what tick size and lookback would be necessary for the dev mining rewards, we likely can adjust our public API accordingly to support it and make a separate PR. Worst case, we can run some ad-hoc SQL for you lol. If this requires more discussion, we can always loop back in the Discord. Just wanted to give full disclosure regarding that.

packages/core/scripts/local/getMedianHistoricalPrice.js
jaxrtech: Thanks, @nicholaspai, I'll go on and take care of updating both of those 👍🏾

nicholaspai: Hey @jaxrtech actually I'm mistaken. There is a difference between the precision used for UMIP's and for the identifiers. Will the BTCDOM ETHDOM be OK to resolve to 18 decimal precision?

jaxrtech: @nicholaspai Should commit 8c940bceb5e4b1e41003d96d6b0f3c381eb06466 be reverted then?

chrismaree: yes. that change needs to be rolled back. In particular, you should not set `IDENTIFIER_NON_18_PRECISION` within `packages/common/src/PriceIdentifierUtils.js` to 2. This will scale your feeds incorrectly. This number `IDENTIFIER_NON_18_PRECISION` is used to scale values returned from the DVM to accommodate non-18 decimal delimitation. For example, an identifier denominated in BTC will be scaled by 1e8. As your collateral type is Dai, you don't want to make this change & it should default to 18.

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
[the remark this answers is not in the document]
chrismaree: ok, that's not a problem. right now if your tokens are listed on coingecko then it's also fine. You have them listed there, right?

In the future, it would be ideal to expand this method to be able to return data looking further back. Regarding the tick size and lookback this is somewhat unimportant for the devmining. As long as there is some arbitrarily high precision data feed that can query up to a week in the past it should be fine. The Devmining is designed to work with different tick size data feeds.

packages/core/scripts/local/getMedianHistoricalPrice.js
jaxrtech: @chrismaree @nicholaspai  Makes sense, I've reverted the commit.

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
chrismaree: `priceFeedDecimals` defines the scaling of the feed itself and can be parameterized by the config passed into the feed. The point of it is to scale feeds in the _same way the DVM would return_ for a given vote. For older EMP products (and some that are currently on mainnet) that use non-18 decimal collateral types we would _scale the price feed_ according to the decimals of the collateral type. For example, contracts collateralized in BTC would have:
- 18 decimal synthetic
- 8 decimal collateral type
- 8 decimal price feed

For these non-18 decimal collateral types, we need to change how the price feed is called to match the collateral.

In https://github.com/UMAprotocol/protocol/pull/2276 we are changing this pattern to accommodate our new contracts that are currently in an audit by OZ. In the future, we will match the decimals of the synthetic and collateral and keep all feeds at 18 decimals. The above BTC example would then look like this:
- 8 decimal synthetic
- 8 decimal collateral type
- 18 decimal price feed

`priceFeedDecimals` within a price feed enables the feed to accommodate _either_ kind of decimals so that the bots remain backward compatible. The actual computation of where this scaling is used can be found within the EMP client, for example, which can be found here: https://github.com/UMAprotocol/protocol/blob/133c4a6e094e7c5dd35673b41a1ee13fd676b86a/packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js#L244

In the case of your product you want the price feed to be scaled to 18 decimals because a) your collateral type is Dai and b) you deployed with the old contracts that expect 18 decimal feeds.

Does that make sense?

jaxrtech: @chrismaree  Thanks, that makes a lot more sense with how the decimals are working from a high level between all the different components 👍🏾

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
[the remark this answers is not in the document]
jaxrtech: @chrismaree I've updated the PR to pull from a newly deployed API of ours that supports adjustable lookback and tick size. The only thing to keep in mind is that it will return a maximum of 5,000 rows per request so you likely will need to adjust the `tickPeriod` parameter to say 5 min = 300 secs (~2,000 rows for 1 week) and `lookback` as appropriate. Figured best to just take care of this now 😅 ...

Let me know if the PR is good to go! 👍🏾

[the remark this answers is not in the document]
chrismaree: Ok cool! will try it out. Should not block this PR though :)

chrismaree (approved):
This looks good to me! thanks for implementing all the changes requested.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -261,6 +261,16 @@ const defaultConfigs = {
...
+  },
+  BTCDOM: {
+    type: "domfi",
+    pair: "BTCDOM",
+    minTimeBetweenUpdates: 60
+  },
+  ALTDOM: {
+    type: "domfi",
+    pair: "ALTDOM",
+    minTimeBetweenUpdates: 60
```
nicholaspai, line 273: How come no lookback in the default even though it's a required param?

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
```diff
@@ -0,0 +1,252 @@
...
+    return `https://live.domination.finance/api/v0/history/${this.pair}?tick=${this.tickPeriod}s&range=${this.lookback}s`;
+  }
+
+  getCurrentPrice() {
+    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+  }
+
+  getHistoricalPrice(time, verbose = false) {
+    if (this.lastUpdateTime === undefined) {
+      return undefined;
```
nicholaspai, line 67: Usually we return null for situations when price cannot be computed. Undefined should only be returned if the function goes down an unexpected path

nicholaspai (approved):
Looks great just left some nits please take a look!

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
jaxrtech: This was the verbatim from the `CryptoWatchPriceFeed.js`. I've updated it there as well.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
jaxrtech: That's my bad, I must have accidentally removed it earlier, I realize `lookback` became `twapLength` in the others...
Anyhow, I've fixed it 👍🏾

PR #2385, opened 2021-01-08
fix(bots): Add feature to denominate synthetic price by another price feed + remove noisy BasketSpreadPriceFeed logs

nicholaspai:

**Motivation**

Currently, the `SyntheticPegMonitor` assumes that the `TOKEN_PRICE_FEED` (i.e. the "synthetic") and the `MEDIANIZER_PRICE_FEED` (i.e. the "peg") are returning prices denominated in the same currency. However, consider the case where the synthetic market is (zeldaCash - USDC) and the peg market is tracking (STABLESPREAD/ETH). This can result in the following erroneous log:

Screen Shot 2021-01-07 at 20 40 29

**Summary**

Adds an optional construction parameter to this monitor called the `denominatorPriceFeed` which will divide the synthetic price if available.

Example environment variable to divide the synthetic price by the price of ETHUSD:

```
TOKEN_DENOMINATOR_PRICE_FEED_CONFIG={"type":"cryptowatch","apiKey":"XXX","exchange":"coinbase-pro","pair":"ethusd","lookback":"3600","minTimeBetweenUpdates":60}
```

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
@@ -64,22 +64,12 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
       return null;
     }
     const experimentalMean = this._computeMean(experimentalPrices);
-    this.logger.debug({
-      at: "BasketSpreadPriceFeed",
-      message: "Average of experimental prices",
-      mean: experimentalMean.toString()
-    });
 
     // Second, compute the average of the baseline pricefeeds.
     if (baselinePrices.length === 0 || baselinePrices.some(element => element === undefined || element === null)) {
       return null;
     }
     const baselineMean = this._computeMean(baselinePrices);
-    this.logger.debug({
-      at: "BasketSpreadPriceFeed",
-      message: "Average of baseline prices",
-      mean: baselineMean.toString()
-    });
 
     // All calculations within this if statement will produce unexpected results if any of the
     // experimental mean, baseline mean, or denominator price are NOT in the same precision as
@@ -89,11 +79,6 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
     // TODO: Parameterize the lower (0) and upper (2) bounds, as well as allow for custom "spreadValue" formulas,
     // for example we might not want to have the spread centered around 1, like it is here:
     let spreadValue = experimentalMean.sub(baselineMean).add(this.convertPriceFeedDecimals("1"));
-    this.logger.debug({
-      at: "BasketSpreadPriceFeed",
-      message: "Basket spread value",
-      spreadValue: spreadValue.toString()
-    });
 
     // Ensure non-negativity
     if (spreadValue.lt(this.toBN("0"))) {
@@ -106,11 +91,6 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
 
     // Optionally divide by denominator pricefeed.
     if (!denominatorPrice) return spreadValue;
-    this.logger.debug({
-      at: "BasketSpreadPriceFeed",
-      message: "Denominator price",
-      denominatorPrice: denominatorPrice.toString()
-    });
     spreadValue = spreadValue.mul(this.convertPriceFeedDecimals("1")).div(denominatorPrice);
 
     return spreadValue;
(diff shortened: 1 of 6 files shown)


packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -64,22 +64,12 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
       return null;
     }
     const experimentalMean = this._computeMean(experimentalPrices);
-    this.logger.debug({
```
nicholaspai, line 67: These clog up the logs, removing now

mrice32: +1, this will help

mrice32 (approved):
Looks awesome. I was trying to consider other ways to handle the pool being against a different currency without having to provide yet another feed, but I don't have a good solution. LGTM!

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -18,7 +18,7 @@ class BalancerPriceFeed extends PriceFeedInterface {
     lookback,
     twapLength,
     poolDecimals = 18,
-    decimals = 18
+    priceFeedDecimals = 18
```
nicholaspai, line 21: This was a nasty bug: because this construction was named `decimals` and not `priceFeedDecimals`, then `DefaultPriceFeedConfigs` was not correctly overriding this precision.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -109,7 +109,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.lookback,
       config.twapLength,
       config.poolDecimals,
-      config.decimals // This defaults to 18 unless supplied by user
+      config.priceFeedDecimals // This defaults to 18 unless supplied by user
```
nicholaspai, line 112: `config.decimals` doesn't actually exist when the config is constructed by `DefaultPriceFeedConfigs`, so I renamed to `priceFeedDecimals` which `CryptoWatchPriceFeed` uses.

packages/monitors/index.js
```diff
@@ -72,21 +74,51 @@ async function run({
...
+    const priceFeedDecimals = medianizerPriceFeed.getPriceFeedDecimals();
+
     // 0. Setup EMP and token instances to monitor.
-    const [networkId, latestBlock, tokenPriceFeed, medianizerPriceFeed] = await Promise.all([
+    const [networkId, latestBlock, tokenPriceFeed, denominatorPriceFeed] = await Promise.all([
       web3.eth.net.getId(),
       web3.eth.getBlock("latest"),
-      createTokenPriceFeedForEmp(logger, web3, networker, getTime, empAddress, tokenPriceFeedConfig),
-      createReferencePriceFeedForEmp(logger, web3, networker, getTime, empAddress, medianizerPriceFeedConfig)
+      createTokenPriceFeedForEmp(logger, web3, networker, getTime, empAddress, {
```
nicholaspai, line 101: Here's some new logic: I enforce that all of the pricefeeds use the same precision.

packages/monitors/src/SyntheticPegMonitor.js
```diff
@@ -3,13 +3,16 @@
 
 const { ConvertDecimals, createFormatFunction, formatHours, createObjectFromDefaultProps } = require("@uma/common");
 
+// TODO: Rename "medianizerPriceFeed" ==> "pegPriceFeed" and "uniswapPriceFeed" ==> "syntheticPriceFeed"
```
mrice32, line 6: +1

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -70,6 +70,7 @@ const defaultConfigs = {
   STABLESPREAD: {
     // This is alternatively known as "STABLESPREAD/ETH"
     type: "basketspread",
+    lookback: 7200,
```
nicholaspai, line 73: These EMP's all have huge lookback's so I want to override them

mrice32, line 73: +1

mrice32 (approved):
LGTM!

PR #2387, opened 2021-01-08
improve(affiliates): add json file to show current state of dev mining

### LINKED ISSUE #2386: Add support for getting active participants in Dev Mining from API

### Why
Yam finance (and probably other third party devleopers) will need a way to estimate APY on their frontends for liquidity mining. Currently this requires knowing all the EMPs participating in dev mining, to correctly attribute the right proportion of UMA rewards to their EMP. 

This information is currently private ( in notion) and should be exposed to the public in a way that can be queried from a frontend. 

### How
This should be as simple as adding an dev mining config for the "current weeks rewards" in github and updating it every week as rewards are paid out.  This is a good enough solution for now and a more accurate solution will require more engineering time.

daywiss:
**Motivation**

#2386 

**Summary**

- renamed payouts/deployers to payouts/devmining for consistency with campaign name
- Add a small json file that can be referenced to calculate dev mining rewards. It contains active EMP participants and the total rewards for the week.  It is assumed that the program will run weekly, but this information could eventually be added to the file as well if needed.
- Small readme is also added to the payouts folder.

**Issue(s)**

Fixes #2386


packages/affiliates/payouts/devmining-status.json
```diff
@@ -0,0 +1,11 @@
+{
+  "empWhitelist": [
+    "0x3a93E863cb3adc5910E6cea4d51f132E8666654F",
+    "0x516f595978D87B67401DaB7AfD8555c3d28a3Af4",
+    "0x4AA79c00240a2094Ff3fa6CF7c67f521f32D84a2",
+    "0xf32219331A03D99C98Adf96D43cc312353003531",
+    "0x1c3f1A342c8D9591D9759220d114C685FD1cF6b8",
+    "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5"
```
smb2796, line 8: Since we expect to support these three ZeldaCash contracts, think we should just add them now so that people can get more accurate rewards predictions. Wdyt?

```suggestion
    "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5",
    "0xeAddB6AD65dcA45aC3bB32f88324897270DA0387",
    "0xf215778F3a5e7Ab6A832e71d87267Dd9a9aB0037",
    "0x267D46e71764ABaa5a0dD45260f95D9c8d5b8195"
```

smb2796 (approved):
Looking good - really love having this so other devs can get this info easily.

Left a suggestion to preemptively add the newly supported contracts for this upcoming week.

mrice32 (approved):

packages/affiliates/payouts/devmining-status.json
daywiss: yea sounds good

PR #2389, opened 2021-01-08
docs(readmes): add readmes for monitors, reporters, and serverless-orchestration packages

pemulis:
**Motivation**

We want to have slim READMEs for all of our packages that provide some basic information and direct the reader to the main docs site for more in-depth instructions to set up and run the bots.

**Summary**

Adds brief READMEs for three packages that don't currently have them.

**Issue(s)**

N/A


mrice32 (approved):
LGTM! I agree that it's probably not necessary at this point to give run instructions for these packages since they are less essential for others to run and some are explained on the docs site.

PR #2390, opened 2021-01-08
improve(price-feed-interface): Disputer should not throw warning if trying to lookup liquidation time beyond price feed lookback

nicholaspai:
**Motivation**

- Adds `earliestTimestamp` to PriceFeedInterface which is the earliest possible timestamp for which a historical price could be available
- Use `pricefeed.earliestTimestamp` in `disputer` and don't throw warning if trying to fetch price for liquidation time earlier than this `earliestTimestamp`. This is possible if the user explicitly sets a shorter `lookback` than the EMP's `liquidationLiveness` window. e.g. This avoids sending out warnings if a liquidation occurred earlier than the pricefeed's lookback window and its liveness has not expired.

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
@@ -89,6 +89,10 @@ class BalancerPriceFeed extends PriceFeedInterface {
     return this.lastUpdateTime;
   }
 
+  getEarliestTime() {
+    return this.earliestTime;
+  }
+
   getCurrentPrice() {
     let currentPrice;
     // If twap window is 0, then return last price
@@ -143,6 +147,7 @@ class BalancerPriceFeed extends PriceFeedInterface {
     await Promise.all(blocks.map(this.priceHistory.update));
 
     this.lastUpdateTime = currentTime;
+    this.earliestTime = this.blockHistory.earliest().timestamp;
   }
   // If priceHistory only encompasses 1 block, which happens if the `lookback` window is 0,
   // then this should return the last and only price.
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -145,6 +145,10 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       });
   }
 
+  getEarliestTime() {
+    return this.earliestTime;
+  }
+
   getLastUpdateTime() {
     return this.lastUpdateTime;
   }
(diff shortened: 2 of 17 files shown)


packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -117,6 +121,7 @@ class UniswapPriceFeed extends PriceFeedInterface {
     this.currentTwap = this._computeTwap(this.events, currentTime - this.twapLength, currentTime);
 
     this.lastUpdateTime = currentTime;
+    this.earliestTime = this.events[0].timestamp;
```
mrice32, line 124: Why not `currentTime - this.historicalLookback`? Also, should we compute this on-demand in the getEarliestTime function rather than using a state variable?

I only ask because this would separate an _expected_ lack of data due to insufficient time range from the case where data is missing because of some event/querying issue.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -145,6 +145,10 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       });
   }
 
+  getEarliestTime() {
+    return this.earliestTime;
```
mrice32, line 149: Similar to below, should this be the _expected_ earliest time based on the price feed construction? Or should it be the earliest time we have data for (which could be now if historical data is lacking or unavailable because of a feed error)?

packages/financial-templates-lib/src/price-feed/BalancerPriceFeed.js
```diff
@@ -89,6 +89,10 @@ class BalancerPriceFeed extends PriceFeedInterface {
     return this.lastUpdateTime;
   }
 
+  getEarliestTime() {
```
mrice32, line 92: Same as below.

mrice32 (commented):
Looks great!  One broader question about what information we want from the getEarliestTime method.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: We can also use `currentTime - this.historicalLookBack` but I thought getting the last block in the block history's timestamp would be an improvement.

> Also, should we compute this on-demand in the getEarliestTime function rather than using a state variable?

This is a fair point, I'll make that change.

nicholaspai: Actually thinking about this more, I feel like `currentTime - this.historicalLookback` is more reasonable to the user and hides implementation details. It can also be consistent across price feeds.

nicholaspai:
Closing in favor of #2391

PR #2391, opened 2021-01-09
improve(price-feed-interface): Add getLookback to interface

nicholaspai:
**Motivation**

`Disputer` should not throw warning if trying to lookup liquidation time beyond price feed lookback (`lastUpdateTimestamp` - `lookback`):

- Adds `getLookback` to PriceFeedInterface which is the earliest possible timestamp for which a historical price could be available
- Usethis information in `disputer` and `ContractMonitor` and don't throw warning if trying to fetch price for liquidation time earlier than historical lookback. This is possible if the user explicitly sets a shorter lookback than the EMP's liquidationLiveness window. e.g. This avoids sending out warnings if a liquidation occurred earlier than the pricefeed's lookback window and its liveness has not expired.


mrice32 (approved):
LGTM! Agree this is much simpler!!

PR #2341, opened 2020-12-23
feat(serverless): Updated spoke logging within GCP

chrismaree:
Signed-off-by: Christopher Maree <christopher.maree@gmail.com>
image

image

**Motivation**

Explain the motivation for making this change.
What existing problem does the pull request solve?
It may be sufficient to just link to an issue, but please explain if more details are needed.

**Summary**

Briefly summarize what changes were made to accomplish the motivation above.

**Details**

This may be unnecessary for some PRs. Catch-all for detailed explanations about the implementation decisions and implications of the change.

**Issue(s)**

Fixes #XXXX

PR #2393, opened 2021-01-11
feat(serverless): add timeout delay to serverless spoke calls

### LINKED ISSUE #2392: Add spoke timeout support to serverless hub

chrismaree:
**Motivation**

issue #2392

**Summary**

This PR places an upper bound on how long a spoke can take to reply to the hub. If it takes longer than a defined period (defaults to 1 min) the hub defines the call as rejected and logs accordingly.

**Details**

Before the PR the hub could timeout if a spoke did not respond. The hub correctly dealt with rejected calls, including re-try logic, but it had no logic built-in for timeout on spokes.

This PR uses `Promise.race` in conjunction with each spoke call and a new method called `_rejectAfterDelay`. This is placed within each spoke call and limits how long each spoke call can take. If a spoke call takes longer than `hubConfig.rejectSpokeDelay` then the hub will log the rejected call.

Note that if a spoke times out the re-try logic is still used. This means a timeout is re-tried.

Looking at past loops within our serverless infrastructure, the hub normally hears back from ALL spokes and resolves all serverless calls within 30 seconds from the initialization of the loop. From this information, I have set the default timeout to 60 seconds per spoke. This means that if a spoke times out it should be caught and re-tried after 60 seconds.

This places an upper bound on the hub execution at 2 minutes (assuming a timeout, retry and timeout again). As the serverless hub runs on a 5 min cron job, we know that by execution +2 mins we will see valid (or rejected) execution of ALL spokes, based off the timeout.

**Issue(s)**

close #2392


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -40,7 +40,14 @@ const { Logger, delay } = require("@uma/financial-templates-lib");
 let logger;
 let spokeUrl;
 let customNodeUrl;
-let hubConfig = { configRetrieval: "localStorage", saveQueriedBlock: "localStorage", spokeRunner: "localStorage" };
+let hubConfig = {};
+
+const defaultHubConfig = {
```
chrismaree, line 45: this small change was made to make it cleaner when overriding configs, especially between test itterations.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -73,13 +81,17 @@ hub.post("/", async (req, res) => {
...
+    // that each promise is a race between the serverlessSpoke command and a `_rejectAfterDelay`. This places an upper
+    // bound on how long each spoke can take to respond, acting as a timeout for each spoke call.
     let promiseArray = [];
     let botConfigs = {};
     for (const botName in configObject) {
       const botConfig = _appendEnvVars(configObject[botName], botName, lastQueriedBlockNumber, latestBlockNumber);
       botConfigs[botName] = botConfig;
-      promiseArray.push(_executeServerlessSpoke(spokeUrl, botConfig));
+      promiseArray.push(
+        Promise.race([_executeServerlessSpoke(spokeUrl, botConfig), _rejectAfterDelay(hubConfig.rejectSpokeDelay)])
```
chrismaree, line 93: this is the key change: each spoke call is a race between the spoke execution and a rejecting delay. whichever ends first wins. i.e if the reject ends first, the call timed out.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -124,7 +136,12 @@ hub.post("/", async (req, res) => {
       });
       let rejectedRetryPromiseArray = [];
       retriedOutputs.forEach(botName => {
-        rejectedRetryPromiseArray.push(_executeServerlessSpoke(spokeUrl, botConfigs[botName]));
+        rejectedRetryPromiseArray.push(
+          Promise.race([
```
chrismaree, line 140: the same behavior is added on re-try logic.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -350,7 +367,13 @@ async function _postJson(url, body) {
 // Takes in a spokeResponse object for a given botKey and identifies if the response includes an error. If it does,
 // append the error information to the errorOutputs. If there is no error, append to validOutputs.
 function _processSpokeResponse(botKey, spokeResponse, validOutputs, errorOutputs) {
-  if (
+  if (spokeResponse.status == "rejected" && spokeResponse.reason.status == "timeout") {
```
chrismaree, line 370: small bit of log processing if the execution timed out.

packages/serverless-orchestration/test-helpers/TimeoutSpokeMock.js
```diff
@@ -0,0 +1,21 @@
+const express = require("express");
```
chrismaree, line 1: mock spoke that can take ` _responseDelay` to respond to a call.

packages/serverless-orchestration/test/ServerlessSpoke.js
```diff
@@ -62,7 +62,7 @@ contract("ServerlessSpoke.js", function(accounts) {
       transports: [new SpyTransport({ level: "debug" }, { spy: spy })]
     });
 
-    // Start the cloud run spoke instance with the spy logger injected.
```
chrismaree, line 65: renamed this because it was inconsistent.

packages/serverless-orchestration/src/ServerlessHub.js
nicholaspai: Why was this an issue before without the race method?

nicholaspai (approved):
This is a nice addition and will make sure that hub always returns a status. Previously it seemed like the hub would just hang until the next iteration. Is my understanding of the issue accurate?

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: Before this, the spokes could take unlimited time to respond. This race places an upper bound on how long a spoke can take to reply.

chrismaree:
> This is a nice addition and will make sure that hub always returns a status. Previously it seemed like the hub would just hang until the next iteration. Is my understanding of the issue accurate?

yup, that's right! this places an upper bound on how long each spoke can run for and in turn how long until the hub returns some info. This will prevent the hub from hanging.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -40,7 +40,14 @@ const { Logger, delay } = require("@uma/financial-templates-lib");
...
 let spokeUrl;
 let customNodeUrl;
-let hubConfig = { configRetrieval: "localStorage", saveQueriedBlock: "localStorage", spokeRunner: "localStorage" };
+let hubConfig = {};
+
+const defaultHubConfig = {
+  configRetrieval: "localStorage",
+  saveQueriedBlock: "localStorage",
+  spokeRunner: "localStorage",
+  rejectSpokeDelay: 60 // 1 min.
```
mrice32, line 49: Is there any reason not to bump this default to 5 minutes to be a bit more lenient? I'm thinking especially about the case when we're using our own nodes that aren't hyper-optimized like infura. Obviously, anything longer than 5 minutes feels a little problematic because it means the next bot iteration starts before the last finishes.

mrice32 (commented):
Looks great -- only one question about the length of the default timeout.

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: I chose this because a timed out spoke will be re-tried. Therefore the max hub execution time is ~2x the `rejectSpokeDelay`. Looking at past loops within our serverless infrastructure, the hub normally hears back from ALL spokes and resolves all serverless calls within 30 seconds from the initialization of the loop. From this information, I have set the default timeout to 60 seconds per spoke. This means that if a spoke times out it should be caught and re-tried after 60 seconds.


You are right that this heuristic only works within the past infura calls and does not work well with anything longer. However, I feel that at an upper bound the `rejectSpokeDelay` should be no more than 1/2 the execution delay of the hub. So the hub runs every 5 mins this value should be strictly less than 2.5 mins. The reason is that I really don't want the previous hub execution still running when the next one starts or it is very difficult to reason about the logs.

The highest upper bound I think we should go for this value is 2 mins. Processes _really_ should not take longer than this and if they are then we need to speed them up pretty drastically! 

WDYT?

nicholaspai: Agreed we can't set this too close to the POLLING_DELAY otherwise the kill signal might not be sent before the next loop starts. 2-2.5 mins seems right for an upper bound.

chrismaree:
Update: this PR was changed to include the functionality originally presented within https://github.com/UMAprotocol/protocol/pull/2341. Log messages have been changed within the spokes to be cleaner and more consumable to remove bloat. Logs now look like this:
image

For each bot. If you want to see more info on the underlying spoke logs you can search the individual bot logs.

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: Agreed @nicholaspai. I have set it to 2 mins by default, making the max hub execution time ~4mins.

PR #2394, opened 2021-01-11
feat(liquidator) update liquidator to work with perp and latest EMP contracts

### LINKED ISSUE #2382: Update liquidation bots to work with Perpetual

chrismaree:
**Motivation**

With the release of the new perpetual contracts & the updated EMP, the UMA bot infrastructure is required to be updated accordingly. This PR updates the liquidator to work with these updated contacts.

issue #2382

**Summary**

The _vast_ majority of the behavior of the perpetual and new EMP contracts is the same as previous EMPs written by UMA. This is useful as most of the interfaces, except for some events and new functions, remain the same. The main challenge with adding this functionality is considering how to _test_ and parameterized one piece of infrastructure (e.g. a liquidator bot) to work with multiple sets of smart contracts. This is achieved in the following ways:

- Bots that supports multiple versions must define a `SUPPORTED_CONTRACT_VERSIONS` which contains information on the specific contracts the bot can support. This PR adds this logic to the liquidator bot enabling it to support the following contract versions:
```
const SUPPORTED_CONTRACT_VERSIONS = [
  { contractType: "ExpiringMultiParty", contractVersion: "1.2.0" },
  { contractType: "ExpiringMultiParty", contractVersion: "1.2.1" },
  { contractType: "ExpiringMultiParty", contractVersion: "1.2.2" },
  { contractType: "ExpiringMultiParty", contractVersion: "latest" },
  { contractType: "Perpetual", contractVersion: "latest" }
];
```
- Bots are responsible for _detecting_ the kind of contract version they are talking to using a newly added utility called `findContractVersion`. This enables the user of the bot to simply define a `financialContractAddress` and the bot will identify what version it is talking to and use the associated contract code.
- The bot operator should be able to override the contract version detected by `findContractVersion` to enable testing and the use of compatible pre-release contracts.
- Bots that can run against multiple versions/contract types should be tested against _all_ contract types that they support (more details on this in the Details section).
- Places where bot logic needs to _branch_ based on different kinds of bots should be kept as close to the implementation as possible. This point aims to keep bot code as simple as possible. This is only viable as the contract changes between versions are _very_ small. This would **not** be the right pattern if we had massive changes in ABI between versions. In the future, we might choose to change this point to use a more sophisticated form of abstraction but for now, this seems to be the cleanest pattern.

**Details**

The main complexity (and diff) in this PR came from changing how we run tests to enable one set of tests to run the _same_ logic against different contract versions and types. This was achieved by a set of custom test helpers to enable the unit tests:
- run against a predefined array of `SUPPORTED_CONTRACT_VERSIONS`
- specify tests that should run on a specific version/contract type or run on `any`
- helper functions to accommodate changes in constructor ABIs, to support changes such as those made in https://github.com/UMAprotocol/protocol/pull/2332. This enables the interfaces to switch within the constructor.

With this functionality, the existing liquidator tests were extended to verify that _all_ previous functionality works with all new contract versions. Additionally, some tests were added to validate contract specific behavior, such as Perpetual specific tests that validate that the bot correctly considers funding rate in the perpetual case.

**How to review this PR**
This PR is biiig, due to the test changes. All previous tests _were not changed_. Rather, the test files were updated to accommodate this multi-version structure. Reviewers should focus primarily on the non-test changes and the _new_ tests that were added that validate specific behavior. Additionally, reviewers should give feedback on the new testing structure. I will draw attention to specific sections for feedback.

**What comes next**
After this PR the next step is to refactor these test helpers into a common utility that can then be re-used between the other packages (Monitor, Disputer). I've left a TODO to this effect. The reason I did not go directly to the common package implementation is that I felt it easier to follow & understand the changes for review with it all in one place. 

After this, the other bot packages should be updated in a similar fashion to this PR.

**Issue(s)**

close #2382


packages/common/src/Enums.js
```diff
@@ -23,10 +23,10 @@ const LiquidationStatesEnum = {
 // Note that these are status translations AFTER a withdrawLiquidation method is called
 const PostWithdrawLiquidationRewardsStatusTranslations = {
   "0": "Liquidation deleted; All rewards have been withdrawn",
-  "3": "Dispute succeeded; Not all rewards have been withdrawn"
```
chrismaree, line 26: this const Enum needs to support the new version of contracts wherein a withdrawal can be processed by any user, and state 4 is posible.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -24,7 +24,8 @@ class ExpiringMultiPartyClient {
     empAddress,
     collateralDecimals = 18,
     syntheticDecimals = 18,
-    priceFeedDecimals = 18
+    priceFeedDecimals = 18,
+    contractType
```
chrismaree, line 28: the client now uses the `contractType` to make small changes to the internal logic.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -138,6 +154,14 @@ class ExpiringMultiPartyClient {
       this.emp.methods.getCurrentTime().call()
     ]);
 
+    if (this.contractType == "Perpetual") {
```
chrismaree, line 157: Set to the current `latestCumulativeFundingRateMultiplier` else default to zero. this makes other computations work as per usual with no changes needed in the non-perpetual case.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -24,7 +24,8 @@ class ExpiringMultiPartyClient {
     empAddress,
     collateralDecimals = 18,
     syntheticDecimals = 18,
-    priceFeedDecimals = 18
+    priceFeedDecimals = 18,
+    contractType = "ExpiringMultiParty" // Default to EMP for now to enable backwards compatibility with other bots. This will be removed as soon as the other bots have been updated to work with these contract types.
```
chrismaree, line 28: this ensures that these changes will not break this client within the other bots that are yet to be changed.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -219,10 +247,13 @@ class ExpiringMultiPartyClient {
         sponsor: this.activeSponsors[index],
         withdrawalRequestPassTimestamp: position.withdrawalRequestPassTimestamp,
         withdrawalRequestAmount: position.withdrawalRequestAmount.toString(),
-        numTokens: position.tokensOutstanding.toString(),
-        amountCollateral: this.toBN(position.rawCollateral.toString())
+        numTokens: this.toBN(position.tokensOutstanding.toString()) // Apply the current funding rate to the sponsor debt.
```
chrismaree, line 250: this is the only change needed for the client: the `numTokens` (sponsor outstanding token debt) is scaled by the current `CumulativeFundingRateMultiplier`

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,944 @@
...
+  await emp.proposeFundingRate({ rawValue: fundingRate }, currentTime);
+  await emp.setCurrentTime(currentTime + 10000);
+};
+
+// If the current version being executed is part of the `supportedVersions` array then return `it` to run the test.
+// Else, do nothing. Can be used exactly in place of a normal `it` to parameterize contract types and versions supported
+// for a given test.eg: versionedIt(["Perpetual-latest"])("test name", async function () { assert.isTrue(true) })
+// Note that a second param can be provided to make the test an `it.only` thereby ONLY running that single test, on
+// the provided version. This is very useful for debugging and writing single unit tests without having ro run all tests.
+const versionedIt = function(supportedVersions, shouldBeItOnly = false) {
```
chrismaree, line 75: this function is meant to replace the mocha `it` through the tests by providing a means to define a set of "contract versions" that is supported.

chrismaree: in the next PR I will pull this to a common helper util.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,944 @@
...
+    throw new Error(
+      `Contract versioned specified ${supportedVersions} is not part of the supported contracts for this test suit`
+    );
+  }
+  // Return if the `currentTestIterationVersion` is part of the supported versions includes any. Returning true
+  // means that test will be run. Else, if returned false, the test will be skipped.
+  return supportedVersions.includes(currentTestIterationVersion) || supportedVersions.includes("any");
+};
+
+const _createConstructorParamsForContractVersion = async function(contractVersion, contractType) {
```
chrismaree, line 92: Different versions require different constructors. this is useful in switching based on the version in question. Will also be pulled to common util.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,944 @@
...
+            financialContractInstance.abi,
+            web3,
+            emp.address,
+            testConfig.collateralDecimals,
+            testConfig.syntheticDecimals,
+            testConfig.priceFeedDecimals,
+            currentTypeTested // either ExpiringMultiParty OR Perpetual depending on the test
+          );
+        });
+        versionedIt(["any"])("Returns all positions", async function() {
```
chrismaree, line 258: here you can see how the new test structure is used. This test is meant to be run on `any` contract version (meaning all contract versions specified should be tested).

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,944 @@
...
+              testConfig.syntheticDecimals,
+              testConfig.priceFeedDecimals,
+              null // some contract name that does not exist
+            );
+          } catch (error) {
+            didThrow = true;
+          }
+          assert.isTrue(didThrow);
+        });
+        versionedIt(["Perpetual-latest"])(
```
chrismaree, line 819: here you can see a Perpetual specific unit test.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: note that the second param here is optional. if included it will make the test exclusive (`it.only`). This could be dangerous syntax as it will exclude other tests within a test file but won't be caught by our linter. Am curious what others think about this.

packages/liquidator/index.js
```diff
@@ -53,24 +61,52 @@ async function run({
...
+          liquidatorConfig.contractVersion
+        } & contractType:${liquidatorConfig.contractType} is not part of ${JSON.stringify(SUPPORTED_CONTRACT_VERSIONS)}`
+      );
 
-    // Setup contract instances.
-    const voting = new web3.eth.Contract(getAbi("Voting"), getAddress("Voting", networkId));
-    const emp = new web3.eth.Contract(getAbi("ExpiringMultiParty"), empAddress);
+    // Setup contract instances. This uses the contract version pulled in from previous step. Voting is hardcoded to latest main net version.
+    const voting = new web3.eth.Contract(getAbi("Voting", "1.2.2"), getAddress("Voting", networkId));
+    const emp = new web3.eth.Contract(
```
chrismaree, line 89: here we pull in the right contract type and version

packages/liquidator/index.js
```diff
@@ -53,24 +61,52 @@ async function run({
...
+      getAbi(liquidatorConfig.contractType, liquidatorConfig.contractVersion),
+      empAddress
+    );
 
     // Returns whether the EMP has expired yet
-    const checkIsExpiredPromise = async () => {
-      const [expirationTimestamp, contractTimestamp] = await Promise.all([
-        emp.methods.expirationTimestamp().call(),
+    const checkIsExpiredOrShutdownPromise = async () => {
```
chrismaree, line 95: this method needs to accommodate expired or shutdown contracts.

packages/liquidator/index.js
```diff
@@ -86,21 +122,17 @@ async function run({
       minSponsorTokens,
       collateralTokenAddress,
       syntheticTokenAddress,
-      isExpired,
```
chrismaree, line 89: I removed this call to simply the logic a bit. this info is only _really_ needed within the main while loop and it is queried there anyways.

packages/liquidator/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "UMA Liquidator",
   "dependencies": {
     "@uma/common": "^1.1.0",
-    "@uma/core": "npm:@uma/core@1.2.2",
+    "@uma/core": "1.2.2",
```
chrismaree, line 7: liquidator now uses "latest" and pulls the associate ABI versions when needed.

packages/liquidator/src/liquidator.js
```diff
@@ -239,7 +253,7 @@ class Liquidator {
         message: "Detected a liquidatable position",
         scaledPrice: scaledPrice.toString(),
         maxCollateralPerToken: maxCollateralPerToken.toString(),
-        position: JSON.stringify(position)
```
chrismaree, line 242: these `JSON.strigify` parts were removed as the logs look way better as structured objects spread out by the transport.

packages/liquidator/src/liquidator.js
```diff
@@ -407,22 +421,22 @@ class Liquidator {
...
         message: "Detected a pending or expired liquidation",
-        liquidation: JSON.stringify(liquidation)
+        liquidation: liquidation
       });
 
       // Construct transaction.
       const withdraw = this.empContract.methods.withdrawLiquidation(liquidation.id, liquidation.sponsor);
 
       // Confirm that liquidation has eligible rewards to be withdrawn.
-      let withdrawAmount, gasEstimation;
```
chrismaree, line 417: between version 1.2.2 and the latest, we changed how withdrawals work. now, this _either_ returns the amount withdrawn _or_ returns an object containing extra info (such as other positions withdrawn).

packages/liquidator/test/Liquidator.js
```diff
@@ -42,640 +33,295 @@ const configs = [
   { tokenSymbol: "BTC", collateralDecimals: 8, syntheticDecimals: 18, priceFeedDecimals: 8 },
   { tokenSymbol: "BTC", collateralDecimals: 8, syntheticDecimals: 8, priceFeedDecimals: 18 }
 ];
-// allows this to be set to null without throwing.
```
chrismaree, line 45: the test changes made here are very similar to the other two sets. See the new tests at the bottom that test the perp specific logic.

packages/serverless-orchestration/test/ServerlessHub.js
```diff
@@ -324,7 +324,8 @@ contract("ServerlessHub.js", function(accounts) {
           CUSTOM_NODE_URL: web3.currentProvider.host,
           POLLING_DELAY: 0,
           EMP_ADDRESS: emp.address,
-          PRICE_FEED_CONFIG: defaultUniswapPricefeedConfig
+          PRICE_FEED_CONFIG: defaultUniswapPricefeedConfig,
+          LIQUIDATOR_CONFIG: { contractVersion: "latest", contractType: "ExpiringMultiParty" }
```
chrismaree, line 328: added here to make the serverless compatible with the new changes. note this is only needed within these unit tests because the contract version deployed here is slightly different due to package versioning. This simply acts to override the default version.

packages/liquidator/test/index.js
```diff
@@ -32,403 +106,440 @@ contract("index.js", function(accounts) {
...
-          lockedCollateral: toWei("130"),
-          liquidationTime: liquidationTime.toString(),
-          liquidator: ZERO_ADDRESS,
-          disputer
+        // To verify decimal detection is correct for a standard feed, check the third log to see it matches expected.
+        assert.isTrue(spyLogIncludes(spy, 3, '"collateralDecimals":18'));
+        assert.isTrue(spyLogIncludes(spy, 3, '"syntheticDecimals":18'));
+        assert.isTrue(spyLogIncludes(spy, 3, '"priceFeedDecimals":18'));
+      });
+      it("Correctly detects contract type and rejects unknown contract types", async function() {
```
chrismaree, line 426: validates the version detection behavour

packages/common/src/Enums.js
nicholaspai: After anyone calls `withdrawLiquidation()`, won't the state go to 0? So I think actually only state 0 is possible post-withdrawal?

packages/common/src/FindContractVersion.js
```diff
@@ -0,0 +1,48 @@
...
+ */
+async function findContractVersion(contractAddress, web3) {
+  assert(web3, "Web3 object must be provided");
+  assert(contractAddress, "Contract Address Must be provided");
+  const contractCode = await web3.eth.getCode(contractAddress);
+  const contractCodeHash = web3.utils.soliditySha3(contractCode);
+  return versionMap[contractCodeHash] || {};
+}
+
+const versionMap = {
```
nicholaspai, line 18: This seems like the cleanest way to detect version, +1. Another idea would have been detecting presence of certain methods that exist in one version but not another, but that seems overly complex.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
nicholaspai: That's a reasonable default

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -138,6 +158,14 @@ class ExpiringMultiPartyClient {
       this.emp.methods.getCurrentTime().call()
     ]);
 
+    if (this.contractType === "Perpetual") {
+      this.latestCumulativeFundingRateMultiplier = this.toBN(
+        (await this.emp.methods.fundingRate().call()).cumulativeMultiplier.rawValue
+      );
+    } else {
+      this.latestCumulativeFundingRateMultiplier = this.fixedPointAdjustment;
```
nicholaspai, line 166: If `latestCumulativeFundingRateMultiplier` defaults to `fixedPointAdjustment` perhaps this class var is better described as `numTokenMultiplier`. Otherwise setting `latestCumulativeFundingRateMultiplier = fixedPointAdjustment` is a bit weird.

packages/liquidator/index.js
```diff
@@ -21,6 +21,14 @@ const {
 const { getAbi, getAddress } = require("@uma/core");
 const { getWeb3 } = require("@uma/common");
 
+const SUPPORTED_CONTRACT_VERSIONS = [
```
nicholaspai, line 24: Do you think that other classes or even dApps would re-use some of these contract versions? Perhaps we should put the list of all possible contract versions in a `common` file and allow each deriving class to specify what versions it supports.

packages/liquidator/index.js
```diff
@@ -53,24 +61,52 @@ async function run({
...
     // Load unlocked web3 accounts and get the networkId.
-    const [accounts, networkId] = await Promise.all([web3.eth.getAccounts(), web3.eth.net.getId()]);
+    const [detectedContract, accounts, networkId] = await Promise.all([
+      findContractVersion(empAddress, web3),
+      web3.eth.getAccounts(),
+      web3.eth.net.getId()
+    ]);
+
+    // Append the contract version and type to the liquidatorConfig, if the liquidatorConfig does not already contain one.
+    if (!liquidatorConfig) liquidatorConfig = {};
```
nicholaspai, line 71: What if the passed in `liquidatorConfig.contractVersion` and type do not match the `detectedContract`'s version and type?

packages/liquidator/index.js
nicholaspai: Do you forsee any issues where voting version and EMP version are different? I think that's not a problem but its  just weird to see the version hardcoded here for `voting` but not for `EMP`

packages/liquidator/index.js
```diff
@@ -53,24 +61,52 @@ async function run({
...
+    const checkIsExpiredOrShutdownPromise = async () => {
+      const [expirationOrShutdownTimestamp, contractTimestamp] = await Promise.all([
+        liquidatorConfig.contractType == "ExpiringMultiParty"
+          ? emp.methods.expirationTimestamp().call()
+          : emp.methods.emergencyShutdownTimestamp().call(),
         emp.methods.getCurrentTime().call()
       ]);
       // Check if EMP is expired.
-      if (Number(contractTimestamp) >= Number(expirationTimestamp)) {
+      if (Number(contractTimestamp) >= Number(expirationOrShutdownTimestamp) && expirationOrShutdownTimestamp > 0) {
```
nicholaspai, line 103: Should the RHS of the `&&` be `Number(expirationOrShutdownTimestamp)` since we do that on the LHS?

packages/liquidator/index.js
```diff
@@ -53,24 +61,52 @@ async function run({
...
         logger.info({
           at: "Liquidator#index",
-          message: "EMP is expired, can only withdraw liquidator dispute rewards 🕰",
-          expirationTimestamp,
+          message: `EMP is ${
+            liquidatorConfig.contractType == "ExpiringMultiParty" ? "expired" : "shutdown"
```
nicholaspai, line 107: I notice you're using double "=="'s throughout this PR, but I didn't want to comment on each one. For consistency with rest of repo can you switch to triple "==="

packages/liquidator/index.js
```diff
@@ -190,49 +223,47 @@ async function run({
...
       priceFeedDecimals: Number(priceFeed.getPriceFeedDecimals()),
-      priceFeedConfig
+      priceFeedConfig,
+      liquidatorConfig
     });
 
     // The EMP requires approval to transfer the liquidator's collateral and synthetic tokens in order to liquidate
     // a position. We'll set this once to the max value and top up whenever the bot's allowance drops below MAX_INT / 2.
-    // If the contract is expired, the liquidator can only withdraw disputes so there is no need to set allowances.
-    if (!IS_EXPIRED) {
```
nicholaspai, line 199: This isn't a problem IMO to remove the `IS_EXPIRED` check, as it prevents sending allowance transactions in only a strange edge case: where the bot is running on an expired EMP that it has not approved yet.

packages/liquidator/index.js
```diff
@@ -305,14 +336,16 @@ async function Poll(callback) {
...
       //   "liquidationMinPrice":0, -> Aborts if the amount of collateral in the position per token is below this ratio
       //   "txnGasLimit":9000000 -> Gas limit to set for sending on-chain transactions.
       //   "whaleDefenseFundWei": undefined -> Amount of tokens to set aside for withdraw delay defense in case position cant be liquidated.
       //   "defenseActivationPercent": undefined -> How far along a withdraw must be in % before defense strategy kicks in.
-      //   "logOverrides":{"positionLiquidated":"warn"}} -> override specific events log levels.
-      liquidatorConfig: process.env.LIQUIDATOR_CONFIG ? JSON.parse(process.env.LIQUIDATOR_CONFIG) : null,
+      //   "logOverrides":{"positionLiquidated":"warn"}, -> override specific events log levels.
+      //   "contractType":"ExpiringMultiParty", -> override the kind of contract the liquidator is pointing at.
+      //   "contractVersion":"1.2.2"":"ExpiringMultiParty"} -> override the contract version the liquidator is pointing at.
+      liquidatorConfig: process.env.LIQUIDATOR_CONFIG ? JSON.parse(process.env.LIQUIDATOR_CONFIG) : {},
```
nicholaspai, line 348: Doesn't this switching of the default value from `null` to `{}` remove the need for [this](https://github.com/UMAprotocol/protocol/pull/2394/files#diff-2eaa6768c5ce2a0129aa4b22dfb8f7f33c0cedd7bbb4cb7120ee3327ced9c4ebR71) change?

packages/liquidator/src/liquidator.js
```diff
@@ -134,6 +134,18 @@ class Liquidator {
...
+      contractType: {
+        value: undefined,
+        isValid: x => {
+          return x == "ExpiringMultiParty" || x == "Perpetual";
+        }
+      },
+      contractVersion: {
+        value: undefined,
+        isValid: x => {
+          return x == "1.2.0" || "1.2.1" || "1.2.2" || x == "latest";
```
nicholaspai, line 147: I feel like these hardcoded version values that are shared between this file and `index/liquidator.js` should be refactored into a shared dictionary.

packages/liquidator/src/liquidator.js
```diff
@@ -407,22 +421,22 @@ class Liquidator {
...
 
       // Construct transaction.
       const withdraw = this.empContract.methods.withdrawLiquidation(liquidation.id, liquidation.sponsor);
 
       // Confirm that liquidation has eligible rewards to be withdrawn.
-      let withdrawAmount, gasEstimation;
+      let withdrawalCallResponse, gasEstimation;
       try {
-        [withdrawAmount, gasEstimation] = await Promise.all([
+        [withdrawalCallResponse, gasEstimation] = await Promise.all([
```
nicholaspai, line 433: +1 better name

packages/liquidator/src/liquidator.js
```diff
@@ -440,11 +454,16 @@ class Liquidator {
         gas: Math.min(Math.floor(gasEstimation * this.GAS_LIMIT_BUFFER), this.txnGasLimit),
         gasPrice: this.gasEstimator.getCurrentFastPrice()
       };
+      // In contract version 1.2.2 and below this function returns one value: the amount withdrawn by the function caller.
```
nicholaspai, line 457: Since versions are linearly increasing, and some features don't change "greater than" or "less than" certain versions, is there an easy way to replicate the logic `if (this.contractVersion <= 1.2.2)`...?

nicholaspai (commented):
I've reviewed all non-test files, will get to those after but wanted to get some initial comments out. This was a TON of code, great start!

packages/liquidator/index.js
chrismaree: Well, the whole point of this change is that we _know_ the canonical deployed version of the voting contract (1.2.2.) and we can update this hardcoded version when we roll out a new release for voting. On the other hand, the EMP could be any version.

Perhaps we should have a common file that indicates the current production versioning of `Voting` that can then be pulled in but for now, I think this is easy enough.

packages/common/src/FindContractVersion.js
chrismaree: ye, I messed around with that & it worked relatively well. However, I felt that this would be more consistent as there is not always going to be a 1:1 mapping between some methods being present and a contract version. for example, how do you tell the difference between 1.2.1 and 1.2.2 if they have the same methods?

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
chrismaree: well `fixedPointAdjustment` simply is `1e18`. So this means that it sets the `cumlativeFundingRateMultipler` to 1 if the contract type is EMP.

`numTokenMultiplier` is a good name but I feel that it's better to keep this really explicit. What it _really_ is the cumulative funding rate multiplier and so it makes it really explicit by keeping the var named accordingly. 

WDYT?

packages/liquidator/index.js
chrismaree: yup! good idea. We can pull these out in a separate PR I think. My next step is to try and generalize a fair amount of the specific changes done in this PR.

packages/liquidator/index.js
chrismaree: then it defaults to the passed in version. that's the preferred version.

packages/liquidator/index.js
chrismaree: ye, good point. done

packages/liquidator/index.js
chrismaree: I have updated this. We should add a linter task for this!

chrismaree: I added a task to track this when we get around to it: https://github.com/UMAprotocol/protocol/issues/2447

packages/liquidator/index.js
chrismaree: ye, that's why I removed it. it was unnecessary logic that would really never apply in reality.

packages/liquidator/index.js
chrismaree: you could still put null as the process variable. this change might be redundant though seeing it is cast on the other side. will test it a bit and get back to you.

packages/liquidator/src/liquidator.js
chrismaree: agreed. also just realized this has an error in it. I have corrected it. I want to leave this for a bit later though.

packages/liquidator/src/liquidator.js
chrismaree: yup, we could build something like that out. would need to create an override for the comparison operator or a function to enable us to do this. For now, though explicit version matching is easiest I think.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: Do you use the `shouldBeItOnly` on any production tests? or is this only for debugging

packages/liquidator/index.js
nicholaspai: >Perhaps we should have a common file that indicates the current production versioning of Voting that can then be pulled in but for now, I think this is easy enough.

I think having some common dictionary listing all contract versions (or in the Voting's case, the one canonical version) would be useful. I can imagine that the voting dApp and other bots would read from this.

packages/common/src/FindContractVersion.js
nicholaspai: Yeah you'd have to test logic of the functions as well which sounds like too much code

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
nicholaspai: I'm fine leaving as it is to reduce churn

packages/liquidator/index.js
nicholaspai: Ok sgtm to leave to the next PR but I do think that these hardcoded dicts should be refactored to another file

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: only for debugging. it acts as `it.only`

packages/common/src/FindContractVersion.js
```diff
@@ -0,0 +1,49 @@
...
+ * @param {Object} web3 instance
+ * @param {string} contractAddress address of the contract in question
+ * @return {Object} contract name & version
+ */
+async function findContractVersion(contractAddress, web3) {
+  assert(web3, "Web3 object must be provided");
+  assert(contractAddress, "Contract address must be provided");
+  const contractCode = await web3.eth.getCode(contractAddress);
+  const contractCodeHash = web3.utils.soliditySha3(contractCode);
+  console.log("contractCodeHash", contractCodeHash);
```
nicholaspai, line 15: leftover console.log

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: `versionedIt` and `runTestForContractVersion` seem like they'd be useful for other clients. do you plan on pulling them into a shared file eventually?

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,943 @@
...
+        rewardRatePerSecond: { rawValue: "0" },
+        proposerBondPercentage: { rawValue: "0" },
+        maxFundingRate: { rawValue: toWei("0.00001") },
+        minFundingRate: { rawValue: toWei("-0.00001") },
+        proposalTimePastLimit: 0
+      },
+      timer.address
+    );
+
+    fundingRateIdentifier = web3.utils.utf8ToHex("TEST_FUNDiNG_IDENTIFIER");
```
nicholaspai, line 130:
```suggestion
    fundingRateIdentifier = web3.utils.utf8ToHex("TEST_FUNDING_IDENTIFIER");
```

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: This is a pretty cool feature! Does this mean that the test suite takes 5x as long to run now?

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
nicholaspai: Also, does this mean that the linter will no longer catch any `it.only()` tests leftover from testing?

nicholaspai (commented):
Responded to more comments! Thanks for reviewing the bulk of them

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: yup. will pull them into a shared util in the next PR.

and you right; the linter will no longer catch `it.only`s. if you were to put `versionedIt(["any"], true)` it be equivalent to `it.only` but the linter will miss this. I will look into a custom linter rule to catch this.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: it takes 3x as long as right now we running on three versions. I'm working (in a separate PR) on speeding up CI so hopefully, it's not too bad to the overall run time!

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: good catch. have fixed this in other places as well.

daywiss:
will try to get through your updates today, taking me a while!

daywiss (commented):
@chrismaree really cool PR and some good ideas in here. I will need to go over rest tomorrow.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
chrismaree: for sure. I will do that kind of validation within the util when I refactor this out.

packages/financial-templates-lib/src/clients/ExpiringMultiPartyClient.js
```diff
@@ -138,6 +158,14 @@ class ExpiringMultiPartyClient {
       this.emp.methods.getCurrentTime().call()
     ]);
 
+    if (this.contractType === "Perpetual") {
```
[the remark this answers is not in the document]
chrismaree: ye, this is why I was so pro the branching logic :). the only real difference is that the perp has a funding rate that modified outstanding token debt.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,943 @@
...
+};
+
+// If the current version being executed is part of the `supportedVersions` array then return `it` to run the test.
+// Else, do nothing. Can be used exactly in place of a normal `it` to parameterize contract types and versions supported
+// for a given test.eg: versionedIt(["Perpetual-latest"])("test name", async function () { assert.isTrue(true) })
+// Note that a second param can be provided to make the test an `it.only` thereby ONLY running that single test, on
+// the provided version. This is very useful for debugging and writing single unit tests without having ro run all tests.
+const versionedIt = function(supportedVersions, shouldBeItOnly = false) {
+  if (shouldBeItOnly) return runTestForContractVersion(supportedVersions) ? it.only : () => {};
+  return runTestForContractVersion(supportedVersions) ? it : () => {};
```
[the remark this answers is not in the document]
chrismaree: if the `runTestForContractVersion` returns false then you want it to do nothing (empty function). I originally put `it.skip` here but then it shows this as "pending" in the test results, which is not exactly what we want.

packages/financial-templates-lib/test/clients/ExpiringMultiPartyClient.js
```diff
@@ -1,621 +1,943 @@
...
-        await emp.create(
-          { rawValue: convertCollateral("10") },
-          { rawValue: convertSynthetic("50") },
-          { from: sponsor1 }
-        );
-        await updateAndVerify(
-          client,
-          [sponsor1],
-          [
+contract("ExpiringMultiPartyClient.js", function(accounts) {
```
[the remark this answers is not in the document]
chrismaree: some are new but most are updates of previous tests. you can see the ones that are new by looking explicitly for the `Perpetual-latest` test below.

packages/liquidator/test/Liquidator.js
```diff
@@ -687,368 +329,483 @@ contract("Liquidator.js", function(accounts) {
...
             { from: sponsor2 }
           );
 
-          // liquidatorBot creates a position with enough tokens to liquidate all positions.
+          // liquidatorBot creates a position to have synthetic tokens to pay off debt upon liquidation.
           await emp.create(
-            { rawValue: convertCollateral("10000") },
-            { rawValue: convertSynthetic("10") },
+            { rawValue: convertCollateral("1000") },
+            { rawValue: convertSynthetic("500") },
```
daywiss, line 451: +1 better numbers and cleaner math

daywiss (approved):

nicholaspai (approved):
LGTM assuming you've responded to david's comments!

packages/common/src/FindContractVersion.js
```diff
@@ -0,0 +1,48 @@
...
+ * @param {Object} web3 instance
+ * @param {string} contractAddress address of the contract in question
+ * @return {Object} contract name & version
+ */
+async function findContractVersion(contractAddress, web3) {
+  assert(web3, "Web3 object must be provided");
+  assert(contractAddress, "Contract address must be provided");
+  const contractCode = await web3.eth.getCode(contractAddress);
+  const contractCodeHash = web3.utils.soliditySha3(contractCode);
+  return versionMap[contractCodeHash] || {};
```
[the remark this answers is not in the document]
chrismaree: we could do that but right now there are scenarios where we don't want it to throw, such as when the user defines their own contract version. In these cases it returning `{}` is ideal as the null state is dealt with downstream.

PR #2395, opened 2021-01-11
improve(oo-client): Filter price request events

### LINKED ISSUE #2381: OptimisticOracleClient: Filter price requests, proposals, and disputes and determine which ones should be responded to
**Why**

On each `update()` loop, update the following state (names likely to change):
- `ProposableRequests`: Price requests that have not been proposed to yet
- `DisputableProposals`: Proposals that can be disputed
- `Withdrawable`: Any requests that has rewards that can be withdrawn

Follow up PR to define the actions to execute against these states

nicholaspai:

**Motivation**

#2381

**Summary**

One breaking change to the smart contracts that I made is to add a param to the `PriceProposed` event: `expirationTimestamp`. Without this param, the OO client would need to query the block timestamp of the `PriceProposed` event and then compute its expiration time, which may require an additional web3 call. Do you think this is a reasonable addition to the contracts?

This should be the last OO client PR before we start implementing the OO keeper which takes the OO client state and determines how to respond to `unproposedRequests` and `undisputedProposals`, for example.

Open questions:
- Should the OO client make additional DVM web3 calls to determine which disputed price requests can be resolved?
- Will the OO keeper settle proposals?

**Issue(s)**

Fixes #2381


packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -92,7 +92,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         bytes32 identifier,
         uint256 timestamp,
         bytes ancillaryData,
-        int256 proposedPrice
+        int256 proposedPrice,
+        uint256 expirationTimestamp
```
nicholaspai, line 96: New contract event param that makes the OO client a lot easier to implement (with fewer web3 calls)

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +84,58 @@ class OptimisticOracleClient {
...
-              request.returnValues.ancillaryData ? request.returnValues.ancillaryData : "0x"
-            )
-            .call(),
-        {
-          concurrency: WEB3_CALLS_BATCH_SIZE
-        }
-      )
-    ]);
+    // Store price requests that have NOT been proposed to yet:
+    const unproposedPriceRequests = requestEvents.filter(event => {
```
nicholaspai, line 95: These array filters/maps are a bit ugly but they are better than making additional web3 calls IMO:

Performance:
- web3 calls: (RequestPrice + ProposePrice + DisputePrice)(`N`)
- Event array computations: (requestEvents + 2 * proposalEvents)(`N`)

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Your changes here do not affect the internal state of the contract in any way, so I think this is probably okay. Probably would be good to just shoot this over to OZ when this gets merged.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
mrice32: +1 -- this method of pulling out the state LGTM!

mrice32 (approved):
This looks great!

mrice32:
@nicholaspai 

> * Should the OO client make additional DVM web3 calls to determine which disputed price requests can be resolved?

Since disputes are pretty rare, I don't think this would be much of a problem, so I would say yes.

> * Will the OO keeper settle proposals?

What if it only settled proposals that it's involved in (including disputes)?

nicholaspai:
> @nicholaspai
> 
> > * Should the OO client make additional DVM web3 calls to determine which disputed price requests can be resolved?
> 
> Since disputes are pretty rare, I don't think this would be much of a problem, so I would say yes.
> 
> > * Will the OO keeper settle proposals?
> 
> What if it only settled proposals that it's involved in (including disputes)?

That's a good idea, we can just filter by disputes linked to the proposer and store that in a state retrievable by the OO keeper

nicholaspai:
@mrice32: The client now stores settleable proposals and disputes filtered on proposer and disputer addresses, respectively.

mrice32 (approved):
LGTM!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -293,7 +294,15 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         }
 
         // Event.
```
chrismaree, line 296: this is a pretty useless comment 😆 .

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,125 @@ class OptimisticOracleClient {
...
+                from: this.oracle.options.address
+              })
+          );
+          if (resolvedPrice !== null) {
+            return disputeEvent;
+          } else {
+            return null;
+          }
+        } catch (error) {
+          return null;
```
chrismaree, line 179: do you want to handle this error a bit more gracefully than just returning `null`?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,125 @@ class OptimisticOracleClient {
...
+    const isExpired = proposal => {
+      return Number(proposal.expirationTimestamp) <= Number(currentTime);
+    };
+    this.expiredProposals = undisputedProposals.filter(proposal => isExpired(proposal));
+    this.undisputedProposals = undisputedProposals.filter(proposal => !isExpired(proposal));
+
+    // Store disputes that can be settled:
+    let resolvedPrices = await Promise.all(
+      disputeEvents.map(async disputeEvent => {
+        try {
```
chrismaree, line 154: do we want to have this `try` nested within a `Promise.all, within a `map`? is there a way to change this to remove the nesting a bit and deal with the error the try is catching a bit sooner?

chrismaree (approved):
This is pretty awesome! good work @nicholaspai. I'm excited to see this OO client come to life!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: true! removing.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: I refactored this a bit to make it more concise, I agree returning `null` is a bit ugly so instead I do nothing (i.e. I don't return the `disputeEvent`) and outside of the `Promise.all(array.map)` I remove undefined entries with a one liner.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: What we could do is `map` the `try-catch`, and then pass the map of promises into a `Promise.all`. I wanted to be more concise, but I understand if readability is decreased a lot.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,118 @@ class OptimisticOracleClient {
...
 
+    // Filter proposals based on their expiration timestamp:
+    const isExpired = proposal => {
+      return Number(proposal.expirationTimestamp) <= Number(currentTime);
+    };
+    this.expiredProposals = undisputedProposals.filter(proposal => isExpired(proposal));
+    this.undisputedProposals = undisputedProposals.filter(proposal => !isExpired(proposal));
+
+    // Store disputes that can be settled:
+    let resolvedPrices = await Promise.all(
```
mrice32, line 152: nit: `resolvedPrices` seems to be a misnomer here and below. It appears as if this array is actually being used to store events. Thoughts on renaming it?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,118 @@ class OptimisticOracleClient {
...
+          if (resolvedPrice !== null) {
+            return disputeEvent;
+          }
+        } catch (error) {
+          // No resolved price available, do nothing.
+        }
+      })
+    );
+    // Remove undefined entries, marking disputes that did not have resolved prices
+    resolvedPrices = resolvedPrices.filter(event => event !== undefined);
```
mrice32, line 182: nit: not strongly opinionated (and this may be less readable), but you could chain this with the last line so you don't have to reassign this variable?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,118 @@ class OptimisticOracleClient {
...
+
+    // Filter out disputes that were already settled and reformat data.
+    resolvedPrices = await Promise.all(
```
mrice32, line 185: Same here on naming.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,54 +100,118 @@ class OptimisticOracleClient {
...
+            requester: event.returnValues.requester,
+            proposer: event.returnValues.proposer,
+            disputer: event.returnValues.disputer,
+            identifier: this.hexToUtf8(event.returnValues.identifier),
+            timestamp: event.returnValues.timestamp
+          };
+        }
+      })
+    );
+    resolvedPrices = resolvedPrices.filter(event => event !== undefined);
```
mrice32, line 208: Same here on potentially chaining to avoid reassignment (totally optional/up to you).

mrice32 (approved):
Looks great! Minor comments on variable names. Feel free to merge whenever.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Good point, renamed to `resolvedDisputeEvents`

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Yes I think this is more readable actually.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Chained some `map.filters` [here](https://github.com/UMAprotocol/protocol/pull/2395/commits/9a44bad632db1cfe586127e4c6a9346c0cb30aa0)

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: Implemented [here](https://github.com/UMAprotocol/protocol/pull/2395/commits/9a44bad632db1cfe586127e4c6a9346c0cb30aa0)

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,53 +100,116 @@ class OptimisticOracleClient {
...
+          identifier: this.hexToUtf8(event.returnValues.identifier),
+          timestamp: event.returnValues.timestamp,
+          proposedPrice: event.returnValues.proposedPrice,
+          expirationTimestamp: event.returnValues.expirationTimestamp
+        };
+      });
+
+    // Filter proposals based on their expiration timestamp:
+    const isExpired = proposal => {
+      return Number(proposal.expirationTimestamp) <= Number(currentTime);
```
daywiss, line 147: this comparison looks right 😄

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -73,53 +100,116 @@ class OptimisticOracleClient {
...
+            return disputeEvent;
+          }
+        } catch (error) {
+          // No resolved price available, do nothing.
         }
-      )
-    ]);
+      })
+      // Remove undefined entries, marking disputes that did not have resolved prices
+    ).filter(event => event !== undefined);
```
daywiss, line 182: I just realized you were using bluebird promises which is why this probably works 👍

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
nicholaspai: So the event Params are slightly different, is there a way to see difference on a subset of the params?

PR #2401, opened 2021-01-12
improve(elastic-stablespread): Support new identifier, update uniswap price feed

### LINKED ISSUE #2380: Make BasketSpreadPriceFeed work with ELASTIC_STABLESPREAD identifier
**Why**

`ELASTIC_STABLESPREAD` includes one coin, BAC, which only has a UniswapPriceFeed. This might only require a change in `CreatePriceFeed` which makes an assumption that each basket price feed is a `MedianizerPriceFeed`.

The UniswapPriceFeed does not have custom precision logic yet and the BAC/DAI uniswap pair returns prices in 6 decimals I assume.

The default lookback for this EMP is 172800, to match its liquidation liveness, but the DEX PriceFeeds (i.e. Uni and Balancer) do NOT handle such a large lookback gracefully. For example, the Uniswap price feed throws:

```
{ code: -32005, message: 'query returned more than 10000 results' }
```

- 4: ~2 hours

nicholaspai:

**Motivation**

#2380

Need to update `UniswapPriceFeed` as well to support non-18 precision

**Summary**

In addition to updating `DefaultPriceFeedConfig` to work with the `ELASTIC_STABLESPREAD` identifier, updates the Uniswap price feed in two important ways:
- Makes the `getPastEvents` lookup much more efficient by setting a `fromBlock` param. This call was actually throwing an error, "query returned more than 10000 results", when testing on the `BAC/DAI` uniswap pool which has a lot trade history than the current uniswap pools that we have tested on (uGas).
- Make UniswapPriceFeed work with non-18 decimal precision. Currently, the price feed assumes that the LP returns prices in 18 decimals and that the user of the PriceFeed wants 18 precision. This PR changes the logic to be consistent with the `BalancerPriceFeed` which works with any precision.

**Testing**
EMP Address shown for Zelda Cash ElasticStablespread/USDC, fill in `apiKey` if you get rate-limited by CryptoWatch:

```
EMP_ADDRESS=0xf215778F3a5e7Ab6A832e71d87267Dd9a9aB0037
PRICE_FEED_CONFIG={"apiKey":"XXX"}
MEDIANIZER_PRICE_FEED_CONFIG={"apiKey":"XXX"}
MONITOR_CONFIG={"deviationAlertThreshold":0.0001,"volatilityWindow":3600,"pegVolatilityAlertThreshold":0.0001,"syntheticVolatilityAlertThreshold":0.0001}
```

`yarn truffle exec ./packages/monitors/index.js --network mainnet_mnemonic`

**Issue(s)**

Fixes #2380

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -86,7 +86,9 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.twapLength,
       config.lookback,
       getTime,
-      config.invertPrice // Not checked in config because this parameter just defaults to false.
+      config.invertPrice, // Not checked in config because this parameter just defaults to false.
+      config.poolDecimals,
+      config.priceFeedDecimals // This defaults to 18 unless supplied by user
     );
   } else if (config.type === "medianizer") {
     const requiredFields = ["medianizedFeeds"];
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -265,7 +265,6 @@ const defaultConfigs = {
       },
       {
         type: "medianizer",
-        computeMean: true,
         medianizedFeeds: [
           // BAC/DAI:
           { type: "uniswap", uniswapAddress: "0xd4405f0704621dbe9d4dea60e128e0c3b26bddbd", twapLength: 2 }
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -250,6 +250,52 @@ const defaultConfigs = {
       ]
     }
   },
+  "ELASTIC_STABLESPREAD/USDC": {
+    type: "basketspread",
+    lookback: 7200,
+    minTimeBetweenUpdates: 60,
+    experimentalPriceFeeds: [
+      {
+        type: "medianizer",
+        medianizedFeeds: [{ type: "cryptowatch", exchange: "uniswap-v2", pair: "fraxusdc" }]
+      },
+      {
+        type: "medianizer",
+        medianizedFeeds: [{ type: "cryptowatch", exchange: "uniswap-v2", pair: "esdusdc" }]
+      },
+      {
+        type: "medianizer",
+        computeMean: true,
+        medianizedFeeds: [
+          // BAC/DAI:
+          { type: "uniswap", uniswapAddress: "0xd4405f0704621dbe9d4dea60e128e0c3b26bddbd", twapLength: 2 }
+        ]
+      }
+    ],
+    baselinePriceFeeds: [
+      {
+        type: "medianizer",
+        medianizedFeeds: [
+          {
+            type: "medianizer",
+            computeMean: true,
+            medianizedFeeds: [
+              { type: "cryptowatch", exchange: "bitfinex", pair: "usdtusd" },
+              { type: "cryptowatch", exchange: "kraken", pair: "usdtusd" }
+            ]
+          },
+          {
+            type: "medianizer",
+            computeMean: true,
+            medianizedFeeds: [
+              { type: "cryptowatch", exchange: "kraken", pair: "usdcusd" },
+              { type: "cryptowatch", exchange: "bitstamp", pair: "usdcusd" }
+            ]
+          }
+        ]
+      }
+    ]
+  },
   "GASETH-TWAP-1Mx1M": {
     type: "uniswap",
     uniswapAddress: "0x25fb29D865C1356F9e95D621F21366d3a5DB6BB0",
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -149,7 +149,7 @@ const defaultConfigs = {
         computeMean: true,
         medianizedFeeds: [
           { type: "cryptowatch", exchange: "binance", pair: "busdusdt" },
-          { type: "cryptowatch", exchange: "uniswap-v2", pair: "busdusdt" }
+          { type: "uniswap", uniswapAddress: "0xa0abda1f980e03d7eadb78aed8fc1f2dd0fe83dd", twapLength: 2 }
         ]
       },
       {
--- a/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
@@ -2,7 +2,7 @@
 
 const { PriceFeedInterface } = require("./PriceFeedInterface");
 const { computeTWAP } = require("./utils");
-const { ConvertDecimals } = require("@uma/common");
+const { ConvertDecimals, averageBlockTimeSeconds } = require("@uma/common");
 class UniswapPriceFeed extends PriceFeedInterface {
   /**
    * @notice Constructs new uniswap TWAP price feed object.
@@ -85,8 +85,12 @@ class UniswapPriceFeed extends PriceFeedInterface {
   }
 
   async update() {
-    // TODO: optimize this call. This may be very slow or break if there are many transactions.
-    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: 0 });
+    // Approximate the first block from which we'll need price data from based on the
+    // lookback and twap length:
+    const lookbackWindow = this.twapLength + this.historicalLookback;
+    const latestBlockNumber = (await this.web3.eth.getBlock("latest")).number;
+    const earliestBlockNumber = latestBlockNumber - Math.ceil(lookbackWindow / (await averageBlockTimeSeconds()));
+    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: earliestBlockNumber });
 
     // If there are no prices, return null to allow the user to handle the absense of data.
     if (events.length === 0) {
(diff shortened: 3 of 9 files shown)


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -149,7 +149,7 @@ const defaultConfigs = {
         computeMean: true,
         medianizedFeeds: [
           { type: "cryptowatch", exchange: "binance", pair: "busdusdt" },
-          { type: "cryptowatch", exchange: "uniswap-v2", pair: "busdusdt" }
+          { type: "uniswap", uniswapAddress: "0xa0abda1f980e03d7eadb78aed8fc1f2dd0fe83dd", twapLength: 2 }
```
nicholaspai, line 152: The cryptowatch endpoint for this pair is particularly flaky, subbing it out with our hardened Uniswap price feed

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -86,7 +86,9 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.twapLength,
       config.lookback,
       getTime,
-      config.invertPrice // Not checked in config because this parameter just defaults to false.
+      config.invertPrice, // Not checked in config because this parameter just defaults to false.
+      config.poolDecimals,
```
mrice32, line 90: Is the pool decimals derivable by looking at the decimals of the two tokens in the pool? In what cases would this not work?

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -55,16 +77,20 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
   }
 
   async update() {
-    // TODO: optimize this call. This may be very slow or break if there are many transactions.
-    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: 0 });
+    // Approximate the first block from which we'll need price data from based on the
+    // lookback and twap length:
+    const lookbackWindow = this.twapLength + this.historicalLookback;
+    const latestBlockNumber = (await this.web3.eth.getBlock("latest")).number;
+    const earliestBlockNumber = latestBlockNumber - Math.ceil(lookbackWindow / (await averageBlockTimeSeconds()));
```
mrice32, line 92: 1. Should we cushion this a little so we don't undershoot in case the average is off?
2. What happens if the pool just doesn't trade much? (i.e. the price is constant) for the entire lookback period?

mrice32 (commented):
Looks awesome!!! Two high-level questions.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
nicholaspai: You know I'm not sure how it actually works. Take the USDC/ETH pool for example which emits a `Sync` event with  `reserve0` in 6 decimal precision, and `reserve1` in 18 decimal precision. Perhaps the `poolDecimals` is equal to `reserve0`, unless the price is inverted?

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: 1. I wanted to avoid hardcoding a value but I'll add a cushion of 10%
2. This is a really good point, perhaps in this case where there are no events in the lookback then we should crawl backwards until we find a Sync event.

nicholaspai: 1. Implemented a hard-coded cushion of 10% for 1, (and changed the Balancer price feed's lookback cushion to be a % rather than a fixed # of blocks)
2. Added a while loop to crawl backwards for events until either `events.length > 0` or the `fromBlock` passed to `getPastEvents` is 0

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -55,18 +81,32 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+
+    // For low-volume pools, it is possible that there are no Sync events within the lookback window.
+    // To cover these cases, we'll keep looking back until we find a window with a sync event.
+    while (fromBlock >= 0 && events.length === 0) {
+      fromBlock -= lookbackBlocks;
+      events = await this.uniswap.getPastEvents("Sync", { fromBlock: Math.max(fromBlock, 0) });
+    }
 
-    // If there are no prices, return null to allow the user to handle the absense of data.
+    // If there are still no prices, return null to allow the user to handle the absense of data.
```
pemulis, line 109: `absence`

nicholaspai: nice catch, fixed

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
chrismaree: in the case of `USDC/ETH` is it 6 decimals? It could be the lower decimals of the two tokens within the pair.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -26,14 +39,26 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+    this.poolPrecision = poolDecimals;
+    this.decimals = priceFeedDecimals;
 
     // Helper functions from web3.
     this.toBN = this.web3.utils.toBN;
     this.toWei = this.web3.utils.toWei;
+
+    // Convert _bn precision from poolDecimals to desired decimals by scaling up or down based
+    // on the relationship between poolPrecision and the desired decimals.
+    this.convertDecimals = ConvertDecimals(this.poolPrecision, this.decimals, this.web3);
```
chrismaree, line 57: nit: can you rename this method `convertPoolDecimalsToPriceFeedDecimals`? I know it's a long function name but `convertDecimals` is too generic I think & I want us to be as consistent as possible within the feeds when talking about decimal conversion. For example, the crypto watch refers to `convertPriceFeedDecimals` so this will be consistent with that naming convention.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -26,14 +39,26 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
     this.historicalLookback = historicalLookback;
     this.invertPrice = invertPrice;
+    // The % of the lookback window (historicalLookback + twapLength) that we want to query for Uniswap
+    // Sync events. For example, 1.1 = 110% meaning that we'll look back 110% * (historicalLookback + twapLength)
+    // seconds, in blocks, for Sync events.
+    this.bufferBlockPercent = 1.1;
+
+    // TODO: Should/Can we read in `poolDecimals` from the this.uniswap?
+    this.poolPrecision = poolDecimals;
+    this.decimals = priceFeedDecimals;
```
chrismaree, line 49: nit: can you store this variable as `this.priceFeedDecimals` for consistency to be explicit what kind of decimals you are referring to?

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -26,14 +39,26 @@ class UniswapPriceFeed extends PriceFeedInterface {
     this.getTime = getTime;
     this.historicalLookback = historicalLookback;
     this.invertPrice = invertPrice;
+    // The % of the lookback window (historicalLookback + twapLength) that we want to query for Uniswap
+    // Sync events. For example, 1.1 = 110% meaning that we'll look back 110% * (historicalLookback + twapLength)
+    // seconds, in blocks, for Sync events.
+    this.bufferBlockPercent = 1.1;
+
+    // TODO: Should/Can we read in `poolDecimals` from the this.uniswap?
+    this.poolPrecision = poolDecimals;
```
chrismaree, line 48: nit: why do you go from calling this `Decimals` to `Precision`? Is this not a scaling factor (decimals) and not a unit of precision?

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -55,18 +81,32 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
 
   getPriceFeedDecimals() {
-    return 18;
+    return this.decimals;
   }
 
   async update() {
-    // TODO: optimize this call. This may be very slow or break if there are many transactions.
-    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: 0 });
+    // Approximate the first block from which we'll need price data from based on the
```
chrismaree, line 92: nit: comment line length

+1 on optimizing this!

chrismaree (approved):
This looks great! nice updates and generalization. I have a few nits on variable names and layout but otherwise, this looks good to me.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
nicholaspai: Its 6 decimals. I THINK its the lower of the two decimals, but I'm not sure if the ordering of `reserve0` or `reserve1` matters. This seems like I need to dig into it a bit more which is why I left it as a todo (same with the BalancerPriceFeed's poolPrecision FYI). Ideally I'd like this not to block this PR

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: Done and made the same change for `BalancerPriceFeed`

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: Done and did the same for `BalancerPriceFeed`

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: Good point I'll rename to `poolDecimals`

mrice32 (approved):
LGTM!

PR #1924, opened 2020-08-31
Feat(bots): Bots Using Named Parameters

kendricktan:
**Motivation**

The amount of parameters each bot takes increases everytime we add a new feature into it. Having them in a object destructuring notation makes it:

1. Easier to eat
2. Less error prone:

i.e.
```javascript
new Bot({
    logger,
    config,
    streamAbi: contractAbi
    botConfig: innerBotConfig,
    priceFeedOracle: priceOracle,
})
```

is a lot easier to and spot errors (when args are out of order) reason about then:

```javascript
new Bot(logger, config, details, innerBotConfig, someAbi, priceOracle, contractAbi)
```

**Summary**

Added object structuring notation / named params to the constructor of `liquidator`, `monitors`, and `reporters`.

**Issue(s)**

Fixes https://github.com/UMAprotocol/protocol/issues/1869

PR #2409, opened 2021-01-13
fix(monitor,disputer,liquidator bots): correctly pass in bot config objects into bots

chrismaree:
**Motivation**
While working on the perp liquidator bot I found a subtle but substantial bug present within the Liquidator & Disputer bots regards to how configs are passed around.

**Summary**

PR https://github.com/UMAprotocol/protocol/pull/1924 merged on 2 Sep 2020 refactored the bots to pass around bot parameters as named values. However, this PR introduced a bug by incorrectly naming the `liquidatorConfig`. This bug can be seen in the following places:

1) in the `index.js` we feed in the `liquidatorConfig` as a named type `liquidatorConfig` here: https://github.com/UMAprotocol/protocol/blob/d782c56503a220a7eaa14baa8d03d925b5e956b6/packages/liquidator/index.js#L184
2) HOWEVER the `liquidator.js` file refers to this variable as `config`, here: https://github.com/UMAprotocol/protocol/blob/d782c56503a220a7eaa14baa8d03d925b5e956b6/packages/liquidator/src/liquidator.js#L37

The same exact issue is present within the disputer between `disputerConfig` and `config` in the `disputer/index.js` and `disputer.js` files respectively. The bug was not presented within the monitor bot but the bots were updated for consistency with the liquidator and disputer.

**Bug implications**
As a result of the bug, no `LiquidatorConfigs` are _actually_ passed into the bots! This means that the following are not passed in:
a) `crThreshold`
b) `liquidationMinPrice`
d) `defenseActivationPercent`
e) `whaleDefenseFundWei`
g) `logOverrides`
h) `liquidationDeadline`
i) `txnGasLimit`

Note that almost none of these overrides were used in production. However, two pretty important ones were, namely: a) the `logOverrides` which was meant to change liquidation events from `info` to `error` (as specified in the bot configs and was not. b) More worryingly all of the `whaleDefenseFund` configs were not correctly passed in.

In the case of the disputer this means the following configs were not correctly passed in:
a) txnGasLimit
b) disputeDelay

Neither of these vars are used in production.

**Details**

To fix this, this PR simply re-maps the key-value pairs correctly.

### DIFF
--- a/packages/disputer/src/disputer.js
+++ b/packages/disputer/src/disputer.js
@@ -15,7 +15,7 @@ class Disputer {
    * @param {String} account Ethereum account from which to send txns.
    * @param {Object} empProps Contains EMP contract state data. Expected:
    *      { priceIdentifier: hex("ETH/BTC") }
-   * @param {Object} [config] Contains fields with which constructor will attempt to override defaults.
+   * @param {Object} [disputerConfig] Contains fields with which constructor will attempt to override defaults.
    */
   constructor({
     logger,
@@ -25,7 +25,7 @@ class Disputer {
     priceFeed,
     account,
     empProps,
-    config
+    disputerConfig
   }) {
     this.logger = logger;
     this.account = account;
@@ -55,7 +55,7 @@ class Disputer {
     this.GAS_LIMIT_BUFFER = 1.25;
 
     // Default config settings. Disputer deployer can override these settings by passing in new
-    // values via the `config` input object. The `isValid` property is a function that should be called
+    // values via the `disputerConfig` input object. The `isValid` property is a function that should be called
     // before resetting any config settings. `isValid` must return a Boolean.
     const defaultConfig = {
       disputeDelay: {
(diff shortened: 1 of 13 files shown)


nicholaspai (approved):
Nice catch LGTM

packages/disputer/src/disputer.js
```diff
@@ -15,7 +15,7 @@ class Disputer {
    * @param {String} account Ethereum account from which to send txns.
    * @param {Object} empProps Contains EMP contract state data. Expected:
    *      { priceIdentifier: hex("ETH/BTC") }
-   * @param {Object} [config] Contains fields with which constructor will attempt to override defaults.
+   * @param {Object} [disputerConfig] Contains fields with which constructor will attempt to override defaults.
```
mrice32, line 18: nit: not related to this line, but the title doesn't conform to conventional commit standards:
- `bug` should be changed to `fix`
- There should be a `:` after `fix(...)`.

Correcting those two, the title should be:
```
fix(monitor,disputer,liquidator bots): correctly pass in bot config objects into bots
```

mrice32 (approved):
Whew. Very nice catch!!

packages/disputer/src/disputer.js
chrismaree: thanks for pointing this out. will follow the convention correctly in the future :)

PR #2413, opened 2021-01-14
improve(voter-app): update app to work with newest voting version 1.2.2

### LINKED ISSUE #1969: Errors trying to decrypt votes in both voterdapp and cli
**Description**
Unable to decode votes causing votedapp to crash and cli tool to not work on reveal.

**Reproduction**

Build protocol/core and follow directions in voter-dapp/run_tests.sh
Get to snapshot phase 6/10 and enter votes in dapp
see that there is an error in dapp and unable to decode.
Move to reveal phase.
Reveal votes in cli, see that you get a bad decrypt.

**Expected Behavior**

Should just work without errors and show your votes.

**Screenshots**

Using CLI:
image

Using VoterDapp:
image

**Version and OS Info**

Running Ubuntu 18.04. 
Chrome fresh user
Lastest protocol/master
Fresh metamask

**Additional Context**
This might be a linux environment issue, most likely will start to debug there.

### LINKED ISSUE #2371: Re-release existing voter dApp to work with new DVM ABIs
Why
There were breaking changes in the new voting contract, specifically around handling ancillary data. The current implementation is backward compatible currently with the dapp, but we want to update it to call the new functions.

AC
- See that voter dapp has no regressions
- see that dapp uses new ancillary functions under the hood

daywiss:

**Motivation**

#2371

**Summary**

Adds support for latest contract abi with ancillary data and fixes some bugs.

**Details**
- updates package file to latest abis
- uses ancillary data in calls when needed. also sets it to 0x since there is no ability to specify in this pr
- fixes infinite loop in requesting resolved requests
- tested with run_test script for 2key and non 2key scenarios
- "fixes" bug when decoding encrypted data which i ran into again, see #1969. this was resolved by updating the library version of `eth-crypto`

**Issue(s)**

Fixes #2371
Fixes #1969

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -42,7 +42,7 @@
     "bignumber.js": "^8.0.1",
     "chalk-pipe": "^3.0.0",
     "dotenv": "^6.2.0",
-    "eth-crypto": "^1.6.0",
+    "eth-crypto": "^1.7.0",
     "hardhat": "^2.0.4",
     "hardhat-gas-reporter": "^1.0.3",
     "minimist": "^1.2.0",
--- a/packages/core/scripts/local/RequestOraclePrice.js
+++ b/packages/core/scripts/local/RequestOraclePrice.js
@@ -35,7 +35,7 @@ async function registerContract(registry) {
 //   2. Registers the deployer's address as a derivative.
 //   3. Adds the specified identifier with Voting.
 //   4. Requests a price at the specified time.
-async function run(deployedFinder, identifier, timeString) {
+async function run(deployedFinder, identifier, timeString, ancillaryData = "") {
   try {
     const deployedRegistry = await Registry.at(
       await deployedFinder.getImplementationAddress(web3.utils.utf8ToHex(interfaceName.Registry))
@@ -43,6 +43,7 @@ async function run(deployedFinder, identifier, timeString) {
     await registerContract(deployedRegistry);
 
     const identifierInBytes = web3.utils.fromAscii(identifier);
+    const ancillaryDataInBytes = ancillaryData.length ? web3.utils.fromAscii(ancillaryData) : null;
 
     let timeInSeconds = parseInt(timeString);
     if (timeInSeconds === 0) {
@@ -61,13 +62,24 @@ async function run(deployedFinder, identifier, timeString) {
       await deployedFinder.getImplementationAddress(web3.utils.utf8ToHex(interfaceName.IdentifierWhitelist))
     );
     await deployedIdentifierWhitelist.addSupportedIdentifier(identifierInBytes);
-    const priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN);
+    let priceExists;
+
+    if (ancillaryDataInBytes) {
+      priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN, ancillaryDataInBytes);
+    } else {
+      priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN);
+    }
     if (priceExists) {
-      console.log(`Price already exists for ${identifier} @ ${time}`);
+      console.log(`Price already exists for ${identifier} @ ${time} : ${ancillaryData}`);
       return;
     }
 
-    await deployedVoting.requestPrice(identifierInBytes, timeInBN);
+    if (ancillaryDataInBytes) {
+      await deployedVoting.requestPrice(identifierInBytes, timeInBN, ancillaryDataInBytes);
+    } else {
+      await deployedVoting.requestPrice(identifierInBytes, timeInBN);
+    }
+
     console.log(`Price requested for ${identifier} @ ${time}`);
   } catch (err) {
     console.error(err);
@@ -102,10 +114,10 @@ const runRequestOraclePrice = async function(callback) {
       }
 
       // Request the price.
-      await run(finder, argv.identifier, argv.time);
+      await run(finder, argv.identifier, argv.time, argv.ancillaryData);
     });
   } else {
-    await run(finder, argv.identifier, argv.time);
+    await run(finder, argv.identifier, argv.time, argv.ancillaryData);
 
     // Note: only call the callback in the non-server case. In the server case, the server is expected to run
     // indefinitely.
--- a/packages/voter-dapp/package.json
+++ b/packages/voter-dapp/package.json
@@ -9,7 +9,6 @@
     "@material-ui/styles": "^4.3.0",
     "@uma/common": "^2.0.1",
     "@uma/core": "^2.0.1",
-    "@uma/core-1-2-0": "npm:@uma/core@1.2.0",
     "@umaprotocol/react-plugin": "^1.5.5",
     "@umaprotocol/store": "^1.5.7",
     "bignumber.js": "^8.0.1",
--- a/packages/voter-dapp/run_tests.sh
+++ b/packages/voter-dapp/run_tests.sh
@@ -39,18 +39,14 @@ echo -e '####################################################################\n'
 # This forces the voter dapp to deploy the _new_ voting contract with all the other contracts unchanged, which will
 # force it to test the old 2key contract with the new voting contract.
 yarn truffle compile
-ARTIFACT_PATH_1_2_0=$(node -p "require.resolve('@uma/core-1-2-0/build/contracts/Voting.json')")
-ARTIFACT_PATH_NEW=$(node -p "require.resolve('@uma/core/build/contracts/Voting.json')")
-BACKUP=$(mktemp -d)/Voting.json
-echo "$ARTIFACT_PATH_1_2_0 is being modified in a breaking way. Backup at $BACKUP."
-cp $ARTIFACT_PATH_1_2_0 $BACKUP
-node -p "JSON.stringify({...require('$BACKUP'), bytecode: require('$ARTIFACT_PATH_NEW').bytecode }, null, 2)" > $ARTIFACT_PATH_1_2_0
 yarn run truffle migrate --reset --network test
 echo "- ✅ Migration complete!"
 
-# Submit 2 normal price requests:
-# - 1 to commit & reveal on browser
-# - 1 to commit & reveal on browser with 2-Key contract
+# Submit 4 normal price requests:
+# - 1 to commit & reveal on browser 8 decimals
+# - 1 to commit & reveal on browser 6 decimals
+# - 1 to commit & reveal on browser 8 decimals and ancillary data
+# - 1 to commit & reveal on browser 6 decimals and ancillary data
 # Submit 1 admin price request:
 # - 1 to commit & reveal on browser
 echo -e '\n####################################################################'
@@ -59,7 +55,9 @@ echo '# 2/16. Submitting price requests                                  #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
 yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier TEST8DECIMALS --time 1570000000
-yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier TEST6DECIMALS --time 1570000000
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier TEST6DECIMALS --time 1570000000 
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier TEST8DECIMALSANCIL --time 1570000000 --ancillaryData='8dec'
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier TEST6DECIMALSANCIL --time 1570000000 --ancillaryData='6dec'
 yarn run truffle exec ../core/scripts/mainnet/ProposeAdmin.js --network test --prod
 echo "- ✅ Price requests submitted!"
 
--- a/packages/voter-dapp/run_tests.sh
+++ b/packages/voter-dapp/run_tests.sh
@@ -187,7 +187,7 @@ echo '#                                                                  #'
 echo '# 11/16. Create a voting proxy contract                            #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
-echo "- Follow this tutorial to set up your 2Key wallet via the voter dApp: https://docs.umaproject.org/tutorials/voting-2key. Use your current address as the Cold Wallet Address and transfer your tokens to the proxy contract. You will be prompted to connect again to the voter dApp, but if this doesn't happen automatically then you can simply refresh the page. The Resolved Requests will no longer show that you voted on the two resolved price requests because your voting account has changed to the proxy contract's address"
+echo "- Follow this tutorial to set up your 2Key wallet via the voter dApp: https://docs.umaproject.org/uma-tokenholders/voting-2key. Use your current address as the Cold Wallet Address and transfer your tokens to the proxy contract. You will be prompted to connect again to the voter dApp, but if this doesn't happen automatically then you can simply refresh the page. The Resolved Requests will no longer show that you voted on the two resolved price requests because your voting account has changed to the proxy contract's address"
 select yn in "Continue" "Help" "Exit"; do
     case $yn in
         Continue ) break;;
--- a/packages/voter-dapp/src/ActiveRequests.js
+++ b/packages/voter-dapp/src/ActiveRequests.js
@@ -36,7 +36,7 @@ import {
   deriveKeyPairFromSignatureMetamask,
   encryptMessage,
   getKeyGenMessage,
-  computeVoteHash,
+  computeVoteHashAncillary,
   getRandomUnsignedInt,
   BATCH_MAX_COMMITS,
   BATCH_MAX_REVEALS,
@@ -141,7 +141,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
         setPendingRequests(nonBlacklistedRequests);
       }
     }
-  }, [allPendingRequests, IDENTIFIER_BLACKLIST, showSpamRequests]);
+  }, [allPendingRequests, hexToUtf8, showSpamRequests]);
 
   const { roundVoteData, getRequestKey } = VoteData.useContainer();
 
@@ -350,7 +350,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
   const decryptionComplete = decryptedCommits && voteStatuses && decryptedCommits.length === voteStatuses.length;
 
   const { send: batchRevealFunction, status: revealStatus } = useCacheSend(votingGateway, "batchReveal");
-  const { send: snapshotCurrentRound, status: snapshotStatus } = useCacheSend(snapshotContract, "snapshotCurrentRound");
+  const { send: snapshotCurrentRound } = useCacheSend(snapshotContract, "snapshotCurrentRound");
 
   const onSnapshotHandler = () => {
     const hashedSnapshotMessage = web3.utils.soliditySha3(snapshotMessage);
@@ -361,6 +361,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       .catch(err => console.log("snapshot error", err));
   };
 
+  // keccak256(abi.encodePacked(price, salt, msg.sender, time, ancillaryData, roundId, identifier))
   const onClickHandler = () => {
     const reveals = [];
     for (const index in checkboxesChecked) {
@@ -368,10 +369,9 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
         reveals.push({
           identifier: pendingRequests[index].identifier,
           time: pendingRequests[index].time,
+          ancillaryData: pendingRequests[index].ancillaryData || DEFAULT_ANCILLARY_DATA,
           price: decryptedCommits[index].price.toString(),
-          salt: decryptedCommits[index].salt,
-          // TODO: This will need to be updated eventually when data input from user
-          ancillaryData: DEFAULT_ANCILLARY_DATA
+          salt: decryptedCommits[index].salt
         });
       }
     }
@@ -398,8 +398,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       const identifierPrecision = getPrecisionForIdentifier(hexToUtf8(pendingRequests[index].identifier));
       const price = parseFixed(editState[index], identifierPrecision).toString();
       const salt = getRandomUnsignedInt().toString();
-      // TODO: This will need to be pulled from user input in the future
-      const ancillaryData = DEFAULT_ANCILLARY_DATA;
+      const ancillaryData = pendingRequests[index].ancillaryData || DEFAULT_ANCILLARY_DATA;
       const encryptedVote = await encryptMessage(
         decryptionKeys[account][currentRoundId].publicKey,
         JSON.stringify({ price, salt })
@@ -408,7 +407,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
         identifier: pendingRequests[index].identifier,
         time: pendingRequests[index].time,
         ancillaryData,
-        hash: computeVoteHash({
+        hash: computeVoteHashAncillary({
           price,
           salt,
           account: votingAccount,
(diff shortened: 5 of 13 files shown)


packages/voter-dapp/src/ResolvedRequests.js
```diff
@@ -84,8 +84,9 @@ function ResolvedRequests({ votingAccount }) {
   const newProposalEvents = useCacheEvents(
     "Governor",
     "NewProposal",
-    useMemo(() => ({ fromBlock: 0 }))
+    useMemo(() => ({ fromBlock: 0 }), [])
```
daywiss, line 87: This is what was causing infinite loop

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -248,11 +248,17 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
     // Get the latest `encryptedVote` for each (identifier, time).
     const encryptedVoteMap = {};
     for (const ev of encryptedVoteEvents) {
-      encryptedVoteMap[toPriceRequestKey(ev.returnValues.identifier, ev.returnValues.time)] =
-        ev.returnValues.encryptedVote;
+      encryptedVoteMap[
+        // Ancillary data is undefined or null in these "returnValues" so we add a default string of "0x".
+        // If not added, the key will improperly index and the votes will not be found.
+        toPriceRequestKey(ev.returnValues.identifier, ev.returnValues.time, ev.returnValues.ancillaryData || "0x")
```
daywiss, line 254: this is ugly but for some reason ev.returnValues.ancillaryData is undefined, even though it was sent as 0x on the vote. This causes an issue if you dont default it to '0x'

packages/core/scripts/mainnet/ProposeAdmin.js
```diff
@@ -32,7 +32,7 @@ async function propose(callback) {
     const signingAccount = (await web3.eth.getAccounts())[0];
     console.group(`Proposer account: ${signingAccount}`);
     if (signingAccount !== "0x2bAaA41d155ad8a4126184950B31F50A1513cE25") {
-      throw new Error("Cannot propose admin votes from this account");
+      console.warn("Warning: Cannot propose admin votes if using a mainnet provider from the current account");
```
nicholaspai, line 35: +1

packages/voter-dapp/run_tests.sh
```diff
@@ -177,7 +178,7 @@ echo '#                                                                  #'
...
 echo -e '####################################################################\n'
-echo "- Follow this tutorial to set up your 2Key wallet via the voter dApp: https://docs.umaproject.org/tutorials/voting-2key. Use your current address as the Cold Wallet Address and transfer your tokens to the proxy contract. You will be prompted to connect again to the voter dApp, but if this doesn't happen automatically then you can simply refresh the page. The Resolved Requests will no longer show that you voted on the two resolved price requests because your voting account has changed to the proxy contract's address"
+echo "- Follow this tutorial to set up your 2Key wallet via the voter dApp: https://docs.umaproject.org/uma-tokenholders/voting-2key. Use your current address as the Cold Wallet Address and transfer your tokens to the proxy contract. You will be prompted to connect again to the voter dApp, but if this doesn't happen automatically then you can simply refresh the page. The Resolved Requests will no longer show that you voted on the two resolved price requests because your voting account has changed to the proxy contract's address"
```
nicholaspai, line 181: Thanks for updating this script!

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -73,9 +73,10 @@ const editStateReducer = (state, action) => {
   }
 };
 
-const toPriceRequestKey = (identifier, time) => time + "," + identifier;
-const toVotingAccountAndPriceRequestKey = (votingAccount, identifier, time) =>
-  votingAccount + "," + time + "," + identifier;
+// toPriceRequestKey( identifier,time, ancillarydata )
+const toPriceRequestKey = (...args) => args.join(",");
+// toVotingAccountAndPriceRequestKey( votingAccount, identifier, time, ancillarydata )
+const toVotingAccountAndPriceRequestKey = (...args) => args.join(",");
```
nicholaspai, line 79: `toPriceRequestKey` and `toVotingAccountAndPriceRequestKey` now implement the same function. I find it a bit more readable to define the different vars that will be joined with commas, similar to how you define it in the comments. Why'd you change it?

packages/voter-dapp/src/ActiveRequests.js
nicholaspai: I experienced this also when testing the `OptimisticOracleClient` and I think using the `var || 0x` is better than how I addressed it which was the ternary operator: `var ? var : 0x`

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -312,9 +318,13 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       const currentVotes = await Promise.all(
         voteStatuses.map(async (voteStatus, index) => {
           if (voteStatus.committedValue) {
-            return JSON.parse(
-              await decryptMessage(decryptionKeys[account][currentRoundId].privateKey, voteStatus.committedValue)
-            );
+            try {
+              return JSON.parse(
+                await decryptMessage(decryptionKeys[account][currentRoundId].privateKey, voteStatus.committedValue)
```
nicholaspai, line 323: Does this still fail since we last talked? Catching this error might confuse users in case the messages cannot be decrypted, but I suppose its better than crashing the App.

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -353,7 +363,8 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
           identifier: pendingRequests[index].identifier,
           time: pendingRequests[index].time,
           price: decryptedCommits[index].price.toString(),
-          salt: decryptedCommits[index].salt
+          salt: decryptedCommits[index].salt,
+          ancillaryData: "0x"
```
nicholaspai, line 367: Shouldn't we be reading the ancillary data from `decryptedCommits` or is the decryption not working?

nicholaspai: Actually, see my comment below in the `commits.push` code.

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -387,6 +398,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       commits.push({
         identifier: pendingRequests[index].identifier,
         time: pendingRequests[index].time,
+        ancillaryData: "0x",
```
nicholaspai, line 401: So clearly the committed `ancillaryData` and the revealed `ancillaryData` have to be identical. Furthermore, in the future we might want to change this field dynamically depending on the type of vote we're having. Therefore I think its sensible to set a constant ANCILLARY_DATA object at the file level that both this and reveals can read from. This way we can more easily upgrade the ancillary data logic to be dynamic based on types of votes.

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -408,7 +420,8 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       const newCommitKey = toVotingAccountAndPriceRequestKey(
         votingAccount,
         pendingRequests[index].identifier,
-        pendingRequests[index].time
+        pendingRequests[index].time,
+        pendingRequests[index].ancillaryData
```
nicholaspai, line 424: Wait, why do you use `pendingRequests[index].ancillaryData` here but hardcode `0x` above?

packages/voter-dapp/src/ResolvedRequests.js
nicholaspai: Why does this fix the infinite loop? Really elegant solution if so

nicholaspai (commented):
The changes are well implemented and overall its not much code which is awesome! I have a comment about how you set/get `ancillaryData` for commits and reveals.

Also, if this PR goes in will the voter dApp no longer be compatible with the current DVM? If so, we should discuss how to handle votes until we migrate the DVM. Perhaps we deploy a version of the web app to a new URL.

packages/voter-dapp/src/ActiveRequests.js
daywiss: Yea good point, i was a bit on the fence about this. The reason why is that it can accept arbitrary arguments now, for instance with or without ancillary data if needed, though this might not be a good enough reason.  I dont mind changing it back.

packages/voter-dapp/src/ActiveRequests.js
daywiss: This was failing for me until I bumped the eth-crypto version from 1.6.0 to 1.7.0 in the common package. I dont know if this will break other things, but doesnt seem like it so far.  I left the try catch in since this wasnt really being handled.

packages/voter-dapp/src/ActiveRequests.js
daywiss: you might be right about this, probably a mistake.

packages/voter-dapp/src/ActiveRequests.js
daywiss: Yea I think i made some mistakes here, good catch

packages/voter-dapp/src/ActiveRequests.js
daywiss: The next PR will be to add a field so the user can change ancillary data, but yea the default should at least be a constant.

packages/voter-dapp/src/ResolvedRequests.js
daywiss: i think because if you dont unclude the `[]` it will run every render, and its probably causing a rerender when it runs. by adding the `[]` it means just run once.

daywiss:
@nicholaspai thanks for the review. I addressed your concerns. I'm still not 100% sure of this but it does work correctly in the test. I think the next addition when adding user input for ancillary data should shake out further issues.

daywiss:
> The changes are well implemented and overall its not much code which is awesome! I have a comment about how you set/get `ancillaryData` for commits and reveals.
> 
> Also, if this PR goes in will the voter dApp no longer be compatible with the current DVM? If so, we should discuss how to handle votes until we migrate the DVM. Perhaps we deploy a version of the web app to a new URL.

I think @mrice32 mentioned he has a plan for this release and noone seemed too concerned about this... Maybe matt can comment

packages/voter-dapp/src/ResolvedRequests.js
nicholaspai: I think the second arg to any of the `useX` hooks is an array of dependencies, so I guess if you leave out the array of dependency then it will always re-render. If you explicitly set the dependencies to non with `[]` then it should only render once.

packages/voter-dapp/src/ActiveRequests.js
nicholaspai: > The reason why is that it can accept arbitrary arguments now, for instance with or without ancillary data if needed

This is definitely useful, but I think its confusing to have two functions implement the same logic. So if you want the flexibility of an unspecified amount of args, maybe consolidate the functions into one `toPriceRequestKey`

packages/voter-dapp/src/ActiveRequests.js
nicholaspai: Got it. I'm OK with leaving that here so the App doesn't crash, but its still going to be jarring for the user if they can't see their committed value (they will wonder, "did my commit go through correctly"?)

packages/voter-dapp/src/ResolvedRequests.js
daywiss: yes thats my understanding and explained a lot better

packages/voter-dapp/src/ActiveRequests.js
daywiss: yea thats a great point but im not aware of any infrastructure within the voter dapp to alert the user to errors. if you have a suggestion on how to do this LMK

mrice32:
> > The changes are well implemented and overall its not much code which is awesome! I have a comment about how you set/get `ancillaryData` for commits and reveals.
> > Also, if this PR goes in will the voter dApp no longer be compatible with the current DVM? If so, we should discuss how to handle votes until we migrate the DVM. Perhaps we deploy a version of the web app to a new URL.
> 
> I think @mrice32 mentioned he has a plan for this release and noone seemed too concerned about this... Maybe matt can comment

@daywiss I think the simplest thing to do is just wait until the upgrade is over to merge this.

pemulis:
While we wait on the upgrade to merge this, should we resolve the merge conflicts?

daywiss:
> While we wait on the upgrade to merge this, should we resolve the merge conflicts?

Ill fix conflicts when we are ready to go with this pr. shouldnt take long

packages/core/scripts/local/RequestOraclePrice.js
```diff
@@ -61,13 +62,24 @@ async function run(deployedFinder, identifier, timeString) {
       await deployedFinder.getImplementationAddress(web3.utils.utf8ToHex(interfaceName.IdentifierWhitelist))
     );
     await deployedIdentifierWhitelist.addSupportedIdentifier(identifierInBytes);
-    const priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN);
+    let priceExists;
+
+    if (ancillaryDataInBytes) {
+      priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN, ancillaryDataInBytes);
+    } else {
+      priceExists = await deployedVoting.hasPrice(identifierInBytes, timeInBN);
```
daywiss, line 70: this script asks for new prices. im allowing this script to run either function to ensure when testing voter dapp both situations work ( and it does)

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -140,7 +141,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
         setPendingRequests(nonBlacklistedRequests);
       }
     }
-  }, [allPendingRequests, IDENTIFIER_BLACKLIST, showSpamRequests]);
+  }, [allPendingRequests, hexToUtf8, showSpamRequests]);
```
daywiss, line 144: modified this due to build warnings. most warnings are gone now

packages/voter-dapp/src/DesignatedVotingDeployment.js
```diff
@@ -53,7 +53,11 @@ function DesignatedVotingDeployment({ votingAccount }) {
       <strong>The two key voting contract is completely optional.</strong>
       <br />
       To deploy one, enter your cold key address. To learn more about the 2 key voting system see our{" "}
-      <a href="https://docs.umaproject.org/uma/oracle/voting_with_UMA_2-key_contract.html" target="_blank">
+      <a
+        href="https://docs.umaproject.org/uma/oracle/voting_with_UMA_2-key_contract.html"
+        rel="noopener noreferrer"
```
daywiss, line 58: this was added to stop compiler warning

packages/voter-dapp/src/ResolvedRequests.js
```diff
@@ -151,7 +152,7 @@ function ResolvedRequests({ votingAccount }) {
         setResolvedEvents(nonBlacklistedRequests);
       }
     }
-  }, [allResolvedEvents, IDENTIFIER_BLACKLIST, showSpamRequests]);
+  }, [hexToUtf8, allResolvedEvents, showSpamRequests]);
```
daywiss, line 155: again recommended change by compiler

packages/voter-dapp/package.json
```diff
@@ -9,7 +9,6 @@
     "@material-ui/styles": "^4.3.0",
     "@uma/common": "^2.0.1",
     "@uma/core": "^2.0.1",
-    "@uma/core-1-2-0": "npm:@uma/core@1.2.0",
```
nicholaspai, line 12: What did this binding do before?

packages/voter-dapp/run_tests.sh
```diff
@@ -39,18 +39,14 @@ echo -e '####################################################################\n'
 # This forces the voter dapp to deploy the _new_ voting contract with all the other contracts unchanged, which will
 # force it to test the old 2key contract with the new voting contract.
 yarn truffle compile
-ARTIFACT_PATH_1_2_0=$(node -p "require.resolve('@uma/core-1-2-0/build/contracts/Voting.json')")
```
nicholaspai, line 42: Great to see this complexity removed.

packages/voter-dapp/src/ActiveRequests.js
nicholaspai: Super strange that this `useEffect()` hook would depend on the value of `hexToUtf8`, which is a function?

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -417,7 +439,8 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
           [encryptionTimestamp]: {
             salt,
             price,
-            identifierPrecision
+            identifierPrecision,
+            ancillaryData
```
nicholaspai, line 443: Nice adding this here, not sure how you would have thought to do so without remembering about this button but +1

packages/voter-dapp/src/DesignatedVotingDeployment.js
nicholaspai: Why does the `rel` fix the compiler warning OOC?

packages/voter-dapp/src/ResolvedRequests.js
```diff
@@ -128,10 +129,10 @@ function ResolvedRequests({ votingAccount }) {
         return { filter: { roundId: showAllResolvedRequests ? undefined : indexRoundId }, fromBlock: 0 };
       }, [currentRoundId, showAllResolvedRequests])
     ) || [];
+
   useEffect(() => {
     setHasSpamRequests(false);
-
-    if (allResolvedEvents) {
+    if (allResolvedEvents && allResolvedEvents.length) {
```
nicholaspai, line 135: +1

nicholaspai: Can you not do `allResolvedEvents?.length`?

packages/voter-dapp/src/ResolvedRequests.js
nicholaspai: Makes no sense...

packages/voter-dapp/src/RetrieveRewards.js
```diff
@@ -225,7 +232,7 @@ function RetrieveRewards({ votingAccount }) {
         })
         .catch(err => console.error(`retrieveRewards.call failed:`, err));
     }
-  }, [votingAccount, rewardsTxn.oldestUnclaimedRound, rewardsTxn.toRetrieve]);
+  }, [web3.utils, drizzle.contracts.Voting, votingAccount, rewardsTxn.oldestUnclaimedRound, rewardsTxn.toRetrieve]);
```
nicholaspai, line 235: Also super strange to see `web3.utils` here

nicholaspai (approved):
Really clean code and its great to see a bunch of hack fixes going away!

packages/voter-dapp/package.json
daywiss: this was due to @mrice32 PR #2586. allowed us to load in new and old version of contract into app so we had both abis avaialble, or something like that

packages/voter-dapp/src/ActiveRequests.js
daywiss: @nicholaspai actually your run_tests script is very thorough and once i added requests with both ancillary and non ancillary data I had to do a lot more updates. pretty much every change in this pr was found thanks to that script 👍

packages/voter-dapp/src/ResolvedRequests.js
daywiss: interesting, yea i havent used this new syntax yet but its a good candidate for it. can update

packages/voter-dapp/src/RetrieveRewards.js
daywiss: yea... i was just following the compiler warnings here, i dont really understand why it wanted this

packages/common/package.json
```diff
@@ -42,7 +42,7 @@
     "bignumber.js": "^8.0.1",
     "chalk-pipe": "^3.0.0",
     "dotenv": "^6.2.0",
-    "eth-crypto": "^1.6.0",
+    "eth-crypto": "^1.7.0",
```
daywiss, line 45: not sure if this is strictly necessary, but running locally without this change doesnt seem to be an option

packages/voter-dapp/src/DesignatedVotingDeployment.js
daywiss: image

packages/voter-dapp/src/ResolvedRequests.js
daywiss: image

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -248,11 +248,23 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
...
     for (const ev of encryptedVoteEvents) {
-      encryptedVoteMap[toPriceRequestKey(ev.returnValues.identifier, ev.returnValues.time)] =
-        ev.returnValues.encryptedVote;
+      encryptedVoteMap[
+        // Ancillary data is undefined or null in these "returnValues" so we add a default string of "0x".
+        // If not added, the key will improperly index and the votes will not be found.
+        toPriceRequestKey(
+          ev.returnValues.identifier,
+          ev.returnValues.time,
+          ev.returnValues.ancillaryData || DEFAULT_ANCILLARY_DATA
```
mrice32, line 257: Is ev.returnValues.ancillaryData null when the bytes field is empty? I would expect it to be `0x`...

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -312,9 +324,13 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
...
           if (voteStatus.committedValue) {
-            return JSON.parse(
-              await decryptMessage(decryptionKeys[account][currentRoundId].privateKey, voteStatus.committedValue)
-            );
+            try {
+              return JSON.parse(
+                await decryptMessage(decryptionKeys[account][currentRoundId].privateKey, voteStatus.committedValue)
+              );
+            } catch (err) {
+              return "";
```
mrice32, line 332: If we're going to catch this error, can we return a string value that indicates the error? I think this should be distinguishable from no vote in some way. Something like "ERROR: DECRYPTION FAILED"? Would be ideal to show that in the "value" field in the dapp. Another (slightly less ideal) solution could be to print out the error using console.error so we can diagnose the issue in prod. Otherwise the error gets swallowed and there's no way to find the cause.

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -345,13 +361,15 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       .catch(err => console.log("snapshot error", err));
   };
 
+  // keccak256(abi.encodePacked(price, salt, msg.sender, time, ancillaryData, roundId, identifier))
```
mrice32, line 364: commented code?

packages/voter-dapp/src/ResolvedRequests.js
mrice32: The second part confuses me @daywiss -- can you shed a light on why the blacklist "doesn't re-render the component"?

mrice32 (approved):
LGTM!

packages/voter-dapp/src/ActiveRequests.js
daywiss: its not that it doesnt depend on it, but react warns that mutations to `IDENTIFIER_BLACKLIST` will not re-render component, so it isnt needed
image

packages/voter-dapp/src/ResolvedRequests.js
daywiss: im guessing its because this is state defined outside of a react component, so react cant react to changes to this variable. on top of that its a constant as well, so its kind of moot, it should never change.

packages/core/scripts/local/RequestOraclePrice.js
daywiss: depends on what you mean by does it work. I've updated the test script to create price requests that hit both of these branches, and it works as expected: price requests are made with and without ancillary data. Its possible that line 70 is still calling the ancillary version of the function though, which could be giving me similar behavior.

packages/voter-dapp/src/ActiveRequests.js
daywiss: yes, this should log to console.error since theres no way to really notifiy the user in this app

PR #2421, opened 2021-01-16
fix(scripts): fix conversion in historical price feed script

mrice32:
**Motivation**

Decimals are off in the getMedianHistoricalPrice script.

**Summary**

Ensures all price feeds are seeded with 18 as their decimals so the `fromWei` call at the end works as expected.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/local/getMedianHistoricalPrice.js
+++ b/packages/core/scripts/local/getMedianHistoricalPrice.js
@@ -17,7 +17,8 @@ const UMIP_PRECISION = {
   USDBTC: 8,
   USDETH: 5,
   BTCDOM: 2,
-  ALTDOM: 2
+  ALTDOM: 2,
+  BCHNBTC: 8
 };
 const DEFAULT_PRECISION = 5;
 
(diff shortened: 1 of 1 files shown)


packages/core/scripts/local/getMedianHistoricalPrice.js
```diff
@@ -43,6 +43,7 @@ async function getMedianHistoricalPrice(callback) {
     let priceFeedConfig = {
       // Empirically, Cryptowatch API only returns data up to ~4 days back.
       lookback: 345600,
+      priceFeedDecimals: 18, // Ensure all prices come out as 18-decimal denominated so the fromWei conversion works at the end.
```
nicholaspai, line 46: This is neccessary because of the `fromWei` below right?

nicholaspai (approved):
One question then LGTM

smb2796 (approved):
Thanks for fixing. LGTM

packages/core/scripts/local/getMedianHistoricalPrice.js
mrice32: Yep.

Issue #2419, opened 2021-01-15
Bot upgrades related to historical price querying and WDF


nicholaspai:
**Description**

- `STABLESPREAD` default price feeds current rely on a known flaky CryptoWatch pair: "USTUSDT on Bittrex". For example, querying the endpoint `https://api.cryptowat.ch/markets/uniswap-v2/ustusdt/ohlc` returns the latest OHLC timestamp as `1610727300`, which is 5 hours behind the current Unix timestamp `1610746046`. We need a strategy to handle missing historical price data. Additionally, we need to provide a way for aggregator price feeds (i.e. Medianizer and BasketSpreadPriceFeed) to add more details about which dependent price feed is the offending one. One idea is to add a callback function that can be called by clients, such as the SyntheticPegMonitor and the Disputer, to try to fetch more information about the missing historical price error.
- When the liquidator purposefully decides not to liquidate a position because it is prior to the slow withdrawal's `defenseActivationPercentage`, the Liquidator should NOT emit an error-level log. Instead, the liquidator should display that it is purposefully ignoring this liquidation.
- Change disputer log about missing historical price to an error level log instead of a warning. This could cause the system to lose funds if the disputer cannot dispute because it is missing a historical price.
- Ability to override the level of the log when the monitor sees a liquidation.

nicholaspai:
Fixed by #2426 and #2432

PR #2426, opened 2021-01-19
improve(price-feed): Add more error details to aggregator PriceFeeds about which sub-pricefeed is missing data

nicholaspai:
**Summary**

- Part of #2419 addressing historical price feed debug logs. The main issue this PR addresses is that the the aggregator pricefeeds like `BasketSpread` and `Medianizer` will return invalid for a price if ANY of their constituent price feeds throw an error, but usually only one price feed is failing. This PR adds logic to the aggregator pricefeeds to query their constituent pricefeeds for additional data.
- Fix `TransferSynthetic` and `TransferCollateral` scripts to work with ABI versioning logic

**Details**

Example new logs for the `BasketSpreadPriceFeed`:
Screen Shot 2021-01-21 at 12 36 17


packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -152,7 +152,7 @@ class UniswapPriceFeed extends PriceFeedInterface {
     this.events = events.slice(i);
 
     // Filter out events where price is null.
-    this.events = events.filter(e => e.price !== null);
+    this.events = this.events.filter(e => e.price !== null);
```
nicholaspai, line 155: Random bug I found, don't think this has really affected anything yet

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -112,6 +113,23 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
...
 
+  debugHistoricalData(time) {
+    let priceFeedErrorDetails = [];
+    this.allPriceFeeds.map(priceFeed => {
+      const hasPrice = priceFeed.getHistoricalPrice(time);
+      if (!hasPrice) {
+        if (priceFeed instanceof MedianizerPriceFeed) {
+          // MedianizerPriceFeed.debugHitoricalData returns an array of error details so we will
+          // concat its error log array.
+          priceFeedErrorDetails = priceFeedErrorDetails.concat(priceFeed.debugHistoricalData(time));
```
mrice32, line 124: Why not just have every price feed return a string as its own error message and higher-level price feeds just join the strings with separators so you don't have to detect the medianizer price feed specifically?

packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
```diff
@@ -36,6 +36,7 @@ class DominationFinancePriceFeed extends PriceFeedInterface {
 
     this.pair = pair;
     this.lookback = lookback;
+    this.uuid = `DominationFinance-${pair}`;
```
mrice32, line 39: very nice -- I really like the uuid concept.

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -49,6 +49,17 @@ class MedianizerPriceFeed extends PriceFeedInterface {
     }
   }
 
+  debugHistoricalData(time) {
+    let priceFeedErrorDetails = [];
+    this.priceFeeds.map(priceFeed => {
+      const hasPrice = priceFeed.getHistoricalPrice(time);
+      if (!hasPrice) {
+        priceFeedErrorDetails.push(`PriceFeed down with UUID: ${priceFeed.uuid}`);
```
mrice32, line 57: Might be useful to include the time that's unavailable in these since it may be particular times that are broken as opposed to the entire feed being down?

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -112,6 +113,23 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
     return this._getSpreadFromBasketPrices(experimentalPrices, baselinePrices, denominatorPrice);
   }
 
+  debugHistoricalData(time) {
```
mrice32, line 116: Just a random idea -- have you thought about returning an error message from the getHistoricalData method itself? That way, we wouldn't need to add a new method to the interface. The logic you have here would just be baked into the normal method. We could even say the second element is optional, so if you don't provide it, you're just failing to provide the caller with info about the error. Something like:
```js
return [price, errorString];
```

`errorString` could be null in the success case and price could be null in the error case?

An even more nuclear and invasive option is to begin allowing these price feeds to throw errors and relying on the calling code to catch those errors?

What do you think?

mrice32 (commented):
Looks great! A few comments and questions.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
daywiss, line 116: could you leave a comment on what this function is supposed to do

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -112,6 +113,23 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
...
+    let priceFeedErrorDetails = [];
+    this.allPriceFeeds.map(priceFeed => {
```
daywiss, line 118: this is just a nit, but you are using map but also pushing data into the priceFeedErrorDetails array. Consider changing this to forEach for side effects like this, or using reduce (removing need for priceFeedErrorDetails) since you are potentially changing length of array.

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -49,6 +49,17 @@ class MedianizerPriceFeed extends PriceFeedInterface {
     }
   }
 
+  debugHistoricalData(time) {
+    let priceFeedErrorDetails = [];
+    this.priceFeeds.map(priceFeed => {
```
daywiss, line 54: same as above, this is probably better as forEach or a reduce.

daywiss (approved):

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: This is a good idea, originally I wanted to shape the errors as an array of strings, but I don't think the array of strings vs concatenated string makes much difference to the end user.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: I think this would create more churn by changing the return interface of `getHistoricalPrice()`

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: However, I still like the idea of the higher level price feeds defining the error strings based on the lower level price feed's UUID's. Ultimately, this approach removes the need for each lower level price feed to implement `debugHistoricalData`, but the tradeoff is that the higher level pricefeeds should check if a lower level price feed has implemented `debugHistoricalData` first before it constructs an error string for the lower price feed. Look at [this commit](https://github.com/UMAprotocol/protocol/pull/2426/commits/be57391633074bd8769656769ded7740ade7f685) for example.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
nicholaspai: Good idea, I am not using `map` how you are supposed to :)

nicholaspai:
@mrice32 I responded to your comments in [this commit](https://github.com/UMAprotocol/protocol/pull/2426/commits/be57391633074bd8769656769ded7740ade7f685)

@daywiss I responded to your comments in [this commit](https://github.com/UMAprotocol/protocol/pull/2426/commits/b8319e8b773ad76364eade7cdd3b5d1833445423)

daywiss (approved):

mrice32 (commented):
I've looked through the recent updates, and I was wondering if you could explain what the requirements are for someone who's writing a price feed class? Is it different if that price feed aggregates vs just reports a price from an API? It seems that we're starting to create a distinction between those two types of interfaces and place different requirements on them? 

My general concern is that we're further muddying the water about what an implementer of the price feed class is expected to do. It seems that some are expected to report debug information and others are not. The reason I'm pushing on this is I think we should aim to make this interface as light and simple as possible for those trying to add price feeds since ultimately we'd like external devs to be able to do it.

nicholaspai:
> I've looked through the recent updates, and I was wondering if you could explain what the requirements are for someone who's writing a price feed class? Is it different if that price feed aggregates vs just reports a price from an API? It seems that we're starting to create a distinction between those two types of interfaces and place different requirements on them?
> 

If the price feed aggregates, then they should implement a `debugHistoricalData` method so that a client of the pricefeed can determine which of its aggregated price feeds errored out. If the price feed does not aggregate, then the pricefeed should implement a `uuid` so that an aggregator price feed can distinguish it. So yes there are now two classes of price feeds:
- `AggregatorPriceFeeds`: apply an aggregation function across more than one `BasePriceFeeds`
- `BasePriceFeeds`: does not rely on any other `PriceFeed`.
- 
> My general concern is that we're further muddying the water about what an implementer of the price feed class is expected to do. It seems that some are expected to report debug information and others are not. The reason I'm pushing on this is I think we should aim to make this interface as light and simple as possible for those trying to add price feeds since ultimately we'd like external devs to be able to do it.

I've considered extending `PriceFeedInterface` to create `AggregatorPriceFeedInterface` and `BasePriceFeedInterface` and every `PriceFeed` should implement one of the latter two. WDYT? Or is there a better way to address the core problem of this PR?

chrismaree (commented):
> > I've looked through the recent updates, and I was wondering if you could explain what the requirements are for someone who's writing a price feed class? Is it different if that price feed aggregates vs just reports a price from an API? It seems that we're starting to create a distinction between those two types of interfaces and place different requirements on them?
> 
> If the price feed aggregates, then they should implement a `debugHistoricalData` method so that a client of the pricefeed can determine which of its aggregated price feeds errored out. If the price feed does not aggregate, then the pricefeed should implement a `uuid` so that an aggregator price feed can distinguish it. So yes there are now two classes of price feeds:
> 
> * `AggregatorPriceFeeds`: apply an aggregation function across more than one `BasePriceFeeds`
> * `BasePriceFeeds`: does not rely on any other `PriceFeed`.
> 
> > My general concern is that we're further muddying the water about what an implementer of the price feed class is expected to do. It seems that some are expected to report debug information and others are not. The reason I'm pushing on this is I think we should aim to make this interface as light and simple as possible for those trying to add price feeds since ultimately we'd like external devs to be able to do it.
> 
> I've considered extending `PriceFeedInterface` to create `AggregatorPriceFeedInterface` and `BasePriceFeedInterface` and every `PriceFeed` should implement one of the latter two. WDYT? Or is there a better way to address the core problem of this PR?

I think this is a good structure, primarily because it will keep the `BasePriceFeeds` as simple as possible. Most products will fall into the `BasePriceFeeds` category and hopefully, we can re-use the aggregator logic in different feed contexts. 

Once this logic is finalized we should document it very clearly to explain to price feed writers what the conditions/requirements are for each kind.

packages/disputer/src/disputer.js
```diff
@@ -116,9 +116,18 @@ class Disputer {
         ? this.toBN(disputerOverridePrice)
         : this.priceFeed.getHistoricalPrice(liquidationTime);
       if (!price) {
+        // If missing historical price, check if there are further debug logs available. This is useful especially
+        // if the pricefeed aggregates pricefeeds, like a Medianizer or a BasketSpreadPriceFeed.
+        let priceFeedErrorDetails;
+        try {
+          priceFeedErrorDetails = this.priceFeed.debugHistoricalData(liquidationTime);
```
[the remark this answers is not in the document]
nicholaspai: ... yeah that'd be perfect actually

nicholaspai:
> > > I've looked through the recent updates, and I was wondering if you could explain what the requirements are for someone who's writing a price feed class? Is it different if that price feed aggregates vs just reports a price from an API? It seems that we're starting to create a distinction between those two types of interfaces and place different requirements on them?
> > 
> > 
> > If the price feed aggregates, then they should implement a `debugHistoricalData` method so that a client of the pricefeed can determine which of its aggregated price feeds errored out. If the price feed does not aggregate, then the pricefeed should implement a `uuid` so that an aggregator price feed can distinguish it. So yes there are now two classes of price feeds:
> > 
> > * `AggregatorPriceFeeds`: apply an aggregation function across more than one `BasePriceFeeds`
> > * `BasePriceFeeds`: does not rely on any other `PriceFeed`.
> > 
> > > My general concern is that we're further muddying the water about what an implementer of the price feed class is expected to do. It seems that some are expected to report debug information and others are not. The reason I'm pushing on this is I think we should aim to make this interface as light and simple as possible for those trying to add price feeds since ultimately we'd like external devs to be able to do it.
> > 
> > 
> > I've considered extending `PriceFeedInterface` to create `AggregatorPriceFeedInterface` and `BasePriceFeedInterface` and every `PriceFeed` should implement one of the latter two. WDYT? Or is there a better way to address the core problem of this PR?
> 
> I think this is a good structure, primarily because it will keep the `BasePriceFeeds` as simple as possible. Most products will fall into the `BasePriceFeeds` category and hopefully, we can re-use the aggregator logic in different feed contexts. Once this logic is finalized we should document it very clearly to explain to price feed writers what the conditions/requirements are for each kind.
> 
> The other option would be to follow more of what @mrice32 is proposing wherein each feed itself can provide debugging information, either via returning multiple values (as proposed [here](https://github.com/UMAprotocol/protocol/pull/2426#discussion_r561338566), via a separate log debugger/helper utility or through a logger.
> 
> We should consider taking a step back and looking at the implications of both techniques and how they generalize given we are wanting to simplify the overall process of adding price feeds.

Great points. On one hand, if we were designing the `PriceFeedInterface` from scratch then I think returning two values `[price, errorString]` would make the most sense, where either `price !== null || errorString !== null`. This would allow the caller to check `!price` and then print out the `errorString` if the `!price` is true.

However implementing this change now would introduce a lot of churn into the code base because every caller (in the tests and clients) would now need to check for the `[0]`th element of `getHistoricalPrice` and there are a lot of such calls. 

I think my PR introduces the least amount of churn but it does "muddy the water about what an implementer of the price feed class is expected to do". I think introducing two extensions of `PriceFeedInterface` could ease this concern: `Base` and `Aggregator` and we could use `Base` to improve documentation. Thoughts?

mrice32 (commented):
As mentioned IRL, after thinking about this more, I think keeping this interface simple and concise is important as it's becoming a focal point for external developers. I think this interface is already larger than it ideally should be, so I would prefer integrating the error messaging into the existing interface either by throwing or by returning objects of some type and having the error optionally added to those objects.

So one of the following would be my preferred approach:
```js
return { price, error };
```
or 
```js
return [price, error];
```
or
```js
assert(price, error);
return price;
```

mrice32:
@nicholaspai As far as creating two types of price feeds, aggregators and base, that feels a little bit like overkill just for the sake of error reporting, but definitely open to other arguments in favor of that approach.

nicholaspai:
> As mentioned IRL, after thinking about this more, I think keeping this interface simple and concise is important as it's becoming a focal point for external developers. I think this interface is already larger than it ideally should be, so I would prefer integrating the error messaging into the existing interface either by throwing or by returning objects of some type and having the error optionally added to those objects.
> 
> So one of the following would be my preferred approach:
> 
> ```js
> return { price, error };
> ```
> 
> or
> 
> ```js
> return [price, error];
> ```
> 
> or
> 
> ```js
> assert(price, error);
> return price;
> ```

Makes sense and I generally agree that creating two sub classes of PriceFeeds seems overkill for such a small feature. I think making the `getXPrice` methods return tuples will allow us to make clients such as the disputer more powerful in the future because they can better understand upstream pricefeed errors. Will plan to implement the `getHistoricalPrice = (time) => return [price, errorString]`  interface.

packages/financial-templates-lib/test/truffle/BalancerPriceFeed.js
```diff
@@ -157,6 +157,31 @@ contract("BalancerPriceFeed.js", function(accounts) {
     assert.equal(dexPriceFeed.getCurrentPrice(), null);
   });
 
+  it("One event within window, several before", async function() {
```
nicholaspai, line 160: Unrelated to the core of this PR, but I noticed that @mrice32 never added tests specific to #2465 . Do you agree that these tests correctly test changes introduced in that PR?

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
```diff
@@ -155,6 +155,51 @@ contract("UniswapPriceFeed.js", function(accounts) {
     assert.equal(uniswapPriceFeed.getCurrentPrice(), null);
   });
 
+  it("One event within window, several before", async function() {
```
nicholaspai, line 158: Same comment as [above](https://github.com/UMAprotocol/protocol/pull/2426/files#r567926321)

nicholaspai:
> > As mentioned IRL, after thinking about this more, I think keeping this interface simple and concise is important as it's becoming a focal point for external developers. I think this interface is already larger than it ideally should be, so I would prefer integrating the error messaging into the existing interface either by throwing or by returning objects of some type and having the error optionally added to those objects.
> > So one of the following would be my preferred approach:
> > ```js
> > return { price, error };
> > ```
> > 
> > 
> > or
> > ```js
> > return [price, error];
> > ```
> > 
> > 
> > or
> > ```js
> > assert(price, error);
> > return price;
> > ```
> 
> Makes sense and I generally agree that creating two sub classes of PriceFeeds seems overkill for such a small feature. I think making the `getXPrice` methods return tuples will allow us to make clients such as the disputer more powerful in the future because they can better understand upstream pricefeed errors. Will plan to implement the `getHistoricalPrice = (time) => return [price, errorString]` interface.

@mrice32 @chrismaree @daywiss as discussed IRL, this PR changes the `getHistoricalPrice` interface so that it throws an error if a historical price is not returned for some reason.

packages/monitors/src/SyntheticPegMonitor.js
```diff
@@ -289,17 +289,31 @@ class SyntheticPegMonitor {
...
+      try {
+        pricefeed.getHistoricalPrice(latestTime);
+      } catch (err) {
+        priceFeedErrorDetails = err.message;
+      }
       return {
         errorData: {
-          latestTime: latestTime ? latestTime : 0
+          latestTime: latestTime ? latestTime : 0,
+          priceFeedErrorDetails
```
mrice32, line 303: nit: thoughts on just getting _calculateHistoricalVolatility() to keep track of its last error and throwing it in the spot where it currently returns null?

mrice32 (commented):

packages/financial-templates-lib/test/truffle/BalancerPriceFeed.js
nicholaspai: Yeah these tests actually indicate the balancer price feed doesn't have the same problem that uniswap did (until your PR fixed it)!

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
nicholaspai: So is it correct to say that the latest data point before the window is used in the TWAP until the first data point within the window?

packages/disputer/src/disputer.js
```diff
@@ -112,31 +112,37 @@ class Disputer {
...
-          this.logger.debug({
+        try {
+          price = this.priceFeed.getHistoricalPrice(liquidationTime);
+        } catch (error) {
+          this.logger.warn({
             at: "Disputer",
-            message: "Detected a disputable liquidation",
-            price: price.toString(),
-            liquidation: JSON.stringify(liquidation)
+            message: "Cannot dispute: price feed returned invalid value",
```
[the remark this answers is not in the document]
nicholaspai: Yeah I think it should be an error since the liquidator sends error level logs for failures to liquidate

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js
```diff
@@ -36,16 +36,25 @@ class MedianizerPriceFeed extends PriceFeedInterface {
...
+      try {
+        return priceFeed.getHistoricalPrice(time, verbose);
+      } catch (err) {
+        errors.push(err);
+      }
+    });
 
-    if (this.computeMean) {
-      return this._computeMean(historicalPrices);
+    if (errors.length > 0 || historicalPrices.length === 0) {
```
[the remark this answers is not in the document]
nicholaspai: No, good point.

packages/monitors/src/SyntheticPegMonitor.js
nicholaspai: Hmm this is pretty interesting. So have _calculateHistoricalVolatility return successfully or throw, and have _checkPricrfeedVol report an error if it exists

packages/disputer/src/disputer.js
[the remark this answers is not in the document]
nicholaspai: Made this an error level

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
mrice32: Yep, that's correct.

mrice32 (approved):
LGTM!

PR #2429, opened 2021-01-19
fix(dec-20-audit): [H01] fix optimistic oracle fee burning logic when final fee == 0

mrice32:
**Motivation**

OpenZeppelin pointed out an issue in the [H01 fix](https://github.com/UMAprotocol/protocol/pull/2329). In the (admittedly unlikely) case that the final fee is 0, the burned portion of the bond gets trapped in the optimistic oracle contract and can never be retrieved.

**Summary**

This fixes the issue by conditioning the payout on the sum of the burned bond _and_ the final fee being > 0, which was the original intention of the code.

**Issue(s)**

N/A


nicholaspai (approved):
LGTM

PR #2430, opened 2021-01-19
fix(voter-dapp): fix QA script

mrice32:
**Motivation**

The QA script was slightly out of date when I tried to use it.

**Summary**

- The `ProposeAdmin.js` script had been moved and args updated. The path and args were updated at the callsite.
- The `ProposeAdmin.js` script was modified to only work on mainnet (it depended on a hardcoded address). A minor update was made to make it work on any network.
- The `ProposeAdmin.js` script was missing an arg in its propose call.
- The voter-dapp depends on artifacts pulled from npm, so the migrate call needs to be made from the voter-dapp directory rather than the core directory. Otherwise, the migration will write to the local core artifacts, which are different from the ones the voter dapp reads from.

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/mainnet/ProposeAdmin.js
+++ b/packages/core/scripts/mainnet/ProposeAdmin.js
@@ -31,9 +31,6 @@ async function propose(callback) {
      *********************************/
     const signingAccount = (await web3.eth.getAccounts())[0];
     console.group(`Proposer account: ${signingAccount}`);
-    if (signingAccount !== "0x2bAaA41d155ad8a4126184950B31F50A1513cE25") {
-      throw new Error("Cannot propose admin votes from this account");
-    }
     console.groupEnd();
 
     /** *******************************
@@ -83,6 +80,7 @@ async function propose(callback) {
         .propose([
           {
             to: identifierWhitelist.address,
+            value: 0,
             data: proposedTx
           }
         ])
--- a/packages/voter-dapp/run_tests.sh
+++ b/packages/voter-dapp/run_tests.sh
@@ -16,7 +16,6 @@ echo '# 0/16. Prerequisites                                              #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
 ## Will need to run some JS scripts from packages/core
-cd ../core
 
 # Prompt user to start Ganache in another window
 echo "- Have you started Ganache in a separate window at port 9545? Be sure to use the following mnemonic: \"candy maple cake sugar pudding cream honey rich smooth crumble sweet treat\" "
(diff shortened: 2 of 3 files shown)


packages/core/scripts/mainnet/ProposeAdmin.js
```diff
@@ -31,9 +31,6 @@ async function propose(callback) {
      *********************************/
     const signingAccount = (await web3.eth.getAccounts())[0];
     console.group(`Proposer account: ${signingAccount}`);
-    if (signingAccount !== "0x2bAaA41d155ad8a4126184950B31F50A1513cE25") {
```
mrice32, line 34: This was unnecessary, as the address is checked against the role explicitly later in the script.

packages/voter-dapp/run_tests.sh
```diff
@@ -16,7 +16,6 @@ echo '# 0/16. Prerequisites                                              #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
 ## Will need to run some JS scripts from packages/core
-cd ../core
```
mrice32, line 19: All truffle commands are now run from the voter-dapp directory, so they will use the same artifacts that the voter dapp reads from.

packages/voter-dapp/run_tests.sh
```diff
@@ -48,9 +47,9 @@ echo '#                                                                  #'
 echo '# 2/16. Submitting price requests                                  #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
-yarn run truffle exec ./scripts/local/RequestOraclePrice.js --network test --identifier USDETH-Commit+Reveal --time 1570000000
-yarn run truffle exec ./scripts/local/RequestOraclePrice.js --network test --identifier USDETH-2key-Commit+Reveal --time 1570000000
-yarn run truffle exec ./scripts/local/ProposeAdmin.js --network test
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier USDETH-Commit+Reveal --time 1570000000
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier USDETH-2key-Commit+Reveal --time 1570000000
+yarn run truffle exec ../core/scripts/mainnet/ProposeAdmin.js --network test --prod
```
mrice32, line 52: --prod flag was added so it would send the transation.

packages/voter-dapp/run_tests.sh
```diff
@@ -48,9 +47,9 @@ echo '#                                                                  #'
 echo '# 2/16. Submitting price requests                                  #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
-yarn run truffle exec ./scripts/local/RequestOraclePrice.js --network test --identifier USDETH-Commit+Reveal --time 1570000000
-yarn run truffle exec ./scripts/local/RequestOraclePrice.js --network test --identifier USDETH-2key-Commit+Reveal --time 1570000000
-yarn run truffle exec ./scripts/local/ProposeAdmin.js --network test
+yarn run truffle exec ../core/scripts/local/RequestOraclePrice.js --network test --identifier USDETH-Commit+Reveal --time 1570000000
```
mrice32, line 50: This directory change was necessary because we're now running these commands from the voter-dapp directory. This is a little hacky as packages shouldn't directly use each other's js files. We should remove this down the road.

daywiss (approved):
looks good thanks

packages/core/scripts/mainnet/ProposeAdmin.js
nicholaspai: @daywiss fixes this in another PR but that one might be blocked for a bit longer, +1

nicholaspai (approved):
Nice lgtm!

PR #2432, opened 2021-01-19
improve(liquidator): Send log when WDF is purposefully not activated

nicholaspai:

**Motivation**

Problem mentioned as part of #2419

Previously, this error log is being emitted even if the WDF is not being activated on purpose (i.e there is no error):
Screen Shot 2021-01-19 at 13 06 30

**Summary**

Liquidator queries `LiquidatorStrategy` to determine if it should emit a specialized log about the WDF not being activated yet.

### DIFF
--- a/packages/liquidator/index.js
+++ b/packages/liquidator/index.js
@@ -126,6 +126,13 @@ async function run({
       withdrawLiveness
     };
 
+    // Add block window into `liquidatorConfig`
+    let overriddenLiquidatorConfig = {
+      ...liquidatorConfig,
+      startingBlock,
+      endingBlock
+    };
+
     // If pollingDelay === 0 then the bot is running in serverless mode and should send a `debug` level log.
     // Else, if running in loop mode (pollingDelay != 0), then it should send a `info` level log.
     logger[pollingDelay === 0 ? "debug" : "info"]({
@@ -136,7 +143,7 @@ async function run({
       errorRetries,
       errorRetriesTimeout,
       priceFeedConfig,
-      liquidatorConfig,
+      liquidatorConfig: overriddenLiquidatorConfig,
       liquidatorOverridePrice
     });
 
@@ -185,9 +192,7 @@ async function run({
       priceFeed,
       account: accounts[0],
       empProps,
-      liquidatorConfig,
-      startingBlock,
-      endingBlock
+      liquidatorConfig: overriddenLiquidatorConfig
     });
 
     logger.debug({
--- a/packages/liquidator/index.js
+++ b/packages/liquidator/index.js
@@ -34,6 +34,8 @@ const { getWeb3 } = require("@uma/common");
  * @param {Object} priceFeedConfig Configuration to construct the price feed object.
  * @param {Object} [liquidatorConfig] Configuration to construct the liquidator.
  * @param {String} [liquidatorOverridePrice] Optional String representing a Wei number to override the liquidator price feed.
+ * @param {Number} [startingBlock] Earliest block to query for contract events that the bot will log about.
+ * @param {Number} [endingBlock] Latest block to query for contract events that the bot will log about.
  * @return None or throws an Error.
  */
 async function run({
@@ -46,7 +48,9 @@ async function run({
   errorRetriesTimeout,
   priceFeedConfig,
   liquidatorConfig,
-  liquidatorOverridePrice
+  liquidatorOverridePrice,
+  startingBlock,
+  endingBlock
 }) {
   try {
     const { toBN } = web3.utils;
@@ -181,7 +185,9 @@ async function run({
       priceFeed,
       account: accounts[0],
       empProps,
-      liquidatorConfig
+      liquidatorConfig,
+      startingBlock,
+      endingBlock
     });
 
     logger.debug({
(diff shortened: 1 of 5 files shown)


packages/liquidator/src/liquidator.js
```diff
@@ -265,9 +265,28 @@ class Liquidator {
...
+        if (
+          this.defenseActivationPercent &&
+          !this.liquidationStrategy.utils.passedDefenseActivationPercent({ position, currentBlockTime })
+        ) {
+          this.logger.info({
+            at: "Liquidator",
+            message: "Withdrawal liveness has not passed WDF activation threshold, skipping✋",
+            sponsor: position.sponsor,
+            inputPrice: scaledPrice.toString(),
+            position: position,
```
chrismaree, line 282:
```suggestion
            position,
```

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -221,11 +229,7 @@ module.exports = (
     )
       return false;
     // liveness timer on withdraw has not passed time
-    if (
-      withdrawProgressPercent(withdrawLiveness, position.withdrawalRequestPassTimestamp, currentBlockTime) <
-      parseFloat(defenseActivationPercent)
-    )
-      return false;
+    if (!passedDefenseActivationPercent({ position, currentBlockTime })) return false;
```
chrismaree, line 232: can you not just return `passedDefenseActivationPercent({ position, currentBlockTime })` at this point?

packages/liquidator/src/liquidator.js
```diff
@@ -265,9 +265,28 @@ class Liquidator {
...
       // This gets logged as an event, see constructor
       if (!liquidationArgs) {
+        // If WDF is active but liveness hasn't passed activation %, then send customized log:
+        if (
+          this.defenseActivationPercent &&
+          !this.liquidationStrategy.utils.passedDefenseActivationPercent({ position, currentBlockTime })
+        ) {
+          this.logger.info({
+            at: "Liquidator",
+            message: "Withdrawal liveness has not passed WDF activation threshold, skipping✋",
```
chrismaree, line 279: shall we use a new emoji? this one is used in other places.

chrismaree (approved):
Looks good! as discussed IRL we need to think about if we always want to be sending out these `info` messages. We _really_ just need to know when the initial WDF liquidation is sent out and if the bot is running low on funds/cant extend a liquidation. The way this is set up is we'll see this info message every 5 mins, which I think is fine, but might be a bit noisy.

nicholaspai:
> Looks good! as discussed IRL we need to think about if we always want to be sending out these `info` messages. We _really_ just need to know when the initial WDF liquidation is sent out and if the bot is running low on funds/cant extend a liquidation. The way this is set up is we'll see this info message every 5 mins, which I think is fine, but might be a bit noisy.

Good point, given the constraint that this bot runs serverlessly this is a bit tricky. Thoughts on changing this to a debug log vs keeping it as an info? I'm not sure what we'd prefer as a team, but I kind of like the idea of having more information than not.

packages/liquidator/src/liquidator.js
```diff
@@ -265,9 +265,28 @@ class Liquidator {
...
+      // we couldnt liquidate, this typically would only happen if our balance is 0 or the withdrawal liveness
+      // has not passed the WDF's activation threshold.
       // This gets logged as an event, see constructor
       if (!liquidationArgs) {
+        // If WDF is active but liveness hasn't passed activation %, then send customized log:
+        if (
+          this.defenseActivationPercent &&
+          !this.liquidationStrategy.utils.passedDefenseActivationPercent({ position, currentBlockTime })
+        ) {
+          this.logger.info({
```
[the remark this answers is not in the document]
nicholaspai: INFO level logs should not set of PagerDuty

packages/liquidator/src/liquidator.js
nicholaspai: key comment, will do

packages/liquidator/src/liquidationStrategy.js
nicholaspai: true!

daywiss (approved):

packages/liquidator/index.js
```diff
@@ -46,7 +48,9 @@ async function run({
   errorRetriesTimeout,
   priceFeedConfig,
   liquidatorConfig,
-  liquidatorOverridePrice
+  liquidatorOverridePrice,
+  startingBlock,
```
chrismaree, line 52: as discussed on slack, you could consider putting these within the `liquidatorConfig`.

nicholaspai: I'm going to pack `startingBlock` and `endingBlock` into `liquidatorConfig` before creating the `Liquidator.js` object, but I still want to read those vars from the environment because the serverless hub sets them conveniently there

nicholaspai: See [this commit](https://github.com/UMAprotocol/protocol/pull/2432/commits/2717599ae0fc024fd0437a07c8f4712dc81f445c) for the proposed changes in response to your comment.

daywiss (approved):
Looks good to me, i just have questions about how starting and ending block gets set. Is this something that gets persisted in gcp somehow and set dynamically in cloud run or what?

packages/liquidator/index.js
```diff
@@ -122,6 +126,13 @@ async function run({
       withdrawLiveness
     };
 
+    // Add block window into `liquidatorConfig`
+    let overriddenLiquidatorConfig = {
```
chrismaree, line 130: does this need to be a new variable? could you not just re-asign `liquidatorConfig`?
```
liquidatorConfig = {...liquidatorConfig, startingBlock, endingBlock };
```

The only reason I bring this up is that it is more consistent with how I was trying to override the liquidator config in the perpetual generalization.

chrismaree (approved):
This looks great! thanks for refining this & adding all the nits.

nicholaspai:
> Looks good to me, i just have questions about how starting and ending block gets set. Is this something that gets persisted in gcp somehow and set dynamically in cloud run or what?

Yep look at how the Hub sets the environment [here](https://github.com/UMAprotocol/protocol/blob/master/packages/serverless-orchestration/src/ServerlessHub.js#L90)

packages/liquidator/index.js
nicholaspai: Good call, I originally had it set this way but ran into an error because I redeclared `liquidatorConfig` with a `let` statement.

packages/liquidator/src/liquidator.js
```diff
@@ -265,9 +280,40 @@ class Liquidator {
...
+        ) {
+          // If `startingBlock` and `endingBlock` are specified and the requested withdrawal was within the
+          // block window, then upgrade the log level:
+          let logLevel = "debug";
+          if (!isNaN(this.startingBlock) && !isNaN(this.endingBlock)) {
+            const blockWindowHasRequestedWithdraw = await this.empContract.getPastEvents("RequestWithdrawal", {
+              fromBlock: this.startingBlock,
+              toBlock: this.endingBlock
+            });
+            if (blockWindowHasRequestedWithdraw) {
```
chrismaree, line 300: tiny nit: This could be done more cleanly using a simply ternary operator.
```
const logLevel = blockWindowHasRequestedWithdraw ? "info" : "debug";
```

Feel free to ignore as well, this is more personal preference.

PR #2443, opened 2021-01-20
improve(core): rename umip 15 scripts to make them general-purpose

### LINKED ISSUE #2439: Write and test deployment scripts to propose DVM upgrade (voting.sol) with ancillary data fields
Created by Matt Rice via monday.com integration. 🎉

mrice32:
**Motivation**

UMIP-15 scripts can be reused for upgrading the Voting contract, generally. To signify this, comments and naming were changed throughout. Otherwise, the scripts seem to work as expected with the new contract.

**Summary**

Almost entirely naming/cosmetic changes to signify that these scripts and Solidity code do not just apply to UMIP-15.

The only minor change that I made was to symlink the Vote simulation script to the one that the other umip folders use.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Note: I ran the code on a mainnet fork to verify that it ran correctly.

**Issue(s)**

Fixes #2439


packages/core/scripts/voting-upgrade-umip/1_Propose.js
```diff
@@ -11,7 +11,7 @@ const Registry = artifacts.require("Registry");
 const Voting = artifacts.require("Voting");
 const VotingToken = artifacts.require("VotingToken");
 const Governor = artifacts.require("Governor");
-const Umip15Upgrader = artifacts.require("Umip15Upgrader");
+const VotingUpgrader = artifacts.require("VotingUpgrader");
```
chrismaree, line 14: here we are using `artifacts.require`. however, in PR https://github.com/UMAprotocol/protocol/pull/2444/files we are using `getTruffleContract`. should we consider changing this to use the same syntax everywhere?

chrismaree (approved):
Looks good! I had a nit on how we are importing things but otherwise, I think this is good to go

nicholaspai (approved):
LGTM

packages/core/scripts/voting-upgrade-umip/1_Propose.js
mrice32: Good point. Fixed.

PR #2444, opened 2021-01-20
improve(core): add optimistic oracle upgrade scripts

### LINKED ISSUE #2440: Write and test deployment scripts to propose optimistic oracle upgrade
Created by Matt Rice via monday.com integration. 🎉

mrice32:
**Motivation**

Scripts are needed to add or upgrade the optimistic oracle in the current DVM.

**Summary**

This PR adds fairly straightforward scripts to add/upgrade the optimistic oracle.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Note: I ran and verified the upgrade on a mainnet fork.

**Issue(s)**

Fixes #2440


packages/core/scripts/optimistic-oracle-umip/1_Propose.js
```diff
@@ -0,0 +1,113 @@
...
+  // 1. Temporarily add the Governor as a contract creator.
+  const addGovernorToRegistryTx = registry.contract.methods
+    .addMember(RegistryRolesEnum.CONTRACT_CREATOR, governor.address)
+    .encodeABI();
+
+  console.log("addGovernorToRegistryTx", addGovernorToRegistryTx);
+
+  // 2. Register the OptimisticOracle as a verified contract.
+  const registerOptimisticOracleTx = registry.contract.methods
+    .registerContract([], optimisticOracle.address)
```
chrismaree, line 50: this does not require any `parties`, as the first param in `registerContract`, right?

packages/core/scripts/optimistic-oracle-umip/1_Propose.js
```diff
@@ -0,0 +1,113 @@
+// This script generates and submits an identifier-add upgrade transaction to the DVM. It can be run on a local ganache
+// fork of the main net or can be run directly on the main net to execute the upgrade transactions.
+// To run this on the localhost first fork main net into Ganache with the proposerWallet unlocked as follows:
+// ganache-cli --fork https://mainnet.infura.io/v3/d70106f59aef456c9e5bfbb0c2cc7164 --unlock 0x2bAaA41d155ad8a4126184950B31F50A1513cE25
+// Then execute the script as: truffle exec ./scripts/identifier-umip/1_Propose.js --network mainnet-fork --identifier USDETH --identifier ETHBTC from core
```
chrismaree, line 5: nit: this comment is wrong. This script does not use `identifier`

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
```diff
@@ -0,0 +1,60 @@
+// This script verify that the upgrade was executed correctly.
+// It can be run on mainnet after the upgrade is completed or on the local Ganache mainnet fork to validate the
+// execution of the previous two scripts. This script does not need any wallets unlocked and does not make any on-chain
+// state changes. It can be run as:
+// truffle exec ./scripts/identifier-umip/3_Verify.js --network mainnet-fork --identifier USDETH --identifier ETHBTC
```
chrismaree, line 5: same nit about this comment being wrong

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
```diff
@@ -0,0 +1,60 @@
...
+  const registry = await Registry.deployed();
+  const governor = await Governor.deployed();
+  const finder = await Finder.deployed();
+
+  console.log("Verifying that Governor doesn't hold the creator role...");
+  assert(!(await registry.holdsRole(RegistryRolesEnum.CONTRACT_CREATOR, governor.address)));
+  console.log("Verified!");
+
+  console.log("Verifying that the OptimisticOracle is registered with the Registry...");
+  assert(await registry.isContractRegistered(argv.deployedAddress));
```
chrismaree, line 35: this should just be assert. it should be `assert.true(...)` or `assert.equal(...)`.

packages/core/scripts/optimistic-oracle-umip/1_Propose.js
```diff
@@ -0,0 +1,113 @@
...
+
+// Use the same ABI's as deployed contracts:
+const { getTruffleContract } = require("../../index");
+const Governor = getTruffleContract("Governor", web3, "1.1.0");
+const Finder = getTruffleContract("Finder", web3, "1.1.0");
+const Registry = getTruffleContract("Registry", web3, "1.1.0");
+const OptimisticOracle = getTruffleContract("OptimisticOracle", web3);
```
chrismaree, line 12: do you want to explicitly define this as "latest" to be really clear that it's the newest code possible?
```suggestion
const OptimisticOracle = getTruffleContract("OptimisticOracle", web3, "latest");
```

packages/core/scripts/optimistic-oracle-umip/1_Propose.js
mrice32: Yep, exactly.

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
mrice32: Does `assert.true` exist for the nodejs assert library? I only used `assert` because I wasn't familiar with any boolean checker in native nodejs.

packages/core/scripts/optimistic-oracle-umip/1_Propose.js
mrice32: Done.

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
mrice32: Good point. Done.

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
daywiss: yea theres no `assert.true`, but `assert` and `assert.ok` are equivalent.

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
```diff
@@ -0,0 +1,60 @@
+// This script verify that the upgrade was executed correctly.
+// It can be run on mainnet after the upgrade is completed or on the local Ganache mainnet fork to validate the
+// execution of the previous two scripts. This script does not need any wallets unlocked and does not make any on-chain
+// state changes. It can be run as:
+// yarn truffle exec ./scripts/optimistic-oracle-umip/3_Verify.js --network mainnet-fork --deployedAddress 0xOPTIMISTIC_ORACLE_ADDRESS
+
+const assert = require("assert").strict;
```
daywiss, line 7: +1 have not seen this befor

daywiss (approved):
lgtm

packages/core/scripts/optimistic-oracle-umip/3_Verify.js
chrismaree: sorry, I meant `assert.isTrue`

mrice32: @chrismaree Ah, I think there's minor miscommunication. The issue is that the docs you reference are from the chai assert library that is _generally_ meant to be an extension of asserts for test environments (it is automatically injected into truffle tests along with mocha). We _could_ import chai in non-test scripts, but we haven't done that in the past as far as I know. The native nodejs assert doesn't have an `isTrue` (see docs here: https://nodejs.org/docs/latest-v12.x/api/assert.html), so we would either need to use `assert()` or `assert.ok()`.

chrismaree: aha. I see. `assert.isTrue` is explicitly used within unit testing, and not the general node Assert context. Fair enough! In that case, assert is the way to go :)

chrismaree (approved):
Looks good to me! thanks for appeasing me with my comments :)

PR #2449, opened 2021-01-21
feat(yarn): enable concurrency between test threads

### LINKED ISSUE #2452: Add concurrency to CI
**Why**

CI can be slow. Adding concurrency between package tests should speed it up.

**How**

Add concurrency and speed up tests & ensure tests still work.

**Resolution**

Concurrency is added and tests work.

**Difficulty Score [1-10]**

- 5: ~4 hours

chrismaree:
**Motivation**

Slow CI and speed up the tests. Issue: 2452

**Summary**

This PR changes how the unit tests are run within the repo by executing tests between packages in parallel. On average this drastically decreases the speed of unit tests.

**Details**

The changes in this PR are 4 fold:
1) add a new `test-concurrency` entry point in the main package.json that runs some packages tests in parallel.
2) add the `concurrency` flag to yarn to run tests between packages in parallel.
3) increase the size of the machine in circleCI to accommodate change (1). we now use a large VM which has 8gb of ram and 4 cores.
4) update all packages to ensure that hardhat uses `--network hardhat`.

**Tests speedup**
How long things take to run in CI can vary a lot based on how many other jobs are running at the same time. On average, the test job can take over an hour. The screenshot below shows the average over the past 90 days (~200 runs):
image

Here we can see that the `test` job takes 1h 13m 28s for the p95.

However, there are cases where it can be _much_ faster than this average. The screenshot below shows a job running all on its own that finished in a much shorter time:

image

After the change of this PR the test job now takes **less than 20 mins** to run ALL tests over all packages, as can be seen in the screenshot below:
image

This is a pretty big improvement! 

We'll need to run this in production for a bit to really see how much of a speedup this has achieved.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

close #2452


daywiss (approved):
awesome work

package.json
```diff
@@ -9,7 +9,9 @@
     "lint-fix": "npm run eslint -- --fix && npm run prettier -- --write",
     "eslint": "eslint './**/*.js'",
     "prettier": "prettier './**/*.js' './**/*.sol' './**/*.md'",
-    "test": "lerna run --stream --concurrency=1 test",
+    "test": "yarn test-no-concurrent && yarn test-concurrent",
```
mrice32, line 12: Do you think it would make sense to have the default "test" run everything serially and have a concurrent option that runs some of the modules concurrently?

Not strongly opinionated, it could just be nice to not have our default test command require substantial system resources. The parallel logs can also be somewhat confusing.

Thoughts?

package.json
```diff
@@ -9,7 +9,9 @@
...
+    "test-no-concurrent": "lerna run --stream --concurrency=1 test --scope @uma/serverless-orchestration --scope @uma/affiliates",
+    "test-concurrent": "lerna run --stream --concurrency=4 test --ignore @uma/serverless-orchestration --ignore @uma/affiliates",
```
mrice32, line 14: Why can these not be run in parallel?

mrice32 (commented):
Looks great! Just a few questions.

mrice32 (commented):
Is it fair to say that the speedup is ~28 minutes -> ~19.5 minutes?

package.json
chrismaree: ye, this is a good point. the parallel logs are basically impossible to read. I'll change this to have a separate entry point that CI uses.

package.json
chrismaree: These two packages, in particular, rely on some block number/time manipulation that seems to break within the concurrent context. I still need to test it out more but for now, this seems to in a stable way.

chrismaree:
> Is it fair to say that the speedup is ~28 minutes -> ~19.5 minutes?

based on my cherry-picked data, ye that's fair to say. I'd say on average though it is actually _more_ of a speedup than just this.

package.json
mrice32: Makes sense.

packages/core/package.json
```diff
@@ -53,7 +53,7 @@
   "scripts": {
     "test": "yarn mocha-test && yarn hardhat-test && yarn truffle-test",
     "truffle-test": "truffle test ./truffle-test/*",
-    "hardhat-test": "node --max-old-space-size=1024 $(yarn bin hardhat) test",
+    "hardhat-test": "hardhat test --network hardhat",
```
mrice32, line 56: nice

packages/monitors/package.json
```diff
@@ -30,7 +30,7 @@
   ],
   "bin": "index.js",
   "scripts": {
-    "test": "hardhat test"
+    "test": "hardhat test --network hardhat"
```
mrice32, line 33: Just curious -- why is the network required here but not in the others?

mrice32 (approved):
LGTM! My question is purely out of curiosity, so please feel free to merge.

packages/monitors/package.json
chrismaree: I added it on all the `hardhat test` parts to ensure that we are using the hardhat network and there is no confusion about the environment that hardhat test is running within.

PR #2450, opened 2021-01-22
feat(m1): update dependencies to run on apple m1 chip

pemulis:

**Motivation**

There is an Intel-specific dependency that breaks compatibility with computers running the new M1 chip from Apple.

**Summary**

Replaces the Intel-specific dependency with one that works with alternate instruction sets.

Note: There are still some failing tests if you run this on a computer running an Apple M1 chip, but it doesn't break existing tests on regular Intel chipsets.

**Details**

This also updates the ganache-cli version to be compatible with node v15. The reason for this is because node v15 works natively with Apple's M1 chip, while earlier versions do not.

Interestingly, all tests passed locally after building and running with node v15 on an older MacBook. That means that this PR would appear to make the project compatible with the latest version of node, although additional changes would be needed to update our cloud infrastructure to run node v15 instead of v12, and we would need to update the docs once we're sure this is working, since devs will not need to run it in v12 anymore.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
-    "ganache-cli": "^6.10.2",
+    "ganache-cli": "^6.12.2",
     "husky": "^4.2.3",
     "lerna": "^3.22.1",
     "lint-staged": "^10.1.3",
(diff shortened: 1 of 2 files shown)


mrice32 (approved):
LGTM!

chrismaree (approved):
This is pretty awesome!

if we can _actually_ bump to node 15 you could try updating the node version within CI here: https://github.com/UMAprotocol/protocol/blob/a52b275f8fceaf61e3b07581b275df6bb7b569e3/.circleci/config.yml#L97

as well as updating the node version within the docker container here: https://github.com/UMAprotocol/protocol/blob/a52b275f8fceaf61e3b07581b275df6bb7b569e3/Dockerfile#L22

If it works as expected then it should pass CI. 

If it does not (for example one or two tests don't work) it's still measurably better as we can support M1 chips but still recommend that users stick to node 12.

PR #2451, opened 2021-01-22
ci(nightly): fix nightly by removing privileged context

mrice32:
**Motivation**

coverage currently requires a privileged context to run, which causes the nightly build to fail because it cannot be run with privileged permissions: see https://circleci.canny.io/cloud-feature-requests/p/allow-restricted-contexts-within-scheduled-workflows.

**Summary**

Coverage seems to only require the coveralls token to run correctly, so I figure it isn't too dangerous to provide that variable to all ci jobs rather than making it privileged (it's not a paid service and the token can be regenerated if compromised). This update has been made in the circleci config.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -233,6 +233,5 @@ workflows:
     jobs:
       - checkout_and_install
       - coverage:
-          context: api_keys
           requires:
             - checkout_and_install


chrismaree (approved):
LGTM!

PR #2455, opened 2021-01-22
improve(node-lts): update circleci and docker config to use the latest lts version of nodejs

pemulis:

**Motivation**

Upgrades to the latest long-term support version of nodejs in our CircleCI and Docker configuration.

**Summary**

This would let us use new node features.

**Details**

We need to see if this passes continuous integration in the pull request process.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   checkout_and_install:
     docker:
-      - image: circleci/node:15
+      - image: circleci/node:lts
     working_directory: ~/protocol
     steps:
       - checkout
@@ -19,7 +19,7 @@ jobs:
             - ~/.ssh
   build:
     docker:
-      - image: circleci/node:15
+      - image: circleci/node:lts
     working_directory: ~/protocol
     steps:
       - restore_cache:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -2,7 +2,7 @@ version: 2
 jobs:
   checkout_and_install:
     docker:
-      - image: circleci/node:12
+      - image: circleci/node:15
```
mrice32, line 5: We should probably use `circleci/node:lts` (which is 14). LTS releases are [recommended](https://nodejs.org/en/about/releases/) for production applications.

mrice32 (commented):
Looks awesome! Just one comment about which version we should use. Also, is there any reference to the node version in the README or docs that we need to update?

.circleci/config.yml
pemulis: I'll swap this for `lts` and see if it builds! Right now this is an exploratory PR to see what happens in the CI process. I'll need to update the docs in at least one place (maybe more) if it passes muster in CI and this can actually be merged.

pemulis:
> Looks awesome! Just one comment about which version we should use. Also, is there any reference to the node version in the README or docs that we need to update?

I'll need to update some docs, but I want to see if it passes CI first.

mrice32 (approved):
LGTM!

chrismaree (approved):

Issue #2461, opened 2021-01-26
getHistoricalPricePeriods output differs between price feed classes


daywiss:
**Description**

Discovered during routine dev mining calculation using non medianizer price feeds. It turns out both dom finance and cryptowatch feeds return data in the format of an object like: 
```
[
{
  openTime, closeTime, openPrice, close Price
}
]
```
While the medianizer price feed returns data in the form of:
```
[
  [timestamp, price]
]
```

The data really should output the same for these calls so that we can process result without context of the class type.

**Reproduction**
Call `getHistoricalPricePeriods` from a medianizer price feed and from cryptowatch feed and look at results. Shape of data will be different.

**Expected Behavior**

Data should be returned in a consistent format between all classes. I suggest to copy the medianizer style in the form of
```
[
[ closeTime, closePrice]
]
```

**Additional Context**

Medinizer price feed uses cryptowatch and dom finance calls to `getHistoricalPricePeriods` under the hood, so it actually expects this behavior. Changing those classes will also require an update to medianizer so that it can properly do its calculations. This ultimately may be too much churn, so use discretion with solution.

PR #2463, opened 2021-01-26
fix(affiliates): switch between price feed output and also convert time to ms

### LINKED ISSUE #2462: Developer mining script is unable to get price history from non medianizer feeds
**Description**

see #2461 which outlines why devmining script breaks on non medianized price feeds. This will require a quick fix in order to run rewards for the week since dom finance has been added and it does not by default load a medianized feed. 

Additionally it turns out medianized prices are timestamped in seconds, so these need to be converted to MS for dev mining to be able to look up prices properly.

**Reproduction**

Run dev mining with a dom finance address such as `0xeFA41F506EAA5c24666d4eE40888bA18FA60a1c7` and see that the app crashes with strange error.

**Expected Behavior**

Dev mining needs to detect output from price feed classes and determine how to properly parse them for now, until these outputs can be standardized. Dev mining also needs to convert timestamps from seconds to ms.

daywiss:
**Motivation**

#2462 

**Summary**
This is  a quick fix that needed to be added in order to run dev mining for this week due to the inclusion of dom finance.

- When querying synthetic price the dev mining processor will switch between different output types by detecting if its an 
array or not
- convert synthetic time from S to MS.
- do not check valid addresses for payout addresses so that we can support running script before partner address is given

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2462

### DIFF
--- a/packages/affiliates/libs/affiliates/devmining.js
+++ b/packages/affiliates/libs/affiliates/devmining.js
@@ -99,8 +99,9 @@ module.exports = ({ queries, empAbi, coingecko, synthPrices, firstEmpDate }) => 
   // Gets the value of synthetics in collateral currency.
   async function getSyntheticPriceHistory(address, start, end) {
     const result = await synthPrices.getHistoricSynthPrices(address, start, end);
-    // Turns out getHistoricPriceFeed differs between price feed classes. This needs to be fixed
-    // but for now we will detect it and switch between decodings.
+    // Turns out getHistoricPriceFeed differs between price feed classes.
+    // This needs to be fixed but for now we will detect it and switch between decodings.
+    // TODO: Fix this behavior, see issue #2461
     const fixedPrices = result.map(price => {
       if (isMedianizerPrice(price)) return decodeMedianizerPrice(price);
       return decodeOtherPrice(price);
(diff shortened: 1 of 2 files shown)


chrismaree (approved):
Looks good! nice to be able to not know the payout address. id argue we want this to be uniform but I understand there is a business need for this functionality.

packages/affiliates/libs/affiliates/devmining.js
```diff
@@ -87,10 +87,25 @@ module.exports = ({ queries, empAbi, coingecko, synthPrices, firstEmpDate }) =>
...
   async function getSyntheticPriceHistory(address, start, end) {
     const result = await synthPrices.getHistoricSynthPrices(address, start, end);
-    return Prices(result);
+    // Turns out getHistoricPriceFeed differs between price feed classes. This needs to be fixed
+    // but for now we will detect it and switch between decodings.
+    const fixedPrices = result.map(price => {
+      if (isMedianizerPrice(price)) return decodeMedianizerPrice(price);
+      return decodeOtherPrice(price);
+    });
+    return Prices(fixedPrices);
```
mrice32, line 108: Yeah, we should definitely fix the underlying price feed. Can you add a TODO?

mrice32 (approved):

packages/affiliates/libs/affiliates/devmining.js
daywiss: for reference see #2461. ive also added a todo

PR #2465, opened 2021-01-26
fix(uniswap-pf): fixes a low volume bug in the uniswap price feed

mrice32:
**Motivation**

The Uniswap price feed has a bug where few trades (last trade long before the lookback window) will cause the uniswap price feed to just take the spot price rather than the TWAP.

**Summary**

Reimplemented the update function to streamline and simplify the implementation by condensing most of the complex looping logic into a single looping block.

To fix the issue, this adds an exponential lookback to quickly make larger and larger historical queries to _always_ find the correct most recent sync event before the window.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Note: this could probably use some specific tests for low volume pools. Will add.

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
@@ -100,17 +100,17 @@ class UniswapPriceFeed extends PriceFeedInterface {
 
     let events = []; // Caches sorted events (to keep subsequent event queries as small as possible).
     let blocks = {}; // Caches blocks (so we don't have to re-query timestamps).
-    let startBlock = Infinity; // Arbitrary initial value > 0.
+    let fromBlock = Infinity; // Arbitrary initial value > 0.
 
     // For loop continues until the start block hits 0 or the first event is before the earlest lookback time.
-    for (let i = 0; !(startBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
+    for (let i = 0; !(fromBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
       // Uses latest unless the events array already has data. If so, it only queries _before_ existing events.
-      const endBlock = events[0] ? events[0].blockNumber - 1 : "latest";
+      const toBlock = events[0] ? events[0].blockNumber - 1 : "latest";
 
       // By taking larger powers of 2, this doubles the lookback each time.
-      startBlock = Math.max(0, latestBlockNumber - lookbackBlocks * 2 ** i);
+      fromBlock = Math.max(0, latestBlockNumber - lookbackBlocks * 2 ** i);
 
-      const newEvents = await this._getSortedSyncEvents(startBlock, endBlock).then(newEvents => {
+      const newEvents = await this._getSortedSyncEvents(fromBlock, toBlock).then(newEvents => {
         // Grabs the timestamps for all blocks, but avoids re-querying by .then-ing any cached blocks.
         return Promise.all(
           newEvents.map(event => {
@@ -153,8 +153,8 @@ class UniswapPriceFeed extends PriceFeedInterface {
     this.lastUpdateTime = currentTime;
   }
 
-  async _getSortedSyncEvents(startBlock, endBlock) {
-    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: startBlock, toBlock: endBlock });
+  async _getSortedSyncEvents(fromBlock, toBlock) {
+    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: fromBlock, toBlock: toBlock });
     // Primary sort on block number. Secondary sort on transactionIndex. Tertiary sort on logIndex.
     events.sort((a, b) => {
       if (a.blockNumber !== b.blockNumber) {
--- a/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
@@ -92,18 +92,45 @@ class UniswapPriceFeed extends PriceFeedInterface {
     // Approximate the first block from which we'll need price data from based on the
     // lookback and twap length:
     const lookbackWindow = this.twapLength + this.historicalLookback;
+    const currentTime = await this.getTime();
+    const earliestLookbackTime = currentTime - lookbackWindow;
     const latestBlockNumber = (await this.web3.eth.getBlock("latest")).number;
-    // Add cushion in case `averageBlockTimeSeconds` overestimates the seconds per block:
-    const lookbackBlocks = Math.ceil((this.bufferBlockPercent * lookbackWindow) / (await averageBlockTimeSeconds()));
-    const earliestBlockNumber = latestBlockNumber - lookbackBlocks;
-    let fromBlock = earliestBlockNumber;
-    let events = await this.uniswap.getPastEvents("Sync", { fromBlock: Math.max(fromBlock, 0) });
+    // Add cushion in case `averageBlockTimeSeconds` underestimates the seconds per block:
+    let lookbackBlocks = Math.ceil((this.bufferBlockPercent * lookbackWindow) / (await averageBlockTimeSeconds()));
 
-    // For low-volume pools, it is possible that there are no Sync events within the lookback window.
-    // To cover these cases, we'll keep looking back until we find a window with a sync event.
-    while (fromBlock >= 0 && events.length === 0) {
-      fromBlock -= lookbackBlocks;
-      events = await this.uniswap.getPastEvents("Sync", { fromBlock: Math.max(fromBlock, 0) });
+    let events = []; // Caches sorted events (to keep subsequent event queries as small as possible).
+    let blocks = {}; // Caches blocks (so we don't have to re-query timestamps).
+    let startBlock = Infinity; // Arbitrary initial value > 0.
+
+    // For loop continues until the start block hits 0 or the first event is before the earlest lookback time.
+    for (let i = 0; !(startBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
+      // Uses latest unless the events array already has data. If so, it only queries _before_ existing events.
+      const endBlock = events[0] ? events[0].blockNumber - 1 : "latest";
+
+      // By taking larger powers of 2, this doubles the lookback each time.
+      startBlock = Math.max(0, latestBlockNumber - lookbackBlocks * 2 ** i);
+
+      const newEvents = await this._getSortedSyncEvents(startBlock, endBlock).then(newEvents => {
+        // Grabs the timestamps for all blocks, but avoids re-querying by .then-ing any cached blocks.
+        return Promise.all(
+          newEvents.map(event => {
+            // If there is nothing in the cache for this block number, add a new promise that will resolve to the block.
+            if (!blocks[event.blockNumber]) {
+              blocks[event.blockNumber] = this.web3.eth.getBlock(event.blockNumber);
+            }
+
+            // Add a .then to the promise that sets the timestamp (and price) for this event after the promise resolves.
+            return blocks[event.blockNumber].then(block => {
+              event.timestamp = block.timestamp;
+              event.price = this._getPriceFromSyncEvent(event);
+              return event;
+            });
+          })
+        );
+      });
+
+      // Adds newly queried events to the array.
+      events = [...newEvents, ...events];
     }
 
     // If there are still no prices, return null to allow the user to handle the absence of data.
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -92,18 +92,45 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
-      events = await this.uniswap.getPastEvents("Sync", { fromBlock: Math.max(fromBlock, 0) });
+    // Add cushion in case `averageBlockTimeSeconds` underestimates the seconds per block:
+    let lookbackBlocks = Math.ceil((this.bufferBlockPercent * lookbackWindow) / (await averageBlockTimeSeconds()));
+
+    let events = []; // Caches sorted events (to keep subsequent event queries as small as possible).
+    let blocks = {}; // Caches blocks (so we don't have to re-query timestamps).
+    let startBlock = Infinity; // Arbitrary initial value > 0.
+
+    // For loop continues until the start block hits 0 or the first event is before the earlest lookback time.
+    for (let i = 0; !(startBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
```
chrismaree, line 106: yes! +100 for `?.` 🤩

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -114,6 +141,20 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+    this.lastBlockPrice = this.events[this.events.length - 1].price;
+
+    // Compute TWAP up to the current time.
+    this.currentTwap = this._computeTwap(this.events, currentTime - this.twapLength, currentTime);
+
+    this.lastUpdateTime = currentTime;
+  }
+
+  async _getSortedSyncEvents(startBlock, endBlock) {
+    const events = await this.uniswap.getPastEvents("Sync", { fromBlock: startBlock, toBlock: endBlock });
```
chrismaree, line 157: tiny nit: it might be cleaner to just refer to `fromBlock` and `toBlock` through out this file so we don't need to have this translation between `fromBlock: startBlock` & `toBlock: endBlock`. feel free to ignore this comment but it just seems that it would be more consistent to always use `fromBlock` and `toBlock` notation for these vars.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -92,18 +92,45 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+    let blocks = {}; // Caches blocks (so we don't have to re-query timestamps).
+    let startBlock = Infinity; // Arbitrary initial value > 0.
+
+    // For loop continues until the start block hits 0 or the first event is before the earlest lookback time.
+    for (let i = 0; !(startBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
+      // Uses latest unless the events array already has data. If so, it only queries _before_ existing events.
+      const endBlock = events[0] ? events[0].blockNumber - 1 : "latest";
+
+      // By taking larger powers of 2, this doubles the lookback each time.
+      startBlock = Math.max(0, latestBlockNumber - lookbackBlocks * 2 ** i);
```
chrismaree, line 111: clever!

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -92,18 +92,45 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+
+      const newEvents = await this._getSortedSyncEvents(startBlock, endBlock).then(newEvents => {
```
chrismaree, line 113: seeing this does not run in parallel any more, what is the upper bound on how long this `update` function can now take to run, assuming this needs to search until the 0th block and never finds any `sync` events( for example a mis parameterized feed)?

chrismaree (approved):
This is great. really clever implementation. I left one small nit and one question but otherwise this looks good to me!

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
mrice32: Agreed. Will change.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -92,18 +92,45 @@ class UniswapPriceFeed extends PriceFeedInterface {
     // Approximate the first block from which we'll need price data from based on the
     // lookback and twap length:
     const lookbackWindow = this.twapLength + this.historicalLookback;
+    const currentTime = await this.getTime();
```
nicholaspai, line 95: Can you assume that getTime is async?

nicholaspai (approved):
Looks good! I'm surprised this doesn't change any of the unit tests, do we currently test the following situation:
- one data point within the look back window: twap should factor in the single data point and the last one before it
- no data point in the window: twap should only factor in the last one before it(?)

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
mrice32: Check my math here:
```
avg_block_time = 13.5;
standard_lookback = 7200; // Standard lookback we use with no TWAP.
total_blocks = 14140850; // ~block number in a year
lookback_in_blocks = standard_lookback / avg_block_time = 533.333333333;
iterations = log2(total_blocks / lookback_in_blocks) = 14.6944718176; // Number of times needed to double to reach the total blocks.
 ```

This means it's basically 15 ish loops, which would equate to O(15) serial calls, since no timestamp calls would be required in the average loop since a timestamp call would mean we found a price, terminating the looping.

Does that match your expectation?

File: packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js (hunk 5 above)
Comment (mrice32) [reply packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js:95]: We cannot, but `await` just passes the value through if it isn't a promise: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await, so no special casing should be necessary.

**Comment by mrice32** [comment]
> Looks good! I'm surprised this doesn't change any of the unit tests, do we currently test the following situation:
> 
> * one data point within the look back window: twap should factor in the single data point and the last one before it
> * no data point in the window: twap should only factor in the last one before it(?)

Both of these are tested (pretty much):
- "Basic Historical TWAP" tests the first, except that there are multiple data points in the TWAP. However, data needs to be considered that comes before the window.
- "All events before window" tests the second exactly.

I may try to add a test where the loop has to repeat to grab enough blocks, although this is kinda difficult with ganache since I don't think you can create blocks with arbitrary block numbers.

File: packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js (hunk 4 above)
Comment (mrice32) [reply packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js]: If we prefer, we can change this from 2 to 3 or even 10, etc if we want to make the ramp up faster. For instance, if we change it to 10, that brings the max iterations down from 15 to 5 ish, but it would mean that the queries would be much more painful for the node to fulfill much more quickly.

**Comment by Reinis-FRP** [comment]
After this PR got merged at cf2570eccb377cadd4b68878192b21301e01be76 I fail to run [getMedianHistoricalPrice.js](/packages/core/scripts/local/getMedianHistoricalPrice.js) with following error:

```
<HOME>/protocol/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js:106
    for (let i = 0; !(fromBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
                                                   ^
SyntaxError: Unexpected token .
    at Module._compile (internal/modules/cjs/loader.js:723:23)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/reinism/protocol/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js:5:30)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at Object.<anonymous> (/home/reinism/protocol/packages/financial-templates-lib/index.js:10:6)
    at Module._compile (internal/modules/cjs/loader.js:778:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
    at Module.load (internal/modules/cjs/loader.js:653:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
    at Function.Module._load (internal/modules/cjs/loader.js:585:3)
    at Module.require (internal/modules/cjs/loader.js:692:17)
    at require (internal/modules/cjs/helpers.js:25:18)
    at require (/home/reinism/protocol/node_modules/truffle/build/webpack:/packages/require/require.js:79:1)
    at /home/reinism/protocol/packages/core/scripts/local/getMedianHistoricalPrice.js:11:55
    at Script.runInContext (vm.js:133:20)
    at Script.runInNewContext (vm.js:139:17)
    at Object.file (/home/reinism/protocol/node_modules/truffle/build/webpack:/packages/require/require.js:94:1)
Truffle v5.1.41 (core: 5.1.41)
Node v10.19.0
error Command failed with exit code 1.
```

Before cf2570eccb377cadd4b68878192b21301e01be76 I could run getMedianHistoricalPrice script without errors. Could this commit be the issue or it has something to do with my local setup?

nicholaspai:
> After this PR got merged at [cf2570e](https://github.com/UMAprotocol/protocol/commit/cf2570eccb377cadd4b68878192b21301e01be76) I fail to run [getMedianHistoricalPrice.js](/packages/core/scripts/local/getMedianHistoricalPrice.js) with following error:
> 
> ```
> <HOME>/protocol/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js:106
>     for (let i = 0; !(fromBlock === 0 || events[0]?.timestamp <= earliestLookbackTime); i++) {
>                                                    ^
> SyntaxError: Unexpected token .
>     at Module._compile (internal/modules/cjs/loader.js:723:23)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
>     at Module.load (internal/modules/cjs/loader.js:653:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
>     at Module.require (internal/modules/cjs/loader.js:692:17)
>     at require (internal/modules/cjs/helpers.js:25:18)
>     at Object.<anonymous> (<HOME>/protocol/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js:5:30)
>     at Module._compile (internal/modules/cjs/loader.js:778:30)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
>     at Module.load (internal/modules/cjs/loader.js:653:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
>     at Module.require (internal/modules/cjs/loader.js:692:17)
>     at require (internal/modules/cjs/helpers.js:25:18)
>     at Object.<anonymous> (<HOME>/protocol/packages/financial-templates-lib/index.js:10:6)
>     at Module._compile (internal/modules/cjs/loader.js:778:30)
>     at Object.Module._extensions..js (internal/modules/cjs/loader.js:789:10)
>     at Module.load (internal/modules/cjs/loader.js:653:32)
>     at tryModuleLoad (internal/modules/cjs/loader.js:593:12)
>     at Function.Module._load (internal/modules/cjs/loader.js:585:3)
>     at Module.require (internal/modules/cjs/loader.js:692:17)
>     at require (internal/modules/cjs/helpers.js:25:18)
>     at require (<HOME>/protocol/node_modules/truffle/build/webpack:/packages/require/require.js:79:1)
>     at <HOME>/protocol/packages/core/scripts/local/getMedianHistoricalPrice.js:11:55
>     at Script.runInContext (vm.js:133:20)
>     at Script.runInNewContext (vm.js:139:17)
>     at Object.file (<HOME>/protocol/node_modules/truffle/build/webpack:/packages/require/require.js:94:1)
> Truffle v5.1.41 (core: 5.1.41)
> Node v10.19.0
> error Command failed with exit code 1.
> ```
> 
> Before [cf2570e](https://github.com/UMAprotocol/protocol/commit/cf2570eccb377cadd4b68878192b21301e01be76) I could run getMedianHistoricalPrice script without errors. Could this commit be the issue or it has something to do with my local setup?

Can you try upgrading your node version to >14? The `?.` syntax is only available in later node versions

Reinis-FRP:
> 
> Can you try upgrading your node version to >14? The `?.` syntax is only available in later node versions

Thanks a lot, this solved the issue!

PR #2476, opened 2021-01-27
improve(truffle): upgrade to latest truffle version

pemulis:

**Motivation**

Upgrades Truffle to v5.1.63 so that it works with the latest versions of nodejs.

**Summary**

Update package.json and `yarn build`.

**Details**

Ran into some difficulty due to this when doing Kovan migrations.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


pemulis:
Marking this as ready for review since it looks like the Slither failure is not specific to the Truffle update, and is being addressed in https://github.com/UMAprotocol/protocol/pull/2478. Not sure how we want to handle PRs in the meantime, since it seems none of them will make it past Slither.

package.json
```diff
@@ -40,6 +40,7 @@
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
     "ganache-cli": "^6.12.2",
+    "hardhat": "^2.0.8",
```
pemulis, line 43: I installed `hardhat` to try to fix the Slither issue, but it didn't work. This dependency might not be necessary after all.

mrice32: I could see it being useful to have it be a top-level dev dependency and/or being a devDependency of every package where it isn't needed as a runtime dep.

mrice32 (commented):
@chrismaree Where do you think the hardhat binary should be installed? As a top-level dev dependency? Or a package-by-package dev dependency?

chrismaree:
> @chrismaree Where do you think the hardhat binary should be installed? As a top-level dev dependency? Or a package-by-package dev dependency?

ye, the PR where I was trying this https://github.com/UMAprotocol/protocol/pull/2478 was just an experiment based on the error message generated in CI. Still need to explore more what's going on exactly.

mrice32:
@pemulis do we still want to merge this or is it unnecessary now?

pemulis:
> @pemulis do we still want to merge this or is it unnecessary now?

I think it's unnecessary. We could look at doing this as part of a broader upgrade of our packages.

PR #2477, opened 2021-01-28
fix(FindContractVersion): address web3.js inconsistant production behaviour with ethers.js

chrismaree:
**Motivation**

The most recent functionality with the perpetual bot support included the ability to automatically detect what contract version we are using. For some or other reason this failed to work in production.

**Summary**

After much painful testing, I was able to get this method to work in production by using ethers.js over web3.js. I'm not actually sure why this works, but it does 🤷 

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/FindContractVersion.js
+++ b/packages/common/src/FindContractVersion.js
@@ -3,16 +3,33 @@ const assert = require("assert");
 /**
  * Get the version and type of a financial contract deployed using the official UMA contract factories.
  * Note: all inputs and outputs are expressed as fixed-point (scaled by 1e18) BNs.
- * @param {Object} web3 instance
+ * @param {Object} web3 instance. This is passed in to re-use the calling context & network of the entry point.
  * @param {string} contractAddress address of the contract in question
  * @return {Object} contract name & version
  */
 async function findContractVersion(contractAddress, web3) {
   assert(web3, "Web3 object must be provided");
   assert(contractAddress, "Contract address must be provided");
-  const contractCode = await web3.eth.getCode(contractAddress);
+
+  // Note: there is an unknown issue in web3.js that means that the `getCode` syntax does not function correctly in
+  // production. However, ethers has proven to work correctly in production. The code below is a patch to still enable
+  // this module to work while we find a better long term solution for the web3.js issue. If running within unit tests
+  // then the web3.js version is required as it is scope according to the unit test.
+  let contractCode;
+  if (require("minimist")(process.argv.slice())?.network?.includes("mainnet")) {
+    const providers = require("ethers").providers;
+    const { getNodeUrl } = require("./TruffleConfig");
+    const argv = require("minimist")(process.argv.slice());
+
+    const provider = new providers.JsonRpcProvider(getNodeUrl(argv.network || "test"));
+    contractCode = await provider.getCode(contractAddress);
+  } else {
+    contractCode = await web3.eth.getCode(contractAddress);
+  }
+
   const contractCodeHash = web3.utils.soliditySha3(contractCode);
-  return versionMap[contractCodeHash] || {};
+
+  versionMap[contractCodeHash] || { contractAddress, contractCodeHash, contractCode: contractCode.substring(0, 1000) };
 }
 
 const versionMap = {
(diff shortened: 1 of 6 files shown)


packages/common/src/ProviderUtils.js
```diff
@@ -48,8 +48,8 @@ function createBasicProvider(url) {
  * @notice Gets a web3 instance based on the network argument using the truffle config in this package.
  * Use this for compatibility for running with or without truffle.
  * @example
- * // If a node app uses getWeb3() and you want to load network 1 with a default wallet
```
chrismaree, line 51: this change is unrelated. the comment style was wrong.

packages/liquidator/index.js
```diff
@@ -89,9 +89,11 @@ async function run({
       ).length == 0
     )
       throw new Error(
-        `Contract version specified or inferred is not supported by this bot. Provided/inferred config: ${JSON.stringify(
```
chrismaree, line 92: change made here helps with debugging as it prints out the whole `detectedContract`

packages/liquidator/test/Liquidator.js
```diff
@@ -31,9 +31,9 @@ const { Liquidator } = require("../src/liquidator.js");
 // 2) non-matching 8 collateral & 18 synthetic for legacy UMA synthetics.
 // 3) matching 8 collateral & 8 synthetic for current UMA synthetics.
 const configs = [
-  { tokenSymbol: "WETH", collateralDecimals: 18, syntheticDecimals: 18, priceFeedDecimals: 18 }
-  // { tokenSymbol: "BTC", collateralDecimals: 8, syntheticDecimals: 18, priceFeedDecimals: 8 },
-  // { tokenSymbol: "BTC", collateralDecimals: 8, syntheticDecimals: 8, priceFeedDecimals: 18 }
+  { tokenSymbol: "WETH", collateralDecimals: 18, syntheticDecimals: 18, priceFeedDecimals: 18 },
```
chrismaree, line 34: un-related change. this should not have been commented out on master.

packages/liquidator/package.json
```diff
@@ -32,7 +32,7 @@
   "bin": "index.js",
   "scripts": {
     "test": "yarn mocha-test && yarn hardhat-test",
-    "hardhat-test": "hardhat test --network hardhat",
```
chrismaree, line 35: this change was made as for some or other unknown reason having this `--network hardhat` broke the `ProviderUtils.js` with the following error:

image

we should consider removing this on all other packages as well but for now they seem to be passing CI.

packages/common/src/FindContractVersion.js
```diff
@@ -3,16 +3,36 @@ const assert = require("assert");
...
   assert(web3, "Web3 object must be provided");
   assert(contractAddress, "Contract address must be provided");
-  const contractCode = await web3.eth.getCode(contractAddress);
+
+  // Note: there is an unknown issue in web3.js that means that the `getCode` syntax does not function correctly in
+  // production. However, ethers has proven to work correctly in production. The code below is a patch to still enable
+  // this module to work while we find a better long term solution for the web3.js issue. If running within unit tests
+  // then the web3.js version is required as it is scope according to the unit test.
+  let contractCode;
+  if (require("minimist")(process.argv.slice())?.network?.includes("mainnet")) {
```
mrice32, line 19: nit: will this use web3 on testnets, like kovan? Will that break? Have you thought about using the isPublicNetwork function that we use in migrations?

A (potentially) more general way to switch is what we do in other parts of the codebase to determine whether we're in a truffle/hardhat test env:
```js
if (global.web3) {
  // This is run inside of truffle or hardhat test.
else {
  // This is run literally anywhere else.
}
```

mrice32 (commented):
Looks awesome -- one minor question.

packages/common/src/FindContractVersion.js
chrismaree: ye, that's probably cleaner. The solution I had put here was specifically looking for the case where I knew this _did not_ work, which is mainnet. However, you are right that it is likely that this also does not work on kovan & there is no reason really to not generalize it further. 

Using your `if (global.web3) {` is probs the cleanest. I have updated it to your implementation, PTAL.

daywiss (approved):
Looks good

mrice32 (approved):
LGTM!

PR #2479, opened 2021-01-28
feat(default-price-feed): add TraderMadePriceFeed

FreshmanQ:
**Motivation**
Follow up on UMAProtocol/UMIPs#97, [umip-32](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-32.md).
Similar to CryptoWatch price feed implementation, this PR can help us to get CNYUSD exchange rate from TraderMade

**Testing**

- [✅ ]  Ran end-to-end test, running the code as in production
- [✅ ]  New unit tests created
- [✅ ]  All existing tests pass

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -6,6 +6,7 @@ const { UniswapPriceFeed } = require("./UniswapPriceFeed");
 const { BalancerPriceFeed } = require("./BalancerPriceFeed");
 const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
 const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
+const { TraderMadePriceFeed } = require("./TraderMadePriceFeed");
 const { PriceFeedMockScaled } = require("./PriceFeedMockScaled");
 const { InvalidPriceFeedMock } = require("./InvalidPriceFeedMock");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
@@ -165,6 +166,31 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.denominatorPriceFeed && (await _createMedianizerPriceFeed(config.denominatorPriceFeed));
 
     return new BasketSpreadPriceFeed(web3, logger, baselinePriceFeeds, experimentalPriceFeeds, denominatorPriceFeed);
+  } else if (config.type === "tradermade") {
+    const requiredFields = ["pair", "apiKey", "minTimeBetweenUpdates"];
+
+    if (isMissingField(config, requiredFields, logger)) {
+      return null;
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating TraderMadePriceFeed",
+      config
+    });
+
+    return new TraderMadePriceFeed(
+      logger,
+      web3,
+      config.apiKey,
+      config.pair,
+      config.minuteLookback,
+      config.hourlyLookback,
+      networker,
+      getTime,
+      config.minTimeBetweenUpdates,
+      config.priceFeedDecimals // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
+    );
   } else if (config.type === "test") {
     const requiredFields = ["currentPrice", "historicalPrice"];
     if (isMissingField(config, requiredFields, logger)) {
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -189,7 +189,8 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       networker,
       getTime,
       config.minTimeBetweenUpdates,
-      config.priceFeedDecimals // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
+      config.priceFeedDecimals, // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
+      config.ohlcPeriod
     );
   } else if (config.type === "test") {
     const requiredFields = ["currentPrice", "historicalPrice"];
--- a/packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
@@ -1,0 +1,364 @@
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const { parseFixed } = require("@uma/common");
+const moment = require("moment");
+
+// An implementation of PriceFeedInterface that uses TraderMade api to retrieve prices.
+class TraderMadePriceFeed extends PriceFeedInterface {
+  /**
+   * @notice Constructs the TraderMadePriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {String} pair Representation of the pair the price feed is tracking.
+   * @param {String} apiKey TraderMade Data API key.
+   * @param {Integer} minuteLookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Integer} hourlyLookback How far in the past the historical prices will be available using getHistoricalPricePeriods.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the current time.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+   * @param {Number} ohlcPeriod Number of minutes interval between ohlc prices requested from TraderMade.
+   */
+  constructor(
+    logger,
+    web3,
+    apiKey,
+    pair,
+    minuteLookback,
+    hourlyLookback,
+    networker,
+    getTime,
+    minTimeBetweenUpdates,
+    priceFeedDecimals = 18,
+    ohlcPeriod = 10 // Only 5, 10, 15 minutes is supported by TraderMade.
+  ) {
+    super();
+    this.logger = logger;
+    this.web3 = web3;
+
+    this.apiKey = apiKey;
+    this.pair = pair;
+
+    this.minuteLookback = minuteLookback;
+    this.hourlyLookback = hourlyLookback;
+    this.uuid = `TraderMade-${pair}`;
+    this.networker = networker;
+    this.getTime = getTime;
+    this.minTimeBetweenUpdates = minTimeBetweenUpdates;
+
+    this.toBN = this.web3.utils.toBN;
+    this.ohlcPeriod = ohlcPeriod;
+
+    this.priceFeedDecimals = priceFeedDecimals;
+    this.ohlcPeriod = ohlcPeriod;
+
+    this.convertPriceFeedDecimals = number => {
+      // Converts price result to wei
+      // returns price conversion to correct decimals as a big number
+      return this.toBN(parseFixed(number.toString(), priceFeedDecimals).toString());
+    };
+  }
+
+  getCurrentPrice() {
+    return this.currentPrice;
+  }
+
+  async getHistoricalPrice(time, verbose = false) {
+    if (this.lastUpdateTime === undefined) {
+      throw new Error(`${this.uuid}: undefined lastUpdateTime`);
+    }
+
+    // Set first price time in `historicalPrices` to first non-null price.
+    let firstPriceTime;
+    for (let p in this.historicalPricesMinute) {
+      if (this.historicalPricesMinute[p] && this.historicalPricesMinute[p].openTime) {
+        firstPriceTime = this.historicalPricesMinute[p];
+        break;
+      }
+    }
+
+    // If there are no valid price time, return null.
+    if (!firstPriceTime) {
+      throw new Error(`${this.uuid}: no valid price time`);
+    }
+
+    // If the time is before the first piece of data in the set, return null because
+    // the price is before the lookback window.
+    if (time < firstPriceTime.openTime) {
+      throw new Error(`${this.uuid}: time ${time} is before firstPriceTime.openTime`);
+    }
+
+    // historicalPrices are ordered from oldest to newest.
+    // This finds the first priceTime whose closeTime is after the provided time.
+    const match = this.historicalPricesMinute.find(price => {
+      return time < price.closeTime;
+    });
+
+    // If there is no match, that means that the time was past the last data point.
+    // In this case, the best match for this price is the current price.
+    let returnPrice;
+    if (match === undefined) {
+      returnPrice = this.currentPrice;
+      if (verbose) {
+        console.group(`\n(${this.pair}) No OHLC available @ ${time}`);
+        console.log(
+          `- ✅ Time is later than earliest historical time, fetching current price: ${this.web3.utils.fromWei(
+            returnPrice.toString()
+          )}`
+        );
+        console.log(
+          `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://marketdata.tradermade.com/api/v1/live?currency=${this.pair}&api_key={api-key}`
+        );
+        console.groupEnd();
+      }
+      return this.currentPrice;
+    }
+
+    returnPrice = match.closePrice;
+    if (verbose) {
+      console.group(`\n(${this.pair}) Historical OHLC @ ${match.closeTime}`);
+      console.log(`- ✅ Close Price:${this.web3.utils.fromWei(returnPrice.toString())}`);
+      console.log(
+        `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://marketdata.tradermade.com/api/v1/timeseries?currency=${this.pair}&api_key={api-key}&start_date=${time}&end_date=${match.closeTime}&format=records&interval=minute&period=${this.ohlcPeriod}`
+      );
+      console.groupEnd();
+    }
+    return returnPrice;
+  }
+
+  getHistoricalPricePeriods() {
+    return this.historicalPricesHourly;
+  }
+
+  getLastUpdateTime() {
+    return this.lastUpdateTime;
+  }
+
+  getMinuteLookback() {
+    return this.minuteLookback;
+  }
+
+  getHourlyLookback() {
+    return this.hourlyLookback;
+  }
+
+  getPriceFeedDecimals() {
+    return this.priceFeedDecimals;
+  }
+
+  async updateLatest(lastUpdateTime) {
+    const currentTime = this.getTime();
+
+    // Return early if the last call was too recent.
+    if (this.lastUpdateTime !== undefined && lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+      this.logger.debug({
+        at: "TraderMadePriceFeed",
+        message: "Update skipped because the last one was too recent",
+        currentTime: currentTime,
+        lastUpdateTimestamp: this.lastUpdateTime,
+        timeRemainingUntilUpdate: this.lastUpdateTimes + this.minTimeBetweenUpdates - currentTime
+      });
+      return;
+    }
+
+    this.logger.debug({
+      at: "TraderMade_PriceFeed",
+      message: "Updating Latest Price",
+      currentTime: currentTime,
+      lastUpdateTimestamp: this.lastUpdateTime
+    });
+
+    // 1. Construct URLs.
+    const priceUrl = `https://marketdata.tradermade.com/api/v1/live?currency=${this.pair}&api_key=${this.apiKey}`;
+
+    // 2. Send requests.
+    const priceResponse = await this.networker.getJson(priceUrl);
+
+    // 4. Parse results.
+    // Return data structure:
+    //  {
+    //     "endpoint": "live",
+    //     "quotes": [
+    //       {
+    //         "ask": 0.15431,
+    //         "base_currency": "CNY",
+    //         "bid": 0.15431,
+    //         "mid": 0.15431,
+    //         "quote_currency": "USD"
+    //       }
+    //     ],
+    //     "requested_time": "Tue, 26 Jan 2021 01:29:52 GMT",
+    //     "timestamp": 1611624593
+    //   }
+    // For more info, see: https://marketdata.tradermade.com/documentation
+    const newPrice = this.convertPriceFeedDecimals(priceResponse.quotes[0].ask);
+
+    // 5. Store results.
+    this.currentPrice = newPrice;
+    this.lastUpdateTime = currentTime;
+  }
+
+  async updateMinute(lastUpdateTime) {
+    const currentTime = this.getTime();
+
+    // Return early if the last call was too recent.
+    if (this.lastUpdateTime !== undefined && lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+      return;
+    }
+
+    this.logger.debug({
+      at: "TraderMade_PriceFeed",
+      message: "Updating Minute Price",
+      currentTime: currentTime,
+      lastUpdateTimestamp: this.lastUpdateTime
+    });
+
+    // Round down to the nearest ohlc period so the queries captures the OHLC of the period *before* this earliest
+    // timestamp (because the close of that OHLC may be relevant).
+    const earliestMinuteHistoricalTimestamp =
+      Math.floor((currentTime - this.minuteLookback) / (this.ohlcPeriod * 60)) * (this.ohlcPeriod * 60);
+    const endDate = this._secondToDateTime(currentTime);
+    const startMinuteDate = this._secondToDateTime(earliestMinuteHistoricalTimestamp);
+
+    // 1. Construct URLs.
+    const ohlcMinuteUrl = `https://marketdata.tradermade.com/api/v1/timeseries?currency=${this.pair}&api_key=${this.apiKey}&start_date=${startMinuteDate}&end_date=${endDate}&format=records&interval=minute&period=${this.ohlcPeriod}`;
+
+    // 2. Send requests.
+    const ohlcMinuteResponse = await this.networker.getJson(ohlcMinuteUrl);
+
+    // 3. Check responses.
+    if (!ohlcMinuteResponse || !ohlcMinuteResponse.quotes || !ohlcMinuteResponse.quotes[0].close) {
+      throw new Error(
+        `🚨Could not parse ohlc minute price result from url ${ohlcMinuteUrl}: ${JSON.stringify(ohlcMinuteResponse)}`
+      );
+    }
+
+    // Return data structure:
+    // {
+    //   "base_currency": "CNY",
+    //   "end_date": "2021-01-26 03:31:00",
+    //   "endpoint": "timeseries",
+    //   "quote_currency": "USD",
+    //   "quotes": [
+    //     {
+    //       "close": 0.1543,
+    //       "date": "2021-01-26 00:01:00",
+    //       "high": 0.1543,
+    //       "low": 0.1543,
+    //       "open": 0.1543
+    //     },
+    //     ...
+    //   ]
+    //   "request_time": "Wed, 27 Jan 2021 01:45:28 GMT",
+    //   "start_date": "2021-01-26-00:01"
+    // }
+    // For more info, see: https://marketdata.tradermade.com/documentation
+    const newHistoricalPricesMinute = ohlcMinuteResponse.quotes
+      .map(ohlcMinute => ({
+        // Output data should be a list of objects with only the open and close times and prices.
+        closePrice: this.convertPriceFeedDecimals(ohlcMinute.close),
+        openTime: this._dateTimeToSecond(ohlcMinute.date) - this.ohlcPeriod * 60,
+        closeTime: this._dateTimeToSecond(ohlcMinute.date)
+      }))
+      .sort((a, b) => {
+        // Sorts the data such that the oldest elements come first.
+        return a.openTime - b.openTime;
+      });
+
+    // 5. Store results.
+    this.historicalPricesMinute = newHistoricalPricesMinute;
+  }
+
+  async updateHourly(lastUpdateTime) {
+    const currentTime = this.getTime();
+
+    // Return early if the last call was too recent.
+    if (this.lastUpdateTime !== undefined && lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+      return;
+    }
+
+    this.logger.debug({
+      at: "TraderMade_PriceFeed",
+      message: "Updating Hourly Price",
+      currentTime: currentTime,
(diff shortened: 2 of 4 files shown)


FreshmanQ:
fyi @jishankai @smb2796 take a look if you have time, thanks

jishankai (commented):
Great work! 

I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
 ```
https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
```

Need more feedback from @smb2796

**Comment by FreshmanQ** [comment]
> Great work!
> 
> I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> 
> ```
> https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> ```
> 
> Need more feedback from @smb2796

Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.

File: packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js [hunk 1]
```diff
@@ -162,6 +163,30 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    return new TraderMadePriceFeed(
+      logger,
+      web3,
+      config.apiKey,
+      config.pair,
+      config.lookback,
+      networker,
+      getTime,
+      config.minTimeBetweenUpdates,
+      config.priceFeedDecimals // Defaults to 6 unless supplied. Informs how the feed should be scaled to match a DVM response.
```
nicholaspai, line 188: I think the default decimals for any price feed should generally be 18 since this is the default for any Ethereum ERC20 contracts

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,255 @@
...
+   * @param {String} apiKey TraderMade Data API key.
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the current time.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+
+   */
+  constructor(logger, web3, apiKey, pair, lookback, networker, getTime, minTimeBetweenUpdates, priceFeedDecimals = 6) {
```
nicholaspai, line 20: As mentioned above, I think its preferred if the default `priceFeedDecimals` is 18

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,255 @@
...
+      let returnPrice = this.currentPrice;
+      if (verbose) {
+        console.group(`\n(${this.pair}) No OHLC available @ ${time}`);
+        console.log(
+          `- ✅ Time is later than earliest historical time, fetching current price: ${this.web3.utils.fromWei(
+            returnPrice.toString()
+          )}`
+        );
+        console.log(
+          `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://marketdata.tradermade.com/api/v1/live?currency=${this.pair}&api_key={api-key}`
```
nicholaspai, line 90: Does this work without an `apiKey`? If not can you add it to the list of required params in `CreatePriceFeed`

nicholaspai:
> > Great work!
> > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > ```
> > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > ```
> > 
> > 
> > Need more feedback from @smb2796
> 
> Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.

What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw

nicholaspai (commented):
Thanks for the submission, I'm really excited to implement this price feed! I left a first pass through set of comments, looking forward to working together.

FreshmanQ:
> > > Great work!
> > > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > > ```
> > > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > > ```
> > > 
> > > 
> > > Need more feedback from @smb2796
> > 
> > 
> > Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.
> 
> What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw

The granularity of the `minute_historical` endpoint is 10 minutes, however, this endpoint doesn't support time range query.

In this PR the granularity of `hourly` endpoint is 1 hour, which supports up to 2 months. The granularity of `minute` endpoint is 5 minutes, which supports up to 2 days. 

Due to the `getHistoricalPricePeriods` needs to calculate historical prices for more than one week of time, so I use the `hourly` endpoint

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
FreshmanQ: sure,

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
FreshmanQ: got it,

nicholaspai:
> > > > Great work!
> > > > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > > > ```
> > > > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > > > ```
> > > > 
> > > > 
> > > > Need more feedback from @smb2796
> > > 
> > > 
> > > Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.
> > 
> > 
> > What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw
> 
> The granularity of the `minute_historical` endpoint is 10 minutes, however, this endpoint doesn't support time range query.
> 
> In this PR the granularity of `hourly` endpoint is 1 hour, which supports up to 2 months. The granularity of `minute` endpoint is 5 minutes, which supports up to 2 days.
> 
> Due to the `getHistoricalPricePeriods` needs to calculate historical prices for more than one week of time, so I use the `hourly` endpoint

How do you propose we get intra-hour prices? Interpolate between hourly prices?

FreshmanQ:
> > > > > Great work!
> > > > > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > > > > ```
> > > > > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > > > > ```
> > > > > 
> > > > > 
> > > > > Need more feedback from @smb2796
> > > > 
> > > > 
> > > > Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.
> > > 
> > > 
> > > What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw
> > 
> > 
> > The granularity of the `minute_historical` endpoint is 10 minutes, however, this endpoint doesn't support time range query.
> > In this PR the granularity of `hourly` endpoint is 1 hour, which supports up to 2 months. The granularity of `minute` endpoint is 5 minutes, which supports up to 2 days.
> > Due to the `getHistoricalPricePeriods` needs to calculate historical prices for more than one week of time, so I use the `hourly` endpoint
> 
> How do you propose we get intra-hour prices? Interpolate between hourly prices?

can I do something like this? use a cache to store historical price, cache will be filled either when update or a cache miss
```
getHistoricalPrice(time) {
  if (priceCache[time]) {
    return priceCache[time];
  }
  // ... fetch price asynchronously and fill the cache so next query should hit
  fetchPrice(time).then(...);
}

async update() {
  // ... fetch current price and write into cache
  priceCache[time] = price;
  // ... other stuff
}
```

nicholaspai:
> > > > > > Great work!
> > > > > > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > > > > > ```
> > > > > > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > > > > > ```
> > > > > > 
> > > > > > 
> > > > > > Need more feedback from @smb2796
> > > > > 
> > > > > 
> > > > > Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.
> > > > 
> > > > 
> > > > What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw
> > > 
> > > 
> > > The granularity of the `minute_historical` endpoint is 10 minutes, however, this endpoint doesn't support time range query.
> > > In this PR the granularity of `hourly` endpoint is 1 hour, which supports up to 2 months. The granularity of `minute` endpoint is 5 minutes, which supports up to 2 days.
> > > Due to the `getHistoricalPricePeriods` needs to calculate historical prices for more than one week of time, so I use the `hourly` endpoint
> > 
> > 
> > How do you propose we get intra-hour prices? Interpolate between hourly prices?
> 
> can I do something like this? use a cache to store historical price, cache will be filled either when update or a cache miss
> 
> ```
> getHistoricalPrice(time) {
>   if (priceCache[time]) {
>     return priceCache[time];
>   }
>   // ... fetch price asynchronously and fill the cache so next query should hit
>   fetchPrice(time).then(...);
> }
> 
> async update() {
>   // ... fetch current price and write into cache
>   priceCache[time] = price;
>   // ... other stuff
> }
> ```

Well what if the price feed is used by a serverless bot that turns on/turns off on each iteration? What you could do is just crawl backwards until you get the latest hourly price. This obviously won't be as accurate as the UMIP might call for (i.e. the UMIP might want minute by minute granularity, and this solution would only give you the latest hourly price) but at least it will return a price.

I think it might be best to use the `/minute` endpoint optimistically, and then once a requested lookback is outside of the 2 day window, then we can fallback to the `/hour` endpoint.

@mrice @chrismaree @daywiss WDYT? The problem right now is that the `/minute` endpoint only allows a lookback of up to 2 days, but the `affiliates` package requires a week long lookback. The `/hour` endpoint has hourly granularity and supports a lookback of up to 2 months.

mrice32:
> > > > > > > Great work!
> > > > > > > I think it maybe is different from CryptoWatch for getting history price, should we change the ohlc endpoint to the historical one?
> > > > > > > ```
> > > > > > > https://marketdata.tradermade.com/api/v1/minute_historical?currency=CNYUSD&date_time=2020-11-11-13:50&api_key={apikey}
> > > > > > > ```
> > > > > > > 
> > > > > > > 
> > > > > > > Need more feedback from @smb2796
> > > > > > 
> > > > > > 
> > > > > > Since `minute_historical` doesn't support time range query, if we want to search like 1-day price, we may have to use a for loop. So, I change the endpoint from historical to timeseries.
> > > > > 
> > > > > 
> > > > > What is the granularity of the `/minute_historical` endpoint? minutes? Minutes is sufficient fwiw
> > > > 
> > > > 
> > > > The granularity of the `minute_historical` endpoint is 10 minutes, however, this endpoint doesn't support time range query.
> > > > In this PR the granularity of `hourly` endpoint is 1 hour, which supports up to 2 months. The granularity of `minute` endpoint is 5 minutes, which supports up to 2 days.
> > > > Due to the `getHistoricalPricePeriods` needs to calculate historical prices for more than one week of time, so I use the `hourly` endpoint
> > > 
> > > 
> > > How do you propose we get intra-hour prices? Interpolate between hourly prices?
> > 
> > 
> > can I do something like this? use a cache to store historical price, cache will be filled either when update or a cache miss
> > ```
> > getHistoricalPrice(time) {
> >   if (priceCache[time]) {
> >     return priceCache[time];
> >   }
> >   // ... fetch price asynchronously and fill the cache so next query should hit
> >   fetchPrice(time).then(...);
> > }
> > 
> > async update() {
> >   // ... fetch current price and write into cache
> >   priceCache[time] = price;
> >   // ... other stuff
> > }
> > ```
> 
> Well what if the price feed is used by a serverless bot that turns on/turns off on each iteration? What you could do is just crawl backwards until you get the latest hourly price. This obviously won't be as accurate as the UMIP might call for (i.e. the UMIP might want minute by minute granularity, and this solution would only give you the latest hourly price) but at least it will return a price.
> 
> I think it might be best to use the `/minute` endpoint optimistically, and then once a requested lookback is outside of the 2 day window, then we can fallback to the `/hour` endpoint.
> 
> @mrice @chrismaree @daywiss WDYT? The problem right now is that the `/minute` endpoint only allows a lookback of up to 2 days, but the `affiliates` package requires a week long lookback. The `/hour` endpoint has hourly granularity and supports a lookback of up to 2 months.

Added @daywiss to the list of reviewers as I think he would have the strongest opinions about how important this is. I could imagine the price feed dynamically switching to the hour endpoint if it has a lookback > 2 days or something.

daywiss (commented):
Looks great

daywiss:
> Added @daywiss to the list of reviewers as I think he would have the strongest opinions about how important this is. I could imagine the price feed dynamically switching to the hour endpoint if it has a lookback > 2 days or something.

I'm not too worried about this on the affiliates side since the affiliates calculation uses `getHistoricalPricePeriods`, which is not implemented in this PR. So optimize the current code for the bots, which I think if you dont expect more than a 2 day look back, then use the higher resolution api. 

For the `getHistoricalPricePeriods` call which affiliates does need, we may have to do something more sophisticated, but I think personally I would be ok with keeping it simple and using hourly prices. I don't think it will make that much of a difference with the final calculation and we could easily add this now. 

what do you guys think, are hourly prices good enough for affiliates historical query? @UMAprotocol/eng

nicholaspai:
> > Added @daywiss to the list of reviewers as I think he would have the strongest opinions about how important this is. I could imagine the price feed dynamically switching to the hour endpoint if it has a lookback > 2 days or something.
> 
> I'm not too worried about this on the affiliates side since the affiliates calculation uses `getHistoricalPricePeriods`, which is not implemented in this PR. So optimize the current code for the bots, which I think if you dont expect more than a 2 day look back, then use the higher resolution api.
> 
> For the `getHistoricalPricePeriods` call which affiliates does need, we may have to do something more sophisticated, but I think personally I would be ok with keeping it simple and using hourly prices. I don't think it will make that much of a difference with the final calculation and we could easily add this now.
> 
> what do you guys think, are hourly prices good enough for affiliates historical query? @UMAprotocol/eng

For affiliates, hourly prices seem perfectly fine to me. But for the liquidator/disputer bots we'll want minute data.

daywiss:
> For affiliates, hourly prices seem perfectly fine to me. But for the liquidator/disputer bots we'll want minute data.

Sounds good, unless anyone has objections, I think we can use the 10 min api for now and add in the hourly api for the `getHistoricalPricePeriods`.  This requires 2 datastructures ( one for minute, one for hour data) which is a slight deviation from previous implementations, but I think thats not a big deal. 

@FreshmanQ Do you mind making these changes? Basically switch your current implementation to use the 10 minute feed, and add in a second datastructure for an hourly feed which is returned with `getHistoricalPricePeriods`.  Theres an example [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L136)

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,255 @@
...
+    const year = +ymd[0];
+    const month = +ymd[1];
+    const day = +ymd[2];
+    second = new Date(year, month, day).getTime() / 1000;
+    const hm = timeserires[1].split(":");
+    const hour = +hm[0];
+    const minute = +hm[1];
+    second += hour * 60 * 60 + minute * 60;
+    return second;
+  }
```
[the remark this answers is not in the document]
FreshmanQ: got it, thanks

FreshmanQ:
> here

Thanks for your help, I'll switch my implementation

FreshmanQ:
@daywiss @nicholaspai @mrice32 
made all the above changes, please take a look

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,278 @@
...
+      console.log(`- ✅ Close Price:${this.web3.utils.fromWei(returnPrice.toString())}`);
+      console.log(
+        `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://marketdata.tradermade.com/api/v1/timeseries?currency=${this.pair}&api_key={api-key}&start_date=${time}&end_date=${match.closeTime}&format=records&interval=minute&period=${this.ohlcPeriod}`
+      );
+      console.groupEnd();
+    }
+    return returnPrice;
+  }
+
+  getHistoricalPricePeriords() {
```
nicholaspai, line 123:
```suggestion
  getHistoricalPricePeriods() {
```

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,278 @@
...
+    // Round down to the nearest ohlc period so the queries captures the OHLC of the period *before* this earliest
+    // timestamp (because the close of that OHLC may be relevant).
+    const earliestHistoricalTimestamp =
+      Math.floor((currentTime - this.lookback) / (this.ohlcPeriod * 60)) * (this.ohlcPeriod * 60);
+    const endDate = this._secondToDateTime(currentTime);
+    const startDate = this._secondToDateTime(earliestHistoricalTimestamp);
+
+    // 1. Construct URLs.
+    const priceUrl = `https://marketdata.tradermade.com/api/v1/live?currency=${this.pair}&api_key=${this.apiKey}`;
+    const ohlcMinuteUrl = `https://marketdata.tradermade.com/api/v1/timeseries?currency=${this.pair}&api_key=${this.apiKey}&start_date=${startDate}&end_date=${endDate}&format=records&interval=minute&period=${this.ohlcPeriod}`;
```
nicholaspai, line 170: Are all integers allowed for the `ohlc` param? Or do only certain round numbers like 10, 1, 50, etc. work?

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,278 @@
...
+      console.log(
+        `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://marketdata.tradermade.com/api/v1/timeseries?currency=${this.pair}&api_key={api-key}&start_date=${time}&end_date=${match.closeTime}&format=records&interval=minute&period=${this.ohlcPeriod}`
+      );
+      console.groupEnd();
+    }
+    return returnPrice;
+  }
+
+  getHistoricalPricePeriords() {
+    return this.historicalPricesHourly;
```
nicholaspai, line 124: @daywiss does this work for dev mining? This returnsr the hourly periods for `getHistoricalPricePeriods` and uses the minute periods for `getHistoricalPrice`

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,278 @@
...
+    //       "open": 0.1543
+    //     },
+    //     ...
+    //   ]
+    //   "request_time": "Wed, 27 Jan 2021 01:45:28 GMT",
+    //   "start_date": "2021-01-26-00:01"
+    // }
+    // For more info, see: https://marketdata.tradermade.com/documentation
+    const newHistoricalPricesMinute = ohlcMinuteResponse.quotes
+      .map(ohlcMinute => ({
```
nicholaspai, line 237: So it seems we don't care about the `open` , `high`, or `low` prices right?

FreshmanQ: yes, only use close price

packages/financial-templates-lib/test/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,303 @@
...
+      {
+        quotes: [
+          {
+            error: "test"
+          }
+        ]
+      }
+    ];
+
+    assert.isTrue(await traderMadePriceFeed.update().catch(() => true), "Update didn't throw");
```
nicholaspai, line 256: You can also use `assert.throws()`

nicholaspai (commented):
Thanks for making the changes, left a new round of comments let me know what you think

packages/financial-templates-lib/test/price-feed/TraderMadePriceFeed.js
FreshmanQ: `assert.throws` for `async` function may be a little complex, can we skip this?

nicholaspai: Yep

nicholaspai: @mrice32 btw this is how you can async test for throws

mrice32: The way you have it now looks right to me!

FreshmanQ:
@daywiss could you help take a look at the latest implementation? thanks,

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,283 @@
...
+      // returns price conversion to correct decimals as a big number
+      return this.toBN(parseFixed(number.toString(), priceFeedDecimals).toString());
+    };
+  }
+
+  getCurrentPrice() {
+    return this.currentPrice;
+  }
+
+  getHistoricalPrice(time, verbose = false) {
```
nicholaspai, line 64: This should be an `async` function now to match the `PriceFeedInterface`. Additionally, instead of returning `null` this function should throw an Error. The [CryptoWatchPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L69) is probably the best model that you can follow for help!

nicholaspai (commented):
Looks good to me, waiting on @daywiss 's go ahead that this would be good for Affiliates

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
daywiss: yes that should work

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -162,6 +163,31 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+      config
+    });
+
+    return new TraderMadePriceFeed(
+      logger,
+      web3,
+      config.apiKey,
+      config.pair,
+      config.botLookback,
+      config.affiliateLookback,
```
daywiss, line 185: I like how you broke out the 2 lookback params, but it breaks the expected API, at least for affiliates, as it expects only a `lookback` param.  I think though just leave this for now and I will change it on my end.

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,283 @@
...
+  getLookback() {
+    return this.lookback;
+  }
+
+  getPriceFeedDecimals() {
+    return this.priceFeedDecimals;
+  }
+
+  async update() {
+    const currentTime = this.getTime();
```
daywiss, line 143: I'm going to suggest something that might be a pain in the butt here, but because some of these queries may be optional, it might be good to have a way to disable certain parts of the update.  Id suggest maybe breaking each one of these out into their own functions:
- updateLatest() -> sets current price and last update time
- updateMinute() -> sets minute history based on bot lookback
- updateHourly() -> sets hourly history based on affiliate lookback

Then in `update()` basically check if the params exist before calling them
```js
  await this.updateLatest()
  if(this.botLookback){
     await this.updateMinute()
  }
  if(this.affiliateLookback){
     await this.updateHourly()
  }
```

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
daywiss: just a  nit here but maybe it might be clearer to understand what these are doing to rename them to something like:
`botLookback -> minuteLookback`
`affiliateLookback -> hourlyLookback`

daywiss (commented):
looks good, just want to see what you think of some of these comments

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
FreshmanQ: changed

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
FreshmanQ: updated

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
FreshmanQ: fixed

daywiss (approved):
Thanks for working on this and the updates. Looks good to me.  In the future don't be afraid to re-request reviews, so we get notified when you are ready.

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -0,0 +1,362 @@
...
+    };
+  }
+
+  getCurrentPrice() {
+    return this.currentPrice;
+  }
+
+  async getHistoricalPrice(time, verbose = false) {
+    if (this.lastUpdateTime === undefined) {
+      throw new Error(`${this.pair}: undefined lastUpdateTime`);
```
nicholaspai, line 66: Would you mind starting all of the error messages with a `this.uuid` field? Look [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L71) for example implementation. Should be pretty trivial to implement, but it helps parse logs when some bots rely on multiple pricefeeds

nicholaspai (approved):
Looks great to me, just one more requested change please

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
FreshmanQ: fixed,

nicholaspai:
@FreshmanQ I authorized the `test` CI to run, apologies, can you finally squash and sign off all commits? See here for example: https://github.com/UMAprotocol/protocol/pull/2346#issuecomment-770594038

daywiss:
looks like this needs to be rebased against master. @FreshmanQ is this somehting you know how to do? If not i might be able to provide some instructions

FreshmanQ:
> looks like this needs to be rebased against master. @FreshmanQ is this somehting you know how to do? If not i might be able to provide some instructions

seems like I use the wrong email to sign off DCO, could you help me to fix that? thanks,

nicholaspai:
> > looks like this needs to be rebased against master. @FreshmanQ is this somehting you know how to do? If not i might be able to provide some instructions
> 
> seems like I use the wrong email to sign off DCO, could you help me to fix that? thanks,

Working on this

daywiss:
@FreshmanQ we cant have all these duplicate commits from other contributors woven in with your changes. It looks like it might be complicated to fix.  You can see in the commit tab, theres all these other unrelated commits in the PR.  Not 100% sure best course of action here, but we might need to branch it, cherry picking commits and rebase on master.

mrice32:
@FreshmanQ These are my recommended steps to fix DCO issues after a long and complex commit history.

Yep! Note: please back up your branch before you do this: `git checkout -b backup_branch`. Then switch back to the branch you've been using.

This is what I'd recommend as a simple way around these issues. Note: I'm using `upstream/master` as your local reference to our master branch. Just swap it out with whatever branch you typically merge into your branch to pull in upstream changes. Note: please make sure that whatever local reference you're using to refer to our master branch (upstream/master in the example below) has already been merged into your branch. If not, you'll get weird changes in the final diff.

- `git reset upstream/master` destroys the git history on your branch and makes it match upstream/master, BUT it keeps all your changes as unstaged changes.
- `git add .` stages all of the changes at once.
- `git commit -s -m "feat(financial-templates): adds a new price feed for the DefiPulseTotalTVL synthetic tokens"` creates a single signed commit on your branch that contains all of your changes.
- `git push -f` force overwrites the remote branch on github.

Note: make sure you backup everything in a place where you can retrieve it if something breaks in this process. Since you're wiping your git history on your branch, you could wind up accidentally wiping your changes, so be sure you have a way to start over if things go bad!

FreshmanQ:
@nicholaspai @daywiss @mrice32 thanks for all your help, please do a final check~

mrice32:
@FreshmanQ It looks like when you ran `git reset upstream/master`, your branch wasn't up-to-date with `upstream/master`. This caused a _lot_ of weirdness in the resulting diff. Basically, you committed a change that _removed_ a lot of our recent changes (that occurred between when you updated last and the branch you reset to). You might want to start over with your backup by doing something like `git reset --hard backup_branch `, then run `git merge upstream/master`. Then start the process outlined above again.

daywiss (approved):

pemulis (approved):

pemulis:
Looks like there's one failing test in CI; should be good once that's resolved.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -165,6 +166,31 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+      logger,
+      web3,
+      config.apiKey,
+      config.pair,
+      config.minuteLookback,
+      config.hourlyLookback,
+      networker,
+      getTime,
+      config.minTimeBetweenUpdates,
+      config.priceFeedDecimals // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
```
mrice32, line 192: Should the user be able to also configure the ohlcPeriod?
```suggestion
      config.priceFeedDecimals, // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
      config.ohlcPeriod
```

FreshmanQ: I see, maybe the backup_branch record an old version..., thanks for pointing this

Issue #2474, opened 2021-01-27
Add CoinMarketCap and CoinGeckgo DAI:PHP and PHP:DAI price feed that serves as basis to mint a Philippine Peso (PHP) stablecoin


ctverceles:
**Why**

[HaloDAO’s](https://halodao.com/) goal is to bring annual billions of real world liquidity from Southeast Asian cross border remittances, consumer payments, online shopping and financial services into DeFi’s TVL. On the business side, we’ve locked up the partnerships to begin doing that for crypto gaming and the Singapore - Philippines corridor, but on the backend we need a permissionless method to mint region specific synthetic stablecoins out of commonly used collateral (starting with DAI). For this reason we need to integrate a DAI:PHP price feed that we would then invert for PHP:DAI so that DAI collateral can be used to mint a Philippine Peso stablecoin (called “UBE” for now). 

**How**

We’re in the process of wrapping up our own contracts that would then route DAI collateral to UMA to outsource the actual minting and redeeming flows.

Relevant to this issue, we propose to add CoinMarketCap’s and CoinGecko's DAI:PHP price feed (because CryptoWatch and the others did not seem to have a DAI:PHP specific quote) via their available endpoint for liquidator/disputer bot integration:

Sample CMC request: 
``` 
curl --location --request GET 'https://pro-api.coinmarketcap.com/v1/cryptocurrency/quotes/latest?symbol=DAI&convert=PHP&CMC_PRO_API_KEY=<free tier api key>'
```
This same request works with unit tests in our local branch. We ended up passing the API key as url parameter since we weren't sure how to modify the headers using UMA’s networker lib. But i guess we'll get some feedback as we go here. 

Sample CMC response:

```
 {
    "status": {
        "timestamp": "2021-01-07T03:59:50.118Z",
        "error_code": 0,
        "error_message": null,
        "elapsed": 17,
        "credit_count": 1,
        "notice": null
    },
    "data": {
        "DAI": {
            "id": 4943,
            "name": "Dai",
            "symbol": "DAI",
            "slug": "multi-collateral-dai",
            "num_market_pairs": 351,
            "date_added": "2019-11-22T00:00:00.000Z",
            "tags": [
                "defi",
                "stablecoin"
            ],
            "max_supply": null,
            "circulating_supply": 1316809873.5376725,
            "total_supply": 1316809899.3578176,
            "platform": {
                "id": 1027,
                "name": "Ethereum",
                "symbol": "ETH",
[...]

PR #2480, opened 2021-01-29
feat(price-feeds): adds CoinMarketCapPriceFeed & CoinGeckoPriceFeed for querying DAI:PHP & PHP:DAI price

schystz:
**Motivation**

PHP:DAI is not supported on CryptoWatch so we need to create a couple of price feeds to support this pair

**Summary**

First, CoinMarketCapPriceFeed.js & CoinGeckoPriceFeed.js is added which is based from CryptoWatchPriceFeed.js. Then we updated CreatePriceFeed.js to add options for the both price feeds. We also added a new tests (CoinMarketCapPriceFeed.js & CoinGeckoPriceFeed.js) and updated CreatePriceFeed.js test to make sure the new changes works and doesn't break existing codes.

**Details**

For the historical price, we simply store the time-price value pair every time update() is called, which is a different approach from CryptoWatchPriceFeed since it uses another API call to fetch the history. This is because of our CoinMarketCap API key plan limitation & CoinGecko doesn't have a history API available.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

To test the changes, simply cd to `packages/financial-templates-lib` then run the below test script:

```
npx hardhat test test/price-feed/CoinMarketCapPriceFeed.js test/price-feed/CoinGeckoPriceFeed.js test/price-feed/CreatePriceFeed.js

// or to test everything from financial-templates-lib package:
yarn truffle test --network test
```

**Issue(s)**

Fixes https://github.com/UMAprotocol/protocol/issues/2474


package.json
```diff
@@ -39,7 +39,7 @@
     "eslint-plugin-prettier": "^3.1.2",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
-    "ganache-cli": "^6.12.2",
+    "ganache-cli": "^6.10.2",
```
nicholaspai, line 42: Why downgrade?

package.json
```diff
@@ -51,9 +51,6 @@
     "solidity-docgen": "^0.5.3",
     "truffle": "^5.1.33"
   },
-  "resolutions": {
```
nicholaspai, line 54: Why this change?

.circleci/config.yml
```diff
@@ -98,6 +98,7 @@ jobs:
       - image: trufflesuite/ganache-cli
         command: ganache-cli -i 1234 -l 9000000 -p 9545
     working_directory: ~/protocol
+    resource_class: large
```
nicholaspai, line 101: I wouldn't recommend making changes to the `ci` files.

nicholaspai (changes requested):
Hey @schystz can you please rebase from master so I can better understand what new changes you are introducing?

schystz:
> Hey @schystz can you please rebase from master so I can better understand what new changes you are introducing?

Hi @nicholaspai, this is odd since we only added 4 files and changed 2 files. I think this is byproduct after we did the DCO suggested sign-off commands. We'll do the rebase to master to clarify the confusion, sorry about that!

schystz:
> > Hey @schystz can you please rebase from master so I can better understand what new changes you are introducing?
> 
> Hi @nicholaspai, this is odd since we only added 4 files and changed 2 files. I think this is byproduct after we did the DCO suggested sign-off commands. We'll do the rebase to master to clarify the confusion, sorry about that!

@nicholaspai had problems rebasing to master locally, I had to cherry pick the changes and combine them into a single commit. It produces the same result anyway but please let me know if this is fine.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.js
```diff
@@ -0,0 +1,158 @@
...
+
+  getPriceFeedDecimals() {
+    return this.priceFeedDecimals;
+  }
+
+  getLookback() {
+    return this.lookback;
+  }
+
+  _convertPriceFeedDecimals(number) {
```
chrismaree, line 136: to follow the pattern done in other price feeds, please define this method at the top within your constructor as `this.convertPriceFeedDecimals`. see: https://github.com/UMAprotocol/protocol/blob/9fcc5128fe3d684f4a87e2efa3d2e49934b96766/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L56

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
```diff
@@ -0,0 +1,168 @@
...
+
+  getPriceFeedDecimals() {
+    return this.priceFeedDecimals;
+  }
+
+  getLookback() {
+    return this.lookback;
+  }
+
+  _convertPriceFeedDecimals(number) {
```
chrismaree, line 146: same comment as in previous file.

packages/financial-templates-lib/test/price-feed/CoinGeckoPriceFeed.js
```diff
@@ -0,0 +1,226 @@
+const { CoinGeckoPriceFeed } = require("../../src/price-feed/CoinGeckoPriceFeed");
```
chrismaree, line 1: great tests!

chrismaree (commented):
This looks like an awesome start! good work on adding the tests and the two feeds. is the intention of this feed to use CoinMarketCap OR CoinGecko? For the Dai:Php pair?

You can also consider adding your default configuration to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js to auto-parameterize you feeds based on the contract identifier.

All products you intend to use with this feed are collateralized in Dai, right?

Can you link to the associated umip? I cant seem to find any umips that use DAI:PHP or any reference to PHP or Philippine peso.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.js
schystz: @chrismaree sure we can follow this convention. But out of curiosity, is there any particular reason why the function is defined in the constructor instead?

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
schystz: Noted @chrismaree !

schystz:
> This looks like an awesome start! good work on adding the tests and the two feeds. is the intention of this feed to use CoinMarketCap OR CoinGecko? For the Dai:Php pair?

Both actually! We plan to use medianizer price feed to get the median price on both CMC and CG:
```
{
  "type": "medianizer",
  "computeMean": true,
  "lookback": 120,
  "minTimeBetweenUpdates": 60,
  "medianizedFeeds": [{
    "type": "coinmarketcap",
    "apiKey": "c2b6c434-27dc-4d7c-b03d-96fa5413f0e9",
    "symbol": "DAI",
    "convert": "PHP"
  }, {
    "type": "coingecko",
    "contractAddress": "0x6b175474e89094c44da98b954eedeac495271d0f",
    "currency": "php"
  }]
}
```

> You can also consider adding your default configuration to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js to auto-parameterize you feeds based on the contract identifier.

Yeah we thought of adding the config here as well but its not in the official docs but now that you bring it up then we'll go right ahead :)

> All products you intend to use with this feed are collateralized in Dai, right?
> 
> Can you link to the associated umip? I cant seem to find any umips that use DAI:PHP or any reference to PHP or Philippine peso.

@ctverceles have we submitted the UMIP yet?

ctverceles:
> This looks like an awesome start! good work on adding the tests and the two feeds. is the intention of this feed to use CoinMarketCap OR CoinGecko? For the Dai:Php pair?
> 
> You can also consider adding your default configuration to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js to auto-parameterize you feeds based on the contract identifier.
> 
> All products you intend to use with this feed are collateralized in Dai, right?
> 
> Can you link to the associated umip? I cant seem to find any umips that use DAI:PHP or any reference to PHP or Philippine peso.

Thanks for the comments and review @chrismaree ! we haven't submitted it actually, let us do that now and comment here.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.js
schystz: Fixed :)

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
schystz: Fixed :)

schystz:
> You can also consider adding your default configuration to https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js to auto-parameterize you feeds based on the contract identifier.

@chrismaree we have added the default config for DAIPHP :)

schystz:
@chrismaree (cc: @ctverceles) here's the link to the UMIP: https://github.com/UMAprotocol/UMIPs/pull/160

schystz:
Hi reviewers (@nicholaspai @chrismaree), please let us know if you need anything else from our side to get the PR moving :) Thank you!

package.json
schystz: This is already outdated so resolving this issue.

package.json
schystz: This is already outdated so resolving this issue.

.circleci/config.yml
schystz: This is already outdated so resolving this issue.

packages/financial-templates-lib/test/price-feed/CoinGeckoPriceFeed.js
```diff
@@ -0,0 +1,226 @@
...
+    );
+  });
+
+  it("getCurrentPrice() returns the latest price", async function() {
+    networker.getJsonReturns = [validResponse];
+
+    await coinGeckoPriceFeed.update();
+
+    const price = coinGeckoPriceFeed.getCurrentPrice();
+    assert.equal(price.toString(), toWei(`${mockPrice}`));
```
chrismaree, line 55: Your tests are great and are pretty extensive! However, this set of tests does not validate that the feed can correctly deal with non-18 decimal place precision. could you please add an additional test that double-checks that the feed can handle a non-18 decimals config? I know your product is used dai and as such it should not matter for your synth but for the sake of completeness I'd like to see a test that validates this. 

To do this I think you should do the following in the unit test:
1) create an instance of the `coinGeckoPriceFeed` with `priceFeedDecimals=6` (for USDC, for example)
2) create a decimal conversion object to test the output. see here: https://github.com/UMAprotocol/protocol/blob/71515a915c9d796014295c18ebd6300229f24364/packages/monitors/test/ContractMonitor.js#L86 this basically wraps `parseFixed` in a nice way.
3) execute some of your previously tested logic, such as validating the `getCurrentPrice()` & `getHistoricalPrice()` methods.

Also, please add equivalent tests to the other feed included in this PR.

chrismaree (commented):
Sorry for the slow reply to this PR, have a lot of stuff flowing through our pipeline rn.

This PR is great! your code is simple, clean and well tested. You've replied to all of my comments. 

I've left one location where I'd like you to add one additional small test to do with decimal conversion. 

Once this is done, this is ready to be merged from my end.

packages/financial-templates-lib/test/price-feed/CoinGeckoPriceFeed.js
schystz: Hi @chrismaree, thanks for the kind words & feedback :)

We have added the unit tests you requested, though 1 note for the `Convert` util - we had to change the code a bit so that it matches the conversion logic used in the price feeds (which we have taken from `CryptoWatchPriceFeed`). Instead of the code above we ended up using:

```
const Convert = decimals => number => parseFixed(number.toString().substring(0, decimals), decimals).toString();
```

antphyohlyandtun (approved):

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.js
```diff
@@ -0,0 +1,158 @@
...
+    this.currentPrice = newPrice;
+    this.lastUpdateTime = currentTime;
+    this.priceHistory.push({ time: currentTime, price: newPrice });
+  }
+
+  getCurrentPrice() {
+    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+  }
+
+  getHistoricalPrice(time) {
```
nicholaspai, line 110: This should be async now. Look at the [CryptowatchPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L69) for an example implementation of an async getHistoricalPrice.

Additionally, instead of returning `undefined`, `getHistoricalPrice` should throw an error where a price can't be found. See the same example implementation.

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
```diff
@@ -0,0 +1,168 @@
...
+    this.currentPrice = newPrice;
+    this.lastUpdateTime = currentTime;
+    this.priceHistory.push({ time: currentTime, price: newPrice });
+  }
+
+  getCurrentPrice() {
+    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+  }
+
+  getHistoricalPrice(time) {
```
nicholaspai, line 120: See same comment about `getHistoricalPrice` that I made for the `CoinGeckoPriceFeed`. Thanks!

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
```diff
@@ -0,0 +1,168 @@
...
+   *      this number of seconds has passed, it will be a no-op.
+   * @param {Bool} invertPrice Indicates if prices should be inverted before returned.
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+   */
+  constructor(
+    logger,
+    web3,
+    apiKey,
+    symbol,
+    convert,
```
nicholaspai, line 25: Can we rename `convert` to `currency` since it looks like it has the same function as the `currency` value in `CoinGecko`? Even more accurate would be to rename both `currency` and `convert` to `quoteCurrency` to indicate that the price should be quoted in this currency

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -362,6 +362,26 @@ const defaultConfigs = {
       { type: "cryptowatch", exchange: "bitfinex", pair: "amplusd" }
     ]
   },
+  PHPDAI: {
+    type: "medianizer",
+    computeMean: true,
+    lookback: 7200,
+    minTimeBetweenUpdates: 60,
+    medianizedFeeds: [
```
nicholaspai, line 370: This is a really cool price feed!

nicholaspai (commented):
Looks great, left some comments about re-implementing `getHistoricalPrice` to match the updated interface.

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
schystz: @nicholaspai renamed both `CMC.convert` & `CG.currency` to `quoteCurrency`

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.js
schystz: @nicholaspai took care of this on the latest commit https://github.com/UMAprotocol/protocol/pull/2480/commits/05822407a7026f1fbfac48b2627c935c84961c61

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.js
schystz: @nicholaspai we have updated the implementation (including tests) to match the updated interface. Please see https://github.com/UMAprotocol/protocol/pull/2480/commits/05822407a7026f1fbfac48b2627c935c84961c61

schystz:
Hi @nicholaspai , thanks for your feedback! We have addressed all of the concerns in the latest commit, please see https://github.com/UMAprotocol/protocol/pull/2480/commits/05822407a7026f1fbfac48b2627c935c84961c61

nicholaspai (approved):
Thanks for implementing my requested changes, I've started the test CI suite again and LGTM assuming everything passes!

chrismaree (approved):
This is awesome & LGTM!

PR #2484, opened 2021-01-30
feat(trader): start trader package with typescript support

mrice32:
**Motivation**

We'd like to build a simple trading bot for keeping perp prices in line. This also seemed like a good opportunity to begin adding ts support!

**Summary**

Added a trader package with some simple ts stub code. This also adds a build process to automatically generate type definitions for packages in the repo that we expect the trader package to use.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


daywiss (approved):
nice work on this, looks good. Can probably wait till a later commit, but I am curious on how you would run the project. Would be good to have an npm script or some docs on that.

mrice32:
> nice work on this, looks good. Can probably wait till a later commit, but I am curious on how you would run the project. Would be good to have an npm script or some docs on that.

@daywiss great point. Will add docs in the package, but the idea is to just run `dist/index.js`. The only difference from the other packages is that 1) it will require a compilation before it's runnable and 2) that the "main" will point to the dist dir rather than a top-level index.js.

PR #2487, opened 2021-02-01
feature(core): Create voting script for GASETH-TWAP-1Mx1M price identifier

smb2796:
**Motivation**

This script will allow UMA voters to query for median 30 day gas prices through Google Big Query.

**Summary**

To run this price feed locally, you need to set a GOOGLE_APPLICATIONS_CREDENTIALS environment variable with GCP admin or BigQuery permissions. 

The BQ query was included in a separate file for readability.

The following packages are also added to core/package.json.
- [highland](https://www.npmjs.com/package/highland)
- [google-cloud/bigquery](https://www.npmjs.com/package/@google-cloud/bigquery)

At a later date, this script should be generalized to accept any GASETH price identifier or any other price identifier that uses Google BQ.


packages/core/scripts/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,103 @@
...
+      callback(err);
+      return;
+    }
+
+    // Scaling by one million as specified in the GASETH-TWAP-1Mx1M UMIP
+    queryPrice = queryPrice * 1000000;
+
+    console.log(`\n⚠️ Truncating price to ${precisionToUse} decimals`);
+    console.log(
+      `\n💹 Median ${queryIdentifier} price @ ${queryTime} = ${Number(fromWei(queryPrice.toString())).toFixed(
```
smb2796, line 75: `toFixed` introduces a slight rounding error when scaling the price identifier. As an example, a return value of `.086` is changed to `0.085999999999999993`. Any idea why this is?

packages/core/scripts/local/GasEthQuery.js
```diff
@@ -0,0 +1,62 @@
...
+  -- Querying for the amount of blocks in the preset time range. This will allow block_count to be compared against a given minimum block amount.
+  SET (block_count, max_block) = (SELECT AS STRUCT (MAX(number) - MIN(number)), MAX(number) FROM \`bigquery-public-data.crypto_ethereum.blocks\` 
+  WHERE timestamp BETWEEN TIMESTAMP('${earlierTimeBound}', 'UTC') AND TIMESTAMP('${laterTimeBound}', 'UTC'));
+
+  CREATE TEMP TABLE cum_gas (
+    gas_price int64,
+    cum_sum int64
+  );
+
+  -- If the minimum threshold of blocks is met, query on a time range
```
chrismaree, line 17: this is the same bigquery we had back when you first designed the uGAS feeds, right? before moving over to the twap pattern?

packages/core/scripts/local/getGasEthHistoricalPrice.js
chrismaree: I suspect this is because your `precisionToUse` is set to 18. normally this value is closer to 5. you might need to do the `toFixed` logic differently seeing that javascript does not work well with this kind of tiny number.

One option would be to use the native bignumber `.toFixed` which will properly work. ([docs](https://mikemcl.github.io/bignumber.js/#toFix)). To do this try out something like:

```
toBN(fromWei(queryPrice.toString())).toFixed(precisionToUse)
```

where you convert to a BN rather than to a `Number`.

packages/core/scripts/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,103 @@
+/**
+ * @notice This is an example script demonstrating how to get a historical median gas price at a specific timestamp.
+ * It could be used for example to query a GASETH-TWAP-1Mx1M price before committing a vote for a DVM price request.
+ *
+ * @dev Prequisites: before running, you will need to set a GOOGLE_APPLICATIONS_CREDENTIALS environment variable for a service account.
+ * @dev This service account will need GCP admin or BigQuery permissions. This guide provides further instructions: https://cloud.google.com/docs/authentication/getting-started
+ * @dev How to run: yarn truffle exec ./scripts/local/getGasEthHistoricalPrice.js --network mainnet_mnemonic --time 1612137600
```
chrismaree, line 7: nit: should the example contain `identifier`?

chrismaree (approved):
This looks good to me! I had a nit and a few questions but overall I'm happy :)

packages/core/scripts/local/GasEthQuery.js
smb2796: Yes it is the same.

packages/core/scripts/local/getGasEthHistoricalPrice.js
smb2796: Hm yeah probably. Was initially thinking this was only usable for one identifier, but it would also work for the `uGAS-FEB21` and `uGAS-MAR21` identifiers, so I will include.

packages/core/scripts/local/getGasEthHistoricalPrice.js
smb2796: Cool, thank you. I addressed with

`BigNumber(fromWei(queryPrice.toString())).toFixed(precisionToUse)`

nicholaspai (approved):
LGTM

packages/core/scripts/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,110 @@
...
+    // If user specified a timestamp, then use it, otherwise default to the current time.
+    let queryTime;
+    if (!argv.time) {
+      queryTime = getTime();
+      console.log(
+        `Optional '--time' flag not specified, defaulting to the current Unix timestamp (in seconds): ${queryTime}`
+      );
+    } else {
+      queryTime = argv.time;
+    }
```
pemulis, line 44: Is there ever a situation where someone would want to query with the current time and not a specified timestamp?

packages/core/scripts/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,110 @@
...
+    queryTime = new Date(queryTime * 1000);
+    // Using moment package to convert queryTime to a usable BQ UTC format.
+    let laterTimeBound = moment(queryTime)
+      .utc()
+      .format(dateFormat);
+
+    // Subtracting 30 days from the current time to give the earlier time bound.
+    let earlierTimeBound = moment(queryTime).subtract(30, "days");
+
+    // Creates format that is easier to read for loggin
```
pemulis, line 57: `logging`

packages/core/scripts/local/getGasEthHistoricalPrice.js
nicholaspai: Yeah for sanity checking or a quick run for example

nicholaspai:
@smb2796 let's merge this after you merge in master!

packages/core/package.json
```diff
@@ -32,6 +32,9 @@
     "truffle-assertions": "^0.9.2",
     "truffle-deploy-registry": "^0.5.1",
     "web3": "1.2.11",
+    "highland": "^2.13.5",
+    "@google-cloud/bigquery": "^5.3.0",
+    "bignumber.js": "^8.0.1"
```
smb2796, line 37:
```suggestion
    "bignumber.js": "^8.0.1",
```

PR #2489, opened 2021-02-01
ci(slither): remove slither from ci

mrice32:
**Motivation**

Slither tends to cause more problems than it solves in ci. We run it in a separate docker image and it consistently breaks when interacting with our toolchain. It also doesn't seem to provide a lot of value -- by my understanding, it has never caught a smart contract bug -- it has caught some minor stylistic or readability issues.

**Summary**

This PR removes slither from ci, but leaves the run script in place in case we want to run it in the future.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,38 +60,6 @@ jobs:
           command: ./scripts/build_docs_site.sh
       - store_artifacts:
           path: build/site
-  slither:
-    docker:
-      - image: trailofbits/eth-security-toolbox
-    working_directory: ~/protocol
-    steps:
-      - checkout
-      - restore_cache:
-          keys:
-            - v7-sec-toolbox-deps-{{ checksum "yarn.lock" }}
-            - v7-sec-toolbox-deps-
-      - run:
-          name: Install node and NPM
-          command: ./ci/install_node_npm.sh
-      - run:
-          name: Change user
-          command: sudo su ethsec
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
LGTM but I'm curious what you have in mind for a substitute (if any) to detect common contract bugs? Also there is a `run_slither.sh` file that we can remove.

mrice32:
> LGTM but I'm curious what you have in mind for a substitute (if any) to detect common contract bugs? Also there is a `run_slither.sh` file that we can remove.

@nicholaspai I left the file just so we could run it manually if we wanted. Does that seem reasonable?

As for a substitute, I don't really have one, but I haven't found slither to be useful for catching any smart contract bugs. Have you found it to be useful? Has it caught real issues for you?

nicholaspai:
> > LGTM but I'm curious what you have in mind for a substitute (if any) to detect common contract bugs? Also there is a `run_slither.sh` file that we can remove.
> 
> @nicholaspai I left the file just so we could run it manually if we wanted. Does that seem reasonable?
> 
> As for a substitute, I don't really have one, but I haven't found slither to be useful for catching any smart contract bugs. Have you found it to be useful? Has it caught real issues for you?

Ah makes sense to leave that file. I like slither but removing it from CI makes sense.

PR #2492, opened 2021-02-02
docs: add MAINTAINING.md

mrice32:
**Motivation**

To keep the repos organized, we need to add an on-duty maintainer structure.

**Summary**

Added `MAINTAINING.md`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/MAINTAINING.md
+++ b/MAINTAINING.md
@@ -1,0 +1,24 @@
+# Maintaining UMA Repositories
+
+As our repos grow in activity, it takes more and more effort to maintain them and ensure code and issues are being
+actively triaged for consideration and review. This document outlines the responsibilities of the on-duty maintainer.
+
+## Responsibilities
+
+- Every business day, the on-duty maintainer is expected to make a pass over _all_ open PRs and new issues in _all_
+  UMA repositories.
+  - On-duty maintainers rotate each week in a round robin fashion
+- For each PR, the on-duty maintainer should make sure the PR is moving forward:
+  - Check the title and description according to the [template](./.github/PULL_REQUEST_TEMPLATE.md). If either of
+    these are incorrect, edit them and add a comment explaining your edits.
+  - Make sure reviews are requested. If not, request them (if the PR is very small, feel free to just review it
+    yourself).
+  - If the PR has stalled, determine whose action is needed and tag them in a comment (or re-request a review if it's
+    the reviewer). If a 3rd party's PR is ready and passing tests, merge it.
+  - DCO is a common problem for many users, please help them by pointing them to the
+    [DCO docs](https://github.com/apps/dco). Note: DCO checks can be overriden for UMA org members because DCO is covered
+    by other agreements, but 3rd party developers _must_ sign off on their commits.
+  - For the full PR process, see [the contributing guidelines](./CONTRIBUTING.md).
+- For each _new_ issue:
+  - If it's a 3rd party issue, traige by assigning or tagging a relevant UMA team member.
+  - If it was posted by an UMA org member, no action is required.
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):
LGTM. Is there a way for GH to manage the rotation of the on-call maintainer? This way an external PR could request the "maintainer" role and this would auto select the on call person.

MAINTAINING.md
```diff
@@ -0,0 +1,23 @@
+# Maintaining UMA Repositories
+
+As our repos grow in activity, it takes more and more effort to maintain them and ensure code and issues are being
+actively triaged for consideration and review. This document outlines the responsibilities of the on-duty maintainer.
+
+## Responsibilities
+
+- Every business day, the on-duty maintainer is expected to make a pass over _all_ open PRs and new issues in _all_
+  UMA repositories.
```
chrismaree, line 9: should this not just apply to the protocol repo? the rules & process on the UMIP/front end repo will properly need to be different, don't you think?

MAINTAINING.md
```diff
@@ -0,0 +1,23 @@
+# Maintaining UMA Repositories
```
chrismaree, line 1: do you want to add a section that mentions how we cycle between weeks of on-call for maintaining? i.e "each eng is on-call for one week in a round robbin fashion" or should this be left off gh?

chrismaree (approved):
Looks good! had 2 questions but otherwise I'm happy :)

mrice32:
> LGTM. Is there a way for GH to manage the rotation of the on-call maintainer? This way an external PR could request the "maintainer" role and this would auto select the on call person.

@nicholaspai That would be so cool. Will look into this.

MAINTAINING.md
mrice32: It's a good question. Maybe it shouldn't apply to UMIPs repo, but my feeling is that it _should_ apply to EMP-tools, EMP-launch, docs, etc? Maybe we just mention the exceptions here? Or do you think there's a better way to organize it?

mrice32:
> > LGTM. Is there a way for GH to manage the rotation of the on-call maintainer? This way an external PR could request the "maintainer" role and this would auto select the on call person.
> 
> @nicholaspai That would be so cool. Will look into this.

@nicholaspai this might do the trick: https://github.com/marketplace/actions/team-rotation

MAINTAINING.md
Tommy1231232: We have a separate process managing the UMIP repo so it makes sense to exclude it. I think all of the other repos are applicable. E.g., if it's a doc PR then tag a non-eng to work on it

MAINTAINING.md
```diff
@@ -0,0 +1,23 @@
...
+
+As our repos grow in activity, it takes more and more effort to maintain them and ensure code and issues are being
+actively triaged for consideration and review. This document outlines the responsibilities of the on-duty maintainer.
+
+## Responsibilities
+
+- Every business day, the on-duty maintainer is expected to make a pass over _all_ open PRs and new issues in _all_
+  UMA repositories.
+- For each PR, the on-duty maintainer should make sure the PR is moving forward:
```
Tommy1231232, line 11: It seems like the ENG team being assigned to PRs and making sure they are continually moving forward is time consuming and takes away from items in the existing sprint. Would it be useful to track the work ENG is doing on these PRs in our existing sprint?

mrice32: It definitely takes time, but I'm not sure how useful it will be to explicitly track the review work in the sprint. I'm definitely open to the idea, though -- what sort of tracking process would you propose?

MAINTAINING.md
mrice32: I don't know if it should be managed off-github, but I think this doc shouldn't prescribe the rotation. IMO, that should be managed in whatever tool/doc we use to actually do the rotation. I think this doc should just be able to point to somewhere you can go to figure out who the current on-call maintainer is.

pemulis:
Anything we're waiting on here before merging @mrice32?

MAINTAINING.md
chrismaree: ye, I think it best to scope this to just protocol for now.

PR #2493, opened 2021-02-02
feat(financial-templates-lib): make getHistoricalPrice async

mrice32:
**Motivation**

Some price feeds cannot do a full 2 hour lookup during `update()` as it would be too inefficient.

**Summary**

Make the standard `getHistoricalPrice` fn async.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/disputer/src/disputer.js
+++ b/packages/disputer/src/disputer.js
@@ -119,7 +119,7 @@ class Disputer {
             price = this.toBN(disputerOverridePrice);
           } else {
             try {
-              price = this.priceFeed.getHistoricalPrice(liquidationTime);
+              price = await this.priceFeed.getHistoricalPrice(liquidationTime);
             } catch (error) {
               this.logger.error({
                 at: "Disputer",
--- a/packages/disputer/src/disputer.js
+++ b/packages/disputer/src/disputer.js
@@ -94,57 +94,59 @@ class Disputer {
 
     // Get the latest disputable liquidations from the client.
     const undisputedLiquidations = this.empClient.getUndisputedLiquidations();
-    const disputeableLiquidations = await Promise.all(
-      undisputedLiquidations.map(async liquidation => {
-        // If liquidation time is before the price feed's lookback window, then we can skip this liquidation
-        // because we will not be able to get a historical price. If a dispute override price is provided then
-        // we can ignore this check.
-        const liquidationTime = parseInt(liquidation.liquidationTime.toString());
-        const historicalLookbackWindow =
-          Number(this.priceFeed.getLastUpdateTime()) - Number(this.priceFeed.getLookback());
-        if (!disputerOverridePrice && liquidationTime < historicalLookbackWindow) {
-          this.logger.debug({
-            at: "Disputer",
-            message: "Cannot dispute: liquidation time before earliest price feed historical timestamp",
-            liquidationTime,
-            historicalLookbackWindow
-          });
-          return null;
-        }
-
-        // If an override is provided, use that price. Else, get the historic price at the liquidation time.
-        let price;
-        if (disputerOverridePrice) {
-          price = this.toBN(disputerOverridePrice);
-        } else {
-          try {
-            price = this.priceFeed.getHistoricalPrice(liquidationTime);
-          } catch (error) {
-            this.logger.error({
+    const disputeableLiquidations = (
+      await Promise.all(
+        undisputedLiquidations.map(async liquidation => {
+          // If liquidation time is before the price feed's lookback window, then we can skip this liquidation
+          // because we will not be able to get a historical price. If a dispute override price is provided then
+          // we can ignore this check.
+          const liquidationTime = parseInt(liquidation.liquidationTime.toString());
+          const historicalLookbackWindow =
+            Number(this.priceFeed.getLastUpdateTime()) - Number(this.priceFeed.getLookback());
+          if (!disputerOverridePrice && liquidationTime < historicalLookbackWindow) {
+            this.logger.debug({
               at: "Disputer",
-              message: "Cannot dispute: price feed returned invalid value",
-              error: error.message
+              message: "Cannot dispute: liquidation time before earliest price feed historical timestamp",
+              liquidationTime,
+              historicalLookbackWindow
             });
             return null;
           }
-        }
 
-        // Price is available, use it to determine if the liquidation is disputable
-        if (
-          this.empClient.isDisputable(liquidation, price) &&
-          this.empClient.getLastUpdateTime() >= Number(liquidationTime) + this.disputeDelay
-        ) {
-          this.logger.debug({
-            at: "Disputer",
-            message: "Detected a disputable liquidation",
-            price: price.toString(),
-            liquidation: JSON.stringify(liquidation)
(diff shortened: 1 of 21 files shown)


nicholaspai (commented):
Couple points
- Try rebasing from master now that I've merged #2426 
- Does this mean that the `disputer` no longer needs to call `pricefeed.update`? If so, where else can we make this optimization

mrice32:
> Couple points
> 
> * Try rebasing from master now that I've merged #2426

Will do!

> * Does this mean that the `disputer` no longer needs to call `pricefeed.update`? If so, where else can we make this optimization

No. At least not right now. This does not change the expected flow (or even the implementation of the feeds), it just adds extra flexibility that the feed may choose to do its async query inside the historical call rather than in update, but it still _can_ query the data up front in `update()`. Removing update() is a different (and probably much more invasive) change. It totally could make sense, but that's not the aim of this PR.

packages/disputer/src/disputer.js
```diff
@@ -94,50 +94,52 @@ class Disputer {
...
+            });
+            return liquidation;
+          } else {
+            return null;
+          }
         }
-      }
-    });
+      })
+    ).filter(liquidation => liquidation !== null);
```
daywiss, line 142: i dont think this will work right unless you've imported bluebird promises or wrap 
`(await Promise.all(...)).filter(x=>x)`. Maybe just double check this

daywiss:
> No. At least not right now. This does not change the expected flow (or even the implementation of the feeds), it just adds extra flexibility that the feed may choose to do its async query inside the historical call rather than in update, but it still can query the data up front in update(). Removing update() is a different (and probably much more invasive) change. It totally could make sense, but that's not the aim of this PR.

It seems like now theres no way to avoid double calling the historical data, since the history calls aren't removed from `update`, and theres also no way to grab the historical data cache if you choose to only call `update`.  Am I missing something here?

packages/financial-templates-lib/test/price-feed/MedianizerPriceFeed.js
```diff
@@ -109,14 +109,14 @@ contract("MedianizerPriceFeed.js", function() {
...
     const arbitraryHistoricalTimestamp = 1000;
-    try {
-      medianizerPriceFeed.getHistoricalPrice(arbitraryHistoricalTimestamp);
-    } catch (err) {
-      assert.equal(err[0].message, "PriceFeedMock expected error thrown");
-      assert.equal(err[1].message, "PriceFeedMock expected error thrown");
-      assert.equal(err.length, 2);
-    }
-    assert.throws(() => medianizerPriceFeed.getHistoricalPrice(arbitraryHistoricalTimestamp));
+    await medianizerPriceFeed.getHistoricalPrice(arbitraryHistoricalTimestamp).then(
```
mrice32, line 112: @nicholaspai .then syntax actually makes this test a little simpler IMO

packages/financial-templates-lib/test/price-feed/DominationFinancePriceFeed.js
```diff
@@ -208,9 +208,9 @@ contract("DominationFinancePriceFeed.js", function() {
     assert.isTrue(await invertedPriceFeed.update().catch(() => true), "Update didn't throw");
 
     assert.equal(priceFeed.getCurrentPrice(), undefined);
-    assert.throws(() => priceFeed.getHistoricalPrice(earliestTick));
+    await priceFeed.getHistoricalPrice(earliestTick).then(() => assert.fail());
```
mrice32, line 211: @nicholaspai this is the syntax that we have to use to replicate `.throws` since `.throws` doesn't handle async functions (at least I don't _think_ it does). It's a little harder to understand, but at least it's not super painful to write.

packages/disputer/src/disputer.js
mrice32: Nice catch -- was just re-reading it and totally agree. Need to wrap in parens.

mrice32: Added parens -- PTAL.

mrice32:
> > No. At least not right now. This does not change the expected flow (or even the implementation of the feeds), it just adds extra flexibility that the feed may choose to do its async query inside the historical call rather than in update, but it still can query the data up front in update(). Removing update() is a different (and probably much more invasive) change. It totally could make sense, but that's not the aim of this PR.
> 
> It seems like now theres no way to avoid double calling the historical data, since the history calls aren't removed from `update`, and theres also no way to grab the historical data cache if you choose to only call `update`. Am I missing something here?
> 
> Edit: sorry i mis understood the changes. It seems the logic is the same but the function sig just has async added

exactly. There's no real logic change here. It just adds flexibility for the price feed implementor. I could see an argument that we should drop update down the road and let the price feed just internally cache after calls to its price request functions, but that's not something I'm ready to tackle right now :).

daywiss (approved):
👍

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -109,31 +109,23 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
...
-    });
-    let baselinePrices = this.baselinePriceFeeds.map(priceFeed => {
-      try {
-        return priceFeed.getHistoricalPrice(time);
-      } catch (err) {
-        errors.push(err);
-      }
-    });
+    const errors = [];
+    const experimentalPrices = await Promise.all(
```
nicholaspai, line 116: +1 more succinct .catch syntax

packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -101,13 +101,16 @@ contract("CryptoWatchPriceFeed.js", function() {
     networker.getJsonReturns = [...validResponses];
     await invertedCryptoWatchPriceFeed.update();
 
-    // Before period 1 should return null.
-    assert.throws(() => invertedCryptoWatchPriceFeed.getHistoricalPrice(1588376339));
+    // Before period 1 should fail.
+    await invertedCryptoWatchPriceFeed.getHistoricalPrice(1588376339).then(
+      () => assert.fail(),
+      () => {}
```
nicholaspai, line 107: Can you explain what's happening here?

nicholaspai: + all the other assert.throws changes you made

nicholaspai: Ok I think I see what you're doing. Instead of returning `() => {}` can you make this more explicitly return `.catch() => assert.succeeds()` or something like that?

nicholaspai (approved):
Looks good to me, lots of small changes but I mostly think that some of the `assert.throws()` refactors are a bit mysterious to read now.

mrice32:
> Looks good to me, lots of small changes but I mostly think that some of the `assert.throws()` refactors are a bit mysterious to read now.

Swapped the syntax for `.catch(() => true)`.

PR #2502, opened 2021-02-03
fix(FindContractVersion): address inconsistent provider behavior

chrismaree:
**Motivation**

`FindContractVersion.js` depends on ethers.js. The previous implementation only worked with `http` and `https` web3 providers and not `wss` ones. This means that the script broke in production, where we use WebSockets. 

This PR changes the implementation to rely on the existing web3 provider with the ethers `Web3Provider` provider.

**Summary**

**Testing**

- [ ]  New unit tests created
- [X]  Untested


daywiss (approved):
good find!

nicholaspai (approved):

Issue #2411, opened 2021-01-13
Build stub for Optimistic Oracle Keeper which follows strategy to respond to price requests


nicholaspai:
**Why**

OO Keeper should pull OO contract state from the OOClient and respond to:
- unproposed price requests
- undisputed proposals

The OO Keeper should handle fetching offchain price indications for any supported identifier (i.e. leveraging the `financial-templates-lib/price-feed` lib) and proposing prices for price requests, and deciding whether to dispute ongoing proposals

Follow up issues will address:
- executing settlements of proposals + disputes

**Work**
- 6: ~1 day

PR #2505, opened 2021-02-04
feat(optimistic-oracle-keeper): Add functionality to propose prices for price requests

nicholaspai:
**Motivation**

#2411

**Summary**

Subsequent PR will add dispute logic for pending proposals.

**Details**

- Keeper bot constructs a new PriceFeed for each pending price request identifier and uses it to send price proposals
- The Keeper bot uses the default precision for the identifier from `@uma/common` and the default pricefeed config from @uma/common
- Add test default pricefeed configs and precision mappings in `@uma/common` to test situation where there are several price requests that require different precisions

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Partial solution to #2411, does not add dispute or settlement logic.


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -62,7 +64,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       getTime,
       config.minTimeBetweenUpdates,
       config.invertPrice, // Not checked in config because this parameter just defaults to false.
-      config.decimals, // This defaults to 18 unless supplied by user
+      config.priceFeedDecimals, // This defaults to 18 unless supplied by user
```
nicholaspai, line 67: This was an error for the DomFi price feed. The file `DefaultPriceFeedConfig` sets the `priceFeedDecimals` param not `decimals`:

```
// Pull in the number of decimals for each identifier from the common getPrecisionForIdentifier. This is used within the
// Voterdapp and ensures that price feeds are consistently scaled through the UMA ecosystem.
Object.keys(defaultConfigs).forEach(identifierName => {
  defaultConfigs[identifierName].priceFeedDecimals = getPrecisionForIdentifier(identifierName);
});
```

Luckily this is not an issue in prod since the default decimal is 18 which the DomFi EMP's use

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -162,6 +164,34 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.denominatorPriceFeed && (await _createMedianizerPriceFeed(config.denominatorPriceFeed));
 
     return new BasketSpreadPriceFeed(web3, logger, baselinePriceFeeds, experimentalPriceFeeds, denominatorPriceFeed);
+  } else if (config.type === "test") {
```
nicholaspai, line 167: The `test` and `invalid` types are used to test that the OO Keeper uses `CreatePriceFeed` correctly to create new price feeds.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -361,6 +361,20 @@ const defaultConfigs = {
...
+  TEST8DECIMALS: {
+    type: "test"
+  },
+  TEST6DECIMALS: {
+    type: "test"
+  },
+  TEST18DECIMALS: {
+    type: "test"
+  },
+  INVALID: {
```
nicholaspai, line 376: These are used by the CreatePriceFeed file to map between test identifiers and the Mock PriceFeeds

packages/financial-templates-lib/src/price-feed/PriceFeedMockScaled.js
```diff
@@ -2,74 +2,89 @@ const { toBN } = require("web3").utils;
 const { PriceFeedInterface } = require("./PriceFeedInterface");
```
nicholaspai, line 2: This file has a lot of diffs because I basically copy and pasted the `PriceFeedMock` code here but added the final conversion step. This file is not currently used in any other tests.

packages/optimistic-oracle/test/keeper.js
```diff
@@ -0,0 +1,305 @@
...
+    await keeper.update();
+    await keeper.sendProposals();
+
+    // Check for the specific failure log emitted by the keeper.
+    assert.equal(lastSpyLogLevel(spy), "error");
+    assert.isTrue(spyLogIncludes(spy, -1, "Failed to construct a PriceFeed for price request"));
+  });
+
+  it("Skip price requests with historical prices that keeper fails to fetch", async function() {
+    // Request a valid identifier but set an invalid price feed config.
```
pemulis, line 276: This is worded in a way that's a bit confusing, since it sounds the same as the earlier invalid price feed config. Maybe something like:

`// Request a valid identifier that is getting bad data from the data source`

packages/financial-templates-lib/src/price-feed/PriceFeedMock.js
```diff
@@ -1,12 +1,6 @@
 const { PriceFeedInterface } = require("./PriceFeedInterface");
 const { toBN } = require("web3").utils;
-
-// An implementation of PriceFeedInterface that medianizes other price feeds.
 class PriceFeedMock extends PriceFeedInterface {
-  // Constructs the MedianizerPriceFeed.
```
chrismaree, line 6: +1

packages/optimistic-oracle/index.js
```diff
@@ -4,6 +4,7 @@ require("dotenv").config();
 const retry = require("async-retry");
 
 const { Logger, waitForLogger, delay, OptimisticOracleClient, GasEstimator } = require("@uma/financial-templates-lib");
+const { OptimisticOracleKeeper } = require("./src/keeper");
```
chrismaree, line 7: do we want to call this a `Keeper`? would it not be more consistent to call it `OptimisticOracleProposer` or something to that effect so it's consistent with the other packages?

packages/optimistic-oracle/index.js
```diff
@@ -52,13 +48,29 @@ async function run({ logger, web3, pollingDelay, errorRetries, errorRetriesTimeo
...
+    // Construct default price feed config passed to all pricefeeds constructed by the keeper.
+    // The keeper needs to query prices for any identifier approved to use the Optimistic Oracle,
+    // so a new pricefeed is constructed for each identifier. This `defaultPriceFeedConfig` contains
+    // properties that are shared across all of these new pricefeeds.
+    const defaultPriceFeedConfig = {
+      lookback: 7200 // Should we pass in this object and/or its properties as input into the `run()` method?
+    };
+    const ooKeeper = new OptimisticOracleKeeper({
+      logger,
+      optimisticOracleClient: ooClient,
```
chrismaree, line 60: nit: might be cleaner to just name the local var `optimisticOracleClient` and then you don't need to name it like this in the params.

packages/optimistic-oracle/index.js
```diff
@@ -106,7 +118,12 @@ async function Poll(callback) {
       // Default to 3 re-tries on error within the execution loop.
       errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
       // Default to 10 seconds in between error re-tries.
-      errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 10
+      errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 10,
```
chrismaree, line 121: This was changed in a recent PR.

```suggestion
      errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 1,
```

packages/optimistic-oracle/index.js
```diff
@@ -106,7 +118,12 @@ async function Poll(callback) {
       // Default to 3 re-tries on error within the execution loop.
       errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
```
chrismaree, line 119:
```suggestion
      errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 3,
```

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
+const { Networker, createReferencePriceFeedForEmp } = require("@uma/financial-templates-lib");
+const { getPrecisionForIdentifier, createObjectFromDefaultProps, MAX_UINT_VAL } = require("@uma/common");
+const { getAbi } = require("@uma/core");
+
+class OptimisticOracleKeeper {
```
chrismaree, line 5: nit about this name, as mentioned in an above comment. Also, Ideally, the class name should match the name of the file. What do you think about calling it `OptimisticOracleProposer`?

Keeper might be a better name as it does both proposals, disputes and requests. if you want to keep it named as such then please match the file name to the class name.

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+
+    // Increase allowances for all relevant collateral currencies.
+    // TODO: Consider whether this should happen in a separate function (i.e. `setAllowances`) or within
+    // the `sendProposals()` method.
+    await this._setAllowances();
+  }
+
+  // Sets allowances for all collateral currencies used in unproposed price requests
+  async _setAllowances() {
+    // TODO: Should we do this also for currencies from undisputedProposals?
```
chrismaree, line 70: I think it's fine to do bulk approval like this for now. When we move over to using DSProxy we can do the approval in the same tx that we sent the request/proposal.

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+          currency: collateralToken.options.address,
+          collateralApprovalTx: collateralApprovalTx.transactionHash
+        });
+      }
+    }
+  }
+
+  // Create the pricefeed for a specific identifier and save it to the state, or
+  // return the saved pricefeed if already constructed.
+  async _createOrGetCachedPriceFeed(identifier) {
```
chrismaree, line 97: nit: do you want to put these internal methods at the bottom of the class?

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+          .approve(this.ooContract.options.address, MAX_UINT_VAL)
+          .send({
+            from: this.account,
+            gasPrice: this.gasEstimator.getCurrentFastPrice()
+          });
+        this.logger.info({
+          at: "OptimisticOracle#Keeper",
+          message: "Approved OO to transfer unlimited collateral tokens 💰",
+          currency: collateralToken.options.address,
+          collateralApprovalTx: collateralApprovalTx.transactionHash
```
chrismaree, line 89: do you want to include the tx in this message?

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+      // Create the transaction.
+      const proposal = this.ooContract.methods.proposePrice(
+        priceRequest.requester,
+        this.utf8ToHex(priceRequest.identifier),
+        priceRequest.timestamp,
+        "0x", // Ancillary data
+        proposalPrice
+      );
+
+      // Simple version of inventory management: simulate the transaction and assume that if it fails, the caller didn't have enough collateral.
```
chrismaree, line 175: nit: line length.

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+
+  // Submit proposals to unproposed price requests.
+  async sendProposals() {
+    this.logger.debug({
+      at: "OptimisticOracleKeeper",
+      message: "Checking for unproposed price requsts to send proposals for"
+    });
+
+    for (let priceRequest of this.optimisticOracleClient.getUnproposedPriceRequests()) {
+      let priceFeed = await this._createOrGetCachedPriceFeed(priceRequest.identifier);
```
chrismaree, line 139: nice! cool how you construct the feed.

do you want to make this a `const`?

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+        });
+        continue;
+      }
+
+      // Create the transaction.
+      const proposal = this.ooContract.methods.proposePrice(
+        priceRequest.requester,
+        this.utf8ToHex(priceRequest.identifier),
+        priceRequest.timestamp,
+        "0x", // Ancillary data
```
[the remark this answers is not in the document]
chrismaree: Should this not be set to the value of the ancillary data from the request generated from the OO?

In the case of the EMP this is set to the address of the collateral currency. see https://github.com/UMAprotocol/protocol/blob/c79f3476fee07736257582f7d92eb7c95932c300/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L797 for where we request the price in the EMP and https://github.com/UMAprotocol/protocol/blob/c79f3476fee07736257582f7d92eb7c95932c300/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L997 for where we generate this ancillary data.

packages/optimistic-oracle/test/keeper.js
```diff
@@ -0,0 +1,305 @@
...
+
+const { toWei, hexToUtf8, utf8ToHex } = web3.utils;
+
+const {
+  OptimisticOracleClient,
+  GasEstimator,
+  SpyTransport,
+  lastSpyLogLevel,
+  spyLogIncludes,
+  PriceFeedMockScaled
```
chrismaree, line 12: I dont think you should be using the `PriceFeedMockScaled` anywhere within these tests like this. IMO this file should actually be deprecated as it's somewhat confusing and is not used anywhere else within the repo after the most recent refactors.

Ideally, this set of tests should just use normal feeds without any kind of scaling like this expected to be applied to them.

packages/optimistic-oracle/test/keeper.js
```diff
@@ -0,0 +1,305 @@
...
+      }
+      let result = client.getUnproposedPriceRequests();
+      assert.deepStrictEqual(result, expectedResults);
+
+      // Now: Execute `sendProposals()` and test that the bot correctly responds to these price proposals
+      await keeper.sendProposals();
+
+      // Check that the onchain requests have been proposed to.
+      for (let i = 0; i < identifiersToTest.length; i++) {
+        await verifyState(OptimisticOracleRequestStatesEnum.PROPOSED, identifiersToTest[i]);
```
chrismaree, line 209: nice!

packages/optimistic-oracle/test/keeper.js
```diff
@@ -0,0 +1,305 @@
...
+          "0x",
+          collateralCurrenciesForIdentifier[i].address,
+          0
+        );
+      }
+
+      // Construct OO Keeper using a valid default price feed config containing any additional properties
+      // not set in DefaultPriceFeedConfig
+      let defaultPriceFeedConfig = {
+        currentPrice: "1", // Mocked current price. This will be scaled to the identifier's precision.
```
chrismaree, line 159: Ideally, I don't think scaling should occur like this and rather we should define a set of `convertPrice` methods that input defined parameters into the feeds see here within the liquidator for example: https://github.com/UMAprotocol/protocol/blob/c79f3476fee07736257582f7d92eb7c95932c300/packages/liquidator/test/Liquidator.js#L126

If we are to follow this pattern then we should test this bot against a set of different decimals to ensure it can correctly deal with all configurations and scalling. 

LMKWYT.

chrismaree (commented):
Overall this is pretty awesome and very exciting! your code is great. I have a few nits/comments and questions as well as one point of contention on the use of the `PriceFeedMockScaled` which I would ideally like to not be used at all in the bots.

packages/financial-templates-lib/src/price-feed/PriceFeedMockScaled.js
```diff
@@ -2,74 +2,89 @@ const { toBN } = require("web3").utils;
...
 const { parseFixed } = require("@ethersproject/bignumber");
 
-// An implementation of PriceFeedInterface that medianizes other price feeds.
+// Adds a final precision conversion step to the PriceFeedMock before returning prices.
 class PriceFeedMockScaled extends PriceFeedInterface {
-  // Constructs the MedianizerPriceFeed.
-  // priceFeeds a list of priceFeeds to medianize. All elements must be of type PriceFeedInterface. Must be an array of
-  // at least one element.
-  constructor(currentPrice, historicalPrice, lastUpdateTime, invertPrice, decimals = 18) {
+  constructor(currentPrice, historicalPrice, lastUpdateTime, priceFeedDecimals = 18, lookback = 3600) {
```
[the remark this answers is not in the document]
nicholaspai: In this pricefeedMock, nothing, because its intended to use only for tests. But every `PriceFeed` that implements `PriceFeedInterface` has to implement `getLookback()` so this sets a value for that function. 

Generally, `lookback` sets the valid window into the past that a pricefeed can search for a historical price.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -162,6 +164,34 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    const requiredFields = ["currentPrice", "historicalPrice"];
```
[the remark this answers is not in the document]
nicholaspai: Yes in general I agree with your statement. But these fields are required for the `test` type of pricefeed because they are required inputs to construct a new `PriceFeedMockScaled`

packages/financial-templates-lib/src/price-feed/PriceFeedMockScaled.js
```diff
@@ -2,74 +2,89 @@ const { toBN } = require("web3").utils;
...
-      this.historicalPrices[_price.timestamp] = this.convertDecimals(_price.price);
+      // allows this to be set to null without throwing.
+      this.historicalPrices[_price.timestamp] = _price.price ? this.convertDecimals(_price.price) : _price.price;
     });
   }
 
-  // only available in mock
   setHistoricalPrice(historicalPrice) {
-    this.historicalPrice = this.convertDecimals(historicalPrice);
+    this.historicalPrice = historicalPrice ? this.convertDecimals(historicalPrice) : historicalPrice;
```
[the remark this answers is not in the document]
nicholaspai: > Does this make this.historicalPrice null if historicalPrice is not passed as a parameter?
Correct

And then if you look at this class's `getHistoricalPriceFeed()` implementation, it will throw an error if `historicalPrice` is null:

```
    if (!this.historicalPrice && !(time in this.historicalPrices)) {
      throw new Error("PriceFeedMock expected error thrown");
```

packages/optimistic-oracle/src/keeper.js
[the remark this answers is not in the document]
nicholaspai: Great catch @chrismaree !

and @pemulis yes `0x` is the default ancillary data value.

packages/optimistic-oracle/src/keeper.js
```diff
@@ -0,0 +1,250 @@
...
+          proposer: this.account,
+          proposalBond,
+          priceRequest,
+          error
+        });
+        continue;
+      }
+      const txnConfig = {
+        from: this.account,
+        gas: Math.min(Math.floor(gasEstimation * this.GAS_LIMIT_BUFFER), this.txnGasLimit),
```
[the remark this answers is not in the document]
nicholaspai: Its incase Truffle underestimates the gas required

packages/optimistic-oracle/test/keeper.js
```diff
@@ -0,0 +1,305 @@
...
+        defaultPriceFeedConfig
+      });
+
+      // Update the bot to read the new OO state.
+      await keeper.update();
+    });
+
+    it("_setAllowances", async function() {
+      // Calling it once should set allowances
+      await keeper._setAllowances();
```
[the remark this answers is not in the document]
nicholaspai: We could... but I don't anticipate doing so, hence the `_` prefix.. I just wanted to unit test the method.

packages/optimistic-oracle/test/keeper.js
nicholaspai: Great point! fixed

packages/common/src/PriceIdentifierUtils.js
```diff
@@ -11,7 +11,10 @@ const IDENTIFIER_NON_18_PRECISION = {
   "STABLESPREAD/BTC": 8,
   "ELASTIC_STABLESPREAD/USDC": 6,
   BCHNBTC: 8,
-  AMPLUSD: 6
+  AMPLUSD: 6,
+  // The following identifiers are used in local test environments only:
+  TEST8DECIMALS: 8,
```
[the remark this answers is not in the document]
nicholaspai: The problem is that `USDBTC` gets mapped to a `MedianizerPriceFeed` that uses `CryptoWatchPriceFeeds` by the `DefaultPriceFeed.js`, and I don't want to test with CryptoWatch data.

packages/optimistic-oracle/index.js
nicholaspai: Well its also a disputer, I think its fine to call it proposer though

nicholaspai: Renamed

packages/optimistic-oracle/src/keeper.js
nicholaspai: I renamed!

packages/optimistic-oracle/src/keeper.js
nicholaspai: Yeah good call. Bulk approval is an upfront cost that I think we should pay so we don't have any approval errors come proposal/dispute time

packages/optimistic-oracle/src/keeper.js
nicholaspai: Will do

packages/optimistic-oracle/src/keeper.js
nicholaspai: We usually just include the hash, look at the liquidator bot's index.js file for approval for example

packages/optimistic-oracle/src/keeper.js
nicholaspai: Yes great call

packages/optimistic-oracle/test/keeper.js
nicholaspai: Note that I changed the PriceFeedMockScaled code. It now copies directly from PriceFeedMock but allows you set the starting precision from a price like "10" --> 10e18, 10e6, etc.

I wanted to test that the proposer would use the same precision as the identifier calls for. For example, USDBTC would call for a price in 6 precision.

WDYT?

packages/optimistic-oracle/test/keeper.js
nicholaspai: For a set of price requests: A, B, C, how would you determine how to `convertPrice` when responding to each request?

For example, imagine that `A` calls for 6 precision, `B` for 8, and `C` for 18.

Currently, I read from the `DefaultPriceFeed.js` to determine the mapping of identifier --> precision. All I do is call the constructed pricefeed (again, implied from `DefaultPriceFeed`) `getHistoricalPrice()` method which returns price in the identifier's specified precision.

What about this flow can be improved?

packages/common/src/PriceIdentifierUtils.js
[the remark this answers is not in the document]
chrismaree: SGTM.

packages/optimistic-oracle/src/keeper.js
chrismaree: sgtm.

packages/optimistic-oracle/test/keeper.js
chrismaree: ye, that makes sense. and to double-check this PriceFeedMockScaled was not used anywhere else, right?

packages/optimistic-oracle/test/keeper.js
chrismaree: ye, I see the issue. given that you are generating feeds within your proposer that makes sense to me. In that case I think it's fine.

packages/optimistic-oracle/test/keeper.js
nicholaspai: That's correct

packages/optimistic-oracle/test/proposer.js
```diff
@@ -0,0 +1,312 @@
...
+    });
+    await proposer.update();
+    await proposer.sendProposals();
+
+    // Check for the specific failure log emitted by the proposer.
+    assert.equal(lastSpyLogLevel(spy), "error");
+    assert.isTrue(spyLogIncludes(spy, -1, "Failed to query historical price for price request"));
+  });
+
+  // TODO:
```
chrismaree, line 309:
```suggestion
  // TODO:
  // Can deal with ancillary data
```

should also have a test to validate ancillary data is correctly dealt with.

chrismaree (approved):
This is great! I think in a future PR we should add tests that validate ancillary data is correctly dealt with within requests.

LGTM!

packages/optimistic-oracle/test/proposer.js
nicholaspai: Ah yes. Can you look at this [commit](https://github.com/UMAprotocol/protocol/pull/2505/commits/8042aa6c391756e9119fc10517f249c930ed24b5) and check that I'm correctly simulating how the Perpetual and EMP will submit ancillary data to the OO?

packages/affiliates/payouts/dappmining/2021-02-01_2021-02-08_YD-BTC-MAR21_0004.json
```diff
@@ -5,10 +5,7 @@
     "empAddress": "0x1c3f1A342c8D9591D9759220d114C685FD1cF6b8",
     "totalRewards": 4595,
     "defaultAddress": "0xa0dfF1F09a0DCEda7641b7eB07c3BDbe66a06C35",
-    "whitelist": [
```
nicholaspai, line 8: I guess these changes were never linted before that's why they're included in diff.

packages/affiliates/payouts/dappmining/2021-02-01_2021-02-08_YD-ETH-MAR21_0004.json
```diff
@@ -5,10 +5,7 @@
     "empAddress": "0xE4256C47a3b27a969F25de8BEf44eCA5F2552bD5",
     "totalRewards": 4422,
     "defaultAddress": "0xa0dfF1F09a0DCEda7641b7eB07c3BDbe66a06C35",
-    "whitelist": [
```
nicholaspai, line 8: Same comment as [here](https://github.com/UMAprotocol/protocol/pull/2505/files#r574575364)

packages/affiliates/payouts/devmining/2021-02-01_21-02-02.json
```diff
@@ -3,80 +3,26 @@
     "startTime": 1612220400000,
     "endTime": 1612825200000,
     "empWhitelist": [
-      [
```
nicholaspai, line 6: Same comment as [here](https://github.com/UMAprotocol/protocol/pull/2505/files#r574575364)

packages/trader/tsconfig.json
```diff
@@ -4,7 +4,5 @@
   "compilerOptions": {
     "outDir": "build"
   },
-  "references": [
```
nicholaspai, line 7: Same comment as [here](https://github.com/UMAprotocol/protocol/pull/2505/files#r574575364)

packages/affiliates/payouts/dappmining/2021-02-01_2021-02-08_YD-BTC-MAR21_0004.json
chrismaree: ye, I had the same diff one of my other PRs.

chrismaree (approved):

PR #2506, opened 2021-02-05
fix(jsontransport): refactor and simplify transport

chrismaree:
**Motivation**

In a recent PR added to better deal with feeds that produce errors, a change was made to accommodate an array of errors being thrown. This can be seen [here](https://github.com/UMAprotocol/protocol/blob/c91e11bad264509efd4ef98044e448e6e5b8b5f0/packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js#L49), for example. In this scenario it is possible for a feed to catch multiple errors and produce an array of logs that would look something like this within gcp:
image.

This works fine and plays nicely with all logging transports.

However, this change is incompatible with how the JsonTransport works. The JsonTransport takes in logs from Winston and formats them as an array of json objects, enabling easy consumption upstream within GCP when the bots run in serverless.

This PR simplifies and refines this transport to accommodate this new structure.


packages/financial-templates-lib/src/logger/JsonTransport.js
```diff
@@ -12,17 +12,13 @@ function createJsonTransport() {
...
-              error
-                .replace(/\r?\n|\r/g, "")
-                .replace(/\s\s+/g, " ") // Remove tabbed chars.
-                .replace(/\\"/g, ""), // Remove escaped quotes.
-              Object.getOwnPropertyNames(error) // Turn the json object into a parsable structure.
-            )
-          );
+          // If there is an error then strip out all punctuation to make it easily consumable by GCP within a log json
+          // object. Note this error object is assumed to be a string, converted within Logger.js that uses this transport.
+          error = error
```
nicholaspai, line 17: So this handles JSON's same as before but also handles arrays now but just to stringing it?

nicholaspai (approved):
Is this backwards compatible with non array errors? Just want to make sure you've tested this with both non array and array errors

mrice32:
Merging to see if it will fix the alerts going off now. If not, we can send out a second PR.

PR #2508, opened 2021-02-05
fix(WDF): add warning if wdf threshold is more than token balance

chrismaree:
**Motivation**

The WDF can behave in unexpected ways if the threshold is set to more than the total number of synthetics the bot is holding. This should warn the bot operator.

**Summary**

Added some branching logic to send an `info` alert if the synthetic balance of the bot is less than the whale defense fund amount.

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/liquidator/src/liquidator.js
+++ b/packages/liquidator/src/liquidator.js
@@ -213,6 +213,20 @@ class Liquidator {
       at: "Liquidator",
       message: "Checking for liquidatable positions and preforming liquidations"
     });
+    if (
+      maxTokensToLiquidateWei &&
+      this.whaleDefenseFundWei &&
+      this.toBN(this.whaleDefenseFundWei).gte(this.toBN(maxTokensToLiquidateWei))
+    ) {
+      this.logger.info({
+        at: "Liquidator",
+        message:
+          "The whale defense fund reserve amount is greater than the liquidators synthetic balance. This might result in unintended consequences. Consider changing this param or adding more synthetics.",
+        botTokenBalance: maxTokensToLiquidateWei.toString(),
+        whaleDefenseFundWei: this.whaleDefenseFundWei.toString()
+      });
+    }
+
     // If an override is provided, use that price. Else, get the latest price from the price feed.
     const price = liquidatorOverridePrice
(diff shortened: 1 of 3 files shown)


packages/liquidator/src/liquidator.js
```diff
@@ -213,6 +213,20 @@ class Liquidator {
...
     });
+    if (
+      maxTokensToLiquidateWei &&
+      this.whaleDefenseFundWei &&
+      this.toBN(this.whaleDefenseFundWei).gte(this.toBN(maxTokensToLiquidateWei))
+    ) {
+      this.logger.info({
+        at: "Liquidator",
+        message:
+          "The whale defense fund reserve amount is greater than the liquidators synthetic balance. This might result in unintended consequences. Consider changing this param or adding more synthetics.",
```
nicholaspai, line 224:
```suggestion
          "The whale defense fund reserve amount is greater than the liquidators synthetic balance. This might result in skipped liquidations. Consider reducing this threshold or adding more synthetics.",
```

packages/liquidator/src/liquidator.js
```diff
@@ -213,6 +213,20 @@ class Liquidator {
       at: "Liquidator",
       message: "Checking for liquidatable positions and preforming liquidations"
     });
+    if (
```
nicholaspai, line 216: As @mrice32 mentioned I think this would be cleaner in the BalanceMonitor. Wdyt ?

nicholaspai (approved):
Thanks for making this change so quickly! I think it might be better placed in the BalanceMonitor however. Otherwise the implementation looks great

packages/liquidator/src/liquidator.js
chrismaree: hmm. I'm not too sure about that. This is a _liquidator_ specific check and is independent of the monitor bot. If, say, I'm only running a liquidator how would placing this in the `BalanceMonitor` detect this invalid config if I'm not running a monitor?

Another option could be to put it within the liquidator's `index.js` file but it was cleaner to put it here after trying both options.

WDYT?

daywiss (approved):

PR #2512, opened 2021-02-07
feat(financial-templates-lib): add BlockFinder

mrice32:
**Motivation**

To build a price feeds that require past block queries at specific _times_, we need a way to find the exact block at or immediately preceding a particular timestamp. Our existing block querying utilities _can_ do this, but they query _all_ blocks over a certain range. This utility serves a slightly different purpose in that it uses a binary search (technically interpolation search) to find the block for a timestamp at any point in the blockchain's history without needing _all_ blocks since that point. Because it's a binary-style search, this can take slightly longer since all queries need to be done in series.

**Summary**

The BlockFinder follows the same style as the other block querying utilities. Its interface is a single function to get the block for a timestamp.

Note: the implementation may be convoluted in certain spots, so please point out where more comments or details are needed.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -192,8 +192,11 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
     return block;
   }
 
+  // Return the latest block, between startBlock and endBlock, whose timestamp is <= timestamp.
   // Effectively, this is an interpolation search algorithm to minimize block requests.
   async function findBlock(startBlock, endBlock, timestamp) {
+    assert(endBlock.number !== startBlock.number, "startBlock cannot equal endBlock");
+
     // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
     // timestamp.
     if (endBlock.timestamp === timestamp) return endBlock;
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -197,8 +197,6 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
   // Effectively, this is an interpolation search algorithm to minimize block requests.
   // Note: startBlock and endBlock _must_ be different blocks.
   async function findBlock(startBlock, endBlock, timestamp) {
-    assert(endBlock.number !== startBlock.number, "startBlock cannot equal endBlock");
-
     // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
     // timestamp.
     if (endBlock.timestamp === timestamp) return endBlock;
@@ -207,6 +205,7 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
     // timestamp to be <= the requested timestamp.
     if (endBlock.number === startBlock.number + 1) return startBlock;
 
+    assert(endBlock.number !== startBlock.number, "startBlock cannot equal endBlock");
     assert(
       timestamp < endBlock.timestamp && timestamp > startBlock.timestamp,
       "timestamp not in between start and end blocks"
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -194,7 +194,12 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
 
   // Effectively, this is an interpolation search algorithm to minimize block requests.
   async function findBlock(startBlock, endBlock, timestamp) {
+    // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
+    // timestamp.
     if (endBlock.timestamp === timestamp) return endBlock;
+
+    // If there's no equality, but the blocks are adjacent, return the startBlock, since we want the returned block's
+    // timestamp to be <= the requested timestamp.
     if (endBlock.number === startBlock.number + 1) return startBlock;
 
     assert(
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -168,7 +168,12 @@ exports.PriceHistory = (getPrice, prices = {}) => {
   };
 };
 
-// A simple class-like structure to find blocks by timestamp, but keep a cache to optimize the search.
+/**
+ * @notice Constructs the BlockFinder, which is a simple class-like object to find blocks by timestamp, but keeps a
+ *      cache to optimize the search.
+ * @param {Function} requestBlock async function, like web3.eth.getBlock that returns a block for a particular block
+ *      number or returns the latest block if no argument is provided. Blocks returned must have a `number` and
+ */
 exports.BlockFinder = (requestBlock, blocks = []) => {
   // Grabs the most recent block and caches it.
   async function getLatestBlock() {
@@ -214,8 +219,10 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
     }
   }
 
-  // Gets the latest block with a timestamp <= the provided timestamp.
-  // Throws if the timestamp is after the latest block or before the 0th block.
+  /**
+   * @notice Gets the latest block whose timestamp is <= the provided timestamp.
+   * @param {number} timestamp timestamp to search.
+   */
   async function getBlockForTimestamp(timestamp) {
     // If the last block we have stored is too early, grab the latest block.
     if (blocks.length === 0 || blocks[blocks.length - 1].timestamp < timestamp) {
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -175,6 +175,7 @@ exports.PriceHistory = (getPrice, prices = {}) => {
  *      number or returns the latest block if no argument is provided. Blocks returned must have a `number` and
  */
 exports.BlockFinder = (requestBlock, blocks = []) => {
+  assert(requestBlock, "requestBlock function must be provided");
   // Grabs the most recent block and caches it.
   async function getLatestBlock() {
     const block = await requestBlock();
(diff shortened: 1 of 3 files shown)


packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+    return block;
+  }
+
+  // Effectively, this is an interpolation search algorithm to minimize block requests.
+  async function findBlock(startBlock, endBlock, timestamp) {
+    // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
+    // timestamp.
+    if (endBlock.timestamp === timestamp) return endBlock;
+
+    // If there's no equality, but the blocks are adjacent, return the startBlock, since we want the returned block's
```
nicholaspai, line 201: I think you should add the comment "Return the block, between `startBlock` and `endBlock`, whose timestamp is <= `timestamp`" above this function because it helped me to understand what the end goal was here.

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+    const block = await requestBlock(number);
+    blocks.splice(index, 0, block); // A simple insert at index.
+    return block;
+  }
+
+  // Effectively, this is an interpolation search algorithm to minimize block requests.
+  async function findBlock(startBlock, endBlock, timestamp) {
+    // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
+    // timestamp.
+    if (endBlock.timestamp === timestamp) return endBlock;
```
nicholaspai, line 199: What if `startBlock` and `endBlock` are the same block?

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+
+    assert(
+      timestamp < endBlock.timestamp && timestamp > startBlock.timestamp,
+      "timestamp not in between start and end blocks"
+    );
+
+    // Interpolating the timestamp we're searching for to block numbers.
+    const totalTimeDifference = endBlock.timestamp - startBlock.timestamp;
+    const totalBlockDistance = endBlock.number - startBlock.number;
+    const blockPercentile = (timestamp - startBlock.timestamp) / totalTimeDifference;
```
nicholaspai, line 213: As mentioned above, if its possible that endBlock == startBlock than this denominator will be 0 and throw unexpectedly

nicholaspai (approved):
Looks really great, this seems like an easy substitution into the BalancerPriceFeed but not so easily for the UniswapPriceFeed which already implements its own block searching algo. We'll want to make sure that the UniswapPriceFeed TWAP still works after using this new `util` for edge cases like when there is one price before period and one inside (i.e. the edge case you just fixed recently)

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: This would fail earlier, since the timestamps inequality above would fail (if they are the same block, the timestamp cannot be both above and below the desired timestamp). Need to think a little more about whether they can ever be the same block at the point of entry. My suspicion is that they cannot -- otherwise, I would expect the tests to sporadically fail. I will add a comment about _why_ this cannot happen and add a comment that a precondition of calling this function is that they are not the same block.

This is called from two places:
- getBlockForTimestamp: in this case the blocks are two different indices from the blocks array, so they can't be the same block.
- recursive call: only one of the blocks is new, and the new block that's requested is "clamped" between the previous start and end block. So I don't think they can be the same block.

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: See https://github.com/UMAprotocol/protocol/pull/2512/files#r571652327, but I don't think the can be at either of the callsites. However, this should be an explicit precondition.

mrice32:
> Looks really great, this seems like an easy substitution into the BalancerPriceFeed but not so easily for the UniswapPriceFeed which already implements its own block searching algo. We'll want to make sure that the UniswapPriceFeed TWAP still works after using this new `util` for edge cases like when there is one price before period and one inside (i.e. the edge case you just fixed recently)

Yeah, not sure this will exactly work as written for uniswap since it needs to find a block before a timestamp that also has a sync event. My guess is that we'll just leave that as is for now, but then sub this in for the Balancer and future price feeds that require it, like the badger and/or sushi ones.

mrice32:
@nicholaspai thanks for the thoughtful review. Will add more comments and tag you on them so you can verify that they clarify the code.

packages/financial-templates-lib/src/price-feed/utils.js
nicholaspai: Agreed maybe you just `assert(endBlock !== startBlock)`?

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+  async function getBlock(number) {
+    const index = lodash.sortedIndexBy(blocks, { number }, "number");
+    if (blocks[index]?.number === number) return blocks[index]; // Return early if block already exists.
+    const block = await requestBlock(number);
+    blocks.splice(index, 0, block); // A simple insert at index.
+    return block;
+  }
+
+  // Effectively, this is an interpolation search algorithm to minimize block requests.
+  async function findBlock(startBlock, endBlock, timestamp) {
```
chrismaree, line 196: do you want to apply any input sanitization on these variables or types?

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+      "timestamp not in between start and end blocks"
+    );
+
+    // Interpolating the timestamp we're searching for to block numbers.
+    const totalTimeDifference = endBlock.timestamp - startBlock.timestamp;
+    const totalBlockDistance = endBlock.number - startBlock.number;
+    const blockPercentile = (timestamp - startBlock.timestamp) / totalTimeDifference;
+    const estimatedBlock = startBlock.number + Math.round(blockPercentile * totalBlockDistance);
+
+    // Clamp is just ensures the estimated block is strictly greater than the start block and strictly less than the end block.
```
chrismaree, line 216:
```suggestion
    // Clamp ensures the estimated block is strictly greater than the start block and strictly less than the end block.
```

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -168,6 +168,97 @@ exports.PriceHistory = (getPrice, prices = {}) => {
...
+    const block = await requestBlock();
+    const index = lodash.sortedIndexBy(blocks, block, "number");
+    if (blocks[index]?.number !== block.number) blocks.splice(index, 0, block);
+    return blocks[index];
+  }
+
+  // Grabs the block for a particular number and caches it.
+  async function getBlock(number) {
+    const index = lodash.sortedIndexBy(blocks, { number }, "number");
+    if (blocks[index]?.number === number) return blocks[index]; // Return early if block already exists.
```
chrismaree, line 189: seeing you using the `?.` syntax, if this is not the type you expect you might have some issues. do you want to sanitize the inputs before, as with the other methods?

chrismaree (approved):
Looks great to me! nice clean code. I had a few comments/nits about input validation but otherwise LGTM

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: This is to deal with the case where index doesn't exist, which means blocks[index] === undefined. This happens (I believe) if the block number should be slotted at the end of the blocks array.

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: This is a great question. Only one function is exported, so the others aren't callable from outside. If these methods are only called internally, should we do input sanitization? Or should we only do it for the externally-callable functions?

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: Sounds good!

mrice32: Done.

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: Done.

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: Added a comment making this an explicit precondition.

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: Two other comments:
- blocks isn't an input, so I don't think sanitizing them would give us guarantees about the types in that array.
- Regarding sanitizing inputs, generally, we _could_ ensure that number is defined, but since it is only called internally (function isn't exported), I'm somewhat conflicted on whether we should check types when we can trivially see all the callsites nearby, but I don't have strong opinions on that!

packages/financial-templates-lib/src/price-feed/utils.js
mrice32: @chrismaree I added some input validation both to the outer function (BlockFinder) and the getBlockForTimestamp.

nicholaspai (approved):
LGTM assuming youve resolved Chris' comments

packages/financial-templates-lib/src/price-feed/utils.js
chrismaree: sounds good to me!

chrismaree (approved):

PR #2513, opened 2021-02-08
feat(price-feeds): add ExpressionPriceFeed

mrice32:
**Motivation**

To support more arbitrary types of price feeds, it is becoming clear that it would be useful to be able to define arbitrary expressions that combine existing price feeds. This is a proposal for a price feed that takes expressions of the form:

```
"BTCUSD * 100 / ETHUSD"
```

Note: the library can actually handle _far more_ complex expressions, even multiline ones. But, for now, we should encourage users to keep them fairly simple.

**Summary**

This is a _proposal_ for a type of expression-parsing price feed to add more generality to the price feed creation process. This would move much of the simple math to the config and leave only the data-acquisition to the price feeds themselves.

I haven't added the configuration and construction portions of this price feed, as I want to ensure the team agrees with this rough direction. However, it's important to note that I think this price feed will be constructed by searching for known strings in the expression and including price feeds associated with each of those known strings. The user would basically be able to use existing default price feed identifiers _and_ they would be able to include a custom mapping in their config for additional price feed names that they'd like to map.

Ultimately, this expression parser could make the medianizer price feed unnecessary, as the library used would allow us to inject a custom medianizer function.

This will also facilitate the badger price feed.

Other notes:
- The `mathjs` library used for expression parsing is quite powerful and large. It has some very convenient scoping mechanisms that allow us to replace "symbols" with values. This sort of functionality could potentially be replaced with a call to `eval` but that's generally considered bad practice due to security and performance concerns. However, it would shrink the size of the dependencies considerably.
- To make _all_ of our default price identifiers work with this sort of strategy, we'll need to strip them of `/` characters (which could be confused with a division operator by the parser) when ingesting them here. That's not a huge concern, but it is a slight added complexity.
- The decimal strategy in this class is quite different from others. In this case, I use a high-precision BigNumber class that comes with the mathjs library to convert all values to decimal before being processed by the library. This allows the library to use its default arithmetic operators, which is far far simpler than trying to make it operate on fixed point values. To ensure this BigNumber can handle our high-precision fixed point numbers, I customized the config to give them 100 decimals of precision, which should be plenty for our use. After the operations, it rounds off the result to drop any precision that's larger than our fixed point numbers allow.

This is how you configure an expression price feed:
```js
{
  type: "expression",
  expression: "ETHBTC + MYCUSTOMFEED",
  customFeeds: {
    MYCUSTOMFEED: {
      type: "cryptowatch",
      exchange: "coinbase-pro",
      pair: "ethusd"
    }
  }
}
```

In this example, the parser would find that the two "symbols" in the expression are `ETHBTC` (notice that we have stripped the `/` character) and `MYCUSTOMFEED`. Preferring customFeeds in the resolution, we would look through customFeeds _and_ the default feeds in the repo searching for each of these symbols. If any resolution fails, the construction fails. In this case, however, it would construct both feeds and use them to evaluate the expression. The nice thing about the early resolution is that any symbols that aren't in the expression are pruned, so there should never be any unnecessary feeds being constructed and/or evaluated.

Note: for feed names that contain special characters that are allowed in identifiers: `/`, ` `, `[`, `]`, they will need to be escaped, like `"\\/"` in the configuration.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -11,6 +11,7 @@
     "@uniswap/sdk": "^2.0.5",
     "bluebird": "^3.7.2",
     "dotenv": "^6.2.0",
+    "mathjs": "^9.2.0",
     "minimist": "^1.2.0",
     "node-fetch": "^2.6.0",
     "node-pagerduty": "^1.2.0",
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -192,8 +192,8 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
   // Returns an ExpressionPriceFeed.
   async function _createExpressionPriceFeed(expressionConfig) {
     // Build list of configs that could be used in the expression including default price feed configs and customFeeds
-    // that the user has provided inside the ExpressionPriceFeed config. Tranform keys by stripping "/" since that
-    // would be interpreted as division.
+    // that the user has provided inside the ExpressionPriceFeed config. Note: default configs are overriden by
+    // customFeeds with the same name. Tranform keys by stripping "/" since that would be interpreted as division.
     const allConfigs = Object.fromEntries(
       Object.entries({ ...defaultConfigs, ...expressionConfig.customFeeds }).map(([key, value]) => {
         return [key.replace("/", ""), value];
@@ -201,8 +201,8 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     );
 
     // This call chain:
-    // 1. Parses the expression tree into "nodes"
-    // 2. Filters for "symbol" nodes, which would be price feed identifiers in our case.
+    // 1. Parses the expression into an expression tree of nodes.
+    // 2. Filters for "symbol" nodes, which would be price feed identifiers in this case.
     // 3. Extract the name property for each of these symbol nodes
     // 4. Puts it all in a set to dedupe repeated symbols.
     const symbols = new Set(
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -1,5 +1,6 @@
 const assert = require("assert");
 const { ChainId, Token, Pair, TokenAmount } = require("@uniswap/sdk");
+const { parse } = require("mathjs");
 const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
 const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
@@ -8,6 +9,7 @@ const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
 const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
 const { getTruffleContract } = require("@uma/core");
+const { ExpressionPriceFeed } = require("./ExpressionPriceFeed");
 
 async function createPriceFeed(logger, web3, networker, getTime, config) {
   const Uniswap = getTruffleContract("Uniswap", web3, "latest");
@@ -162,6 +164,19 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.denominatorPriceFeed && (await _createMedianizerPriceFeed(config.denominatorPriceFeed));
 
     return new BasketSpreadPriceFeed(web3, logger, baselinePriceFeeds, experimentalPriceFeeds, denominatorPriceFeed);
+  } else if (config.type === "expression") {
+    const requiredFields = ["expression"];
+    if (isMissingField(config, requiredFields, logger)) {
+      return null;
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating ExpressionPriceFeed",
+      config
+    });
+
+    return await _createExpressionPriceFeed(config);
   }
 
   logger.error({
@@ -173,6 +188,63 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
   return null;
 
   // Internal helper methods:
+
+  // Returns an ExpressionPriceFeed.
+  async function _createExpressionPriceFeed(expressionConfig) {
+    // Build list of configs that could be used in the expression including default price feed configs and customFeeds
+    // that the user has provided inside the ExpressionPriceFeed config. Tranform keys by stripping "/" since that
+    // would be interpreted as division.
+    const allConfigs = Object.fromEntries(
+      Object.entries({ ...defaultConfigs, ...expressionConfig.customFeeds }).map(([key, value]) => {
+        return [key.replace("/", ""), value];
+      })
+    );
+
+    // This call chain:
+    // 1. Parses the expression tree into "nodes"
+    // 2. Filters for "symbol" nodes, which would be price feed identifiers in our case.
+    // 3. Extract the name property for each of these symbol nodes
+    // 4. Puts it all in a set to dedupe repeated symbols.
+    const symbols = new Set(
+      parse(expressionConfig.expression)
+        .filter(node => node.isSymbolNode)
+        .map(node => node.name)
+    );
+
+    // This is a complicated looking map that maps each symbol into an entry in an object with its value the price
+    // feed created from the mapped config in allConfigs.
+    const priceFeedMap = Object.fromEntries(
+      await Promise.all(
+        symbols.values().map(async symbol => {
+          const config = allConfigs[symbol];
+
+          // If there is no config for this symbol, insert null and send an error.
+          if (!config) {
+            logger.error({
+              at: "_createExpressionPriceFeed",
+              message: `No price feed config found for symbol: ${symbol} 🚨`,
+              expressionConfig
+            });
+            return [symbol, null];
+          }
+
+          // These configs will inherit the expression config values (except type), but prefer the individual config's
+          // value when present.
+          const combinedConfig = { ...expressionConfig, type: undefined, ...config };
+
+          // If this returns null, just return upstream since the error has already been logged and the null will be
+          // detected upstream.
+          const priceFeed = await createPriceFeed(logger, web3, networker, getTime, combinedConfig);
+          return [symbol, priceFeed];
+        })
+      )
+    );
+
+    // Return null if any of the price feeds in the map are null (meaning there was an error).
+    if (Object.values(priceFeedMap).some(priceFeed => priceFeed === null)) return null;
+
+    return new ExpressionPriceFeed(priceFeedMap, expressionConfig.expression);
+  }
 
   // Returns a MedianizerPriceFeed
   async function _createMedianizerPriceFeed(medianizerConfig) {
@@ -196,6 +268,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     }
     return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
   }
+
   // Returns an array or "basket" of MedianizerPriceFeeds
   async function _createBasketOfMedianizerPriceFeeds(medianizerConfigs) {
     return await Promise.all(medianizerConfigs.map(config => _createMedianizerPriceFeed(config)));
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -245,7 +245,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     // Return null if any of the price feeds in the map are null (meaning there was an error).
     if (Object.values(priceFeedMap).some(priceFeed => priceFeed === null)) return null;
 
-    return new ExpressionPriceFeed(priceFeedMap, expressionConfig.expression);
+    return new ExpressionPriceFeed(priceFeedMap, expressionConfig.expression, expressionConfig.priceFeedDecimals);
   }
 
   // Returns a MedianizerPriceFeed
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -1,6 +1,5 @@
 const assert = require("assert");
 const { ChainId, Token, Pair, TokenAmount } = require("@uniswap/sdk");
-const { parse } = require("mathjs");
 const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
 const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
@@ -9,7 +8,7 @@ const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
 const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
 const { getTruffleContract } = require("@uma/core");
-const { ExpressionPriceFeed } = require("./ExpressionPriceFeed");
+const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
 
 async function createPriceFeed(logger, web3, networker, getTime, config) {
   const Uniswap = getTruffleContract("Uniswap", web3, "latest");
@@ -193,10 +192,10 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
   async function _createExpressionPriceFeed(expressionConfig) {
     // Build list of configs that could be used in the expression including default price feed configs and customFeeds
     // that the user has provided inside the ExpressionPriceFeed config. Note: default configs are overriden by
-    // customFeeds with the same name. Tranform keys by stripping "/" since that would be interpreted as division.
+    // customFeeds with the same name. Tranform keys by escaping any special characters in the identifier names..
     const allConfigs = Object.fromEntries(
       Object.entries({ ...defaultConfigs, ...expressionConfig.customFeeds }).map(([key, value]) => {
-        return [key.replace("/", ""), value];
+        return [escapeSpecialCharacters(key), value];
       })
     );
 
@@ -207,7 +206,8 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     // 4. Puts it all in a set and converts back to an array to dedupe any repeated values.
     const symbols = Array.from(
       new Set(
-        parse(expressionConfig.expression)
+        math
+          .parse(expressionConfig.expression)
           .filter(node => node.isSymbolNode)
           .map(node => node.name)
       )
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -204,18 +204,20 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     // 1. Parses the expression into an expression tree of nodes.
     // 2. Filters for "symbol" nodes, which would be price feed identifiers in this case.
     // 3. Extract the name property for each of these symbol nodes
-    // 4. Puts it all in a set to dedupe repeated symbols.
-    const symbols = new Set(
-      parse(expressionConfig.expression)
-        .filter(node => node.isSymbolNode)
-        .map(node => node.name)
+    // 4. Puts it all in a set and converts back to an array to dedupe any repeated values.
+    const symbols = Array.from(
+      new Set(
+        parse(expressionConfig.expression)
+          .filter(node => node.isSymbolNode)
+          .map(node => node.name)
+      )
     );
 
     // This is a complicated looking map that maps each symbol into an entry in an object with its value the price
     // feed created from the mapped config in allConfigs.
     const priceFeedMap = Object.fromEntries(
       await Promise.all(
-        symbols.values().map(async symbol => {
+        symbols.map(async symbol => {
           const config = allConfigs[symbol];
 
           // If there is no config for this symbol, insert null and send an error.
--- a/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
@@ -1,0 +1,117 @@
+const assert = require("assert");
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const Web3 = require("web3");
+const { create, all } = require("mathjs");
+const math = create(all, {
+  number: "BigNumber",
+  precision: 100
+});
+
+// Gets balancer spot and historical prices. This price feed assumes that it is returning
+// prices as 18 decimals of precision, so it will scale up the pool's price as reported by Balancer contracts
+// if the user specifies that the Balancer contract is returning non-18 decimal precision prices.
+class ExpressionPriceFeed extends PriceFeedInterface {
+  constructor(priceFeedMap, expression) {
+    super();
+    this.expressionCode = math.parse(expression).compile();
+    this.priceFeedMap = priceFeedMap;
+  }
+
+  async getHistoricalPrice(time) {
+    const historicalPrices = {};
+    const errors = [];
+    await Promise.all(
+      Object.entries(this.priceFeedMap).map(async ([name, pf]) => {
+        const price = await pf.getHistoricalPrice(time).catch(err => errors.push(err));
+        historicalPrices[name] = this._convertToDecimal(price);
+      })
+    );
+
+    if (errors.length > 0) {
+      throw errors;
+    }
+
+    return this._convertToFixed(this.expressionCode.evaluate(historicalPrices));
+  }
+
+  getLastUpdateTime() {
+    const lastUpdateTimes = Object.values(this.priceFeedMap).map(pf => pf.getLastUpdateTime());
+
+    // If any constituents returned an invalid value, bubble it up.
+    if (lastUpdateTimes.some(time => time === null || time === undefined)) {
+      return null;
+    }
+
+    // Take the max.
+    return Math.max(...lastUpdateTimes);
+  }
+
+  getLookback() {
+    const lookbacks = Object.values(this.priceFeedMap).map(priceFeed => priceFeed.getLookback());
+
+    // If any constituents returned an invalid value, bubble it up.
+    if (lookbacks.some(lookback => lookback === undefined || lookback === null)) {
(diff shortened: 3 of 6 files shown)


packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,117 @@
...
+  precision: 100
+});
+
+// Gets balancer spot and historical prices. This price feed assumes that it is returning
+// prices as 18 decimals of precision, so it will scale up the pool's price as reported by Balancer contracts
+// if the user specifies that the Balancer contract is returning non-18 decimal precision prices.
+class ExpressionPriceFeed extends PriceFeedInterface {
+  constructor(priceFeedMap, expression) {
+    super();
+    this.expressionCode = math.parse(expression).compile();
```
nicholaspai, line 16: Will this throw if `expression` is invalid? I think thats a good outcome but just want to make sure

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,117 @@
...
+        const price = await pf.getHistoricalPrice(time).catch(err => errors.push(err));
+        historicalPrices[name] = this._convertToDecimal(price);
+      })
+    );
+
+    if (errors.length > 0) {
+      throw errors;
+    }
+
+    return this._convertToFixed(this.expressionCode.evaluate(historicalPrices));
```
nicholaspai, line 34: Does this imply that the `priceFeedMap` has to list pricefeeds in the exact order that they are used in the `expressionCode`?

nicholaspai (approved):
I really like the overall direction of this PR and 100% agree it will be useful. I'd like to see the following implemented before we merge (which I'm sure you're working on)
- Better comments on new pricefeed class and functions
- Comment above constructor explaining what types of expressions are valid
- Comment explaining how the pricefeed map should be constructed 

One cool thing we can do with this is replace the `inverse` boolean flag from the `Medianizer` and just use this pricefeed instead and define the expression as `1 / MedianizedPriceFeed`

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: It depends on _how_ it is invalid. It will throw if the expression is unparsable, but it will _not_ throw if it's parsable by the engine, but doesn't resolve to a number or if one of the symbols isn't resolved by the scope.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: This is a great question. I will add comments as you suggest and we should probably have more checks around detecting if the output is valid. Here's how this roughly works:
```js
let scope = {
  a: 5,
  b: 6
};
let output = math.parse("a + b").compile().evaluate(scope);
```

In this case output is 11. You can use a and b however you like in the expression and they can be any order in the scope. The scope just defines how it resolves each symbol.

However, I'm not sure what happens in this case:
```js
let scope = {
  a: 5
};
let output = math.parse("a + b").compile().evaluate(scope);
```

My guess is that it would return something like "b + 5", but I need to do more testing to make sure we error gracefully in that case and maybe catch it early.

mrice32:
> * Comment explaining how the pricefeed map should be constructed

I think this PR should add to CreatePriceFeed.js to allow users to create/configure this price feed through the config. Will do this and re-request your review.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,117 @@
...
+  }
+
+  async update() {
+    // Update all constituent price feeds.
+    await Promise.all(Object.values(this.priceFeedMap).map(pf => pf.update()));
+  }
+
+  // Takes a BN fixed point number and converts it to a math.bignumber decimal number that the math library can handle.
+  _convertToDecimal(price) {
+    const decimals = math.bignumber(this.getPriceFeedDecimals());
```
chrismaree, line 97: so this limits this feed to only outputting at the decimals of the underlying price feed. It might be worth while to set this as a parameter. for example, there could be a feed that is comprised of a number of underlying feeds that use say 6 decimals but the resultant feed is set to 18 decimals. Or, if running on an old emp, it is possible the decimals don't match underlying feeds.

In ether case, do you think it worthwhile to add a constructor param that scales the output according to some `priceFeedDecimals` in the same way we treat other feed types?

chrismaree (approved):
I had a few nits/comments and questions but overall this is epic and super powerful! Great work.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,117 @@
+const assert = require("assert");
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const Web3 = require("web3");
+const { create, all } = require("mathjs");
+const math = create(all, {
+  number: "BigNumber",
+  precision: 100
+});
+
+// Gets balancer spot and historical prices. This price feed assumes that it is returning
```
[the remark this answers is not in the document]
mrice32: Ha, yeah, this is just copy paste :) will fix

packages/financial-templates-lib/package.json
```diff
@@ -11,6 +11,7 @@
     "@uniswap/sdk": "^2.0.5",
     "bluebird": "^3.7.2",
     "dotenv": "^6.2.0",
+    "mathjs": "^9.2.0",
```
pemulis, line 14: As a general comment about smart contract interface development patterns, and not this pull request specifically, I wonder if it would be better for interfaces that are mostly reading and writing data from smart contracts to do most of their calculation (and maybe rendering) on the server side.

This came up in the engineering team discussion, where we determined that asking people who use this package to run a server would be a heavy lift, versus doing calculations in the browser (which `mathjs` allows for) or hitting an API we provide, which would be a centralized point of failure in addition to taking a lot of time to build and maintain.

In the long run, I think that smart contract front-ends that pull most of the work out of the client and put it on the server will win out, since very little of what you want to do with smart contracts resembles the social media clients that React _et al._ were really designed for. Every user interaction takes at least a few seconds to finalize, and data updates also take a few seconds. Doing most of the stuff on the server side with a light client-side interface feels like the right trade-off.

I suppose this is also some background for the general community on why we're going in this direction, since `mathjs`, while quite heavy, will work in the browser, while some other alternatives we considered would not.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,114 @@
+const assert = require("assert");
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const Web3 = require("web3");
+const { create, all } = require("mathjs");
+const math = create(all, { number: "BigNumber", precision: 100 });
+
+// Gets balancer spot and historical prices. This price feed assumes that it is returning
+// prices as 18 decimals of precision, so it will scale up the pool's price as reported by Balancer contracts
+// if the user specifies that the Balancer contract is returning non-18 decimal precision prices.
+class ExpressionPriceFeed extends PriceFeedInterface {
```
pemulis, line 10: Another meta comment: If you wanted a price feed that is the result of two existing, approved price feeds run through this calculator, does that price feed need to be approved through the UMIP process, or would people be free to do price feed mash-ups in new synthetics without an extra vote?

mrice32: @pemulis this is a great question. To put it simply, the latter. However, I want to make a larger point. None of our price feeds know anything about approved identifiers. They only know about the default configs in the price feed directory. Those default configs allow us to automatically construct price feeds based on the identifier the EMP is using without needing a custom config input. When an UMIP is approved to add an identifier, we generally add a default config for it.

In addition to expressions automatically being able to use default configs with no additional information, you will also be able to specify a `customFeeds` map inside the ExpressionPriceFeed config that maps your own custom symbols so you're not limited to the default configs. Here's an example of a working config:
```js
{
  type: "expression",
  expression: "ETHBTC + MYCUSTOMFEED",
  customFeeds: {
    MYCUSTOMFEED: {
      type: "cryptowatch",
      exchange: "coinbase-pro",
      pair: "ethusd"
    }
  }
}
```

mrice32:
@chrismaree @nicholaspai I re-requested y'all's reviews. I added the configuration/construction code, which is pretty important for this sort of price feed that's primarily meant for making configuration and construction easier. I added this to the description, but copying below so you see it.

This is how you configure an expression price feed:
```js
{
  type: "expression",
  expression: "ETHBTC + MYCUSTOMFEED",
  customFeeds: {
    MYCUSTOMFEED: {
      type: "cryptowatch",
      exchange: "coinbase-pro",
      pair: "ethusd"
    }
  }
}
```

In this example, the parser would find that the two "symbols" in the expression are `ETHBTC` (notice that we have stripped the `/` character) and `MYCUSTOMFEED`. Preferring customFeeds in the resolution, we would look through customFeeds _and_ the default feeds in the repo searching for each of these symbols. If any resolution fails, the construction fails. In this case, however, it would construct both feeds and use them to evaluate the expression. The nice thing about the early resolution is that any symbols that aren't in the expression are pruned, so there should never be any unnecessary feeds being constructed and/or evaluated.

chrismaree:
> @chrismaree @nicholaspai I re-requested y'all's reviews. I added the configuration/construction code, which is pretty important for this sort of price feed that's primarily meant for making configuration and construction easier. I added this to the description, but copying below so you see it.
> 
> This is how you configure an expression price feed:
> 
> ```js
> {
>   type: "expression",
>   expression: "ETHBTC + MYCUSTOMFEED",
>   customFeeds: {
>     MYCUSTOMFEED: {
>       type: "cryptowatch",
>       exchange: "coinbase-pro",
>       pair: "ethusd"
>     }
>   }
> }
> ```
> 
> In this example, the parser would find that the two "symbols" in the expression are `ETHBTC` (notice that we have stripped the `/` character, which is a requirement because the parser will interpret it as division) and `MYCUSTOMFEED`. Preferring customFeeds in the resolution, we would look through customFeeds _and_ the default feeds in the repo searching for each of these symbols. If any resolution fails, the construction fails. In this case, however, it would construct both feeds and use them to evaluate the expression. The nice thing about the early resolution is that any symbols that aren't in the expression are pruned, so there should never be any unnecessary feeds being constructed and/or evaluated.

This is so cool. only real downside I can see is stripping out the `/` char as now your feed won't match the actual price identifiers.

The ability to define a custom feed like this is epic. This means you can define as many constituent feeds that you need then define the `expression` to join them together.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -174,6 +189,65 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
   // Internal helper methods:
 
+  // Returns an ExpressionPriceFeed.
+  async function _createExpressionPriceFeed(expressionConfig) {
+    // Build list of configs that could be used in the expression including default price feed configs and customFeeds
+    // that the user has provided inside the ExpressionPriceFeed config. Note: default configs are overriden by
+    // customFeeds with the same name. Tranform keys by stripping "/" since that would be interpreted as division.
+    const allConfigs = Object.fromEntries(
+      Object.entries({ ...defaultConfigs, ...expressionConfig.customFeeds }).map(([key, value]) => {
+        return [key.replace("/", ""), value];
```
chrismaree, line 199: hold on. just to double-check I understand. I can't put any feed names that have `/` within the feed expression, right? this is just parsing the list of default feeds to match this rule that expression feeds cant contain any `/`?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -174,6 +189,65 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+      })
+    );
+
+    // This call chain:
```
chrismaree, line 203: is there any way to detect if the user has incorrectly placed a `/` within the feed within an expression? I guess if it parses this as deciding two parts of a feed it would error out? ETH/BTC would be ETH divided by BTC and we don't have a feed that is just `ETH` or `BTC` so it would error, right?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: The way it previously worked (and there was a subtle bug :)) was that you would enter an expression, like:
```
"ETHBTC + STABLESPREADUSDC"
```

That would match the two existing identifiers:
```
ETH/BTC
STABLESPREAD/USDC
```
because we strip out the `/` before searching through the default list.

The weird caveat is that it would also match:
```
ETHBTC
ET/HBTC
E/THBTC
...
```

This has now changed to using escape characters. PTAL :)

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Exactly. 💯 . That's effectively our only way to verify the correctness -- rely on mathjs to tell us if there are any unresolved symbols.

packages/financial-templates-lib/package.json
mrice32: Yeah, I could imagine employing a much heavier server-side implementation. If we really wanted our UIs to be resilient, they could have a fallback setup where they _only_ pull down the price feed js code if the backend/server-side is down for some reason. This, while complicated, could be the best of both worlds, although I doubt we'll ever have the engineering resources to build something like that :)

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: It's actually nicer @nicholaspai. It throws if any of the symbols are not resolved at runtime.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: It's a fair point. Will add.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: Just to follow-up here. It's difficult to guarantee valid-ness of the expression without passing in values for all of the symbols, but we also can't assume which symbol values would be valid in the expression. For instance, passing in 1 for all values would cause the following to throw:
```
(A + B) / (A - B)
```

mrice32: Technically, we could check that all symbols are defined in the map, but I don't think we could check for all types of expression errors.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
pemulis: The only thing actually on-chain is the identifier string, right?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -174,6 +188,66 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    // This is a complicated looking map that maps each symbol into an entry in an object with its value the price
+    // feed created from the mapped config in allConfigs.
+    const priceFeedMap = Object.fromEntries(
+      await Promise.all(
+        symbols.map(async symbol => {
+          const config = allConfigs[symbol];
+
+          // If there is no config for this symbol, insert null and send an error.
+          if (!config) {
+            logger.error({
```
daywiss, line 225: why doesnt this just throw an error, or log error and throw error?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -174,6 +188,66 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+          // If this returns null, just return upstream since the error has already been logged and the null will be
+          // detected upstream.
+          const priceFeed = await createPriceFeed(logger, web3, networker, getTime, combinedConfig);
+          return [symbol, priceFeed];
+        })
+      )
+    );
+
+    // Return null if any of the price feeds in the map are null (meaning there was an error).
+    if (Object.values(priceFeedMap).some(priceFeed => priceFeed === null)) return null;
```
daywiss, line 246: if error was thrown earlier, i dont think this would be needed

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,154 @@
...
+    this.priceFeedMap = priceFeedMap;
+    this.decimals = decimals;
+  }
+
+  async getHistoricalPrice(time) {
+    const historicalPrices = {};
+    const errors = [];
+    await Promise.all(
+      Object.entries(this.priceFeedMap).map(async ([name, pf]) => {
+        const price = await pf.getHistoricalPrice(time).catch(err => errors.push(err));
```
daywiss, line 66: why accumulate errors here rather than throwing on the first one? you cant continue if you have a single error anyway

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,154 @@
...
+  return Array.from(input)
+    .map((char, index, array) => {
+      if (allowedSpecialCharacters.includes(char) && array[index - 1] !== "\\") {
+        return `\\${char}`;
+      } else {
+        return char;
+      }
+    })
+    .join("");
+}
```
daywiss, line 35: this function looks tricky, you might want a test for this, though im pretty sure its working by proxy of testing everything else

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -713,6 +713,119 @@ contract("CreatePriceFeed.js", function(accounts) {
...
+
+  it("ExpressionPriceFeed: constant expression", async function() {
+    const config = {
+      type: "expression",
+      expression: "1 + 2"
+    };
+
+    const expressionPriceFeed = await createPriceFeed(logger, web3, networker, getTime, config);
+
+    assert.isNotNull(expressionPriceFeed);
```
daywiss, line 826: nice test suite, wondering if its worth also testing for the error log?

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,154 @@
...
+        assert(price !== undefined && price !== null, "Valid price must be returned");
+        prices[name] = this._convertToDecimal(price, pf.getPriceFeedDecimals());
+      } catch (err) {
+        errors.push(err);
+      }
+    });
+
+    if (errors.length > 0) return null;
+
+    return this._convertToFixed(this.expressionCode.evaluate(prices));
```
daywiss, line 117: wonder if we can make this DRYer since the logic is really similar to getting historical price. its almost like you could create a higher order function which takes in a `getPrice` call which can represent getCurrentPrice or getHistorical price. wdyt?

packages/financial-templates-lib/test/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,163 @@
...
+    };
+
+    const expressionPriceFeed = new ExpressionPriceFeed(
+      priceFeedMap,
+      "(ETHUSD + COMPUSD) * (BTCUSD + COMPUSD) / 7.321"
+    );
+
+    // Should return the sum of the current prices.
+    // (1 + 9) * (2 + 9) / 7.321 = 15.025269771889086190 (rounded to 18 decimals).
+    assert.equal(expressionPriceFeed.getCurrentPrice(), toWei("15.025269771889086190"));
```
daywiss, line 59: really cool to see this in action!

packages/financial-templates-lib/test/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,163 @@
...
+    it("Escapes characters correctly", async function() {
+      assert.equal(escapeSpecialCharacters("USD-[bwBTC/ETH SLP]"), "USD\\-\\[bwBTC\\/ETH\\ SLP\\]");
+      assert.equal(escapeSpecialCharacters("-/][a--]rr]"), "\\-\\/\\]\\[a\\-\\-\\]rr\\]");
+    });
+
+    it("Processes escaped expressions correctly", async function() {
+      const escapedSymbol = "\\[1\\/\\-2\\]\\ \\-\\ 17";
+      assert.equal(math.evaluate(`${escapedSymbol} - 2 * ${escapedSymbol}`, { [escapedSymbol]: 5 }).toString(), "-5");
+
+      const unescapedSymbol = "USD-[bwBTC/ETH SLP]";
```
daywiss, line 151: would have been good for the tests above to use this identifier in one of the expressions

daywiss (commented):
really interesting approach and solution. you've touched on the implications of using this pattern which could be quite powerful!

packages/financial-templates-lib/test/price-feed/ExpressionPriceFeed.js
```diff
@@ -0,0 +1,163 @@
...
+      COMPUSD: new PriceFeedMock(toBN(toWei("9")), toBN(toWei("10")), 25, 17)
+    };
+
+    const invalidExpressionPriceFeed = new ExpressionPriceFeed(inValidDecimalsPriceFeeds, "ETHUSD * BTCUSD * COMPUSD");
+    assert.throws(() => invalidExpressionPriceFeed.getPriceFeedDecimals());
+  });
+
+  describe("Expression parsing with escaped characters", async function() {
+    it("Escapes characters correctly", async function() {
+      assert.equal(escapeSpecialCharacters("USD-[bwBTC/ETH SLP]"), "USD\\-\\[bwBTC\\/ETH\\ SLP\\]");
```
chrismaree, line 143: nice.

chrismaree (approved):
overall this looks great to me! only issue I can see is that currently you are not implementing `getHistoricalPricePeriods` so this won't work out the box with developer mining. However, this is probs fine as for now we dont have all feeds implementing this method always.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: This is the way we do this in the medianizer feed, so I just continued the pattern. The idea is that _all_ lower-level feed errors should be bubbled up for the purpose of logging in case multiple are failing. This has been useful for the complex stablespread feeds.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: This is how the others work I believe -- just continuing the pattern from above where we log an error and return null.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -174,6 +188,66 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
 
   // Internal helper methods:
 
+  // Returns an ExpressionPriceFeed.
+  async function _createExpressionPriceFeed(expressionConfig) {
+    // Build list of configs that could be used in the expression including default price feed configs and customFeeds
+    // that the user has provided inside the ExpressionPriceFeed config. Note: default configs are overriden by
+    // customFeeds with the same name. Tranform keys by escaping any special characters in the identifier names..
+    const allConfigs = Object.fromEntries(
```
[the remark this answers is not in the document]
mrice32: I hadn't either! It's a pretty new function from what I can tell. I doubt we could use it on node12 :)

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: It has a test (not sure if this was true when you reviewed).

mrice32: Sorry, I just now realized that I forgot to push a few changes, but there's definitely a test in there now :)

packages/financial-templates-lib/test/price-feed/ExpressionPriceFeed.js
mrice32: Done! Added it to one of the earlier tests.

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
mrice32: Done!

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: Yep!

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Agreed. However, the current pattern is to return null and log an error. I agree that this may be a bit nonsensical looking back, but I would prefer to update this pattern in a follow-up PR across this entire file rather than 1) being inconsistent or 2) creating a lot of unrelated churn by changing all of the creation functions.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
mrice32: Two issues I see with that approach (the first likely making this a nonstarter):
- getHistoricalPrice is async and this higher level function would likely need to be async as a result of calling getHistoricalPrice on other feeds. If this function is async, getCurrentPrice can't call it. _Technically_ there may be ways around this by trying to do the async stuff outside of the function, but I think that might be so complex as to defeat the purpose of this refactor.
- I suspect this may hurt readability. I'm not sure, though (I'd have to try to implement it to really tell).

Thoughts?

mrice32:
> only issue I can see is that currently you are not implementing `getHistoricalPricePeriods` so this won't work out the box with developer mining. However, this is probs fine as for now we dont have all feeds implementing this method always.

@chrismaree it's a great point. I think there are two good reasons for _not_ doing this:
- It's not clear that we're actually going to keep that method medium-term. It seems like we may be moving towards historical capture for dev mining.
- Not all constituent price feeds implement it, so we would have to assume that the passed in price feeds implement to call them even though they might not. I guess the consequence is that we'd have a method that will _sometimes_ always error depending on the sub-price feeds that you use in your expression.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js
daywiss: yes the sync/async issue does break that idea! thanks for thinking about this

daywiss (approved):
nice work, thanks for updates

PR #2514, opened 2021-02-08
fix(serverless-hub): Address dropped spoke logs and mismatch timedout spokes

chrismaree:
**Motivation**
This PR addresses two issues identified within the serverless configuration:

- Often spoke logs are dropped at the end of a spoke execution and are not logged correctly.
- There was a mismatch between the timed out spokes and the error spokes reported in hub logs. 

**Details**

**Problem 1:**
When debugging hub/spoke interactions, you can sometimes find that the bottom-most set of logs can get lost. The screenshot below shows the logs within a spoke for a given bot. This bot executed correctly and all logs were produced. however, some did not show up within GCP correctly. 
image

To deal with this issue, each bot's `index.js` file was updated to accommodate a 2-second timeout before the process terminates. This is a pretty shitty patch but it solves the issue and for now it'll be a bandaid until we can better deal with terminating transport issues.

**Problem 2:**
While working with the bots over last week it kept looking like the name of the bot reported in the `retriedSpokes` did not match the correct bot as the `retriedSpokes` always looked as if they completed their loops correctly. In particular, this log is nonsensical:
image

here you can see that the hub reported that the `retriedSpokes` is `stablespreadethmarch-mainnet-disputer` but the `errorOutputs` is `usdperl-mainnet-disputer` with this bot exceeding the timeout. This does not make sense as `usdperl-mainnet-disputer` shows up just fine within the `validOutputs`. This indicates that the `retriedSpokes` contains the wrong bot names.

After investigating locally I uncovered a small but meaningful error in how the hub logs are produced that mismatched these spokes. After this change, the re-tried spokes are correctly logged within the slack & GCP messages.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Note that this does not yet solve the underlying timeout issues but it should help a bit in getting to the resolution.

### DIFF
--- a/packages/disputer/index.js
+++ b/packages/disputer/index.js
@@ -178,6 +178,7 @@ async function run({
           message: "End of serverless execution loop - terminating process"
         });
         await waitForLogger(logger);
+        await delay(2); // waitForLogger does not always work 100% correctly in serverless. add a delay to ensure logs are captured upstream.
         break;
       }
       logger.debug({
@@ -207,10 +208,10 @@ async function Poll(callback) {
       empAddress: process.env.EMP_ADDRESS,
       // Default to 1 minute delay. If set to 0 in env variables then the script will exit after full execution.
       pollingDelay: process.env.POLLING_DELAY ? Number(process.env.POLLING_DELAY) : 60,
-      // Default to 5 re-tries on error within the execution loop.
-      errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
-      // Default to 10 seconds in between error re-tries.
-      errorRetriesTimeout: process.env.ERROR_RETRIES__TIMEOUT ? Number(process.env.ERROR_RETRIES__TIMEOUT) : 10,
+      // Default to 3 re-tries on error within the execution loop.
+      errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 3,
+      // Default to 1 seconds in between error re-tries.
+      errorRetriesTimeout: process.env.ERROR_RETRIES__TIMEOUT ? Number(process.env.ERROR_RETRIES__TIMEOUT) : 1,
       // Read price feed configuration from an environment variable. This can be a crypto watch, medianizer or uniswap
       // price feed Config defines the exchanges to use. If not provided then the bot will try and infer a price feed
       // from the EMP_ADDRESS. EG with medianizer: {"type":"medianizer","pair":"ethbtc",
--- a/packages/liquidator/index.js
+++ b/packages/liquidator/index.js
@@ -300,6 +300,7 @@ async function run({
           message: "End of serverless execution loop - terminating process"
         });
         await waitForLogger(logger);
+        await delay(2); // waitForLogger does not always work 100% correctly in serverless. add a delay to ensure logs are captured upstream.
         break;
       }
       logger.debug({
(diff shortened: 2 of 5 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -142,7 +142,7 @@ hub.post("/", async (req, res) => {
       });
       const rejectedRetryResults = await Promise.allSettled(rejectedRetryPromiseArray);
       rejectedRetryResults.forEach((result, index) => {
-        _processSpokeResponse(Object.keys(configObject)[index], result, validOutputs, errorOutputs);
+        _processSpokeResponse(retriedOutputs[index], result, validOutputs, errorOutputs);
```
chrismaree, line 145: this was the main bug in the severless hub

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -186,7 +186,7 @@ hub.post("/", async (req, res) => {
               spokeName: spokeName,
               errorReported: errorOutput.errorOutputs[spokeName].execResponse
                 ? errorOutput.errorOutputs[spokeName].execResponse.stderr
-                : JSON.stringify(errorOutput.errorOutputs[spokeName])
+                : errorOutput.errorOutputs[spokeName]
```
chrismaree, line 189: this enhances logs a bit. before:
image


after:
image

packages/disputer/index.js
```diff
@@ -178,6 +178,7 @@ async function run({
           message: "End of serverless execution loop - terminating process"
         });
         await waitForLogger(logger);
+        await delay(2); // waitForLogger does not always work 100% correctly in serverless. add a delay to ensure logs are captured upstream.
```
nicholaspai, line 181: Why was this failing before? Because GCP imposes some default timeout?

nicholaspai: Maybe its worth extending this even longer in case the spokes become more beefy in the future

packages/disputer/index.js
```diff
@@ -207,10 +208,10 @@ async function Poll(callback) {
       empAddress: process.env.EMP_ADDRESS,
       // Default to 1 minute delay. If set to 0 in env variables then the script will exit after full execution.
       pollingDelay: process.env.POLLING_DELAY ? Number(process.env.POLLING_DELAY) : 60,
-      // Default to 5 re-tries on error within the execution loop.
-      errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
-      // Default to 10 seconds in between error re-tries.
-      errorRetriesTimeout: process.env.ERROR_RETRIES__TIMEOUT ? Number(process.env.ERROR_RETRIES__TIMEOUT) : 10,
+      // Default to 3 re-tries on error within the execution loop.
```
nicholaspai, line 211: +1, 5 is a lot and 3 is enough to indicate an error I imagine

packages/disputer/index.js
nicholaspai: Shold we add this to the `disputerConfig` obj

packages/liquidator/index.js
```diff
@@ -300,6 +300,7 @@ async function run({
           message: "End of serverless execution loop - terminating process"
         });
         await waitForLogger(logger);
+        await delay(2); // waitForLogger does not always work 100% correctly in serverless. add a delay to ensure logs are captured upstream.
```
nicholaspai, line 303: same comment as disputer, should we parameterize this in `liquidatorConfig`

packages/monitors/index.js
```diff
@@ -274,6 +274,7 @@ async function run({
           message: "End of serverless execution loop - terminating process"
         });
         await waitForLogger(logger);
+        await delay(2); // waitForLogger does not always work 100% correctly in serverless. add a delay to ensure logs are captured upstream.
```
nicholaspai, line 277: Should we parameterize in `monitorConfig`

packages/serverless-orchestration/src/ServerlessHub.js
nicholaspai: aha nice catch

nicholaspai (approved):
Nice catches, left some comments about where to parameterize hard coded values

packages/monitors/index.js
chrismaree: > Nice catches, left some comments about where to parameterize hard coded values

right now I don't think we should parameterize them. Not really sure of a situation where we'd want it to not be the full 2-second delay. What scenario are you picturing where we'd want to change it?

PR #2520, opened 2021-02-09
improve(trader): add hardhat test setup

mrice32:
**Motivation**

The trader package doesn't currently have a test setup.

**Summary**

It was a little tricky to get typescript tests set up in hardhat using truffle5 and web3, but it's doable. This PR adds a sample test and shuffles the package configuration to make it so `hardhat test` works as one would expect. I suspect there may be other issues down the road, but this should be a good place to start and has the basics.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/trader/hardhat.config.ts
+++ b/packages/trader/hardhat.config.ts
@@ -1,6 +1,9 @@
-const { getHardhatConfig } = require("@uma/common");
+import "@nomiclabs/hardhat-web3";
+import "@nomiclabs/hardhat-truffle5";
 
-const path = require("path");
+import { getHardhatConfig } from "@uma/common";
+import path from "path";
+
 const coreWkdir = path.dirname(require.resolve("@uma/core/package.json"));
 const packageWkdir = path.dirname(require.resolve("@uma/optimistic-oracle/package.json"));
 
@@ -10,7 +13,7 @@ const configOverride = {
     sources: `${coreWkdir}/contracts`,
     artifacts: `${coreWkdir}/artifacts`,
     cache: `${coreWkdir}/cache`,
-    tests: `${packageWkdir}/dist/build/test`
+    tests: `${packageWkdir}/test`
   }
 };
 
--- a/packages/trader/package.json
+++ b/packages/trader/package.json
@@ -29,7 +29,7 @@
     "url": "git+https://github.com/UMAprotocol/protocol.git"
   },
   "files": [
-    "/src/**/*.js"
+    "/dist/**/*.js"
   ],
   "bin": "dist/index.js",
   "types": "dist/index.d.ts",
--- a/packages/trader/package.json
+++ b/packages/trader/package.json
@@ -9,9 +9,14 @@
     "dotenv": "^8.2.0"
   },
   "devDependencies": {
-    "typescript": "^4.1.3",
     "@tsconfig/node14": "^1.0.0",
-    "@types/async-retry": "^1.4.2"
+    "@types/async-retry": "^1.4.2",
+    "@types/chai": "^4.2.14",
+    "@types/mocha": "^8.2.0",
+    "@types/node": "^14.14.25",
+    "chai": "^4.3.0",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.3"
   },
   "homepage": "https://umaproject.org",
   "license": "AGPL-3.0-or-later",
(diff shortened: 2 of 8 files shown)


packages/trader/hardhat.config.ts
```diff
@@ -1,6 +1,9 @@
-const { getHardhatConfig } = require("@uma/common");
+import "@nomiclabs/hardhat-web3";
+import "@nomiclabs/hardhat-truffle5";
```
mrice32, line 2: This is needed to ensure that the type system pulls web3 and truffle artifacts into the hardhat import within the tests.

packages/trader/package.json
```diff
@@ -9,9 +9,14 @@
...
-    "typescript": "^4.1.3",
     "@tsconfig/node14": "^1.0.0",
-    "@types/async-retry": "^1.4.2"
+    "@types/async-retry": "^1.4.2",
+    "@types/chai": "^4.2.14",
+    "@types/mocha": "^8.2.0",
+    "@types/node": "^14.14.25",
+    "chai": "^4.3.0",
+    "ts-node": "^9.1.1",
+    "typescript": "^4.1.3"
```
mrice32, line 19: These are all necessary for hardhat's typescript testing setup.

packages/trader/package.json
```diff
@@ -24,15 +29,14 @@
...
+    "/dist/**/*.js"
   ],
   "bin": "dist/index.js",
   "types": "dist/index.d.ts",
   "main": "dist/index.js",
   "scripts": {
-    "build": "tsc && tsc -p tsconfig-build.json",
-    "test": "echo \"No tests configured!\"",
-    "test-log": "truffle test ./test/*.js --network=test logInTest"
+    "build": "tsc -b",
```
mrice32, line 38: `-b` allows us to use the "references" field in our config (explained below).

packages/trader/src/tsconfig.json
```diff
@@ -0,0 +1,9 @@
+{
```
mrice32, line 1: There are now two typescript "projects". This seems to be the new recommended setup for separating test and prod code in ts.

Effectively, everything in src is considered a separate project. It can be compiled on its own, but it's generally compiled as a dependency of the outer config that exists in the top level trader directory.

This has some performance enhancements, but the main benefit is that we can have a separate output dir -- `dist`. Hardhat automatically uses the tsconfig.json at the top level for testing (they don't let you point it to a different filename, which was the setup we had before), so that config _must_ build the test files. The output in that project, which are _distinct_ from those in this project, will be dumped in `build`. That will include the compiled hardhat config and all the compiled test files, etc.

packages/trader/test/index.ts
```diff
@@ -0,0 +1,23 @@
+import { run } from "../src/index";
+import { web3, assert } from "hardhat";
+
+describe("index.js", function() {
```
mrice32, line 4: This is a dumb sample test just to provide and example and prove the setup works.

packages/trader/tsconfig.json
```diff
@@ -1,8 +1,10 @@
 {
   "extends": "@tsconfig/node14/tsconfig.json",
-  "include": ["src/**/*.ts"],
+  "include": ["test/**/*.ts", "hardhat.config.ts"],
```
mrice32, line 3: Note: this only includes non-production files (aka all ts files outside of src).

chrismaree (approved):
image

packages/trader/package.json
```diff
@@ -24,15 +29,14 @@
     "url": "git+https://github.com/UMAprotocol/protocol.git"
   },
   "files": [
-    "/src/**/*.js"
+    "/dist/**/*.js"
```
daywiss, line 32: is this an input our output dir?

packages/trader/src/tsconfig.json
daywiss: i think i get this. so theres a tsconfig in src which builds bot independently from tests, then there is a tsconfig in root which builds tests as well as src? If this is the case, what are your thoughts about keeping them _more_ seperate? ie tests and src contain their own tsconfigs and they are siblings rather parent child? 

I guess my thought process is i would rather not treat tests as a special case (root tsconfig), and use the same abstraction you did for src, treat it as its own project.

daywiss (approved):
just have a comment on organization but nothing to block you

packages/trader/src/tsconfig.json
mrice32: The issue is that I believe the hardhat config needs to be in the same directory where 1) you run hardhat and 2) the test tsconfig.json is. So unless we always want to run hardhat from test/, we need to have the test tsconfig at the root of the package. Thoughts?

packages/trader/package.json
mrice32: files (at least as far as I understand) tells the release process what files to include in the npm release.

In this case, it would just be the files compiled by the prod tsconfig, which get dumped in dist. Everything else is for local development if I understand correctly.

packages/trader/src/tsconfig.json
mrice32: @daywiss gonna go ahead and merge because I don't want to block @chrismaree, but happy to modify this setup in a follow-up!

PR #2530, opened 2021-02-10
refactor(liquidator): Simplify liquidatorStrategy class and remove `whaleDefenseFundWei`

nicholaspai:

**Motivation**

The helper methods in this class are renamed to better reflect their return value, and the possible paths that `processPosition` can take are better outlined.

We are able to remove `whaleDefenseFundWei` from the `liquidatorConfig` now. The only modification to the liquidator flow now is for the situation where:
- **there is a withdrawal pending AND the whale defense strategy is active AND the liquidator does NOT have enough balance to fully liquidate the position**

Currently, the bot will use up as much of its balance minus `whaleDefenseFundWei` to send a liquidation.

This PR removes the `whaleDefenseFundWei` parameter and will NOT liquidate the position in this situation, until the `defenseActivationThreshold` is passed.

**Summary**

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -50,6 +51,62 @@ describe("LiquidatorStrategy", () => {
     result = strat.utils.withdrawProgressPercent(1000, 2000, 1500);
     assert.equal(result.toFixed(2), "50.00");
   });
+  it("hasWithdrawRequestPending", () => {
```
nicholaspai, line 54: I added unit tests for each helper function in the `utility` export for liquidatorStrategy.js

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -66,12 +61,11 @@ module.exports = (
   assert(BN, "requires BN");
 
   // these variables are optional and are only checked if WDF is activated
-  whaleDefenseFundWei &&
-    assert(
-      defenseActivationPercent >= 0 && defenseActivationPercent <= 100,
-      "Requires defenseActivationPercent to bet set between 0 and 100"
-    );
-  whaleDefenseFundWei && assert(withdrawLiveness > 0, "requires withdrawLiveness");
+  const wdfActive = !isNaN(defenseActivationPercent) && defenseActivationPercent > 0;
```
nicholaspai, line 64: Previously, setting `whaleDefenseFundWei > 0` would activate the WDF. Now, setting `defenseActivationPercent >0` activates it.

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -142,46 +129,44 @@ module.exports = (
...
+    let tokensToLiquidate = toBN(0);
 
-    // WDF strategy not activated, calculate tokens to liquidate respecting all constraints
-    const tokensToLiquidate = calculateTokensToLiquidate({
-      syntheticTokenBalance,
-      positionTokens: position.numTokens,
-      maxTokensToLiquidateWei,
-      withdrawRequestPending: hasWithdrawRequestPending({ position, currentBlockTime })
-    });
+    if (!wdfActive || !hasWithdrawRequestPending({ position, currentBlockTime })) {
```
nicholaspai, line 134: Hopefully this if-else if-else layout is a bit clearer to read. If its not, then this PR is kind of pointless :D

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -219,24 +204,24 @@ module.exports = (
   // Returns true if enough of the withdrawal liveness has passed that the WDF strategy can be activated.
   function passedDefenseActivationPercent({ position, currentBlockTime }) {
     return (
-      hasWithdrawRequestPending({ position, currentBlockTime }) &&
```
nicholaspai, line 222: I decoupled helper methods from each other, this way the function names make it more clear what they are returning

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -219,24 +204,24 @@ module.exports = (
...
       withdrawProgressPercent(withdrawLiveness, position.withdrawalRequestPassTimestamp, currentBlockTime) >=
-        parseFloat(defenseActivationPercent)
+      parseFloat(defenseActivationPercent)
     );
   }
 
-  // Should we try to delay this withdrawal. Update this logic to change conditions to run delay strategy.
-  function shouldLiquidateMinimum({ position, syntheticTokenBalance, currentBlockTime }) {
```
nicholaspai, line 229: This function has too many branches and is the main reason why I think this refactor will make the class more readable.

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -142,46 +129,44 @@ module.exports = (
...
+        position: position,
+        maxLiquidationPrice: maxCollateralPerToken.toString(),
+        syntheticTokenBalance: syntheticTokenBalance.toString(),
+        maxTokensToLiquidateWei: maxTokensToLiquidateWei ? maxTokensToLiquidateWei.toString() : null,
+        currentBlockTime,
+        ...logInfo
+      });
+      tokensToLiquidate = minSponsorSize;
+    } else {
+      // 4) Do nothing.
```
daywiss, line 168: just trying to follow the logic, this would be a case where:
- wdf not active and there's a pending withdraw request and we cant liquidate for some reason
- wdf active and there's no pending withdraw request and we cant liquidate for some reason
-
I think its probably right, and in this case we return nothing from this function which might trigger a log in the liquidator. Is that ok and does that log mean anything to us?

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -142,46 +129,44 @@ module.exports = (
...
+    }
 
     // check if we should try to liquidate this position
     if (shouldLiquidate({ tokensToLiquidate })) {
```
daywiss, line 172: i wonder if this function is still needed after all the checks in the if/else?

packages/liquidator/test/Liquidator.js
```diff
@@ -1272,185 +1272,10 @@ contract("Liquidator.js", function(accounts) {
             assert.ok(liquidator);
           });
           versionedIt([{ contractType: "any", contractVersion: "any" }])(
-            "full liquidation: should enable and not affect existing logic if not triggered",
```
daywiss, line 1275: im good with removing tests. just would like your comment on why this is possible now

daywiss (commented):
Really like how you simplified utility functions and added tests for them. I think the new logic flow makes sense, but would be good to get some more eyes on just to 👍 Couple issues with tests but for the most part looks good.

packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -73,119 +130,71 @@ describe("LiquidatorStrategy", () => {
...
       syntheticTokenBalance: 100,
       positionTokens: 10,
-      maxTokensToLiquidateWei: "1000",
-      withdrawRequestPending: true
+      maxTokensToLiquidateWei: "1000"
     });
     // should respect position size
     assert.equal(result.toString(), "10");
 
-    strat = Strategy(
```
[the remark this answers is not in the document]
nicholaspai: Because these tests either dealt with the removed `whaleDefenseFundWei` value or with a removed `withdrawPending` parameter

packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -73,119 +130,71 @@ describe("LiquidatorStrategy", () => {
...
-
     result = strat.utils.calculateTokensToLiquidate({
       syntheticTokenBalance: 100,
       positionTokens: 101,
-      maxTokensToLiquidateWei: "1000",
-      withdrawRequestPending: true
+      maxTokensToLiquidateWei: "1000"
     });
     // should respsect empminsponsor size
     assert.equal(result.toString(), "1");
```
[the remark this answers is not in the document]
nicholaspai: Nope. Position has 101 tokens and liquidator has only 100 tokens to liquidate. Min sponsor size is 100, so liquidator either has to liquidate 101 or 1 to keep position above min sponsor size.

packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -197,131 +206,153 @@ describe("LiquidatorStrategy", () => {
...
       currentBlockTime: 500,
       maxCollateralPerToken: "0"
     });
-    // expect 1 log notifying user of wdf
-    assert.equal(eventlist.length, 1);
-    // This should liquidate entire position
-    assert(result[3].rawValue, config.minSponsorSize);
-    events.removeAllListeners("log");
-
+    assert(result[3].rawValue, "100");
```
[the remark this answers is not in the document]
nicholaspai: Great catch and fixed other similar instances

packages/liquidator/src/liquidationStrategy.js
nicholaspai: > wdf not active and there's a pending withdraw request and we cant liquidate for some reason
If "wdf is not active", then the first case gets triggered `if (!wdfActive || !hasWithdrawRequestPending({ position, currentBlockTime }))` and the liquidator will use whatever balance it has.

> wdf active and there's no pending withdraw request and we cant liquidate for some reason
If the "wdf is active", and there is pending withdrawal, then the first case also gets triggered because of the RHS. This fourth case occurs where the wdf is active AND the liquidator cannot fully liquidate a position BUT either the `defensePassedActivationPercent` has not triggered or the liquidator doesn't have enough balance to submit a min liquidation.

Does that make sense?

> I think its probably right, and in this case we return nothing from this function which might trigger a log in the liquidator. Is that ok and does that log mean anything to us?

Yes right now we actually try to parse this situation in `liquidator` and send a cutomized log in this [section](https://github.com/UMAprotocol/protocol/blob/master/packages/liquidator/src/liquidator.js#L300)

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Currently the function checks that `tokensToLiquidate >= 0`, but I'm not sure that's even possible now. I can remove if you feel OK with that.

packages/liquidator/test/Liquidator.js
nicholaspai: This test checks that the liquidator will use its "full balance" if the `whaleDefnseFundWei` is set to a trivially small value:

```
in this test, the WDF is not triggered because the liquidator has enough available capital
to liquidate a full position.
```

So this case is not relevant anymore

packages/liquidator/test/Liquidator.js
```diff
@@ -1272,185 +1272,10 @@ contract("Liquidator.js", function(accounts) {
...
-                (await emp.getLiquidations(sponsor1))[0].state,
-                LiquidationStatesEnum.UNINITIALIZED
-              );
-
-              // The other two positions should not have any liquidations associated with them.
-              assert.deepStrictEqual(await emp.getLiquidations(sponsor2), []);
-            }
-          );
-          versionedIt([{ contractType: "any", contractVersion: "any" }])(
-            "should trigger and only submit the emp minimum position",
```
nicholaspai, line 1388: @daywiss This test currently checks that the liquidator only has `balance - whaleDefenseFundWei` tokens to possibly liquidate with and therefore can only submit min liquidations on large enough positions. So this test isn't relevant

packages/liquidator/test/Liquidator.js
```diff
@@ -1272,185 +1272,10 @@ contract("Liquidator.js", function(accounts) {
...
-              // minimum liquidation size / outstanding tokens * (outstanding collateral - withdraw amount)
-              // 5 / 100 * ( 120 -10) = 5.5
-              assert.equal(liquidationObject.liquidatedCollateral, convertCollateral("5.5"));
-              // minimum liquidation size / outstanding tokens * outstading collateral
-              // 5 / 100 * 120 = 6
-              assert.equal(liquidationObject.lockedCollateral, convertCollateral("6"));
-            }
-          );
-          versionedIt([{ contractType: "any", contractVersion: "any" }])(
-            "trigger multiple extensions and finally full liquidation",
```
nicholaspai, line 1446: @daywiss This test is renamed and is able to check all possible cases in one:

- WDF is active
- Sponsor1 and sponsor2 create 100 token positions
- Liquidator starts with 150 tokens
- Sponsor1 and sponsor2 both request withdrawals
- Liquidator liquidates `sponsor1` who can be fully liquidated, bot has 50 tokens left
- Liquidator can only submit min liquidations on sponsor2, but not until the activation threshold has passed
- Finally, once sponsor2's withdrawal has expired, check that liquidator uses its full balance to liquidate sponsor2

packages/liquidator/src/liquidationStrategy.js
daywiss: yea that makes a lot of sense. thanks for breaking it down. Also looked at logs and that looks good to me

daywiss (approved):

packages/liquidator/src/liquidationStrategy.js
nicholaspai: I do think we should move all the log logic to the `liquidatorStrategy` rather than reading it from `liquidator` but I'll save for another PR and curious what other eng's think

packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -50,6 +51,62 @@ describe("LiquidatorStrategy", () => {
...
+    // current block = 900, defense activation = 0%
+    // ==> FALSE
+    let result = strat.utils.passedDefenseActivationPercent({
+      position: { withdrawalRequestPassTimestamp: 1000 },
+      currentBlockTime: 900
+    });
+    assert(!result);
+    // current block = 950, defense activation = 50%
+    // ==> FALSE
+    result = strat.utils.passedDefenseActivationPercent({
```
pemulis, line 97: I think I have a handle on all of this except for the defense activation percent. Could you explain what this is and how it's used in the context of the whale defense fund strategy?

nicholaspai: When a withdrawal is requested it has a `liveness` period during which it cannot be executed. If a position with a pending withdrawal is liquidated, then its liveness gets extended. For example, if a liveness period is 100 seconds long, and 20 seconds have elapsed when a liquidation occurs, then another 100 seconds gets added to the withdrawal's liveness. In this example, the 80 seconds remaining get another 100 seconds for a total of 180 seconds of liveness.

The `defenseActivatioPercent` controls when liquidations are sent, so that bots can efficiently extend withdrawal periods. If the `defenseActivationPercent` is 50%, then a liquidation is only sent when at least 50% of the withdrawal's liveness has expired. Using the previous example of 100 second withdrawal liveness, the first liquidation can be submitted after 50 seconds have elapsed. The new withdrawal liveness will be 50 + 100 = 150 seconds. The next liquidation will occur after 75 seconds. And so on.

packages/liquidator/mocha-test/liquidationStrategy.js
```diff
@@ -67,125 +124,77 @@ describe("LiquidatorStrategy", () => {
     assert.equal(result[1].rawValue, "0");
     assert.equal(result[2].rawValue, params.maxCollateralPerToken);
     assert.equal(result[3].rawValue, params.tokensToLiquidate);
-    // curren blocktime=100 plus default liquidationDeadline=300
-    assert(result[4], 400);
```
chrismaree, line 71: poison syntax! this was not testing anything before 💀

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -142,46 +129,44 @@ module.exports = (
...
+        positionTokens: position.numTokens,
+        maxTokensToLiquidateWei
+      });
+    } else if (
+      wdfActive &&
+      hasWithdrawRequestPending({ position, currentBlockTime }) &&
+      passedDefenseActivationPercent({ position, currentBlockTime }) &&
+      canLiquidateMinimum({ position, syntheticTokenBalance })
+    ) {
+      // 3) If WDF is active and withdraw has passed liveness, then submit min liquidation
```
chrismaree, line 156:
```suggestion
      // 3) If WDF is active and withdraw has passed WDF activation percentage, then submit min liquidation
```

packages/liquidator/src/liquidationStrategy.js
chrismaree, line 168: do you want to emit some kind of log in the "do nothing case"

chrismaree (approved):
This is great! good simplification & removal of unsanitary variables. I'm happy with the outcome of this PR

LGTM!

packages/liquidator/src/liquidationStrategy.js
mrice32, line 64: nit: why did you choose isNaN? If you're trying to verify the type, would it be clearer to check `typeof defenseActivationPercent === "number"`?

mrice32 (commented):
Looks great! I think @chrismaree and @daywiss have a strong handle on this one, so I'll defer to them.

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Yes but let's leave that for another PR. There is some complex logic in the `liquidator` that tries to parse the liquidatorStrategy's choice and emit a log. See [this discussion](https://github.com/UMAprotocol/protocol/pull/2530#discussion_r574551579)

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Yeah that's reasonable its a more strict check

packages/liquidator/src/liquidationStrategy.js
daywiss: i think its reasonable to do it in a later PR if needed

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -168,15 +168,15 @@ module.exports = (
...
-      // We should liquidate the entire position using our capital - wdf
+    if (toBN(tokensToLiquidate).gt(toBN(0))) {
       return createLiquidationParams({
         sponsor: position.sponsor,
         tokensToLiquidate,
         currentBlockTime,
         maxCollateralPerToken
       });
+    } else {
+      return null;
```
daywiss, line 179: just a nit, not a big deal, but this is redundant. can just remove the whole else block and the function will return undefined. if you are explicitly checking for `null` then just return null without the else, since previous if block exits early.

daywiss (approved):

nicholaspai:
@daywiss @chrismaree I know I already said that I'm not going to fix up the customized log logic explaining why a liquidation was skipped, but I ended up doing so in [this commit](https://github.com/UMAprotocol/protocol/pull/2530/commits/38de5e93eff271bbfc832d87a65321138b7eb0d0) and this cleans up the `liquidator.js` code a lot. This moves all of the complex logging logic to `liquidatorStrategy.processPositions` wdyt?

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Actually, I'm going to implement this change, see below

nicholaspai: Actually, I'm going to implement this change, see below

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -115,17 +115,29 @@ module.exports = (
...
     syntheticTokenBalance,
     currentBlockTime,
     maxCollateralPerToken,
     // mandated token to liquidate amount, an override
     maxTokensToLiquidateWei,
+    shouldEmitWDFSkipLogs,
     ...logInfo
   }) {
-    assert(syntheticTokenBalance, "requires syntheticTokenBalance");
+    if (!syntheticTokenBalance || toBN(syntheticTokenBalance).lte(toBN(0))) {
```
daywiss, line 131: im ok with this still throwing an error rather than returning. wdyt?

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -115,17 +115,29 @@ module.exports = (
...
+  // Responsible for emitting customized logs since its already parsing
+  // whether a liquidation is valid to send and for how many tokens.
   function processPosition({
     position,
     syntheticTokenBalance,
     currentBlockTime,
     maxCollateralPerToken,
     // mandated token to liquidate amount, an override
     maxTokensToLiquidateWei,
+    shouldEmitWDFSkipLogs,
```
daywiss, line 128: not a big fan of injecting a logging flag into this function, would rather like to see if this could be handled in the event handler in liquidator. i realize though this may be the best way as it is

edit: saw this is based on `blockWindowHasRequestedWithdraw`.  i would actually be ok with this  name as it represents how the boolean is calculated. its not logging specific, and theoretically could be used as additional info to process position.

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -156,27 +168,68 @@ module.exports = (
...
+          message: "Liquidator bot skipping liquidation, withdrawal liveness has not passed WDF activation threshold😴",
+          position,
+          currentBlockTime,
+          defenseActivationPercent,
+          ...logInfo
+        });
+      }
+      // Case 2: Liquidator doesn't have enough balance to liquidate the minimum sponsor size
+      else if (!canLiquidateMinimum({ position, syntheticTokenBalance })) {
+        console.log("test");
```
daywiss, line 205: stray

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -156,27 +168,68 @@ module.exports = (
...
+      // Case 1: WDF is active but liveness threshold has not passed yet
+      if (
+        wdfActive &&
+        hasWithdrawRequestPending({ position, currentBlockTime }) &&
+        !passedDefenseActivationPercent({ position, currentBlockTime })
+      ) {
+        // If requested withdrawal was within the block window, then upgrade the log level.
+        // This makes it possible to emit only one log about skipping a liquidation
+        // due to the WDF activation threshold not being passed.
+        let logLevel = shouldEmitWDFSkipLogs ? "info" : "debug";
```
daywiss, line 193: does info trigger PD and debug does not?

packages/liquidator/src/liquidator.js
```diff
@@ -282,61 +293,15 @@ class Liquidator {
         // minimim position size, as well as minimum liquidation to extend withdraw
         empMinSponsorSize: this.empMinSponsorSize,
         currentBlockTime,
+        // should bot emit specialized logs about skipping liquidations due to WDF
+        shouldEmitWDFSkipLogs: blockWindowHasRequestedWithdraw,
```
daywiss, line 297: i think this will trigger info logs when there have been withdraws which are expired. is this a problem?

daywiss (commented):
i agree with the direction of this pr, it makes it a lost easier to reason about the logs. i have concerns mainly around triggering that log which can be either info or debug. i think maybe we should just simplify it and choose to always do like `debug` and then allow the monitor to actually trigger PD when balance gets low.

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Yeah I actually changed this to make an existing test that checks for an `error` log pass, but I'll just remove that test. I'd rather reduce bot churn

packages/liquidator/src/liquidationStrategy.js
nicholaspai: Let's just make this `debug` only for now

packages/liquidator/src/liquidationStrategy.js
nicholaspai: `info` triggers a Slack alert and debug does not. For sake of simplicity and little added benefit since no PD lett's just revert to debug

packages/liquidator/src/liquidator.js
nicholaspai: No if the withdraw has expired, then `hasWithdrawRequestPending` will be false so the liquidator will use its full balance to liquidate

nicholaspai:
@daywiss responded to your comments and I agree with changing the skipping-wdf log to `debug` only to simplify logic. I changed tests accordingly.

daywiss (approved):

packages/liquidator/src/liquidationStrategy.js
daywiss: for the record i am fine with either debug or info, just was concerned if it was a PD level alert

packages/liquidator/src/liquidationStrategy.js
daywiss: yea np. i actually meant that it could both log and throw error. but i see you changed it to just throw error, which i am fine with too.

packages/liquidator/src/liquidationStrategy.js
nicholaspai: If its info it will keep spamming the Slack channel so better to keep debug for now

Issue #1808, opened 2020-08-05
Create new Kovan release with updated contracts


mrice32:
**Why**

Contracts on Kovan are somewhat disorganized (many deployments) and out of date.

**How**

Do a redeployment of the contracts and verify them (this will require https://github.com/UMAprotocol/protocol/issues/1807).

**Resolution**

Newly deployed contracts should be verified and _at least_ listed in `core/networks/42.json`.

**Difficulty Score [1-10]**

4: ~2 hours

PR #2531, opened 2021-02-10
feat(deployment): example deploy scripts for EMP and Perpetual

pemulis:

**Motivation**

We want to make it easy to deploy new EMP and Perpetual contracts on Kovan, so the deployment scripts have been updated to accommodate the new contracts.

https://github.com/UMAprotocol/protocol/issues/1808

**Summary**

Updated EMP and Perpetual deployment scripts, and added the new addresses to `42.json`.

**Details**

Currently breaks the test suite but deploys nicely to Kovan. Trying to figure out why.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://github.com/UMAprotocol/protocol/issues/1808

### DIFF
--- a/packages/core/scripts/local/DeployEMP.js
+++ b/packages/core/scripts/local/DeployEMP.js
@@ -28,7 +28,7 @@ const argv = require("minimist")(process.argv.slice(), {
   boolean: ["test"],
   string: ["identifier", "collateral", "cversion"]
 });
-const abiVersion = argv.cversion || "latest"; // Default to most recent mainnet deployment, latest.
+const abiVersion = argv.cversion || "1.2.2"; // Default to old EMP contract by default, for tests.
 
 // Deployed contract ABI's and addresses we need to fetch.
 const ExpiringMultiPartyCreator = getTruffleContract("ExpiringMultiPartyCreator", web3, abiVersion);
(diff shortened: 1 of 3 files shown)


pemulis:
@mrice32 I'm trying to better understand how our internal deployment scripts are used, since I think that's at the root of whatever is going wrong here in the tests. Could you give me some background on these scripts?

Also, would it be a better use of my time at the moment to shift to documenting the steps for deploying to Kovan, so we can apply it towards mainnet, and creating a perpetual stub repo? This pull request is where I've gotten so far in cleaning up the scripts I used to deploy to Kovan, so that others could use them, but it sounds like this is the wrong repo for that, anyway.

packages/core/scripts/local/DeployEMP.js
```diff
@@ -28,11 +28,12 @@ const argv = require("minimist")(process.argv.slice(), {
   boolean: ["test"],
   string: ["identifier", "collateral", "cversion"]
 });
-const abiVersion = argv.cversion || "1.2.2"; // Default to most recent mainnet deployment, 1.2.2.
+const abiVersion = argv.cversion || "latest"; // Default to most recent mainnet deployment, latest.
```
nicholaspai, line 31: I don't think we should change the default on this script yet since other partners might be using this script on the older factory

packages/core/scripts/local/DeployEMP.js
```diff
@@ -58,6 +59,8 @@ const deployEMP = async callback => {
     const finder = await Finder.deployed();
     const store = await Store.deployed();
     const tokenFactory = await TokenFactory.deployed();
+    const expiringMultiPartyLib = await ExpiringMultiPartyLib.deployed();
```
nicholaspai, line 62: Why do you need the EMPLib?

mrice32:
@pemulis is this PR still relevant?

pemulis:
> @pemulis is this PR still relevant?

I think it's irrelevant now that we have the `launch-emp` and `launch-perpetual` stubs updated.

PR #2534, opened 2021-02-12
feat(price-feeds) : Add PerlinXPriceFeed 

pisuthd:
**Motivation**

XAUPERL and XAUUSD will need to be supported according to UMIP-26 (https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-26.md)

**Summary**

This adapter and codebase was inspired by `DominationFinancePriceFeed` and `CryptoWatchPriceFeed` with rarely changes its logic. Those identifiers are request data from various sources as following:
* XAUPERL - XAUUSD from Tradermade API divided by PERLUSDT from CryptoWatch API (Binance)
* XAUUSD - XAUUSD from Tradermade API

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


pemulis (commented):
Can you sign your old commit messages so that the pull request passes the DCO check? @mrice32 shared a good walk-through here:

https://github.com/UMAprotocol/protocol/pull/2346#pullrequestreview-586251293

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -361,6 +361,19 @@ const defaultConfigs = {
       { type: "cryptowatch", exchange: "gateio", pair: "amplusdt" },
       { type: "cryptowatch", exchange: "bitfinex", pair: "amplusd" }
     ]
+  },
+  XAUUSD: {
```
nicholaspai, line 365: Are there any financial contracts that are currently using this identifier? If not then no need to add XAUUSD to this file.

packages/financial-templates-lib/src/price-feed/PerlinXPriceFeed.js
```diff
@@ -0,0 +1,291 @@
...
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+   * @param {Number} periods Number of minutes interval between price entries.
+   */
+  constructor(
+    logger,
+    web3,
+    tradermadeApiKey,
+    cryptowatchApiKey,
+    pair,
+    convertToPerl,
```
nicholaspai, line 29:
```suggestion
    convertToPerl = false,
```

Instead of setting `config.convertToPerl || false` in the `CreatePriceFeed.js` I think its a bit cleaner to set a default value in this constructor

packages/financial-templates-lib/src/price-feed/PerlinXPriceFeed.js
```diff
@@ -0,0 +1,291 @@
...
+      // Note: Must ensure that `number` has no more decimal places than `priceFeedDecimals`.
+      return this.toBN(parseFixed(number.toString().substring(0, priceFeedDecimals), priceFeedDecimals).toString());
+    };
+  }
+
+  getCurrentPrice() {
+    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+  }
+
+  getHistoricalPrice(time, verbose = false) {
```
nicholaspai, line 68: This should be an `async` method now and should throw errors instead of returning `null` where a price can't be retrieved. Follow [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L69) for reference please

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
mrice32: Do we generally not preemptively add to this file once an identifier has been approved?

mrice32 (commented):
@pisuthd We have a new feature that we think would make it so you don't need to write a new price feed! As you can see [here](https://github.com/UMAprotocol/protocol/pull/2479), there's a tradermade price feed that should be merged soon. We also have a new "expression" price feed that allows you to arbitrarily combine price feeds in mathematical expressions. To summarize all that, I think you should be able to just add a new configuration (once the aforementioned PR is merged). Something like the following might work:
```js
XAUPERL: {
  type: "expression",
  expression: "XAUUSD * USDPERL"
},
XAUUSD: {
  type: "tradermade",
  pair: "XAUUSD",
  minTimeBetweenUpdates: 300,
  minuteLookback: 7200,
  hourlyLookback: 7200
}
```

@nicholaspai and @smb2796, could y'all check this config and make sure my reasoning makes sense?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: Generally no but there's no problem preemptively doing so. However if we're going to use the `ExpressionPriceFeed` here then we'll want the XAUUSD one

nicholaspai:
> @pisuthd We have a new feature that we think would make it so you don't need to write a new price feed! As you can see [here](https://github.com/UMAprotocol/protocol/pull/2479), there's a tradermade price feed that should be merged soon. We also have a new "expression" price feed that allows you to arbitrarily combine price feeds in mathematical expressions. To summarize all that, I think you should be able to just add a new configuration (once the aforementioned PR is merged). Something like the following might work:
> 
> ```js
> XAUPERL: {
>   type: "expression",
>   expression: "XAUUSD * USDPERL"
> },
> XAUUSD: {
>   type: "tradermade",
>   pair: "XAUUSD",
>   minTimeBetweenUpdates: 300,
>   minuteLookback: 7200,
>   hourlyLookback: 7200
> }
> ```
> 
> @nicholaspai and @smb2796, could y'all check this config and make sure my reasoning makes sense?

So just specifying the expression `XAUUSD * USDPERL` will work because `XAUUSD` and `USDPERL` are both specified in the `DefaultPriceFeedConfig`?

mrice32:
> > @pisuthd We have a new feature that we think would make it so you don't need to write a new price feed! As you can see [here](https://github.com/UMAprotocol/protocol/pull/2479), there's a tradermade price feed that should be merged soon. We also have a new "expression" price feed that allows you to arbitrarily combine price feeds in mathematical expressions. To summarize all that, I think you should be able to just add a new configuration (once the aforementioned PR is merged). Something like the following might work:
> > ```js
> > XAUPERL: {
> >   type: "expression",
> >   expression: "XAUUSD * USDPERL"
> > },
> > XAUUSD: {
> >   type: "tradermade",
> >   pair: "XAUUSD",
> >   minTimeBetweenUpdates: 300,
> >   minuteLookback: 7200,
> >   hourlyLookback: 7200
> > }
> > ```
> > 
> > 
> > @nicholaspai and @smb2796, could y'all check this config and make sure my reasoning makes sense?
> 
> So just specifying the expression `XAUUSD * USDPERL` will work because `XAUUSD` and `USDPERL` are both specified in the `DefaultPriceFeedConfig`?

yep, exactly.

smb2796:
> @nicholaspai and @smb2796, could y'all check this config and make sure my reasoning makes sense?

Yep, this reasoning makes sense to me!

> Do we generally not preemptively add to this file once an identifier has been approved?

Not for older price ids that didn't have supporting price feeds at the time of their approval (like XAUUSD). Also missing for a few recently approved price identifiers that use CW or Uniswap feeds, but don't actually have live contracts yet. 

We should make this part of our price identifier approval process though. I can go back through existing identifiers and add all of the ones that can be currently supported.

daywiss:
will need to re check into this PR once expressions feed goes in. maybe can be updated to utilize that.

pisuthd:
I'm working on other priority tasks at the moment, hope to continue working on this PR in coming days.

smb2796:
I think this PR can be closed. The [ExpressionPriceFeed](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.js) has been merged, so like @mrice32 explained, this should be able to be supported by just adding a default configuration. I opened a PR to do this: #2582

pisuthd:
@smb2796  Alright. I'm closing it now.

PR #2537, opened 2021-02-14
feat(pricefeeds): Add COMPUSDC-APR-MAR28/USDC

evanmays:
**Motivation**

This price identifier is self referential. It tracks the price of the synthetic that uses it.
UMIP-38 for price identifier spec: https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-38.md

**Summary**

Added config for tracking the Uniswap pool for new synthetic.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -343,7 +343,8 @@ const defaultConfigs = {
   "COMPUSDC-APR-MAR28/USDC": {
     type: "uniswap",
     uniswapAddress: "0xd8ecab1d50c3335d01885c17b1ce498105238f24",
-    twapLength: 7200
+    twapLength: 7200,
+    poolDecimals: 6
   },
   BTCDOM: {
     type: "domfi",
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -340,6 +340,11 @@ const defaultConfigs = {
     uniswapAddress: "0x683ea972ffa19b7bad6d6be0440e0a8465dba71c",
     twapLength: 7200
   },
+  "COMPUSDC-APR-MAR28/USDC": {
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -340,6 +340,11 @@ const defaultConfigs = {
     uniswapAddress: "0x683ea972ffa19b7bad6d6be0440e0a8465dba71c",
     twapLength: 7200
   },
+  "COMPUSDC-APR-MAR28/USDC": {
+    type: "uniswap",
+    uniswapAddress: "0xd8ecab1d50c3335d01885c17b1ce498105238f24",
+    twapLength: 7200
```
mrice32, line 346:
```suggestion
    twapLength: 7200,
    poolDecimals: 6,
    priceFeedDecimals: 6
```

mrice32 (commented):
Looks great! I think we need to specify the poolDecimals for the uniswap feed to function correctly. @nicholaspai can you check my suggestion? I assume this would've _technically_ worked before since the pool decimals and price feed decimals matched, but I think we should be explicit that the pool will output a 1e6 value and that the price feed should also output a 1e6 value. Thoughts?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: I agree with this

mrice32: edited to remove the priceFeedDecimals

evanmays: Cool, I'll test again

evanmays:
Just did an end to end test. Still works! `poolDecimals` is included now

nicholaspai (approved):
LGTM

mrice32 (approved):
LGTM!

chrismaree (approved):

smb2796 (approved):

PR #2546, opened 2021-02-16
feat(optimistic-oracle-proposer): Add dispute logic

nicholaspai:

**Motivation**

Part of #2411

Adds dispute logic to Optimistic Oracle proposer bot.

**Summary**

- Added event parameters to `OptimisticOracle.sol` to simplify bot logic. See comments for more details.
- Implemented `sendDisputes` in the `packages/OptimisticOracle: proposer.js` bot which checks proposed prices for `undisputed proposals` and sends disputes where `disputePrice != proposePrice`

Future work:
- Add a disputePrice buffer configuration variable to the OO bot which allows `disputePrice` and `proposePrice` to diverge beyond a certain precision, (e.g. 5 decimals of precision), without having to send a dispute.
- Add settlePrice feature

**Details**

- Adding event parameters to `OptimisticOracle.sol` should be harmless but its important to note that this would change Contract bytecode from that in the latest OpenZeppelin audit.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -93,7 +93,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         uint256 timestamp,
         bytes ancillaryData,
         int256 proposedPrice,
-        uint256 expirationTimestamp
+        uint256 expirationTimestamp,
+        address currency
     );
     event DisputePrice(
         address indexed requester,
@@ -101,7 +102,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         address indexed disputer,
         bytes32 identifier,
         uint256 timestamp,
-        bytes ancillaryData
+        bytes ancillaryData,
+        int256 proposedPrice
     );
     event Settle(
         address indexed requester,
@@ -300,7 +302,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
             timestamp,
             ancillaryData,
             proposedPrice,
-            request.expirationTime
+            request.expirationTime,
+            address(request.currency)
         );
 
         // Callback.
@@ -390,7 +393,15 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
             request.currency.safeTransfer(requester, refund);
         }
 
-        emit DisputePrice(requester, request.proposer, disputer, identifier, timestamp, ancillaryData);
+        emit DisputePrice(
+            requester,
+            request.proposer,
+            disputer,
+            identifier,
+            timestamp,
+            ancillaryData,
+            request.proposedPrice
+        );
 
         // Callback.
         if (address(requester).isContract())
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
@@ -66,19 +66,11 @@ class OptimisticOracleClient {
     });
   }
 
-  // Returns disputes that can be settled and that involved the caler as the disputer
+  // Returns disputes that can be settled and that involved the caller as the disputer
   getSettleableDisputes(caller) {
     return this.settleableDisputes.filter(event => {
       return event.disputer === caller;
     });
-  }
-
-  // Returns an array of Price Request objects for each position that someone has proposed a price for and whose
-  // proposal can be disputed. The proposal can be disputed because the proposed price deviates from the `inputPrice` by
-  // more than the `errorThreshold`.
-  getDisputablePriceProposals() {
-    // TODO
-    return;
   }
 
   // Returns the last update timestamp.
@@ -140,7 +132,8 @@ class OptimisticOracleClient {
           ancillaryData: event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x",
           timestamp: event.returnValues.timestamp,
           proposedPrice: event.returnValues.proposedPrice,
-          expirationTimestamp: event.returnValues.expirationTimestamp
+          expirationTimestamp: event.returnValues.expirationTimestamp,
+          currency: event.returnValues.currency
         };
       });
 
@@ -210,10 +203,7 @@ class OptimisticOracleClient {
     ).filter(event => event !== undefined);
     this.settleableDisputes = unsettledResolvedDisputeEvents;
 
-    // Determine which undisputed proposals SHOULD be disputed based on current prices:
-    // TODO: This would involve having a pricefeed for each identifier. This logic might be
-    // better placed in the OO Keeper instead of the client.
-
+    // Update timestamp and end update.
     this.lastUpdateTimestamp = currentTime;
     this.logger.debug({
       at: "OptimisticOracleClient",
--- a/packages/optimistic-oracle/src/proposer.js
+++ b/packages/optimistic-oracle/src/proposer.js
@@ -214,8 +214,7 @@ class OptimisticOracleProposer {
         continue;
       }
 
-      // If proposal price is not equal to the dispute price, then
-      // prepare dispute.
+      // If proposal price is not equal to the dispute price, then prepare dispute
       // TODO: Implement dispute buffer feature to only dispute if proposal and dispute price
       // differ using some margin of error. For example, we could define a variable `proposalErrorPrecision`
       // and set it to 5, to only send disputes if prices differ after rounding to 5 decimals of precision.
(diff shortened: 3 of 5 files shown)


packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -93,15 +93,17 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         uint256 timestamp,
         bytes ancillaryData,
         int256 proposedPrice,
-        uint256 expirationTimestamp
+        uint256 expirationTimestamp,
+        address currency
```
nicholaspai, line 97: This event parameter allows the bot to directly approve any collateral currencies from `ProposePrice` events, which would be used by the bot if for some reason the collateral currency from a proposal was not already approved when it was pre-proposal.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -93,15 +93,17 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
...
     );
     event DisputePrice(
         address indexed requester,
         address indexed proposer,
         address indexed disputer,
         bytes32 identifier,
         uint256 timestamp,
-        bytes ancillaryData
+        bytes ancillaryData,
+        int256 proposedPrice
```
nicholaspai, line 106: This event parameter allows the bot to more easily report the proposedPrice for a dispute transaction, which is useful for the bot owner to know.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -66,21 +66,13 @@ class OptimisticOracleClient {
...
   getSettleableDisputes(caller) {
     return this.settleableDisputes.filter(event => {
       return event.disputer === caller;
     });
   }
 
-  // Returns an array of Price Request objects for each position that someone has proposed a price for and whose
-  // proposal can be disputed. The proposal can be disputed because the proposed price deviates from the `inputPrice` by
-  // more than the `errorThreshold`.
-  getDisputablePriceProposals() {
```
nicholaspai, line 79: Now handled in `proposer.js` code

packages/optimistic-oracle/src/proposer.js
```diff
@@ -191,29 +309,45 @@ class OptimisticOracleProposer {
...
-          currency: collateralToken.options.address,
-          collateralApprovalTx: collateralApprovalTx.transactionHash
-        });
+        approvalPromises.push(collateralApprovalPromise);
       }
+    };
+
+    // The OO requires approval to transfer the proposed price request's collateral currency in order to post a bond.
+    // We'll set this once to the max value and top up whenever the bot's allowance drops below MAX_INT / 2.
+    for (let priceRequest of this.optimisticOracleClient.getUnproposedPriceRequests()) {
```
nicholaspai, line 341: This is where [this new contract event](https://github.com/UMAprotocol/protocol/pull/2546/files#r577015775) is used.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Just playing devil's advocate: could we just approve if we decide to dispute and look up the full request in that case? Or does that cause other problems? I only say that because each of these adds about 200 gas to each log message.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Similar to the above, aren't disputes pretty rare? Would it be too painful to just directly query the request in question if there's a dispute?

mrice32 (commented):
Just wanted to give quick feedback on the contract changes: while they don't seem like they would cause any security issues, just curious _how much_ extra trouble not having them creates for the bot? Both seem like a lookup would only be required in the case of a dispute, which (at least for now) we expect to be fairly rare, so the extra lookup wouldn't be too bad, at least on our request count. But would be curious to hear how much worse they would be for our bot logic.

nicholaspai:
> Just wanted to give quick feedback on the contract changes: while they don't seem like they would cause any security issues, just curious _how much_ extra trouble not having them creates for the bot? Both seem like a lookup would only be required in the case of a dispute, which (at least for now) we expect to be fairly rare, so the extra lookup wouldn't be too bad, at least on our request count. But would be curious to hear how much worse they would be for our bot logic.

We've already made changes to the `OptimisticOracle.sol` contract bytecode since the OZ audit, see [this PR](https://github.com/UMAprotocol/protocol/pull/2395) for example. I don't see how this PR adds additional uncertainty. WDYT?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
chrismaree: also if we moved over to a DSProxy model we would be able to do the approval in the same time as the dispute logic and would not need do any pre-approval logic.

However, not really any downside (other than gas) to adding thins IMHO.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -175,8 +175,126 @@ class OptimisticOracleProposer {
...
+          at: "OptimisticOracleProposer#sendDisputes",
+          message: "Failed to query historical price for price request",
+          disputePrice,
+          error
+        });
+        continue;
+      }
+
+      // If proposal price is not equal to the dispute price, then
+      // prepare dispute.
```
chrismaree, line 218:
```suggestion
      // If proposal price is not equal to the dispute price, then prepare dispute
```

packages/optimistic-oracle/src/proposer.js
```diff
@@ -175,8 +175,126 @@ class OptimisticOracleProposer {
...
+      // TODO: Implement dispute buffer feature to only dispute if proposal and dispute price
```
chrismaree, line 219: this is a great point and something we need to consider quite closely when it comes to funding rates. ideally a funding rate within a small margin of error cant be disputed as it'll become basically impossible to create a third party implementation that does not constantly get disputed due to small variants. There needs to be some window of tolerance for marginally wrong proposals that _dont_ get disputed. WDYT?

packages/optimistic-oracle/src/proposer.js
```diff
@@ -175,8 +175,126 @@ class OptimisticOracleProposer {
...
+          priceRequest,
+          proposalPrice,
+          disputePrice,
+          txnConfig
+        });
+
+        // Send the transaction or report failure.
+        let receipt;
+        try {
+          receipt = await dispute.send(txnConfig);
```
chrismaree, line 269: do you not want to use `ynatm`?

chrismaree (approved):
LGTM! nice and clean code, well commented and tested. good work

packages/optimistic-oracle/src/proposer.js
nicholaspai: I'll open up a subsequent PR on this I have one idea where we basically check precision up to X decimals of precision. For example, 1.2345 and 1.23456 are "equal" up to 4 decimals of precision

packages/optimistic-oracle/src/proposer.js
nicholaspai: I think we should deprecate using `ynatm` since we don't use it in the `disputer`, and its easier to not have to maintain an external repo

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: The reason why I like blanket approving all collateral during the `update` method is so that we can better isolate failures during the `sendDisputes` and `sendProposals` method to the actual propose/dispute logic. If for some reason the approvals fail, then we'd want these failures to happen as early as possible, not right before a dispute needs to be sent.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: Disputes are rare, but this is also why I see little harm adding a param to this event

mrice32: That's fair, the gas impact will probably be low.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: My very weak opinion is that it doesn't seem worth the extra 400 gas (200 for the read and 200 for the extra log field) for a bot optimization -- however, I'm cool either way. LGTM!

mrice32 (approved):
LGTM!

PR #2547, opened 2021-02-16
refactor(packages): Deprecate CLI

nicholaspai:

Removes unsupported and out of date CLI package.


daywiss (approved):
good bye cli.

as a side note, was any of the logic there worth saving, things like getting all markets or expired markets, etc?

nicholaspai:
> good bye cli.
> 
> as a side note, was any of the logic there worth saving, things like getting all markets or expired markets, etc?

Not really, nothing related to financial contracts that's unique from EMP tools

PR #2548, opened 2021-02-16
refactor(packages): Deprecate Reporters

nicholaspai:

Remove unused daily reporter package


packages/serverless-orchestration/package.json
```diff
@@ -8,6 +8,9 @@
     "@uma/financial-templates-lib": "^1.2.0"
   },
   "devDependencies": {
+    "@google-cloud/datastore": "^6.0.0",
```
nicholaspai, line 11: Not having these packages previously was a false positive because they were already included by the reporter

mrice32 (approved):
LGTM, but would be interested to hear what @hlambur thinks.

packages/serverless-orchestration/package.json
chrismaree: good catch.

chrismaree (approved):
RIP reporter 💔

nicholaspai:
> LGTM, but would be interested to hear what @hlambur thinks.

We discussed offline and he doesn't use this feature anymore nor plans to

PR #2545, opened 2021-02-16
feat(trader): uniswap trading broker

chrismaree:
**Motivation**

With the introduction of the perp, we require a bot that can trade against tokens that are misspriced in the market. This is vital in the early days of launching the perp until market forces coupled with the funding rate are sufficient to keep the perp trading inline.

**Summary**

This PR adds a simple "broker" that can be used to move a uniswap market to a target "true" price. Implementation consists of a broker contract that leverages existing uniswap libraries and implementation to preform the associated trades.

**Details**

This PR adds a smart contract that implements one main function `swapToPrice` which moves a uniswap market from it's current price to a "true" price, paramaterized in the function arguments.
Numerically, This function works using the following logic & arithmetic (screenshot taken of a google doc to enable mathematical typesetting):

image

This `swapToPrice` method implements the equation arrived at the end of the derivation above. 

For more info on this derivation please see [this](https://arxiv.org/abs/1911.03380) paper on useful uniswap math & [this](https://github.com/Uniswap/uniswap-v2-periphery/blob/master/contracts/examples/ExampleSwapToPrice.sol) reference repo, which inspired much of the design presented in this PR. 

**Testing**

Check a box to describe how you tested these changes and list the steps for reviewers to test.

- [ ]  Ran end-to-end test, running the code as in production
- [X]  New unit tests created
- [ ]  Existing tests adequate, no new tests required
- [ ]  All existing tests pass
- [ ]  Untested

**Issue(s)**

partly solves #2560

PR #2559, opened 2021-02-17
feat(bot-infrastructure): add a generic DSProxy client

chrismaree:
**Motivation**

As part of an effort to import our bot infrastructure, we requireDSProxy as part of our bot infra tx processing infra. DSProxy enables bots to execute complex, atomic operations within one transaction. For example, swapping a reserver to a collateral currency, minting synthetics and preforming a liquidation all within 1 tx. This also enables the bots to have more sophisticated inventory management wherein multiple bots can re-use one DSProxy wallet, with multiple EOAs accessing the smart contract wallet. Lastly, DSProxy solves the nonce collision issue that can occur in our current bot design.

**Summary**

This PR adds a generic DSProxy "Manager" to the `financial-templates-lib` that can be re-used by all the bots down the road when we build out a transaction manager. For now, this library will be used by the trader bot in a follow on PR.

**Details**

The goal of this Manager is to abstract away the complexity of dealign with DSProxy, while providing a simple interface to work with. In particular this interface behaves similarly to other `financial-template-lib` clients by providing turn key methods for common logic. The main behaviour this client provides is as follows:

- `getDSProxyFactoryAddress`: returns the address of the managers DSProxy factory. 
- `getDSProxyAddress`: returns the address of the Managers DSPoxy. This is the DSProxy owned by the initialised EOA.
- `initializeDSProxy`: Checks on-chain if the EOA already has a DSProxy. If it does, load it in. If it does not then create one.
- `callFunctionOnExistingLibrary`: lets the user execute `callData` on a already deployed `libraryAddress` from the context of the DSProxy.
- `callFunctionOnNewlyDeployedLibrary`: lets the user execute `callData` on a freshly deployed library, defined in `callCode`. When using this method the DSProxy will first deploy the library then call the method on it, within one tx. 

All methods use our current `logger` pattern to make this as interoperable as possible with our current bots and design pattern.

Note that this implementation is not yet part of the trading bot as it is placed within the `financial-templates-lib`. A subsequent PR will integrate this into the bots, along with the uniswap broker defined in #2545.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

partly solves #2560


packages/core/contracts/common/implementation/DSProxyFactory.sol
```diff
@@ -0,0 +1,203 @@
+// This file was originally take from from Maker: DS Proxy Factory and modified. The original source code can be found
+// here: https://etherscan.io/address/0xa26e15c895efc0616177b7c1e7270a4c7d51c997#code
+// Changes are limited to updating the Solidity version and some stylistic modifications.
```
nicholaspai, line 3: To make reviewing easier can you comment where you made changes from the copied and battle-tested code?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+   * @notice Constructs new Liquidator bot.
+   * @param {Object} logger Module used to send logs.
+   * @param {Object} web3 Web3 object to submit transactions and process on-chain info.
+   * @param {Object} gasEstimator Module used to estimate optimal gas price with which to send txns.
+   * @param {String} account Ethereum account from which to send txns.
+   * @param {String} dsProxyFactoryAddress address of the DSProxy factory to create new DSProxies and the like.
+   */
+  constructor({ logger, web3, gasEstimator, account, dsProxyFactoryAddress, dsProxyFactoryAbi, dsProxyAbi }) {
+    assert(web3.utils.isAddress(account), "Account needs to be a valid address");
+    assert(web3.utils.isAddress(dsProxyFactoryAddress), "dsProxyFactoryAddress needs to be a valid contract address");
```
nicholaspai, line 14: Will the `eth.Contract(abi, address)` not  fail if `dsProxyFactoryAddress` is not a valid address?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+    // Helper functions from web3.
+    this.BN = this.web3.utils.BN;
+    this.toBN = this.web3.utils.toBN;
+    this.toWei = this.web3.utils.toWei;
+    this.fromWei = this.web3.utils.fromWei;
+    this.utf8ToHex = this.web3.utils.utf8ToHex;
+    this.isAddress = this.web3.utils.isAddress;
+
+    // Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+    this.GAS_LIMIT_BUFFER = 1.25;
```
nicholaspai, line 34: You should probably set this in a default config object like the other bots, but this is a nit and can be handled in a subsequent PR

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    // The user does not yet have a DSProxy. Create them one.
+    if (events.length == 0) {
+      this.logger.debug({
+        at: "DSProxyManager",
+        message: "No DSProxy found for EOA. Deploying new DSProxy",
+        account: this.account
+      });
+      const dsProxyCreateTx = await this.dsProxyFactory.methods.build().send({
```
nicholaspai, line 79: Thoughts on simulating the .send() via .estimateGas() in order to catch failed txns?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+      this.logger.info({
+        at: "DSProxyManager",
+        message: "DSProxy has been deployed for the EOA 🚀",
+        dsProxyAddress: this.dsProxyAddress,
+        tx: dsProxyCreateTx.transactionHash,
+        account: this.account
+      });
+    }
+  }
+  // Encode with `yourTruffleContractInstance.yourMethod(params1,param2).encodeABI()
```
nicholaspai, line 94: Good comment explaining context for this method

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+    assert(this.isAddress(libraryAddress), "Library address must be valid address");
+    assert(typeof callData === "string", "Call data must be a string");
+    this.logger.debug({
+      at: "DSProxyManager",
+      message: "Executing function on deployed library",
+      libraryAddress,
+      callData
+    });
+
+    const executeTransaction = await this.dsProxy.methods["execute(address,bytes)"](libraryAddress, callData).send({
```
nicholaspai, line 105: How do you think we should handle reverts?

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,163 @@
...
+
+const { getTruffleContract } = require("@uma/core");
+
+const winston = require("winston");
+const sinon = require("sinon");
+
+// Script to test
+const { DSProxyManager } = require("../../src/proxy-wallet-transaction-handler/DSProxyManager.js");
+
+const TokenSender = getTruffleContract("TokenSender", web3);
```
nicholaspai, line 13: I think you should explicitly set "latest" for readability

packages/trader/src/index.ts
```diff
@@ -5,7 +5,7 @@ import { TraderConfig } from "./TraderConfig";
 
 export async function run(): Promise<void> {
   // Config Processing
-  const config = new TraderConfig(process.env);
+  // const config = new TraderConfig(process.env);
```
nicholaspai, line 8: Leftover comment?

nicholaspai (approved):
This is an amazing feature I'm excited to integrate into our repo. There's generally little risk to add this to the repo so I'm curious what the future features building on top of this will bring.

packages/core/contracts/common/implementation/DSProxyFactory.sol
```diff
@@ -0,0 +1,203 @@
+// This file was originally take from from Maker: DS Proxy Factory and modified. The original source code can be found
```
mrice32, line 1: Similar to uniswap, are there packages exposed that include the ABIs and bytecode for this factory?

packages/core/contracts/common/implementation/DSProxyFactory.sol
mrice32: +1

packages/core/contracts/common/test/TokenSender.sol
```diff
@@ -0,0 +1,17 @@
...
+
+contract TokenSender {
+    function transferERC20(
+        address tokenAddress,
+        address recipientAddress,
+        uint256 amount
+    ) public returns (bool) {
+        IERC20 token = IERC20(tokenAddress);
+        token.transfer(recipientAddress, amount);
+        return true;
```
mrice32, line 15: Just curious, why return true?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+  getDSProxyFactoryAddress() {
+    return this.dsProxyFactoryAddress;
+  }
+
+  getDSProxyAddress() {
+    return this.dsProxyAddress;
+  }
+
+  // Load in an existing DSProxy for the account EOA if one already exists OR create a new one for the user.
+  async initializeDSProxy() {
```
mrice32, line 46: Should this return early if this.dsProxyAddress and this.dsProxy are already set?

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,163 @@
...
+        gasEstimator,
+        account: "not an address",
+        dsProxyFactoryAddress: dsProxyFactory.address,
+        dsProxyFactoryAbi: DSProxyFactory.abi,
+        dsProxyAbi: DSProxy.abi
+      });
+    } catch (error) {
+      didThrow = true;
+    }
+    assert.isTrue(didThrow);
```
mrice32, line 73: nit: can this be refactored to:
```js
assert.throws(() => {
  new DSProxyManager({
      logger: spyLogger,
      web3,
      gasEstimator,
      account: "not an address",
      dsProxyFactoryAddress: dsProxyFactory.address,
      dsProxyFactoryAbi: DSProxyFactory.abi,
      dsProxyAbi: DSProxy.abi
    });
});
```

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,163 @@
...
+    await dsProxyManager.initializeDSProxy();
+
+    dsProxyFactory = await DSProxyFactory.at(dsProxyManager.getDSProxyFactoryAddress());
+
+    // Seeing the EOA had no DSProxy before, it should have deployed a new DSProxy with the contractCreator set as owner.
+    const logs = await dsProxyFactory.getPastEvents("Created", { fromBlock: 0 });
+
+    assert.equal(logs.length, 1);
+    assert.equal(logs[0].returnValues.owner, contractCreator);
+    assert.equal(await (await DSProxy.at(logs[0].returnValues.proxy)).owner(), contractCreator);
```
mrice32, line 86: ❤️ double await

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+    });
+
+    const executeTransaction = await this.dsProxy.methods["execute(bytes,bytes)"](callCode, callData).send({
+      from: this.account,
+      gasPrice: this.gasEstimator.getCurrentFastPrice()
+    });
+
+    this.logger.info({
+      at: "DSProxyManager",
+      message: "Executed function on a freshly minted library, created in the same tx 🌝",
```
mrice32, line 137:
```suggestion
      message: "Executed function on a freshly minted library, created in the same tx 🤗",
```

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,163 @@
...
+    // any tokens that it might have within it's wallet. It is used to showcase how a more complex transaction would occur
+    // with the DSProxy acting in place of an EOA.
+    tokenSender = await TokenSender.new();
+
+    // Encode the send transaction. This is what will be executed within the DSProxy.
+    const sendTokenTx = tokenSender.contract.methods
+      .transferERC20(testToken.address, contractCreator, toWei("10"))
+      .encodeABI();
+
+    // Call the method on the tokenSender(library) with the sendTokenTxData. Node this is called from the perspective
```
mrice32, line 117:
```suggestion
    // Call the method on the tokenSender(library) with the sendTokenTxData. Note this is called from the perspective
```

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,148 @@
...
+      at: "DSProxyManager",
+      message: "Executed function on deployed library 📸",
+      libraryAddress,
+      callData,
+      tx: executeTransaction.transactionHash
+    });
+    return executeTransaction;
+  }
+  // Extract call code using the `.abi` syntax on a truffle object or the `getABI(contractType,contractVersion)` from common.
+  async callFunctionOnNewlyDeployedLibrary(callCode, callData) {
```
mrice32, line 120: Is there a simple way (without a bunch of extra methods or code) to allow callers to simulate these transactions and grab the return value? aka just a .call rather than a .send()?

mrice32 (commented):
Looks awesome!!

packages/core/contracts/common/implementation/DSProxyFactory.sol
chrismaree: the only changes were updating the solidity version & the associated syntactical changes that changed between versions (emit on events, `memory` on bytes params) and linting the file to fit to our style guide.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: ye, properly felt that this was more explicit though.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: ye, could do. let's leave that for a follow up PR though.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: I think we can do a `.send` and then pass that back to the logger and the caller of the client.

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
chrismaree: sure & done.

packages/trader/src/index.ts
chrismaree: this was failing the linter.

packages/core/contracts/common/implementation/DSProxyFactory.sol
chrismaree: not that I can find but that is a good point! I don think this is too bloated though, wdyt?

packages/core/contracts/common/test/TokenSender.sol
chrismaree: returned true to validate that the returned params from the libraries can be seen from the DSProxy that use the libraries :)

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: ye, good point. I also changed it to return `this.dsProxyAddress`.

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
chrismaree: spending too long playing with node haha

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: ye, should be pretty easy to do I guess. we could make a separate method to do it and then just return that value there. WDYT?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,150 @@
...
+    // Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+    this.GAS_LIMIT_BUFFER = 1.25;
+  }
+
+  getDSProxyFactoryAddress() {
+    return this.dsProxyFactoryAddress;
+  }
+
+  getDSProxyAddress() {
+    return this.dsProxyAddress;
```
daywiss, line 42: imo these should throw if uninitialized, wdyt?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
daywiss: is this idea that this function needs to be called right after construction? sort of like price feeds `update`

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,150 @@
...
+
+    if (this.dsProxy && this.dsProxyAddress) this.dsProxyAddress;
+    const fromBlock = await getFromBlock(this.web3);
+    const events = await this.dsProxyFactory.getPastEvents("Created", {
+      fromBlock,
+      filter: { owner: this.account }
+    });
+
+    // The user already has a DSProxy deployed. Load it in from the events.
+    if (events.length > 0) {
```
daywiss, line 61: i feel like this function should be split up into multiple functions which handle parts of the logic
async hasProxy() => checks cache / chain to see if user has already deployed this
async getProxy() => loads cached proxy or pulls it from chain to cache. throws if does not exist.
async createProxy() => only creates if proxy does not exist, throws if it does

i think this kind of makes it easier to reason about and easier to unit test and these would be composed to create your init function. wdyt?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,150 @@
...
+    assert(this.isAddress(libraryAddress), "Library address must be valid address");
+    assert(typeof callData === "string", "Call data must be a string");
+    this.logger.debug({
+      at: "DSProxyManager",
+      message: "Executing function on deployed library",
+      libraryAddress,
+      callData
+    });
+
+    const executeTransaction = await this.dsProxy.methods["execute(address,bytes)"](libraryAddress, callData).send({
```
daywiss, line 107: i feel like since this is relying on initializing the class, this should be calling a getter which throws if `this.dsProxy` does not exist.  same with other references to `this.dsProxy`. thoughts?

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,159 @@
...
+    tokenSender = await TokenSender.new();
+
+    // Encode the send transaction. This is what will be executed within the DSProxy.
+    const sendTokenTx = tokenSender.contract.methods
+      .transferERC20(testToken.address, contractCreator, toWei("10"))
+      .encodeABI();
+
+    // Call the method on the tokenSender(library) with the sendTokenTxData. Note this is called from the perspective
+    // of the DSProxy and should send some of the DSProxy tokens.
+    const dsProxyCallReturn = await dsProxyManager.callFunctionOnExistingLibrary(tokenSender.address, sendTokenTx);
```
daywiss, line 115: this is really cool to see.  i dont really fully understand how this is possible. The ds proxy does not need to know anything about the token library other than address? I guess the sendTokenTx has all the info it needs to execute a call on the contract?

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,159 @@
...
+    await testToken.mint(dsProxyAddress, toWei("1000"), { from: contractCreator });
+
+    // As before we encode the tx to send to the library.
+    const contract = new web3.eth.Contract(TokenSender.abi);
+    const callData = contract.methods.transferERC20(testToken.address, contractCreator, toWei("10")).encodeABI();
+
+    // The library also needs to code of the contract to deploy.
+    const callCode = TokenSender.bytecode;
+
+    const dsProxyCallReturn = await dsProxyManager.callFunctionOnNewlyDeployedLibrary(callCode, callData);
```
daywiss, line 145: can you explain what this call is doing. what are the on chain side effects of this?

daywiss (commented):
very cool to see this pattern

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: sure. done.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: correct. this is needed as you cant have any async calls within a constructor in JS.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -0,0 +1,150 @@
...
+
+  // Load in an existing DSProxy for the account EOA if one already exists OR create a new one for the user.
+  async initializeDSProxy() {
+    this.logger.debug({
+      at: "DSProxyManager",
+      message: "Initalizing...Looking for existing DSProxies or deploying a new one for the provided EOA",
+      dsProxyFactoryAddress: this.dsProxyFactoryAddress
+    });
+
+    if (this.dsProxy && this.dsProxyAddress) this.dsProxyAddress;
```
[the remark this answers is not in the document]
chrismaree: yup. have fixed, thanks!

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: ye we could do that. however I feel that seeing there is only two branching states this is probs fine to leave like this for now. I can add this refactor if you feel strongly about it but I dont think it's really warranted at this point. 

Also, I tried to implement this change but it ended up taking more code and had to repeat the log query method against the DSProxy factory. For now I think it cleaner to just leave it like this. In places where initialisation must happen first I added an assert to check that an address is set. I think this amounts to the same behaviour but with fewer async calls on chain and less code.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: I added an assert to check that the DSProxy is set.

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
chrismaree: yup! the dsProxy just needs to know where the code is deployed (the `tokenSender.address`) and then it can execute the transaction  `sendTokenTx` on the library. Take a look at the DSProxy smart contract code and see the `execute` method. This function either runs a function on a existing library (in the case of this call) or if overloaded runs on a contract that is first deployed. 

The key part of the `execute` method to look at is the assembly that uses the `delegatecall` opcode. this runs the provided transaction data (in this case `sendTokenTx`) in the context of the contract (the DSProxy) and executes the code at the location specified in the address.

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
chrismaree: This one is similar to the above function except first the DSProxy will deploy the `callCode` byte code and then execute `callData` on the call code. This differs from `dsProxyManager.callFunctionOnExistingLibrary` which uses an already deployed byte code that is already on-chain. 

The steps are something like this:
1) DSProxy checks if the `callCode` is already deployed  and is in the cache. if it is, load it in. else, deploy the contract
2) execute call data on `callCode`. in this tx the "context" is the DSProxy. this is the key detail when working with proxies; the `msg.sender` is no longer the EOA that called the proxy but is now the proxy itself
3) the transaction `callData` is executed on the `callCode` contract with `msg.sender` now coming from the DSProxy due to the use of `delegatecall`.

Point 3, and the switching of `msg.sender` is the reason that DSProxy is so powerful: the contract can "act" as an EOA, while having all the power of a smart contract. the `callCode` could be as simple as sending an erc20 token (like in this example) or far far more complex like a) swapping from reserve to collateral currency, b) minting a synthetic and c) liquidating a position all in one tx.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
daywiss: ok yea sounds good. not super opinionated on it, just wanted to get your take

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
daywiss: very interesting. I think i undestand, thanks for explanation. So just to be clear, the callCode gets persisted on chain after this call, its not like it gets cleaned up or deleted at the end of the call?

daywiss (approved):

packages/financial-templates-lib/test/proxy-transaction-handler/DSProxyManager.js
chrismaree: correct. the cache can re-use it later on :)

PR #2561, opened 2021-02-17
improve(core): update voting upgrade script to pass proposer wallet as migrated address

mrice32:
**Motivation**

To allow the foundation hot wallet to claim rewards on behalf of voters in the old voting contract, we need to set it as the migrated address.

**Summary**

Added a parameter to the upgrader contract and passed the foundation hot wallet to the constructor in the upgrade script.

Note: the migrated address has the following abilities:
- Migrated address can request a price to "start a vote"
- Nobody can cast votes (so any incomplete votes cannot be finalized and, therefore generate rewards).
- Migrated address can call claim rewards.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/umips/VotingUpgrader.sol
+++ b/packages/core/contracts/umips/VotingUpgrader.sol
@@ -26,23 +26,31 @@ contract VotingUpgrader {
     // Addresses to upgrade.
     address public newVoting;
 
+    // Address to call setMigrated on the old voting contract.
+    address public setMigratedAddress;
+
     /**
      * @notice Removes an address from the whitelist.
      * @param _governor the Governor contract address.
      * @param _existingVoting the current/existing Voting contract address.
      * @param _newVoting the new Voting deployment address.
      * @param _finder the Finder contract address.
+     * @param _setMigratedAddress the address to set migrated. This address will be able to continue making calls to
+     *                            old voting contract (used to claim rewards on others' behalf). Note: this address
+     *                            can always be changed by the voters.
      */
     constructor(
         address _governor,
         address _existingVoting,
         address _newVoting,
-        address _finder
+        address _finder,
+        address _setMigratedAddress
     ) public {
         governor = _governor;
         existingVoting = Voting(_existingVoting);
         newVoting = _newVoting;
         finder = Finder(_finder);
+        setMigratedAddress = _setMigratedAddress;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


packages/core/scripts/voting-upgrade-umip/3_Verify.js
```diff
@@ -37,7 +38,7 @@ async function runExport() {
 
   const newVotingContract = await Voting.at(votingAddress);
   const currentOwner = await newVotingContract.owner();
-  assert.equal(web3.utils.toChecksumAddress(currentOwner), web3.utils.toChecksumAddress(Governor.address));
+  assert.equal(web3.utils.toChecksumAddress(currentOwner), web3.utils.toChecksumAddress(governor.address));
```
chrismaree, line 41: curious question: why was this change required? if you are running this with truffle can you not pull the address directly from the truffle artefact?

packages/core/contracts/umips/VotingUpgrader.sol
```diff
@@ -26,23 +26,31 @@ contract VotingUpgrader {
...
+     *                            old voting contract (used to claim rewards on others' behalf). Note: this address
+     *                            can always be changed by the voters.
      */
     constructor(
         address _governor,
         address _existingVoting,
         address _newVoting,
-        address _finder
+        address _finder,
+        address _setMigratedAddress
```
chrismaree, line 47: you could just use `msg.sender` for this, no? I guess it's more generalised in how you've done it now and in future we might want to set the migrated address to the new voting contract so not really any downside of doing it this way.

chrismaree (approved):

packages/core/scripts/voting-upgrade-umip/3_Verify.js
mrice32: I think this broke when we moved over to using our own `getTruffleArtifact` method. `.address` is seeded by some async method being called on the instance so it's able to grab the network id from web3. Truffle does this for it's artifacts on startup, but since this hadn't been done yet for the artifact that we created, the instance doesn't know what address to grab.

packages/core/contracts/umips/VotingUpgrader.sol
mrice32: Good point. We could. But yeah, I thought it might be better to keep it explicit and agnostic of the deployer.

nicholaspai (approved):
LGTM

PR #2563, opened 2021-02-17
refactor(uniswap-pricefeed): Remove poolDecimals param

### LINKED ISSUE #2557: UniswapPriceFeed: Detect quote currency precision from `token0` or `token`
**Why**

Currently, `UniswapPriceFeed` will only work if base currency (reserve0 if not inverted) uses 18 decimal precision. Additionally, this pricefeed forces user to specify `poolDecimals` which is really the quote currency precision (reserve1 if not inverted).

We should remove the `poolDecimals` constructor argument and read this automatically from the contract.

**How**

Will have to add `token0` and `token1` to `UniswapMock.sol` to test properly. Need to update tests as well.

**Difficulty**

- 4: ~2 hours

nicholaspai:

**Motivation**

Fixes #2557

**Summary**

`poolDecimals` is replaced by internal class variable `this.baseCurrencyDecimals` which is autodetected from `Uniswap` contract. `Uniswap.sol` was updated with `token0` and `token1` public methods.

Follow up, but less urgent PR, makes same change to `BalancerPriceFeed`. No identifiers use this pricefeed currently, thus less urgent.

**Testing**

- [X]  Ran end-to-end test, running the code as in production. Tested on uGAS and X-STABLESPREAD identifiers which use `UniswapPriceFeed` in production.
- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2577

### DIFF
--- a/packages/core/contracts/common/interfaces/Uniswap.sol
+++ b/packages/core/contracts/common/interfaces/Uniswap.sol
@@ -8,4 +8,8 @@ pragma solidity ^0.6.0;
 abstract contract Uniswap {
     // Called after every swap showing the new uniswap "price" for this token pair.
     event Sync(uint112 reserve0, uint112 reserve1);
+    // Base currency.
+    address public token0;
+    // Quote currency.
+    address public token1;
 }
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -81,8 +81,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xc50ef7861153c51d383d9a7d48e6c9467fb90c38",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -94,8 +93,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xa0abda1f980e03d7eadb78aed8fc1f2dd0fe83dd",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -154,8 +152,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xc50ef7861153c51d383d9a7d48e6c9467fb90c38",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -167,8 +164,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xa0abda1f980e03d7eadb78aed8fc1f2dd0fe83dd",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -220,8 +216,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xc50ef7861153c51d383d9a7d48e6c9467fb90c38",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -233,8 +228,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xa0abda1f980e03d7eadb78aed8fc1f2dd0fe83dd",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -292,8 +286,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0x97c4adc5d28a86f9470c70dd91dc6cc2f20d2d4d",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
@@ -304,8 +297,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0x88ff79eb2bc5850f27315415da8685282c7610f9",
-            twapLength: 2,
-            poolDecimals: 6
+            twapLength: 2
           }
         ]
       },
--- a/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
@@ -90,31 +90,33 @@ class UniswapPriceFeed extends PriceFeedInterface {
   }
 
   async update() {
-    // Read token0 and token1 precision from Uniswap contract.
-    const [token0Address, token1Address] = await Promise.all([
-      this.uniswap.methods.token0().call(),
-      this.uniswap.methods.token1().call()
-    ]);
-    this.token0 = new this.web3.eth.Contract(this.erc20Abi, token0Address);
-    this.token1 = new this.web3.eth.Contract(this.erc20Abi, token1Address);
-    const [token0Precision, token1Precision] = await Promise.all([
-      this.token0.methods.decimals().call(),
-      this.token1.methods.decimals().call()
-    ]);
-    // The base currency of the Uniswap pool pair, which is `token1` unless
-    // `invertPrice == True` is the precision that the internal method `_getPriceFromSyncEvent()`
-    // returns prices in, and `convertBaseDecimalsToPriceFeedDecimals` will convert from this precision
-    // to the user's desired `priceFeedDecimals`.
-    this.baseCurrencyDecimals = Number(this.invertPrice ? token0Precision : token1Precision);
-    this.quoteCurrencyDecimals = Number(this.invertPrice ? token1Precision : token0Precision);
-    // Now that we've determined the base precision, create a function that converts
-    // _bn precision from baseCurrencyDecimals to desired decimals by scaling up or down based
-    // on the relationship between pool precision and the desired decimals.
-    this.convertBaseDecimalsToPriceFeedDecimals = ConvertDecimals(
-      this.baseCurrencyDecimals,
-      this.priceFeedDecimals,
-      this.web3
-    );
+    // Read token0 and token1 precision from Uniswap contract if not already cached:
+    if (!this.quoteCurrencyDecimals || !this.convertBaseDecimalsToPriceFeedDecimals) {
+      const [token0Address, token1Address] = await Promise.all([
+        this.uniswap.methods.token0().call(),
+        this.uniswap.methods.token1().call()
+      ]);
+      this.token0 = new this.web3.eth.Contract(this.erc20Abi, token0Address);
+      this.token1 = new this.web3.eth.Contract(this.erc20Abi, token1Address);
+      const [token0Precision, token1Precision] = await Promise.all([
+        this.token0.methods.decimals().call(),
+        this.token1.methods.decimals().call()
+      ]);
+      // The base currency of the Uniswap pool pair, which is `token1` unless
+      // `invertPrice == True` is the precision that the internal method `_getPriceFromSyncEvent()`
+      // returns prices in, and `convertBaseDecimalsToPriceFeedDecimals` will convert from this precision
+      // to the user's desired `priceFeedDecimals`.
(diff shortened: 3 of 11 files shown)


packages/serverless-orchestration/test/ServerlessHub.js
```diff
@@ -124,11 +124,10 @@ contract("ServerlessHub.js", function(accounts) {
 
     uniswap = await UniswapMock.new();
 
-    defaultUniswapPricefeedConfig = {
```
nicholaspai, line 127: Removing reliance on UniswapMock because this package uses @uma/core version 1.2 which doesn't have latest changes to UniswapMock.sol

packages/liquidator/test/index.js
```diff
@@ -20,7 +20,6 @@ const Poll = require("../index.js");
 let collateralToken;
 let syntheticToken;
 let financialContract;
-let uniswap;
```
nicholaspai, line 23: Removing reliance on Uniswap mock from these tests beacuse the actual pricefeed type doesn't matter. And now we have the `test` type (i.e. the PriceFeedMock) to use

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -81,8 +81,7 @@ const defaultConfigs = {
           {
             type: "uniswap",
             uniswapAddress: "0xc50ef7861153c51d383d9a7d48e6c9467fb90c38",
-            twapLength: 2,
```
chrismaree, line 84: +1

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -86,14 +82,40 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
   async update() {
+    // Read token0 and token1 precision from Uniswap contract.
+    const [token0Address, token1Address] = await Promise.all([
+      this.uniswap.methods.token0().call(),
+      this.uniswap.methods.token1().call()
+    ]);
+    this.token0 = new this.web3.eth.Contract(this.erc20Abi, token0Address);
+    this.token1 = new this.web3.eth.Contract(this.erc20Abi, token1Address);
+    const [token0Precision, token1Precision] = await Promise.all([
+      this.token0.methods.decimals().call(),
```
chrismaree, line 101: what happens if the token does not have a decimals method?

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
```diff
@@ -418,13 +469,13 @@ contract("UniswapPriceFeed.js", function(accounts) {
       // The historical TWAP for 1 hour ago (the earliest allowed query) should be 100 for the first half and then 90 for the second half -> 95.
       assert.equal(
         (await scaleUpPriceFeed.getHistoricalPrice(currentTime - 3600)).toString(),
-        toBN(toWei("95"))
+        toBN(parseFixed("95", baseCurrencyPrecision).toString())
```
chrismaree, line 472: do you want to create a `convertPrice` or `convertBaseCurrency` method like we do in other places that basically will wrap `parseFixed`?

chrismaree (approved):
This looks great! I just have one question and a nit but otherwise this is a good simplification. good work!

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: I think we can handle this two ways. One ignore it and assume that we can only work with pools that have tokens that implement decimals. I think this is reasonable since any tokens that don't implement decimals might not be that robust. [This is the same assumption that the `liquidator/index.js` makes for example](https://github.com/UMAprotocol/protocol/blob/master/packages/liquidator/index.js#L159).

Or we can catch and assume 18 decimals.

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
nicholaspai: Done

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
mrice32: I think either is probably fine. I'm not sure of many examples of tokens that don't implement decimals, but I would assume that any that don't are probably 18. It's probably a _little_ safer to throw an error in the absence of the method, however.

nicholaspai: Yeah agreed I'd like to know if a token doesn't implement `decimals()`

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -86,14 +82,40 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+    return this.lastBlockPrice && this.convertBaseDecimalsToPriceFeedDecimals(this.lastBlockPrice);
   }
 
   getPriceFeedDecimals() {
     return this.priceFeedDecimals;
   }
 
   async update() {
+    // Read token0 and token1 precision from Uniswap contract.
+    const [token0Address, token1Address] = await Promise.all([
```
mrice32, line 94: Since we don't expect token0 and token1 to change precisions, could we cache the results by doing something like:
```js
if (!this.quoteCurrencyDecimals || !this.convertBaseDecimalsToPriceFeedDecimals) {
  // Logic to set this.quoteCurrencyDecimals and this.convertBaseDecimalsToPriceFeedDecimals
}

// Grab price info from uniswap contract.
```

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -86,14 +82,40 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+    const [token0Precision, token1Precision] = await Promise.all([
+      this.token0.methods.decimals().call(),
+      this.token1.methods.decimals().call()
+    ]);
+    // The base currency of the Uniswap pool pair, which is `token1` unless
+    // `invertPrice == True` is the precision that the internal method `_getPriceFromSyncEvent()`
+    // returns prices in, and `convertBaseDecimalsToPriceFeedDecimals` will convert from this precision
+    // to the user's desired `priceFeedDecimals`.
+    this.baseCurrencyDecimals = Number(this.invertPrice ? token0Precision : token1Precision);
+    this.quoteCurrencyDecimals = Number(this.invertPrice ? token1Precision : token0Precision);
```
mrice32, line 109: Just want to check my understanding.

Non inverted:
Quote Currency == token0
Base Currency == token1
price = token1 / token0
price decimals == token1 decimals (base?)

Inverted:
Quote Currency == token1
Base currency == token0
price decimals == token0 decimals (base?)


Is my understanding correct?

mrice32 (commented):
Looks great!  A few minor questions/comments.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: Yeah good point. I'll implement

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
nicholaspai: Correct but `priceDecimals` in the context of this PriceFeed is equal the precision that the user wants to return prices. So, the `_getPriceFromSyncEvent()` returns prices in the same precision as the base currency (token1 unless inverted), and then that precision is converted into `priceDecimals`.

For example, say the base decimals is 8, but the user wants prices in 10, then the return value of `_getPriceFromSyncEvent()` will get scaled up by 10**2

packages/core/contracts/common/interfaces/Uniswap.sol
```diff
@@ -8,4 +8,8 @@ pragma solidity ^0.6.0;
 abstract contract Uniswap {
     // Called after every swap showing the new uniswap "price" for this token pair.
     event Sync(uint112 reserve0, uint112 reserve1);
+    // Base currency.
+    address public token0;
+    // Quote currency.
+    address public token1;
```
pemulis, line 14: Based on the rest of the code, isn't this the other way around, with `token0` the quote currency and `token1` the quote currency?

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
```diff
@@ -301,43 +312,61 @@ contract("UniswapPriceFeed.js", function(accounts) {
...
+      // The UniswapPriceFeed's _getPriceFromSyncEvent will return prices in same precision as
+      // the base currency. But we also change the quote currency precision to test that the
+      // UniswapPriceFeed correctly handles it.
+      let quoteToken = await Token.new("Uni Token", "UQ", quoteCurrencyPrecision, { from: owner });
+      let baseToken = await Token.new("Uni Token", "UB", baseCurrencyPrecision, { from: owner });
+
+      // Not inverted so quote token is token0
+      await uniswapMock.setTokens(quoteToken.address, baseToken.address);
+
+      // Here we specify that the pool is returning a 6 decimal precision price, and we want prices in
```
pemulis, line 332: Nit: The pool is returning a 6 decimal precision price, but it doesn't look like we actually specify it here anymore; it's pulled from the pool, right? The part where it's specified in the function is deleted. Same nit for `scaleDownPriceFeed`.

packages/liquidator/test/index.js
```diff
@@ -464,7 +455,8 @@ contract("index.js", function(accounts) {
         // To create an error within the liquidator bot we can create a price feed that we know will throw an error.
         // Specifically, creating a uniswap feed with no `sync` events will generate an error. We can then check
         // the execution loop re-tries an appropriate number of times and that the associated logs are generated.
-        uniswap = await UniswapMock.new();
+        const uniswap = await UniswapMock.new();
+        await uniswap.setTokens(collateralToken.address, collateralToken.address);
```
pemulis, line 459: Quote token and base token are the same?

mrice32 (commented):
As discussed IRL, it's probably less confusing to just remove base and quote currency since the interpretation seems to differ based on the reader.

packages/core/contracts/common/interfaces/Uniswap.sol
mrice32: @nicholaspai and I discussed this and we differed in our interpretation of how to translate the base and quote currencies into the uniswap feed. So @nicholaspai suggested that we just remove that terminology from the PR to make it easier to reason about.

packages/liquidator/test/index.js
nicholaspai: I'm gonna remove the base and quote nomenclature but I'll make it more clear that for this specific test it's just arbitrary and I just need to initiate these values

packages/core/contracts/common/interfaces/Uniswap.sol
nicholaspai: I removed all references to "quote" and "base" currency because my usage was ambiguous. I think the standard is that for the currency pair listed as "X-Y", X is the base and Y is the quote.

packages/financial-templates-lib/test/truffle/UniswapPriceFeed.js
nicholaspai: Correct the price is pulled from one of the Uniswap pair's constituent tokens. The `uniswapMock.setTokens()` method sets the tokens so I'll make this more clear.

nicholaspai: Removed all references to `base` and `quote` from the tests and also made it clear why we're expecting the pool to return 6 precision and how we convert that either into 12 or 4

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
chrismaree: right. I agree with your point here nick. I think the key is that the decimals that the feed returns at will be different to that of the pair and should always be scaled by the `priceFeedDecimals`, irrespective of the number of decimals in the pooled tokens.

mrice32 (approved):
LGTM!

PR #2564, opened 2021-02-17
feat(optimistic-oracle-proposer): Only submit disputes where historical prices diverge beyond a margin of allowed error

nicholaspai:
**Motivation**

Part of #2411 

Currently disputes are sent whenever the `optimistic-oracle` bot queries a different historical price from the proposed price, even if the prices are off by 0.00001, for example. This PR adds an optional param that enables a margin of error.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/optimistic-oracle/index.js
```diff
@@ -121,6 +121,8 @@ async function Poll(callback) {
       errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 10,
       // If there is an optimistic oracle config, add it. Else, set to null. Example config:
       // {
+      //   "disputePricePrecisionOfError":4 -> Proposal prices that do not equal the dispute price
```
chrismaree, line 124: the default you have in the `proposer.js` is 6, not 4.
```suggestion
      //   "disputePricePrecisionOfError":6 -> Proposal prices that do not equal the dispute price
```

packages/optimistic-oracle/src/proposer.js
```diff
@@ -48,6 +54,16 @@ class OptimisticOracleProposer {
         isValid: x => {
           return x >= 6000000 && x < 15000000;
         }
+      },
+      disputePricePrecisionOfError: {
```
chrismaree, line 58: this is a good way way deal with this.

I have a question about the best way to generalise this. When we talk about precision we are normally talking about the number of digits after the decimal point. however, in the case of really large numbers, it might be better to consider precision interns of the number of significant figures within the number.

For example, consider the number 12.3456789. If you said you wanted to round this to a precision of 6 would my understanding of your implementation is that it would turn this into 12.3456790, rounding to the 6th decimal place. 

How would you then round the number 123456789 to a precision of 6? based off your implementation I _think_ it would result in not modifying the number. however, there is an equal number of significant figures in this second case and you could argue that it should be rounded equally to 123456000. In this case we look for the number of significant figures and "round" to that. If this pattern was followed with the first example the rounding would have yielded 12.3456000, thereby considering all significant figures in the process, rather than just those after the decimal point. 

The point I'm trying to make is that we should ensure we converge on a consistant definition of precision for small and big numbers alike. To be clear: the proposal I'm making it to consider precision interns of significant figures, rather than decimal points. WDYT?

chrismaree (commented):
this is pretty awesome. I have 1 nit and one comment about how we should deal with precision. curious to hear what you think.

packages/optimistic-oracle/src/proposer.js
nicholaspai: This is a really good point and I think large numbers we'd be interested in rouding to a number left of the decimal point. The easyish fix for this is to replace `toFixed()` with `toPrecision()` in determining the `rounded` values. WDYT?

`toPrecision` sets the # of significant digits, which includes both left and right sides of the decimal

nicholaspai: Alternatively, we could just define the buffer as a percentage rather than a # of decimals of precision. Say we allow prices to differ by 10% before we'd dispute. WDYT?

nicholaspai: I implemented this change as I think its more flexible ^

packages/optimistic-oracle/index.js
```diff
@@ -18,9 +18,19 @@ const { getWeb3 } = require("@uma/common");
  *     mode which will exit after the loop.
  * @param {Number} errorRetries The number of times the execution loop will re-try before throwing if an error occurs.
  * @param {Number} errorRetriesTimeout The amount of milliseconds to wait between re-try iterations on failed loops.
+ * @param {Object} [commonPriceFeedConfig] Common configuration to pass to all PriceFeeds constructed by proposer.
```
nicholaspai, line 21: `commonPriceFeedConfig` was not actually added as a constructor param so I added it, and also renamed it from the less descriptive `defaultPriceFeedConfig`. `common` better explains how its a config passed to all constructed pricefeeds

packages/optimistic-oracle/index.js
```diff
@@ -119,8 +129,14 @@ async function Poll(callback) {
       errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
       // Default to 10 seconds in between error re-tries.
       errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 10,
+      // Common price feed configuration passed along to all those constructed by proposer.
+      commonPriceFeedConfig: process.env.OOPRICE_FEED_CONFIG
```
pemulis, line 133: Why the name difference, `OOPRICE_FEED_CONFIG` vs. `commonPriceFeedConfig`? Should this be `COMMON_PRICE_FEED_CONFIG`?

packages/optimistic-oracle/index.js
```diff
@@ -119,8 +129,14 @@ async function Poll(callback) {
...
+        ? JSON.parse(process.env.OOPRICE_FEED_CONFIG)
+        : { lookback: 7200 },
       // If there is an optimistic oracle config, add it. Else, set to null. Example config:
       // {
+      //   "disputePriceErrorPercent":0.05 -> Proposal prices that do not equal the dispute price
```
pemulis, line 138: Does this mean 0.05% or 5%? Looks like 5% from reading the code, but the comment could be more explicit.

packages/optimistic-oracle/test/proposer.js
```diff
@@ -441,6 +458,33 @@ contract("OptimisticOracle: proposer.js", function(accounts) {
...
+    } catch (err) {
+      assert(true);
+    }
+    try {
+      new OptimisticOracleProposer({
+        logger: spyLogger,
+        optimisticOracleClient: client,
+        gasEstimator,
+        account: botRunner,
+        ooProposerConfig: { disputePriceErrorPercent: 1 }
```
pemulis, line 480: I can see why, practically speaking, you would never want a > 100% margin of error, but is there a technical reason for this as well?

nicholaspai: I'm gonna remove the base and quote nomenclature but I'll make it more clear that for this specific test it's just arbitrary and I just need to initiate these values

packages/optimistic-oracle/index.js
nicholaspai: Yeah agreed, and changed.

packages/optimistic-oracle/index.js
nicholaspai: 5%, clarified in the code.

packages/optimistic-oracle/test/proposer.js
nicholaspai: I suppose there is technically no limit....do you think we should allow it? Curious what @mrice32 @daywiss think too

packages/optimistic-oracle/src/proposer.js
chrismaree: Good changes! I'm happy with how you modified this to be a percentage; this way it's not about decimal points/precision but rather an absolute error relative to the "true" value. I'm more happy with this! thank for changing it.

chrismaree (approved):
Thanks for changing this to be percentage based, much easier to reason about even though the code is a bit more complex.

LGTM!

nicholaspai:
> Thanks for changing this to be percentage based, much easier to reason about even though the code is a bit more complex.
> 
> LGTM!

I think the code is actually less complex, but I was just thinking it might be nice to round to a certain precision like we specify in the UMIPs.  Look at [uStonks](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md#technical-specifications) technical specification for example.

But I think its ok to just set a % buffer for the disputer

PR #2568, opened 2021-02-18
fix(liquidator): Remove unneccessary event query

nicholaspai:

**Motivation**

Removing code that never should have been committed from #2530 and that fails when a liquidation is detected and both START_BLOCK and END_BLOCK are set (such as on GCP)

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


chrismaree (approved):
hhhmm. I see the issue. can we double check there is no other issues like this left? I'll re-read through all this logic my side as well to try and catch any potential issue.

PR #2569, opened 2021-02-18
feat(optimistic-oracle-proposer): Can settle proposals and disputes

### LINKED ISSUE #2411: Build stub for Optimistic Oracle Keeper which follows strategy to respond to price requests

nicholaspai:

**Motivation**

Fixes #2411 

**Summary**

Proposer bot can settle proposals and disputes it was involved in, as a proposer and disputer respectively.

Also improved `OptimisticOracleClient` to filter out proposals that have already been settled.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2411

### DIFF
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
@@ -117,32 +117,45 @@ class OptimisticOracleClient {
       };
     });
 
-    // Store proposals that have NOT been disputed:
-    let undisputedProposals = proposalEvents
-      .filter(event => {
-        const key = this._getPriceRequestKey(event);
-        const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
-        return hasDispute === undefined;
+    // Store proposals that have NOT been disputed and have NOT been settled, and reformat data.
+    let undisputedProposals = proposalEvents.filter(event => {
+      const key = this._getPriceRequestKey(event);
+      const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
+      return hasDispute === undefined;
+    });
+    let unsettledProposals = await Promise.all(
+      undisputedProposals.map(async event => {
+        const state = await this.oracle.methods
+          .getState(
+            event.returnValues.requester,
+            event.returnValues.identifier,
+            event.returnValues.timestamp,
+            event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x"
+          )
+          .call();
+
+        // For unsettled proposals, reformat the data:
+        if (state !== OptimisticOracleRequestStatesEnum.SETTLED) {
+          return {
+            requester: event.returnValues.requester,
+            proposer: event.returnValues.proposer,
+            identifier: this.hexToUtf8(event.returnValues.identifier),
+            ancillaryData: event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x",
+            timestamp: event.returnValues.timestamp,
+            proposedPrice: event.returnValues.proposedPrice,
+            expirationTimestamp: event.returnValues.expirationTimestamp,
+            currency: event.returnValues.currency
+          };
+        }
       })
-      .map(event => {
-        return {
-          requester: event.returnValues.requester,
-          proposer: event.returnValues.proposer,
-          identifier: this.hexToUtf8(event.returnValues.identifier),
-          ancillaryData: event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x",
-          timestamp: event.returnValues.timestamp,
-          proposedPrice: event.returnValues.proposedPrice,
-          expirationTimestamp: event.returnValues.expirationTimestamp,
-          currency: event.returnValues.currency
-        };
-      });
+    ).filter(event => event !== undefined);
 
     // Filter proposals based on their expiration timestamp:
     const isExpired = proposal => {
       return Number(proposal.expirationTimestamp) <= Number(currentTime);
     };
-    this.expiredProposals = undisputedProposals.filter(proposal => isExpired(proposal));
-    this.undisputedProposals = undisputedProposals.filter(proposal => !isExpired(proposal));
+    this.expiredProposals = unsettledProposals.filter(proposal => isExpired(proposal));
+    this.undisputedProposals = unsettledProposals.filter(proposal => !isExpired(proposal));
 
     // Store disputes that were resolved and can be settled:
     let resolvedDisputeEvents = await Promise.all(
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
@@ -123,7 +123,7 @@ class OptimisticOracleClient {
       const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
       return hasDispute === undefined;
     });
-    let unsettledProposals = await Promise.all(
+    const unsettledProposals = await Promise.all(
       undisputedProposals.map(async event => {
         const state = await this.oracle.methods
           .getState(
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
@@ -118,7 +118,7 @@ class OptimisticOracleClient {
     });
 
     // Store proposals that have NOT been disputed and have NOT been settled, and reformat data.
-    let undisputedProposals = proposalEvents.filter(event => {
+    const undisputedProposals = proposalEvents.filter(event => {
       const key = this._getPriceRequestKey(event);
       const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
       return hasDispute === undefined;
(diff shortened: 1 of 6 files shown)


packages/optimistic-oracle/src/proposer.js
```diff
@@ -296,8 +296,86 @@ class OptimisticOracleProposer {
...
+    this.logger.debug({
+      at: "OptimisticOracleProposer#settleRequests",
+      message: "Checking for proposals and disputes to settle"
+    });
+
+    for (let priceRequest of this.optimisticOracleClient
+      .getSettleableProposals(this.account)
+      .concat(this.optimisticOracleClient.getSettleableDisputes(this.account))) {
+      // Create the transaction.
+      const settle = this.ooContract.methods.settle(
```
daywiss, line 310: this looks good to me, the only nit i have here is that id like to see a function which takes in a single priceRequest and has this logic, rather than it being in a loop. then the loop can call that function

daywiss (approved):
@nicholaspai this looks good to me, but you might want to get a second opinion from chris or matt

packages/optimistic-oracle/src/proposer.js
nicholaspai: Good feedback, I refactored this method and `sendDisputes()` and `sendProposals()`

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -117,32 +117,45 @@ class OptimisticOracleClient {
...
     });
 
-    // Store proposals that have NOT been disputed:
-    let undisputedProposals = proposalEvents
-      .filter(event => {
-        const key = this._getPriceRequestKey(event);
-        const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
-        return hasDispute === undefined;
+    // Store proposals that have NOT been disputed and have NOT been settled, and reformat data.
+    let undisputedProposals = proposalEvents.filter(event => {
```
mrice32, line 121:
```suggestion
    const undisputedProposals = proposalEvents.filter(event => {
```

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -117,32 +117,45 @@ class OptimisticOracleClient {
...
+      const key = this._getPriceRequestKey(event);
+      const hasDispute = disputeEvents.find(disputeEvent => this._getPriceRequestKey(disputeEvent) === key);
+      return hasDispute === undefined;
+    });
+    let unsettledProposals = await Promise.all(
```
mrice32, line 126:
```suggestion
    const unsettledProposals = await Promise.all(
```

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
     });
 
     for (let priceRequest of this.optimisticOracleClient.getUnproposedPriceRequests()) {
-      const priceFeed = await this._createOrGetCachedPriceFeed(priceRequest.identifier);
+      await this._sendProposal(priceRequest);
```
mrice32, line 76: Maybe add a TODO here where we can optionally configure a reward threshold here so the bot only proposes for requests that are adequately rewarded?

This isn't a trivial feature, but I think it's worth noting that this is probably required to keep people from spamming the bot.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
...
+
+    // With pricefeed successfully constructed, confirm the proposal price
+    await priceFeed.update();
+    let disputePrice;
+    try {
+      disputePrice = (await priceFeed.getHistoricalPrice(priceRequest.timestamp)).toString();
+    } catch (error) {
+      this.logger.error({
+        at: "OptimisticOracleProposer#sendDisputes",
+        message: "Failed to query historical price for price request",
```
mrice32, line 240: nit: might be useful to include the timestamp and certain fields of the price request in this error log.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
...
+        disputePrice,
+        error
+      });
+      return;
+    }
 
+    // If proposal price is not equal to the dispute price, then prepare dispute
+    // TODO: Implement dispute buffer feature to only dispute if proposal and dispute price
+    // differ using some margin of error. For example, we could define a variable `proposalErrorPrecision`
+    // and set it to 5, to only send disputes if prices differ after rounding to 5 decimals of precision.
```
mrice32, line 250: +1

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
...
+      this.utf8ToHex(priceRequest.identifier),
+      priceRequest.timestamp,
+      priceRequest.ancillaryData,
+      proposalPrice
+    );
+
+    // Simple version of inventory management: simulate the transaction and assume that if it fails,
+    // the caller didn't have enough collateral.
+    let proposalBond, gasEstimation;
+    try {
```
mrice32, line 153: Not something we should do in this PR, but some of this transaction sending logic seems like it could be generalized and tossed into a library that could be used here and in the other bots since we do this in so many places. There are _some_ differences in the logic that make this a little difficult, but I could basically imagine a function that's called like this:
```js
const proposal = this.ooContract.methods.proposePrice(
      priceRequest.requester,
      this.utf8ToHex(priceRequest.identifier),
      priceRequest.timestamp,
      priceRequest.ancillaryData,
      proposalPrice
    );
await runTransaction(proposal).catch(e => {
  if (e.type === "call") {
    logger.error({
      ...
    });
  } else if (e.type === "send") {
    logger.error({
      ...
    });
  }
});
```

mrice32 (approved):
LGTM! A few minor comments around things we might want to add in the future but looks great overall -- I think this puts us way closer to being able to run this in prod! Very very nice work

packages/optimistic-oracle/index.js
```diff
@@ -69,8 +69,8 @@ async function run({ logger, web3, pollingDelay, errorRetries, errorRetriesTimeo
         async () => {
           await ooProposer.update();
           await ooProposer.sendProposals();
-          // await ooProposer.sendDisputes();
-          // await ooProposer.settleRequests();
+          await ooProposer.sendDisputes();
```
pemulis, line 72: Should this be `optimisticOracleProposer` to be more explicit? We use the expanded name elsewhere.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
...
+        at: "OptimisticOracleProposer#sendProposals",
+        message: "Failed to query historical price for price request",
+        priceRequest,
+        error
+      });
+      return;
+    }
+
+    // Create the transaction.
+    const proposal = this.ooContract.methods.proposePrice(
```
pemulis, line 142: Similar question about whether `ooContract` should expanded to `optimisticOracleContract`.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -73,60 +73,209 @@ class OptimisticOracleProposer {
...
+        at: "OptimisticOracleProposer#sendDisputes",
+        message: "Failed to query historical price for price request",
+        disputePrice,
+        error
+      });
+      return;
+    }
 
+    // If proposal price is not equal to the dispute price, then prepare dispute
+    // TODO: Implement dispute buffer feature to only dispute if proposal and dispute price
```
pemulis, line 248: This is addressed in https://github.com/UMAprotocol/protocol/pull/2564?

packages/optimistic-oracle/src/proposer.js
nicholaspai: Greet feature. Will definitely todo for this pr and implement in next pr

packages/optimistic-oracle/src/proposer.js
nicholaspai: Ah this is a typo, this should print out `priceRequest` not `proposalPrice` which should be undefined if we catch an error here.

packages/optimistic-oracle/src/proposer.js
nicholaspai: Will do in next PR.

packages/optimistic-oracle/index.js
nicholaspai: Yeah good point

packages/optimistic-oracle/src/proposer.js
nicholaspai: Yup will do

packages/optimistic-oracle/src/proposer.js
nicholaspai: Correct

packages/optimistic-oracle/src/proposer.js
mrice32: Because we need to translate reward amounts between currencies, this may be something we save for a v2 since it may be pretty complicated to implement... but definitely interested to hear if you have a clever solution to solve this in a simpler way.

mrice32 (approved):
LGTM!

packages/optimistic-oracle/src/proposer.js
nicholaspai: Yeah simple solution is no need to transalate reward amounts, just denominate the reward threshold in the collateral currency.

PR #2573, opened 2021-02-19
refactor(optimistic-oracle-proposer): Create common library to manage transaction lifecycle

### LINKED ISSUE #2572: optimistic-oracle-proposer: refactor transaction sending logic
**Description**

`sendProposals`, `sendDisputes`, and `settle` re-use common logic in simulating and sending a transaction, this should be refactored to an external library.

Ideally this external library should be re-used by disputer and liquidator.

nicholaspai:
**Motivation**

#2572

Refactors shared logic out of internal methods of `optimisticOracleProposer` that make `.call()` and `.send()` calls on web3 transactions. Future work can use this library in other bots that manage transactions like `liquidator` and `disputer`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2572

### DIFF
--- a/packages/common/src/ContractUtils.js
+++ b/packages/common/src/ContractUtils.js
@@ -42,7 +42,7 @@ const runTransaction = async ({ transaction, config }) => {
   try {
     [returnValue] = await Promise.all([transaction.call(config)]);
   } catch (error) {
-    error.type === "call";
+    error.type = "call";
     throw error;
   }
 
@@ -55,7 +55,7 @@ const runTransaction = async ({ transaction, config }) => {
       returnValue
     };
   } catch (error) {
-    error.type === "send";
+    error.type = "send";
     throw error;
   }
 };
--- a/packages/common/src/ContractUtils.js
+++ b/packages/common/src/ContractUtils.js
@@ -27,6 +27,61 @@ const revertWrapper = result => {
   return result;
 };
 
+/**
+ * Simulate transaction via .call() and then .send() and return receipt. If an error is thrown,
+ * return the error and add a flag denoting whether it was sent on the .call() or the .send().
+ * @param {*Object} transaction Transaction to call `.call()` and subsequently `.send()` on from `senderAccount`.
+ * @param {*String} senderAccount Sender of transaction. Injected to `transactionConfig`.
+ * @param {*Object} transactionConfig { gasPrice } passed to web3 transaction.
+ * @param {*Number} transactionGasLimit Maximum gas limit to set for sending transaction.
+ * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async (
+  transaction,
+  senderAccount,
+  transactionConfig,
+  transactionGasLimit,
+  gasLimitBuffer = 1.25
+) => {
+  // First try to simulate transaction and also extract return value if its
+  // a state-modifying transaction. If the function is state modifying, then successfully
+  // sending it will return the transaction receipt, not the return value, so we grab it here.
+  let returnValue, gasEstimation;
+  try {
+    [returnValue, gasEstimation] = await Promise.all([
+      transaction.call({ from: senderAccount }),
+      transaction.estimateGas({ from: senderAccount })
+    ]);
+  } catch (error) {
+    return {
+      error,
+      type: "call"
+    };
+  }
+
+  // .call() succeeded, now broadcast transaction.
+  transactionConfig = {
+    ...transactionConfig,
+    from: senderAccount,
+    gas: Math.min(Math.floor(gasEstimation * gasLimitBuffer), transactionGasLimit)
+  };
+  let receipt;
+  try {
+    receipt = await transaction.send(transactionConfig);
+    return {
+      receipt,
+      returnValue
+    };
+  } catch (error) {
+    return {
+      error,
+      type: "send"
+    };
+  }
+};
+
 module.exports = {
-  revertWrapper
+  revertWrapper,
+  runTransaction
 };
--- a/packages/common/src/ContractUtils.js
+++ b/packages/common/src/ContractUtils.js
@@ -38,9 +38,9 @@ const runTransaction = async ({ transaction, config }) => {
   // First try to simulate transaction and also extract return value if its
   // a state-modifying transaction. If the function is state modifying, then successfully
   // sending it will return the transaction receipt, not the return value, so we grab it here.
-  let returnValue;
+  let returnValue, estimatedGas;
   try {
-    [returnValue] = await Promise.all([transaction.call(config)]);
+    [returnValue, estimatedGas] = await Promise.all([transaction.call(config), transaction.estimateGas(config)]);
   } catch (error) {
     error.type = "call";
     throw error;
@@ -49,7 +49,7 @@ const runTransaction = async ({ transaction, config }) => {
   // .call() succeeded, now broadcast transaction.
   let receipt;
   try {
-    receipt = await transaction.send(config);
+    receipt = await transaction.send({ ...config, gas: estimatedGas });
     return {
       receipt,
       returnValue
--- a/packages/common/src/ContractUtils.js
+++ b/packages/common/src/ContractUtils.js
@@ -31,53 +31,32 @@ const revertWrapper = result => {
  * Simulate transaction via .call() and then .send() and return receipt. If an error is thrown,
  * return the error and add a flag denoting whether it was sent on the .call() or the .send().
  * @param {*Object} transaction Transaction to call `.call()` and subsequently `.send()` on from `senderAccount`.
- * @param {*String} senderAccount Sender of transaction. Injected to `transactionConfig`.
- * @param {*Object} transactionConfig { gasPrice } passed to web3 transaction.
- * @param {*Number} transactionGasLimit Maximum gas limit to set for sending transaction.
- * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @param {*Object} config transaction config, e.g. { gasPrice, from }, passed to web3 transaction.
  * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
  */
-const runTransaction = async (
-  transaction,
-  senderAccount,
-  transactionConfig,
-  transactionGasLimit,
-  gasLimitBuffer = 1.25
-) => {
+const runTransaction = async ({ transaction, config }) => {
   // First try to simulate transaction and also extract return value if its
   // a state-modifying transaction. If the function is state modifying, then successfully
   // sending it will return the transaction receipt, not the return value, so we grab it here.
-  let returnValue, gasEstimation;
+  let returnValue;
   try {
-    [returnValue, gasEstimation] = await Promise.all([
-      transaction.call({ from: senderAccount }),
-      transaction.estimateGas({ from: senderAccount })
-    ]);
+    [returnValue] = await Promise.all([transaction.call(config)]);
   } catch (error) {
-    return {
-      error,
-      type: "call"
-    };
+    error.type === "call";
+    throw error;
   }
 
   // .call() succeeded, now broadcast transaction.
-  transactionConfig = {
-    ...transactionConfig,
-    from: senderAccount,
-    gas: Math.min(Math.floor(gasEstimation * gasLimitBuffer), transactionGasLimit)
-  };
   let receipt;
   try {
-    receipt = await transaction.send(transactionConfig);
+    receipt = await transaction.send(config);
     return {
       receipt,
       returnValue
     };
   } catch (error) {
-    return {
-      error,
-      type: "send"
-    };
+    error.type === "send";
+    throw error;
   }
 };
 
--- a/packages/optimistic-oracle/src/proposer.js
+++ b/packages/optimistic-oracle/src/proposer.js
@@ -221,13 +221,13 @@ class OptimisticOracleProposer {
         expirationTimestamp: receipt.events.ProposePrice.returnValues.expirationTimestamp
       };
       this.logger.info({
-        at: "OptimisticOracleProposer#sendProposals",
+        at: "OptimisticOracleProposer#sendDisputes",
         message: "Proposed price!💍",
         priceRequest,
         proposalBond: returnValue,
         proposalPrice,
         txnConfig,
-        proposalResult: logResult
+        disputeResult: logResult
       });
     }
   }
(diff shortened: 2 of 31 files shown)


packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+ * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async (
+  transaction,
+  senderAccount,
+  transactionConfig,
+  transactionGasLimit,
+  gasLimitBuffer = 1.25
+) => {
```
mrice32, line 46: Thoughts on making this an object input? i.e:
```js
const runTransaction = async ({
  transaction,
  senderAccount,
  transactionConfig,
  transactionGasLimit,
  gasLimitBuffer = 1.25
}) => {
```

I only suggest that because with this many arguments, it can be challenging for the caller to get the ordering right.

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+      error,
+      type: "call"
+    };
+  }
+
+  // .call() succeeded, now broadcast transaction.
+  transactionConfig = {
+    ...transactionConfig,
+    from: senderAccount,
+    gas: Math.min(Math.floor(gasEstimation * gasLimitBuffer), transactionGasLimit)
```
mrice32, line 67: I'm guessing this is something we do elsewhere, but why do we need a max gas limit? Is this to avoid going over the block gas limit? Or something else?

Just wondering because it might make sense to have a "sensible default" for this value if we normally pass the same max.

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+ * Simulate transaction via .call() and then .send() and return receipt. If an error is thrown,
+ * return the error and add a flag denoting whether it was sent on the .call() or the .send().
+ * @param {*Object} transaction Transaction to call `.call()` and subsequently `.send()` on from `senderAccount`.
+ * @param {*String} senderAccount Sender of transaction. Injected to `transactionConfig`.
+ * @param {*Object} transactionConfig { gasPrice } passed to web3 transaction.
+ * @param {*Number} transactionGasLimit Maximum gas limit to set for sending transaction.
+ * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async (
```
mrice32, line 40: really cool that we don't even need to pass web3. +1

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+  gasLimitBuffer = 1.25
+) => {
+  // First try to simulate transaction and also extract return value if its
+  // a state-modifying transaction. If the function is state modifying, then successfully
+  // sending it will return the transaction receipt, not the return value, so we grab it here.
+  let returnValue, gasEstimation;
+  try {
+    [returnValue, gasEstimation] = await Promise.all([
+      transaction.call({ from: senderAccount }),
+      transaction.estimateGas({ from: senderAccount })
```
mrice32, line 54: Should these also use the transactionConfig?
```js
    [returnValue, gasEstimation] = await Promise.all([
      transaction.call({ ...transactionConfig, from: senderAccount }),
      transaction.estimateGas({ ...transactionConfig, from: senderAccount })
```

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+ * @param {*Object} transactionConfig { gasPrice } passed to web3 transaction.
+ * @param {*Number} transactionGasLimit Maximum gas limit to set for sending transaction.
+ * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async (
+  transaction,
+  senderAccount,
+  transactionConfig,
+  transactionGasLimit,
```
mrice32, line 44: Thoughts on shortening these variable names?
```suggestion
  sender,
  config,
  gasLimit,
```

packages/optimistic-oracle/src/proposer.js
```diff
@@ -157,80 +159,77 @@ class OptimisticOracleProposer {
...
+        return;
+      } else {
+        this.logger.error({
+          at: "OptimisticOracleProposer#sendProposals",
+          message: "Failed to propose price🚨",
+          priceRequest,
+          error: transactionResult.error
+        });
+        return;
+      }
```
mrice32, line 206: I think you can shorten this because the log is exactly the same except for the message (you should be able to do the same elsewhere).
```suggestion
        const message = transactionResult.error.type === "call" ? "Cannot propose price: not enough collateral (or large enough approval)✋" : "Failed to propose price🚨";
        this.logger.error({
          at: "OptimisticOracleProposer#sendProposals",
          message,
          priceRequest,
          error: transactionResult.error
        });
        return;
```

packages/optimistic-oracle/src/proposer.js
```diff
@@ -286,159 +285,152 @@ class OptimisticOracleProposer {
...
 
-      // Send the transaction or report failure.
-      let receipt;
-      try {
-        receipt = await dispute.send(txnConfig);
-      } catch (error) {
-        this.logger.error({
+      // Handle error:
+      if (transactionResult.error) {
+        if (transactionResult.error.type === "call") {
```
mrice32, line 317: nit: here and elsewhere, I think this should be `transactionResult.type` rather than `transactionResult.error.type` since the type is returned _next to_ the error, not as a field inside of it.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -286,159 +285,152 @@ class OptimisticOracleProposer {
...
+          this.logger.error({
+            at: "OptimisticOracleProposer#sendDisputes",
+            message: "Cannot dispute price: not enough collateral (or large enough approval)✋",
+            priceRequest,
+            error: transactionResult.error
+          });
+          return;
+        } else {
```
mrice32, line 325: I think you can get rid of this if, else in place of a ternary -- see the suggestion for the other error handling logic above.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -286,159 +285,152 @@ class OptimisticOracleProposer {
...
+    if (transactionResult.error) {
+      if (transactionResult.error.type === "call") {
+        this.logger.error({
+          at: "OptimisticOracleProposer#settleRequests",
+          message: "Cannot settle for unknown reason☹️",
+          priceRequest,
+          error: transactionResult.error
+        });
+        return;
+      } else {
```
mrice32, line 399: Same suggestion here as above -- only the message differs, so you should be able to get rid of this if and just use a ternary to determine the `message`.

mrice32 (commented):
Looks awesome! Really quick turnaround on this library. Very very cool. Some minor comments.

daywiss (commented):
im just curious @nicholaspai, are gas estimations that bad that web3 needs all this help around sending the TX?

packages/common/src/ContractUtils.js
nicholaspai: I think we just have one to allow more transaction customization, but let's just hardcode a limit for now since we never change this.

packages/common/src/ContractUtils.js
nicholaspai: Yeah but the config will be slightly different for the `.send()` because we will use the `gasEstimation`

nicholaspai: but we can pass the gas price at least.

packages/common/src/ContractUtils.js
nicholaspai: Done

packages/optimistic-oracle/src/proposer.js
nicholaspai: Great optimization

packages/optimistic-oracle/src/proposer.js
nicholaspai: Ah yes agreed

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+  let returnValue, gasEstimation;
+  try {
+    [returnValue, gasEstimation] = await Promise.all([
+      transaction.call({ from: senderAccount }),
+      transaction.estimateGas({ from: senderAccount })
+    ]);
+  } catch (error) {
+    return {
+      error,
+      type: "call"
```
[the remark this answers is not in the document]
nicholaspai: Interesting, we should build out our own custom Errors in a future PR.

nicholaspai:
> im just curious @nicholaspai, are gas estimations that bad that web3 needs all this help around sending the TX?

I'm not sure, I was just following patterns from other bots. But i'll remove gas estimations optimistically :)

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,61 @@ const revertWrapper = result => {
...
+ * @param {*Number} transactionGasLimit Maximum gas limit to set for sending transaction.
+ * @param {*Number} gasLimitBuffer Multiplier applied to Truffle's estimated gas limit for a transaction to send.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async (
+  transaction,
+  senderAccount,
+  transactionConfig,
+  transactionGasLimit,
+  gasLimitBuffer = 1.25
```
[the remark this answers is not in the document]
nicholaspai: Yeah that'd be great, i followed your other suggestion and changed `runTransaction` to take only 2 params: `transaction` and `config`

packages/common/src/ContractUtils.js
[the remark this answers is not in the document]
nicholaspai: I ended up just relying on web3 estimation, we'll see in war games if this works

nicholaspai:
@mrice32 @daywiss thanks for the great suggestions, I greatly simplified the `config` object passed to `runTransaction` and simplified error handling

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,40 @@ const revertWrapper = result => {
...
+    [returnValue] = await Promise.all([transaction.call(config)]);
+  } catch (error) {
+    error.type === "call";
+    throw error;
+  }
+
+  // .call() succeeded, now broadcast transaction.
+  let receipt;
+  try {
+    receipt = await transaction.send(config);
```
mrice32, line 52: How well does web3 estimate the gas to send? I think we used to have an issue where it was always sending some huge amount, like 6MM gas, which caused problems for other people who ran bots with less ETH in their account.

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,40 @@ const revertWrapper = result => {
...
+ */
+const runTransaction = async ({ transaction, config }) => {
+  // First try to simulate transaction and also extract return value if its
+  // a state-modifying transaction. If the function is state modifying, then successfully
+  // sending it will return the transaction receipt, not the return value, so we grab it here.
+  let returnValue;
+  try {
+    [returnValue] = await Promise.all([transaction.call(config)]);
+  } catch (error) {
+    error.type === "call";
```
daywiss, line 45: should be just a single equal to set, and same with send

packages/optimistic-oracle/src/proposer.js
```diff
@@ -286,159 +266,126 @@ class OptimisticOracleProposer {
...
+          priceRequest,
+          disputePrice,
+          disputeBond: returnValue,
+          allowedError: this.disputePriceErrorPercent,
+          transactionConfig,
+          disputeResult: logResult
+        });
       } catch (error) {
+        const message =
+          error.type === "call"
```
daywiss, line 319: how hard it is to make a test case to check this is working, ie check both branches of the log message, so it would have caught the issue of this not setting correctly?

packages/common/src/ContractUtils.js
```diff
@@ -27,6 +27,40 @@ const revertWrapper = result => {
...
+ * @param {*Object} config transaction config, e.g. { gasPrice, from }, passed to web3 transaction.
+ * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
+ */
+const runTransaction = async ({ transaction, config }) => {
+  // First try to simulate transaction and also extract return value if its
+  // a state-modifying transaction. If the function is state modifying, then successfully
+  // sending it will return the transaction receipt, not the return value, so we grab it here.
+  let returnValue;
+  try {
+    [returnValue] = await Promise.all([transaction.call(config)]);
```
mrice32, line 43: Why not drop the Promise.all?
```js
returnValue = await transaction.call(config);
```

mrice32 (approved):
Looks great! Two pretty minor comments/questions.

packages/common/src/ContractUtils.js
mrice32: +1

nicholaspai: Done

packages/optimistic-oracle/src/proposer.js
nicholaspai: I'd like to unit test the `common/ContractUtils/runTransaction` method but I'm having trouble thinking of a situation where `.call()` succeeds but `.send()` fails... any ideas? @chrismaree @mrice32

packages/common/src/ContractUtils.js
nicholaspai: Not sure, but I do recall that issue. I think its reasonable to add an `estimateGas()` call to the promise array with the `.call()` and just use that value as the `.gas`

packages/common/src/ContractUtils.js
nicholaspai: See above, otherwise yes I should

packages/optimistic-oracle/src/proposer.js
nicholaspai: I added a unit test for `runTransaction` and left a TODO for how to test the `.send()` path

yarn.lock
```diff
@@ -3830,31 +3830,6 @@
   resolved "https://registry.yarnpkg.com/@protobufjs/utf8/-/utf8-1.1.0.tgz#a777360b5b39a1a2e5106f8e858f2fd2d060c570"
   integrity sha1-p3c2C1s5oaLlEG+OhY8v0tBgxXA=
 
-"@sendgrid/client@^6.5.5":
```
nicholaspai, line 3833: I think we actually need to commit this `yarn.lock` since we changed package.json's recently by removing CLI and reporters packages. Thoughts?

daywiss (approved):

packages/optimistic-oracle/src/proposer.js
daywiss: i see yea, that makes sense. its not a huge deal though

PR #2574, opened 2021-02-19
fix(PriceIdentifierUtils): Remove identifiers that should be scaled to 18 decimals

smb2796:
**Motivation**

The new EMP and perp templates require that price identifiers be scaled to 18 decimal places. Price identifiers that are already being used in legacy EMP contracts will need to remain on this list. Price ids without live contracts should be removed and have their accompanying UMIPs updated to note that the price identifier should be scaled to 18 decimals.

**Summary**

This PR updates the list of price identifiers that need to be scaled to non-18 decimals.

This is accompanied by this [PR](https://github.com/UMAprotocol/UMIPs/pull/194) to the UMIPs repo. This PR "deprecates" the identifiers that remain on the PriceIdentifierUtils list, and specifies 18 decimals scaling for all identifiers that are not on this list.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

**Issue(s)**
N/A


packages/common/src/PriceIdentifierUtils.js
```diff
@@ -12,16 +12,8 @@ const IDENTIFIER_NON_18_PRECISION = {
   "ELASTIC_STABLESPREAD/USDC": 6,
   BCHNBTC: 8,
   AMPLUSD: 6,
-  "uVOL-BTC-APR21": 6,
```
nicholaspai, line 15: To clarify: these are being removed because there are no live EMP's using them correct?

smb2796: Yes that is correct

smb2796:
Closing until the new factories are ready to be used and we are sure that these non 18 decimal identifiers are not being used by any contract from the old factory.

mrice32 (approved):

PR #2575, opened 2021-02-19
feat(financial-templates-lib): add LPPriceFeed

mrice32:
**Motivation**

We'd like to be able to price LP tokens. This LPPriceFeed can basically be used wherever there's a pool of tokens (one or more) and you'd like to know how much of a particular token is redeemable for a single LP share.

For instance, in a Uniswap pool with 50 WETH, 1 WBTC, and 25 LP shares, you could configure this feed to tell you how much WETH you'd get back for an LP share. The "price" that it would return would be 2 since the 50 WETH is spread over 25 shares. You could configure another instance of this feed to tell you how many WBTC each share is redeemable for. The "price" for that feed would be 0.04.

**Summary**

The LPPriceFeed was implemented very similarly to the VaultPriceFeed given that they do very similar things: make a smart contract call or two, do some math, and return the result in the right decimals.

Note: this PR also makes a few minor fixes to the VaultPriceFeed.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -16,6 +16,7 @@ const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
 const { getTruffleContract, getAbi } = require("@uma/core");
 const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
 const { VaultPriceFeed } = require("./VaultPriceFeed");
+const { LPPriceFeed } = require("./LPPriceFeed");
 const { BlockFinder } = require("./utils");
 
 async function createPriceFeed(logger, web3, networker, getTime, config) {
@@ -327,9 +328,30 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       ...config,
       logger,
       web3,
+      getTime,
       vaultAbi: getAbi("VaultInterface", "latest"),
       erc20Abi: getAbi("IERC20Standard", "latest"),
       vaultAddress: config.address,
+      blockFinder: getSharedBlockFinder(web3)
+    });
+  } else if (config.type === "lp") {
+    const requiredFields = ["poolAddress", "tokenAddress"];
+    if (isMissingField(config, requiredFields, logger)) {
+      return null;
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating LPPriceFeed",
+      config
+    });
+
+    return new LPPriceFeed({
+      ...config,
+      logger,
+      web3,
+      getTime,
+      erc20Abi: getAbi("IERC20Standard", "latest"),
       blockFinder: getSharedBlockFinder(web3)
     });
   }
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -13,7 +13,7 @@ const { TraderMadePriceFeed } = require("./TraderMadePriceFeed");
 const { PriceFeedMockScaled } = require("./PriceFeedMockScaled");
 const { InvalidPriceFeedMock } = require("./InvalidPriceFeedMock");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
-const { getTruffleContract, getAbi } = require("@uma/core");
+const { getTruffleContract } = require("@uma/core");
 const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
 const { VaultPriceFeed } = require("./VaultPriceFeed");
 const { LPPriceFeed } = require("./LPPriceFeed");
@@ -23,6 +23,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
   const Uniswap = getTruffleContract("Uniswap", web3, "latest");
   const ERC20 = getTruffleContract("ExpandedERC20", web3, "latest");
   const Balancer = getTruffleContract("Balancer", web3, "latest");
+  const VaultInterface = getTruffleContract("VaultInterface", web3, "latest");
 
   if (config.type === "cryptowatch") {
     const requiredFields = ["exchange", "pair", "lookback", "minTimeBetweenUpdates"];
@@ -329,8 +330,8 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       logger,
       web3,
       getTime,
-      vaultAbi: getAbi("VaultInterface", "latest"),
-      erc20Abi: getAbi("IERC20Standard", "latest"),
+      vaultAbi: VaultInterface.abi,
+      erc20Abi: ERC20.abi,
       vaultAddress: config.address,
       blockFinder: getSharedBlockFinder(web3)
     });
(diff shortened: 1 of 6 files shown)


packages/financial-templates-lib/test/price-feed/VaultPriceFeed.js
```diff
@@ -113,29 +113,6 @@ contract("VaultPriceFeed.js", function(accounts) {
     assert.equal(vaultPriceFeed.getLastUpdateTime(), mockTime); // Update time should have no incremented.
   });
 
-  it("Update Frequency", async function() {
```
mrice32, line 116: This test was accidentally copied twice in this file.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -327,11 +328,32 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       ...config,
       logger,
       web3,
+      getTime,
```
mrice32, line 331: This argument was missing from the VaultPriceFeed.

packages/financial-templates-lib/src/price-feed/VaultPriceFeed.js
```diff
@@ -29,6 +29,15 @@ class VaultPriceFeed extends PriceFeedInterface {
     priceFeedDecimals = 18
   }) {
     super();
+
+    // Assert required inputs.
+    assert(logger, "logger required");
```
mrice32, line 34: These checks were added to ensure that all required arguments are provided.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -327,11 +328,32 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+      message: "Creating LPPriceFeed",
+      config
+    });
+
+    return new LPPriceFeed({
+      ...config,
+      logger,
+      web3,
+      getTime,
+      erc20Abi: getAbi("IERC20Standard", "latest"),
```
nicholaspai, line 354: Why don't you grab this same way as UniswapPriceFeed?

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
```diff
@@ -0,0 +1,129 @@
...
+    tokenAddress,
+    getTime,
+    blockFinder,
+    minTimeBetweenUpdates = 60,
+    priceFeedDecimals = 18
+  }) {
+    super();
+
+    // Assert required arguments.
+    assert(logger, "logger required");
```
nicholaspai, line 36: Are these asserts needed for any fields that are in the requiredFields dict in CreatePriceFeed?

nicholaspai (approved):
LGTM

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
mrice32: I guess we _could_ leave them out, but I was thinking of them as a way to fail early if the CreatePriceFeed implementation was wrong. Thoughts?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Not super opinionated. Just figured this call was more directly what I needed since I don't need to construct a truffle contract to just get an ABI. Happy to change if you think `getTruffleContract` would be better.

mrice32: My guess is that the truffle one is that way because we did a large find-and-replace to change artifacts.require to getTruffleContract. However, now that we're using the core methods directly, getAbi is all that's needed here.

pemulis (approved):

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: OH! The truffle contract is already created at the top, so I can just pull it from there.

mrice32: Will update.

pemulis:
Looks good, and tests are all green.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Done.

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
```diff
@@ -0,0 +1,129 @@
...
+      this.poolDecimals = parseInt(await this.pool.methods.decimals().call());
+    }
+
+    return { poolDecimals: this.poolDecimals, tokenDecimals: this.tokenDecimals };
+  }
+
+  async _getPrice(blockNumber = "latest") {
+    const lpTotalSupply = this.toBN(await this.pool.methods.totalSupply().call(undefined, blockNumber));
+    const tokensInPool = this.toBN(
+      await this.token.methods.balanceOf(this.pool.options.address).call(undefined, blockNumber)
```
daywiss, line 101: How is it this has infinite lookback, is it only with archive node?

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
```diff
@@ -0,0 +1,129 @@
...
+    // Return infinity since this price feed can technically look back as far as needed.
+    return Infinity;
+  }
+
+  getPriceFeedDecimals() {
+    return this.priceFeedDecimals;
+  }
+
+  async update() {
+    const currentTime = await this.getTime();
```
daywiss, line 79: At this point the update func feels like it might as well be getPrice. But thats a comment for another pr 😄

packages/financial-templates-lib/test/price-feed/LPPriceFeed.js
```diff
@@ -0,0 +1,152 @@
...
+
+  it("Zero LP shares", async function() {
+    await token.mint(pool.address, parseFixed("25", tokenDecimals));
+
+    await lpPriceFeed.update();
+
+    assert.equal(lpPriceFeed.getCurrentPrice().toString(), "0");
+  });
+
+  it("Update Frequency", async function() {
```
daywiss, line 104: Wonder if this rate limit logic should be a util

daywiss (approved):
Nice implementation and lots of tests :godmode:

packages/financial-templates-lib/test/price-feed/LPPriceFeed.js
mrice32: Agreed. We repeat it in every price feed.

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
mrice32: Yeah, good point. For now, we assume the lookback is infinite because we don't know what type of node we're connected to.

packages/financial-templates-lib/src/price-feed/LPPriceFeed.js
mrice32: haha, yuup.

PR #2579, opened 2021-02-22
feat(trader): add Ranger Trader, Exchange Adapter and initial trader implementation

### LINKED ISSUE #2560: Implement range trading bot

chrismaree:
**Motivation**

#2560

Continuation of PR https://github.com/UMAprotocol/protocol/pull/2559 which added a generic DSProxy client and PR https://github.com/UMAprotocol/protocol/pull/2545 which added a smart contract to trade uniswap to a desired price. 

**Summary**

This PR adds the plumbing required to run a trading bot which leveraged the Uniswap Broker & DSProxy client to keep a synthetic in line with a desired price. The implementation presented in this PR is simpler than that originally proposed in the 
internal "Range Trading Regulator Doc". This doc describes the desire for more complex inventory management wherein a delay is used to define when a trade should occur. Specifically, the Regulator would only preform a trade if the market movement was lager then a desired % over a given window. This functionality can be added later once we are happy with the basic logic proposed within this PR.

This PR follows the same pattern that the other bots use:

1) an `index.ts` file which acts as the bots main entry point. This file pulls together the price feeds, initialises the clients and preforms the execution loop.
2) sub models, which have scoped domains of responsibility. in particular:
2.1) `RangeTrader.ts` which takes in two price feeds and preforms trades if the price deviation is larger than a given threshold.
2.2) A standard interface for an `ExchangeAdapter` which acts to wrap Dev & exchange logic. This includes an `ExchangeAdapterInterface` which is similar to the `PriceFeedInterface` to wrap common logic, a `CreateExchangeAdapter` which makes new exchange adapters and a `UniswapTrader` which implements the interface and facilities interaction with the `UniswapBroker` Added in a previous PR.

**Details**

The bot created in this PR follows a very simple set of heuristics to define if a trade should be preformed. When creating the bot, two prices feeds are ingested: a) a `tokenPriceFeed` which represents the price of the synthetic trading within a decentralised exchange (for now Uniswap) and b) a `referencePriceFeed` which represents the "true" market price for the synthetic.

If the bot detects a `deviationError` `δ` larger than the provided `tradeExecutionThreshold`, then the bot will execute a correcting trade. The bot will trade the price back to a price within the window of `targetPriceSpread` `△` away from the "true" price. This is defined by a `priceScalar` which informs how far off the resultant price should be from the "true" price. This `priceScalar` is governed by the following equation:
image

By default, the bot sets the `tradeExecutionThreshold` to 0.2, such that only errors larger than 0.2 will yield a trade and a `targetPriceSpread` of 0.05, such that the desired price is 5% off of the "true" market price. These parameters can be tuned according to the desired bot behaviour.

**Numerical example**
For example, say a synthetic is trading at a price of 1250, with the reference price feed at 1000. This yields a `deviationError` of δ = (observed - expected) / expected = (1250 - 1000) / 1000 = 0.25. As this is larger than the `tradeExecutionThreshold` a trade will be executed. The `priceScalar = targetPriceSpread + 1 = 1 + 0.05 = 1.05`. The bot will therefore try to trade the price _down_ to 1.05x the desired price, or 1050. 

Similarly, if the synthetic is trading at 750 then the deviation error would be -0.25. Then, the the scalar = 1 - 0.05 = 0.95. Therefore the bot will trade the price _up_ to 950.

This implementation amounts to the bot trying to trade the price back to parity, with a window on either side of how close to the "true" price the synthetic should be.

In future, we can add an additional trading module that implements the "Regulator" behaviour described in the internal doc.

**Questions to ENG**
There are a few places where I think I could have been way more strict about typscript interfaces. Rather, I simply defaulted to the `any` type. I did this as I was not exactly sure on what the best practice is with this kind of interfacing. I would love some feedback on the best way to approach this in the review.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

close #2560


packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
```diff
@@ -23,57 +23,56 @@ contract UniswapBroker {
      * possible to the truePrice.
      * @dev True price is expressed in the ratio of token A to token B.
      * @dev The caller must approve this contract to spend whichever token is intended to be swapped.
+     * @param tradingAsEOA bool to indicate if the UniswapBroker is being called by a DSProxy or an EOA.
```
chrismaree, line 26: few small changes the the `UniswapBroker`. The two main changes are:
1) group common vars into arrays (i.e `tokenA` & `tokenB`) are joined together to form an array of addresses.
2) add `tradingAsEOA` that adds an extra token transfer that is only required when using EOA to enable DSProxy support in this module.

packages/core/test/trader/uniswap-broker/UniswapBroker.js
```diff
@@ -147,14 +147,12 @@ contract("UniswapBroker", function(accounts) {
 
     // Now we can actually execute the swapToPrice method to ensure that the contract correctly modifies the spot price.
     await uniswapBroker.swapToPrice(
+      true, // The swap is being executed as an EOA. This ensures that the correct token transfers are done.
```
chrismaree, line 150: these changes just update the interface to match the new contract.

packages/financial-templates-lib/index.js
```diff
@@ -7,9 +7,11 @@ module.exports = {
   ...require("./src/helpers/GasEstimator"),
   ...require("./src/logger/Logger"),
   ...require("./src/logger/SpyTransport"),
+  ...require("./src/price-feed/UniswapPriceFeed"),
```
chrismaree, line 10: additional exports that are needed to test the trader correctly.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -566,6 +566,17 @@ async function createUniswapPriceFeedForFinancialContract(
 
   const userConfig = config || {};
 
+  // Check if there is an override for the getTime method in the price feed config. Specifically, we can replace the
```
chrismaree, line 569: this `if` logic before enabled the uniswap feeds to be used correctly in hard hat when created from the `createReferencePriceFeedForFinancialContract` method. I added this branching logic to the uniswap feeds as well as this kind of config is needed when testing the trader bot when using the `createTokenPriceFeedForFinancialContract` method. This only effects unit tests.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -701,8 +712,12 @@ async function createReferencePriceFeedForFinancialContract(
 }
 
 function getFinancialContractIdentifierAtAddress(web3, financialContractAddress) {
-  const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, "1.2.0");
-  return new web3.eth.Contract(ExpiringMultiParty.abi, financialContractAddress);
+  try {
```
chrismaree, line 715: if you provide this method with an unknown identifier or an invalid EMP it tends to throw a really nasty hard to debug error. As a result, I've wrapped it in a try catch to try and help with debugging these cases.

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,171 @@
+const { ConvertDecimals, createFormatFunction, createObjectFromDefaultProps } = require("@uma/common");
```
chrismaree, line 1: This is the key logic for the RangeTrader that executes trades based on a range movement of a synthetic price.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
```diff
@@ -0,0 +1,47 @@
+const assert = require("assert");
```
chrismaree, line 1: This file is very similar to the `CreatePriceFeed` except it create `ExchangeAdapters`. For now, there is just one exchange adapter. when we add additional exchange support to this bot we will add them here.

packages/trader/src/exchange-adapters/ExchangeAdapterInterface.ts
```diff
@@ -0,0 +1,19 @@
+class ExchangeAdapterInterface {
```
chrismaree, line 1: all exchange adapters need to implement this interface, just like with price feeds.

packages/trader/src/exchange-adapters/UniswapTrader.ts
```diff
@@ -0,0 +1,92 @@
+import assert from "assert";
```
chrismaree, line 1: this is the uniswap exchange adapter. it simply implements the interface and executes the trade using the `DSProxyManager` using the `UniswapBroker` implementation.

packages/trader/src/index.ts
```diff
@@ -1,33 +1,136 @@
 import { config } from "dotenv";
```
chrismaree, line 1: this is the bots main entry point. it is structure very similarly to other bot's `index.js` files. The main difference is the configs come from the `TraderConfig` object, rather than from within this `index` file. I think this is a great pattern as it makes things cleaner with a more well defined separation of concerns.

packages/trader/src/index.ts
```diff
@@ -1,33 +1,136 @@
...
+      gasEstimator,
+      account: accounts[0],
+      dsProxyFactoryAddress: config.dsProxyFactoryAddress,
+      dsProxyFactoryAbi: getAbi("DSProxyFactory"),
+      dsProxyAbi: getAbi("DSProxy")
+    });
+
+    await dsProxyManager.initializeDSProxy();
+
+    const [tokenPriceFeed, referencePriceFeed, exchangeAdapter] = await Promise.all([
```
chrismaree, line 59: This bot relies on two main feeds: a `tokenPriceFeed` which is the synth market price (Uniswap feed, for example) and `referencePriceFeed` which is the "true" token price (desired) from the external market.

packages/trader/test/RangeTrader.ts
```diff
@@ -0,0 +1,338 @@
+import { web3, assert } from "hardhat";
```
chrismaree, line 1: These tests test the `RangeTrader` module directly, just like the `Liquidator.js` tests test the liquidator. They verify the range trading behaviour against expected market movements.

packages/trader/test/index.ts
```diff
@@ -1,23 +1,187 @@
 import { run } from "../src/index";
```
chrismaree, line 1: This set of tests correctly validates that the `index` can start up with no error. In future this should include re-try logic tests as well as some others.

chrismaree:
Note it looks like this PR is failing the linter. I'm looking into why husky is not working as expected with ts.

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
...
+
+    // Validate and set config settings to class state.
+    const configWithDefaults = createObjectFromDefaultProps(rangeTraderConfig, defaultConfig);
+    Object.assign(this, configWithDefaults);
+
+    this.fixedPointAdjustment = this.toBN(this.toWei("1"));
+  }
+
+  async update() {
+    await Promise.all([this.tokenPriceFeed.update(), this.referencePriceFeed.update()]);
```
daywiss, line 85: is this function needed anymore, considering it can be done as needed in other functions?

daywiss (approved):
great work chris. i think we will really need to do some trial and error on how to get third party types in and reduce the use of any.  most my comments are nits so implement as you see fit.

packages/core/test/trader/uniswap-broker/UniswapBroker.js
```diff
@@ -221,14 +219,12 @@ contract("UniswapBroker", function(accounts) {
...
       router.address,
       factory.address,
-      tokenA.address,
-      tokenB.address,
-      "1000", // The "true" price of the pair is expressed as the ratio of token A to token B. A price of 1000 is simply 1000/1.
-      "1",
-      MAX_UINT_VAL, // Set to the max posable value as we want to let the broker trade as much as needed in this example.
-      MAX_UINT_VAL,
+      [tokenA.address, tokenB.address],
+      ["1000", "1"], // The "true" price of the pair is expressed as the ratio of token A to token B. A price of 1000 is simply 1000/1.
```
mrice32, line 226: Unrelated to this line (and definitely not something we should do in this PR), but how hard do you think it would be to deploy a uniswap pair for this test and using it to vet the calculation?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -566,6 +566,17 @@ async function createUniswapPriceFeedForFinancialContract(
...
+  // Check if there is an override for the getTime method in the price feed config. Specifically, we can replace the
+  // get time method with the current block time.
+  if (userConfig.getTimeOverride) {
+    if (userConfig.getTimeOverride.useBlockTime) {
+      getTime = async () =>
+        web3.eth.getBlock("latest").then(block => {
+          return block.timestamp;
+        });
+    }
+  }
```
mrice32, line 578:
```suggestion
  if (userConfig.getTimeOverride?.useBlockTime) {
    getTime = () => web3.eth.getBlock("latest").then(block => block.timestamp);
  }
```

Or

```js
  if (userConfig.getTimeOverride?.useBlockTime) {
    getTime = async () => (await web3.eth.getBlock("latest")).timestamp;
  }
```

mrice32 (commented):
Flushing some comments -- still working through

packages/trader/src/index.ts
```diff
@@ -1,33 +1,136 @@
...
+  delay,
+  DSProxyManager
+} = require("@uma/financial-templates-lib");
+
+const { RangeTrader } = require("./RangeTrader");
+const { createExchangeAdapter } = require("./exchange-adapters/CreateExchangeAdapter");
 import { TraderConfig } from "./TraderConfig";
+config();
+
+export async function run(logger: any, web3: any): Promise<void> {
```
mrice32, line 22: logger might be another thing that we could type explicitly -- winston almost certainly has a logger type that we can pull in here.

mrice32 (commented):
Looks great! A few comments. I agree with your initial assessment that ts will require some getting used to. We'll definitely need to establish some new patterns, especially around interfaces, classes, and inheritance. I _do_ think that our object-oriented approach is much better supported by typescript than javascript, so I'm hopeful that it will allow us to keep our current approach at least at a high level.

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
...
+
+  // TODO: is there any way to not have to declare these as part of the type?
+  readonly BN: any;
+  readonly toBN: any;
+  readonly toWei: any;
+  readonly fromWei: any;
+
+  readonly tradeExecutionThreshold: any;
+  readonly targetPriceSpread: any;
+  readonly fixedPointAdjustment: any;
```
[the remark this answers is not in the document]
chrismaree: this means no longer creating a class for the `RangeTrader` and just baking everything into one nested function? that's a really interesting pattern, I have not seen that before. What are the implications seeing we no longer have a "class" object. I mean, it amounts to effectively the same thing I guess but what are the tradeoffs with this pattern?

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
+const { ConvertDecimals, createFormatFunction, createObjectFromDefaultProps } = require("@uma/common");
+import assert from "assert";
+
+export class RangeTrader {
+  readonly logger: any;
+  readonly web3: any;
```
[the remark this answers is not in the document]
chrismaree: awesome! I was able to simply get the types of Winston and web3 by importing 
```
import Web3 from "web3";
```
then using
```
readonly logger: winston.Logger;
readonly web3: Web3;
```

Would we want to define a common file that contains all the types that we would need and then import that into typescript files? this way we won't need to define the extra imports within each file. What is the best practice with this?

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
...
+      ...commonLogObject,
+      postTradeSpotPrice: this.formatDecimalString(exchangeSpotPriceAfterTrade),
+      postTradePriceDeviationError: this.formatDecimalString(postTradePriceDeviationError.muln(100)) + "%",
+      tx: tradeExecutionTransaction.transactionHash
+    });
+  }
+
+  // TODO: replace the any type with bignumber types. I'm not exactly sure what the best practice is to do this in typescript.
+  // TODO: this method was taken from the SyntheticPegMonitor verbatim. Ideally it should be refactored into a common utility that both can use.
+  _calculateDeviationError(observedValue: any, expectedValue: any) {
```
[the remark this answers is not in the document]
chrismaree: cool, that worked after importing `import BigNumber from "bignumber.js";`

packages/trader/src/TraderConfig.ts
```diff
@@ -2,15 +2,40 @@ import Web3 from "web3";
 import assert from "assert";
 
 export interface ProcessEnv {
-  [key: string]: string | undefined;
+  [key: string]: any;
```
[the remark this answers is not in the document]
chrismaree: I got this error when it was set to `string | undefined`
image

chrismaree:
https://github.com/UMAprotocol/protocol/pull/2579#discussion_r582292518

Not sure why I could not reply to this specific comment @daywiss  but I linked it above ^^

This would look something like this, right?
```
interface ExchangeAdapterInterface {
  // Take in a desired price and execute the trades required to move the market from the current price to desiredPrice.
  tradeMarketToDesiredPrice: (desiredPrice: BigNumber) => BigNumber;

  // Returns the current spot price within the exchange.
  getExchangeSpotPrice: BigNumber;
}
```

My issue that I had with this is that does this enforce that the function is implemented the same way we do with the `_abstractFunctionCalled` method?

packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
```diff
@@ -23,57 +23,56 @@ contract UniswapBroker {
...
 
-        address tokenIn = aToB ? tokenA : tokenB;
-        address tokenOut = aToB ? tokenB : tokenA;
-        TransferHelper.safeTransferFrom(tokenIn, msg.sender, address(this), amountIn);
+        address tokenIn = aToB ? swappedTokens[0] : swappedTokens[1];
+        address tokenOut = aToB ? swappedTokens[1] : swappedTokens[0];
+
         TransferHelper.safeApprove(tokenIn, address(router), amountIn);
 
+        if (tradingAsEOA) TransferHelper.safeTransferFrom(tokenIn, msg.sender, address(this), amountIn);
```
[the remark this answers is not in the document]
chrismaree: that's correct. Only if it's trading from an EOA do we need to move the tokens into this contract. if it's called from a DSProxy then the context of `this` becomes the DSProxy and the logic within the `UniswapBroker` is effectively executed _as_ the DSProxy.

packages/core/test/trader/uniswap-broker/UniswapBroker.js
chrismaree: Take a look at the UniswapBroker tests in core ([here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/trader/uniswap-broker/UniswapBroker.js)). They deploy a full uniswap factory & pair and goes through the motions of validating this behaviour. Is this sufficient?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
chrismaree: nice! I updated it as well as the previous instance of this override logic.

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
...
+   * @param {Object} logger Module used to send logs.
+   * @param {Object} web3 Provider from Truffle/node  to connect to Ethereum network.
+   * @param {Object} tokenPriceFeed Price feed to fetch the current synthetic token trading price. EG a Dex price feed.
+   * @param {Object} referencePriceFeed Price feed to fetch the "real" identifier price. EG a Cryptowatch price feed.
+   * @param {Object} exchangeAdapter Interface to interact with on-chain exchange. EG: Uniswap.
+   * @param {Object} rangeTraderConfig: Config to parameterize the range trader. Expected:
+   *      { tradeExecutionThreshold: 0.2,  -> error amount which must be exceeded for a correcting trade to be executed.
+            targetPriceSpread: 0.05 }      -> target price that should be present after a correcting trade has concluded.
+   */
+    logger: any,
```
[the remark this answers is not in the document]
chrismaree: If I dont add it to the declaration then I cant set things like `this.logger` and access the logger later on. I get the following error:
image


If I remove the `this` syntax then I cant access the `logger` as it's not correctly scoped. What is the correct way to deal with this?

packages/trader/src/exchange-adapters/UniswapTrader.ts
```diff
@@ -0,0 +1,92 @@
...
+    const reserves = await this.uniswapPair.getReserves();
+
+    return reserves.reserve1.mul(this.web3.utils.toBN(this.web3.utils.toWei("1"))).div(reserves.reserve0);
+  }
+
+  // TODO: This method was pulled from the uniswapBroker tests. it should be refactored to work with generic implementations.
+  // potentially the getTruffleContract method should be modified to enable creation of truffle contracts from json
+  // as a fallback keeping the getter interface generic.
+  createContractObjectFromJson(contractJsonObject: any) {
+    const contract = require("@truffle/contract");
```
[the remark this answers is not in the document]
chrismaree: ye, there was a reason. This method originally came from the `UniswapBroker` tests in core. There, the test would simply not run if this import was at the top. However, it worked fine when it was in the method. I _think_ it's due to a scoping issue. if it's imported at the top then it overrides the global `contract` import that is injected into the tests. However, if it is imported within the method then the scoping is correct.

I just tried re-naming this import from `contract to `truffleContract` within the `UniswapBroker` tests & this fixed the no running test issue, which confirms my original suspicion around scoping. I've implemented the same pattern here within the `UniswapTrader`.

packages/trader/src/index.ts
chrismaree: indeed it does! I've imported from `Winston.Logger`

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
daywiss: yea thats a good question, im not sure what the best practice is here. I'm ok either importing them in every file or creating a top level types file.

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
daywiss: Yes the drawback is there are not really any classes, though with typescript you can create interfaces to check this.  This class though in particular is not meant to really be extended or changed, so I dont think you lose much in this case. Ultimately this is more of a style thing and a way to get around using `this` and having all this extra ceremony around storing class variables and setting visibility, functionally is 99% equivalent.  It pains me to see so much extra code that could be avoided.

packages/trader/src/TraderConfig.ts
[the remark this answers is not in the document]
daywiss: oh i see yea, you probably can get around this by either throwing an error if `DS_PROXY_FACTORY_ADDRESS` is undefined, or setting a default string like calling with `DS_PROXY_FACTORY_ADDRESS || ''`

one thing that is really annoying about typescript is i dont think it will detect `assert` as properly gaurding against null checks. You may have to explicitly use `if(DS_PROXY_FACTORY_ADDRESS == null) throw new Error('blah')`

daywiss:
> My issue that I had with this is that does this enforce that the function is implemented the same way we do with the _abstractFunctionCalled method?

Its true, you are trading a run time check for a compile time check though, so it still will be enforced (theoretically). Typescript takes the stance that the compile time checks are superior to runtime checks 😄

mrice32:
> ```
>   getExchangeSpotPrice: BigNumber;
> ```

Shouldn't this be a function?
Something like:
```ts
getExchangeSpotPrice: () => BigNumber;
```

or maybe:
```ts
getExchangeSpotPrice: (void) => BigNumber;
```

Not 100% sure how to type no-arg functions.

> My issue that I had with this is that does this enforce that the function is implemented the same way we do with the `_abstractFunctionCalled` method?

Shouldn't it enforce that? My understanding is that if a class inherits from this interface (and it's instantiated), it _must_ have the properties in the interface, but maybe I'm misunderstanding?

packages/trader/src/exchange-adapters/UniswapTrader.ts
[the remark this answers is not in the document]
mrice32: Ah, makes sense. Sounds good!

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
mrice32: In your snippet above, you seem to be missing `readonly`.
```ts
// Won't compile.
class RangeTrader {
  constructor(/* no readonly here means the logger is just a normal var*/ logger: any) {}
}
```

That snippet _will not_ declare logger automatically as a class var.

_This_ snippet will, though:
```ts
// Will compile.
class RangeTrader {
  constructor(/* readonly makes it work! */ readonly logger: any) {}
}

File: packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol (hunk 22 above)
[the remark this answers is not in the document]
Comment (mrice32) [reply packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol:74]: Is there a similar transfer _out_ that needs to be enabled/disabled?

File: packages/trader/src/TraderConfig.ts (hunk 21 above)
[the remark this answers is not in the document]
Comment (mrice32) [reply packages/trader/src/TraderConfig.ts]: ^+1. This is just typescript forcing you to "narrow" the type (aka create a logic branch for string and for undefined) before you use it where a string is expected.

File: packages/trader/src/RangeTrader.ts (hunk 18 above)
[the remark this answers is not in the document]
Comment (mrice32) [reply packages/trader/src/RangeTrader.ts:25]: Yeah, it's a style thing, and I totally get the anti-verbosity argument.

I _personally_ don't mind (and even prefer) the verbosity of explicitly declaring instance/class variables. @daywiss, you and I have had really good discussions on this before, but I think using classes makes the _scoping_ of data very clear via declarations, especially in typescript where they are declared up-front. I feel like closures kind of mask scoping, and they make object lifetimes and data availability a little more difficult to reason about.

One other big advantage of classes IMO is tooling. Classes are a much more common pattern and my guess is that typescript type inference, IDE tooling, etc _just works better_ with classes than they do with this closure-class pattern.

The last point I'll make in favor of classes is that variable declaration is a little verbose up-front, but it's really not _that_ bad. My guess is that VSCode has some feature where it can add them for you.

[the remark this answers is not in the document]
Comment (chrismaree) [reply packages/trader/src/RangeTrader.ts:25]: thanks for the input above guys ^^. I'd say that for now we leave it as a class (even though it is ugly and verbose, with you on that @daywiss ) and then we can converge on a _common_ styling pattern in the future. we can always refactor this later if we choose but for now consistency is king 👑

File: packages/trader/src/RangeTrader.ts (hunk 14 above)
Comment (chrismaree) [reply packages/trader/src/RangeTrader.ts:88]: not strictly; we could put the updates within the parent class. I've done just this. it's a bit cleaner.

File: packages/trader/src/exchange-adapters/ExchangeAdapterInterface.ts (hunk 8 above)
Comment (chrismaree) [reply packages/trader/src/exchange-adapters/ExchangeAdapterInterface.ts:1]: ok cool! I've gotten this to work. I export the interface as:
```
export default interface ExchangeAdapterInterface {
```

Then import it and use it as:
```
class UniswapTrader implements InstanceType<typeof ExchangeAdapterInterface> {
```

File: packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol (hunk 22 above)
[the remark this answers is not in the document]
Comment (chrismaree) [reply packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol:74]: right now the contract should not be left with any extra tokens are a trade; it pulls the exact amount needed for the trade and sends everything left over to the calling EOA.

This does open up a separate meta point that I think we can leave for a bit later: we'll want some more general DSProxy methods for doing things like:
- deposit/withdraw/ transfer tokens from the DSProxy
- wrapper methods for dealing with EMPs/perps
- admin methods for adding other admins etc

File: packages/trader/src/RangeTrader.ts (hunk 23 above)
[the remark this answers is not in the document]
Comment (chrismaree) [reply packages/trader/src/RangeTrader.ts]: OK great! this is a _much_ cleaner pattern. I've implemented this change through and was able to remove a bunch of state variables in the process.

File: packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts [hunk 25]
```diff
@@ -0,0 +1,50 @@
...
+async function createExchangeAdapter(logger: winston.Logger, web3: Web3, dsProxyManager: any, config: any) {
+  assert(config.type, "Exchange adapter must have a type. EG uniswap for a uniswap dex");
+
+  if (config.type === "uniswap") {
+    const requiredFields = ["tokenAAddress", "tokenBAddress"];
+    if (isMissingField(config, requiredFields, logger)) return null;
+
+    // TODO: refactor these to be pulled from a constants file somewhere.
+    const uniswapRouterAddress = "0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D";
+    const uniswapFactoryAddress = "0x5C69bEe701ef814a2B6a3EDD4B1652CB9cc5aA6f";
```
mrice32, line 16: Could we use the uniswap SDK to get these for whichever network we're using?

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
```diff
@@ -0,0 +1,50 @@
...
+      config.tokenAAddress,
+      config.tokenBAddress,
+      dsProxyManager
+    );
+  }
+  return null;
+}
+
+// TODO: this method was taken verbatim from the create price feed class. it should be refactored to a common util.
+function isMissingField(config: any, requiredFields: Array<string>, logger: any) {
```
mrice32, line 33:
```suggestion
function isMissingField(config: object, requiredFields: Array<string>, logger: winston.Logger) {
```

mrice32 (commented):
Looks great! Mostly minor ts-related comments. I think we should try to make types as specific as possible (even if it just means `object`. I may have been overly aggressive in some of my suggestions, so I'm totally cool swapping back to any if it's too difficult to get it to compile otherwise.

.eslintrc.js
```diff
@@ -25,7 +25,10 @@ module.exports = {
         "eslint:recommended",
         "plugin:@typescript-eslint/eslint-recommended",
         "plugin:@typescript-eslint/recommended"
-      ]
+      ],
+      rules: {
+        "@typescript-eslint/no-var-requires": 0
```
nicholaspai, line 30: What's this rule?

packages/financial-templates-lib/index.js
nicholaspai: Weird that this wasn't exported previously...

packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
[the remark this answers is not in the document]
chrismaree: the `to` parameter ensures this. If an EOA is calling this then they can set the `to` to be their address. This will send the output of the trade to the EOA. In the DSProxy case the caller must set the `to` to be the DSProxy wallet. If this is set then the DSProxy will hold the tokens after the trade.

I've updated the comments in the `UniswapTrader` to reflect this fully.

packages/core/test/trader/uniswap-broker/UniswapBroker.js
chrismaree: these tests _are_ using a real thing though. The `RangeTrader` tests deploy a `uniswapFactory` and create a `pair` from it. This dev is then seeded with an amount of tokens to create the initial price. The tests also use the normal `UniswapPriceFeed`, pointed at this "real" uniswap contract.

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,183 @@
...
+      exchangeSpotPriceAfterTrade,
+      currentReferencePrice
+    );
+
+    this.logger.info({
+      at: "RangeTrader",
+      message: "The exchange adapter has executed a trade successfully 👉 👈",
+      ...commonLogObject,
+      postTradeSpotPrice: this.formatDecimalString(exchangeSpotPriceAfterTrade),
+      postTradePriceDeviationError: this.formatDecimalString(postTradePriceDeviationError.muln(100)) + "%",
```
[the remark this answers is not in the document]
chrismaree: I actually replied to this; not sure why it was lost. I originally added this param for this specific reason where we could generate an error if it exceeds some threshold. However, I was not really sure how to test it as the smart contract always trades within 0.1% of the threshold. I guess we could miss-paramaterize the contract to detect a deviation error? 

A simple heuristic could be to generate an error if the `postTradePriceDeviationError` is larger than `targetPriceSpread` or do you think it should be a separate param?

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,177 @@
...
+      tx: tradeExecutionTransaction.transactionHash
+    });
+  }
+
+  // TODO: replace the any type with bignumber types. I'm not exactly sure what the best practice is to do this in typescript.
```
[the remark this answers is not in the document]
chrismaree: yup!

packages/trader/src/exchange-adapters/UniswapTrader.ts
```diff
@@ -0,0 +1,87 @@
...
+const { MAX_UINT_VAL } = require("@uma/common");
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
+
+const IUniswapV2Factory = require("@uniswap/v2-core/build/IUniswapV2Factory.json");
+const IUniswapV2Pair = require("@uniswap/v2-core/build/IUniswapV2Pair.json");
+
+class UniswapTrader implements InstanceType<typeof ExchangeAdapterInterface> {
+  readonly tradeDeadline: any;
+  readonly UniswapBroker: any;
```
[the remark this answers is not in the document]
chrismaree: UniswapBroker is a contract. How exactly would you declare the type for that? I tried something like 
```
Web3.eth.Contract
```
but this did not work. any ideas?

packages/trader/src/exchange-adapters/UniswapTrader.ts
```diff
@@ -0,0 +1,87 @@
...
+  uniswapPair: any;
+
+  constructor(
+    readonly logger: winston.Logger,
+    readonly web3: Web3,
+    readonly uniswapRouterAddress: string,
+    readonly uniswapFactoryAddress: string,
+    readonly tokenAAddress: string,
+    readonly tokenBAddress: string,
+    readonly dsProxyManager: any
```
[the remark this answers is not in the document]
chrismaree: this does not work without a type interface (unless I'm doing it wrong). I get this if it's an object
image

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,177 @@
...
+   * @param {Object} referencePriceFeed Price feed to fetch the "real" identifier price. EG a Cryptowatch price feed.
+   * @param {Object} exchangeAdapter Interface to interact with on-chain exchange. EG: Uniswap.
+   * @param {Object} rangeTraderConfig: Config to parameterize the range trader. Expected:
+   *      { tradeExecutionThreshold: 0.2,  -> error amount which must be exceeded for a correcting trade to be executed.
+            targetPriceSpread: 0.05 }      -> target price that should be present after a correcting trade has concluded.
+   */
+    readonly logger: winston.Logger,
+    readonly web3: Web3,
+    readonly tokenPriceFeed: any,
+    readonly referencePriceFeed: any,
```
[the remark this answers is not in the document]
chrismaree: same as before, I cant call methods on this if it's an object
image

Again, I might be doing something wrong, IDK.

packages/trader/src/exchange-adapters/UniswapTrader.ts
```diff
@@ -0,0 +1,87 @@
...
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
+
+const IUniswapV2Factory = require("@uniswap/v2-core/build/IUniswapV2Factory.json");
+const IUniswapV2Pair = require("@uniswap/v2-core/build/IUniswapV2Pair.json");
+
+class UniswapTrader implements InstanceType<typeof ExchangeAdapterInterface> {
+  readonly tradeDeadline: any;
+  readonly UniswapBroker: any;
+  uniswapPair: any;
```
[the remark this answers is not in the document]
chrismaree: this is similar issue to the `UniswapBroker`. you cant set it to an object  without defining the interface. it cant be undefined or it shows errors later down the list. Perhaps there needs to be a way to define the interface from a contract instance?

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,177 @@
...
+import assert from "assert";
+
+export class RangeTrader {
+  readonly normalizePriceFeedDecimals: any;
+  readonly formatDecimalString: any;
+
+  readonly BN: any;
+  readonly toBN: any;
+  readonly toWei: any;
+  readonly fromWei: any;
```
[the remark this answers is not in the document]
chrismaree: I really like this pattern but could not get it to work :/ I was however able to do this fine which is still measurably cleaner:
```
import Web3 from "web3";
const { toWei, toBN } = Web3.utils;
```

I also added a new function:
```
const toBNWei = (number: string | number) => toBN(toWei(number.toString()).toString());
```
to clean up some of the conversions through. this is worse than normal JS as typescript gets upset with things like
```
toBN(toWei(numberThatIsAString))
```

as apparently `toWei` can return a Number so you end up needing to do horrible things like:
```
toBN(toWei(yourNumber.toString()).toString()
```

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,177 @@
...
+   * @param {Object} rangeTraderConfig: Config to parameterize the range trader. Expected:
+   *      { tradeExecutionThreshold: 0.2,  -> error amount which must be exceeded for a correcting trade to be executed.
+            targetPriceSpread: 0.05 }      -> target price that should be present after a correcting trade has concluded.
+   */
+    readonly logger: winston.Logger,
+    readonly web3: Web3,
+    readonly tokenPriceFeed: any,
+    readonly referencePriceFeed: any,
+    readonly exchangeAdapter: typeof ExchangeAdapterInterface,
+    readonly rangeTraderConfig: any
```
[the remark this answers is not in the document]
chrismaree: > I tried this out and made _some_ progress. first off, I was able to define the type of the `rangeTraderConfig`. 

Secondly, I did not implement your `type DefaultConfig<T> =` as I'm not actually sure what this will enforce. The object params cant be anything other than a `number` as the type in the param enforces this. Note that this was actually the case previous as the `defaultConfig` enforced hat the `isValid` method was a `number` for both params. What will your type definition above also prevent?

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
chrismaree: `object` on it's own did not work. `(config: { [key: string]: string }...)` did though

.eslintrc.js
chrismaree: this lets us use `require` for imports. it's needed for compatibility with our existing JS code.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
chrismaree: I took a look at this but the SDK only has the `FACTORY_ADDRESS` on main net: https://uniswap.org/docs/v2/SDK/other-exports/

packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
nicholaspai: Does 1 impact gas at all?

packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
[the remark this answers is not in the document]
nicholaspai: Instead of the `tradingAsEOA` step, why not just allow this contract to be withdrawn from? The downside is that it would have to be Ownable however, or store balances for account

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,169 @@
...
+    readonly web3: Web3,
+    readonly tokenPriceFeed: any,
+    readonly referencePriceFeed: any,
+    readonly exchangeAdapter: typeof ExchangeAdapterInterface,
+    readonly rangeTraderConfig: {
+      tradeExecutionThreshold: number;
+      targetPriceSpread: number;
+    }
+  ) {
+    assert(tokenPriceFeed.priceFeedDecimals === referencePriceFeed.priceFeedDecimals, "pricefeed decimals must match");
```
nicholaspai, line 41: I think you should be calling `getPriceFeedDecimals()` instead of reading directly from the class state

packages/trader/src/RangeTrader.ts
```diff
@@ -0,0 +1,169 @@
...
+
+    this.logger = logger;
+    this.web3 = web3;
+    this.tokenPriceFeed = tokenPriceFeed;
+    this.referencePriceFeed = referencePriceFeed;
+    this.exchangeAdapter = exchangeAdapter;
+
+    this.normalizePriceFeedDecimals = ConvertDecimals(tokenPriceFeed.priceFeedDecimals, 18, this.web3);
```
nicholaspai, line 49: Same here

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
mrice32: Yeah, this may not be the right solution. I think defining an interface makes sense to start and I will think more about generic common functions down the road.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
mrice32: Hmmmm, weird

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
mrice32: I see. I'm kinda surprised they don't export that 🤷

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
mrice32: haha, gross.

I actually think this is the really nice thing about typescript. We've had so many difficult-to-diagnose errors because of the input/output types of the web3 functions. I think ts will finally force us to handle these correctly rather than waiting for runtime errors. It's a little messier, but it's more fool-proof :)

packages/trader/src/RangeTrader.ts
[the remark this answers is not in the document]
mrice32: No you're probably right. Typescript may classify an `object` type as a type where _no_ properties are accessible. Seems kinda useless, but 🤷 .

Leaving these as is makes sense for now, and we'll be more specific once we convert financial-templates-lib.

packages/core/contracts/trader/uniswap-broker/UniswapBroker.sol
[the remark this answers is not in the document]
mrice32: @chrismaree that makes sense! Thanks for the explanation!

mrice32 (approved):
LGTM!

Thanks for working through this. I know this was a pretty brutal PR.

PR #2581, opened 2021-02-22
improve(proposer): Send alert when price request detected, improve scripts to enhance OO bot testing

nicholaspai:

**Motivation**

Proposal bot changes:
- Proposer bot now emits an ERROR level log when a price request is detected.
- `transactionConfig` is removed from logs because it does not explicitly set the `gas` param, which causes the SlackTransporter to error out when it attempts to compute `.length` on `{ ...txnConfig, gas: undefined }`. Its a bit esoteric but I see no real drawback to removing this config from the logs. The other option is to explicitly set `gas` to the result of an `estimateGas` call

Script changes:
- (DeployEMP.js) Builds off of #2531 and replaces the "DeployEMP.js` diff from #2531
- (DeployEMP.js): New EMP's have slightly different constructor arguments that this script needs to handle.
- (SettleTokens.js): New script to easily settle tokens at expiry for an `emp`
- (ProposePriceEMP.js): New script to manually propose a price to the OO for an `emp`. Useful to test bot dispute logic

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/optimistic-oracle/src/proposer.js
+++ b/packages/optimistic-oracle/src/proposer.js
@@ -164,7 +164,8 @@ class OptimisticOracleProposer {
       at: "OptimisticOracleProposer#sendProposals",
       message: "Detected price request, and proposing new price",
       priceRequest,
-      potentialProposalPrice: proposalPrice
+      potentialProposalPrice: proposalPrice,
+      proposer: this.account
     });
     try {
       const transactionResult = await runTransaction({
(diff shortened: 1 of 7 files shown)


pemulis (approved):
LGTM! Simpler than the earlier version of this script.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -195,7 +193,6 @@ class OptimisticOracleProposer {
         priceRequest,
         proposalBond: returnValue,
         proposalPrice,
-        transactionConfig,
```
nicholaspai, line 198: Removed from logs because `transactionConfig` is missing a `gas` param (which we purposefully omit to allow Truffle to estimate gas) which causes issues for SlackTransport. SlackTransport sees `gas: undefined` in the `transactionConfig` object. Here thrown [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/logger/SlackTransport.js#L78)

PR #2586, opened 2021-02-23
improve(voter-dapp): add new addresses to dapp with as few changes as possible

mrice32:
**Motivation**

We'd like to temporarily patch the voter dapp to be able to interact with the new voting contract without updating any abis for ancillary data. This will avoid the problem of requiring a 2-key contract migration until we do the full voter dapp upgrade.

**Summary**

I just overrode the voting addresses that the voter dapp uses.

EDIT: this required some ABI changes to make the dapp work as expected. I modified the test script and some of the versioning in the dapp (as few changes as I could to make it work). A lot of this custom versioning is rather hacky and once we get a full-fledged update, we should roll this PR back.

**Testing**

- [ ]  New unit tests created
- [x]  Untested (will be running tests)

**Issue(s)**

N/A


daywiss (approved):

packages/voter-dapp/run_tests.sh
```diff
@@ -34,6 +34,17 @@ echo '#                                                                  #'
 echo '# 1/16. Deploying voting contracts on test network                 #'
 echo '#                                                                  #'
 echo -e '####################################################################\n'
+
+# This hacky snippet just copies the newer voting bytecode into the package artifacts that the voter dapp uses.
+# This forces the voter dapp to deploy the _new_ voting contract with all the other contracts unchanged, which will
+# force it to test the old 2key contract with the new voting contract.
```
nicholaspai, line 40: This is a well done hack and I really like how you backup the old ABI. Perhaps a more stable fix longer term to this test would allow user to enter in contract versions that they want to test (for 2key and voting) and perform backups where appropriate.

packages/voter-dapp/src/Dashboard.js
```diff
@@ -5,7 +5,7 @@ import ActiveRequests from "./ActiveRequests.js";
 import ResolvedRequests from "./ResolvedRequests.js";
 import DesignatedVotingDeployment from "./DesignatedVotingDeployment.js";
 import DesignatedVotingTransfer from "./DesignatedVotingTransfer.js";
-import DesignatedVoting from "@uma/core/build/contracts/DesignatedVoting.json";
+import DesignatedVoting from "@uma/core-1-2-0/build/contracts/DesignatedVoting.json";
```
nicholaspai, line 8: Maybe Add todo to revert this?

packages/voter-dapp/src/drizzleOptions.js
```diff
@@ -1,7 +1,13 @@
-import DesignatedVotingFactory from "@uma/core/build/contracts/DesignatedVotingFactory.json";
-import Governor from "@uma/core/build/contracts/Governor.json";
-import Voting from "@uma/core/build/contracts/Voting.json";
-import VotingToken from "@uma/core/build/contracts/VotingToken.json";
+import DesignatedVotingFactory from "@uma/core-1-2-0/build/contracts/DesignatedVotingFactory.json";
```
nicholaspai, line 1: Why can't you use the getAbi(contract, versionString) method here?

nicholaspai (commented):
Hacky but really cleanly written and commented. Left some comments that are mostly requests for clarification

packages/voter-dapp/run_tests.sh
mrice32: Agreed, that would be awesome. When we do a full dapp upgrade to the new contracts, we should roll this hacky change back in its entirety. I don't want this code to last more than a week :)

packages/voter-dapp/src/drizzleOptions.js
mrice32: Not sure any of those methods work in the browser since they require `fs` calls which I think are only available in node.

packages/voter-dapp/src/Dashboard.js
mrice32: Will do!

mrice32: Done.

chrismaree (approved):
This looks great to me!

I spent some time this verifying the voter dapp works with your hot fix. I was able to go through 2 full voting cycles including commit, reveal, snapshot and claim rewards on the voterdapp without the 2key and 1 cycle with the 2 key. I’m pretty sure it’s all working as expected.

image

PR #2590, opened 2021-02-23
refactor(create-price-feeds)!: Each PriceFeed renames apiKey config param

### LINKED ISSUE #2588: PriceFeeds: `apiKey` is reused for different pricefeeds but each one should use its own unique name for this key
**Why**

Allows easier config setting by the user

nicholaspai:

**Motivation**

#2588

**Summary**

When this PR is merged, bot configs need to be changed that are using `apiKey` as a config param.

**Issue(s)**

Fixes #2588

### DIFF
--- a/packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
@@ -323,7 +323,7 @@ contract("CreatePriceFeed.js", function(accounts) {
   it("Valid CryptoWatch config", async function() {
     const config = {
       type: "cryptowatch",
-      apiKey,
+      cryptowatchApiKey: apiKey,
       exchange,
       pair,
       lookback,
@@ -359,7 +359,7 @@ contract("CreatePriceFeed.js", function(accounts) {
   it("Invalid CryptoWatch config", async function() {
     const validConfig = {
       type: "cryptowatch",
-      apiKey,
+      cryptowatchApiKey: apiKey,
       exchange,
       pair,
       lookback,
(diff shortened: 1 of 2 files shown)


packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -323,7 +323,7 @@ contract("CreatePriceFeed.js", function(accounts) {
   it("Valid CryptoWatch config", async function() {
     const config = {
       type: "cryptowatch",
-      apiKey,
+      cryptowatchApiKey: apiKey,
```
mrice32, line 326: Are you worried that this change (and the others for existing feeds in this PR) will break any 3rd party production bots? If we don't want to do a soft deprecation (i.e. take `apiKey`, but prefer `cryptowatchApiKey` if provided), then we should probably change the title to (notice the `!`):
```
refactor(create-price-feeds)!: Each PriceFeed renames apiKey config param
```

This will trigger a major version bump to denote the breaking change.

mrice32 (commented):
Looks good. Minor comment on versioning/deprecation.

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
nicholaspai: Good point!

nicholaspai: This is a really neat feature that it triggers a major version bump

mrice32 (approved):

PR #2501, opened 2021-02-02
improve(pricefeed): Add getHistoricalPricePeriods to BasketSpread and Uniswap

nicholaspai:
Signed-off-by: Nick Pai <npai.nyc@gmail.com>

**Motivation**

Part of #2461

**Summary**

Implements `getHistoricalPricePeriods() => [time, price]` for Uniswap and BasketSpread price feeds.

**Testing**

TODO: Does this work for `affiliates`?

- [x]  Ran end-to-end test, running the code as in production
- [ ]  New unit tests created
- [ ]  Existing tests adequate, no new tests required
- [ ]  All existing tests pass
- [x]  Untested

PR #2595, opened 2021-02-23
feat(affiliates): add composable dev/dapp mining apps for future automation

### LINKED ISSUE #2519: Implement some of the proposals for payout scripts to automate dev mining

daywiss:
**Motivation**

#2519

**Summary**

This adds a bunch of new "apps" which can be composed to create an automation pipeline. They can take input as a combination from standard in as json strings, env vars and a config file. These can be piped together using unix pipes to compose input and output. The idea is that they can be run independently if needed to verify parts of the pipeline. 

**Details**

For review purposes, you can ignore changes in the `scripts` dir, as these were just prototypes to get to this functionality. Production code changes are in `apps`, `libs/affiliates` and `tests/mocha`.

An example of how to run a dev mining pipeline:
` node apps/GetSheetsDevminingStatus.js | node apps/GenerateDevMiningConfig.js |  node apps/OpenDevMiningIssue.js | node apps/DevMiningRewards.js --network-mainnet_mnemonic  | node apps/DevMiningSaveResult`

This essentially pulls a whitelist from the google sheet, passes it into a config generator to generate proper config params to run the rewards. A github issue is opened with this information using a template, then params are passed to run the actual dev mining rewards. The output of this gets passed to a file saver, which formats the file as well as the filename and saves it to disk. 

If you could pick apart this pipeline, as a common use case would just be to verify dev mining. Assuming you had a config file you can run it like this:
`cat config.json | node apps/DevMiningRewards.js --network=mainnet_mnemonic` 

which will just output the results to console for you to compare.

If you just want to see what is currently whitelisted you can just run:

`node apps/GetSheetsDevminingStatus.js` (requires .env file)

Similarly you can compose a Dapp mining pipeline this way, though it is slightly different. 

The main point of this pattern is that each successive app only adds to an object that gets continually passed through to the next. The next app knows it can pull fields off an object, and add its own output parameters to it. Theres also an assumption that each app should add a unique key, and for now that requires some subjectivity.

Theres much more that needs to be done, including comments and more extensive tests, but most of this has been tested in production.  I will be working with @evaldofelipe to start automating this work. 

This also depends on #2501 going in.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2519


packages/affiliates/apps/DappMiningRewards.js
```diff
@@ -18,15 +20,12 @@ async function App(config) {
...
+    result
   };
-}
+};
 
-const config = Config();
-
-App(config)
-  .then(x => console.log(JSON.stringify(x, null, 2)))
+makeUnixPipe(App)
```
daywiss, line 28: this got converted to take its config from standard in. it expects a json object which is not documented yet.

packages/affiliates/apps/DevMiningRewards.js
```diff
@@ -81,16 +85,13 @@ async function App(config, env) {
...
+    result
   };
-}
-
-const config = Config();
+};
 
-App(config, process.env)
-  .then(x => console.log(JSON.stringify(x, null, 2)))
+makeUnixPipe(App(process.env))
```
daywiss, line 94: again converted from a config file to a json blob piped through standard in

packages/affiliates/apps/MultiDappMiningRewards.js
```diff
@@ -0,0 +1,44 @@
+const lodash = require("lodash");
```
daywiss, line 1: example of a pipeline that is all built into a single script. it is kind of nice but theres no way to call each stage in the pipeline independently without editing code, which is really annoying if you just want to generate the config file, or verify your github issue template looks correct.

packages/affiliates/apps/OpenDappMiningIssue.js
```diff
@@ -0,0 +1,20 @@
...
+const App = env => async params => {
+  const { config } = params;
+  assert(config, "requires config");
+  const issueTemplate = await dappMiningTemplate(config);
+  const githubIssue = await createGithubIssue({ auth: env.github, ...issueTemplate });
+  return {
+    // data.number is the issue number
+    issueNumber: githubIssue.data.number,
+    issueTemplate,
+    ...params
```
daywiss, line 14: this opens an issue, and also logs out the issue number so it can be consumed by another app. this would typically be needed when creating the PR to reference the issue number.

packages/affiliates/libs/affiliates/utils.js
```diff
@@ -0,0 +1,281 @@
+// collection of functions to compose a automation pipeline
```
daywiss, line 1: this utility file is just a bunch of helper functions that can be mostly unit tested

packages/affiliates/test/mocha/affiliates-utils.js
```diff
@@ -0,0 +1,115 @@
+const { assert } = require("chai");
```
daywiss, line 1: unit tests for the utility file which needs to be expanded on, though most of the functions are tested in production

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.js
```diff
@@ -1,5 +1,6 @@
 const { PriceFeedInterface } = require("./PriceFeedInterface");
 const { parseFixed } = require("@ethersproject/bignumber");
+const assert = require("assert");
```
daywiss, line 3: everything below here can be ignored as they are part of #2501

mrice32:
@daywiss is work still ongoing on this?

daywiss:
@mrice32 yes trying to get this out of draft this week

packages/affiliates/apps/MultiDappMiningRewards.js
```diff
@@ -0,0 +1,44 @@
...
+const App = env => async params => {
+  const web3 = getWeb3();
+
+  const empAbi = getAbi("ExpiringMultiParty");
+  const client = new BigQuery();
+  const queries = Queries({ client });
+
+  const dappmining = DappMining({ empAbi, queries, web3 });
+  params = lodash.castArray(params);
+  await Promise.map(params, async param => {
```
mrice32, line 31: Are you expecting param to be a specific thing? I really like the idea of piping between these different "apps", but I think it would be nice to document what each expects as an input and what it outputs since unlike most unix CLI tools, this looks like it's expecting something a bit more specified than raw text.

As another side note: I think variable names like `param` in general are somewhat opaque. Would be nice to have a more friendly name so it's a little more clear what type of param is expected since I'm assuming there are some requirements.

packages/affiliates/libs/affiliates/utils.js
```diff
@@ -0,0 +1,323 @@
...
+//   payoutAddress,
+//   name,
+//   identifier,
+//   enabled
+// }]
+// And returns it in the whitelist format consumable by dev mining:
+// [
+//   [ address, payoutAddress],
+//   [ address, payoutAddress],
+// ]
```
mrice32, line 164: +1 Until we move everything to ts, this is super helpful!

packages/affiliates/libs/affiliates/utils.js
```diff
@@ -0,0 +1,323 @@
+// collection of functions to compose a automation pipeline with regards to dev/dapp mining.
```
mrice32, line 1: High-level comment: these function feel more contextual than just utils. Not saying we should change the structure, but I could imagine these being split out into more well-defined libraries or class-type objects that have a more refined purpose.

For instance, the config generation functions seem quite specialized and maybe more suited for a special file or class just dedicated to generating configs.

mrice32 (approved):
Looks great! Only minor comments. Really like the idea of using piping this way!!

packages/affiliates/apps/MultiDappMiningRewards.js
nicholaspai: Agreed with matt on opaque naming. Cool use of `Promise.map` I'm going to start replacing my Promise.all's with this.

daywiss: yes this is my bad, params is a bad name and needs to be documented. I have some thoughts around making this more of a common convention to send data between each app to make it more clear, and also be able to throw meaningful errors when expected data is missing. Also yes typescript would be really useful here to both document and enforce these types.  

I will update this with better comments and naming.

packages/affiliates/libs/affiliates/utils.js
daywiss: I agree, though I dont think i want to do this for this PR.  This could be organized better ( though im not sure how yet), i'm using utils as a catch all for these loosely affiliated static functions.

packages/affiliates/apps/DevMiningRewards.js
chrismaree: this is cool.

packages/affiliates/payouts/2021-02-08_2021-02-15.json
```diff
@@ -1,80 +0,0 @@
-{
```
chrismaree, line 1: why was this file removed?

chrismaree (approved):
This is awesome! great work on these scripts @daywiss . I cant wait until these processes are all automated. I really like your piping technique that you've got going here.

packages/affiliates/libs/affiliates/utils.js
mrice32: SG!

packages/affiliates/payouts/2021-02-08_2021-02-15.json
daywiss: good question. i think this was added accidentally in a past PR!

daywiss: actually looking at it now, i think it was committed to the wrong directory, because its not in payouts/devmining where it should be. thanks for noticing this. will move it rather than delete

PR #2598, opened 2021-02-24
feat(merkle-distributor): Initial Merkle Distribution contract + unit tests + merkle root generator script

nicholaspai:

**Motivation**

Based on simplest MerkleDistributor contract made by [Uniswap](https://github.com/Uniswap/merkle-distributor), and borrows features from [Balancer](https://github.com/balancer-labs/erc20-redeemable) and [PieDao](https://github.com/pie-dao/vested-token-migration-app) distribution contracts:
- Stores list of merkle roots so that we can reuse contract for multiple distributions. Each root is inserted at an "index" and associated with a window start + end time + reward currency.
- Because Each merkle root entry is mapped to a window time and a reward token, each distribution can only begin after a start time and is vested until the window end, and can be paid with a specific reward token.
- Admin features to pause distributions, deposit and remove rewards, and reset merkle roots for emergency situations.

**Summary**

Full unit test suite is added

The point of this PR is to get consensus on Smart contract design. Follow up PR will implement off-chain SDK that clients can use to interact with contracts both as a distributor  and claimant.

**Testing**

- [X]  New unit tests created
- []  Untested


pemulis:
Do you need to pull from master to fix the CI issue? cf https://github.com/UMAprotocol/protocol/pull/2603

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+     *
+     * Public functions
+     *
+     ****************************/
+
+    // TODO: This method could get pretty gas intensive; is there a way we can reduce the amount
+    // of external `transfer` calls if we precompute all the rewards for each account?
+    // It's a bit tricky because there are multiple reward currencies possible.
+    function claimWindows(Claim[] memory claims) public nonReentrant() {
+        for (uint256 i = 0; i < claims.length; i++) {
```
daywiss, line 155: is the point of this to like allow someone to claim everyones unclaimed rewards?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+
+        // Fetch how much the account has previously claimed for this window.
+        uint256 amountPreviouslyClaimed = amountClaimedFromWindow[claim.windowIndex][claim.account];
+
+        // The account can claim their maximum vested amount minus any previous claim amounts.
+        amountVestedNetPreviousClaims = amountVested.sub(amountPreviouslyClaimed);
+
+        // The user has now claimed their full vested amount this far into the window.
+        // Update their previously claimed amount.
+        amountClaimedFromWindow[claim.windowIndex][claim.account] = amountVested;
```
daywiss, line 275: this is being set with the amountVested, which means its possible to have additional unclaimed funds if you claimed before expiry? This would make more sense to me if it was set to claim.amount instead, but I could be wrong on this

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+
+    // Claim `amount` of reward tokens for `account`. If `amount` and `account` do not exactly match the values stored
+    // in the merkle proof for this `windowIndex` this method will revert.
+    function claimWindow(Claim memory claim) public nonReentrant() {
+        _claimWindow(claim);
+    }
+
+    // Checks {account, amount} against Merkle root at given window index.
+    function verifyClaim(Claim memory claim) public view returns (bool valid) {
+        bytes32 leaf = keccak256(abi.encodePacked(claim.account, claim.amount));
```
daywiss, line 168: i want to be sure here, but you cannot call claim with an amount other than what was original encoded in the merkle proof? Even though it might make sense to claim less than what you are owed, that is not possible with this?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+    // It's a bit tricky because there are multiple reward currencies possible.
+    function claimWindows(Claim[] memory claims) public nonReentrant() {
+        for (uint256 i = 0; i < claims.length; i++) {
+            _claimWindow(claims[i]);
+        }
+    }
+
+    // Claim `amount` of reward tokens for `account`. If `amount` and `account` do not exactly match the values stored
+    // in the merkle proof for this `windowIndex` this method will revert.
+    function claimWindow(Claim memory claim) public nonReentrant() {
```
daywiss, line 162: how do you prevent someone from claiming for someone else before vesting is over, is this a problem?

daywiss (commented):
still going through this but left some questions

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+    ) external nonReentrant() onlyOwner {
+        uint256 indexToSeed = lastSeededIndex;
+        lastSeededIndex = indexToSeed.add(1);
+
+        _seedWindow(indexToSeed, totalRewardsDistributed, windowStart, windowEnd, rewardToken, merkleRoot);
+    }
+
+    // Flip window lock to opposite of current value. When `window.locked` is true,
+    // claims are blocked for that window.
+    function toggleWindowLock(uint256 windowIndex) external nonReentrant() onlyOwner {
```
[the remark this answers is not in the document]
nicholaspai: Hm maybe I agree. Would there ever be a situation where you'd call `setWindowLock` but NOT want to flip the `lock` bool?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: yea exactly, or claim multiple week's worth of rewards in one go. We might want the dApp to have a "claim all" feature that goes through and checks all of your unclaimed distributions

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: So imagine that `claim.amount > amountVested` (because you're in the middle of the vesting period), then the `amountClaimedFromWindow` should be set to the total amount vested. The assumption we're making is that every call to `claimWindow` claims ALL of your vested tokens, so the `amountClaimedFromWindow` is always set to the maximum vested amount when you call claim.

Does that make sense?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: >i want to be sure here, but you cannot call claim with an amount other than what was original encoded in the merkle proof

Correct

> Even though it might make sense to claim less than what you are owed, that is not possible with this?
This could be possible if we allowed the user to pass in an additional parameter `amountToClaim` to `claimWindow`. They would still have to include the `claim.amount` to verify the `merkleProof`, but they could also specify that they only want to receive `amountToClaim`. Then we can increment `amountClaimedFromWindow` accordingly. Do you think this is a worthwhile feature? @mrice32 @chrismaree 

Personally I'm not sure that partial claims are a useful feature. I'm not even convinced vesting is worth the added complexity. Some of the vesting logic can be repurposed to handle partial claims however.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: You can't prevent it, but I don't see the problems with that. Say person A is owed 100 tokens but wants to wait until they are fully vested to claim all of their tokens, for gas reasons (or tax reasons?). But person B for whatever reason calls `claimWindow` for person A every single second, such that A ends up getting a steady trickle of rewards.

Are there situations where this is a big problem for A?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
daywiss: I see now, yea so you never actually lose rewards by claiming early. This makes a lot of sense and most my comments are around this confusion.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
daywiss: Its more that when you are making the transaction, you know exactly what the result will be without needing to read the state first.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+        uint256 amount;
+        address account;
+        bytes32[] merkleProof;
+    }
+
+    // Windows are mapped to arbitrary indices.
+    mapping(uint256 => Window) public merkleWindows;
+
+    // Keep track of the amount an account has claimed from a window.
+    mapping(uint256 => mapping(address => uint256)) public amountClaimedFromWindow;
```
pemulis, line 59: So this enables vesting but not partial claims?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
daywiss: No im not suggesting partial claims, unecessary complexity. Was just trying to work through the logic to find edge cases

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+     * Admin functions
+     *
+     ****************************/
+
+    // Set merkle root for the next available window index and seed allocations. Callable by owner of this
+    // contract. Importantly, we assume that the owner of this contract
+    // correctly chooses an amount `totalRewardsDistributed` that is sufficient
+    // to cover all claims within the `merkleRoot`. Otherwise, a race condition
+    // can be created either within a window (last claimants within a window don't get their claim)
+    // or across windows (claimants from window T take rewards from window T+1).
```
pemulis, line 100: How would a claimant approved for window T be able to claim extra rewards from window T+1? Wouldn't their claims for window T automatically expire when the window ends?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
daywiss: Thanks yea makes a lot of sense this way

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
pemulis: How many claims do you think this could get through before it runs out of gas?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,285 @@
...
+            return amount;
+        } else {
+            // Assumptions:
+            // - `time - windowStart` > 0
+            // - `windowEnd - windowStart` > 0
+            // Calculate:
+            // - [ amount * (time - start) ] / [ end - start ]
+            // Description:
+            // - Linearly interpolate amount based on time's location between start and end
+            return amount.mul(time.sub(windowStart)) / windowEnd.sub(windowStart);
```
pemulis, line 194: 👍

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: The more I'm thinking about this the more i'm thinking we should allow partial claims. It does add more complexity and adds a couple more test cases, but shouldn't be much more smart contract code. PieDAO's vesting-enabled distributor does something like this where the user specifies both a `_amount` and a `_windowAmount` (equivalent to our claim.amount): https://github.com/pie-dao/vested-token-migration-app/blob/master/contracts/VestedTokenMigration.sol#L71

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Yeah basically you just call `claim()` with some proof parameters, and if the proof succeeds, you get back the maximum that you are owed.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Correct but it also could be used to allow partial claims. See [this discussion](https://github.com/UMAprotocol/protocol/pull/2598#discussion_r583924839) where you'd need to pass an additional `partialAmount` param into the `claimWindow()` method, and the changed logic in the claim method would be to set `amountClaimedFromWindow = min(partialAmount, vestedAmount)` instead of `amountClaimedFromWindow = vestedAmount`

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Actually if the `current time > end timestamp T` then the full amount can be claimed. Rewards never "expire" currently, they become fully vested after the window ends.

This is the code from `calcVestedAmount`

```
        // If time is after window end or window end is not greater than start,then return all tokens.
        else if (time >= windowEnd || windowEnd <= windowStart) {
            return amount;

```

nicholaspai: Actually claimed rewards never expire and after a window ends, the amount just gets fully vested instead of retired. See this code from `calcVestedAmount`:

```
        // If time is after window end or window end is not greater than start,then return all tokens.
        else if (time >= windowEnd || windowEnd <= windowStart) {
            return amount;
```

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Not sure, this is definitely a potential problem and we'll have to add an empirical test to see how many claims we can make in one batch claim. See [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/implementation/Voting.sol#L565) for how we commented on this for the `batchX` methods in `Voting.sol`. We actually empirically tested these transactions in this [script](https://github.com/UMAprotocol/protocol/blob/master/packages/core/scripts/local/BatchGasEstimation.js) (which might be outdated by now).

I'll add a comment for now to test this in the future.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
pemulis: Gotcha, that makes sense. I'm still not sure how claimants from window T could take rewards from window T+1. Can you explain that one?

nicholaspai: This would arise from an undesired race condition. Imagine the following situation:

Window 1: Deposit 100 tokens and inject a Merkle Proof that gives claimant A 50 tokens and claimant B 51 tokens. (This is an error on the contract owner's part for not setting the `totalRewardAmount` of 100 correctly to 101)

Window 2: Deposit 1 token and inject a Merkle Proof that gives claimant A 1 token.

At this point contract has 100 +1 = 101 tokens.

Now, imagine the following sequence of claims:
- Claimant A claims 50 tokens for window 1 (contract now has 101 - 50 = 51 tokens)
- Claimant B claims 51 tokens for window 1 (contract now has 51 - 51 = 0 tokens)
- Claimant A now tries to claim 1 token for window 2 but fails

Thus, the contract Owner allowed "window 1 to steal from window 2"

How could I change the comments to better describe this?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
nicholaspai: That makes sense and is more explicit, I'll make the switch

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
chrismaree: in what scenario would you want partial claims?

nicholaspai: Allows more flexibility to users? Might end up being an unused feature with added complexity.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    // Emergency methods to transfer rewards into and out of the contract
+    // incase the contract was configured improperly.
+    function withdrawRewards(address rewardCurrency, uint256 amount) external nonReentrant() onlyOwner {
+        IERC20(rewardCurrency).safeTransfer(msg.sender, amount);
+    }
+
+    function depositRewards(address rewardCurrency, uint256 amount) external nonReentrant() onlyOwner {
+        IERC20(rewardCurrency).safeTransferFrom(msg.sender, address(this), amount);
```
daywiss, line 141: these probably should have events

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,286 @@
...
+    ) external nonReentrant() onlyOwner {
+        uint256 indexToSeed = lastSeededIndex;
+        lastSeededIndex = indexToSeed.add(1);
+
+        _seedWindow(indexToSeed, totalRewardsDistributed, windowStart, windowEnd, rewardToken, merkleRoot);
+    }
+
+    // When `window.locked` is true, claims are blocked for that window.
+    function setWindowLock(uint256 windowIndex, bool lockValue) external nonReentrant() onlyOwner {
+        merkleWindows[windowIndex].locked = lockValue;
```
daywiss, line 116: this probably should have an event

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Good catch

nicholaspai: In addition to the `Transfer` event that will be emitted?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
pemulis: This is a really clear example. Maybe you could include a stripped-down version of the example as a comment?

nicholaspai: Done

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Done

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Done

nicholaspai:
Have removed the vesting feature which cut down the code and tests a lot

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+
+    // Claim `amount` of reward tokens for `account`. If `amount` and `account` do not exactly match the values stored
+    // in the merkle proof for this `windowIndex` this method will revert.
+    function claimWindow(Claim memory claim) public nonReentrant() {
+        _markClaimed(claim);
+        _disburse(merkleWindows[claim.windowIndex].rewardToken, claim.account, claim.amount);
+    }
+
+    // Checks {account, amount} against Merkle root at given window index.
+    function verifyClaim(Claim memory claim) public view returns (bool valid) {
```
nicholaspai, line 184: Should we keep this function public? I'm not sure if its helpful

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+    // a generalized `setWindowMerkleRoot` whereby the owner can set a merkle root for any index?
+    function resetWindowMerkleRoot(
+        uint256 windowIndex,
+        uint256 totalRewardsDistributed,
+        uint256 windowStart,
+        address rewardToken,
+        bytes32 merkleRoot
+    ) external nonReentrant() onlyOwner {
+        _seedWindow(windowIndex, totalRewardsDistributed, windowStart, rewardToken, merkleRoot);
+    }
```
mrice32, line 140: Do we need this function?

In my mind, it seems bad to be able to overwrite a merkel root after claims have already been submitted. The mapping can't be deleted or overwritten, so in the new merkel root, those who have already claimed would still be marked "claimed". There would be no way to distribute to them in the updated root.

I could see two potential replacements for this functionality:
1. Lock + withdraw.
2. Delete + withdraw.

Curious to hear your thoughts?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+    // incase the contract was configured improperly.
+    function withdrawRewards(address rewardCurrency, uint256 amount) external nonReentrant() onlyOwner {
+        IERC20(rewardCurrency).safeTransfer(msg.sender, amount);
+        emit WithdrawRewards(msg.sender, amount);
+    }
+
+    function depositRewards(address rewardCurrency, uint256 amount) external nonReentrant() onlyOwner {
+        IERC20(rewardCurrency).safeTransferFrom(msg.sender, address(this), amount);
+        emit DepositRewards(msg.sender, amount);
+    }
```
mrice32, line 152: nit: just want to double check that we _want_ an explicit method for this as opposed to just doing raw transfers to the contract?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+        lastSeededIndex = indexToSeed.add(1);
+
+        _seedWindow(indexToSeed, totalRewardsDistributed, windowStart, rewardToken, merkleRoot);
+    }
+
+    // When `window.locked` is true, claims are blocked for that window.
+    function setWindowLock(uint256 windowIndex, bool lockValue) external nonReentrant() onlyOwner {
+        merkleWindows[windowIndex].locked = lockValue;
+        emit SetWindowLock(msg.sender, windowIndex, lockValue);
+    }
```
mrice32, line 126: What do you think the tradeoffs of the lock functionality vs delete?

1. Lock seems better if we're worried about accidentally deleting the wrong merkel root. I could see the argument that once these are uploaded, we should only be able to disable them, not remove them. Otherwise a fat-finger could cause us to have to regenerate and reupload a modified root that removes the claims that had already happened on the deleted root.

2. Delete seems better if we're worried about accidentally re-enabling a distribution that we had been intended to be locked permanently (and rewards withdrawn). This also makes claiming a _little_ cheaper.

Thoughts? I personally think 1 sounds a bit stronger.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+    //     - At this point contract owns 100 + 1 = 101 tokens. Now, imagine the following sequence:
+    //       (1) Claimant A claims 50 tokens for Window 1, contract now has 101 - 50 = 51 tokens.
+    //       (2) Claimant B claims 51 tokens for Window 1, contract now has 51 - 51 = 0 tokens.
+    //       (3) Claimant A tries to claim 1 token for Window 2 but fails because contract has 0 tokens.
+    //     - In summary, the contract owner created a race for step(2) and step(3) in which the first
+    //       claim would succeed and the second claim would fail, even though both claimants would expect
+    //       their claims to suceed.
+    function setWindowMerkleRoot(
+        uint256 totalRewardsDistributed,
+        uint256 windowStart,
```
mrice32, line 112: Is there a strong reason to keep the start time? Can you think of any scenarios where we'd need to specify some future start time rather than just starting the distribution immediately?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+        address account
+    ) public nonReentrant() {
+        uint256 amountToClaim = 0;
+        for (uint256 i = 0; i < claims.length; i++) {
+            Claim memory claim = claims[i];
+            require(claim.account == account, "Invalid account in batch claim");
+            _markClaimed(claim);
+            amountToClaim = amountToClaim.add(claim.amount);
+        }
+        _disburse(IERC20(rewardToken), account, amountToClaim);
```
mrice32, line 173: +1 to splitting out disburse and claim to save some gas.

mrice32 (commented):
This is awesome!! A few comments

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,236 @@
...
+    struct Window {
+        // Claims for this window can begin after `start`.
+        uint256 start;
+        // Merkle root describing the distribution.
+        bytes32 merkleRoot;
+        // Currency in which reward is processed.
+        IERC20 rewardToken;
+        // Total amount of rewards distributed this window. This is not enforced
+        // but might be useful to query.
+        uint256 totalRewardsDistributed;
```
mrice32, line 40: nit: thoughts on changing this to just `rewards`, or `totalRewards`?

Would also like to double check that this variable isn't causing any increase in gas costs during the claim. I only ask because I could see solidity doing something dumb like reading this slot when the struct is accessed even though _this slot_ isn't explicitly used in the code. If that is the case, I would argue that we should rely on the event to access this var. Otherwise, definitely on board with keeping it.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: >  those who have already claimed would still be marked "claimed". There would be no way to distribute to them in the updated root.

Great point, although we could just store the `claimed` mapping inside the Window struct to reset it on this call.

I do think that overwriting merkle roots introduces a lot of uncertainty, in practice the owner might just find it safer to lock a Window, withdraw remaining rewards, and set a new Window.

I'm OK to remove this functionality as it doesn't seem vital. @chrismaree wdyt?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I am not strongly opinionated actually. It's simpler to just go with raw transfers

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I thought we had discussed the desired functionality of "scheduling" distributions in the future, which is the main power of the window start.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Yeah this variable is mostly just a sanity check, or a convenience variable, so I'm happy to remove it.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I don't see the owner accidentally re-enabling a locked distribution, but I see your point about. making the claim cheaper. However, deleting a root seems bit dangerous and I'd rather enforce the functionality that once state is stored on this contract it remains.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
daywiss: thanks matt for bringing this up. it is wierd that you can just overwrite the root, especially seeing the examples you provide. would feel more comfortable with lock, withdraw and create.  maybe we add a seperate cleanup operation that will remove locked or expired windows/claims.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Removed this function

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I removed this variable because there is no strict reason why it needs to be stored on chain. It could easily be stored offchain on IPFS for example

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I think Delete + withdraw might be the best because we don't have to store the `locked` variable on chain.

nicholaspai: @chrismaree wdyt

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Actually, I'm thinking about this some more and I think removing lock in favor of being able to destroy merkle roots seems best. It saves gas for claims because we remove the `lock` data, and I don't see why claims would ever need to be locked temporarily. If an error in the tree was made, then its best to just delete it (and stop all claims immediately), and re-upload (+ optionally withdraw any rewards).

@chrismaree wdyt?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: BTW removing the `require(currentTime >= windowStart)` saves 5000/96000 gas, so its not inconsequential

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: @nicholaspai 

> although we could just store the claimed mapping inside the Window struct to reset it on this call.

Sadly, in solidity, there is no way to delete an entire mapping wholesale in solidity. See the docs here: https://docs.soliditylang.org/en/v0.6.12/security-considerations.html#clearing-mappings.

nicholaspai: Can you totally delete a struct that contains no mappings? That's what we're doing now in the latest code

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: Deleting SGTM! Would be interested to hear @chrismaree's perspective!

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: @nicholaspai yes, as far as I'm aware, normal structs containing only primitives (or nested primitives) can be deleted. I know mappings cannot be deleted. I forget whether `delete` works on arrays.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,217 @@
...
+        uint256 amount
+    ) private {
+        // TODO: Should we revert claims for 0 tokens?
+        if (amount > 0) {
+            token.safeTransfer(account, amount);
+        }
+    }
+
+    // Checks {account, amount} against Merkle root at given window index.
+    function _verifyClaim(Claim memory claim) private view returns (bool valid) {
```
mrice32, line 213: Was this previously public? I could see it making sense to keep it public 🤷‍♀️

packages/merkle-distributor/src/merkleTree.js
```diff
@@ -0,0 +1,131 @@
+const web3 = require("web3");
+const ethereumjsUtil = require("ethereumjs-util");
+
+class MerkleTree {
+  constructor(elements) {
+    // Filter empty strings and hash elements
+    // NOTE do not double has leafs
```
mrice32, line 7: typo?

packages/merkle-distributor/src/merkleTree.js
```diff
@@ -0,0 +1,131 @@
...
+    // this.elements = elements.filter(el => el).map(el => keccakFromString(el));
+    this.elements = elements;
+    // Sort elements
+    this.elements.sort();
+    // Deduplicate elements
+    // this.elements = this.bufDedup(this.elements);
```
mrice32, line 13: Commented code?

packages/merkle-distributor/src/merkleTree.js
```diff
@@ -0,0 +1,131 @@
+const web3 = require("web3");
+const ethereumjsUtil = require("ethereumjs-util");
+
+class MerkleTree {
+  constructor(elements) {
```
mrice32, line 5: Could you describe what type elements is?

packages/merkle-distributor/src/merkleTree.js
```diff
@@ -0,0 +1,131 @@
+const web3 = require("web3");
+const ethereumjsUtil = require("ethereumjs-util");
+
+class MerkleTree {
```
mrice32, line 4: Just curious -- is this code inspired or pulled from somewhere else?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,217 @@
...
+     * Public functions
+     *
+     ****************************/
+
+    // Batch claims for a reward currency for an account to save gas.
+    function claimWindows(
+        Claim[] memory claims,
+        address rewardToken,
+        address account
+    ) public nonReentrant() {
```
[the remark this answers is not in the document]
nicholaspai: This is a really good point, I don't think there is reason to have the reentrancy guards. One consequence is that a malicious owner might seed the contract with an ERC20 that attempts to re-enter the MerkleDistributor contract on the `transfer` call, but at that point it can't really exploit anything from what I can tell.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,217 @@
...
+        bytes32 merkleRoot
+    ) external nonReentrant() onlyOwner {
+        uint256 indexToSeed = lastSeededIndex;
+        lastSeededIndex = indexToSeed.add(1);
+
+        _seedWindow(indexToSeed, totalRewardsDistributed, windowStart, rewardToken, merkleRoot);
+    }
+
+    // Delete merkle root at window index. Likely to be followed by a withdrawRewards call to clear contract state.
+    function destroyMerkleRoot(uint256 windowIndex) external nonReentrant() onlyOwner {
```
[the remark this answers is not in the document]
nicholaspai: Better name +1

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -0,0 +1,217 @@
...
+    //     - Window 2 Tree: Owner sets `totalRewardsDistributed=1` and insert proofs that give
+    //                      claimant A 1 token. The owner correctly set `totalRewardsDistributed` this time.
+    //     - At this point contract owns 100 + 1 = 101 tokens. Now, imagine the following sequence:
+    //       (1) Claimant A claims 50 tokens for Window 1, contract now has 101 - 50 = 51 tokens.
+    //       (2) Claimant B claims 51 tokens for Window 1, contract now has 51 - 51 = 0 tokens.
+    //       (3) Claimant A tries to claim 1 token for Window 2 but fails because contract has 0 tokens.
+    //     - In summary, the contract owner created a race for step(2) and step(3) in which the first
+    //       claim would succeed and the second claim would fail, even though both claimants would expect
+    //       their claims to suceed.
+    function setWindowMerkleRoot(
```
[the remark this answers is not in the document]
nicholaspai: Consistent with `deleteWindow` so I agree with this

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Yeah, but I don't really see a reason to keep it public vs private so by default I lean private. I think you could achieve the same "sanity" check by `.call()`-ing on `claimWindow`

packages/merkle-distributor/src/merkleTree.js
nicholaspai: Probably? @chrismaree wrote this

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
nicholaspai: This saves 4000 bytes btw

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
chrismaree: yup! I agree with your analysis. the scenario we'd want this would be an error in the payouts. we'd want to manually re-compute the tree with removed claimers and then add a new hash. in this scenario deleting and re-uploading is the best outcome.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
chrismaree: SGTM. I've got this in the SDK right now already.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
chrismaree: agreed! not really needed.

packages/merkle-distributor/src/merkleTree.js
chrismaree: nope, I did not. this script was copied directly from the Piedao implementation, found [here](https://github.com/pie-dao/vested-token-migration-app/blob/master/scripts/merkleTree.js). The one change I did make was changing this to work with web3 over Ethers.


Note that in the SDK version I'm building out on a separate PR this file is moved over to typescript and copies the uniswap implementation.

packages/merkle-distributor/src/merkleTree.js
chrismaree: ye does not matter in this context and came from the complied implementation from piedao.

packages/merkle-distributor/src/merkleTree.js
chrismaree: I'll be sure to add better comments in the SDK. Elements is an array of hashed leaves.

packages/merkle-distributor/src/merkleTree.js
chrismaree: see above ^^ https://github.com/UMAprotocol/protocol/pull/2598#discussion_r586316304

packages/merkle-distributor/src/merkleTree.js
mrice32: Can we drop this file from this PR then? Or do we still need it?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: That's fair, although there is one minor difference -- if you've already claimed this would return true but claimWindow would revert. Not sure that's a huge deal because you can just || between the mapping and the `.call` to see if that person was included in a merkle root. Not strongly opinionated, and not sure we'd ever really need to do this.

mrice32 (approved):
Looks great! I'm still a little confused about the Merkle tree offchain implementation and whether we need it in this PR.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
nicholaspai: Yea its been dropped

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Yeah that confusion alone is worth removing for me.

pemulis (approved):
LGTM! Answered all of my questions and tests are green.

packages/merkle-distributor/src/merkleTree.js
chrismaree: it's needed in the unit tests to generate the proofs.

chrismaree (approved):
image

PR #2600, opened 2021-02-25
feat(core)!: deploy slightly modified designated voting factory

mrice32:
**Motivation**

The existing DesignatedVotingFactory _doesn't_ allow users to redeploy or reset their DesignatedVoting contract after setting it. To make DesignatedVotingFactory contracts easier to use for hot wallets, it would be nice if we could remove this restriction.

We also need to deploy new updated DesignatedVoting contracts that are compatible with the upgraded DVM (specifically ancillary data).

**Summary**

This PR makes a minor change to allow users to reset or deploy new designated voting factories after their (two removed requires).

This PR also adds deployed addresses for these new DesignatedVotingFactory contracts on kovan and mainnet.

Note: I _think_ this is safe to change without audit since the changes only concern the hot-wallet mapping and they do not touch the DesignatedVoting contract, itself, which holds the tokens. The changes are also quite minor and limited in scope. However, I think we should have OpenZeppelin audit this as soon as we start our next audit phase. However, I would appreciate feedback on this point.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
looks good to me! I dont see any major security issues with letting someone have multiple designated voting contracts.

Looks like your tests are failing due to changed metadata, which changed the byte code. We might want to consider building the byte code version for _local_ testing in a different way. For example, the `FindContractVersion` could be populated with an automated script that would ensure that the compiled code matches the listed versions for each version of the core package. Otherwise, any PR that makes a change to the smart contracts directory could break CI.

PR #2611, opened 2021-02-27
improve(tradermade-pricefeed): Fall back to hourly OHLC interval when minute interval fails

nicholaspai:

**Motivation**

TraderMade API expectedly returns no timeseries data over the weekend, but the price identifier UMIPs that suggest using the TraderMade API request that pricefeeds fall back to the last known price, see [UMIP-32](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-32.md#weekend-price) for an example of this recommendation.

[The PR that implemented the TraderMadePriceFeed originally was supposed to implement this fallback](https://github.com/UMAprotocol/protocol/pull/2479#issuecomment-771237776) but the feature got lost to no individual fault. This PR therefore adds the fallback functionality to both the `update()` and `getHistoricalPrice` methods:

- `update` fetches historical price data from the `minute` timeseries but also should fallback to the `hourly` timeseries if the `minute` update fails.
- `getHistoricalPrices` should use `historicalPricesHourly` if `historicalPricesMinute` was not constructed properly during `update`.

Note: Another TraderMade UMIP for XAUUSD should add in language that addresses weekend-price, [UMIP-26](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-26.md).

**Summary**

TODO: add unit tests for the following cases:
- `minuteLookback` and `hourlyLookback` are specified, `minute` timeseries is empty and `hourly` timeseries is OK, error should NOT be thrown on `update()`
- `minuteLookback` is specified, `hourlyLookback` is NOT specified, `minute` timeseries is empty and `hourly` timeseries is OK, error should  be thrown on `update()` because fallback to `hourly` should be skipped.
- `minuteLookback` is NOT specified, `hourlyLookback` is specified, `minute` timeseries is empty and `hourly` timeseries is OK, error should  NOT be thrown on `update()` because `updateMinute` should be skipped.

Current unit tests test the following case:
- `minuteLookback` and `hourlyLookback` are specified, but `minute` and `hourly` timeseries are empty, error should be thrown on `update()`

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -386,7 +386,8 @@ const defaultConfigs = {
     pair: "CNYUSD",
     minuteLookback: 7200, // interval=minute allowed lookback up to 2 days
     hourlyLookback: 604800, // interval=hour allowed lookback up to 2 months
-    minTimeBetweenUpdates: 600
+    minTimeBetweenUpdates: 600,
+    ohlcPeriod: 10 // CNYUSD only available at 10 minute granularity
   },
   PHPDAI: {
     type: "medianizer",
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -384,8 +384,6 @@ const defaultConfigs = {
   CNYUSD: {
     type: "tradermade",
     pair: "CNYUSD",
-    minuteLookback: 7200, // interval=minute allowed lookback up to 2 days
-    hourlyLookback: 604800, // interval=hour allowed lookback up to 2 months
     minTimeBetweenUpdates: 600,
     ohlcPeriod: 10 // CNYUSD only available at 10 minute granularity
   },
@@ -475,9 +473,7 @@ const defaultConfigs = {
   XAUUSD: {
     type: "tradermade",
     pair: "XAUUSD",
-    minTimeBetweenUpdates: 60,
-    minuteLookback: 7200,
-    hourlyLookback: 604800
+    minTimeBetweenUpdates: 60
   },
   // The following identifiers can be used to test how `CreatePriceFeed` interacts with this
   // default price feed config.
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -385,6 +385,13 @@ const defaultConfigs = {
     type: "tradermade",
     pair: "CNYUSD",
     minTimeBetweenUpdates: 600,
+    minuteLookback: 172800,
+    // Set minute lookback to maximum of 2 days allowed by minute-interval endpoint. This is longer
+    // than the expected liquidationLiveness of 2 hours because we might need to get the last price
+    // before the tradermade API stopped serving prices for the weekend.
+    hourlyLookback: 604800,
+    // Hourly interval data will be used as fallback if minute data is empty, so
+    // set conservatively to 1 week.
     ohlcPeriod: 10 // CNYUSD only available at 10 minute granularity
   },
   PHPDAI: {
@@ -473,6 +480,13 @@ const defaultConfigs = {
   XAUUSD: {
     type: "tradermade",
     pair: "XAUUSD",
+    minuteLookback: 172800,
+    // Set minute lookback to maximum of 2 days allowed by minute-interval endpoint. This is longer
+    // than the expected liquidationLiveness of 2 hours because we might need to get the last price
+    // before the tradermade API stopped serving prices for the weekend.
+    hourlyLookback: 604800,
+    // Hourly interval data will be used as fallback if minute data is empty, so
+    // set conservatively to 1 week.
     minTimeBetweenUpdates: 60
   },
   // The following identifiers can be used to test how `CreatePriceFeed` interacts with this
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -385,17 +385,15 @@ const defaultConfigs = {
     type: "tradermade",
     pair: "CNYUSD",
     minTimeBetweenUpdates: 600,
-    // Since TradeMadeAPI timeseries data is unavailable over the weekend,
-    // set either the `minuteLookback` or the `hourlyLookback` to be longer than 3 days (259200 seconds)
-    // so that we can fetch the last known price prior to the weekend start.
-    // For example, if `minuteLookback = 7200`, then 2 hours after the weekend has "begun",
-    // `TraderMadePriceFeed.updateMinute()` will fail to find any prices. Simultaneously
-    // setting `hourLookback = 604800` ensures that if `updateMinute()` fails then `updateHourly()`
-    // will return the last hourly price. Alternatively, you can set `minuteLookback = 172800`, or
-    // 2 days (the maximum minute interval lookback) which will allow `updateMinute()` to fetch
-    // data for longer into the off-market period. We set `minuteLookback` to 7200 and `hourlyLookback`
-    // to 604800 so that we can both reduce the amount of data to parse from the minute-interval
-    // and ensure that we can always find a price when weekend prices are not available.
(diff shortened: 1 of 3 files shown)


packages/financial-templates-lib/test/price-feed/TraderMadePriceFeed.js
```diff
@@ -148,87 +176,303 @@ contract("TraderMadePriceFeed.js", function() {
...
+      // Last log should be about updating minute interval.
+      assert.isTrue(spyLogIncludes(spy, -1, "Updating Minute Price"));
+
+      assert.equal(traderMadePriceFeed.getCurrentPrice().toString(), toWei("0.1553"));
+      assert.isTrue(
+        await traderMadePriceFeed.getHistoricalPrice(1611608300 - timezeoneOffsetSeconds).catch(() => true)
+      );
+      assert.equal(traderMadePriceFeed.getHistoricalPricePeriods(), undefined);
+    });
+    it("Fallback to hourly interval ENABLED, latest, minute and hourly intervals all fail to respond", async function() {
```
nicholaspai, line 306: This is the same test as the one previously labeled by "No or bad response"

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -384,9 +384,8 @@ const defaultConfigs = {
   CNYUSD: {
     type: "tradermade",
     pair: "CNYUSD",
-    minuteLookback: 7200, // interval=minute allowed lookback up to 2 days
```
nicholaspai, line 387: We should be setting defaults for `minuteLookback` and `hourlyLookback` within the `TraderMadePriceFeed`. The `lookback` will be automatically set by the client `liquidator`, `disputer`, or `monitor` bot but we can't really assume the window of time for which we want to query the timeseries data.

So we'll basically just ask for more data than we neccessarily need to lookback through.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -384,9 +384,21 @@ const defaultConfigs = {
...
+    // set either the `minuteLookback` or the `hourlyLookback` to be longer than 3 days (259200 seconds)
+    // so that we can fetch the last known price prior to the weekend start.
+    // For example, if `minuteLookback = 7200`, then 2 hours after the weekend has "begun",
+    // `TraderMadePriceFeed.updateMinute()` will fail to find any prices. Simultaneously
+    // setting `hourLookback = 604800` ensures that if `updateMinute()` fails then `updateHourly()`
+    // will return the last hourly price. Alternatively, you can set `minuteLookback = 172800`, or
+    // 2 days (the maximum minute interval lookback) which will allow `updateMinute()` to fetch
+    // data for longer into the off-market period. We set `minuteLookback` to 7200 and `hourlyLookback`
+    // to 604800 so that we can both reduce the amount of data to parse from the minute-interval
+    // and ensure that we can always find a price when weekend prices are not available.
```
chrismaree, line 398: small nit: Due to our styling guide, this can be represented with a few smaller lines if you try take each line to 120 chars.

```suggestion
// Since TradeMadeAPI timeseries data is unavailable over the weekend, set either the `minuteLookback` or the
// `hourlyLookback` to be longer than 3 days (259200 seconds) so that we can fetch the last known price prior to the
// weekend start. For example, if `minuteLookback = 7200`, then 2 hours after the weekend has "begun",
// `TraderMadePriceFeed.updateMinute()` will fail to find any prices. Simultaneously setting `hourLookback = 604800`
// ensures that if `updateMinute()` fails then `updateHourly()` will return the last hourly price. Alternatively, you
// can set `minuteLookback = 172800`, or 2 days (the maximum minute interval lookback) which will allow `updateMinute()`
// to fetch data for longer into the off-market period. We set `minuteLookback` to 7200 and `hourlyLookback` to 604800 so
// that we can both reduce the amount of data to parse from the minute-interval and ensure that we can always find a price when 
// weekend prices are not available.
```

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -474,9 +486,20 @@ const defaultConfigs = {
   XAUUSD: {
     type: "tradermade",
     pair: "XAUUSD",
-    minTimeBetweenUpdates: 60,
+    // Since TradeMadeAPI timeseries data is unavailable over the weekend,
```
chrismaree, line 489: do we want to repeat this? we could also just put the comment within the feed I guess?

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -30,7 +30,7 @@ class TraderMadePriceFeed extends PriceFeedInterface {
     getTime,
     minTimeBetweenUpdates,
     priceFeedDecimals = 18,
-    ohlcPeriod = 10 // Only 5, 10, 15 minutes is supported by TraderMade.
+    ohlcPeriod = 1 // Some pairs like CNYUSD are only available at higher granularities like 10
```
chrismaree, line 33: is there any way to sanitise this or catch if it is misconfigured?

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
```diff
@@ -70,9 +70,20 @@ class TraderMadePriceFeed extends PriceFeedInterface {
...
-      if (this.historicalPricesMinute[p] && this.historicalPricesMinute[p].openTime) {
-        firstPriceTime = this.historicalPricesMinute[p];
+
+    // If no minute-interval prices are available, try the hourly-interval:
+    // Note: The TraderMade API does not update /timeseries data over the weekend,
+    // so to handle this case we can fall back on the longer lookback window of the
+    // hourly timeseries. (The lookback limit for the minute inverval is 2 days, while the limit
+    // for the hourly interval is 2 months). This fall back logic will only work
+    // if `this.hourlyLookback` is configured long enough such that there is an hourly price
+    // available.
```
chrismaree, line 80: again nit about line length
```suggestion
// Note: The TraderMade API does not update /timeseries data over the weekend, so to handle this case we can fall back
// on the longer lookback window of the hourly timeseries. (The lookback limit for the minute inverval is 2 days, while
// the limit for the hourly interval is 2 months). This fall back logic will only work if `this.hourlyLookback` is
// configured long enough such that there is an hourly price available.
```

BTW you can set VS code to show you vertical lines at our line length. add this to your vs code config:
```
"editor.rulers": [
        80,
        120
    ],
```

packages/financial-templates-lib/test/price-feed/TraderMadePriceFeed.js
```diff
@@ -148,87 +174,303 @@ contract("TraderMadePriceFeed.js", function() {
...
+        networker,
+        getTime,
+        minTimeBetweenUpdates,
+        18
+      );
+      // Update should not throw and the historical price should correspond to hourly interval data.
+      await traderMadePriceFeed.update();
+
+      // Second to last log should be about falling back to hourly interval:
+      assert.isTrue(spyLogIncludes(spy, -2, "updateMinute failed, falling back to updateHourly"));
```
chrismaree, line 235: nice. before not this feed did not check any spy stuff.

chrismaree (approved):
LGTM! thanks for fixing this over the weekend!

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: Good point, I'll refactor to the feed

packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.js
nicholaspai: Yes I added these lines to top of constructor:

```
    // Sanitize API specific parameters:
    assert(VALID_OHLC_PERIODS.includes(ohlcPeriod), `ohlcPeriod must be one of ${JSON.stringify(VALID_OHLC_PERIODS)}`);
    assert(minuteLookback <= MAX_MINUTE_LOOKBACK, `minuteLookback must be < ${MAX_MINUTE_LOOKBACK}`);
    assert(hourlyLookback <= MAX_HOURLY_LOOKBACK, `hourlyLookback must be < ${MAX_HOURLY_LOOKBACK}`);

```

where:

```
const VALID_OHLC_PERIODS = [1, 5, 10, 15, 30];
const MAX_MINUTE_LOOKBACK = 172800; // 2 days
const MAX_HOURLY_LOOKBACK = 5184000; //2 months
```

PR #2612, opened 2021-02-28
patch(disputer): Cast liquidation price BN --> String before logging

nicholaspai:

**Motivation**

After dispute is sent, DIsputer logs details about the transaction receipt including the historical price it used to submit the dispute. The price is not correctly cast to a String before sending to the logger, which causes the Slack Transport to error.

Notice in the logs below how the `price` looks strange, because the `price` is actually a BN object.
Screen Shot 2021-02-28 at 16 03 19

**Testing**

Tested on Kovan

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/disputer/src/disputer.js
+++ b/packages/disputer/src/disputer.js
@@ -153,7 +153,7 @@ class Disputer {
               liquidation: JSON.stringify(liquidation)
             });
 
-            return { ...liquidation, price };
+            return { ...liquidation, price: price.toString() };
           }
 
           return null;
@@ -205,7 +205,6 @@ class Disputer {
         at: "Disputer",
         message: "Disputing liquidation",
         liquidation: disputeableLiquidation,
-        inputPrice,
         txnConfig
       });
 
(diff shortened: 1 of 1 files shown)


packages/disputer/src/disputer.js
```diff
@@ -153,7 +153,7 @@ class Disputer {
               liquidation: JSON.stringify(liquidation)
             });
 
-            return { ...liquidation, price };
+            return { ...liquidation, price: price.toString() };
```
nicholaspai, line 156: This price is only added here to make the ultimate Slack logs have more data.

packages/disputer/src/disputer.js
```diff
@@ -199,13 +199,10 @@ class Disputer {
         gasPrice: this.gasEstimator.getCurrentFastPrice()
       };
 
-      const inputPrice = disputeableLiquidation.price;
-
       this.logger.debug({
         at: "Disputer",
```
nicholaspai, line 203: `inputPrice` is no longer necessary to add to the logs because it is included in the `liquidation` object

chrismaree (approved):

PR #1753, opened 2020-07-17
feat(liquidatable): #1553 add new event params to liquidatable events

daywiss:
specifically updates `LiquidationCreated` and `LiquidationWithdrawn`

image

There might be room for one more event param if needed.

this references #1553

PR #2613, opened 2021-03-01
patch(disputer,liquidator): Handle changed LiquidationWithdrawn event params

nicholaspai:

**Motivation**

Resolved disputes are withdrawn successfully but event log parsing is mishandled if EMP version is not a legacy (<=1.2.2) one.

Additionally, we can query `settlementPrice` from the event logs instead of making a call to the VotingContract.getPrice(), which removes the need for the bots to construct the Voting contract.

Tested log output to Slack successfully on Kovan.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/disputer/index.js
+++ b/packages/disputer/index.js
@@ -19,7 +19,7 @@ const {
 } = require("@uma/financial-templates-lib");
 
 // Truffle contracts.
-const { getAbi, getAddress } = require("@uma/core");
+const { getAbi } = require("@uma/core");
 const { getWeb3 } = require("@uma/common");
 
 /**
@@ -63,10 +63,9 @@ async function run({
     });
 
     // Load unlocked web3 accounts and get the networkId.
-    const [detectedContract, accounts, networkId] = await Promise.all([
+    const [detectedContract, accounts] = await Promise.all([
       findContractVersion(financialContractAddress, web3),
-      web3.eth.getAccounts(),
-      web3.eth.net.getId()
+      web3.eth.getAccounts()
     ]);
 
     // Append the contract version and type to the disputerConfig, if the disputerConfig does not already contain one.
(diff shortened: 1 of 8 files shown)


packages/disputer/index.js
```diff
@@ -89,7 +88,6 @@ async function run({
       );
 
     // Setup contract instances.
-    const voting = new web3.eth.Contract(getAbi("Voting", "1.2.2"), getAddress("Voting", networkId));
```
nicholaspai, line 92: No more need to pass in `voting` to Disputer, which was previously used to make a `getPrice` call to figure out the settlement price. `settlementPrice` is part of event params!

packages/disputer/src/disputer.js
```diff
@@ -230,7 +232,6 @@ class Disputer {
         at: "Disputer",
         message: "Position has been disputed!👮‍♂️",
         liquidation: disputeableLiquidation,
-        txnConfig,
```
nicholaspai, line 233: No reason to print out `txnConfig`, remove log clutter

packages/disputer/src/disputer.js
```diff
@@ -328,31 +322,30 @@ class Disputer {
...
-        .call({
-          from: this.financialContract.options.address
-        });
-
-      const logResult = {
+      let logResult = {
         tx: receipt.transactionHash,
         caller: receipt.events.LiquidationWithdrawn.returnValues.caller,
-        withdrawalAmount: receipt.events.LiquidationWithdrawn.returnValues.withdrawalAmount,
+        settlementPrice: receipt.events.LiquidationWithdrawn.returnValues.settlementPrice,
```
daywiss, line 328: does this work with all previous contracts?

daywiss (approved):

packages/disputer/src/disputer.js
nicholaspai: Yep I double checked. You actually added this in a while ago (prior to version 1.1.0 even) but we never took advantage of it in the bots. This event change was made in July, a month before we set up our versioning!

https://github.com/UMAprotocol/protocol/pull/1753

packages/disputer/index.js
mrice32: Might be a dumb question, but won't we still need it if we're interacting with an older EMP?

mrice32: Ah, saw your comment below about changing this a while back.

mrice32 (approved):
LGTM! Nice cleanup!

PR #2616, opened 2021-03-01
feat(encoding): add EncodeParams example script

pemulis:

**Motivation**

This is a lightweight script for generating encoded parameters for an EMP. It's useful if the EMP was deployed through `launch-emp` or some other method, and could be modified to encode parameters for Perpetuals in the future. We need properly encoded parameters to verify the EMP contract on Etherscan.

**Summary**

Using `DeployEMP` as a starting point, stripped the script down to just what was needed to encode the parameters.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

n/a

### DIFF
--- a/packages/core/scripts/local/EncodeParams.js
+++ b/packages/core/scripts/local/EncodeParams.js
@@ -1,0 +1,86 @@
+/**
+ * @notice The purpose of this script is to encode ExpiringMultiParty parameters.
+ *
+ * Spinning up mainnet fork: ganache-cli --fork https://mainnet.infura.io/v3/5f56f0a4c8844c96a430fbd3d7993e39 --unlock 0x2bAaA41d155ad8a4126184950B31F50A1513cE25 --unlock 0x7a3a1c2de64f20eb5e916f40d11b01c441b2a8dc --port 9545
+ * Example: yarn truffle exec ./packages/core/scripts/local/EncodeParams.js --cversion latest --network mainnet-fork
+ */
+const { toWei } = web3.utils;
+const { ZERO_ADDRESS } = require("@uma/common");
+const { getAbi, getTruffleContract } = require("../../index");
+const argv = require("minimist")(process.argv.slice(), {
+  string: ["cversion"]
+});
+const abiVersion = argv.cversion || "latest"; // Default to most recent mainnet deployment, 1.2.2.
+
+// Deployed contract ABI's and addresses we need to fetch.
+const ExpiringMultiPartyCreator = getTruffleContract("ExpiringMultiPartyCreator", web3, abiVersion);
+const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, abiVersion);
+const Finder = getTruffleContract("Finder", web3, abiVersion);
+const TokenFactory = getTruffleContract("TokenFactory", web3, abiVersion);
+
+/** ***************************************************
+ * Main Script
+ /*****************************************************/
+const encodeParams = async callback => {
+  try {
+    const expiringMultiPartyCreator = await ExpiringMultiPartyCreator.deployed();
+    const finder = await Finder.deployed();
+    const tokenFactory = await TokenFactory.deployed();
+
+    // Replace these with the EMP constructor params you want to encode
+    let constructorParams = {
+      expirationTimestamp: "1619812800",
+      collateralAddress: "0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48", // USDC
+      priceFeedIdentifier: "0x7553544f4e4b535f4150523231",
+      syntheticName: "uSTONKS Index Token April 2021",
+      syntheticSymbol: "uSTONKS_APR21",
+      collateralRequirement: { rawValue: toWei("1.25") },
+      disputeBondPercentage: { rawValue: toWei("0.1") },
+      sponsorDisputeRewardPercentage: { rawValue: toWei("0.05") },
+      disputerDisputeRewardPercentage: { rawValue: toWei("0.2") },
+      minSponsorTokens: { rawValue: toWei("1") },
+      liquidationLiveness: 7200,
+      withdrawalLiveness: 7200
+    };
+
+    // Inject constructor params neccessary for "latest" version of the EMPCreator:
+    if (abiVersion === "latest") {
+      constructorParams = {
+        ...constructorParams,
+        financialProductLibraryAddress: ZERO_ADDRESS
+      };
+    }
+
+    const emp = await ExpiringMultiParty.at("0x5A7f8F8B0E912BBF8525bc3fb2ae46E70Db9516B");
+
+    let empConstructorParams = {
+      ...constructorParams,
+      finderAddress: finder.address,
+      tokenFactoryAddress: tokenFactory.address,
+      timerAddress: await expiringMultiPartyCreator.timerAddress()
+    };
+
+    // Grab `tokenAddress` from newly constructed EMP and add to `empConstructorParams` for new EMP's
+    if (abiVersion === "latest") {
+      empConstructorParams = {
+        ...empConstructorParams,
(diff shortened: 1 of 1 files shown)


nicholaspai (commented):
Drive by comment: why not modify the encodeParams logic in DeployEMP to also print out the encoded params foe the perp?

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+
+    // Inject constructor params neccessary for "latest" version of the EMPCreator:
+    if (abiVersion === "latest") {
+      constructorParams = {
+        ...constructorParams,
+        financialProductLibraryAddress: ZERO_ADDRESS
+      };
+    }
+
+    const emp = await ExpiringMultiParty.at("0x5A7f8F8B0E912BBF8525bc3fb2ae46E70Db9516B");
```
nicholaspai, line 54: Should we make this address customizable via the command line args?

nicholaspai:
> Drive by comment: why not modify the encodeParams logic in DeployEMP to also print out the encoded params foe the perp? 

Ah just ready the description more so. My bad. I think this script would be really useful if you could specify the EMP (or Perp?) and it would return the ABI encoded constructor params

nicholaspai (approved):

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+// Deployed contract ABI's and addresses we need to fetch.
+const ExpiringMultiPartyCreator = getTruffleContract("ExpiringMultiPartyCreator", web3, abiVersion);
+const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, abiVersion);
+const Finder = getTruffleContract("Finder", web3, abiVersion);
+const TokenFactory = getTruffleContract("TokenFactory", web3, abiVersion);
+
+/** ***************************************************
+ * Main Script
+ /*****************************************************/
+const encodeParams = async callback => {
```
daywiss, line 24: why is this a callback and not just return the result or throw the error naturally?

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+      syntheticName: "uSTONKS Index Token April 2021",
+      syntheticSymbol: "uSTONKS_APR21",
+      collateralRequirement: { rawValue: toWei("1.25") },
+      disputeBondPercentage: { rawValue: toWei("0.1") },
+      sponsorDisputeRewardPercentage: { rawValue: toWei("0.05") },
+      disputerDisputeRewardPercentage: { rawValue: toWei("0.2") },
+      minSponsorTokens: { rawValue: toWei("1") },
+      liquidationLiveness: 7200,
+      withdrawalLiveness: 7200
+    };
```
daywiss, line 44: are these supposed to be able to be overridden?

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+    console.table(empConstructorParams);
+  } catch (err) {
+    console.error(err);
+    callback(err);
+    return;
+  }
+  callback();
+};
+
+module.exports = encodeParams;
```
daywiss, line 86: i dont understand how this script gets executed, it looks like its more like a library unless maybe im missing something about the truffle runtime environment that makes this magically work

daywiss (commented):
i dont quite understand what the script is used for. left some questions, but maybe the idea is that this is more about documentation and less about using as an actual script?

packages/core/scripts/local/EncodeParams.js
mrice32: When run with truffle exec, it has to be written this way. If we want to run it _outside_ of truffle exec (similar to how we run the bots) we could avoid this weird syntax. If we were to do that, not sure if we should put this script in a separate dir or something.

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+ * Main Script
+ /*****************************************************/
+const encodeParams = async callback => {
+  try {
+    const expiringMultiPartyCreator = await ExpiringMultiPartyCreator.deployed();
+    const finder = await Finder.deployed();
+    const tokenFactory = await TokenFactory.deployed();
+
+    // Replace these with the EMP constructor params you want to encode
+    let constructorParams = {
```
mrice32, line 31: Could we decode these from the on-chain transaction?

Something like this, maybe:

```js
// At the top
const abiDecoder = require("abi-decoder");

...
// this could also just be a single abi if you know which one you want.
const abis = [getAbi("ExpiringMultiParty", "1.1.0"), getAbi("ExpiringMultiParty", "1.2.0") ....];
abis.forEach(abi => abiDecoder.addAbi(abi));
let constructorParams = abiDecoder.decode(transactionData).params;
```

This may be a little annoying because of some of the versioning below, but just wanted to mention it as a possibility. Not saying we necessarily _should_ do this.

packages/core/scripts/local/EncodeParams.js
```diff
@@ -0,0 +1,86 @@
...
+      timerAddress: await expiringMultiPartyCreator.timerAddress()
+    };
+
+    // Grab `tokenAddress` from newly constructed EMP and add to `empConstructorParams` for new EMP's
+    if (abiVersion === "latest") {
+      empConstructorParams = {
+        ...empConstructorParams,
+        tokenAddress: await emp.tokenCurrency()
+      };
+    }
```
mrice32, line 69: Does this break if we just include tokenAddress in the constructor params for all versions? The parameter encoding _may_ just ignore fields it doesn't recognize.

packages/core/scripts/local/EncodeParams.js
mrice32: Just a truffle exec thing.

mrice32 (commented):
Looks awesome!

pemulis:
> i dont quite understand what the script is used for. left some questions, but maybe the idea is that this is more about documentation and less about using as an actual script?

This is useful for when you need to generate encoded parameters to verify a contract on Etherscan that wasn't deployed using `DeployEMP`. The specific reason it was written was to verify `uSTONKS`, which was deployed with `launch-emp` and was the first new-style EMP deployed to mainnet.

packages/core/scripts/local/EncodeParams.js
pemulis: Yep!

packages/core/scripts/local/EncodeParams.js
pemulis: Hmm, I think we probably could. I'll try rolling this into a follow-up PR when we encode params for the perpetual.

packages/core/scripts/local/EncodeParams.js
pemulis: I think so!

I tried adding an unrecognized method, `emp.exampleAddress()`, when running against the new EMP contract, and it failed. I think it would fail in the same way when trying to `await emp.tokenCurrency()` for the old EMP.

However, this could be simplified by rolling `tokenAddress` into the `empConstructorParams` object that is first defined, since we don't need to wait on deployment.

pemulis:
> > Drive by comment: why not modify the encodeParams logic in DeployEMP to also print out the encoded params foe the perp?
> 
> Ah just ready the description more so. My bad. I think this script would be really useful if you could specify the EMP (or Perp?) and it would return the ABI encoded constructor params

I agree! I'll update this script when we need to go through this process for the perpetual.

packages/core/scripts/local/EncodeParams.js
mrice32: @pemulis does `tokenCurrency` not exist on old emps? I just thought the constructor arg was what changed, not the accessor method...

PR #2618, opened 2021-03-01
feat(core): ExecuteAdmin script can now execute multiple proposals

mrice32:
**Motivation**

ExecuteAdmin isn't particularly useful in a world where there are multiple admin proposals each week.

It needs to be able to execute multiple proposals at once. It should also be smarter about the gas prices it uses to execute these txns.

**Summary**

ExecuteAdmin now takes two params `--start` and `--end` that allow the user to specify which admin proposals they wish to execute.

If neither is provided, the script will default to only executing the most recent proposal.

`--end` defaults to the most recent proposal.

`--start` defaults to `--end` (i.e. only execute a single proposal if not provided).

These defaults ensure that the script won't change behavior by default.

The script also now uses the GasEstimator to determine the gas price it should use.

Example command:
```sh
# Run from protocol/packages/core
MNEMONIC="YOUR_MNEMONIC" CUSTOM_NODE_URL=https://your.node.url truffle exec ./scripts/mainnet/ExecuteAdmin.js --network mainnet_mnemonic --start 58 --end 61
```

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/scripts/mainnet/ExecuteAdmin.js
```diff
@@ -1,30 +1,43 @@
...
-    // for every transactions within the proposal
-    for (let i = 0; i < proposal.transactions.length; i++) {
-      console.log(`${i}: Submitting tx...`);
-      let tx = await governor.executeProposal(proposalId.toString(), i.toString());
-      console.log(`${i}: Done: `, tx.tx);
-    }
-
-    // Check that the identifier whitelist has been updated.
-    assert.equal(await identifierWhitelist.isIdentifierSupported(web3.utils.utf8ToHex("TEST-NEW-IDENTIFIER")), true);
+    const gasEstimator = new GasEstimator(
```
nicholaspai, line 19: What are these constants?

packages/core/scripts/mainnet/ExecuteAdmin.js
```diff
@@ -1,30 +1,43 @@
...
-    console.log("Admin proposal to whitelist new identifier executed!");
+    for (let j = start; j <= end; j++) {
+      const proposal = await governor.getProposal(j);
+      // for every transactions within the proposal
+      console.log(`Starting proposal ${j} with ${proposal.transactions.length} transactions.`);
+      for (let i = 0; i < proposal.transactions.length; i++) {
+        await gasEstimator.update();
+        console.log(`Proposal ${j}, Transaction ${i}: Submitting tx...`);
+        let tx = await governor.executeProposal(j.toString(), i.toString(), {
+          gasPrice: gasEstimator.getCurrentFastPrice()
```
nicholaspai, line 35: +1 using gasEstimator in scripts we should do this more...

nicholaspai (approved):

packages/core/scripts/mainnet/ExecuteAdmin.js
mrice32: Good point -- added comments!

PR #2619, opened 2021-03-02
Adding configs for btc and eth 3m/6m basis

chicfilabae:
As requested by @smb2796 , here are the BTC and ETH 3m/6m configs.

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -409,6 +409,78 @@ const defaultConfigs = {
       }
     ]
   },
+  "ETH-BASIS-6M/USDC": {
+    type: "expression",
+    expression: `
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
LGTM!

None of these have been launched yet, right? So they will be launched with the new EMP that will use 18 decimal price feeds by default, correct?

mrice32:
@chicfilabae you'll need to fix DCO. Happy to help debug if needed!

chicfilabae:
@mrice32 

Whoops, forgot to do that. Fixed now.

"None of these have been launched yet, right? So they will be launched with the new EMP that will use 18 decimal price feeds by default, correct?"

Correct, was waiting to deploy until now for that purpose. So should be all good there

daywiss (approved):
this looks fine, but hard to ensure it works from viewing the diff.  going to approve, but it probably should be tested even if its just running it to see it works.

smb2796 (approved):
Tested these price feeds locally against a manual calculation and everything looks to check out!

PR #2632, opened 2021-03-02
feat(perpetual-proposer): Initialize PerpetualFundingRateProposer package

### LINKED ISSUE #2622: Add client to SDK to fetch all PerpetualMultiparty contracts
**Difficulty Score [1-10]**

- 4: ~2 hours

### LINKED ISSUE #2623: Add `perpetual` package that includes a funding rate proposer bot
**Difficulty Score [1-10]**

- 3: ~1 hour

### LINKED ISSUE #2624: Perpetual funding rate proposer bot: Should be able to fetch funding rates for each perpetual and propose new ones
**Why**

What is the motivation for this feature or task?

**How**

- Should be able to predict future rewards for proposing new funding rate
- Should be able to settle optimistic oracle proposals that have expired, and then apply the funding rate changes to the perp and collect rewards.

**Difficulty Score [1-10]**

- 7: ~3 days

### LINKED ISSUE #2642: Move `funding-rate-proposer/` + `FinancialContractFactoryEventClient` from Truffle to Hardhat testing
**Why**

Truffle is really slow but makes deploying EMP/Perp factory contracts easy, which require linking to EMP/Perp library contracts respectively.

This is the error I get when I try to run `ExpiringMultiPartyCreator.new()` for example:

```
Error: ExpiringMultiPartyCreator contains unresolved libraries. You must deploy and link the following libraries before
you can deploy a new version of ExpiringMultiPartyCreator: $585a446ef18259666e65e81865270bd4dc$
```

**Difficulty Score [1-10]**

- 5: ~4 hours

nicholaspai:

**Motivation**

#2622
#2623
#2624
#2642

**Summary**

- Adds new event client that tracks `createdExpiringMultiParty` and `createdPerpetual` events emitted by financial contract factories
- Perpetual proposer bot can use this client to get a list of all live perpetuals
- Proposer bot has update function to fetch on-chain state and store it in its state cache, and `updateFundingRates()` method which proposes new rates where appropriate.

Additionally, refactors were made for logic that is shared between bot packages:
- `setAllowance` is now in `financial-templates-lib/helpers/allowance.js`. Used by all bots.
- `computeDeviationError` is now in `financial-templates-lib/helpers/math.js`. Used by SyntheticPegMonitor and RangeTrader.
- `isDeviationOutsideErrorMargin` is now in `financial-templates-lib/helpers/math.js` Used by OptimisticOracleProposer and FundingRateProposer.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2622 
Fixes #2623
Fixes #2624
Fixes #2642


packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,109 @@
...
+      : await this.web3.eth.getBlockNumber();
+
+    // Define a config to bound the queries by.
+    const blockSearchConfig = {
+      fromBlock: this.firstBlockToSearch,
+      toBlock: lastBlockToSearch
+    };
+
+    // Look for events on chain from the previous seen block number to the current block number.
+    const eventToSearchFor = this.contractType === "Perpetual" ? "CreatedPerpetual" : "CreatedExpiringMultiParty";
```
chrismaree, line 75: nice

packages/financial-templates-lib/test/truffle/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,248 @@
+// Note: This is placed within the /truffle folder because EmpCreator/PerpCreator.new() fails due to incorrect library linking by hardhat.
```
chrismaree, line 1: rip.

packages/perpetual/index.js
```diff
@@ -0,0 +1,179 @@
+#!/usr/bin/env node
```
chrismaree, line 1: 2 initial nits:
1) do you want to call the package `Funding-Rate-Proposer` or maybe `Perpetual-Proposer`? IDK, Perpetual does not feel like a good name.
2) do you want to turn this into typescript, seeing this is a new package?

packages/perpetual/index.js
```diff
@@ -0,0 +1,179 @@
...
+  commonPriceFeedConfig,
+  perpetualProposerConfig
+}) {
+  try {
+    const [networkId, accounts] = await Promise.all([web3.eth.net.getId(), web3.eth.getAccounts()]);
+    // If pollingDelay === 0 then the bot is running in serverless mode and should send a `debug` level log.
+    // Else, if running in loop mode (pollingDelay != 0), then it should send a `info` level log.
+    logger[pollingDelay === 0 ? "debug" : "info"]({
+      at: "PerpetualFundingRateProposer#index",
+      message: "Perpetual funding rate proposer started 🌊",
```
chrismaree, line 48: new emoji please.

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
+const { Networker, createReferencePriceFeedForFinancialContract } = require("@uma/financial-templates-lib");
+const { getPrecisionForIdentifier, createObjectFromDefaultProps, MAX_UINT_VAL } = require("@uma/common");
+const { getAbi } = require("@uma/core");
+
+class FundingRateProposer {
+  /**
+   * @notice Constructs new Perpetual FundingRate Proposer bot.
+   * @param {Object} logger Module used to send logs.
+   * @param {Object} perpetualFactoryClient Module used to query for live perpetual contracts
```
chrismaree, line 9:
```suggestion
   * @param {Object} perpetualFactoryClient Module used to query for live perpetual contracts.
```

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
...
+
+    // Validate and set config settings to class state.
+    const configWithDefaults = createObjectFromDefaultProps(perpetualProposerConfig, defaultConfig);
+    Object.assign(this, configWithDefaults);
+  }
+
+  async update() {
+    await Promise.all([this.perpetualFactoryClient.update(), this.gasEstimator.update()]);
+
+    // Create contract object for each perpetual address created.
```
chrismaree, line 75: nice.

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
...
+    }
+
+    // Increase allowances for all relevant collateral currencies.
+    await this._setAllowances();
+  }
+
+  async updateFundingRates() {
+    this.logger.debug({
+      at: "PerpetualProposer#updateFundingRates",
+      message: "Checking for contract funding rates to update"
```
chrismaree, line 92: do you want to add:
```
perpetualsChecked: Object.keys(this.contractCache)
```
to the log to the addresses? might help a bit with debugging.

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
...
+        }
+      }
+    };
+
+    // Validate and set config settings to class state.
+    const configWithDefaults = createObjectFromDefaultProps(perpetualProposerConfig, defaultConfig);
+    Object.assign(this, configWithDefaults);
+  }
+
+  async update() {
```
chrismaree, line 72: do you not need to update the price feeds within `this.priceFeedCache`?

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
...
+      this.logger.error({
+        at: "PerpetualProposer#updateFundingRates",
+        message: "Failed to construct a PriceFeed for funding rate identifier",
+        fundingRateIdentifier
+      });
+      return;
+    }
+
+    // With pricefeed successfully constructed, get the current funding rate
+    await priceFeed.update();
```
chrismaree, line 129: do we not want to group the updates into a `Promise.all` in the `update` method?

packages/perpetual/src/fundingRateProposer.js
chrismaree: I see you do this in line 129 per feed. would we not want to do them all at once in this update method?

packages/perpetual/src/fundingRateProposer.js
```diff
@@ -0,0 +1,223 @@
...
+    this.logger.debug({
+      at: "OptimisticOracleProposer",
+      message: "Created pricefeed configuration for identifier",
+      commonPriceFeedConfig: this.commonPriceFeedConfig,
+      identifier
+    });
+
+    // Create a new pricefeed for this identifier. We might consider caching these price requests
+    // for re-use if any requests use the same identifier.
+    let newPriceFeed = await createReferencePriceFeedForFinancialContract(
```
chrismaree, line 207: this is super fresh. good job.

chrismaree (commented):
This is pretty awesome! great work on getting these done so quickly. I have a few nits on general updating methods and a few others. I will do a second more detailed review once these ones have been cleared.

packages/perpetual/index.js
nicholaspai: 1. Renamed to `funding-rate-proposer`
2. Yes, but I wanted to get this out as quickly as possible. I'm thinking of either converting to TS in a subsequent PR or just biting bullet and doing it now. Do you think TS and truffle work well together?

packages/perpetual/src/fundingRateProposer.js
nicholaspai: This is tough. I think the code would be cleaner if the `update()` for all pricefeeds was called in this `update()` method, but I left them for the `_updateFUndingRate()` method so that the pricefeeds could be updated as close as possible to their usage, to minimize gaps between when prices are fetched and whent hey are compared against on chain prices.

packages/perpetual/src/fundingRateProposer.js
nicholaspai: see [this comment](https://github.com/UMAprotocol/protocol/pull/2632#discussion_r586364110)

nicholaspai:
@UMAprotocol/eng I've updated this PR to be less of an MVP and more of a fully fleshed-out new bot.

The main frustration I've been having is that using the `emp/perpCreators` in hardhat has given me trouble because of linked library issue. I'm not sure if there is an easy fix to make these tests run in hardhat, but perhaps that should be left for the next PR.

Here's a comment I've left in the tests:

```
// Note: This is placed within the /truffle folder because EmpCreator/PerpCreator.new() fails due to incorrect library linking by hardhat.
// `Error: ExpiringMultiPartyCreator contains unresolved libraries. You must deploy and link the following libraries before
//         you can deploy a new version of ExpiringMultiPartyCreator: $585a446ef18259666e65e81865270bd4dc$`
// We should look more into library linking via hardhat within a script: https://hardhat.org/plugins/hardhat-deploy.html#handling-contract-using-libraries
```

nicholaspai:
> @UMAprotocol/eng I've updated this PR to be less of an MVP and more of a fully fleshed-out new bot.
> 
> The main frustration I've been having is that using the `emp/perpCreators` in hardhat has given me trouble because of linked library issue. I'm not sure if there is an easy fix to make these tests run in hardhat, but perhaps that should be left for the next PR.
> 
> Here's a comment I've left in the tests:
> 
> ```
> // Note: This is placed within the /truffle folder because EmpCreator/PerpCreator.new() fails due to incorrect library linking by hardhat.
> // `Error: ExpiringMultiPartyCreator contains unresolved libraries. You must deploy and link the following libraries before
> //         you can deploy a new version of ExpiringMultiPartyCreator: $585a446ef18259666e65e81865270bd4dc$`
> // We should look more into library linking via hardhat within a script: https://hardhat.org/plugins/hardhat-deploy.html#handling-contract-using-libraries
> ```

Turns out I was able to move all tests to hardhat pretty easily by importing the `truffle-fixture.js` file from `@uma/core`

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,109 @@
...
+   * @return None or throws an Error.
+   */
+  constructor(
+    logger,
+    financialContractFactoryAbi,
+    web3,
+    financialContractFactoryAddress,
+    startingBlockNumber = 0,
+    endingBlockNumber = null,
+    contractType = "Perpetual"
```
mrice32, line 21: One quick note: we could make this the name of the factory contract (ExpiringMultiPartyCreator or PerpetualCreator). Then that could be pulled directly from the artifact. Not strongly opinionated, but that seems a bit more obvious since we call these contracts different names (EMP vs ExpiringMultiParty vs ExpiribgMultiPartyCreator).

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,109 @@
...
+    this.createdContractEvents = [];
+
+    // First block number to begin searching for events after.
+    this.firstBlockToSearch = startingBlockNumber;
+
+    // Last block number to end the searching for events at.
+    this.lastBlockToSearchUntil = endingBlockNumber;
+    this.lastUpdateTimestamp = 0;
+
+    this.contractType = contractType;
```
mrice32, line 44: Should we check that the input is a valid type? What happens if someone passes "Perpteual"?

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+  // Delete all events within the client
+  async clearState() {
+    this.createdContractEvents = [];
+  }
+
+  getAllCreatedContractEvents() {
```
mrice32, line 51: Do you think it might be nice to replace this (or have a separate method) that returns just a list of deployed addresses rather than the full event?

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,109 @@
...
+    return this.createdContractEvents;
+  }
+
+  // Returns the last update timestamp.
+  getLastUpdateTime() {
+    return this.lastUpdateTimestamp;
+  }
+
+  async update() {
+    // The last block to search is either the value specified in the constructor (useful in serverless mode) or is the
```
mrice32, line 61: Will we use this this way? Does this mean we'll only get contracts that are deployed within the last block window? Or are we effectively going to be searching from block 0 in serverless mode?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+    // use in `updateFundingRates()` such as the latest funding rates and config store settings.
+    await this._cachePerpetualContracts();
+
+    // The following updates can be done independently after contract state is fetched:
+    await Promise.all([
+      // Increase allowances for all contracts to spend the bot owner's respective collateral currency.
+      this._setAllowances(),
+      // For each contract, create, save, and update a pricefeed instance for its identifier,
+      // or just update the pricefeed if the identifier already has cached an instance.
+      await this._cacheAndUpdatePriceFeeds()
```
mrice32, line 97:
```suggestion
      this._cacheAndUpdatePriceFeeds()
```

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+      if (receipt) {
+        this.logger.info({
+          at: "PerpetualProposer",
+          message: "Approved Perpetual contract to transfer unlimited collateral tokens 💰",
+          perpetual: receipt.spenderAddress,
+          currency: receipt.currencyAddress,
+          collateralApprovalTx: receipt.tx.transactionHash
+        });
+      }
+    });
```
mrice32, line 275:
```suggestion
    await Promise.map(Object.keys(this.contractCache), async contractAddress => {
      // The Perpetual requires approval to transfer the contract's collateral currency in order to post a bond.
      // We'll set this once to the max value and top up whenever the bot's allowance drops below MAX_INT / 2.
     const receipt = await setAllowance(
        this.web3,
        this.gasEstimator,
        this.account,
        contractAddress,
        this.contractCache[contractAddress].collateralAddress
      );
      if (receipt) {
        this.logger.info({
          at: "PerpetualProposer",
          message: "Approved Perpetual contract to transfer unlimited collateral tokens 💰",
          perpetual: receipt.spenderAddress,
          currency: receipt.currencyAddress,
          collateralApprovalTx: receipt.tx.transactionHash
        });
      }
    });
```

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  async _cacheAndUpdatePriceFeeds() {
+    for (let contractAddress of Object.keys(this.contractCache)) {
```
mrice32, line 279: nit: thoughts on doing this as a big `Promise.map` so we can run these inner async calls in parallel?

mrice32 (commented):
Flushing some comments. Will continue working through.

packages/perpetual/index.js
mrice32: On the typescript question, I think I would be in favor of starting with ts if we hadn't already written most of the code. However, since it's already written in js, I would say we should just _convert_ to ts subsequently in the name of unblocking the perp rollout.

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+        priceFeed = await createReferencePriceFeedForFinancialContract(
+          this.logger,
+          this.web3,
+          new Networker(this.logger),
+          () => Math.round(new Date().getTime() / 1000),
+          null, // No EMP Address needed since we're passing identifier explicitly
+          this.commonPriceFeedConfig,
+          fundingRateIdentifier
+        );
+        this.priceFeedCache[fundingRateIdentifier] = priceFeed;
```
mrice32, line 303: What if this comes back null (because we don't know how to construct this pf)? Should we send an error log? Do we just expect the bot to crash? It seems like we should probably just fire off an error log and keep going?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+        maxFundingRate,
+        offchainFundingRate
+      });
+      return;
+    }
+
+    // If the saved funding rate is not equal to the current funding rate within margin of error, then
+    // prepare request to update. We're assuming that the `offchainFundingRate` is the baseline or "expected"
+    // price.
+    let isPriceDisputable = isDeviationOutsideErrorMargin(
```
mrice32, line 177: isPriceDisputable seems like the wrong name for this variable...? shoulduUdateFundingRate, maybe?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,360 @@
...
+        message: "Potential proposed funding rate is outside allowed funding rate range",
+        fundingRateIdentifier,
+        minFundingRate,
+        maxFundingRate,
+        offchainFundingRate
+      });
+      return;
+    }
+
+    // If the saved funding rate is not equal to the current funding rate within margin of error, then
```
mrice32, line 174: Probably don't need to do anything for this PR, but it could make sense for us to use a time/or ROI metric to determine how often to propose new funding rates. I think the current code makes sense for a bot that _doesn't care about profit_, but a profit-seeking bot would only care how many rewards they would get. This is aligned with UMA-run bots, but may not be as interesting for 3rd party bot runners (which may actually be a feature, not a bug since we'd like them to write their own bots!)

mrice32 (commented):
Looks awesome! A few more minor comments.

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: This is a great point, I'm thinking the generalized solution is to make the user of this client (i.e. the FundingRateProposer), just set `startingBlock = 0` when constructing this client.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: This is an async method because `createRefernecePriceFeed...` is async

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Yes, but I don't really see a difference for now. If we find that we need this more (perhaps for the monitor) its easy enough to add the method. For now its just simpler to not do any filtering on the raw event data

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Definitely

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Done good catch

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Perhaps we should rename this client to just a `FinancialContractFactoryClient` and we can use it to precompute some other contract data, such as `fundingRateIdentifier`, `collateralCurrency`, etc., most of which we do in the `funding-rate-proposer/proposer/src` code right now. WDYT?

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Nice catch definitely will do

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Done.

packages/funding-rate-proposer/index.js
```diff
@@ -0,0 +1,180 @@
...
+      // Default to 1 minute delay. If set to 0 in env variables then the script will exit after full execution.
+      pollingDelay: process.env.POLLING_DELAY ? Number(process.env.POLLING_DELAY) : 60,
+      // Default to 3 re-tries on error within the execution loop.
+      errorRetries: process.env.ERROR_RETRIES ? Number(process.env.ERROR_RETRIES) : 5,
+      // Default to 10 seconds in between error re-tries.
+      errorRetriesTimeout: process.env.ERROR_RETRIES_TIMEOUT ? Number(process.env.ERROR_RETRIES_TIMEOUT) : 10,
+      // Common price feed configuration passed along to all those constructed by proposer.
+      commonPriceFeedConfig: process.env.COMMON_PRICE_FEED_CONFIG
+        ? JSON.parse(process.env.COMMON_PRICE_FEED_CONFIG)
+        : { lookback: 7200 },
```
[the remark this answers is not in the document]
nicholaspai: Hm should we? Sholdn't it be inferred from the OptimisticOracle's proposal liveness? Regardless I'll add a TODO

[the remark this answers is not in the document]
nicholaspai: Actually, I'm not sure there's any sort of lookback that makes sense, because the funding rate proposer is always checking out the current price of the funding rate, never historical. WDYT

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Yeah I think we should skip setting it in the cache during this method, and then emit an error log in the `updateFUndingRates()` call

packages/funding-rate-proposer/src/proposer.js
nicholaspai: I'll add an issue!

packages/funding-rate-proposer/src/proposer.js
mrice32: +1, that sounds like the right way to do it

mrice32: You don't even need to skip setting it here since it'll set it to null, and the other method should be able to recognize that null.

packages/funding-rate-proposer/index.js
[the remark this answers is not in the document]
mrice32: Yep, you're right. Nice catch. We probably don't want to set it to 0 or anything, so just leaving the default seems right.

packages/funding-rate-proposer/src/proposer.js
mrice32: But it's being called inside a `Promise.all`, no? _setAllowances above is also async, but you don't put an await in front of it because the promise.all handles that if I'm reading this correctly.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: - I added an `if(pricefeed)` before the subsequent `await pricefeed.update()`
- I catch `!pricefeed` in the `updateFundingRates()` method and emit error log (and exit early) if no pricefeed.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: oh true, done

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
mrice32: I see. So this is modeled after the other event clients. I could imagine it pulling down other contract metadata, but I could also imagine that being overkill for some/most of the other use cases.

The reason I think it could be nice to return a list of contract addresses is that that seems pretty universal. Put another way, if we wanted a liquidator bot to monitor _all_ emps, it would use this client for the same purpose: grabbing all addresses. Some other metadata may be useful in some cases, but certainly the fundingRateIdentifier wouldn't be needed by the liquidator bot, for example, and it might be a waste to query for that.

Still not strongly opinionated -- just a thought about generalizability.

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
mrice32: Yeah, I actually think this could have a cool double-use. In the monitor, we may want to know about _new_ EMPs. Here, we want to know about _all_ EMPs.

packages/funding-rate-proposer/index.js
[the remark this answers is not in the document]
mrice32: Bump.

packages/funding-rate-proposer/src/proposer.js
mrice32: Bump.

packages/funding-rate-proposer/src/proposer.js
mrice32: Bump.

packages/funding-rate-proposer/src/proposer.js
mrice32: Bump.

packages/funding-rate-proposer/src/proposer.js
mrice32: Bump.

mrice32 (approved):
Looks great! I put Bump on all the threads that I think still need to be resolved. Otherwise, LGTM!

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,118 @@
...
+    this.createdContractEvents = [];
+
+    // First block number to begin searching for events after.
+    this.firstBlockToSearch = startingBlockNumber;
+
+    // Last block number to end the searching for events at.
+    this.lastBlockToSearchUntil = endingBlockNumber;
+    this.lastUpdateTimestamp = 0;
+
+    if (contractType != "ExpiringMultiPartyCreator" && contractType != "PerpetualCreator")
```
mrice32, line 44:
```suggestion
    if (contractType !== "ExpiringMultiPartyCreator" && contractType !== "PerpetualCreator")
```

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
mrice32: Nothing necessarily needs to be changed here. Just want to make sure you see this thread.

packages/funding-rate-proposer/src/proposer.js
mrice32: Ignore this^. LGTM

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: > I see. So this is modeled after the other event clients. I could imagine it pulling down other contract metadata, but I could also imagine that being overkill for some/most of the other use cases.

Yeah problem would be if a user just wants contract addresses and no state, precomputing the state would slow the address query down.

I ultimately agree with you that this methid would be useful to return all addresses, so I'll rename this client to signify that its not just an event client, and add the method.

packages/disputer/test/index.js
```diff
@@ -32,7 +32,7 @@ let pollingDelay = 0; // 0 polling delay creates a serverless bot that yields af
 let errorRetries = 1;
 let errorRetriesTimeout = 0.1; // 100 milliseconds between preforming retries
 let identifier = "TEST_IDENTIFIER";
-let fundingRateIdentifier = "TEST_FUNDiNG_IDENTIFIER";
```
chrismaree, line 35: GoOd CaTcH.

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,118 @@
...
+   */
+  constructor(
+    logger,
+    financialContractFactoryAbi,
+    web3,
+    financialContractFactoryAddress,
+    startingBlockNumber = 0,
+    endingBlockNumber = null,
+    contractType = "PerpetualCreator"
+    // Default to PerpetualMultiParty for now since the first intended user is the funding rate proposer bot
```
chrismaree, line 22:
```suggestion
    // Default to PerpetualMultiParty for now since the first intended user is the funding rate proposer bot.
```

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
```diff
@@ -0,0 +1,118 @@
...
+    // First block number to begin searching for events after.
+    this.firstBlockToSearch = startingBlockNumber;
+
+    // Last block number to end the searching for events at.
+    this.lastBlockToSearchUntil = endingBlockNumber;
+    this.lastUpdateTimestamp = 0;
+
+    if (contractType != "ExpiringMultiPartyCreator" && contractType != "PerpetualCreator")
+      throw new Error(
+        `Invalid contract type provided: ${contractType}! The financial product factory client only supports ExpiringMultiPartyCreator or PerpetualCreator`
```
chrismaree, line 46:
```suggestion
        `Invalid contract type provided: ${contractType}! The financial product factory event client only supports ExpiringMultiPartyCreator or PerpetualCreator`
```

packages/financial-templates-lib/src/helpers/allowances.js
```diff
@@ -0,0 +1,26 @@
+const { MAX_UINT_VAL } = require("@uma/common");
+const { getAbi } = require("@uma/core");
+
+// Sets `owner` allowance for `spender` to MAX, unless `spender` already has an allowance > 1/2 of the MAX_UINT.
+// Return array of successful approval transactions or undefined for skipped approvals.
```
chrismaree, line 5: this comment implies this can do multiple approvals in one go. Wont it only do one at a time?

packages/financial-templates-lib/src/helpers/math.js
```diff
@@ -0,0 +1,37 @@
+const assert = require("assert");
```
chrismaree, line 1: 🤩 

Will be more to add to this in the future for sure. thanks for the initial refactor!

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -557,30 +557,6 @@ const defaultConfigs = {
     minuteLookback: 7200,
     hourlyLookback: 604800,
     minTimeBetweenUpdates: 60
-  },
```
chrismaree, line 560: why was uSTONKS removed?

packages/financial-templates-lib/test/truffle-fixture.js
```diff
@@ -0,0 +1,2 @@
+// Export the truffle-fixture file from `/core/test` so that we can use the `.deployed()` syntax in our tests.
```
chrismaree, line 1: where are you using this? are you using this in other packages? IMHO this is something we should strictly _not_ do. This somewhat defeats a lot of the gains we get from using hardhat. You dont want to re-run all the migrations for a test and rather deploy what you need in the tests. I think this should be the pattern that we follow for all packages that are _not_ `core`. This is how other packages are set up right now. Liquidator/disputer/monitor dont rely on any truffle migrations for this exact reason.

chrismaree: Looking further down, I see you are doing things like
```
timer = await Timer.deployed();
perpFactory = await PerpetualCreator.deployed();
```

IMHO you should refactor this to _not_ be dependent on the core migrations. this will drastically speed up your unit tests and follow the patterns in the other non-core packages.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: This is done.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Done

packages/funding-rate-proposer/src/proposer.js
chrismaree: ye, agreed. could be tuned via a parameter that defines the "profit seeking" threshold of the bot. we'd just set this to zero but it can default for others to only do this in profitable contexts.

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,370 @@
...
+        };
+        this.logger.info({
+          at: "PerpetualProposer#updateFundingRate",
+          message: "Proposed new funding rate!🌻",
+          proposalBond: returnValue,
+          proposalResult: logResult
+        });
+      } catch (error) {
+        const message =
+          error.type === "call"
```
chrismaree, line 237: nice

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -0,0 +1,370 @@
...
+        message: "Skipping proposal because current rate is within allowed margin of error",
+        fundingRateIdentifier,
+        proposedRate: offchainFundingRate,
+        currentRate: onchainFundingRate,
+        allowedError: this.fundingRateErrorPercent
+      });
+    }
+  }
+  // Sets allowances for all collateral currencies used live perpetual contracts.
+  async _setAllowances() {
```
chrismaree, line 260: I like this

chrismaree (approved):
This is a beast of a PR! excellent work on getting it out so quickly and with such high quality, well commented code!

LGTM, granting you make a refactor I requested on the migration-reuse from core.

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Done.

packages/funding-rate-proposer/index.js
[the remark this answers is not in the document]
nicholaspai: I removed this

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Done

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Done

packages/financial-templates-lib/src/clients/FinancialContractFactoryEventClient.js
nicholaspai: Actually ended up renaming this to just factory client, no event

packages/financial-templates-lib/src/helpers/allowances.js
nicholaspai: Yup you're right, will fix comment

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: Oops I'll merge this back in! Great catch

packages/financial-templates-lib/test/truffle-fixture.js
nicholaspai: @chrismaree I need this because the `EMP/PerpCreator` contracts can't be deployed unless I link their their EMPLib/PerpLib contracts which I can't seem to figure out how to do in hardhat

nicholaspai: Ended up fixing this by adding `await PerpetualCreator.link(await PerpetualLib.new())`

nicholaspai:
@chrismaree @mrice32 I've addressed all of your comments PTAL when you have time! Any leftover issues are collected in this [project board card](https://github.com/UMAprotocol/protocol/projects/4#card-46767632)

mrice32 (approved):
LGTM!

PR #2640, opened 2021-03-03
feat(voter-dapp): fetch users 2key contract state, and show migration banner if required

daywiss:
**Motivation**

Contract upgrades have required a 2key upgrade as well, and we will need to warn users who have not migrated to the newest 2key, to do so. 

**Summary**

The dashboard now has logic which looks up if a user has UMA tokens at the old 2key contract, if so, a banner component has been added which directs the user to the migration guide.

**Details**

This is an example of what it looks like, in production it will not show if a user has 0 uma at address
image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


nicholaspai (approved):
Seems like it does the job, lgtm

pemulis (approved):
LGTM! I saw the banner for my address using a 2-key wallet and didn't see the banner for a regular address.

PR #2643, opened 2021-03-04
feat(Merkle-distributor) Add scripts, functions and helpers for dealing with Merkle proofs

chrismaree:
**Motivation**

As part of the KPI options initiative, we require some scripts and helper functions to aid in accessing merkle information. This PR adds a simple package to the protocol repo, `merkle-distributor`, that includes a number of useful functions, scripts and helpers. There include:

**Summary**

There are two main components to this PR:
1) Useful scripts for producing Merkle trees from recipient files. These include adding Merkle proofs to IPFS and Cloudflare as well as uploading the Merkle hash to the `MerkleDistributor.sol` smart contract.
2) Helper functions to be used in the front end such as `getClaimsForAddress` which will return all claims objects for a given address.

**Details**

Files added to IPFS contain the full payout for a particular set of recipients & window, represented as a stringified json file. The file added to IPFS is produced using the `1_CreateClaimsForWindow.ts` file. An example of the payout object can be seen below:
```
{
  "chainId": 42,
  "rewardToken": "0x47B1EE6d02af0AA5082C90Ea1c2c14c70399186c",
  "windowIndex": 0,
  "totalRewardsDistributed": "15000000000000000000",
  "windowStart": 1614850539,
  "merkleRoot": "0x1b1b3e8a64b815fe4dd6f42c11dae9e9ff9ceb3d8f295052bc0c02a326aa9dc2",
  "claims": {
    "0x00b591bc2b682a0b30dd72bac9406bfa13e5d3cd": {
      "amount": "1000000000000000000",
      "metaData": { "reason": ["YD-WETH-21 Liquidity Mining Week 27"] },
      "windowIndex": 0,
      "proof": [
        "0x62465ff033d171eb8bdeee2601be357027af3b7bea19d422b3797ad5c803f576",
        "0xf0987f2e0ec07c35d7cfdd4f5831d8ef2a1829413321611caecbf08eaddf567c",
        "0xedd0b39fe99a555994df53dfca180b84437abbd1b808fbadd5e8a70f7f2a40cf"
      ]
    },
    ... for all recipients
  }
}
```

The key components in this file are the `merkleRoot` which is the string added on-chain to verify the proofs and within the `claims` you can find the `proof` which each recipient adds to.

The outputs of this file are then ingested by the `2_PublishClaimsForWindow.ts` script. This script does the following:
1) Adds the files to IPFS. For a full version of the IPFS output, viewable through an online gateway, see [here](https://cloudflare-ipfs.com/ipfs/QmfVMHgoWpTSZqovo7vhM7Wcmz6EeX4QBYbCk4DZTNM8u3) which has a test output file.
2) Pins the files the Infura & Pinata IPFS servers.
3) Uploads the proofs data in a structured format to Cloudflare KV. Cloudflare KV is a fast, distributed key value store that enables the front end to fetch information about a particular recipient in constant time. More info on the KV setup can be found below.
4) Checks sender allowance in reward tokens.
5) Uploads the Merkle root to the smart contract. A sample Kovan test contract can be found [here](https://kovan.etherscan.io/address/0xAfCd2405298C2FABB2F7fCcEB919B4505A6bdDFC).

**Cloudflare KV setup**
Each key added KV is the `${chainId}:${windowIndex}:${account}` for a particular recipient. The value is the full stringified object, shown within the json object above within the `claims` section for a particular account. This structure enables the querier to search for the `amount`, `metaData`, `windowIndex` and `proof` for a particular `chainId`, `windowIndex` and `account` for a particular recipient. An example from Cloudflare can be seen below:
image

Secondly, KV also stores a mapping between `${chainId}` and an object containing information about all windows added to that chain. This enables the user of KV to first search for a given `chainId` to fetch all posible window indices and then search against the second the `${chainId}:${windowIndex}:${account}` key to find all proofs over the set of `windowIndex`s. An example of this can be seen below:
image

The combination of these two data structures makes fetching all information about a particular window, and then combining this information to find all proofs for a particular `account`, simple and fast.

**Additional helper methods**
There second part of this PR is the addition of helper scripts. The main helper script is `MerkleDistributorHelper.ts`. This script provides two main methods that can be consumed when working with merkle proofs:

1. `createMerkleDistributionProofs(recipientsData, windowIndex: number)` takes in an object mapped between recipient address, amount and metaData and returns the provided data with appended merkle proof and the merkleRoot of the merkle tree. This is used in the scripts described above.
2. `getClaimsForAddress(merkleDistributorAddress, claimerAddress, chainId)` takes in the address of the merkleDistributor, a claimerAddress and a chainID and returns a data structure containing all information for the claimer on the provided chainId, including the merkkleproofs, if the rewards has been claimed and additional information about the each claim window such as the IPFS hash for the claims file. See the code for exactly how this works.

**Integration of theses helpers into the front end**
Note that these methods are designed to run in node exclusively and will not run in the browser. To access the `getClaimsForAddress` in the browser, it is recommended to wrap this method in a serverless function, such as [Vercel](https://vercel.com/docs/serverless-functions/introduction). This will protect your cloudflare API keys and enable some ingress verification. This would work by simply creating a basic Serverless function that wraps the node `getClaimsForAddress` function and can be called by the claims dapp to return all posible claims for a particular wallet.

**Some additional points to note**
There are two main points of contention I'd like feedback from @UMAprotocol/eng on.
1) This PR adds `merkleTree.js` to the unit tests for the `Merkle-distributor` in `core`. This is effectively duplicated code as that found in the `Merkle-distributor` package. As the core unit tests are written in javascript, we cant import the typscript version of this file into the unit tests. Once core has migrated over to typescript this file can be removed. I spent quite some time trying to get the imports to work but could not. I want to check with eng that this is ok and that we are happy with this stop gap solution.
2) I added an additional field to the `MerkleDistributor.sol` contract to store the IPFS hash. My reasoning with this is that I want recipients to be able to claim rewards by simply looking at the smart contracts. By having the IPFS hash, they can always: a) find their proof b) re-construct their proof & Merkle root & c) claim reward tokens without needing anything other than the smart contract and etherscan.

LMK what you guys think on the above 2 points ^^

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/merkle-distributor/hardhat.config.ts
```diff
@@ -0,0 +1,20 @@
+import "@nomiclabs/hardhat-web3";
+import "@nomiclabs/hardhat-truffle5";
+
+import { getHardhatConfig } from "@uma/common";
+import path from "path";
+
+const coreWkdir = path.dirname(require.resolve("@uma/core/package.json"));
+const packageWkdir = path.dirname(require.resolve("@uma/optimistic-oracle/package.json"));
```
nicholaspai, line 8:
```suggestion
const packageWkdir = path.dirname(require.resolve("@uma/merkle-distributor/package.json"));
```

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,84 @@
...
+    throw new Error(`recipients object missing expected key: ${expectedKey}`);
+  }
+});
+
+async function main() {
+  console.log("Running claims creation script 🧞‍♂️");
+
+  // Do some basic sanity checks. In particular, verify that totalRewardsDistributed equals the sum of all amounts.
+  let totalCalculatedRewards = toBN("0");
+  Object.keys(recipientsObject.recipients).forEach((recipientAddress: any) => {
```
nicholaspai, line 54: Do we want to handle when `recipientAddress` is duplicated in the `recipientsObject` and consolidate those payouts? Or should we just assume that the creator of that JSON file "knows what they are doing" and can choose to make duplicate payouts.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,126 @@
...
+  console.log("0. Running some basic sanity checks on the payout file 🔎");
+  assert(web3.utils.isAddress(options.merkleDistributorAddress), "Invalid merkleDistributorAddress");
+  assert((await merkleDistributor.methods.owner().call()) == account, "Unlocked account does not own the distributor");
+  assert((await web3.eth.net.getId()) == claimsObject.chainId, "The connected network not match your  JSON chainId");
+  assert(web3.utils.isAddress(claimsObject.rewardToken), "Invalid rewardToken" + claimsObject.rewardToken);
+  assert((await merkleDistributor.methods.lastSeededIndex().call()) == claimsObject.windowIndex, "Wrong windowIndex");
+
+  console.log("File passed the checks\n\n1. Adding claims file to IPFS 🛰");
+
+  const ipfsHash = await IpfsHelper.uploadFile(JSON.stringify(claimsObject.claims));
```
nicholaspai, line 79: How long does this method take? If a long time, maybe log out some other info to track the async transaction?

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  console.log("Claims file added to IPFS with hash:", ipfsHash, "\n\n2. Pinning claims file to IPFS 📌");
+
+  await IpfsHelper.pinHash(ipfsHash);
```
nicholaspai, line 83: What's the difference between pinning and uploading? Would be great to add to comments too

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  console.log("Claims file pinned to IPFS!\n\n3. Adding claims file to cloudflare KV 🗺");
+
+  await CloudflareKVHelper.addClaimsToKV(claimsObject.claims, claimsObject.chainId, claimsObject.windowIndex);
+
+  console.log("Claims file added to cloudflareKV!\n\n4. Updating the lookup claims indices 🔭");
+
+  await CloudflareKVHelper.updateChainWindowIndicesFromKV(
```
nicholaspai, line 91: If one of the following steps breaks, is there any clean up thats neccessary to do or is it safe to re-upload or pin IPFS files for example:

1. Upload file to IPFS
2. Pin file hash to IPFS
3. Upload claims to cloudflare
4. Update look claims indices on cloudflare

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,126 @@
...
+
+  console.log("5. Checking allowance in payment token 💸");
+  const rewardToken = new web3.eth.Contract(getAbi("ExpandedERC20"), claimsObject.rewardToken);
+  const currentRewardAllowance = await rewardToken.methods.allowance(account, options.merkleDistributorAddress).call();
+
+  if (toBN(currentRewardAllowance).lt(toBN(MAX_UINT_VAL).div(toBN("2")))) {
+    console.log("Reward token allowance is too little to make this payment. Sending a transaction to increase it...");
+    const approveTx = await rewardToken.methods
+      .approve(options.merkleDistributorAddress, MAX_UINT_VAL)
+      .send({ from: account });
```
nicholaspai, line 108: Should we use the gas estimator to set `gasPrice`?

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,126 @@
...
+    claimsObject.rewardToken,
+    claimsObject.totalRewardsDistributed,
+    claimsObject.windowStart
+  );
+
+  console.log("5. Checking allowance in payment token 💸");
+  const rewardToken = new web3.eth.Contract(getAbi("ExpandedERC20"), claimsObject.rewardToken);
+  const currentRewardAllowance = await rewardToken.methods.allowance(account, options.merkleDistributorAddress).call();
+
+  if (toBN(currentRewardAllowance).lt(toBN(MAX_UINT_VAL).div(toBN("2")))) {
```
nicholaspai, line 104: Maybe just compare `currentRewardAllowance` to `claimsObject.totalRewardsDistributed`

packages/merkle-distributor/scripts/example.json
```diff
@@ -0,0 +1,41 @@
...
+        "reason": [
+          "YD-WETH-21 Liquidity Mining Week 27",
+          "Governance Votes KPI Options Round 1",
+          "Token Minter KPI Options Round 1"
+        ]
+      }
+    },
+    "0x01b01c6fb158e55c83883219a533f1efd2abfdf4": {
+      "amount": "5000000000000000000",
+      "metaData": { "reason": ["Governance Votes KPI Options Round 1"] }
```
nicholaspai, line 38: Do you set `metaData` in any of the scripts or is this done offline?

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+) {
+  const KV = Object.keys(recipientsData).map(account => {
+    const claim = recipientsData[account];
+    return {
+      key: `${chainId}:${windowIndex}:${account}`,
+      value: JSON.stringify(claim)
+    };
+  });
+
+  const BATCH_SIZE = 10_000; // limit how many wee add in each bulk upload. Cloud flare limits 10k per bulk put.
```
nicholaspai, line 45:
```suggestion
  const BATCH_SIZE = 10_000; // limit how many we add in each bulk upload. Cloud flare limits 10k per bulk put.
```

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -30,10 +30,12 @@ contract MerkleDistributor is Ownable, Testable {
...
         // Claims for this window can begin after `start`.
         uint256 start;
-        // Merkle root describing the distribution.
-        bytes32 merkleRoot;
         // Currency in which reward is processed.
         IERC20 rewardToken;
+        // Merkle root describing the distribution.
+        bytes32 merkleRoot;
+        // IPFS hash of the merkle tree. Can by used to independently fetch recipient proofs.
+        string ipfsHash;
```
nicholaspai, line 38: I'm probably missing this in an obvious place but where does the value come from adding `ipfsHash` to the onchain struct?

nicholaspai (approved):
Really awesome work + well tested, left mostly nits and questions

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+      error: new Error(error)
+    };
+  }
+}
+
+async function pinHash(hashToPin: string) {
+  const infuraResponse = await nodeFetch(`https://ipfs.infura.io:5001/api/v0/pin/add?arg=${hashToPin}`);
+  if (infuraResponse.status != 200) throw { message: "Failed to pin on infura", error: new Error(infuraResponse) };
+
+  if (process.env.PINATA_API_KEY && process.env.PINATA_SECRET_API_KEY) {
```
daywiss, line 30: again prefer not to reference env in there funcs.

packages/merkle-distributor/src/MerkleDistributorHelper.ts
```diff
@@ -0,0 +1,83 @@
...
+
+  return { recipientsDataWithProof, merkleRoot: merkleTree.getHexRoot() };
+}
+
+async function getClaimsForAddress(merkleDistributorAddress: string, claimerAddress: string, chainId: number) {
+  // Create a new ethers contract instance to fetch on-chain contract information.
+  const infuraApiKey = process.env.INFURA_API_KEY || null;
+  const ethersProvider = new ethers.providers.InfuraProvider(chainId, infuraApiKey);
+  const signer = new ethers.VoidSigner(claimerAddress, ethersProvider);
+  const merkleDistributorContract = new ethers.Contract(merkleDistributorAddress, getAbi("MerkleDistributor"), signer);
```
daywiss, line 43: can we pass in the provider as a dep to this function? also im not sure you need this void signer to call the contract, you can just [pass in the provider](https://docs.ethers.io/v5/api/contract/example/#example-erc-20-contract--connecting-to-a-contract--erc20contract)

daywiss (commented):
did not get a chance to finish reviewing, but wanted to get my comments out! looks good so far!

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
chrismaree: I think that it should be less opinionated and any kind of reduction/joining should happen upstream.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
chrismaree: takes like 0.5 seconds. dont think it needs any better logging imho.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
chrismaree: once it is uploaded the pinning makes sure it is persistently accessible on the pinning services.

I have added a comment.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
chrismaree: If it breaks we can do the remaining steps manually pretty easily. As long as the merkle distribution(created in script 1) works as expected, we can do all the other steps on our own. This is the main reason I split the script.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
chrismaree: We could totally do that. There are 2 transactions in this class, both of which could use the gas estimator.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
chrismaree: This would mean needing to re-approve every time. I opted for this method to ensure that we don't have to send multiple approval transactions. I am happy to do it that way if you think it better.

packages/merkle-distributor/scripts/example.json
chrismaree: All information in this file is generated pre-generated from a separate script. This is just an example to show what you can put in `metaData` any arbitrary data type.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
chrismaree: It is produced by the IPFS upload step; this returns the IPFS hash. Take a look in the `2_PublishClaimsForWindow.ts`. In this script it is the last param in the `setWindow` function call.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -105,12 +107,13 @@ contract MerkleDistributor is Ownable, Testable {
         uint256 totalRewardsDistributed,
         uint256 windowStart,
         address rewardToken,
-        bytes32 merkleRoot
+        bytes32 merkleRoot,
+        string memory ipfsHash
```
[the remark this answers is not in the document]
chrismaree: Hhhhmm. so the IPFS hash is really just a hash of the payout window which is strictly bound to that window. If we put it in a no-sql db (for example) we would still be able to add a hash of the payouts, thereby proving the correctness between some payouts and the reward window. We can totally just forgo this string in its entirety if we wanted to.

My reasoning with this is that I want recipients to be able to claim rewards by simply looking at the smart contracts. By having the IPFS hash, they can always: a) find their proof b) re-construct their proof & Merkle root & c) claim reward tokens without needing anything other than the smart contract, etherscan and access to IPFS.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,84 @@
+// Read in a json file containing information for a particular window distribution and produce a claims file containing
+// merkle proofs and a merkle root. This input file is produced from the liquidity/developer/KPI outputs output scripts.
+// The output of this script is then fed into the PublishClaimsForWindow.ts script to add claims to IPFS and Ethereum.
+
+// The input file should implement the following formatting:
+// {
+// "chainId": 42,
+// "rewardToken": "0x47B1EE6d02af0AA5082C90Ea1c2c14c70399186c",
+// "windowIndex": 0,
+// "totalRewardsDistributed": "15000000000000000000",
```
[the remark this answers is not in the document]
chrismaree: It's really just included in the outputs to inform the front end of extra information about the total set of payouts. In the case of precision loss then there should be rounding to accommodate this. 

In the script it actually doubles checks that this total matches the sum of all payouts.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,84 @@
+// Read in a json file containing information for a particular window distribution and produce a claims file containing
+// merkle proofs and a merkle root. This input file is produced from the liquidity/developer/KPI outputs output scripts.
+// The output of this script is then fed into the PublishClaimsForWindow.ts script to add claims to IPFS and Ethereum.
+
+// The input file should implement the following formatting:
+// {
+// "chainId": 42,
+// "rewardToken": "0x47B1EE6d02af0AA5082C90Ea1c2c14c70399186c",
+// "windowIndex": 0,
```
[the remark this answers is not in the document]
chrismaree: The window is double-checked to match what is expected in the second script. it will error out if it is not what is expected.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
chrismaree: At the end of the day the IPFS hash is just a hash of all information associated with a payout window.

My reasoning with this is that I want recipients to be able to claim rewards by simply looking at the smart contracts. By having the IPFS hash, they can always: a) find their proof b) re-construct their proof & Merkle root & c) claim reward tokens without needing anything other than the smart contract, etherscan & IPFS.

We can remove this; it is totally optional and adjacent to the primary function of this contract. I just thought it was more "complete" to have this included.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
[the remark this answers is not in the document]
chrismaree: the script does a sanity check to ensure that the totals match. This is really just included for convenience in the front end and is totally optional.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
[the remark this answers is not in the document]
chrismaree: it totally could be. it can also be pulled from the contract. Right now I included it to ensure consistency between our expected index and the true index. in the second script, the on-chain index is pulled and compared to the file and triple checked.

packages/merkle-distributor/scripts/example.json
```diff
@@ -0,0 +1,41 @@
...
+  "totalRewardsDistributed": "15000000000000000000",
+  "windowStart": 1614850539,
+  "recipients": {
+    "0x00b591bc2b682a0b30dd72bac9406bfa13e5d3cd": {
+      "amount": "1000000000000000000",
+      "metaData": { "reason": ["YD-WETH-21 Liquidity Mining Week 27"] }
+    },
+    "0x00e4846e2971bb2b29cec7c9efc8fa686ae21342": {
+      "amount": "2000000000000000000",
+      "metaData": { "reason": ["YD-WETH-21 Liquidity Mining Week 27", "Governance Votes KPI Options Round 1"] }
```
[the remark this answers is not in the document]
chrismaree: it is not. the proof just consists of the hash of amount + wallet. it will change to include index in @nicholaspai latest PR.

It is simply included to enrich the front end.

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+}
+
+async function fetchClaimsFromKV(chainId: number, windowIndex: number, account: string) {
+  try {
+    return await _fetchDataFromKV(`${chainId}:${windowIndex}:${account}`);
+  } catch (error) {
+    return error;
+  }
+}
+async function fetchChainWindowIndicesFromKV(chainId: number) {
```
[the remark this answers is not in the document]
chrismaree: sure. I did this on this file and on the IPFS helper.

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
+// This file contains some helper functions to wrap cloudflare Key Value store.
+
+const nodeFetch = require("node-fetch");
+
+if (!process.env.CLOUDFLARE_ACCOUNT_ID || !process.env.CLOUDFLARE_NAMESPACE_ID || !process.env.CLOUDFLARE_TOKEN) {
```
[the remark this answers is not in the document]
chrismaree: Interesting. So I originally actually did something akin to this. My original implementation actually just had them as param. It just seemed much cleaner like this as you don't need all this extra code and to pass all this stuff around. The way it is now worked out to be about a third as much code. I optimized for this over the other pattern just because it was cleaner.

I'm happy to change it though! can you explain what your main reasoning for the change is? do you just not like this pattern or feel that it is hard to test? For me, it just seemed a bit messier so I avoided it.

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+const BASE_URL = `https://api.cloudflare.com/client/v4/accounts/${process.env.CLOUDFLARE_ACCOUNT_ID}/storage/kv/namespaces/${process.env.CLOUDFLARE_NAMESPACE_ID}`;
+const AUTHENTICATED_HEADER = {
+  Authorization: `Bearer ${process.env.CLOUDFLARE_TOKEN}`,
+  "Content-Type": "application/json"
+};
+
+// Bulk add to Cloudflare KV. Data should be a stringified array of [{key:<your-key>,value:<your-value>}]
+async function _addDataToKV(data: string) {
+  const response = await nodeFetch(`${BASE_URL}/bulk`, { method: "put", body: data, headers: AUTHENTICATED_HEADER });
+  if (response.status != 200) throw { response, error: new Error("Something went wrong adding data to KV") };
```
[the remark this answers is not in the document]
chrismaree: ye, I ended up changing this a whole bunch as well. The main reason I wanted to do it this was is that how it now preserves all error logs within the `response, including the `.text`. it _also_ include a bunch of other useful info. My idea with adding the error string is that we can _now_ see where exactly the error was thrown. But 100% you are right that you could just pull this out of the stack trace as well.

I'll try experimenting with the `response.ok` syntax and refactoring accordingly.

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+    return {
+      key: `${chainId}:${windowIndex}:${account}`,
+      value: JSON.stringify(claim)
+    };
+  });
+
+  const BATCH_SIZE = 10_000; // limit how many wee add in each bulk upload. Cloud flare limits 10k per bulk put.
+  let i = 0;
+  while (i < KV.length) {
+    await _addDataToKV(JSON.stringify(KV.slice(i, (i += BATCH_SIZE))));
```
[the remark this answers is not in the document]
chrismaree: It's just max number of key-value pairs in one `put`. I've updated the comment accordingly.

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,44 @@
+// This script adds some useful functions for dealing with IPFS.
+
+const ipfsApi = require("ipfs-http-client");
+const nodeFetch = require("node-fetch");
+
+const ipfs = ipfsApi({ host: "ipfs.infura.io", port: "5001", protocol: "https" });
```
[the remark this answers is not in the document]
chrismaree: Similar reasoning as before tbh. cleaner and shorter code the way it is now. Again though I am happy to change it.

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+async function viewFile(fileHash: string) {
+  try {
+    const response = await nodeFetch(`"https://ipfs.infura.io:5001/api/v0/get?arg=${fileHash}`);
+    const json = await response.json();
+    return json;
+  } catch (error) {
+    throw {
+      message: "Something went wrong fetching your file! It might not be json formatted on IPFS",
```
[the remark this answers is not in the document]
chrismaree: it was the same reason as before with error handling to make it easier to see where things are failing but I do agree this is a bit verbose. I've refactored this to be a bit cleaner.

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+const ipfs = ipfsApi({ host: "ipfs.infura.io", port: "5001", protocol: "https" });
+
+async function uploadFile(file: any) {
+  const { path } = await ipfs.add({ content: Buffer.from(JSON.stringify(file)) });
+  return path;
+}
+
+async function viewFile(fileHash: string) {
+  try {
+    const response = await nodeFetch(`"https://ipfs.infura.io:5001/api/v0/get?arg=${fileHash}`);
```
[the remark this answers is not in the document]
chrismaree: I'm really just using the ipfs library for upload. for viewing, the infura portal is by far easier. I spent a few hours trying to get the `ipfs` object to return json objects using the `view` & `cat` methods but they were inconstant. Going directly to the infra portal worked the most easily.

I properly just need to spend more time trying to get this to work correctly. The other option would be to refractor away the IPFS library complete and change the `uploadFile` to _also_ use the infura ipfs portal. However, when I tried this it was more limited in _how_ the uploads could be done. In particular, it expected a file upload (from local directory) rather than a string/buffer, as the current implementation support. this somewhat janky hybrid was the result of pain in working with these libraries between uploading and viewing files.

packages/merkle-distributor/src/MerkleDistributorHelper.ts
chrismaree: Yup! you _could_. The reason I did it this way is that I wanted the _caller_ of this function to not have to have a provider as it will most likely be called by a serverless function, that the front end would call. It also decouples the method from the provider and makes it provider agnostic.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
daywiss: i think after i heard your explanation, and how its pretty standard across other projects it makes me feel a bit better about it. i did like your idea though of changing this to like metaHash or something like that, as its a bit more generalized, but in general i approve of this.

packages/merkle-distributor/src/CloudflareKVHelper.ts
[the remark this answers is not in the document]
daywiss: i agree in general this is cleaner and less code than passing the vars into every function. 

i think my main problem with this pattern is that I look at these files as low level libraries, and this is not a standard way to configure a library.

its also not clear how you would ever instantiate multiple instances of a library configured this way. I dont think this is a problem for our use case, but its a weird side effect of configuring with envs this way. 

testability is another benefit, but these would be hard to test regardless and not my primary concern, though it would be nice if we had a standard pattern we could use to make it more testable. 

If someone was inspecting your higher level scripts to see how to use the libraries, it would not be obvious that there are mandatory env vars which need to be set in order to use it, unless going into the lib code or actually running it. Compare this to a configuration object that is passed in on instantiation, 1. its much more obvious and 2. this is pretty much how 99% of js libraries are done. 

that said, i dont know if it really matters that much, as all the issues ive outlined are mainly around making it more consumable for broader usage and dont affect functionality in your use case. this are things we can always kick the can on and address if it every becomes a problem.

packages/merkle-distributor/src/IpfsHelper.ts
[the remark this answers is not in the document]
daywiss: ok got it! yea that sucks, but probably not worth spending a lot more time on

packages/merkle-distributor/src/MerkleDistributorHelper.ts
daywiss: i see what you are saying. yea it does make it convenient for a serverless call, but the logic just feels out of place among the rest of the functions in this file.  my last suggestion would be to create serverless folder which has thin js wrapper for configuring the dependencies and making a call to the low level function in this file.  this folder could also store any other future serverless calls. Again this is just more about design, separating config from business logic, but functionally it will work the same.

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,44 @@
+// This script adds some useful functions for dealing with IPFS.
+
+const ipfsApi = require("ipfs-http-client");
+const nodeFetch = require("node-fetch");
```
daywiss, line 4: im not sure if this will solve all our fe problems, but there is an [isomorphic-fetch](https://www.npmjs.com/package/isomorphic-fetch) which basically switches between node fetch or fetch based on environment detected

packages/merkle-distributor/hardhat.config.ts
```diff
@@ -0,0 +1,20 @@
...
+const coreWkdir = path.dirname(require.resolve("@uma/core/package.json"));
+const packageWkdir = path.dirname(require.resolve("@uma/merkle-distributor/package.json"));
+
+const configOverride = {
+  paths: {
+    root: coreWkdir,
+    sources: `${coreWkdir}/contracts`,
+    artifacts: `${coreWkdir}/artifacts`,
+    cache: `${coreWkdir}/cache`,
+    tests: `${packageWkdir}/test`
```
mrice32, line 16: nit: do you want to run the validation-test directory as part of the hardhat tests?

packages/merkle-distributor/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "url": "git+https://github.com/UMAprotocol/protocol.git"
+  },
+  "bin": "dist/index.js",
+  "types": "dist/index.d.ts",
+  "main": "dist/index.js",
+  "files": [
+    "/dist/**/*"
+  ],
+  "scripts": {
+    "build": "tsc -b"
```
mrice32, line 36: nit: related to the above, if you have a test, we should probably add a test script here.

mrice32 (commented):
Minor comments around file structure and ts -- I think @nicholaspai and @daywiss have provided a pretty deep logic review.

packages/merkle-distributor/src/CloudflareKVHelper.ts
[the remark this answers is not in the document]
chrismaree: Thanks again for this feedback David. I've gone and followed the pattern you laid out here on both helpers in this PR. Please take a look and let me know what you think!

packages/merkle-distributor/src/IpfsHelper.ts
chrismaree: sure and done!

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,84 @@
...
+
+  // Save the file to disk.
+  const savePath = `${path.resolve(__dirname)}/../proof-files/chain-id-${recipientsObject.chainId}-reward-window-${
+    recipientsObject.windowIndex
+  }-claims-file.json`;
+  fs.writeFileSync(savePath, JSON.stringify(outputData));
+  console.log("🗄  File successfully written to", savePath);
+}
+
+main();
```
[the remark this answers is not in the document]
chrismaree: my testing it plays just fine. the only _small_ issue is that it does not exit if it crashes and you need to `ctrl+c` to end the process. I played around with a `.then` & `.catch` pattern but it did not always handle error cases correctly always :/ it could be a type script artefact, I'm not sure.

I left it like this because it is min code to achieve the goal while still being 100% compatible with the console.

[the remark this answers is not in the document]
chrismaree: I've refactored this to use:
```
main().catch(e => {
  console.log(e);
  process.exit(1);
});

```
and it works great like this.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,84 @@
...
+//   "0x00b591bc2b682a0b30dd72bac9406bfa13e5d3cd": {
+//     "amount": "1000000000000000000",
+//     "metaData": { "reason": ["YD-WETH-21 Liquidity Mining Week 27"] }
+//   }... for all recipients
+// }
+
+// example execution: ts-node ./scripts/1_CreateClaimsForWindow.ts -i ./scripts/example.json
+
+const assert = require("assert");
+const path = require("path");
```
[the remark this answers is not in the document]
chrismaree: sure. have changed.

packages/merkle-distributor/hardhat.config.ts
chrismaree: we _could_. issue is these tests require Cloudflare keys + an Infura API Key as they ensure that the end to end data retrieval works as expected. would you like this kind of test to be in CI?

packages/merkle-distributor/package.json
chrismaree: I added a script for the `integration-test`. I dont think we want to run this in CI imho.

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
+// This file contains some helper functions to wrap cloudflare Key Value store.
+
+const nodeFetch = require("node-fetch");
+const assert = require("assert");
+
+export default (accountId: string | undefined, namespaceId: string | undefined, token: string | undefined) => {
+  assert(accountId, "Cloudflare helper requires an Account ID");
+  assert(namespaceId, "Cloudflare helper requires a data store namespace");
+  assert(token, "Cloudflare helper requires an Account token");
```
daywiss, line 9: nice, i feel like this is much more portable now

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,92 @@
...
+
+  const baseUrl = `https://api.cloudflare.com/client/v4/accounts/${accountId}/storage/kv/namespaces/${namespaceId}`;
+  const authenticatedHeader = { Authorization: `Bearer ${token}`, "Content-Type": "application/json" };
+
+  // Bulk add to Cloudflare KV. Data should be a stringified array of [{key:<your-key>,value:<your-value>}]
+  async function _addDataToKV(data: string) {
+    const response = await nodeFetch(`${baseUrl}/bulk`, { method: "put", body: data, headers: authenticatedHeader });
+    if (response.status != 200) throw { response, error: new Error("Something went wrong adding data to KV") };
```
daywiss, line 17: would still like to see normal errors thrown, but use your judgement

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -23,17 +21,25 @@ let leaf;
...
   assert.isTrue(
     Object.keys(recipient).every(val => ["account", "amount", "accountIndex"].includes(val)),
     "recipient does not contain required keys"
   );
-  return web3.utils.soliditySha3(
-    { t: "address", v: recipient.account },
-    { t: "uint256", v: recipient.amount },
-    { t: "uint256", v: recipient.accountIndex }
+
+  return Buffer.from(
```
nicholaspai, line 34: Why'd you add Buffer.from wrapper?

packages/merkle-distributor/hardhat.config.ts
nicholaspai: I wonder if you can easily separate out the unit tests that don't require secrets. Would be good to have as much test coverage in CI as possible

nicholaspai (approved):
Second round of comments, lots of good work.

packages/core/test/merkle-distributor/MerkleDistributor.js
chrismaree: the typescript library returns a buffer. the previous one returned a string.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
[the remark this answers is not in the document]
mrice32: Nope, if string is the canonical representation that others use and altogether easier for the end-user, I'd say keep as a string! Just was curious what the canonical representation was.

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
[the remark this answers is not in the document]
mrice32: Sounds good. All I wanted to make sure is that:
- The process terminates itself in all cases.
- The exit code is correct (i.e. errors are reported).

packages/merkle-distributor/hardhat.config.ts
mrice32: @nicholaspai agreed. It would be nice to factor our ci that way.

@chrismaree didn't realize these required API keys. Not running it in ci makes sense, at least given our current setup!

mrice32 (approved):
LGTM! All of my comments are resolved.

packages/merkle-distributor/integration-test/MerkleDistributorHelper.ts
```diff
@@ -0,0 +1,114 @@
...
+// They are simply meant to showcase the correct behavior of the helpers under a number of different situations and inputs.
+
+const { ethers } = require("ethers");
+const { assert } = require("chai");
+
+const {
+  createLeaf,
+  createMerkleDistributionProofs,
+  getClaimsForAddress
+} = require("../src/MerkleDistributorHelper.ts");
```
mrice32, line 12: nit: import?

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,92 @@
...
+//   }... for all recipients
+// }
+
+// example execution: ts-node ./scripts/1_CreateClaimsForWindow.ts -i ./scripts/example.json
+
+import assert from "assert";
+import path from "path";
+import { program } from "commander";
+import fs from "fs";
+const { getWeb3 } = require("@uma/common");
```
mrice32, line 24: nit: import?

packages/merkle-distributor/scripts/1_CreateClaimsForWindow.ts
```diff
@@ -0,0 +1,92 @@
...
+const web3 = getWeb3();
+const { toBN } = web3.utils;
+
+import MerkleDistributorHelper = require("../src/MerkleDistributorHelper");
```
mrice32, line 28: nit: why can't this be
```ts
import MerkleDistributor from "../src/MerkleDistributorHelper";
```

or 

```ts
import * as MerkleDistributorHelper from "../src/MerkleDistributorHelper";
```

if both of these don't work for some reason, totally fine to leave as is.

packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts
```diff
@@ -0,0 +1,129 @@
...
+// }
+
+// example execution: ts-node ./scripts/2_PublishClaimsForWindow.ts -i ./proof-files/chain-id-42-reward-window-0-claims-file.json -m 0xAfCd2405298C2FABB2F7fCcEB919B4505A6bdDFC --network kovan_mnemonic
+
+import assert from "assert";
+import { program } from "commander";
+import fs from "fs";
+
+const { getAbi } = require("@uma/core");
+const { getWeb3, MAX_UINT_VAL } = require("@uma/common");
```
mrice32, line 37: nit: import?

packages/merkle-distributor/src/CloudflareKVHelper.ts
```diff
@@ -0,0 +1,82 @@
+// This file contains some helper functions to wrap cloudflare Key Value store.
+
+const nodeFetch = require("node-fetch");
+const assert = require("assert");
```
mrice32, line 4: nit: import?

packages/merkle-distributor/src/IpfsHelper.ts
```diff
@@ -0,0 +1,37 @@
...
```
mrice32, line 4: nit: import?

packages/merkle-distributor/src/MerkleDistributorHelper.ts
```diff
@@ -0,0 +1,92 @@
+// This helper includes a set of useful functions when dealing with merkle trees and payout claims.
+const { ethers } = require("ethers");
+const { getAbi } = require("@uma/core");
```
mrice32, line 3: nit: import?

mrice32 (commented):
Bunch of tiny nits around using import over require in typescript.

Please don't block the merge on this, just wanted to point them out in case we could move some/all to import.

daywiss (approved):
@chrismaree im really looking foward to getting this integrated into the kpi portal. because this is such a big PR, and it works, i'm ok with shaking out additional design changes as needed when we start integrating. great job on this and thanks for your patience with review!

packages/merkle-distributor/integration-test/MerkleDistributorHelper.ts
chrismaree: fixed.

packages/merkle-distributor/src/CloudflareKVHelper.ts
chrismaree: done! I had to install `@types/node-fetch` but it worked.

chrismaree:
> Bunch of tiny nits around using import over require in typescript.
> 
> Please don't block the merge on this, just wanted to point them out in case we could move some/all to import.

I fixed all the type imports you identified! thanks for pointing them out :)

PR #2647, opened 2021-03-05
feat(core): general improvements to claim rewards script

mrice32:
**Motivation**

Over the past few weeks, I've had to modify the claim rewards script a few times as a part of our upgrade process. To make this process easier in the future, I thought it might be nice to publish these updates.

**Summary**

A few notable updates:
- The script can now exclude addresses.
- The script can now claim for all rounds if unspecified or user-specified round(s).
- The script can now be used to only claim for one or multiple addresses instead of all addresses.
- The script can talk to multiple Voting versions at once.
- The script now takes an option just to print a summary of unclaimed rewards rather than claiming.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/local/ClaimAllRewards.js
+++ b/packages/core/scripts/local/ClaimAllRewards.js
@@ -27,8 +27,8 @@ const { toBN, toWei, toChecksumAddress } = web3.utils;
 // This script claims all voter's rewards for the round provided.
 async function claimRewards() {
   const rounds = argv.round ? lodash.castArray(argv.round) : [];
-  const excludeAddresses = argv.excludeAddress ? lodash.castArray(argv.excludeAddress) : [];
-  const claimAddresses = argv.claimAddress ? lodash.castArray(argv.claimAddress) : [];
+  const excludeAddresses = argv.excludeAddress ? lodash.castArray(argv.excludeAddress).map(toChecksumAddress) : [];
+  const claimAddresses = argv.claimAddress ? lodash.castArray(argv.claimAddress).map(toChecksumAddress) : [];
   const versions = argv.version ? lodash.castArray(argv.version) : ["latest", "1.2.2"];
 
   const account = (await web3.eth.getAccounts())[0];
@@ -38,6 +38,9 @@ async function claimRewards() {
     if (version.startsWith("1")) {
       return new web3.eth.Contract(getAbi("Voting", version), getAddress("Voting", networkId, version));
     } else {
+      // This uses the VotingAncillaryInterfaceTesting to create a voting contract that has _only_ the ancillary
+      // functions and events. This interface also includes almost all the regular voting methods that are unrelated
+      // to the ancillary data change. Using this interface effectively avoids the overload problem.
       return new web3.eth.Contract(
         getAbi("VotingAncillaryInterfaceTesting", "latest"),
         getAddress("Voting", networkId, version)
@@ -189,7 +192,7 @@ async function claimRewards() {
 
 async function wrapper(callback) {
   try {
-    await claimRewards(argv.multisig);
+    await claimRewards();
   } catch (e) {
     // Forces the script to return a nonzero error code so failure can be detected in bash.
     callback(e);
--- a/packages/core/scripts/local/ClaimAllRewards.js
+++ b/packages/core/scripts/local/ClaimAllRewards.js
@@ -1,39 +1,98 @@
-// Usage: $(npm bin)/truffle exec ./scripts/ClaimAllRewards.js --round <round_id> --batcherAddress 0x82458d1C812D7c930Bb3229c9e159cbabD9AA8Cb --network mainnet_mnemonic
+// Usage: $(npm bin)/truffle exec ./scripts/ClaimAllRewards.js --batcherAddress 0x82458d1C812D7c930Bb3229c9e159cbabD9AA8Cb --network mainnet_mnemonic
+// Other optional args:
+// --round <round id> Example: `--round 9344 --round 9342`. This only considers rewards for rounds 9344 and 9342.
+// --claimAddress <address> Example: `--claimAddress 0x1234 --claimAddress 0x5678`. This only considers rewards for
+//   0x1234 and 0x5678. Defaults to claiming for _all_ addresses.
+// --excludeAddress <address> Example: `--excludeAddress 0x1234 --excludeAddress 0x5678`. This excludes addresses
+//   0x1234 and 0x5678. Defaults to excluding no addresses.
+// --onlyPrint. This flag takes no arguments. It tells the script to just print out the available rewards rather than
+//   trying to claim them. Default off.
+// --version <version string>. Example: `--version 1.2.2 --version latest` DVM versions to consider when claiming.
+//   Since a new Voting contract was deployed in version 2.0.0, that version will only claim for that contract, not
+//   older ones. Defaults to 1.2.2 and latest.
 
-const Voting = artifacts.require("Voting");
-const VotingInterfaceTesting = artifacts.require("VotingInterfaceTesting");
+const { getAbi, getAddress } = require("../../index");
 const TransactionBatcher = artifacts.require("TransactionBatcher");
+const lodash = require("lodash");
+const winston = require("winston");
+const { GasEstimator } = require("@uma/financial-templates-lib");
 
-const argv = require("minimist")(process.argv.slice(), { string: ["round", "batcherAddress"] });
+const argv = require("minimist")(process.argv.slice(), {
+  string: ["round", "batcherAddress", "claimAddress", "excludeAddress", "version"],
+  boolean: "onlyPrint"
+});
 
-const { toBN, toWei } = web3.utils;
+const { toBN, toWei, toChecksumAddress } = web3.utils;
 
 // This script claims all voter's rewards for the round provided.
 async function claimRewards() {
-  const voting = await VotingInterfaceTesting.at((await Voting.deployed()).address);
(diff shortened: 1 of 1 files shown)


packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
         }
       })
     )
   ).filter(element => element !== null);
 
-  const dataArray = retrievableRewards.map(([voter, priceRequests]) => {
-    return voting.contract.methods.retrieveRewards(voter, argv.round, priceRequests).encodeABI();
+  if (argv.onlyCompute) {
+    const groupedByVoter = lodash.groupBy(retrievableRewards, el => el.voter);
+    Object.entries(groupedByVoter).forEach(([voter, rewardArray]) => {
```
chrismaree, line 115: o boy. what a console log.

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
-// Usage: $(npm bin)/truffle exec ./scripts/ClaimAllRewards.js --round <round_id> --batcherAddress 0x82458d1C812D7c930Bb3229c9e159cbabD9AA8Cb --network mainnet_mnemonic
+// Usage: $(npm bin)/truffle exec ./scripts/ClaimAllRewards.js --batcherAddress 0x82458d1C812D7c930Bb3229c9e159cbabD9AA8Cb --network mainnet_mnemonic
+// Other optional args:
+// --round <round id> Example: `--round 9344 --round 9342`. This only considers rewards for rounds 9344 and 9342.
+// --claimAddress <address> Example: `--claimAddress 0x1234 --claimAddress 0x5678`. This only considers rewards for
+//   0x1234 and 0x5678. Defaults to claiming for _all_ addresses.
+// --excludeAddress <address> Example: `--excludeAddress 0x1234 --excludeAddress 0x5678`. This excludes addresses
```
chrismaree, line 6: this is so cool.

chrismaree (approved):

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
 
-  await transactionBatcher.batchSend(targetArray, valuesArray, dataArray);
+        if (gasEstimate > 6000000) {
+          throw "The transaction requires too much gas. Will need to be split up.";
+        }
+        await gasEstimator.update();
+
+        const receipt = await txn.send({
+          gasPrice: gasEstimator.getCurrentFastPrice(),
+          gas: gasEstimate * 2,
```
nicholaspai, line 181: *2 seems potentially dangerous if this script is run during high gas times, but at least it will execute! Would be cool to add a quick metamask style prompt here for user to select between standard, fast, and instant gas: https://www.digitalocean.com/community/tutorials/nodejs-interactive-command-line-prompts

Perhaps something we should add as a common lib and import into all mainnet scripts

nicholaspai (approved):
This is a huge improvement, LGTM

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
-const { toBN, toWei } = web3.utils;
+const { toBN, toWei, toChecksumAddress } = web3.utils;
 
 // This script claims all voter's rewards for the round provided.
 async function claimRewards() {
-  const voting = await VotingInterfaceTesting.at((await Voting.deployed()).address);
+  const rounds = argv.round ? lodash.castArray(argv.round) : [];
+  const excludeAddresses = argv.excludeAddress ? lodash.castArray(argv.excludeAddress) : [];
+  const claimAddresses = argv.claimAddress ? lodash.castArray(argv.claimAddress) : [];
+  const versions = argv.version ? lodash.castArray(argv.version) : ["latest", "1.2.2"];
```
pemulis, line 32: Aren't all of the rewards for `1.2.2` claimed at this point?

pemulis (approved):

packages/core/scripts/local/ClaimAllRewards.js
daywiss: yea i agree this could get really crazy. would be nice maybe specify the gas as instant, fast, normal as a cmd line arg when running script.

packages/core/scripts/local/ClaimAllRewards.js
daywiss: this is interesting, never heard of console.group!

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
-    toBlock: "latest"
+  const account = (await web3.eth.getAccounts())[0];
+  const networkId = await web3.eth.net.getId();
+
+  const votingContracts = versions.map(version => {
+    if (version.startsWith("1")) {
+      return new web3.eth.Contract(getAbi("Voting", version), getAddress("Voting", networkId, version));
+    } else {
+      return new web3.eth.Contract(
+        getAbi("VotingAncillaryInterfaceTesting", "latest"),
```
daywiss, line 42: why do we use testing interface on v2? might want to leave a comment here. i cant even remember why and i remember doing something like this in voter dapp

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
-    if (votersToPriceRequests[voter]) {
-      votersToPriceRequests[voter].push(newPriceRequest);
+    const round = event.returnValues.roundId;
+    const contractAddress = event.address;
+    const key = [voter, round, contractAddress].join("|");
+    if (excludeAddresses.includes(toChecksumAddress(voter))) {
+      console.log("Found noclaim voter", voter);
+      continue;
+    }
+    const newPriceRequest = {
```
daywiss, line 70: was trying to understand why price requests have anything to do with this, but i think i realized. is it the case that even governance votes are also considered price requests? so this basically covers all votes

daywiss (approved):
this looks good to me and really cool usage of lodash functions in several places.  i just want to mention this as an alternate design to this, but just for thought:

I think when you try to handle the aggregation of "things" as a plural, rather than a thing as a singular, it creates more complicated code and extra structure that makes it harder to reason about. For instance, this script handles claiming across several voting contracts. I can imagine a function which takes in a single voting contract, fetches its events and basically does the same things you are doing with all the contracts at once. Then at the top level you fetch all the voting contracts and promise.map against this function. 

I think theres some optimizations you gain by processing everything together  (like with the batching), and im not 100% convinced what i am suggesting would be simpler for this use case, but i find myself making this decision a lot and tend to go towards the singular implementation.

packages/core/scripts/local/ClaimAllRewards.js
mrice32: Note: this is the _gas_ to send, not the price. Not sure if we should multiply it by this much. I mainly did this because I'm not sure about how refunds are handled and these rewards transactions get a lot of refunds. Probably deserves a little more research.

packages/core/scripts/local/ClaimAllRewards.js
mrice32: No. Some whales requested to not have their rewards claimed for tax reasons.

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -1,72 +1,190 @@
...
+    }),
+    60, // Time between updates.
+    100 // Default gas price.
+  );
+
+  // These nested calls just chunk up the above arrays into 30 transaction chunks, then organize them in groups for
+  // sending on-chain.
+  await Promise.all(
+    lodash
+      .zip(...[targetArray, valuesArray, dataArray].map(arr => lodash.chunk(arr, 30)))
```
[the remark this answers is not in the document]
mrice32: Took me a few tries to get right :)

packages/core/scripts/local/ClaimAllRewards.js
mrice32: Will add a comment. TLDR: the newer voting contract has overloads (web3 picks blindly between the overloads) and we want to make sure to _only_ use the ancillary data methods throughout.

packages/core/scripts/local/ClaimAllRewards.js
mrice32: Yep, exactly!

mrice32:
> this looks good to me and really cool usage of lodash functions in several places. i just want to mention this as an alternate design to this, but just for thought:
> 
> I think when you try to handle the aggregation of "things" as a plural, rather than a thing as a singular, it creates more complicated code and extra structure that makes it harder to reason about. For instance, this script handles claiming across several voting contracts. I can imagine a function which takes in a single voting contract, fetches its events and basically does the same things you are doing with all the contracts at once. Then at the top level you fetch all the voting contracts and promise.map against this function.
> 
> I think theres some optimizations you gain by processing everything together (like with the batching), and im not 100% convinced what i am suggesting would be simpler for this use case, but i find myself making this decision a lot and tend to go towards the singular implementation.

@daywiss I tried something almost exactly like this on my first try and I felt like the nested functions actually made it more difficult to reason about 🤷 . At least my version of what you're talking about wasn't much simpler, but I could've done it in a suboptimal way!

The other tricky thing is that in some places we want to map over every single vote or round and other places we just want to map over each voting contract and other places we want to map over every transaction. So it was a bit difficult (at least for me!) to build this into a lowest-common-denominator pipeline, but I'm sure there are better ways to do it than I could think of :)

PR #2648, opened 2021-03-05
feat(affiliates): add minting lookback technique as dappmining v2

### LINKED ISSUE #2644: Change dApp mining algorithm to minting lookback

daywiss:
**Motivation**

#2644

**Summary**

This adds a new dapp mining attribution technique and adds a way to call "versions" of dapp mining so the previous technique can still be run for comparison. No new configs are needed, and by default this version will be used when running dapp mining unless specified otherwise in the config. 

**Details**
When comparing to the first dapp mining algorithm, not much has changed other than the attribution model and some small summations. The attribution model now maintains a history of all tagged mints per sponser, and will search through them whenever there needs to be an attribution calculation from a users balance. 

Running based on last weeks YD-ETH dapp mining rewards shows a subtle difference in the calculation:

| Developer | V1 Payout | V2 Payout |
| -- | -- | -- |
| UMA | 4922.852075837127248111  | 4567.838090481865205757 |
|opendao|204.951478474104721601 | 204.951478474104721601 |
|chickflock|204.951478474104721601 | 586.210431044030061924 |

Interestingly this did not affect the opendao team at all, but increases chickflock rewards over 100% since their mints were very recent.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2644


packages/affiliates/apps/DappMiningRewards.js
```diff
@@ -9,12 +10,17 @@ const Queries = require("../libs/bigquery");
 // This is the main function which configures all data sources for the calculation.
 async function App(config) {
   const web3 = getWeb3();
+  const { version = "v2" } = config;
+  assert(
+    DappMining[version],
+    "Invalid version in dappmining config, must be one of: " + Object.keys(DappMining).join(", ")
+  );
```
daywiss, line 17: this allows us to specify a version of dapp mining to run. This is a backward compatible change, so no configs will need to change, unless specifically running v1.

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
+const assert = require("assert");
```
daywiss, line 1: this file is almost identical to dappmining_v1. Will highlight major changes

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+const highland = require("highland");
+const moment = require("moment");
+
+const { DecodeTransaction, DecodeLog } = require("../contracts");
+const { EmpAttributions, EmpBalances } = require("../processors");
+const { AttributionLookback, Balances } = require("../models");
```
daywiss, line 7: We use the AttributionLookback model, rather than SharedAttribution model

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+  function calculateBlockReward(
+    { attributions, balances, whitelist },
+    // return all tags with a percent value for rewards
+    rewards = {}
+  ) {
+    // for every user with a token balance
+    balances.keys().forEach(userid => {
+      // get users balance
+      const balance = balances.get(userid);
+      const attributionPercents = attributions.getAttributionPercents(userid, balance);
```
daywiss, line 63: this function now returns an object keyed by developer addresses that show attribution % based on the users balance.

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+  // integer percent calculation where percent is passed in as wei
+  function calculatePercent(amount, percent) {
+    return fromWei(toBN(amount.toString()).mul(toBN(percent.toString())));
+  }
+
+  // calculates attribution weights for multiple blocks by simply taking last snapshot state and multiplying by
+  // the number of blocks elapsed.
+  function sumAttributions(attributions, blocksElapsed = 1, sum = AttributionLookback()) {
+    blocksElapsed = toBN(blocksElapsed);
+    attributions.forEach((userid, affiliate, amount, index) => {
```
daywiss, line 92: this calculation changed pretty significantly, but the structure is similar to before. This updates multiplies all attributions based on number of blocks elapsed and saves them to a new lookback table.

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+    const startBlock = await getStartBlock(startTime);
+    const endBlock = await getEndBlock(endTime);
+
+    // Wire up streaming logic
+    const attributionPromise = attributionStream
+      .through(
+        ProcessAttributionStream({
+          startBlock: startBlock.number,
+          endBlock: endBlock.number,
+          attributions: EmpAttributions(empAbi, defaultAddress, AttributionLookback())
```
daywiss, line 237: EmpAttribution process now optionally takes the attribution model, so we pass it the new lookback model to be populated by tagged mint events.

packages/affiliates/test/mocha/dappmining.js
```diff
@@ -262,3 +262,260 @@ describe("DappMining", function() {
     });
   });
 });
+describe("DappMining V2", function() {
```
daywiss, line 265: These new tests copy most from v1, but subtly changes some of the scenarios to test that the lookback is working correctly

daywiss:
btw i know this is hard to review. i had a hard time going back and understanding what this code did in the previous implementation (that i wrote!), it probably took at least an hour to figure that out. i dont really know how to make this easier to understand. more comments might help but even that i dont think is enough.

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+  function sumAttributions(attributions, blocksElapsed = 1, sum = AttributionLookback()) {
+    blocksElapsed = toBN(blocksElapsed);
+    attributions.forEach((userid, affiliate, amount, index) => {
+      let currentSum;
+      try {
+        currentSum = sum.getByIndex(userid, index).amount;
+      } catch (err) {
+        currentSum = "0";
+      }
+      const weighted = toBN(amount)
```
mrice32, line 99: I feel like I'm missing something in this calculation. Why do you weight by number of blocks elapsed?

packages/affiliates/libs/affiliates/dappmining_v2.js
nicholaspai: +1

packages/affiliates/libs/affiliates/dappmining_v2.js
daywiss: i made an observation about the math for dapp mining ( and could apply to dev mining), that allowed the calculation to be optimized. 

the first way we can do this calculation is that we must know the state of the system at every block, this includes every user balance, all attributions and the per block reward. We then accumulate the rewards at every block until we get to the last block. This is hard to do without iterating and snapshotting the state at every block once, then iterating again to attribute rewards. This is exactly what dev mining does. 

In dapp mining, the math works out the same if we basically sum up the state of balances and attributions independently and apply the rewards calculation at the very end. You can imagine if someone's balance was 100 at block 1 though 10, this basically says their balance weight for 10 blocks is 100 * 10. Its the same thing for attributions, these will also be accumulated every block any time there is a state change.  

hopefully this makes sense!

mrice32: Yep, it does!

mrice32 (approved):
LGTM!

packages/affiliates/libs/affiliates/dappmining_v2.js
chrismaree: this is pretty awesome.

packages/affiliates/libs/affiliates/dappmining_v2.js
```diff
@@ -0,0 +1,278 @@
...
+    const [result] = blocks.map(block => {
+      return {
+        ...block,
+        timestamp: moment(block.timestamp.value).valueOf()
+      };
+    });
+    return result;
+  }
+
+  // Broken out logic for consuming and processing attribution stream. This can now be tested
```
chrismaree, line 140: thank you for these comments. helps with understanding a lot.

chrismaree (approved):
Great work David!

Some super sophisticated stuff going on here. Really like that we can now accommodate both versions and be backwards compatible.

LGTM!

PR #2649, opened 2021-03-05
refactor(merkle-distributor): Gas optimizations

nicholaspai:
**Motivation**

- After offline discussions we find it unlikely that we'd take advantage of the `windowStart` param, which could be used to "schedule" distributions", so this PR removes it to save gas for the `claim` function. This also removes the `Testable` contract.
- Changes `claimed` mapping to bitmap
- Removes `if(amount == 0)` check in `_disburse`

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -103,14 +91,13 @@ contract MerkleDistributor is Ownable, Testable {
     //       their claims to suceed.
```
mrice32, line 91: Unrelated to this line, can we drop passing in timerAddress, inheriting from Testable, and the constructor entirely since we no longer have any time-dependent logic?

mrice32 (approved):
Looks awesome! One minor comment

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: yes! will do

nicholaspai: Weirdly removing this added 8 gas

mrice32: lol

mrice32: Have you tried nicely asking the solidity compiler to optimize harder?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -184,9 +166,7 @@ contract MerkleDistributor is Ownable, Testable {
         // Check the account has not yet claimed for this window.
         require(!claimed[claim.windowIndex][claim.account], "Account has already claimed for this window");
 
-        // Proof is correct and claim has not occurred yet; check that claim window has begun.
-        require(getCurrentTime() >= merkleWindows[claim.windowIndex].start, "Claim window has not begun");
-
+        // Proof is correct and claim has not occurred yet, mark claimed complete.
         claimed[claim.windowIndex][claim.account] = true;
         emit Claimed(
             msg.sender,
```
mrice32, line 172: unrelated to this line, but I think we can drop this if statement below to kill the last TODO in this contract :) :
```
// TODO: Should we revert claims for 0 tokens?
if (amount > 0)
```

We shouldn't be putting 0-token distributions in there anyway, and most tokens don't revert on 0-token transfers, so I would argue just to drop the if statement and blindly transfer. The only way I could imagine the user getting there is passing 0 claims into the batcher, which IMO isn't a big concern and easy to diagnose.

mrice32 (approved):
This LGTM! Curious to hear how much gas savings (if any) the index 0 claim + index 1 claim gets.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: you think thats worth it? Its at 199 runs

nicholaspai:
Before merging I want to add a test similar to this one by [Uniswap](https://github.com/Uniswap/merkle-distributor/blob/master/test/MerkleDistributor.spec.ts#L287) which tests for average gas on a  large tree (~100,000 leaves)

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -19,17 +18,13 @@ import "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 import "@openzeppelin/contracts/cryptography/MerkleProof.sol";
 import "@openzeppelin/contracts/access/Ownable.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
-import "../../common/implementation/Testable.sol";
 
-contract MerkleDistributor is Ownable, Testable {
```
chrismaree, line 24: +1

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -155,43 +142,56 @@ contract MerkleDistributor is Ownable, Testable {
         _disburse(merkleWindows[claim.windowIndex].rewardToken, claim.account, claim.amount);
     }
 
+    function isClaimed(uint256 windowIndex, uint256 accountIndex) public view returns (bool) {
```
chrismaree, line 145: can you add some comments above this?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -155,43 +142,56 @@ contract MerkleDistributor is Ownable, Testable {
...
-        emit CreatedWindow(windowIndex, totalRewardsDistributed, windowStart, rewardToken, msg.sender);
+        emit CreatedWindow(windowIndex, totalRewardsDistributed, rewardToken, msg.sender);
     }
 
     function _markClaimed(Claim memory claim) private {
         // Check claimed proof against merkle window at given index.
         require(_verifyClaim(claim), "Incorrect merkle proof");
         // Check the account has not yet claimed for this window.
-        require(!claimed[claim.windowIndex][claim.account], "Account has already claimed for this window");
+        require(!isClaimed(claim.windowIndex, claim.accountIndex), "Account has already claimed for this window");
```
chrismaree, line 186: nice

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -24,22 +22,29 @@ let rewardLeafs;
...
   assert.isTrue(
-    Object.keys(recipient).every(val => ["account", "amount"].includes(val)),
+    Object.keys(recipient).every(val => ["account", "amount", "accountIndex"].includes(val)),
     "recipient does not contain required keys"
   );
-  return web3.utils.soliditySha3({ t: "address", v: recipient.account }, { t: "uint256", v: recipient.amount });
+  return web3.utils.soliditySha3(
+    { t: "address", v: recipient.account },
+    { t: "uint256", v: recipient.amount },
+    { t: "uint256", v: recipient.accountIndex }
```
chrismaree, line 36: I'll update the SDK to accommodate this.

chrismaree (approved):
Great refinement! always good to save gas. 

In PR [2643](https://github.com/UMAprotocol/protocol/pull/2643) I added the IPFS hash of the claims window to the claims object. Given the changes made here, we _could_ argue that we dont actually want this functionality. what do you guys think? My reasoning with this is that I want recipients to be able to claim rewards by simply looking at the smart contracts. By having the IPFS hash, they can always: a) find their proof b) re-construct their proof & Merkle root & c) claim reward tokens without needing anything other than the smart contract and etherscan.

nicholaspai:
> By having the IPFS hash, they can always: a) find their proof b) re-construct their proof & Merkle root & c) claim reward tokens without needing anything other than the smart contract and etherscan

So if we don't store the IPFS hash on-chain, the user can still construct their proof right? But it'd just be more involved?

Why would users want to construct their own proof by the way?

nicholaspai:
> This LGTM! Curious to hear how much gas savings (if any) the index 0 claim + index 1 claim gets.

Added in tests that show case gas savings with bitmap, its as expected around ~15000 gas for consecutive claims.

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -346,135 +313,271 @@ contract("MerkleDistributor.js", function(accounts) {
...
+          account: leaf.account,
+          accountIndex: leaf.accountIndex,
+          amount: leaf.amount,
+          merkleProof: proof
+        });
+        total = total.addn(tx.receipt.gasUsed);
+        count++;
+      }
+      const average = total.divn(count);
+      assert.equal(Math.floor(average.toNumber()), 70463);
```
nicholaspai, line 552: @UMAprotocol/eng here is the large gas savings with the bitmap

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -142,17 +130,33 @@ contract MerkleDistributor is Ownable, Testable {
         for (uint256 i = 0; i < claims.length; i++) {
             Claim memory claim = claims[i];
             require(claim.account == account, "Invalid account in batch claim");
-            _markClaimed(claim);
+            _verifyAndMarkClaimed(claim);
             amountToClaim = amountToClaim.add(claim.amount);
         }
-        _disburse(IERC20(rewardToken), account, amountToClaim);
+        IERC20(rewardToken).safeTransfer(account, amountToClaim);
```
daywiss, line 136: this seems dangerous. dont you also need to verify all claim windows match the rewardToken?

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -315,166 +304,283 @@ contract("MerkleDistributor.js", function(accounts) {
...
-  describe("(claimWindows)", function() {
-    let rewardRecipients1, rewardRecipients2;
-    let rewardLeafs1, rewardLeafs2;
-    let merkleTree1, merkleTree2;
-    beforeEach(async function() {
-      // Assume we start at first windowIndex. Disable vesting.
-      windowIndex = 0;
-      const currentTime = await timer.getCurrentTime();
-      windowStart = currentTime;
+    describe("(claimWindows)", function() {
```
daywiss, line 355: can we also add another test where we have claim windows with different erc20 token payouts, and then a batch claim where a user submits claim across multiple tokens?

daywiss (commented):
like the simplicity and gas savings of these changes so far. just had a comment i wanted to make sure isnt a problem

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: Haha, no -- just joking.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -161,56 +165,48 @@ contract MerkleDistributor is Ownable, Testable {
...
+    function _verifyAndMarkClaimed(Claim memory claim) private {
         // Check claimed proof against merkle window at given index.
-        require(_verifyClaim(claim), "Incorrect merkle proof");
+        require(verifyClaim(claim), "Incorrect merkle proof");
         // Check the account has not yet claimed for this window.
-        require(!claimed[claim.windowIndex][claim.account], "Account has already claimed for this window");
-
-        // Proof is correct and claim has not occurred yet; check that claim window has begun.
-        require(getCurrentTime() >= merkleWindows[claim.windowIndex].start, "Claim window has not begun");
+        require(!isClaimed(claim.windowIndex, claim.accountIndex), "Account has already claimed for this window");
```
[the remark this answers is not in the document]
nicholaspai: I think clients would simulate this call so moving this require wouldn't change anything

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: ohh yes great point, should add an `if(claim.rewardToken == rewardToken)` to the loop above

packages/core/test/merkle-distributor/MerkleDistributor.js
nicholaspai: Yep

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: This is a great point. That is definitely exploitable.

mrice32: We probably need a similar check to `claim.account == account` for `rewardToken`.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -142,17 +130,33 @@ contract MerkleDistributor is Ownable, Testable {
...
-        _disburse(IERC20(rewardToken), account, amountToClaim);
+        IERC20(rewardToken).safeTransfer(account, amountToClaim);
     }
 
     // Claim `amount` of reward tokens for `account`. If `amount` and `account` do not exactly match the values stored
     // in the merkle proof for this `windowIndex` this method will revert.
     function claimWindow(Claim memory claim) public {
-        _markClaimed(claim);
-        _disburse(merkleWindows[claim.windowIndex].rewardToken, claim.account, claim.amount);
+        _verifyAndMarkClaimed(claim);
```
mrice32, line 142: nit: should we name this function just `claim`?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -142,17 +130,33 @@ contract MerkleDistributor is Ownable, Testable {
         for (uint256 i = 0; i < claims.length; i++) {
```
mrice32, line 130: If you name the function below, you could name this one claimMulti.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -142,17 +130,33 @@ contract MerkleDistributor is Ownable, Testable {
...
     function claimWindow(Claim memory claim) public {
-        _markClaimed(claim);
-        _disburse(merkleWindows[claim.windowIndex].rewardToken, claim.account, claim.amount);
+        _verifyAndMarkClaimed(claim);
+        merkleWindows[claim.windowIndex].rewardToken.safeTransfer(claim.account, claim.amount);
+    }
+
+    // Returns True if the claim for `accountIndex` has already been completed for the Merkle
+    // root at `windowIndex`.
+    function isClaimed(uint256 windowIndex, uint256 accountIndex) public view returns (bool) {
```
mrice32, line 148: Really like this new interface -- isClaimed + verifyClaim

mrice32 (approved):
LGTM. Only minor nits

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Great catch, fixed

packages/core/test/merkle-distributor/MerkleDistributor.js
nicholaspai: Done

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Yeah less verbose is better

nicholaspai: Saves 50 gas too 🤨

mrice32: haha

mrice32 (approved):

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -315,166 +322,350 @@ contract("MerkleDistributor.js", function(accounts) {
...
-      // Account 0 should have gained claimed amount from both leaves.
-      const batchedClaimAmount = toBN(leaf1.amount).add(toBN(leaf2.amount));
-      assert.equal(
-        (await rewardToken.balanceOf(leaf1.account)).toString(),
-        accountBalanceBefore.add(batchedClaimAmount).toString()
-      );
+        // But batching them together fails.
+        assert(await didContractThrow(merkleDistributor.claimMulti(invalidClaims, rewardToken.address, leaf1.account)));
+      });
+      it("Two trees: Can only batch claim for one reward currency", async function() {
```
nicholaspai, line 512: @daywiss new test based on the bug you found, showing that its now impossible to make batch claim for multiple reward currencies.

nicholaspai:
> looks good! just want to mention the slack convo, theres currently not a way to generate additional deposit events. this means its really hard to mirror state through pure events. i dont think this is a blocker since it cant be strongly enforced but would be nice if we had this option.

@mrice32 @chrismaree PTAL

I see no downside to adding the deposit method except added gas for the caller (versus just transferring it in). We could alternatively strictly enforce the deposit method by keeping an internal reward balance.

chrismaree:
> > looks good! just want to mention the slack convo, theres currently not a way to generate additional deposit events. this means its really hard to mirror state through pure events. i dont think this is a blocker since it cant be strongly enforced but would be nice if we had this option.
> 
> @mrice32 @chrismaree PTAL
> 
> I see no downside to adding the deposit method except added gas for the caller (versus just transferring it in). We could alternatively strictly enforce the deposit method by keeping an internal reward balance.

I think having an additional deposit method is a good idea. It makes the process of adding additional tokens more formalised and the additional event will ensure that we can re-build state. I'm not a huge fan of internal balances to enforce this though, especially anyone could always send tokens to the contract. Not really any issues if this was to happen though.

nicholaspai:
> > > looks good! just want to mention the slack convo, theres currently not a way to generate additional deposit events. this means its really hard to mirror state through pure events. i dont think this is a blocker since it cant be strongly enforced but would be nice if we had this option.
> 
> > 
> 
> > @mrice32 @chrismaree PTAL
> 
> > 
> 
> > I see no downside to adding the deposit method except added gas for the caller (versus just transferring it in). We could alternatively strictly enforce the deposit method by keeping an internal reward balance.
> 
> 
> 
> I think having an additional deposit method is a good idea. It makes the process of adding additional tokens more formalised and the additional event will ensure that we can re-build state. I'm not a huge fan of internal balances to enforce this though, especially anyone could always send tokens to the contract. Not really any issues if this was to happen though.

I don't like having a deposit method that isn't necessary to ever use (ie you can just transfer in, and it's cheaper to do it that way). I'm going to abstain from adding that method.

PR #2655, opened 2021-03-05
feat(price-feed): generalizes the DefiPulseTotalPriceFeed to work for individual DeFi projects

bryanjcampbell1:
**Title**
feat(price-feed): generalizes the DefiPulseTotalPriceFeed to work for individual DeFi projects

**Motivation**
As it stands, the DefiPulseTotalPriceFeed only creates a price feed for the sum of all projects listed on defipulse.com.  This proposed change opens up the creation of TVL price feeds for individual projects as well as ratios like SUSHIUNI as described in UMIP 24.

**Summary**
Changes were made to the DefiPulseTotalPriceFeed, CreatePriceFeed, and DefaultPriceFeedConfigs.  Tests were also modified to be compatible with these changes.

**Details**
An extra config param called "project" was added to specify the protocol that we are interested in fetching TVL data for.  

**Issue(s)**

Fixes the current lack of a price feed for individual defi projects.


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -107,28 +107,29 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
-      message: "Creating DefiPulseTotalPriceFeed",
+      message: "Creating DefiPulsePriceFeed",
       config
     });
 
-    return new DefiPulseTotalPriceFeed(
+    return new DefiPulsePriceFeed(
       logger,
       web3,
-      config.defipulseApiKey,
```
nicholaspai, line 126: You should leave this named `defipulseApiKey`. The reason is this [issue](https://github.com/UMAprotocol/protocol/issues/2588). If the `apiKey` has the same variable name for multiple price feeds, then it can make it tricky for liquidator bot runners to not overload the variable

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -377,9 +377,26 @@ const defaultConfigs = {
...
     lookback: 604800,
-    minTimeBetweenUpdates: 600
+    minTimeBetweenUpdates: 600,
+    project: "all"
+  },
+  DEFI_PULSE_SUSHI_TVL: {
+    type: "defipulse",
+    lookback: 604800,
+    minTimeBetweenUpdates: 600,
+    project: "all"
```
nicholaspai, line 389: This should be `sushiswap` or something like that right? Same with Uniswap below

packages/financial-templates-lib/src/price-feed/DefiPulsePriceFeed.js
```diff
@@ -15,8 +15,19 @@ class DefiPulseTotalPriceFeed extends PriceFeedInterface {
...
+  constructor(
+    logger,
+    web3,
+    apiKey,
+    lookback,
+    networker,
+    getTime,
+    minTimeBetweenUpdates,
+    priceFeedDecimals = 18,
+    project
```
nicholaspai, line 29: Can you validate the value of `project` in the constructor so that the user knows what is acceptable values? For example:

```
constructor(...project) {
const VALID_PROJECTS = ["sushiswap", "uniswap"]
assert(VALID_PROJECTS.includes(project), "Invalid project string");
}
```

This way we won't get potentially obscure errors later when making a request to a URL with a bad `project=` param

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -1221,4 +1223,38 @@ contract("CreatePriceFeed.js", function(accounts) {
...
+    const validConfig = {
+      type: "defipulse",
+      lookback: 604800,
+      apiKey: apiKey,
+      minTimeBetweenUpdates: 600,
+      project: "SushiSwap"
+    };
+
+    assert.equal(await createPriceFeed(logger, web3, networker, getTime, { ...validConfig, apiKey: undefined }), null);
+    assert.equal(await createPriceFeed(logger, web3, networker, getTime, { ...validConfig, project: undefined }), null);
```
nicholaspai, line 1250: Related to the comment I had about sanity checking `project`, can you add some tests for invalid project names like "Sushi"

nicholaspai (commented):
This looks great overall and is going to be an awesome feature. Can you review the comments I made? I also like the name change from `DefiPulseTotalPriceFeed` to the. more generalized `DefiPulsePriceFeed`!

daywiss:
@bryanjcampbell1 just fyi i've updated the title of pr to match conventional commit guidelines

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
bryanjcampbell1: Done

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
bryanjcampbell1: Since the type="defipulse" is used in CreatePriceFeed.js to identify the use of DefiPulsePriceFeed, shouldn't the type stay as "defipulse" for both suhi and uni?

packages/financial-templates-lib/src/price-feed/DefiPulsePriceFeed.js
bryanjcampbell1: Added check for project in CreatePriceFeed

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
bryanjcampbell1: Done

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -107,28 +107,34 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
-    const requiredFields = ["lookback", "minTimeBetweenUpdates", "defipulseApiKey"];
+  } else if (config.type === "defipulse") {
+    const requiredFields = ["lookback", "minTimeBetweenUpdates", "defipulseApiKey", "project"];
+    const VALID_PROJECTS = ["all", "SushiSwap", "Uniswap"];
 
     if (isMissingField(config, requiredFields, logger)) {
       return null;
     }
 
+    if (!VALID_PROJECTS.includes(config.project)) {
```
daywiss, line 118: do you think project should default to one of these, maybe `all`?

daywiss: also can we make this a case insensitive comparison

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
daywiss: i think he means project:"all" should be project:"SushiSwap"? ditto with uni

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -377,9 +377,26 @@ const defaultConfigs = {
...
+  },
+  DEFI_PULSE_UNISWAP_TVL: {
+    type: "defipulse",
+    lookback: 604800,
+    minTimeBetweenUpdates: 600,
+    project: "all"
+  },
+  SUSHIUNI: {
+    type: "expression",
+    expression: "10 * DEFI_PULSE_SUSHI_TVL / DEFI_PULSE_UNISWAP_TVL"
```
daywiss, line 399: getting a lot of mileage out of this feature @mrice32 🚚

daywiss (commented):
Nice upgrade! Just a couple comments but looks good for the most part

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
bryanjcampbell1: I think we stay away from defaulting to all.  IMO the "all" synth may be hard to sustain without a large fee paid to LPs. I don't know many people who are shorting Defi in the long term.  Most likely we will find more users willing to short individual projects.

Also, it seems the case matters in the API request. "all" returns a response but "All" returns an error.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
bryanjcampbell1: Ah got ya. Made change.

daywiss (approved):
LGTM

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -107,28 +107,34 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.priceFeedDecimals, // This defaults to 18 unless supplied by user
       uniswapBlockCache
     );
-  } else if (config.type === "defipulsetvl") {
-    const requiredFields = ["lookback", "minTimeBetweenUpdates", "defipulseApiKey"];
+  } else if (config.type === "defipulse") {
+    const requiredFields = ["lookback", "minTimeBetweenUpdates", "defipulseApiKey", "project"];
+    const VALID_PROJECTS = ["all", "SushiSwap", "Uniswap"];
```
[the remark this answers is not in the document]
nicholaspai: For example, add an `assert(VALID_PROJECTS.includes(project), "invalid project name")` to the constructor and then add the requisite tests

nicholaspai (changes requested):
Thanks for responding to my previous comments, have one more change i'd like to see

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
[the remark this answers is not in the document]
bryanjcampbell1: I'm trying to write a test to check that an error is thrown when an invalid name is passed as a param in DefiPulsePriceFeed.js 

`it("Bad project name", async function() {

    badProjectPriceFeed = new DefiPulsePriceFeed(
      dummyLogger,
      web3,
      defipulseApiKey,
      lookback,
      networker,
      getTime,
      minTimeBetweenUpdates,
      decimals, // Add arbitrary decimal conversion and prove this works.
      "NotAProject"
    );
   
    assert(badProjectPriceFeed instanceof assert.AssertionError);

  });`

The problem is that my test throws an error because DefiPulsePriceFeed() thows an assert error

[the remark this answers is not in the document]
bryanjcampbell1: Any tips would be greatly appreciated!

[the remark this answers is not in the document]
daywiss: theres probably cleverer ways but ive seen this in some other tests.  you could also assert on error.message that it matches expected message
```js
            let errorThrown = false;
            try {
              await liquidator.liquidatePositions();
            } catch (error) {
              errorThrown = true;
            }
            assert.isTrue(errorThrown);
```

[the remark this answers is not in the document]
nicholaspai: yeah you could also do probably do something like:

```
(new PriceFeed(...badConfigs)).catch(err => assert(true)).then(assert(false)
```

nicholaspai (approved):
LGTM, check the one comment I have on making the test less robust, but the current implementation works and the tests look right

PR #2662, opened 2021-03-08
fix(affiliates): explicitly set version of emp abi

### LINKED ISSUE #2661: dev mining event decoding throws an error when running
**Description**

When running dev mining ever since 2.0 release, there is a crash. This is due to wrong emp abi being used to decode events on already deployed emps. 

**Reproduction**
Run dev mining as normal. This will eventually throw an error. 

**Expected Behavior**

Should return dev mining results.

**Screenshots**

```js
Error: no matching event (argument="topichash", value="0x342d1375efc923bbfe4f5b6575b7d47f40247499113546f8689206d3fd1cf82f", code=INVALID_ARGUMENT, version=abi/5.0.7)
    at Logger.makeError (/protocol/node_modules/@ethersproject/logger/src.ts/index.ts:205:28)
    at Logger.throwError (/protocol/node_modules/@ethersproject/logger/src.ts/index.ts:217:20)
    at Logger.throwArgumentError (/protocol/node_modules/@ethersproject/logger/src.ts/index.ts:221:21)
    at Interface.getEvent (/protocol/node_modules/@ethersproject/abi/src.ts/interface.ts:219:20)
    at Interface.parseLog (/protocol/node_modules/@ethersproject/abi/src.ts/interface.ts:553:29)
    at /protocol/packages/affiliates/libs/contracts.js:19:16
    at /protocol/packages/affiliates/libs/affiliates/devmining.js:33:16
    at /protocol/node_modules/highland/lib/index.js:2221:25
    at Stream.s._send (/protocol/node_modules/highland/lib/index.js:1560:9)
    at Stream.write (/protocol/node_modules/highland/lib/index.js:1661:18)
    at Stream._send (/protocol/node_modules/highland/lib/index.js:984:26)
    at Stream.write (/protocol/node_modules/highland/lib/index.js:1661:18)
    at ResourceStream.ondata (internal/streams/readable.js:719:22)
    at ResourceStream.emit (events.js:315:20)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9) {
  reason: 'no matching event',
  code: 'INVALID_ARGUMENT',
  argument: 'topichash',
  value: '0x342d1375efc923bbfe4f5b6575b7d47f40247499113546f8689206d3fd1cf82f'
}
```

daywiss:

**Motivation**

#2661

**Summary**

Allows a way to specify EMP version for the abi when decoding events.  This adds additional config param when specifying emp address and payout address. 

**Details**
Several other important changes were grouped into this PR for convenience:
- bug fix: cryptowatch key was not being set correctly causing rate limits
- feat: optional tradermade key configuration added
- feat: added ability to fetch fallback price from spreadsheet
- feat: logged out the json object which caused errors when serializing data in unix pipe. this helps with diagnosing reason why stringify fails. 
- improve: commented out error logging when parsing rows in spreadsheet to reduce confusing logs. errored rows are still ignored.
- improve: a promise.map was changed to promise.mapSeries to slow down api requests for requesting historical price feed since so many emps are being included.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2661

### DIFF
--- a/packages/affiliates/apps/DevMiningRewards.js
+++ b/packages/affiliates/apps/DevMiningRewards.js
@@ -27,10 +27,6 @@ const App = env => async params => {
   assert(endTime, "requires endTime");
   assert(totalRewards, "requires totalRewards");
 
-  // This just sets a default abi version in case no abi is passed along with the emp address.
-  // This should default to the latest version
-  const defaultEmpAbi = getAbi("ExpiringMultiParty");
-
   const emp = Emp({ web3 });
   const client = new BigQuery();
   const queries = Queries({ client });
@@ -43,10 +39,13 @@ const App = env => async params => {
 
   const rewards = DevMining({
     queries,
-    defaultEmpAbi,
     coingecko,
     synthPrices
   });
+
+  // This just sets a default abi version in case no abi is passed along with the emp address.
+  // This should default to the latest version
+  const defaultEmpAbi = getAbi("ExpiringMultiParty");
 
   // API has changed, we need to validate input. Emps will be required to include payout address.
   empWhitelist = empWhitelist.map(empInput => {
@@ -60,10 +59,7 @@ const App = env => async params => {
 
     // this converts a version number from the config into an abi which gets passed into the calculator
     // so each emp contract will have the correct abi passed along with it.
-    let empAbi;
-    if (empVersion) {
-      empAbi = getAbi("ExpiringMultiParty", empVersion);
-    }
+    const empAbi = empVersion ? getAbi("ExpiringMultiParty", empVersion) : defaultEmpAbi;
 
     return [empAddress, payoutAddress, empAbi];
   });
--- a/packages/affiliates/apps/DevMiningRewards.js
+++ b/packages/affiliates/apps/DevMiningRewards.js
@@ -27,7 +27,9 @@ const App = env => async params => {
   assert(endTime, "requires endTime");
   assert(totalRewards, "requires totalRewards");
 
-  const empAbi = getAbi("ExpiringMultiParty");
+  // we will eventually need to know the versions of each emp being run in order to decode events
+  // but for now this works to run on all current emps.
+  const empAbi = getAbi("ExpiringMultiParty", "1.2.0");
 
   const emp = Emp({ web3 });
   const client = new BigQuery();
--- a/packages/affiliates/apps/DevMiningRewards.js
+++ b/packages/affiliates/apps/DevMiningRewards.js
@@ -27,19 +27,23 @@ const App = env => async params => {
   assert(endTime, "requires endTime");
   assert(totalRewards, "requires totalRewards");
 
-  // we will eventually need to know the versions of each emp being run in order to decode events
-  // but for now this works to run on all current emps.
-  const empAbi = getAbi("ExpiringMultiParty", "1.2.0");
+  // This just sets a default abi version in case no abi is passed along with the emp address.
+  // This should default to the latest version
+  const defaultEmpAbi = getAbi("ExpiringMultiParty");
 
   const emp = Emp({ web3 });
   const client = new BigQuery();
   const queries = Queries({ client });
   const coingecko = Coingecko();
-  const synthPrices = SynthPrices({ web3, apiKey: env.CRYPTOWATCH_KEY });
+  const synthPrices = SynthPrices({
+    web3,
+    cryptowatchApiKey: env.CRYPTOWATCH_KEY,
+    tradermadeApiKey: env.TRADERMADE_KEY
+  });
 
   const rewards = DevMining({
     queries,
-    empAbi,
+    defaultEmpAbi,
     coingecko,
     synthPrices
   });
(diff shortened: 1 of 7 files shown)


chrismaree (approved):

chrismaree:
Future versions could use the `FindContractVersion` util that the bots use but for now this is quick and easy. LGTM

packages/affiliates/apps/DevMiningRewards.js
```diff
@@ -27,7 +27,9 @@ const App = env => async params => {
   assert(endTime, "requires endTime");
   assert(totalRewards, "requires totalRewards");
 
-  const empAbi = getAbi("ExpiringMultiParty");
+  // we will eventually need to know the versions of each emp being run in order to decode events
+  // but for now this works to run on all current emps.
+  const empAbi = getAbi("ExpiringMultiParty", "1.2.0");
```
mrice32, line 32: Does this work for this week? There are 2.0.0 EMPs that are deployed and have $$$ in them.

daywiss: thats a good question. probably not

daywiss:
i've updated this to take emp version input per contract from a config. It will fallback to the default abi (latest) if no version is provided so changes are backwards technically backward compatible, though older contracts must specify their version explicitly in order for this to run.  These changes were used to run the latest dev mining PR #2667 

Other changes:
- bug fix: cryptowatch key was not being set correctly causing rate limits
-  feat: optional tradermade key configuration added

packages/affiliates/apps/GetSheetsDevminingStatus.js
```diff
@@ -82,7 +84,8 @@ function parseSheet(sheet, columns = SHEET_COLUMNS) {
       try {
         result[name] = map ? map(row[index]) : row[index];
       } catch (err) {
-        console.error("error in row:" + row, err);
+        // These errors can be distracting, and we really just ignore them for now.
```
mrice32, line 87: Are these errors common? What causes them?

packages/affiliates/libs/affiliates/devmining.js
```diff
@@ -12,16 +12,18 @@ const { EmpBalancesHistory } = require("../processors");
...
   assert(coingecko, "requires coingecko api");
   assert(synthPrices, "requires synthPrices api");
 
   // use firstEmpDate as a history cutoff when querying for events. We can safely say no emps were deployed before Jan of 2020.
   firstEmpDate = firstEmpDate || moment("2020-01-01", "YYYY-MM-DD").valueOf();
 
-  async function getBalanceHistory(empAddress, start, end) {
+  async function getBalanceHistory(empAddress, start, end, empAbi = defaultEmpAbi) {
+    assert(empAddress, "requires empAddress");
+    assert(empAbi, "requires empAbi");
```
mrice32, line 26: Is this check just to guard against null being passed in? Because empAbi has a default...?

mrice32 (approved):
Looks great! Mostly nits and a few questions, but I think this is good to merge whenever.

packages/affiliates/apps/DevMiningRewards.js
```diff
@@ -47,7 +53,19 @@ const App = env => async params => {
...
+
+    // we want to make sure these addresses are standarized since we do many lookups internally
+    empAddress = rewards.utils.toChecksumAddress(empAddress);
+
+    // this converts a version number from the config into an abi which gets passed into the calculator
+    // so each emp contract will have the correct abi passed along with it.
+    let empAbi;
+    if (empVersion) {
+      empAbi = getAbi("ExpiringMultiParty", empVersion);
+    }
```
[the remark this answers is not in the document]
daywiss: i can consider changing this.  the only reason i didnt was because this would change the behavior of the lower level functions and more tests would need to be updated.

packages/affiliates/apps/GetSheetsDevminingStatus.js
daywiss: the spreadsheet can be quite messy, with misformatted fields, missing data or other weird things. If a row errors during parsing we ignore it. This was causing a lot of chatter when evaldo runs automation which i think was confusing for him since the errors could be ignored. 

we just have to make sure on the spreadsheet side that if a row needs to be included, that is formatted correctly, and we can reconcile this with the output from the spreadsheet script.

packages/affiliates/libs/affiliates/devmining.js
daywiss: it was to maintain backward compatibilility if you did not provide an explicit emp Abi. But I think i might pass it in the way you recommended on first comment

packages/affiliates/libs/affiliates/devmining.js
```diff
@@ -300,7 +306,7 @@ module.exports = ({ queries, empAbi, coingecko, synthPrices, firstEmpDate }) =>
         collateralTokens,
         async address => await getCoingeckoPriceHistory(address, "usd", startTime, endTime)
       ),
-      Promise.map(
+      Promise.mapSeries(
```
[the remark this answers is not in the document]
daywiss: this was a change when trying to understand why i was rate limited. it would slow down the calls slightly, executing requests one at a time vs all at once. I left it in after realizing it was an issue with the api key, because its probably better to call the api slower anyway.

packages/affiliates/libs/affiliates/utils.js
```diff
@@ -329,7 +340,14 @@ function makeUnixPipe(cb, stdin = process.stdin) {
       // we need to do this to extract the result of the promise in the stream.
       .flatMap(highland)
       // we want our result to be also a valid json string, so we can continue chaining the pipeline
-      .map(x => JSON.stringify(x, null, 2))
+      .map(x => {
```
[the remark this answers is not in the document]
daywiss: ive updated the PR description with all these changes you have pointed out. thanks for being so throrough

PR #2672, opened 2021-03-09
feat(merkle-distributor): Make claimMulti more generalized

nicholaspai:
**Motivation**

Caller should be able to batch arbitrary group of unclaimed claims together and claim them in one transaction. Unit test is added to show gas savings.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


nicholaspai:
Compare this PR to #2677 implementaiton which performs some precomputations in exchange for code complexity.

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -617,6 +600,21 @@ contract("MerkleDistributor.js", function(accounts) {
...
+        batchedClaims.push({
+          windowIndex: windowIndex,
+          account: leaf.account,
+          accountIndex: leaf.accountIndex,
+          amount: leaf.amount,
+          merkleProof: proof
+        });
+      }
+      const tx = await merkleDistributor.claimMulti(batchedClaims);
+      assert.equal(Math.floor(tx.receipt.gasUsed / batchedClaims.length), 56846);
```
nicholaspai, line 616: Here

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -639,6 +637,21 @@ contract("MerkleDistributor.js", function(accounts) {
...
+        batchedClaims.push({
+          windowIndex: windowIndex,
+          account: leaf.account,
+          accountIndex: leaf.accountIndex,
+          amount: leaf.amount,
+          merkleProof: proof
+        });
+      }
+      const tx = await merkleDistributor.claimMulti(batchedClaims);
+      assert.equal(Math.floor(tx.receipt.gasUsed / batchedClaims.length), 38401);
```
nicholaspai, line 653: Here

nicholaspai:
Should be replaced by #2677

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -127,32 +127,16 @@ contract MerkleDistributor is Ownable {
...
+    // Batch claims to reduce gas versus individual submitting all claims.
+    function claimMulti(Claim[] memory claims) external {
         for (uint256 i = 0; i < claims.length; i++) {
-            Claim memory claim = claims[i];
-            require(claim.account == account, "Invalid account in batch claim");
-            address _rewardToken = address(merkleWindows[claim.windowIndex].rewardToken);
-            require(_rewardToken == rewardToken, "Invalid rewardToken in batch claim");
-            _verifyAndMarkClaimed(claim);
-            amountToClaim = amountToClaim.add(claim.amount);
+            claim(claims[i]);
```
chrismaree, line 133: personally, I prefer this implementation over the one presented in [2677](https://github.com/UMAprotocol/protocol/pull/2677/files). it's easier to reason about and cleaner. downside is more token transfers. However, from your tests it looks like it uses _less_ gas, which is quite surprising!

nicholaspai: Also [this test from 2677](https://github.com/UMAprotocol/protocol/pull/2677/files#r591517458) is the most realistic IMO (1 account, many trees, many reward currencies) shows that this implementation performs better than #2677

daywiss: how is this possible?

daywiss (approved):
this simplifies the contract a lot and reuses the regular claim call which is great.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -127,34 +127,36 @@ contract MerkleDistributor is Ownable {
...
-        Claim[] memory claims,
-        address rewardToken,
-        address account
-    ) external {
-        uint256 amountToClaim = 0;
+    // Batch claims to reduce gas versus individual submitting all claims. Optimistically tries to batch
+    // together consecutive claims for the same account and same reward token to reduce gas.
+    function claimMulti(Claim[] memory claims) external {
+        uint256 batchedAmount = 0;
         for (uint256 i = 0; i < claims.length; i++) {
```
daywiss, line 134: does caching claims.length in a variable outside the loop give you any gas savings? so then you are referencing that variable in the loop rather than claims.length

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -127,34 +127,36 @@ contract MerkleDistributor is Ownable {
...
+            batchedAmount = batchedAmount.add(_claim.amount);
+
+            // If the next claim is NOT the same account or the same token (or this claim is the last one),
+            // then disburse the `batchedAmount` to the current claim's account for the current claim's reward token.
+            uint256 nextI = i.add(1);
+            address currentRewardToken = address(merkleWindows[_claim.windowIndex].rewardToken);
+            if (
+                nextI == claims.length || // This claim is last claim.
+                address(merkleWindows[claims[nextI].windowIndex].rewardToken) != currentRewardToken || // Next claim reward token is different than current one.
+                claims[nextI].account != _claim.account // Next claim account is different than current one.
```
daywiss, line 146: not quite as clean as the single claim call in the loop, but i do like this solution

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: It saves ~10 gas per call! nice catch

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: Because #2677 only gives you gas-reductions if some of the claims re-use the same {account, rewardToken} combo, which this test did not, so it was the worst case test for #2677

nicholaspai:
I'm going to close #2677 in favor of this one due to (1) less code and (2) better gas optimization due to the latest implementation which optimistically batches transfers (h/t @mrice32 for the idea).

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -127,34 +127,37 @@ contract MerkleDistributor is Ownable {
...
+        uint256 batchedAmount = 0;
+        uint256 claimCount = claims.length;
+        for (uint256 i = 0; i < claimCount; i++) {
+            Claim memory _claim = claims[i];
+            _verifyAndMarkClaimed(_claim);
+            batchedAmount = batchedAmount.add(_claim.amount);
+
+            // If the next claim is NOT the same account or the same token (or this claim is the last one),
+            // then disburse the `batchedAmount` to the current claim's account for the current claim's reward token.
+            uint256 nextI = i.add(1);
```
mrice32, line 142: We might be able to drop the safemath since this can't overflow.

So something like:
```
uint256 nextI = i + 1;
```

Thoughts?

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -127,34 +127,37 @@ contract MerkleDistributor is Ownable {
...
+            address currentRewardToken = address(merkleWindows[_claim.windowIndex].rewardToken);
+            if (
+                nextI == claimCount || // This claim is last claim.
+                address(merkleWindows[claims[nextI].windowIndex].rewardToken) != currentRewardToken || // Next claim reward token is different than current one.
+                claims[nextI].account != _claim.account // Next claim account is different than current one.
+            ) {
+                uint256 amountToDisburse = batchedAmount;
+                batchedAmount = 0;
+                IERC20(currentRewardToken).safeTransfer(_claim.account, amountToDisburse);
```
mrice32, line 151: Why can't we do something like this?
```suggestion
                IERC20(currentRewardToken).safeTransfer(_claim.account, batchedAmount);
                batchedAmount = 0;
```

mrice32 (approved):
LGTM!

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I'd rather the external call come last for reentrancy reasons

mrice32: I don't think re-entrancy can affect this variable because batchedAmount is an in-memory variable that's local to this level in the stack. i.e. if you re-entered, batchedAmount would be pointing to a different spot in memory and couldn't interfere with the original call.

But I also get being cautious around re-entrancy. Up to you.

mrice32 (approved):
LGTM!

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
nicholaspai: I'll test the change you proposed for gas changes

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -698,6 +688,31 @@ contract("MerkleDistributor.js", function(accounts) {
...
+            // Gets the reward token for the window data for this claim
+            // and injects it into the claim object.
+            return {
+              ...claimsArray[i],
+              rewardToken: windowData.rewardToken
+            };
+          })
+          .sort((a, b) => {
+            // If a.account == b.account, then sorts by rewardToken
+            return a.account.localeCompare(b.account) || a.rewardToken.localeCompare(b.rewardToken);
```
mrice32, line 712: Took me a minute to digest this -- neat implementation of primary + secondary sort on strings where we don't care about order, just sort consistency and grouping.

mrice32 (approved):

packages/core/test/merkle-distributor/MerkleDistributor.js
nicholaspai: Yeah the previous implementation wasn't correctly sorting, because you can't assume that different window indices don't re-use reward tokens. (before I was sorting on window index + account).

Also, took me a while to realize that you can't make the `.sort()` comparator function `async` otherwise it just doesn't sort.

PR #2682, opened 2021-03-10
feat(KPI-Options): add simple script for calculating UMA TVL

chrismaree:
**Motivation**

As part of KPI options we require the ability to quickly and easily calculate the UMA TVL over _all_ financial contracts. This PR adds a simple script that computes the current TVL. It also produces some useful metadata that we might want to store later on to produce pie charts of different TVL contributors and time series data to plot contract sizes over time.

**Summary**

This set of functions are intended to sit behind a serveless function. In the future, it could be called by a persistent, stateful API. Or, alternatively, a periodic servless function that stores the information into a database for later retrieval.

**Details**
The main method of consideration in this PR is `calculateCurrentTvl`. This function, when called, will return the current UMA tvl. 

**How the TVL is computed**
The script follows these basic steps:
1) Find all contracts registered in the `Registry`. This includes _all_ contracts that can make requests to the DVM, including all financial contracts including EMPs, perps, and non-UMA contracts like Jarvis's contracts. Note that the OO would also show up in this list.
2) for each registered contract, find information about the collateral currency. In particular, find `collateralCurrency`.
3) In a set of batched calls, query at the same time the following information about the collateral currency:
3.1) the number of `decimals` the collateral has. This is needed to scale values.
3.2) the `balanceOf` of the financial contract in the collateral currency. I.e for a given financial contract, how much collateral is within the contract.
3.3) the price of the collateral currency using Coingecko. Coingecko provides a very convenient API end point that can query an array of addresses at once to find their value.
Note that 3.1,3.2,3.3 these queries all can be done in parallel.
4) We now have completed all async info we need to compute TVL. Next, join the data together to form an array of financial contract information that contains: a) address of the financial contract b) address of the collateral currency c) financial contract balance of collateral d) the number of decimals in the collateral currency e) the price of each unit of collateral in USD.
5) using the above information produced in 4, we can compute the `collateralValueInUsd` for each contract as the product of the amount of collateral and the price per collateral, scaled by the decimals of the collateral currency.
6) finally, we can preform a reduce operation over all contracts to get the total TVL.

Note that the goal of this PR was to implement the final step, number 6 in the list above. However, there is a wealth of other useful information also produced in this script which we can cache at a later point for graphing. A subsequent PR will be release to add this data to a persistent storage solution for retrieval.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/merkle-distributor/kpi-options-helpers/calculate-uma-tvl.ts
```diff
@@ -0,0 +1,154 @@
...
+
+  // If the contract does not have the method called then return null.
+  return collateralAddresses.map((response: any) =>
+    response.status === "fulfilled" ? response.value.toLowerCase() : null
+  );
+}
+
+// Returns an array of decimals associated with an array of collateral addresses. Note if any contract is not implemented
+// (null) or the decimals method is not implemented this will return null for that collateral token.
+export async function fetchCollateralDecimals(collateralTokenAddresses: Array<string>) {
```
daywiss, line 108: this is just fetching decimals from an erc20, doesnt need to be specific to collateral

daywiss (approved):
this is great chris and really quick turn around. most my comments are about design and i dont think its necessarily a blocker for this PR. But as I am seeing this and thinking about it more, i see this generalized as more of a "client" type api, similar to what the bots have and what i kind of did in [affiliates here](https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/libs/contracts.js#L114) where its a general client to get useful data back in a structured way. 

A generalized client could be used in the broader uma sdk in both front end and backend stuff, as well as in these type of bespoke value calculators, affiliates, etc. let me know what you think!

packages/merkle-distributor/kpi-options-helpers/calculate-uma-tvl.ts
```diff
@@ -0,0 +1,154 @@
...
+      contractInstance ? contractInstance.methods.balanceOf(financialContractAddresses[index]).call() : null
+    )
+  );
+
+  return collateralDecimals.map((response: any) => (response.status === "fulfilled" ? response.value : null));
+}
+
+// Return an array of spot prices for an array of collateral addresses in one async call. Note we might in future
+// want to change this to re-use the bot's price feeds for more complex collateral types like LP tokens.
+export async function fetchCollateralValue(collateralTokenAddresses: Array<string>) {
```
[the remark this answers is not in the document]
chrismaree: sure and done! agreed we could use the lib in future.

packages/merkle-distributor/kpi-options-helpers/calculate-uma-tvl.ts
chrismaree: good point! it's not just for collateral. I renamed it to be contract decimals. I did the same thing for balances as well.

PR #2684, opened 2021-03-11
fix(affiliates): winston logger configured to log only to stderr

### LINKED ISSUE #2683: winston logging to console breaks affiliate script composibility
**Description**

When errors happen during dev mining run within a price feed, winston is currently configured to log to console. This ends up breaking the output when piping it into the next app for consumption.

**Reproduction**
This is currnetly happening when trying to run the tradermade feed for ustonks.

**Expected Behavior**

Errors should log to std.err and any level below that can be ignored.

daywiss:
**Motivation**

#2683 

**Summary**

Just adjust winston logger to send errors to std.err

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2683

### DIFF
--- a/packages/affiliates/libs/synthPrices.js
+++ b/packages/affiliates/libs/synthPrices.js
@@ -9,7 +9,14 @@ module.exports = ({ web3, cryptowatchApiKey, tradermadeApiKey, decimals = 18 } =
     from = from / 1000;
     to = to / 1000;
     const priceFeed = await createReferencePriceFeedForFinancialContract(
-      winston.createLogger({ transports: [new winston.transports.Console()] }),
+      winston.createLogger({
+        transports: [
+          new winston.transports.Console({
+            level: "error",
+            stderrLevels: ["error"]
+          })
+        ]
+      }),
       web3,
       new Networker(),
       () => to, // starting time


packages/affiliates/libs/synthPrices.js
```diff
@@ -9,7 +9,14 @@ module.exports = ({ web3, cryptowatchApiKey, tradermadeApiKey, decimals = 18 } =
     from = from / 1000;
     to = to / 1000;
     const priceFeed = await createReferencePriceFeedForFinancialContract(
-      winston.createLogger({ transports: [new winston.transports.Console()] }),
+      winston.createLogger({
+        transports: [
+          new winston.transports.Console({
+            level: "error",
+            stderrLevels: ["emerg", "alert", "crit", "error"]
```
chrismaree, line 16: I dont think you need to define the log levels above the log you are curious in. you should be able to just define `error` and it will log everything above that.

chrismaree (approved):

packages/affiliates/libs/synthPrices.js
```diff
@@ -9,7 +9,14 @@ module.exports = ({ web3, cryptowatchApiKey, tradermadeApiKey, decimals = 18 } =
     from = from / 1000;
     to = to / 1000;
     const priceFeed = await createReferencePriceFeedForFinancialContract(
-      winston.createLogger({ transports: [new winston.transports.Console()] }),
+      winston.createLogger({
+        transports: [
+          new winston.transports.Console({
+            level: "error",
```
nicholaspai, line 15: Why do you define `level: "error"`? I don't think this affects what gets sent to `stdErr`, which is defined in the `stderrLevels` key below. 

I think what you're doing here by setting `level: error` is only logging error or higher logs, which might be what you're intending but FYI

mrice32: Agreed. @daywiss why not just apply this to _all_ log levels?

mrice32 (approved):
LGTM outside of @nicholaspai's comment.

packages/affiliates/libs/synthPrices.js
daywiss: maybe my understanding is wrong, but i thought this would log all messages at level `error` or more critical, and suppress any info/debug logs.

Then in stderLevels i want any logs error and above to log to `error` and not `out`

nicholaspai: Ah then I think you're right. You're specifying you want want logs at error or higher

PR #2686, opened 2021-03-11
fix(docker): point docker build to use commands in package.json rather than hardcoded commands

chrismaree:
**Motivation**

The docker build process was previously using commands that are repeated in the package.json. This PR updates this to have one source of truth.


daywiss (approved):

nicholaspai (approved):
Nice consolidation

mrice32 (approved):
LGTM!

PR #2687, opened 2021-03-11
fix(funding-rate-proposer): Make sure funding rate proposal timestamps use current EVM time, not current real world time

nicholaspai:
**Motivation**

When war games testing the FundingRateProposer, @mrice32 helped me to realize that the `requestTimestamp` was being set to the current clock time, which is technically in the future from the EVM's point of view, causing unexpected reverts. 

I've added a `test` mode option to the proposer bot so that unit tests against the Timer's artificial time can still be run. The default setting is `test=false` which uses the latest block time as the request timestamp.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/funding-rate-proposer/index.js
```diff
@@ -20,7 +20,6 @@ const { getWeb3 } = require("@uma/common");
  * @notice Runs strategies that request and propose new funding rates for Perpetual contracts.
  * @param {Object} logger Module responsible for sending logs.
  * @param {Object} web3 web3.js instance with unlocked wallets used for all on-chain connections.
- * @param {String} perpetualAddress Contract address of the Perpetual Contract.
```
nicholaspai, line 23: Unused variable in this bot, perp addresses auto detected by querying PerpFactory past events

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -98,7 +98,7 @@ class FundingRateProposer {
     ]);
   }
 
-  async updateFundingRates() {
+  async updateFundingRates(useRealTime = false) {
```
mrice32, line 101: Might be worth a note here about what environments useRealTime is set to true for.

mrice32 (approved):
LGTM!

chrismaree (approved):
great catch!

Gamaranto:
Missing DCO then gtg @nicholaspai

PR #2689, opened 2021-03-12
feat(provider-utils): enable network parameterization 

chrismaree:
**Motivation**

Right now the `getWeb3()` method is limited to pulling the network from the args used when running the node process. This is somewhat restrictive, especially when it comes to using this function in serverless. 

**Summary**

This PR enables the network to be parameterized when using this method.

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@uma/common",
-  "version": "2.0.1",
+  "version": "2.0.2",
   "description": "Common js utilities used by other UMA packages",
   "homepage": "http://umaproject.org",
   "license": "AGPL-3.0-or-later",
--- a/packages/common/src/ProviderUtils.js
+++ b/packages/common/src/ProviderUtils.js
@@ -57,14 +57,14 @@ function createBasicProvider(url) {
  * which load a hardcoded infura key. Default port is 9545.
  *
  */
-function getWeb3() {
+function getWeb3(parameterizedNetwork = "test") {
   // If a web3 instance has already been initialized, return it.
   if (web3) {
     return web3;
   }
 
   // Create basic web3 provider with no wallet connection based on the url alone.
-  const network = argv.network || "test"; // Default to the test network (local network).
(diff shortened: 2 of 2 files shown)


nicholaspai (approved):

packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uma/common",
-  "version": "2.0.1",
+  "version": "2.0.2",
```
mrice32, line 3: If we're going to bump the version, should we manually write release notes? Or we could separate these into 2 PRs and do a full release on the following PR to generate release notes automatically.

Either way is fine with me -- we should just denote the version bump and reasoning in the release notes.

mrice32 (commented):
Looks great!

Minor comment around release notes.

Since we're depending on this in production, I think it may make sense for us to add an automatic `@next` release to ci using lerna. This means on each push, all packages would be released and tagged to `@next`. We could point any of our production systems to that so we can use the bleeding edge without needing to pull the entire repo or do manual releases to get access to it.

Thoughts?

chrismaree:
> Looks great!
> 
> Minor comment around release notes.
> 
> Since we're depending on this in production, I think it may make sense for us to add an automatic `@next` release to ci using lerna. This means on each push, all packages would be released and tagged to `@next`. We could point any of our production systems to that so we can use the bleeding edge without needing to pull the entire repo or do manual releases to get access to it. Just curious to hear your thoughts. I don't think we need to change anything immediately. Just might be something we should do as a follow-up.
> 
> Thoughts?

This would be _awesome_ and is something right now that I'm actually needing in production. For example, Some serverless functions I've been working on require the _latest_ contracts we only just added to master but are not part of any package. If I could simply make this dependent on @next, knowing that everything we've got in master would be there, it would be great. We Can then bump things to work with the latest stable version when we do releases.

packages/common/package.json
chrismaree: I'm not sure opinionated on this tbh. I'm happy either way :)

PR #2690, opened 2021-03-12
feat(packages): update version of truffle-ledger-provider to avoid libudev

chrismaree:
**Motivation**

Our codebase was previously dependent on `libudev` due to the `truffle-ledger-provider`. This limited the portability of much of the code by preventing it to run unless this module was installed. Normally, this is not an issue as you can install it in Mac/Linux pretty easily. However, this is not always the case, as in serverless functions where the access to the core operating system is more limited.

This PR updates the packages in common to avoid this requirement.

Note that this PR bumps the version of common from 2.0.1 to 2.0.3 as RP [2689](https://github.com/UMAprotocol/protocol/pull/2689) should be merged first.


packages/common/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@uma/common",
-  "version": "2.0.1",
+  "version": "2.0.3",
```
mrice32, line 3: nit: do we need to release a new version? If you install `2.0.1` fresh, it picks up the patch automatically since truffle-ledger-provider has a `^`. Also why a 2-version bump?

mrice32 (commented):
Not sure we need to release a new version of common. Other than that, everything else seems good!

packages/common/package.json
chrismaree: that works for me and makes sence! Let's get the #2689 PR merged and then we can rebase this onto of that one with the 2.0.2 version.

chrismaree:
@mrice32 I've reverted the changes to common. this PR now just updates the lock file.

mrice32 (approved):

PR #2691, opened 2021-03-12
feat(Networker): enable support for POST requests in getJson()

### LINKED ISSUE #2692: Support for POST requests in Networker.getJson()
**Why**

What is the motivation for this feature or task?

Price feeds that want to use subgraphs as data source requires support for POST requests in [Networker.getJson()](https://github.com/UMAprotocol/protocol/blob/cadccb00a7cf01a82c2e30b1b5af263876419793/packages/financial-templates-lib/src/price-feed/Networker.js#L15).
Usually to fetch data from subgraphs we need to send a POST request with GraphQL query as a payload.
It would be great if Networker can have support for this. Using Networker makes testing much easier.

**How**

#2691

**Resolution**

POST requests can be made using `Network.getJson()` 

**Difficulty Score [1-10]**

- 1: <=5 minutes

ashutoshvarma:
**Motivation**
For price feed that wants to use subgraphs as data source requires support for POST requests in `Networker.getJson()`.
Usually to fetch data from subgraphs we need to send a POST request with GraphQL query as a payload.

While working on[ RAI x UMA challenge ](https://gitcoin.co/issue/reflexer-labs/geb/97/100024834) I needed to write price feed using subgraphs.
It would be great if `Networker` can have support for this. Using `Networker` makes testing much easier.

**Summary**

- It doesn't require any big change, just allowing [`option`](https://www.npmjs.com/package/node-fetch#fetchurl-options) param to `fetch` call is required.
```diff
-  async getJson(url) {
-    const response = await fetch(url);
+  async getJson(url, options) {
+    const response = await fetch(url, options);
```
- Also, this doesn't break existing code depending on `Networker`.

To make POST request,
```js
data = {
  query: `
    {
      redemptionRates{
        perSecondRate
      }
    }
  `,
};
options = {
  method: "POST",
  body: JSON.stringify(data)
}

networker.getJson("SUBGRAPH_URL", options)
```

**Testing**

I don't think this needs a test, 

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2692

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/Networker.js
+++ b/packages/financial-templates-lib/src/price-feed/Networker.js
@@ -12,8 +12,8 @@ class Networker {
     this.logger = logger;
   }
 
-  async getJson(url) {
-    const response = await fetch(url);
+  async getJson(url, options) {
+    const response = await fetch(url, options);
     const json = await response.json();
     if (!json) {
       this.logger.error({


mrice32 (approved):
This is awesome! Agreed, this is a low-risk change that makes the class much more flexible.

PR #2695, opened 2021-03-12
improve(dependencies): Upgrade web3 and truffle related packages

nicholaspai:
…

**Motivation**

Packages install different versions of truffle and web3 dependencies independently, so I bumped all to latest. Also added the packages to the root's `devDependencies` and removed them from individual package dev deps

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/merkle-distributor/package.json
```diff
@@ -39,5 +36,8 @@
   },
   "bugs": {
     "url": "https://github.com/UMAprotocol/protocol/issues"
+  },
+  "devDependencies": {
```
nicholaspai, line 40: This was needed for me to run build this package

packages/serverless-orchestration/package.json
```diff
@@ -23,9 +23,7 @@
     "moment": "^2.24.0",
     "node-fetch": "^2.3.0",
     "supertest": "^4.0.2",
-    "truffle": "^5.1.16",
```
nicholaspai, line 26: This package was 30 patches behind the other packages haha

packages/core/package.json
```diff
@@ -34,11 +34,9 @@
     "moment": "^2.24.0",
     "node-fetch": "^2.3.0",
     "solidity-coverage": "^0.7.2",
-    "truffle": "^5.1.16",
```
nicholaspai, line 37: This was also 30 patches behind other packages

package.json
```diff
@@ -26,6 +26,10 @@
     "type": "git",
     "url": "https://github.com/UMAprotocol/protocol.git"
   },
+  "dependencies": {
```
mrice32, line 29: I don't think this will work.

If these packages are installed through npm, they will not inherit these root-level packages.

So, for instance, if you installed `@uma/common` and used `getWeb3()` the web3 dep would be missing I believe since that dep wouldn't be in `@uma/common`'s package.json.

There may be other ways to do this though, as I share your sentiment that it would be nice to standardize our dep versions across packages.

Keeping dev-dependencies at the root level _could_ make sense because dev dependencies are only meant to be installed/used in the repo and not when installing through npm.

Thoughts?

chrismaree (approved):
Wow. Some wild differences. Hope this makes things more stable!

package.json
nicholaspai: I thought each package inherited deps from the root deps?

mrice32: Not that I'm aware of.

See docs here: https://classic.yarnpkg.com/en/docs/workspaces/.

> Be careful when publishing packages in a workspace. If you are preparing your next release and you decided to use a new dependency but forgot to declare it in the package.json file, your tests might still pass locally if another package already downloaded that dependency into the workspace root. However, it will be broken for consumers that pull it from a registry, since the dependency list is now incomplete so they have no way to download the new dependency. Currently there is no way to throw a warning in this scenario.

mrice32: Let me know if you see something in the docs that implies otherwise!

nicholaspai: Hmmm this is a good point, and probably not detected in the tests because the root `yarn build` is run first, which allows each package to access web3 and truffl

nicholaspai: Ok so I've re-added all `web3` and `truffle`'s to individual package package.jsons, but also moved them to the root `devDependencies` (and removed individual package dev dependencies).

packages/merkle-distributor/.d.ts
```diff
@@ -0,0 +1 @@
+declare module "ethereumjs-util";
```
nicholaspai, line 1: Had to add this file because of the error when trying to build `merkle-distributor`:

```
@uma/merkle-distributor: src/MerkleTree.ts(3,40): error TS7016: Could not find a declaration file for module 'ethereumjs-util'. '/home/circleci/protocol/node_modules/ethereumjs-util/dist/index.js' implicitly has an 'any' type.
@uma/merkle-distributor:   Try `npm i --save-dev @types/ethereumjs-util` if it exists or add a new declaration (.d.ts) file containing `declare module 'ethereumjs-util';`
@uma/merkle-distributor: error Command failed with exit code 1.
@uma/merkle-distributor: info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

packages/merkle-distributor/package.json
```diff
@@ -8,15 +8,13 @@
     "@uma/core": "^2.0.1",
     "chai": "^4.3.0",
     "commander": "^7.1.0",
+    "ethereumjs-util": "^7.0.9",
```
nicholaspai, line 11: Took me a while to figure this out: this package includes the `types` needed for line 3 of src/MerkleTree.ts, not intuitively the `@types/ethereumjs-util` package.

packages/common/package.json
```diff
@@ -36,7 +36,7 @@
     "@nomiclabs/hardhat-etherscan": "^2.1.0",
     "@nomiclabs/hardhat-truffle5": "^2.0.0",
     "@nomiclabs/hardhat-web3": "^2.0.0",
-    "@truffle/hdwallet-provider": "^1.0.25",
+    "@truffle/hdwallet-provider": "^1.2.3",
```
nicholaspai, line 39: Bumped this as well because its involved with signing transactions from our custom provider

packages/merkle-distributor/package.json
mrice32: Does it only need the types? Are there any runtime deps pulled from it? If it's _only_ types, we could move it to devDependencies.

package.json
mrice32: Sounds good. Will take another look. Note: this will never be caught by the tests unless the tests are run from _outside the repo_ by installing the package.

packages/funding-rate-proposer/package.json
```diff
@@ -8,7 +8,7 @@
     "@uma/financial-templates-lib": "^2.0.1",
     "async-retry": "^1.3.1",
     "dotenv": "^6.2.0",
-    "truffle": "^5.1.35"
+    "truffle": "^5.2.3"
```
mrice32, line 11: Does truffle need to be a regular dep in this package? I don't know if it _ever_ needs to be a regular dependency... unless there's a `require("truffle")` floating around...?

packages/liquidator/package.json
```diff
@@ -9,7 +9,7 @@
     "@umaprotocol/ynatm": "^0.0.1",
     "async-retry": "^1.3.1",
     "dotenv": "^6.2.0",
-    "truffle": "^5.1.35"
+    "truffle": "^5.2.3"
```
mrice32, line 12: Same question here.

packages/monitors/package.json
```diff
@@ -11,7 +11,7 @@
     "@uma/financial-templates-lib": "^2.0.1",
     "async-retry": "^1.3.1",
     "dotenv": "^6.2.0",
-    "truffle": "^5.1.35"
+    "truffle": "^5.2.3"
```
mrice32, line 14: Same question here.

packages/optimistic-oracle/package.json
```diff
@@ -8,7 +8,7 @@
     "@uma/financial-templates-lib": "^2.0.1",
     "async-retry": "^1.3.1",
     "dotenv": "^6.2.0",
-    "truffle": "^5.1.35"
+    "truffle": "^5.2.3"
```
mrice32, line 11: Same question here.

packages/serverless-orchestration/package.json
```diff
@@ -5,7 +5,9 @@
   "dependencies": {
     "@truffle/contract": "^4.2.20",
     "@uma/common": "^2.0.1",
-    "@uma/financial-templates-lib": "^2.0.1"
+    "@uma/financial-templates-lib": "^2.0.1",
+    "truffle": "^5.2.3",
+    "web3": "^1.3.4"
```
mrice32, line 10: Is there a `require("truffle")` or `require("web3")` in any non-tests in this package? If not, I think we could probably drop these and rely on the top-level dev dependencies.

mrice32 (commented):
A few minor comments -- also, once you finish moving these deps around, we should rerun `yarn` to make sure the yarn.lock file gets updated with the most recent changes.

packages/serverless-orchestration/package.json
mrice32: I think there is a `require("web3")` in this package, but we can probably drop truffle.

packages/financial-templates-lib/package.json
```diff
@@ -22,7 +22,6 @@
   },
   "devDependencies": {
```
mrice32, line 23: I think we need to actually add `web3` as a runtime dep of this package even though it currently isn't there.

packages/funding-rate-proposer/package.json
daywiss: agreed yea was wondering if it should be a dev dep

daywiss (approved):
looks good, i think theres just some questions around what should be in dev dependencies

package.json
nicholaspai: yeah tahts a really good point: each package's tests should start with a fresh environment ideally to test under worst case conditions

packages/merkle-distributor/package.json
nicholaspai: It only needs types i can move to devDeps

packages/funding-rate-proposer/package.json
nicholaspai: This is a great point, i'll try moving all truffle imports to devDeps

packages/financial-templates-lib/package.json
nicholaspai: Yeah at the very least its imported in ExpressionPriceFeed

nicholaspai:
@daywiss @pemulis @mrice32 @chrismaree Have requested you all because this PR potentially introduces some breaking changes. I've noticed that bumping the web3 version changes how certain types are parsed, for example I've had to use `web3.utils.padRight` to cast `utf8ToHex("identifier")`'s to the correct `bytes` form to call methods that require an identifier. Additionally, I noticed that `getRandomUnsignedInt()` no longer returns a correct length `int256` data type. Will add comments where relevant.

I've also tested the voting dApp locally to make sure these version bumps don't change anything there but am triple checking

packages/core/test/common/SignedFixedPoint.js
```diff
@@ -393,7 +393,7 @@ contract("SignedFixedPoint", function() {
     // +-1e-18 / +-1e19 = -1e-37, which can't be represented and gets floor'd to 0.
     quotient = await fixedPoint.wrapDiv("-1", toWei(toWei("10")));
     assert.equal(quotient.toString(), "0");
-    quotient = await fixedPoint.wrapDiv("1", toWei(toWei("-10")));
+    quotient = await fixedPoint.wrapDiv("1", toBN(toWei(toWei("-10"))));
```
nicholaspai, line 396: Without this `toBN` I receive the following error:

```
  1) Contract: SignedFixedPoint
       Division:
     Error: invalid BigNumber string (argument="value", value="0000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000-10000000000000000000000000000000000000", code=INVALID_ARGUMENT, version=bignumber/5.0.15)
      at Context.<anonymous> (test/common/SignedFixedPoint.js:396:33)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiParty.js
```diff
@@ -29,7 +29,7 @@ contract("ExpiringMultiParty", function(accounts) {
       collateralAddress: collateralToken.address,
       tokenAddress: syntheticToken.address,
       finderAddress: finder.address,
-      priceFeedIdentifier: web3.utils.utf8ToHex("TEST_IDENTIFIER"),
+      priceFeedIdentifier: padRight(utf8ToHex("TEST_IDENTIFIER"), 64),
```
nicholaspai, line 32: This casting is now neccessary otherwise I get invalid length errors

packages/core/test/oracle/Governor.js
```diff
@@ -182,14 +182,20 @@ contract("Governor", function(accounts) {
...
+      ...pendingRequests[1],
+      identifier: padRight(pendingRequests[1].identifier, 64)
+    };
     assert.equal(web3.utils.hexToUtf8(request1.identifier), `Admin ${id1}`);
     assert.equal(web3.utils.hexToUtf8(request2.identifier), `Admin ${id2}`);
 
     // Execute the proposals to clean up.
     const vote = toWei("1");
-    const salt = getRandomUnsignedInt();
+    const salt = getRandomSignedInt();
```
nicholaspai, line 198: These salts all now have to be random ints not random uints, otherwise I get invalid length errors:

```
  8) Contract: Governor
       Unresolved vote:
     Error: value out-of-bounds (argument="salt", value="99956580464701296385646310288893919711651560260231478960720509407514263887743", code=INVALID_ARGUMENT, version=abi/5.0.0-beta.153)
      at Context.<anonymous> (test/oracle/Governor.js:608:18)
      at runMicrotasks (<anonymous>)
      at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

packages/core/test/oracle/Voting.js
```diff
@@ -1502,12 +1502,17 @@ contract("Voting", function(accounts) {
 
     await voting.snapshotCurrentRound(signature);
 
-    assert(await didContractThrow(voting.revealVote(identifier, time, getRandomSignedInt(), getRandomUnsignedInt())));
-    await voting.revealVote(identifier, time, retrievedVote.price, retrievedVote.salt);
+    assert(await didContractThrow(voting.revealVote(identifier, time, getRandomSignedInt(), getRandomSignedInt())));
+    await voting.revealVote(
+      identifier,
+      time,
+      web3.utils.toBN(retrievedVote.price),
```
nicholaspai, line 1509: similar issue as in `SignedFIxedPoint.js` where I have to cast a negative `int` to a BN before passing it as input into a function call

packages/serverless-orchestration/package.json
nicholaspai: I dropped truffle, kept web3

packages/core/test/oracle/Finder.js
```diff
@@ -11,8 +12,8 @@ contract("Finder", function(accounts) {
   it("General methods", async function() {
     const finder = await Finder.deployed();
 
-    const interfaceName1 = web3.utils.hexToBytes(web3.utils.utf8ToHex("interface1"));
-    const interfaceName2 = web3.utils.hexToBytes(web3.utils.utf8ToHex("interface2"));
```
nicholaspai, line 15: I guess truffle or web3 no longer casts `bytes` correctly anymore to pass into a method like `changeImplementationAddress` which expects a `bytes32` argument?

packages/core/test/oracle/Governor.js
mrice32: Seems reasonable. I think they should've been ints anyway.

packages/voter-dapp/src/ActiveRequests.js
```diff
@@ -398,7 +398,7 @@ function ActiveRequests({ votingAccount, votingGateway, snapshotContract }) {
       }
       const identifierPrecision = getPrecisionForIdentifier(hexToUtf8(pendingRequests[index].identifier));
       const price = parseFixed(editState[index], identifierPrecision).toString();
-      const salt = getRandomUnsignedInt().toString();
+      const salt = getRandomSignedInt().toString();
```
mrice32, line 401: Did you QA this change?

mrice32 (approved):
LGTM!

packages/voter-dapp/src/ActiveRequests.js
nicholaspai: Yeah I did

PR #2696, opened 2021-03-13
fix(networker): throw on error rather than logging an error

chrismaree:
**Motivation**
The networker has a error log that should be changed to throw an error. When the Networker fails to complete a request it will produce an error even though it correctly fetched pricing information in the subsequent execution.

An example of the error produced can be seen below:
image

However, the bot _did_ correctly finish its execution cycle, after the retry logic re-initiated the price request. see screenshot below:
image

As a result, This PR updates the networker to rather throw an error such that the error can be bubbled up to the price feed using it, enabling the bot to re-run its execution and optimistically return a value price on the next retry loop.

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/Networker.js
+++ b/packages/financial-templates-lib/src/price-feed/Networker.js
@@ -16,6 +16,7 @@ class Networker {
     const response = await fetch(url);
     const json = await response.json();
     if (!json) {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
Makes sense to throw as early as possible

nicholaspai (approved):
Makes sense to throw as early as possible

PR #2697, opened 2021-03-15
feat(cloud-functions): initial simple cloud functions for calculating UMA TVL and fetching KPI options recipients

chrismaree:
**Motivation**

This PR adds three initial serverless functions to aid the deployment of KPI options. 

**Summary**

The two functions added do the following:
1. `uma-tvl-calculator` wraps the `merkle-distributor-helper` package's `calculateCurrentTvl` method to create a simple interface for computing the current UMA TVL. Once the TVL is calculated, this method will store it within GCP data store with the Key being the computation timestamp. This should be run by a cron job to ensure the data store always has the most up to data information.
 2. `uma-tvl-fetcher` pulls data from GCP data store to quickly return the most recent TVL.
 3. `Merkle-distributor-helper` wraps the the `merkle-distributor-helper` package's `getClaimsForAddress` method to create a simple interface to fetch claims for a particular address.

**Details**

These functions are designed to be as simple as posible and simply wrap the functions in a serverless compatible interface. Note that NO unit tests are added. We _could_ add unit tests quite easily using the `@google-cloud/functions-framework` package which lets you run serverless functions locally. However, seeing these functions wrap already tested underling logic, the testing was deemed unimportant at this point.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested
- [X] Tested locally using the `@google-cloud/functions-framework` package.


nicholaspai (commented):
Drive-by: do you forsee us longer-term adding the `serverless-orchestration` scripts into the `cloud-functions` package?

packages/cloud-functions/.gitignore
```diff
@@ -0,0 +1 @@
+*/yarn.lock
```
nicholaspai, line 1: This seems like you only needed to ignore the `yarn.lock` locally, probably shouldn't go in this PR

nicholaspai (approved):
I really support the naming of the `cloud-functions` package, seems like this package has a lot of potential to scale the ecosystem

packages/cloud-functions/.gitignore
chrismaree: main reason I had this is that I dont want the yarn lock files to be included within any of the sub packages of the `cloud-functions`. The lock file will be produced when building locally but I dont want this to be included in the repo for these scripts seeing they tend to have only one or two dependencies.

packages/cloud-functions/README.md
```diff
@@ -0,0 +1,13 @@
+# @uma/merkle-distributor
+
+This package contains a number of serverless functions used to wrap core UMA logic in a serverless consumable format.
+
+## Key packages included:
```
mrice32, line 5: Do you want these packages to be first-class citizens within our lerna setup? Or do you want lerna to effectively ignore them?

packages/cloud-functions/merkle-distributor-helper/index.js
```diff
@@ -0,0 +1,18 @@
+const { getClaimsForAddress } = require("@uma/merkle-distributor");
```
mrice32, line 1: Just curious -- why not typescript?

packages/cloud-functions/uma-tvl-calculator/index.js
```diff
@@ -0,0 +1,27 @@
+const { Datastore } = require("@google-cloud/datastore");
+const datastore = new Datastore();
+
+const { calculateCurrentTvl } = require("@uma/merkle-distributor");
+
+exports.calculateCurrentTvl = async (req, res) => {
```
mrice32, line 6: In the GCP configuration, is this one going to be publicly accessible too? Or have you configured it s.t. this one is only accessible via cloud scheduler, but the accessors are public?

packages/cloud-functions/uma-tvl-calculator/index.js
```diff
@@ -0,0 +1,27 @@
+const { Datastore } = require("@google-cloud/datastore");
```
mrice32, line 1: Not relevant for this PR, but having a git-ops style configuration for these might be cool -- where the config part gets committed and how the deployment gets triggered might take some thought, but it could have some nice advantages.

mrice32 (commented):
This is really really cool. I love the simplicity of these functions and the separation of concern. Really elegant.

The only real questions I have are around how we set this cloud-functions package up, how we test it, and how it's handled by lerna/the repo. We don't need all the answers now, but I just want to get the conversation started.

packages/cloud-functions/README.md
chrismaree: Ideally, I think Lerna should ignore them. these are meant to be really self standing with minimal code within them. Right now there is a separate .gitignore that skips their `yarn.lock` files. Do you agree? 

My main reasoning is people should not really install/deploy/use the packages on their own outside of serveless and so having it in Lerna is not really any gain.

packages/cloud-functions/merkle-distributor-helper/index.js
chrismaree: Looks like Cloud functions _do_ support it. Additionally, the package `@google-cloud/functions-framework` contains the appropriate types for the exported methods. I avoided it initially as all the googling I did around it involves a fair amount _more_ setup, tsconfig, and a few other dependencies to get it to function 100%. We can try changing this in the future I think.

I also saw a few cool patterns around building more complex sets of serverless functions such as [this](https://github.com/flatfisher/cloud-functions-typescript-template) repo and [this](https://github.com/robisim74/firebase-functions-typescript-starter) repo.

Also, the functions-framework-nodejs (a package I used to run tests against the serverless functions locally) supports typescript. check it out [here](https://github.com/GoogleCloudPlatform/functions-framework-nodejs)

chrismaree: In fact the functions-framework-nodejs has some very simple [getting started on using typescript](https://github.com/GoogleCloudPlatform/functions-framework-nodejs/blob/master/docs/typescript.md). would recommend you look at this @mrice32 before building the cashing engine for the price feeds we discussed IRL during spring planning.

packages/cloud-functions/uma-tvl-calculator/index.js
chrismaree: this is not accessible publicly. it's called directly from a scheduler on a recurring basis (every 5 mins right now)

packages/cloud-functions/uma-tvl-calculator/index.js
chrismaree: for sure! future improvement :)

packages/cloud-functions/uma-tvl-calculator/index.js
mrice32: +1 -- thanks for the explanation. Makes sense!

packages/cloud-functions/merkle-distributor-helper/index.js
mrice32: >  I avoided it initially as all the googling I did around it involves a fair amount more setup, tsconfig, and a few other dependencies to get it to function 100%. We can try changing this in the future I think.

Makes sense. This is a pretty new setup, so I totally get trying to reduce the number of unknowns. 

From what you said it sounds like it's highly doable, but agree that it's probably not worth it until after we get TVL options up and running and stable.

packages/cloud-functions/README.md
mrice32: I generally agree with the sentiment, but I want to make sure that this doesn't cause any unintended side effects. Some things I'm a bit worried about:
- `yarn` at root will not install their packages. You will have to manually jump to each directory and install.
- Similarly `yarn test` will not automatically pick up the `test` script in these packages if it exists. So we would either need to manually add the test command somewhere or not run tests in these packages.
- If these are implemented in typescript, the same thing would apply. Any calls to `yarn qbuild` or `yarn build` would not automatically hit these packages.

We can include things in lerna/yarn workspaces and _not_ publish them. The voter-dapp works this way. We just need to set it as private in the package.json and it will never be published.

mrice32 (commented):
Looks great!

I thought a little more about the setup, and I think these packages should be included in yarn workspaces/lerna: https://github.com/UMAprotocol/protocol/pull/2697/files#r596984938. Happy to help configure it (shouldn't be too hard).

packages/cloud-functions/README.md
mrice32: Just tried this out locally. All you should need to do to have lerna and yarn pick up these new packages is change the root level package.json:
```
  "workspaces": [
    "packages/**"
  ]
```

and lerna.json:
```
  "packages": ["packages/**"],
```

chrismaree: nice find! I've made the updates :) thanks for the input!

packages/cloud-functions/.gitignore
mrice32, line 1: @chrismaree now that these are in the yarn workspace, you should be able to drop this I think....? yarn shouldn't produce a lock file I don't think (at least when running in the repo).

packages/cloud-functions/merkle-distributor-helper/package.json
```diff
@@ -0,0 +1,7 @@
+{
```
mrice32, line 1: To prevent these from trying to publish, you might want to add:
```
"private": true
```

to each of these package.json files.

mrice32 (approved):

packages/cloud-functions/.gitignore
chrismaree: good point! have removed it.

packages/cloud-functions/merkle-distributor-helper/package.json
chrismaree: yup! done.

PR #2698, opened 2021-03-15
feat(Merkle-distributor): Added payout ingestor

chrismaree:
**Motivation Summary**
This PR adds an ingestor script that reads in a number of constituent payouts and generates a combined payouts file. An example of this usage would be to take in a) liquidity b) developer and c) dapp mining payouts and produce one combined output that would distribute all underlying rewards. The initial usage of this script is to inject the KPI outputs payouts (defined in the official UMA repo [here](https://github.com/UMAprotocol/token-distribution)) which contains rewards for a) governance_recipients, uma_holder_recipients and uma_user_recipients and build one payout file containing all three.

**Example usage**

The script is designed to accommodate n number arbitrary input files to be combined into one payout file. These should be formatted sequentially after the execution flag. See the example execution to understand the expected format. The example below combines 3 payout files, for window index 0, on network id 42 (Kovan) with a test UMA Token. Each input file has no key within it and is simply a mapping of addresses to payout amounts.

```
ts-node ./scripts/0_MerklePayoutIngestor.ts --input ../../outputs/governance_recipients.json ../../outputs/uma_holder_recipients.json \
../../outputs/uma_user_recipients.json -- key null null null --decimals 18 18 18 -- reason "KPI Options drop 1: Governance Recipient" \
"KPI Options drop1: UMA Token Holder" "KPI Options drop1: UMA User" -- rewardToken 0x47B1EE6d02af0AA5082C90Ea1c2c14c70399186c -- chainId 42 -- windowIndex 0
```

Note that in this example there are three input files. therefore, the` input` has three strings after it, the `key` has 3 nulls (all three have no nested keys), there are three `decimals` and three `reason`s.

Once executed, this file will produce an output file that has the following structure:
```
{
  "chainId": 42,
  "rewardToken": "0x47B1EE6d02af0AA5082C90Ea1c2c14c70399186c",
  "windowIndex": 0,
  "totalRewardsDistributed": "976200000000000000000000",
  "recipients": {
    "0x1aba978cf66d4bd8ce54ab5e3aaff177ab901fb6": {
      "amount": "270000000000000000000",
      "metaData": [
        "KPI Options drop 1: Governance Recipient",
        "KPI Options drop1: UMA Token Holder",
        "KPI Options drop1: UMA User"
      ],
      "accountIndex": 0
    },
    "0x9efab36288565dc343b6909350767cb392aa4f69": {
      "amount": "270000000000000000000",
      "metaData": [
        "KPI Options drop 1: Governance Recipient",
        "KPI Options drop1: UMA Token Holder",
        "KPI Options drop1: UMA User"
      ],
      "accountIndex": 1
    },
... for all recipients
}
```

**Testing**

- [ ]  New unit tests created
- [X]  Untested


packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
```diff
@@ -0,0 +1,103 @@
...
+          metaData: [o.reason[i]],
+          accountIndex: 0
+        };
+      }
+    });
+  }
+
+  // Sort the outputs payment amount.
+  outputObject.recipients = Object.fromEntries(
+    Object.entries(outputObject.recipients).sort(([, a], [, b]) => (toBN(a.amount).gt(toBN(b.amount)) ? -1 : 0))
```
nicholaspai, line 81: Why not just make the comparator function used as the `.sort()` callback: `return toBN(b.amount).gt(toBN(a.amount))`?

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
```diff
@@ -0,0 +1,103 @@
...
+  );
+
+  // Append an index to each recipient and count the totalRewardsDistributed.
+  let totalRewardsDistributed = toBN("0");
+  Object.keys(outputObject.recipients).forEach((recipientAddress: string, index: number) => {
+    outputObject.recipients[recipientAddress].accountIndex = index;
```
nicholaspai, line 87: I'd run a sanity check, before  writing the file to the desk, that all account indices are unique. If they are not unique it could mean `verifyProof` fails unexpectedly for some claims. You could try using `lodash`'s `uniqBy` method like so:

```js
const _ = require("lodash")
const uniqueIndices = _.uniqBy(outputObject.recipients, "accountIndex")
assert.equal(uniqueIndices.length, outputObject.recipients.length, "duplicate account indices")
```

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
```diff
@@ -0,0 +1,103 @@
...
+    recipients: {} // Built up from the n input files defined in the params.
+  };
+
+  // Build up payouts for recipients for each input file.
+  for (let i = 0; i < o.input.length; i++) {
+    const inputFile = JSON.parse(fs.readFileSync(o.input[i], { encoding: "utf8" }));
+
+    // Fetch the recipients from the input file. The key donates the identifier within the input file. if this is null
+    // then the input file has no nested structures and only contains the recipients.
+    const recipients = o.key == "null" ? inputFile[o.key] : inputFile;
```
nicholaspai, line 53: Just confirming my understanding of the `key` param, say we're using a [gas-rebates JSON as input](https://github.com/UMAprotocol/protocol/blob/master/packages/affiliates/gas-rebate/rebates/Rebate_7.json):

Would `key = "shareHolderPayout"`?

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
```diff
@@ -0,0 +1,103 @@
...
+    Object.keys(recipients).forEach((recipientAddress: string) => {
+      // Scale the amount by the number of decimals for that particular input.
+      const recipientAmountScaled = ConvertDecimals(0, o.decimals[i], Web3)(recipients[recipientAddress]);
+
+      // If the output file already contains information for this particular recipient, then append and add their rewards.
+      // Else, simply init the object with their values from the file. Note that accountIndex in both cases is set to 0.
+      // This will be set once the full data structure is built and ordered by payout amount.
```
nicholaspai, line 60: Maybe set the default `accountIndex` to some more clearly invalid value like -1, or -99999? Something to indicate to the client that something is wrong if it hasn't been reset

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
```diff
@@ -0,0 +1,103 @@
...
+  };
+
+  // Build up payouts for recipients for each input file.
+  for (let i = 0; i < o.input.length; i++) {
+    const inputFile = JSON.parse(fs.readFileSync(o.input[i], { encoding: "utf8" }));
+
+    // Fetch the recipients from the input file. The key donates the identifier within the input file. if this is null
+    // then the input file has no nested structures and only contains the recipients.
+    const recipients = o.key == "null" ? inputFile[o.key] : inputFile;
+    Object.keys(recipients).forEach((recipientAddress: string) => {
```
nicholaspai, line 54: Should we `checksum` the `recipientAddress`, so that `0xaD` and `0xAd` don't get treated as separate recipients?

nicholaspai (approved):
Nice script, seems really generalizable and powerful. Left some nits. Can you also add a comment to the top of the file similar to your PR description where you just show an example execution command?

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: because the sort function needs a number less than zero to swap (hence -1) or 0 or create to not swap.

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: ye, that's a good check! will add it.

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: exactly. it just makes it a bit more flexible to enable any format of payout json file.

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: sure

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: also a good call. done

packages/merkle-distributor/scripts/0_MerklePayoutIngestor.ts
chrismaree: using `[...new Set(Object.keys(outputObject.recipients))]` worked out much faster than lodash. using this. (this is es10 syntax I believe)

chrismaree:
> Nice script, seems really generalizable and powerful. Left some nits. Can you also add a comment to the top of the file similar to your PR description where you just show an example execution command?

Sure! have added a bunch of comments throughout the script and at the top.

PR #2717, opened 2021-03-16
fix(affiliates): error process if no price found

### LINKED ISSUE #2716: dev mining should error out if no price can be found for emp
**Description**

Currently dev mining will log and swallow errors if no price is found via synthetic price feed, fallback price or coingecko price. This should terminate the process with an error for this condition.

daywiss:
**Motivation**

#2716 

**Summary**

We want process to exit with a relevant error and readable logs if no price is found.
This also adds a secondary minor fix: moves dapp mining rewards from a previous week into correct folder.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2716


nicholaspai (approved):
Won't this suppress the stack trace?

daywiss:
> Won't this suppress the stack trace?

yes but only for non critical errors. these errors will just log the message and keep going. the critical error will show full stack trace

PR #2718, opened 2021-03-16
feat(fallback-pricefeed): Create FallBackPriceFeed and ForexPriceFeed

nicholaspai:
**Motivation**

The TraderMadePriceFeed has been empirically unstable, inspiring this new pricefeed which allows the user to specify a priority list of pricefeeds to "fallback" through. For example, the user might use this new pricefeed to first try to get a price from the TraderMadePriceFeed for the EURUSD identifier, but if that fails then they can use the more reliable but less granular ForexPriceFeed for daily EURUSD prices.

**Summary**

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -3,6 +3,7 @@ const { ChainId, Token, Pair, TokenAmount } = require("@uniswap/sdk");
 const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
 const { FallBackPriceFeed } = require("./FallBackPriceFeed");
 const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
+const { ForexDailyPriceFeed } = require("./ForexDailyPriceFeed");
 const { DefiPulsePriceFeed } = require("./DefiPulsePriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
 const { BalancerPriceFeed } = require("./BalancerPriceFeed");
@@ -107,6 +108,30 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.invertPrice, // Not checked in config because this parameter just defaults to false.
       config.priceFeedDecimals, // This defaults to 18 unless supplied by user
       uniswapBlockCache
+    );
+  } else if (config.type === "forexdaily") {
+    const requiredFields = ["base", "symbol", "lookback"];
+
+    if (isMissingField(config, requiredFields, logger)) {
+      return null;
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating ForexDailyPriceFeed",
+      config
+    });
+
+    return new ForexDailyPriceFeed(
+      logger,
+      web3,
+      config.base,
+      config.symbol,
+      config.lookback,
+      networker,
+      getTime,
+      config.priceFeedDecimals, // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
+      config.minTimeBetweenUpdates // Defaults to 43200 (12 hours) unless supplied.
     );
   } else if (config.type === "defipulse") {
     const requiredFields = ["lookback", "minTimeBetweenUpdates", "defipulseApiKey", "project"];
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -1,6 +1,7 @@
 const assert = require("assert");
 const { ChainId, Token, Pair, TokenAmount } = require("@uniswap/sdk");
 const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
+const { FallBackPriceFeed } = require("./FallBackPriceFeed");
 const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
 const { DefiPulsePriceFeed } = require("./DefiPulsePriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
@@ -154,6 +155,28 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
 
     // Loop over all the price feeds to medianize.
     return await _createMedianizerPriceFeed(config);
+  } else if (config.type === "fallback") {
+    const requiredFields = ["orderedFeeds"];
+
+    if (isMissingField(config, requiredFields, logger)) {
+      return null;
+    }
+
+    if (config.orderedFeeds.length === 0) {
+      logger.error({
+        at: "createPriceFeed",
+        message: "FallBackPriceFeed configured with 0 feeds🚨"
+      });
+      return null;
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating FallBackPriceFeed",
+      config
+    });
+
+    return await _createFallBackPriceFeed(config);
   } else if (config.type === "balancer") {
     const requiredFields = ["balancerAddress", "balancerTokenIn", "balancerTokenOut", "lookback", "twapLength"];
 
@@ -436,11 +459,22 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
     return new ExpressionPriceFeed(priceFeedMap, expressionConfig.expression, expressionConfig.priceFeedDecimals);
   }
 
-  // Returns a MedianizerPriceFeed
   async function _createMedianizerPriceFeed(medianizerConfig) {
-    const priceFeedsToMedianize = [];
-    for (const _priceFeedConfig of medianizerConfig.medianizedFeeds) {
-      // The medianized feeds should inherit config options from the parent config if it doesn't define those values
+    const priceFeedsToMedianize = await _createConstituentPriceFeed(medianizerConfig.medianizedFeeds);
+    if (!priceFeedsToMedianize) return null;
+    return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
+  }
+
+  async function _createFallBackPriceFeed(fallbackConfig) {
+    const orderedPriceFeeds = await _createConstituentPriceFeed(fallbackConfig.orderedFeeds);
+    if (!orderedPriceFeeds) return null;
+    return new FallBackPriceFeed(orderedPriceFeeds);
+  }
+
+  async function _createConstituentPriceFeed(priceFeedConfigs) {
+    const priceFeeds = [];
+    for (const _priceFeedConfig of priceFeedConfigs) {
+      // The constituent feeds should inherit config options from the parent config if it doesn't define those values
       // itself.
       // Note: ensure that type isn't inherited because this could create infinite recursion if the type isn't defined
       // on the nested config.
@@ -454,9 +488,9 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
         return null;
       }
 
-      priceFeedsToMedianize.push(priceFeed);
+      priceFeeds.push(priceFeed);
     }
-    return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
+    return priceFeeds;
   }
 
   // Returns an array or "basket" of MedianizerPriceFeeds
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -485,18 +485,18 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
   }
 
   async function _createMedianizerPriceFeed(medianizerConfig) {
-    const priceFeedsToMedianize = await _createConstituentPriceFeed(medianizerConfig.medianizedFeeds);
+    const priceFeedsToMedianize = await _createConstituentPriceFeeds(medianizerConfig.medianizedFeeds);
     if (!priceFeedsToMedianize) return null;
     return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
   }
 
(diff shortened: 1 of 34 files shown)


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -436,11 +484,22 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
   async function _createMedianizerPriceFeed(medianizerConfig) {
-    const priceFeedsToMedianize = [];
-    for (const _priceFeedConfig of medianizerConfig.medianizedFeeds) {
-      // The medianized feeds should inherit config options from the parent config if it doesn't define those values
+    const priceFeedsToMedianize = await _createConstituentPriceFeed(medianizerConfig.medianizedFeeds);
+    if (!priceFeedsToMedianize) return null;
+    return new MedianizerPriceFeed(priceFeedsToMedianize, medianizerConfig.computeMean);
+  }
+
+  async function _createFallBackPriceFeed(fallbackConfig) {
```
nicholaspai, line 493: Refactored this internal method a bit since we often want to create multiple pricefeeds from an array of configs

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -436,11 +484,22 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    const orderedPriceFeeds = await _createConstituentPriceFeed(fallbackConfig.orderedFeeds);
+    if (!orderedPriceFeeds) return null;
+    return new FallBackPriceFeed(orderedPriceFeeds);
+  }
+
+  async function _createConstituentPriceFeed(priceFeedConfigs) {
```
mrice32, line 499: Would plural make sense considering it takes in and returns an array?

```suggestion
  async function _createConstituentPriceFeeds(priceFeedConfigs) {
```

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -399,19 +399,41 @@ const defaultConfigs = {
...
-    hourlyLookback: 604800,
-    ohlcPeriod: 10 // CNYUSD only available at 10 minute granularity
+    type: "fallback",
+    orderedFeeds: [
+      {
+        type: "tradermade",
+        pair: "CNYUSD",
+        minTimeBetweenUpdates: 600,
+        minuteLookback: 7200,
+        hourlyLookback: 604800,
```
mrice32, line 409: Not necessarily suggesting we change it, but why are the default hourly lookbacks for these feeds 7 days? I could see making it 3 hours so we ensure that it at least covers the full 2 hours of the minutely lookback, but it seems like there's no reason the bots would need 7 days? Reducing this may also help with reliability since we're not asking for as much data.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -399,19 +399,41 @@ const defaultConfigs = {
...
+        ohlcPeriod: 10 // CNYUSD only available at 10 minute granularity
+      },
+      {
+        type: "forexdaily",
+        base: "CNY",
+        symbol: "USD",
+        lookback: 604800
```
mrice32, line 416: Similar question as above -- could make sense to shorten this to 1 or 2 days (if the feed is daily). We basically want to ensure we _at least_ cover the standard 2 hour period, but any more than that seems excessive. But maybe I'm missing something?

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
```diff
@@ -0,0 +1,99 @@
...
+
+    // If no pricefeeds return successfully, indicate failure.
+    return null;
+  }
+
+  async getHistoricalPrice(time, verbose = false) {
+    // If failure to fetch any constituent historical prices, then throw
+    // array of errors.
+    let errors = [];
+    for (let _priceFeed of this.priceFeeds) {
```
mrice32, line 38: nice and simple -- I like it.

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
```diff
@@ -0,0 +1,99 @@
...
+      }
+    }
+
+    throw errors;
+  }
+
+  // Note: This method will fail if one of the pricefeeds has not implemented `getHistoricalPricePeriods`, which
+  // is basically every price feed except for the CryptoWatchPriceFeed.
+  getHistoricalPricePeriods() {
+    throw new Error("getHistoricalPricePeriods Unimplemented for FallBackPriceFeed");
```
mrice32, line 53: +1 for implementing and erroring.

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // Gets the update time for the first pricefeed that successfully updated.
+  getLastUpdateTime() {
+    for (let _priceFeed of this.priceFeeds) {
+      const lastUpdateTime = _priceFeed.getLastUpdateTime();
+      if (!lastUpdateTime) continue;
+      return lastUpdateTime;
+    }
```
mrice32, line 62: Falling back on methods like this are a little weird. I'm not exactly sure how to handle it.

What about taking the latest of all successful return values? Not strongly opinionated, but this method seems to rely on the update having _never_ succeeded, which may not be true...? Really not sure :/

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
```diff
@@ -0,0 +1,99 @@
...
+    for (let _priceFeed of this.priceFeeds) {
+      try {
+        return await _priceFeed.update();
+      } catch (err) {
+        errors.push(err);
+        continue;
+      }
+    }
+
+    throw errors;
```
mrice32, line 93: What happens if the historical price query _succeeds_, but returns empty data on update. Then when you call the method to grab the historical prices, it finds no data and falls back. However, since we never called update on the fallback, it has no data either. Thoughts? Should we update _all_ fallbacks just in case? What do you think?

This is another reason that it might be nice to ultimately drop the update method entirely.

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
```diff
@@ -0,0 +1,271 @@
...
+    // 2. Send request.
+    const historyResponse = await this.networker.getJson(url);
+
+    // 3. Check responses.
+    if (
+      !historyResponse ||
+      !historyResponse.rates ||
+      Object.keys(historyResponse.rates).length === 0 ||
+      !historyResponse.rates[Object.keys(historyResponse.rates)[0]][this.symbol]
+    ) {
```
mrice32, line 191: Thoughts on doing this to check every element and simplify a bit?

```suggestion
    if (
      !historyResponse?.rates
      Object.keys(historyResponse.rates).length === 0 ||
      Object.values(historyResponse.rates).some(rate => !rate[this.symbol])
    ) {
```

mrice32 (commented):
Looks awesome! Mostly minor comments around the fallback price feed and other minutia. Thanks for putting this together so quickly!!

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: Well we want this to at least cover the period of time in which no data is being served over the weekend, so if the `minuteLookback` starts failing (i.e. returns NO data) 2 hours into the weekend (which begins at 22:00 UTC on Friday), then we would want `hourLookback` to cover a bit less than 2 days (but probably more over extended holiday periods). So I think a worst case setting here is 4 days, but I agree 7 days is too much. WDYT

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: That's reasonable, I'm going to set to 3 days for the reason mentioned before

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
nicholaspai: Shouldn't this return the update time of the pricefeed that is used for the pricefeed, i.e. the first pricefeed to return successfully?

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
nicholaspai: > What happens if the historical price query succeeds, but returns empty data on update
Hmmmm I think the issue you're pointing out is that  its possible right now that `update` succeeds but `historicalPrice` throws an error. So from that perspective it would make sense to `update` them all.

The downside is unneccessary updates and fetched data, which makes this pricefeed a bit less powerful.

And yes, the `update` is definitely overkill here, as it would be much better if `await getHistoricalPrice` was all we needed to call get a historical price.

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
nicholaspai: Ok so as mentioned [here](https://github.com/UMAprotocol/protocol/pull/2718#discussion_r595626631), I'm leaning towards updating ALL pricefeeds (and catching + ignoring any errors), and then using the latest update time as the update time here.

nicholaspai:
@mrice32 responded to your comments in [this commit](https://github.com/UMAprotocol/protocol/pull/2718/commits/99bb14f4da0d3ea71a7f5cfc263fda7b2296c8b1) (with comments)

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
mrice32: Yeah, I think that makes sense.

> Shouldn't this return the update time of the pricefeed that is used for the pricefeed, i.e. the first pricefeed to return successfully?
Ideally, yes, but I don't think it's true that this method errors when getCurrentPrice or getHistoricalPrice error, so you would effectively always get the first feed if that feed had ever succeeded.

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
mrice32: Agreed. It does make this a bit less powerful.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
mrice32: Seems reasonable to me!

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -760,6 +764,98 @@ contract("CreatePriceFeed.js", function(accounts) {
...
+        }
+      ]
+    };
+
+    const validFallbackFeed = await createPriceFeed(logger, web3, networker, getTime, config);
+
+    assert.equal(validFallbackFeed.priceFeeds[0].lookback, lookbackOverride);
+  });
+
+  it("Fallback feed cannot have 0 nested feeds to medianize", async function() {
```
mrice32, line 819:
```suggestion
  it("Fallback feed cannot have 0 nested feeds", async function() {
```

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
```diff
@@ -0,0 +1,270 @@
...
+    this.historicalPricePeriods = newHistoricalPricePeriods;
+    this.lastUpdateTime = currentTime;
+  }
+
+  _secondToDateTime(inputSecond) {
+    return moment.unix(inputSecond).format("YYYY-MM-DD");
+  }
+  _dateTimeToSecond(inputDateTime) {
+    return moment(inputDateTime, "YYYY-MM-DD").unix();
+  }
```
mrice32, line 228: I'm not sure how precise we want to be around these seconds -> day conversions, but the ECB data is [published at 16:00 CET](https://www.ecb.europa.eu/stats/policy_and_exchange_rates/euro_reference_exchange_rates/html/index.en.html) (computed at 14:15 CET, which might also be a good choice for stamping it), which is 3:00PM UTC and 11:AM EST. Using the terminology from this price feed, I would think the open time for the date string "2021-03-12" would be 2021-03-11 at 16:00 CET and the close time would be 2021-03-12 at 16:00 CET (or 14:15 if you prefer the compute time).

Thoughts?

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
nicholaspai: Yeah. Having the notion of "meta pricefeeds" such as Fallback and Medianizer makes the `update`-`getPrice` pattern inefficient, whereas a more robust `async getPrice` method would be better for the meta pricefeeds. This way the meta pricefeeds don't have to blindly update all constituent price feeds

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
nicholaspai: Great point, check out [this commit](https://github.com/UMAprotocol/protocol/pull/2718/commits/c7bf6695d06d7add6ff41fd3648cefa23e41831a)

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
mrice32: This is really really elegant. +1

PR #2727, opened 2021-03-17
feat(proposal-scripts): add gas price estimation to reusable proposal scripts

mrice32:
**Motivation**

Proposal scripts that are used often should use embedded gas price estimation.

**Summary**

Added the gas estimator class to these scripts to estimate the gas price.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/scripts/collateral-umip/1_Propose.js
+++ b/packages/core/scripts/collateral-umip/1_Propose.js
@@ -110,7 +110,6 @@ async function runExport() {
   await gasEstimator.update();
   const txn = await governor.propose(transactionList, {
     from: proposerWallet,
-    gas: 2000000,
     gasPrice: gasEstimator.getCurrentFastPrice()
   });
   console.log("Transaction: ", txn?.tx);
--- a/packages/core/scripts/collateral-umip/1_Propose.js
+++ b/packages/core/scripts/collateral-umip/1_Propose.js
@@ -14,11 +14,13 @@ const ERC20 = artifacts.require("ERC20");
 const Voting = artifacts.require("Voting");
 
 const { interfaceName } = require("@uma/common");
+const { GasEstimator } = require("@uma/financial-templates-lib");
 
 const { parseUnits } = require("@ethersproject/units");
 const { getDecimals } = require("./utils");
 
 const _ = require("lodash");
+const winston = require("winston");
 
 const argv = require("minimist")(process.argv.slice(), { string: ["collateral", "fee", "decimals"] });
 
@@ -27,6 +29,14 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 async function runExport() {
   console.log("Running Upgrade🔥");
   console.log("Connected to network id", await web3.eth.net.getId());
+
+  const gasEstimator = new GasEstimator(
+    winston.createLogger({
+      silent: true
+    }),
+    60, // Time between updates.
+    100 // Default gas price.
+  );
 
   if (!argv.collateral || !argv.fee) {
     throw new Error("Must provide --fee and --collateral");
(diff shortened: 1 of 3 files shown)


packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -28,6 +30,14 @@ async function runExport() {
   console.log("Running Upgrade🔥");
   console.log("Connected to network id", await web3.eth.net.getId());
 
+  const gasEstimator = new GasEstimator(
+    winston.createLogger({
+      silent: true
+    }),
+    60, // Time between updates.
+    100 // Default gas price.
```
nicholaspai, line 38: +1, probably time to change the default GasEstimator gas to 100

packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -97,7 +107,12 @@ async function runExport() {
   console.log(`Sending to governor @ ${governor.address}`);
 
   // Send the proposal
-  const txn = await governor.propose(transactionList, { from: proposerWallet, gas: 2000000 });
+  await gasEstimator.update();
+  const txn = await governor.propose(transactionList, {
+    from: proposerWallet,
+    gas: 2000000,
```
nicholaspai, line 113: Do you need the `gas` param? I think the `gas` should be auto filled

packages/core/scripts/creator-umip/1_Propose.js
```diff
@@ -34,7 +46,7 @@ async function runExport() {
         data: addCreatorToRegistryTx
       }
     ],
-    { from: proposerWallet, gas: 2000000 }
+    { from: proposerWallet, gas: 2000000, gasPrice: gasEstimator.getCurrentFastPrice() }
```
nicholaspai, line 49: Same question as before: do you need the `gas`?

packages/core/scripts/identifier-umip/1_Propose.js
```diff
@@ -9,14 +9,26 @@ const { getTruffleContract } = require("../../index");
...
+
 const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
   console.log("Connected to network id", await web3.eth.net.getId());
 
+  const gasEstimator = new GasEstimator(
+    winston.createLogger({
+      silent: true
```
nicholaspai, line 26: What does `silent` do? Why not just pass a `dummyLogger` which just prints to console:

```
dummyLogger = winston.createLogger({
            level: "info",
            transports: [new winston.transports.Console()]
          });
```

nicholaspai (approved):
THanks for finally paying this technical debt back in adding the GasEstimator to our scripts!

packages/core/scripts/identifier-umip/1_Propose.js
mrice32: It makes it print nothing. The bot-style logs that the gas estimator produces don't jive super well with the logs these scripts produce (they are super verbose and bulky). I can put the console logger in, but it just seems like it would distract from the other logs. Thoughts?

packages/core/scripts/creator-umip/1_Propose.js
mrice32: I think we originally did this because ganache-fork blows up sometimes when asked to estimate gas, so I just left it as it was.

packages/core/scripts/collateral-umip/1_Propose.js
mrice32: https://github.com/UMAprotocol/protocol/pull/2727#discussion_r596515482

packages/core/scripts/identifier-umip/1_Propose.js
```diff
@@ -45,7 +57,8 @@ async function runExport() {
     };
   });
 
-  await governor.propose(transactions, { from: proposerWallet });
+  await gasEstimator.update();
+  await governor.propose(transactions, { from: proposerWallet, gasPrice: gasEstimator.getCurrentFastPrice() });
```
pemulis, line 61: Unrelated, but should we have consistency in how transaction(s) are passed to the proposal method? We have three slightly different ways of doing this: `transactions` (identifier), `transactionList` (collateral), and a hardcoded array with a single element (creator).

pemulis (approved):
LGTM! Should make proposals easier.

packages/core/scripts/identifier-umip/1_Propose.js
mrice32: Not super opinionated. The creator one only allows a single transaction, so it is a little different from the others.

I think there's more stuff we could factor out and synchronize across these scripts, so I might suggest doing a refactor of all these scripts in a follow up.

PR #2728, opened 2021-03-18
feat(pricefeed): Add QuandlPriceFeed and use as backup for commodity identifiers

nicholaspai:
**Motivation**

Quandl offers a free API for certain commodity time series data on the daily granularity.

**Summary**

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -591,11 +591,23 @@ const defaultConfigs = {
...
+        type: "tradermade",
+        pair: "XAUUSD",
+        minuteLookback: 7200,
+        hourlyLookback: 259200,
+        minTimeBetweenUpdates: 60
+      },
+      {
+        type: "quandl",
+        // https://www.quandl.com/data/CHRIS/CME_MGC1-E-micro-Gold-Futures-Continuous-Contract-1-MGC1-Front-Month
+        datasetCode: "CHRIS",
```
nicholaspai, line 606: This is the "e-micro gold futures", but there are a few other "gold" related datasets. This one seemed like the best free option but curious what others think

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
```diff
@@ -0,0 +1,248 @@
...
+    //   }
+    // }
+    const newHistoricalPricePeriods = historyResponse.dataset_data.data
+      .map(dailyData => ({
+        // Output data should be a list of objects with only the open and close times and prices.
+        // Note: Data is formatted as [Date, Open, High, Low, Last, Change, Settle, Volume, Previous Day Open Interest]
+        openTime: this._dateTimeToSecond(dailyData[0]) - 24 * 3600,
+        closeTime: this._dateTimeToSecond(dailyData[0]),
+        openPrice: this.convertPriceFeedDecimals(dailyData[1]),
+        closePrice: this.convertPriceFeedDecimals(dailyData[4])
```
daywiss, line 225: probably not a huge deal, but it might make it easier (maybe) to use this data if its in the form 
```
[
  [time,price]
]
```
Its slightly more portable (easier to implement getHistoricalPricePeriods) and I dont think theres anywhere currently that requires this olhc shape.

You could do something like this to capture both open and close times, though your lookup function would need to change:

```js
.map(({openTime,openPrice,...etc})=>{
  return [
    [openTime,openPrice],
    [closeTime,closePrice],
  ]
}).flat() // should flatten this structure out to an array of [time,prices]
```

wdyt?

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
```diff
@@ -0,0 +1,248 @@
...
+    //      ...more data for different days
+    //     ],
+    //     ...other data we don't care about
+    //   }
+    // }
+    const newHistoricalPricePeriods = historyResponse.dataset_data.data
+      .map(dailyData => ({
+        // Output data should be a list of objects with only the open and close times and prices.
+        // Note: Data is formatted as [Date, Open, High, Low, Last, Change, Settle, Volume, Previous Day Open Interest]
+        openTime: this._dateTimeToSecond(dailyData[0]) - 24 * 3600,
```
daywiss, line 222: i dont think this is right, since you are converting this with moment, it would be giving you the start of the day.  you would have to add this to close time, you probably want something like 

`moment(time,'YYYY-MM-DD').endOf('day').unix()`

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
```diff
@@ -0,0 +1,248 @@
...
+  async getHistoricalPrice(time, verbose = false) {
+    if (this.lastUpdateTime === undefined) {
+      throw new Error(`${this.uuid}: undefined lastUpdateTime`);
+    }
+
+    // Set first price period in `historicalPricePeriods` to first non-null price.
+    let firstPricePeriod;
+    for (let p in this.historicalPricePeriods) {
+      if (this.historicalPricePeriods[p] && this.historicalPricePeriods[p].openTime) {
+        firstPricePeriod = this.historicalPricePeriods[p];
```
daywiss, line 72: is this just finding the first non empty instance or something?

daywiss (commented):
looks good for the most part just want to double check that timestamp

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
nicholaspai: Yup, copied from the CryptoWatch code. It finds the first price period (assuming they are sorted oldest first), in order to check that the request timestamp >= firstPricePeriod.openTime, otherwise the timestamp is before all data and we can exit early. Also if there is no `firstPricePeriod` after this loop, then we can also exit early because there is no data.

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
nicholaspai: Aha thats a really good catch. Also need to fix this in the ForexDailyPriceFeed as well.

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.js
nicholaspai: So I do think we should standardize how price periods are stored in all price feeds to `[time,price]`. For now I'm going to hold off since `getHistoricalPricePeriods = () => this.historicalPricePeriods.map(ohlc => [ohlc.closeTime, ohlc.closePrice])` is easy enough to implement right now. But once we re-evaluate the `update()` and `getPrice`() architecture, we're going to change a lot of the internal data structures and move towards a standard interface. Good reminder

daywiss (approved):

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
mrice32: This is a futures contract that is expected to be highly correlated with the gold price, so I think it's fine to use, but is it scaled such that it lines up with XAUUSD?

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
```diff
@@ -230,13 +233,21 @@ class ForexDailyPriceFeed extends PriceFeedInterface {
...
+        .tz(inputDateTime, "YYYY-MM-DD", "Europe/Berlin")
+        .endOf("day")
+        .add(16, "hours")
+        .unix();
+    } else {
+      return moment
+        .tz(inputDateTime, "YYYY-MM-DD", "Europe/Berlin")
+        .add(16, "hours")
+        .unix();
+    }
```
mrice32, line 250: What is this change?

mrice32 (approved):
Looks great! Only minor comments

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: It passes the eyeball test that's all  I can say.

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
nicholaspai: So before I added the `.endOf("day")`, the conversion from datetime string to unix was using the BEGINNING of the day by default, but sometimes I actually wanted the end of the day. For example, I was setting closeTime = beginning of the day (unintnetionally), and subtrackign 24 hours to get the openTime, which was actually setting the time to 24 hours behind what it should be.

packages/financial-templates-lib/src/price-feed/ForexDailyPriceFeed.js
```diff
@@ -205,8 +205,11 @@ class ForexDailyPriceFeed extends PriceFeedInterface {
     const newHistoricalPricePeriods = Object.keys(historyResponse.rates)
       .map(dateString => ({
         // Output data should be a list of objects with only the open and close times and prices.
-        openTime: this._dateTimeToSecond(dateString) - 24 * 3600,
-        closeTime: this._dateTimeToSecond(dateString),
```
nicholaspai, line 209: @mrice32 this was the error before: `openTime` should be beginning of the day, not beginning of the day prior. This is because `closeTime` was getting set to the beginning of the desired day.

PR #2742, opened 2021-03-19
feat(liquidation-reporter): Add excel spreadsheet generator that reports on collateral drawdown prices

chrismaree:
**Motivation**

issue #2743

As part of risk management over all collateral currencies, we require a report that shows the dollar amount of liquidation in each collateral currency at various draw down periods

This should include:

Overall TVL

- TVL in each collateral currency
- The number of liquidations available If the price of the collateral dropped by 10%/20%/30%/40% 

This enables us to see how much capital we need to perform liquidations in the event of a big dip in the market.

**Summary**

This PR builds on the `uma-tvl-calculator` that is used in KPI options to fetch info about all UMA contracts within the ecosystem. It also uses the graph to fetch position information about all sponsors in all contracts. Lastly, this PR leverages our existing price feed framework to get identifier prices.

**Details**

Drawdown prices are basically computed by finding the current price of each given identifier and then computing how much collateral would be liquidated for each 10% drop in that identifier price. See the implementation and comments for more info on this.

**Example output**
This script will send slack messages that look like this:
image

**The spreadsheet looks like this**
image

you can download a sample of the spreadsheet [here](https://file.io/jsY2629xoSZV)

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/merkle-distributor/kpi-options-helpers/calculate-uma-tvl.ts
```diff
@@ -24,7 +24,7 @@ export async function calculateCurrentTvl() {
 export async function getAllFinancialContractsData() {
```
chrismaree, line 24: modified this helper to include the collateral symbol. should not affect any other usage of this helper.

packages/merkle-distributor/scripts/example.json
```diff
@@ -1,8 +1,8 @@
 {
```
chrismaree, line 1: this was wrong. updated to match the true test case. unrelated to the main PR.

packages/reporters/liquidation-reporter/emp-position-helper.ts
```diff
@@ -0,0 +1,32 @@
+const fetch = require("node-fetch");
```
chrismaree, line 1: this graph query returns all positions within all EMPs. pretty cool stuff.

packages/reporters/scripts/SendSlackReport.sh
```diff
@@ -7,34 +7,31 @@ if [ -z "${BOT_IDENTIFIER}" ] || [ -z "${SLACK_TOKEN}" ] || [ -z "${SLACK_CHANNE
     exit 1
```
chrismaree, line 7: this is a re-work of the original daily reporter script :)

packages/reporters/liquidation-reporter/emp-position-helper.ts
```diff
@@ -0,0 +1,32 @@
...
+
+export async function getAllEmpsPositions() {
+  const subgraphUrl = "https://api.thegraph.com/subgraphs/name/umaprotocol/uma";
+  const query = `
+    query activePositions {
+      financialContracts {
+        id
+        positions(first: 1000, where: { collateral_gt: 0 }) {
```
nicholaspai, line 9: is 1000 too small? Why not set to something insanely high like 1000000

packages/reporters/liquidation-reporter/excel-writer.ts
```diff
@@ -0,0 +1,116 @@
...
+      .cell(2, 7)
+      .string(`Financial Contracts that use ${collateralData.collateralSymbol}`)
+      .style({ font: { size: 14, bold: true } });
+    sheet
+      .cell(3, 7)
+      .string("Contract Address")
+      .style(boldStyle);
+    sheet.column(7).setWidth("Contract Address".length);
+    sheet
+      .cell(3, 8)
```
nicholaspai, line 49: THese hard coded coordinates (3,8) seem like they will be nightmare to maintain and update if we want to change the format of the sheet. Thoughts on declaring these as constants and then just adding offsets?

For example:
```
COL_COLLATERAL_SYMBOL = 1
COL_COLLATERAL_TVL = 2
```

packages/reporters/liquidation-reporter/excel-writer.ts
```diff
@@ -0,0 +1,116 @@
...
+    sheet.cell(3, 2).number(Number(Number(collateralData.collateralValueInUsd).toFixed(2)));
+    sheet
+      .cell(4, 1)
+      .string("Collateral Address")
+      .style(boldStyle);
+    sheet.cell(4, 2).string(collateralAddress);
+
+    // Financial contracts list
+    sheet
+      .cell(2, 7)
```
nicholaspai, line 40: Why does this start at row(?) 7?

packages/reporters/liquidation-reporter/index.ts
```diff
@@ -0,0 +1,17 @@
...
+const { createExcelSheetFromLiquidationDrawDownData } = require("./excel-writer");
+
+async function fetchDataAndWriteToExcelFile() {
+  const ecosystemData = await fetchUmaEcosystemData();
+  createExcelSheetFromLiquidationDrawDownData(ecosystemData);
+}
+
+fetchDataAndWriteToExcelFile()
+  .then(() => {
+    setTimeout(function() {
```
nicholaspai, line 11: Why is this timeout neccessary?

packages/reporters/liquidation-reporter/liquidation-reporter.ts
```diff
@@ -0,0 +1,269 @@
...
+        // considers either the collateral value dropping OR the synthetic value increasing. We can set all three
+        // common API keys. If the feed does not need one of these it just ignores it.
+        const samplePriceFeed = await createReferencePriceFeedForFinancialContract(
+          logger,
+          web3,
+          networker,
+          getTime,
+          financialContractAddress,
+          {
+            cryptowatchApiKey: process.env.CRYPTO_WATCH_API_KEY,
```
nicholaspai, line 153: do you not need to import `require('dotenv').config()` to use this?

packages/reporters/liquidation-reporter/emp-position-helper.ts
chrismaree: what financial contract has that many positions? could change it if need be but right now we dont have even half that.

packages/reporters/liquidation-reporter/liquidation-reporter.ts
```diff
@@ -0,0 +1,269 @@
...
+  let index = -1;
+  array.forEach((object, i) => {
+    if (object[objectKey] == objectValue) index = i;
+  });
+
+  if (index > -1) array.splice(index, 1);
+  return array;
+}
+
+function computeCollatererlizationRatio(collateral: any, debt: any, tokenPrice: any) {
```
nicholaspai, line 264: typo "collateralization"

packages/reporters/liquidation-reporter/excel-writer.ts
chrismaree: ye, I agree. it's not ideal. it's bound very specifically to the design of this sheet in particular. not really any better way of making it more generalised giving it's building up on a per cell basis.

nicholaspai (commented):
Wow this is a really good feature, I have some concerns about how many hardcoded values you use in the excel-writer script however and think we can make it much more pleasant for future maintainers

packages/reporters/liquidation-reporter/excel-writer.ts
chrismaree: the content is off on the right of the sheet.

packages/reporters/liquidation-reporter/index.ts
chrismaree: without it the script would sometimes not write correctly. I'm not actually sure why.

packages/reporters/liquidation-reporter/liquidation-reporter.ts
chrismaree: dont think so? it seemed to work fine. I think that's if you want to read in a `.env` file. if you want to read directly from the `export ENV=xxx` then it's ok. this will be the case in production.

daywiss (approved):
always impressed how fast you complete this stuff. i assume it works since we saw the xls in slack. lgtm

packages/reporters/liquidation-reporter/index.ts
```diff
@@ -0,0 +1,17 @@
...
+      process.exit(0);
```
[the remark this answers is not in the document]
chrismaree: ye, I think that was the issue. it could also be a ts compatibility thing. really not sure why I had to do it like this.

packages/reporters/liquidation-reporter/emp-position-helper.ts
```diff
@@ -0,0 +1,32 @@
...
+  `;
+  const response = await fetch(subgraphUrl, {
+    method: "POST",
+    headers: {
+      Accept: "application/json",
+      "Content-Type": "application/json"
+    },
+    body: JSON.stringify({ query })
+  });
+  return (await response.json()).data.financialContracts;
```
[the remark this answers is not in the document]
chrismaree: I think the script will just error out downstream

packages/reporters/liquidation-reporter/liquidation-reporter.ts
```diff
@@ -0,0 +1,269 @@
...
+  array: Array<any>,
+  objectKey: string | number,
+  objectValue: string | number | undefined
+) {
+  let index = -1;
+  array.forEach((object, i) => {
+    if (object[objectKey] == objectValue) index = i;
+  });
+
+  if (index > -1) array.splice(index, 1);
```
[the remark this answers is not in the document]
chrismaree: ye, I dont think there will be more than one duplicate within the array.

packages/reporters/liquidation-reporter/liquidation-reporter.ts
```diff
@@ -0,0 +1,269 @@
...
+        await samplePriceFeed.update();
+
+        const startingPrice = samplePriceFeed.getCurrentPrice();
+        const priceFeedDecimals = samplePriceFeed.getPriceFeedDecimals();
+
+        // Some feeds are inverted. This can either be at the top level within the object or within the array of priceFeeds
+        // if the feed consists of multiple feeds within it.
+        let invertedPrice = false;
+        if (samplePriceFeed.invertPrice) invertedPrice = true;
+        if (samplePriceFeed.priceFeeds && samplePriceFeed.priceFeeds[0].invertPrice) invertedPrice = true;
```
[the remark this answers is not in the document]
chrismaree: hhhmm. well this logic relies on the default price feeds that we've got and they contain the inverted behaviour. we could override all price feeds to be `invertedPrice=false` but that kind of changes the definition of the identifier and I think that would be even more difficult to reason about 🤔

PR #2744, opened 2021-03-19
improve(merkle-distributor): Assorted fixes + natspec comments

nicholaspai:
**Motivation**

Addressing several recommendations to improve the quality of the code base:
- The `lastCreatedIndex` variable name and comments suggest it indicates an existing window. In fact, it is the index that will be assigned to the next window. Consider updating the name and comments accordingly.
- The `WithdrawRewards` event does not specify the reward currency. Consider including it as well.
- The `setWindow` function performs the token transfer before emitting the event. If the reward token follows the ERC777 specification, it is possible that the caller has a `tokensToSend` hook that re-enters the `setWindow` function, which would cause the `CreatedWindow` events to be emitted in the wrong order. This may fall outside the threat model, since the function can only be called by the owner, who is assumed to behave honestly. Nevertheless, as a matter of good practice, consider emitting the event before performing the token transfer or introducing a reentrancy guard.
- The code does not use Ethereum Natural Specification Format (NatSpec) comments. Consider following this specification on everything that is part of the contract's public API.
- The `setWindow` function implicitly assumes the caller has granted an appropriate allowance to the contract. Consider stating this explicitly in the function comments.
- The `setWindow` function comment has an extra closing bracket.
- The `withdrawRewards` function comment is missing the space in the phrase "in case".

**Summary**

Addressed all comments, usually in separate commits.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
+++ b/packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
@@ -206,9 +206,9 @@ contract MerkleDistributor is Ownable {
         window.rewardToken = IERC20(rewardToken);
         window.ipfsHash = ipfsHash;
 
-        window.rewardToken.safeTransferFrom(msg.sender, address(this), rewardsDeposited);
-
         emit CreatedWindow(windowIndex, rewardsDeposited, rewardToken, msg.sender);
+
+        window.rewardToken.safeTransferFrom(msg.sender, address(this), rewardsDeposited);
     }
 
     // Verify claim is valid and mark it as completed in this contract.
--- a/packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
+++ b/packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
@@ -100,7 +100,7 @@ contract MerkleDistributor is Ownable {
      *      subsequently transfer in rewards or the following situation can occur).
      *      Example race situation:
      *          - Window 1 Tree: Owner sets `rewardsToDeposit=100` and insert proofs that give claimant A 50 tokens and
-     *            claimant B 51 tokens. The owner has made an error by not setting the `rewardsToDeposit` correctly to 101).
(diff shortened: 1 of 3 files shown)


packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -71,65 +80,87 @@ contract MerkleDistributor is Ownable {
...
         address owner
     );
-    event WithdrawRewards(address indexed owner, uint256 amount);
+    event WithdrawRewards(address indexed owner, uint256 amount, address indexed currency);
     event DeleteWindow(uint256 indexed windowIndex, address owner);
 
     /****************************
      *
-     * Admin functions
+     * ADMIN FUNCTIONS
```
chrismaree, line 88: :)

chrismaree: do you also want to remove the extra `*` above and below this to be more consistant with our other styling guides in other contracts? so it would look like this:
```
     *              ADMIN FUNCTIONS                 *

```

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -128,8 +128,8 @@ contract("MerkleDistributor.js", function(accounts) {
       assert.equal(windowState.rewardToken, rewardToken.address);
       assert.equal(windowState.ipfsHash, sampleIpfsHash);
 
-      // Check that latest created index has incremented.
-      assert.equal((await merkleDistributor.lastCreatedIndex()).toString(), (windowIndex + 1).toString());
+      // Check that next created index has incremented.
```
chrismaree, line 131: please also update this script:
https://github.com/UMAprotocol/protocol/blob/69b87e98283b6970375358010d906f55fa29dd9b/packages/merkle-distributor/scripts/2_PublishClaimsForWindow.ts#L81

chrismaree (approved):
Pretty much only styling and formatting changes. Thanks for making them so quickly! 

Please update the script I linked to as well or this change will break the script.

PR #2750, opened 2021-03-22
feat(liquidator-contracts): add atomic-liquidator contracts enabling Swap, mint, liquidate behaviour

chrismaree:
**Motivation**

#2751

**Summary**

As part of the effort to improve the capital efficiency of UMA liquidator bots, we require the ability for liquidators to hold hold reserve currency for n number of different uma contracts. This PR adds the initial smart contracts and unit tests that enable this.

**Details**

The basic flow of execution looks like this:
1) The bot will call the DSProxy deployed for the bot. This DSProxy holds the reserve currency.
2) DSProxy executes the following within one transaction (facilitated by the new `ReserveCurrencyLiquidator.sol` contract added in this PR):
2.a) Calculate how much collateral needed to preform the liquidation. This assumes the mint will occur at the GCR to make it as capital efficient as posible. The caller of this method defines how much synthetics to liquidate.
2.b) Calculate how much collateral is needed for the final fee.
2.c) Swap reserve currency to get required collateral for the mint + final fee.
2.d) Use the collateral currency to mint the synthetics at the GCR.
2.e) Liquidate the sponsor.
3) After some time, the liquidation is withdrawn by the bot.
4) After some time, a privileged EOA can unwind the position created in 2.d. This is manual for now but could be automated later on if desired.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Partially solves #2751


packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+        FixedPoint.Unsigned calldata minCollateralPerTokenLiquidated,
+        FixedPoint.Unsigned calldata maxCollateralPerTokenLiquidated,
+        FixedPoint.Unsigned calldata maxTokensToLiquidate,
+        uint256 deadline
+    ) public {
+        IUniswapV2Router01 router = IUniswapV2Router01(uniswapRouter);
+        FinancialContractInterface fc = FinancialContractInterface(financialContract);
+
+        // 1. Calculate how much collateral needed to mint maxTokensToLiquidate.
+        FixedPoint.Unsigned memory gcr = fc.pfc().div(fc.totalTokensOutstanding());
```
nicholaspai, line 52: This makes me think that we should make gcr() a public method on the financial contracts

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+        // 2. Calculate how much collateral is needed for the final fee.
+        FinderInterface finder = FinderInterface(fc.finder());
+        StoreInterface store = StoreInterface(finder.getImplementationAddress("Store"));
+        FixedPoint.Unsigned memory totalCollateralNeeded =
+            collateralToMintAtGcr.add(store.computeFinalFee(fc.collateralCurrency()));
+
+        // 3. Swap reserve currency to get required collateral for the mint + final fee.
+        address[] memory path = new address[](2);
+        path[0] = reserveCurrency;
+        path[1] = fc.collateralCurrency();
```
nicholaspai, line 64: Is the path valid if path[0] == path[1] (i.e. the same currency?)

packages/liquidator/src/liquidationStrategy.js
```diff
@@ -75,11 +75,11 @@ module.exports = (
...
     return [
-      sponsor,
-      { rawValue: liquidationMinPrice.toString() },
-      { rawValue: maxCollateralPerToken.toString() },
-      { rawValue: tokensToLiquidate.toString() },
-      parseInt(currentBlockTime) + liquidationDeadline
+      sponsor, // sponsor
+      { rawValue: liquidationMinPrice.toString() }, // minCollateralPerToken
+      { rawValue: maxCollateralPerToken.toString() }, // maxCollateralPerToken
+      { rawValue: tokensToLiquidate.toString() }, // maxTokensToLiquidate
```
nicholaspai, line 81: +1 comments

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+
+        TransferHelper.safeApprove(reserveCurrency, address(router), maxReserverTokenSpent.rawValue);
+        router.swapTokensForExactTokens(
```
nicholaspai, line 67: What about the edge case where the Uniswap exchange rate jumps between when the caller submitted the transaction and when it is mined? Should we make the user set a `maxExchangeRate`, similar to how we make the `createLiquidation()` caller set a `maxCollateralPeTokenLiquidated` ratio to protect themselves from being front-ran?

nicholaspai (approved):
Pretty straightforward and well implemented. Main concern is protecting the caller from being front-ran, which we get for free in the `createLiquidation()` transaction because of how Liquidatable contracts are implemented, but might need to explicitly protect against for the UniswapBroker

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: Ye I'm pretty sure this is not an issue as the liquidator can provide the `maxReserverTokenSpent` to limit how much they should spend in the reserve currency. using this they can limit any kind of slippage they would have received .

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: This is actually a great point. If the reserve currency is the collateral currency then we should not do any trade. It is posible that the uniswap router will gracefully deal with this but it might be better to have a bit of branching logic within the contract. I will add some unit tests to validate this and update the contracts accordingly if this needs to be changed.

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: 100%. equally, it would be great for the contract to expose methods on computing fees so I dont need both a finder and a store to get this number.

mrice32: Wow, great point. Yeah, we really should expose the GCR in some way.

The return values would ideally help you in computing fees and capital required, but you can't really `.call` those until you _have_ the tokens required or the method will revert 🤔

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+        FixedPoint.Unsigned memory collateralToMintAtGcr = maxTokensToLiquidate.mul(gcr);
+
+        // 2. Calculate how much collateral is needed for the final fee.
+        FinderInterface finder = FinderInterface(fc.finder());
+        StoreInterface store = StoreInterface(finder.getImplementationAddress("Store"));
+        FixedPoint.Unsigned memory totalCollateralNeeded =
+            collateralToMintAtGcr.add(store.computeFinalFee(fc.collateralCurrency()));
+
+        // 3. Swap reserve currency to get required collateral for the mint + final fee.
+        address[] memory path = new address[](2);
```
mrice32, line 62: Does uniswap do any smart routing on-chain? Or does this assume a highly-liquid market between token A and token B, specifically.

Put differently, if there are three pools:
- A <-> C: $100MM in pool
- B <-> C: $100MM in pool
- A <-> B: $100k in pool

Will this go through the A <-> B pool without considering the additional slippage incurred?

This is totally fine/reasonable in v1 (our reserve currency will likely be liquid against almost anything in uniswap, etc etc), but I just wanted to make sure I was understanding this correctly.

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+     * @param maxCollateralPerTokenLiquidated abort the liquidation if the position's collateral per token exceeds this value.
+     * @param maxTokensToLiquidate max number of tokens to liquidate. For a full liquidation this is the full position debt.
+     * @param deadline abort the trade and liquidation if the transaction is mined after this timestamp.
+     **/
+    function swapMintLiquidate(
+        address uniswapRouter,
+        address financialContract,
+        address reserveCurrency,
+        address liquidatedSponsor,
+        FixedPoint.Unsigned calldata maxReserverTokenSpent,
```
mrice32, line 42:
```suggestion
        FixedPoint.Unsigned calldata maxReserveTokenSpent,
```

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -0,0 +1,136 @@
...
+        StoreInterface store = StoreInterface(finder.getImplementationAddress("Store"));
+        FixedPoint.Unsigned memory totalCollateralNeeded =
+            collateralToMintAtGcr.add(store.computeFinalFee(fc.collateralCurrency()));
+
+        // 3. Swap reserve currency to get required collateral for the mint + final fee.
+        address[] memory path = new address[](2);
+        path[0] = reserveCurrency;
+        path[1] = fc.collateralCurrency();
+
+        TransferHelper.safeApprove(reserveCurrency, address(router), maxReserverTokenSpent.rawValue);
```
mrice32, line 66: nit: tiny tiny gas optimization we might consider down the road: we might want to just set this to max approval so we don't have to re-approve each time.

Note: I think this would only be better if we're not sure how many tokens the contract will pull. If the contract will pull _exactly_ the approved tokens, it is probably better to set it on each run since we would benefit from storage write netting: 0 -> nonzero -> 0 === 2 sloads (not two sstores).

mrice32 (approved):
This is delightfully simple!! Very nice!

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: in this implementation, it assumes a liquid market between reserve and collateral. This might be a bad assumption but for now it's easiest to reason about. I think if we wanted to do better routing we would need to define the router off-chain and provide it to the contract. 

Thinking more about this, it makes sense to keep the reserve currency in ETH so that we always have a liquid pair (95% of liquidity & trades on uniswap happen in eth denominated pairs).

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: ye, that's a good point. the only downside is right now I actually sue this approval to limit the size of the trade. `maxReserverTokenSpent` bounds how many tokens can be spent!

PR #2752, opened 2021-03-22
fix(liquidator): should submit raw # of position tokens, not funding-rate adjusted #

nicholaspai:

**Motivation**

Currently, when the FinancialContractClient determines which positions are undercollateralized based on their position collateral and funding-rate adjusted token debt outstanding. Next, the client passes to the liquidator a list of positions to liquidate. However, the liquidator uses the funding-rate adjusted token debt to determine how many tokens to liquidate. This isn't a problem where the `cumulativeFundingRateMultiplier` (CFRM) is a positive number, because it causes the liquidator to overestimate the amount of tokens to liquidate, but it can be a problem if the CFRM is negative, potentially causing the liquidator to liquidate fewer tokens that it can and should.

**Summary**

I add the `rawTokens` parameter (the actual amount of synthetic token debt for a position) to the position object sent from the client to the liquidator bot, and use this # for the liquidator. This is easy to use because the liquidator bot doesn't perform any additional calculations on whether a position is correctly collateralized.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/FinancialContractClient.js
+++ b/packages/financial-templates-lib/src/clients/FinancialContractClient.js
@@ -213,6 +213,8 @@ class FinancialContractClient {
           sponsor: liquidation.sponsor,
           id: id.toString(),
           state: liquidation.state,
+          // Note: The `tokensOutstanding` entry in the Perpetual's on-chain Liquidation struct is adjusted
+          // for the funding rate multiplier at the time of liquidation.
           numTokens: liquidation.tokensOutstanding.toString(),
           liquidatedCollateral: liquidation.liquidatedCollateral.toString(),
           lockedCollateral: liquidation.lockedCollateral.toString(),
--- a/packages/financial-templates-lib/src/clients/FinancialContractClient.js
+++ b/packages/financial-templates-lib/src/clients/FinancialContractClient.js
@@ -247,15 +247,23 @@ class FinancialContractClient {
         sponsor: this.activeSponsors[index],
         withdrawalRequestPassTimestamp: position.withdrawalRequestPassTimestamp,
         withdrawalRequestAmount: position.withdrawalRequestAmount.toString(),
-        numTokens: this.toBN(position.tokensOutstanding.toString()) // Apply the current funding rate to the sponsor debt.
+        numTokens: this.toBN(position.tokensOutstanding.toString())
           .mul(this.latestCumulativeFundingRateMultiplier)
           .div(this.fixedPointAdjustment)
           .toString(),
-        rawTokens: position.tokensOutstanding.toString(),
-        amountCollateral: this.toBN(position.rawCollateral.toString()) // Apply the current outstanding fees to collateral.
+        // Applies the current funding rate to the sponsor's raw amount of tokens outstanding. `numTokens` is equal
+        // to `rawTokens * cumulativeFundingRateMultplier`, and this client uses `numTokens` to determine whether
+        // a position is liquidatable or not. If a position is liquidatable, then `rawTokens` amount of debt needs
+        // to be repaid by the liquidator to liquidate the position.
+        rawTokens: position.tokensOutstanding.toString(), // Raw tokens is the `tokensOutstanding` parameter in the
+        // on-chain position struct. It represents the actual amount of tokens outstanding for a sponsor and is the
+        // maximum amount of tokens needed to fully liquidate a position. The liquidator bot therefore uses this
+        // amount.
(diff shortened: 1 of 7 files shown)


packages/liquidator/src/liquidator.js
```diff
@@ -237,7 +237,9 @@ class Liquidator {
       at: "Liquidator",
       message: "Checking for under collateralized positions",
       liquidatorOverridePrice: liquidatorOverridePrice ? liquidatorOverridePrice.toString() : null,
-      latestCumulativeFundingRateMultiplier: this.financialContractClient.getLatestCumulativeFundingRateMultiplier(),
+      latestCumulativeFundingRateMultiplier: this.financialContractClient
```
nicholaspai, line 240: This `.toString()` is added becuse it is currently logging the multiplier as a hex string

pemulis (commented):
These changes make sense to me! The only problem I see is in the tests, which still expect `numTokens` in the assertions but get `rawTokens` instead.

packages/liquidator/src/liquidator.js
chrismaree: good catch

packages/financial-templates-lib/src/clients/FinancialContractClient.js
```diff
@@ -251,6 +251,7 @@ class FinancialContractClient {
           .mul(this.latestCumulativeFundingRateMultiplier)
           .div(this.fixedPointAdjustment)
           .toString(),
+        rawTokens: position.tokensOutstanding.toString(),
```
chrismaree, line 254: so the `numTokens`, considering the funding rate, is still used when the bot decided if a position is liquidatable, correct? then, if a position deemed liquidatable the `rawTokens` is liquidated. This way you still factor in the funding rate to the liquidation decision but use the `rawTokens` (the actual amount in the position) when doing the liquidation, right?

Also could you please add a comment here to explain this distinction between `numTokens` and `rawTokens`.

packages/financial-templates-lib/test/clients/FinancialContractClient.js
```diff
@@ -909,6 +924,7 @@ contract("FinancialContractClient.js", function(accounts) {
                     .mul(toBN(toWei("1.01")))
                     .div(toBN(toWei("1")))
                     .toString(), // the funding rate should be applied to the num of tokens
+                  rawTokens: convertSynthetic("100"),
```
chrismaree, line 927: this test is useful as the `numTokens` is adjusted by the funding rate.

chrismaree (approved):
Great catch @nicholaspai ! This would have been really hard to catch in production. War games save the day!

I left a comment that I'd like you to clarify on + some additional commenting I'd like to see.

packages/financial-templates-lib/src/clients/FinancialContractClient.js
nicholaspai: > so the numTokens, considering the funding rate, is still used when the bot decided if a position is liquidatable, correct?

Correct.

Added comment:
```
numTokens: this.toBN(position.tokensOutstanding.toString()) 
          .mul(this.latestCumulativeFundingRateMultiplier)
          .div(this.fixedPointAdjustment)
          .toString(),
        // Applies the current funding rate to the sponsor's raw amount of tokens outstanding. `numTokens` is equal
        // to `rawTokens * cumulativeFundingRateMultplier`, and this client uses `numTokens` to determine whether
        // a position is liquidatable or not. If a position is liquidatable, then `rawTokens` amount of debt needs
        // to be repaid by the liquidator to liquidate the position.
        rawTokens: position.tokensOutstanding.toString(), // Raw tokens is the `tokensOutstanding` parameter in the
        // on-chain position struct. It represents the actual amount of tokens outstanding for a sponsor and is the
        // maximum amount of tokens needed to fully liquidate a position. The liquidator bot therefore uses this
        // amount.
```

mrice32 (approved):
Nice catch. I still find the rawTokens and numTokens terminology a bit confusing -- it seems like `adjustedTokens` (rather than `numTokens`) and `numTokens` (rather than `rawTokens`) would make sense since `rawTokens` is the _true_ number of outstanding tokens and `numTokens` is really an adjustment/abstraction to make the tokens match the index. However, that's unrelated to this PR.

nicholaspai:
> adjustedTokens

Good point, I changed all the internal variable from `numTokens` to `adjustedTokens` and `rawTokens` to `numTokens` to make this distinction clearer

PR #2756, opened 2021-03-23
feat(liquidator): remove one inch intergration

chrismaree:
**Motivation**

As part of adding single reserve currency support, we need to deprecate some duplicate, redundant and unmaintained code. This PR removes all remnants of one inch from the repo.


mrice32 (approved):

daywiss (approved):
nice

PR #2757, opened 2021-03-23
feat(funding-rate-proposer): Call applyFundingRate if it hasn't been called recently

nicholaspai:
**Motivation**

The `FinancialContractClient` is in charge of determining position collateralization ratios. It performs the calculation: `(collateral value / token value * cumulativeFundingRateMultiplier (CFRM))`
- The CFRM is read on-chain directly from the Perpetual contract: `(await perpetual.fundingRate()).cumulativeFundingRateMultiplier()`
- The CFRM is only updated on chain when an` applyFundingRate()` method is called (this is called as a modifier on every single contract transaction).
- Therefore, it is possible that the CFRM doesn’t update for some time, even though it will update correctly on the next contract interaction. Therefore, it is possible that the liquidator bot can be delayed in liquidating until the CFRM is updated.
- This is potentially exploitable by a malicious sponsor who knows that the pending CFRM will cause their position to go undercollateralized, but the CFRM just doesn’t get updated because this contract doesn’t have a lot of users for example.

This adds a new feature to the funding rate proposer bot whereby it sets an upper bound on the amount of time that a pending proposal can sit unpublished for before the bot will call `applyFundingRate()`. Setting the update threshold time is a subjective task, and I'm also currently investigating simulating calling `applyFundingRate().call()` and `cumulativeFundingRateMultiplier()` atomically to get the to-be-published funding rate without having to broadcast a transaction.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/funding-rate-proposer/index.js
+++ b/packages/funding-rate-proposer/index.js
@@ -84,6 +84,7 @@ async function run({
       await retry(
         async () => {
           await fundingRateProposer.update();
+          await fundingRateProposer.applyFundingRates();
           await fundingRateProposer.updateFundingRates(isTest);
           return;
         },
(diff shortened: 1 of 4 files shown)


nicholaspai:
I don't think this PR is neccessary anymore because of #2762

PR #2762, opened 2021-03-23
feat(multicall): Using Multicall contract to simulate contract state post-state modifying transactions

nicholaspai:

**Motivation**

We want to simulate how contract state would change after sending state-modifying transactions (i.e. `applyFundingRate()`) without having to actually send the transaction. We therefore attempt to integrate the Multicall contract to simulate multiple on-chain transactions.

You should notice that in the bot tests I removed a several `applyFundingRate()` calls that are no longer needed because the bot can now accurately predict what the effective funding rates will be.

**Details**

Added a helper module in `financial-templates-lib` that passes lists of transaction ABI data to a deployed Multicall contract and returns the `.call()` result. In the future we can add a helper module that actually executes the Multicall.

Unfortunately, any bot that uses the `FinancialContractClient` now needs to pass in a Multicall address into the client in order to use it now.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/test/clients/FinancialContractClient.js
```diff
@@ -824,23 +831,26 @@ contract("FinancialContractClient.js", function(accounts) {
...
-            assert.equal(client.getLatestCumulativeFundingRateMultiplier(), toWei("1.05"));
+            // // After updating the client the funding rate is applied.
+            // await client.update();
+            // assert.equal(client.getLatestCumulativeFundingRateMultiplier(), toWei("1.05"));
 
             // after advancing time with the same funding rate the client value should not change
             await _setFundingRateAndAdvanceTime(toWei("0"));
-            await financialContract.applyFundingRate();
+            // await financialContract.applyFundingRate();
             assert.equal(client.getLatestCumulativeFundingRateMultiplier().toString(), toWei("1.05"));
```
nicholaspai, line 849: I've commented out the explicit `applyFundingRate()` calls here, and therefore the client should be able to detect the pending funding rate if the multicontract simulation is working

packages/financial-templates-lib/src/clients/FinancialContractClient.js
```diff
@@ -159,9 +165,35 @@ class FinancialContractClient {
...
+        // that this will not take into account any pending fundingrates that will be
+        // published on the next contract interaction.
+        this.latestCumulativeFundingRateMultiplier = this.toBN(
+          (await this.financialContract.methods.fundingRate().call()).cumulativeMultiplier.rawValue
+        );
+      } else {
+        // Simulate calling `applyFundingRate()` on the perpetual before reading
+        // `fundingRate()`, in order to account for any unpublished, pending funding rates.
+        const applyFundingRateCall = {
+          target: this.financialContract.options.address,
```
chrismaree, line 179: is this not accessible directly from `.address` or `._address`? going via options is a bit different to what we normally do, no?

packages/financial-templates-lib/src/clients/FinancialContractClient.js
```diff
@@ -159,9 +165,35 @@ class FinancialContractClient {
...
+          callData: this.financialContract.methods.fundingRate().encodeABI()
+        };
+        const outputs = await aggregateTransactionsAndCall(this.multicallContractAddress, this.web3, [
+          applyFundingRateCall,
+          fundingRateCall
+        ]);
+        const fundingRateData = this.web3.eth.abi.decodeParameters(
+          contractMethodABIs["fundingRate"],
+          // The return data for `fundingRate()` is the second return argument.
+          outputs.returnData[1]
```
chrismaree, line 193: does the returned data not also offer key-values on the returned types? is it _just_ an array of elements? often I've seen these kinds of methods return both the array and a key-valued type. just mentioning as it will be way easier to read if it is not referencing an array index.

packages/financial-templates-lib/src/helpers/multicall.js
```diff
@@ -0,0 +1,75 @@
...
+      "transaction expected in form {target: address, callData: bytes}"
+    );
+  }
+  return await multicallContract.methods.aggregate(transactions).call();
+};
+
+// Dictionary of ABI's for contract methods that we can use to decode
+// to Javascript type.
+const contractMethodABIs = {
+  fundingRate: [
```
chrismaree, line 23: is there no way to extract this directly from the contract being called? why does it need to be hard coded like this? this will limit how extendable this helper is. could you not find this same ABI directly from the truffle artefact?

chrismaree (commented):
This is really awesome! good work on getting this set up. I left a few comments/nits/question but overall I think this is great!

packages/financial-templates-lib/src/clients/FinancialContractClient.js
nicholaspai: Changed to `this.financialContractAddress`

packages/financial-templates-lib/src/clients/FinancialContractClient.js
nicholaspai: Here's an example return output:

```
Result {
  '0': '96',
  '1': [
    '0x',
    '0x0000000000000000000000000000000000000000000000000000000000000000756e646566696e65645f46554e44494e475f4944454e544946494552000000000000000000000000000000000000000000000000000000000de0b6b3a764000000000000000000000000000000000000000000000000000000000003adb1765300000000000000000000000000000000000000000000000000000003adb176540000000000000000000000000000000000000000000000000000000000000000'
  ],
  blockNumber: '96',
  returnData: [
    '0x',
    '0x0000000000000000000000000000000000000000000000000000000000000000756e646566696e65645f46554e44494e475f4944454e544946494552000000000000000000000000000000000000000000000000000000000de0b6b3a764000000000000000000000000000000000000000000000000000000000003adb1765300000000000000000000000000000000000000000000000000000003adb176540000000000000000000000000000000000000000000000000000000000000000'
  ]
}
```

You'll notice that the returnData is just a simple array

packages/financial-templates-lib/src/helpers/multicall.js
nicholaspai: Good point, replaced with `this.financialContractAbi.find(method => method.name === "fundingRate").outputs`

packages/financial-templates-lib/src/helpers/multicall.js
```diff
@@ -0,0 +1,23 @@
+// Provides convenience methods for interacting with deployed Multicall contract on network.
+const { getAbi } = require("@uma/core");
+const multicallAddressMap = require("@makerdao/multicall/src/addresses.json");
+const assert = require("assert");
+
```
mrice32, line 5: We could have this method automatically decode the outputs:
```js
const { getAbiDecoder } = require("@uma/common");

let abiDecoder;

function _decodeOutput(callData, returnData, web3) {
  let abiDecoder = abiDecoder || getAbiDecoder(); // Only load once because it's expensive.
  const methodAbi = abiDecoder.getMethodIDs()[callData.slice(2,10)];
  return web3.eth.abi.decodeParameters(methodAbi.outputs, returnData);
}
```

Then we could call this in `aggregateTransactionsAndCall` to auto-decode.

Thoughts?

We could also do something like this in a follow-up.

mrice32 (commented):
Looks awesome. Two minor comments

packages/core/contracts/common/test/MulticallMock.sol
```diff
@@ -0,0 +1,24 @@
+pragma solidity >=0.5.0;
+pragma experimental ABIEncoderV2;
+
+/// @title Multicall - Aggregate results from multiple read-only function calls
+/// @author Michael Elliot <mike@makerdao.com>
+/// @author Joshua Levine <joshua@makerdao.com>
+/// @author Nick Johnson <arachnid@notdot.net>
+
+contract MulticallMock {
```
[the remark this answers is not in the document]
nicholaspai: I actually kind of like keeping the interface that the bot uses as barebones as possible so I'm going to keep for now.

packages/financial-templates-lib/src/helpers/multicall.js
nicholaspai: Very nice, it worked and I refactored

packages/financial-templates-lib/src/clients/FinancialContractClient.js
mrice32: @chrismaree this is how you access the address for web3 contracts I believe. Truffle contracts offer the .address I think.

packages/core/contracts/common/test/MulticallMock.sol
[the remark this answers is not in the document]
mrice32: Fair enough!

packages/financial-templates-lib/src/helpers/multicall.js
```diff
@@ -0,0 +1,40 @@
...
+      "transaction expected in form {target: address, callData: bytes}"
+    );
+  }
+
+  // Decode return data, which is an array of the same length as `transactions`:
+  const returnData = (await multicallContract.methods.aggregate(transactions).call()).returnData;
+  const decodedOutputs = [];
+  returnData.forEach((data, i) => {
+    decodedOutputs.push(_decodeOutput(transactions[i].callData, data, web3));
+  });
```
mrice32, line 33: nit: might be slightly simpler as a map

```suggestion
  return returnData.map((data, i) => _decodeOutput(transactions[i].callData, data, web3));
```

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/helpers/multicall.js
chrismaree: I really like this change. good implementation!

chrismaree (approved):
This is awesome. thanks for refactoring out the ABI stuff. I'm pretty happy with this and excited to see where we can apply this same pattern in other places!

packages/financial-templates-lib/src/clients/FinancialContractClient.js
chrismaree: ye, this is a web3 contract as we are within the bots, not in a truffle test. both `.options.address` and `._address` are bound to the same thing. Looks like we actually use both methods within the repo if you search for the strings :)

PR #2768, opened 2021-03-24
feat(financial-product-library): Create a KPI Options financial product library

smb2796:
**Motivation**

To create KPI options, we require a simple library to transform the collateral requirement and price of an EMP. 

Behavior:
- Pre-expiry the price is always set to 2.
- Post-expiry the price is not transformed
- The collateral requirement is always set to 1

The price identifier that will be used with this library is the `uTVL_KPI_UMA` price identifier, defined in [UMIP-65](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-65.md). `uTVL_KPI_UMA` sets a price ceiling of 2, so enforcement a pre-expiry collateralization requirement of 2 tokens will ensure that positions cannot become under collateralized by means of an increase in the token price.

**Summary**

`KpiOptionsFinancialProductLibrary.sol` added with associated unit tests.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
@@ -6,12 +6,35 @@ import "../../../common/implementation/Lockable.sol";
 
 /**
  * @title KPI Options Financial Product Library
- * @notice Returns a transformed collateral requirement. The collateralization requirement should
- * always be equivalent to being collateralized by 2 tokens, no matter what the price or expiration status is.
+ * @notice Adds custom tranformation logic to modify the price and collateral requirement behavior of the expiring multi party contract.
+ * If a price request is made pre-expiry, the price should always be set to 2 and the collateral requirement should be set to 1.
+ * Post-expiry, the collateral requirement is left as 1 and the price is left unchanged.
  */
 contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Ownable, Lockable {
     /**
-     * @notice Returns a transformed collateral requirement that is always set to 2 tokens.
+     * @notice Returns a transformed price for pre-expiry price requests.
+     * @param oraclePrice price from the oracle to be transformed.
+     * @param requestTime timestamp the oraclePrice was requested at.
+     * @return transformedPrice the input oracle price with the price transformation logic applied to it.
+     */
+    function transformPrice(FixedPoint.Unsigned memory oraclePrice, uint256 requestTime)
+        public
+        view
+        override
+        nonReentrantView()
+        returns (FixedPoint.Unsigned memory)
+    {
+        // If price request is made before expiry, return 2. Thus we can keep the contract 100% collateralized with
+        // each token backed 1:2 by collateral currency. Post-expiry, leave unchanged.
+        if (requestTime < ExpiringContractInterface(msg.sender).expirationTimestamp()) {
+            return FixedPoint.fromUnscaledUint(2);
+        } else {
+            return oraclePrice;
+        }
+    }
+
+    /**
+     * @notice Returns a transformed collateral requirement that is set to be equivalent to 2 tokens pre-expiry.
      * @param oraclePrice price from the oracle to transform the collateral requirement.
      * @param collateralRequirement financial products collateral requirement to be scaled to a flat rate.
      * @return transformedCollateralRequirement the input collateral requirement with the transformation logic applied to it.
@@ -20,7 +43,7 @@ contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Ownable, 
         FixedPoint.Unsigned memory oraclePrice,
         FixedPoint.Unsigned memory collateralRequirement
     ) public view override nonReentrantView() returns (FixedPoint.Unsigned memory) {
-        // Always return 2 because the option must be collateralized by 2 tokens.
-        return FixedPoint.fromUnscaledUint(2);
+        // Always return 1.
+        return FixedPoint.fromUnscaledUint(1);
     }
 }
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
@@ -1,0 +1,26 @@
+pragma solidity ^0.6.0;
+pragma experimental ABIEncoderV2;
+import "./FinancialProductLibrary.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "../../../common/implementation/Lockable.sol";
+
+/**
+ * @title KPI Options Financial Product Library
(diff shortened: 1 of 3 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -0,0 +1,26 @@
...
+ * always be equivalent to being collateralized by 2 tokens, no matter what the price or expiration status is.
+ */
+contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Ownable, Lockable {
+    /**
+     * @notice Returns a transformed collateral requirement that is always set to 2 tokens.
+     * @param oraclePrice price from the oracle to transform the collateral requirement.
+     * @param collateralRequirement financial products collateral requirement to be scaled to a flat rate.
+     * @return transformedCollateralRequirement the input collateral requirement with the transformation logic applied to it.
+     */
+    function transformCollateralRequirement(
```
nicholaspai, line 19: Should we also be implementing `transformPrice` which returns `2` is the request time is before the expiration time? @mrice32 @chrismaree

smb2796: Yeah I think you're right. It doesn't make much sense as is right now without any price transformation. Setting a CR of 2 isn't actually a constant token amount, unless we also make the price constant pre-expiry.

pemulis (approved):
LGTM! Tests are green locally that that missing `toWei` got added. Thanks for walking me through how the collateralization ratio is calculated from these values.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -0,0 +1,49 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "../../../common/implementation/Lockable.sol";
+
+/**
+ * @title KPI Options Financial Product Library
+ * @notice Adds custom tranformation logic to modify the price and collateral requirement behavior of the expiring multi party contract.
+ * If a price request is made pre-expiry, the price should always be set to 2 and the collateral requirement should be set to 1.
+ * Post-expiry, the collateral requirement is left as 1 and the price is left unchanged.
+ */
+contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Ownable, Lockable {
```
pemulis, line 13: Is it possible in the financial product library framework to deploy a generalized version of this, that would map EMP addresses to pre-expiry prices (in this case, `2`), to save on gas if a single entity was deploying multiple KPI options? If not, is that because deployed EMP addresses are non-deterministic, and maybe this would be possible with CREATE2?

nicholaspai: > If not, is that because deployed EMP addresses are non-deterministic, and maybe this would be possible with CREATE2?

There is probably a way around this, by creating a mapping of `addressees => prices` and modifying the factory code to take in an array of prices to set for deployed contracts and deploying new contracts with these prices.

mrice32, line 13: nit: can we drop ownable and lockable? I don't think they're used in this lib.

mrice32 (approved):
Looks good! One minor comment. Agree with others that we could probably build something more generalized to set an arbitrary pre-expiry price, but 🤷‍♂️ . Probably not worthwhile until we have a reason to generalize :)

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
smb2796: Isn't lockable needed for `nonReentrantView()`?

smb2796: Or is `nonReentrantView()` also not needed because there is no `nonReentrant()` method?

nicholaspai: I think we can drop nonReentrantView yes. It doesn't make external calls

nicholaspai: Hmm technically this makes external contract calls but we also trust these contracts. Hard to say. Is the main benefit to removing Ownable and Lockable byte code and gas? @mrice32

smb2796: Removed Ownable since it seemed completely unused, and left Lockable for now.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
nicholaspai: I'm thinking about this some more, and I think `transformPrice` might actually have to return `1` pre-expiry. The liquidation calculation pre-expiry is as follows:

```
liquidation.settlementPrice = transformPrice(liquidationTime);
FixedPoint.Unsigned memory tokenRedemptionValue =
            liquidation.tokensOutstanding.mul(liquidation.settlementPrice);
FixedPoint.Unsigned memory requiredCollateral =
            tokenRedemptionValue.mul(transformCollateralRequirement(liquidation.settlementPrice));
disputeSucceeded = liquidation.liquidatedCollateral.isGreaterThanOrEqual(requiredCollateral);
```

So if `transformCollateralRequirement` always returns `2` and `transformPrice` always returns 1, then as long as the contract is seeded with at least 2x collateral tokens than tokens minted, then the contract will always be correctly collateralized pre-expiry (assuming no withdrawals are allowed).

nicholaspai: Ok disregard because I see the `transformCollateralRequirement` always returns 1 now

nicholaspai (approved):

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
smb2796: Sorry, my outdated comment was probably confusing. The collateral requirement transformation was updated to always return 1.

chrismaree (approved):
This looks good to me! I initially was confused why we needed the `transformCollateralRequirement` method but it makes sence seeing we are limited to the value we can provide on the CR within an EMP.

chrismaree:
It looks like you had failing tests due to changed byte code. I pulled down your code and addressed the issue. it should pass tests now 🤞

smb2796:
Thanks all for the review and help!

PR #2770, opened 2021-03-25
feat(price-feeds): add funding rate multiplier price feed

mrice32:
**Motivation**

This FRM price feed will enable simple tracking of perpetual funding rate multiplier values for use in determining the correct fair value of a perpetual token.

**Summary**

Added a simple price feed that returns the value of the funding rate multiplier at a particular time.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -435,6 +437,26 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+      config
+    });
+
+    return new FundingRateMultiplierPriceFeed({
+      ...config,
+      logger,
+      web3,
+      getTime,
+      perpetualAbi: Perpetual.abi,
+      blockFinder: getSharedBlockFinder(web3)
```
nicholaspai, line 458: Unrelated: How come we don't use `getSharedBlockFinder` in BalancerPriceFeed? Just haven't gotten around to it?

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
```diff
@@ -0,0 +1,89 @@
...
+  async update() {
+    const currentTime = await this.getTime();
+    if (this.lastUpdateTime === undefined || currentTime >= this.lastUpdateTime + this.minTimeBetweenUpdates) {
+      this.price = await this._getPrice();
+      this.lastUpdateTime = currentTime;
+    }
+  }
+
+  async _getPrice(blockNumber = "latest") {
+    const { cumulativeMultiplier } = await this.perpetual.methods.fundingRate().call(undefined, blockNumber);
```
nicholaspai, line 82: Now that `multicall` is merged, should you use it to handle case where `fundingRate` hasn't been published yet, but the effective funding rate would be different than the current funding rate for a particular block?

mrice32: Yep! I can add that!

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
```diff
@@ -0,0 +1,89 @@
...
+  /**
+   * @notice Constructs new price feed object that tracks the funding rate multiplier in a particular perpetual contract.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} perpetualAbi Perpetual abi object to create a contract instance.
+   * @param {Object} web3 Provider from Truffle instance to connect to Ethereum network.
+   * @param {String} perpetualAddress Ethereum address of the perpetual to monitor.
+   * @param {Function} getTime Returns the current time.
+   * @param {Function} [blockFinder] Optionally pass in a shared blockFinder instance (to share the cache).
+   * @param {Integer} [minTimeBetweenUpdates] Minimum amount of time that must pass before update will actually run
+   *                                        again.
```
chrismaree, line 16:
```suggestion
   * @param {Integer} [minTimeBetweenUpdates] Minimum time that must pass before update will actually run again.
```

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
```diff
@@ -0,0 +1,89 @@
...
+    minTimeBetweenUpdates = 60,
+    priceFeedDecimals = 18
+  }) {
+    super();
+
+    // Assert required inputs.
+    assert(logger, "logger required");
+    assert(perpetualAbi, "perpetualAbi required");
+    assert(web3, "web3 required");
+    assert(perpetualAddress, "perpetualAddress required");
```
chrismaree, line 36:
```suggestion
    assert(web3.utils.isAddress(perpetualAddress), "perpetualAddress required");
```

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
```diff
@@ -0,0 +1,89 @@
...
+    this.logger = logger;
+    this.web3 = web3;
+
+    this.perpetual = new web3.eth.Contract(perpetualAbi, perpetualAddress);
+    this.uuid = `FundingRateMultiplier-${perpetualAddress}`;
+    this.getTime = getTime;
+    this.priceFeedDecimals = priceFeedDecimals;
+    this.minTimeBetweenUpdates = minTimeBetweenUpdates;
+    this.blockFinder = blockFinder || BlockFinder(web3.eth.getBlock);
+    this.convertDecimals = ConvertDecimals(18, priceFeedDecimals, web3);
```
chrismaree, line 48: question: in what context would this _not_ be 18 decimals? does a perpetual with a non-18 decimal collateral still have a funding rate of 18 decimals? If it does, we should probs consider removing any notion of decimals from this feed to simplify it a bit.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
chrismaree, line 82: this assumes an archival node for anything with a long lookback. I think that's fine though as your internal use of this method does not provide anything to the `blockNumber` param.

chrismaree (approved):
Nice and simple. good code. I left a few small code change recommendations. Otherwise, this looks great! will be cool to see this pulled together within the expression feed.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: It uses a different library for block caching. Specifically, it uses a library that pulls in _all_ blocks between two points for the purpose of computing a TWAP. The sharedBlockCache is using a library that does a binary search, and therefore produces a sparse block list.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
mrice32: It does require an archival node to serve historical prices. This is also required in the vault and LP price feeds.

@nicholaspai and I discussed this. I think the most reasonable solution is to (in a follow-up) add an env variable to specify whether you're using an archival node. If not, these price feeds should have a lookback of 30 minutes. If so, they keep their current lookback value on `Infinite`. Thoughts?

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
mrice32: Yes, the value retrieved from the blockchain is always 18 decimals since it is a percentage. However, we still follow the pattern of returning the value in whatever decimals the user specifies. Should we be doing something different in this price feed?

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
mrice32: Done!

yarn.lock
```diff
@@ -1363,7 +1363,7 @@
   dependencies:
     regenerator-runtime "^0.13.4"
 
-"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.11.2", "@babel/runtime@^7.2.0", "@babel/runtime@^7.3.1", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.4", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.2", "@babel/runtime@^7.8.3", "@babel/runtime@^7.8.4", "@babel/runtime@^7.8.7":
+"@babel/runtime@^7.0.0", "@babel/runtime@^7.10.2", "@babel/runtime@^7.11.2", "@babel/runtime@^7.2.0", "@babel/runtime@^7.3.1", "@babel/runtime@^7.3.4", "@babel/runtime@^7.4.4", "@babel/runtime@^7.5.5", "@babel/runtime@^7.6.3", "@babel/runtime@^7.7.2", "@babel/runtime@^7.7.4", "@babel/runtime@^7.8.3", "@babel/runtime@^7.8.4", "@babel/runtime@^7.8.7":
```
mrice32, line 1366: When running yarn to install packages, it seems that some previous packages had not been included in the lockfile, so they were committed here.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
chrismaree: ye, that seems reasonable.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
chrismaree: Right but the whole point of defining a `prifeFeedDecimals` is to match the value that the Price feed returns to the value that the DVM would return. If the DVM would scale a price (such as an old BTC identifier/USDC) then the price feed should scale it the same way.

In the case of this feed the DVM will always return the price scaled at 18 decimals and so I dont see a reason to let the user of the feed parameterize this as it will by default match up with what dvm returns at this scalling.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -435,6 +438,40 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    }
+
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating FundingRateMultiplierPriceFeed",
+      config
+    });
+
+    // Grab multicall address from network (will be undefined if any lookups fail).
+    const networkId = await web3.eth.net.getId();
```
nicholaspai, line 454: Can you skip this part if `config.multicallAddress` is provided by the caller?

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
nicholaspai: I see no problem with enabling the flexibility for the user to decide.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  getCurrentPrice() {
+    return this.price;
+  }
+
+  async getHistoricalPrice(time) {
+    const block = await this.blockFinder.getBlockForTimestamp(time);
+    return this._getPrice(block.number);
```
nicholaspai, line 61: what if `block == undefined`?

mrice32: Then that should be an error, right?

nicholaspai (approved):
Really enjoyed reading this code, it was very clean. LGTM

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
mrice32: Yeah, but this price feed value isn't generally used alone. This output could be used directly or it could be used in some sort of aggregation where a particular decimal value is needed (fallback or medianizer price feeds). I think keeping it consistent with the other feeds is the safest option even if it makes this feed a little more complex.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Yep, I'll change this.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.js
mrice32: I don't think the blockfinder returns undefined. It will error out if it cannot find a block.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
mrice32: Done.

PR #2772, opened 2021-03-25
feat(price-feeds): add twap to cryptowatch

mrice32:
**Motivation**

CW price feeds need to be able to compute a TWAP to allow us to generate more stable prices for funding rate calculations.

**Summary**

Uses the TWAP utility function to compute the TWAP across the cryptowatch array of price periods.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -1,5 +1,7 @@
 const { PriceFeedInterface } = require("./PriceFeedInterface");
 const { parseFixed } = require("@uma/common");
+const { computeTWAP } = require("./utils");
+const lodash = require("lodash");
 
 // An implementation of PriceFeedInterface that uses CryptoWatch to retrieve prices.
 class CryptoWatchPriceFeed extends PriceFeedInterface {
@@ -33,7 +35,8 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     minTimeBetweenUpdates,
     invertPrice,
     priceFeedDecimals = 18,
-    ohlcPeriod = 60 // One minute is CryptoWatch's most granular option.
+    ohlcPeriod = 60, // One minute is CryptoWatch's most granular option.
+    twapLength = 0 // No TWAP by default.
   ) {
     super();
     this.logger = logger;
@@ -43,6 +46,7 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     this.exchange = exchange;
     this.pair = pair;
     this.lookback = lookback;
+    this.twapLength = twapLength;
     this.uuid = `CryptoWatch-${exchange}-${pair}`;
     this.networker = networker;
     this.getTime = getTime;
@@ -71,6 +75,15 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       throw new Error(`${this.uuid}: undefined lastUpdateTime`);
     }
 
+    // Return early if computing a TWAP.
+    if (this.twapLength) {
+      const twapPrice = this._computeTwap(time, this.historicalPricePeriods);
+      if (!twapPrice) {
+        throw new Error(`${this.uuid}: historical TWAP computation failed due to no data in the TWAP range`);
+      }
+      return twapPrice;
+    }
+
     // Set first price period in `historicalPricePeriods` to first non-null price.
     let firstPricePeriod;
     for (let p in this.historicalPricePeriods) {
@@ -80,7 +93,7 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       }
     }
 
-    // If there are no valid price periods, return null.
+    // If there are no valid price periods, throw.
     if (!firstPricePeriod) {
       throw new Error(`${this.uuid}: no valid price periods`);
     }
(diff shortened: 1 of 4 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -181,46 +194,56 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
...
+      : await this._getImmediatePrice();
+
+    // 5. Store results.
+    this.currentPrice = newPrice;
+    this.historicalPricePeriods = newHistoricalPricePeriods;
+    this.lastUpdateTime = currentTime;
+  }
 
-    // 1. Construct URLs.
+  async _getImmediatePrice() {
```
nicholaspai, line 211: Is the idea eventually to make getCurrentPrice just return _getImmediatePrice()? I.e. skip the `update()` process

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -554,7 +554,7 @@ const defaultConfigs = {
         uniswapAddress: "0xcd7989894bc033581532d2cd88da5db0a4b12859",
         invertPrice: true
       },
-      BADGER_USD_HUOBI: { type: "cryptowatch", exchange: "huobi", pair: "badgerusdt" },
+      BADGER_USD_HUOBI: { type: "cryptowatch", exchange: "huobi", pair: "badgerusdt", twapLength: 0 },
```
nicholaspai, line 557: Doesn't this default to 0?

packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -178,6 +179,87 @@ contract("CryptoWatchPriceFeed.js", function() {
...
+    // Inject data.
+    networker.getJsonReturns = [...validResponses];
+
+    await cryptoWatchPriceFeed.update();
+
+    // ((12 * 60) + (11 * 60)) / 120
+    assert.equal((await cryptoWatchPriceFeed.getHistoricalPrice(1588376460)).toString(), web3.utils.toWei("1.15"));
+  });
+
+  it("TWAP fails if period starts before data", async function() {
```
nicholaspai, line 234: How does TWAP perform if some data is missing inside the twap window? Think about the cases where data is missing in the middle, beginning, and end of window

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -1,5 +1,7 @@
 const { PriceFeedInterface } = require("./PriceFeedInterface");
 const { parseFixed } = require("@uma/common");
+const { computeTWAP } = require("./utils");
```
nicholaspai, line 3: +1 for not having to implement the TWAP algo again!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -252,11 +275,21 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
...
-    this.currentPrice = newPrice;
-    this.historicalPricePeriods = newHistoricalPricePeriods;
-    this.lastUpdateTime = currentTime;
+  _computeTwap(endTime, ohlcs) {
+    // Combine open and close to get more data fidelity at the edges of the range.
+    const priceTimes = lodash.flatten(
+      ohlcs.map(pricePeriod => {
+        return [
+          [pricePeriod.openTime, pricePeriod.openPrice],
+          [pricePeriod.closeTime, pricePeriod.closePrice]
```
nicholaspai, line 286: Is there any time between a period's `closeTime` and the next's `openTime`?

nicholaspai (approved):
It's nice that the default CW pricefeed behavior stays the same, +1 for re-using the `utils.js` computeTwap method. I just want to make sure you've tested edge case behavior with missing data

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
mrice32: Yeah, but we define `twapLength` a level above for the uniswap feeds, so that gets inherited by this config. To set it back to 0, it needs to be specified.

packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js
mrice32: It uses the same backing algorithm as the uniswap feed, which is to say that it doesn't know anything about the expected data frequency. The CW feed, in general, is agnostic to how "filled" the data is. All it cares about is that there is data _before_ the earliest point you care about (which is what this test case is testing). Is there a test case that you think should be added on this point?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: There shouldn't be unless:
- An ohlc point is missing.
- The data point is the first or last, so there is only an `openTime` (beginning of the data) or a `closeTime` (end of the data).

Both of these shouldn't affect the output too much, but I think including both makes the calculation a little more robust.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: Not exactly. `_getImmediatePrice` is only used when twapLength === 0. So if we were to remove `update`, `getCurrentPrice` would be a ternary that either returns `_getImmediatePrice` or `_computeTwap` depending on twapLength.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -181,46 +194,56 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     // Round down to the nearest ohlc period so the queries captures the OHLC of the period *before* this earliest
     // timestamp (because the close of that OHLC may be relevant).
-    const earliestHistoricalTimestamp = Math.floor((currentTime - this.lookback) / this.ohlcPeriod) * this.ohlcPeriod;
+    const earliestHistoricalTimestamp =
+      Math.floor((currentTime - (this.lookback + this.twapLength)) / this.ohlcPeriod) * this.ohlcPeriod;
```
daywiss, line 198: probably wrong, but this logic is confusing. i can imagine though if you used `Math.ceil` instead, you would get a longer lookback time because of the multiplication by `ohlcPeriod` at the end, adding an additional ohlc period if there is a decimal, rather than truncating it.  Not totally sure though if that is the intent

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -252,11 +275,21 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
...
       });
+  }
 
-    // 5. Store results.
-    this.currentPrice = newPrice;
-    this.historicalPricePeriods = newHistoricalPricePeriods;
-    this.lastUpdateTime = currentTime;
+  _computeTwap(endTime, ohlcs) {
+    // Combine open and close to get more data fidelity at the edges of the range.
+    const priceTimes = lodash.flatten(
```
daywiss, line 282: `flat()` is built into arrays fyi, also awesome way of creating the price array

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
daywiss: just as a side note, i find it very weird that historical data is represented as ohlc if closeTime/ next openTime are identical and open/close price are identical. its really redundant. I'm assuming that its this way because they want the ability to differentiate between close/open price/time if needed.

daywiss (approved):
looks great

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: oh good to know -- will use that instead!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: Yeah, I'm really not sure why it's represented this way. This is the traditional "candle" way of viewing data. My guess is that it mostly has to do with the fact that they want a single candle to be self-contained. It's also somewhat unclear what they mean by price -- is it trade price or midpoint. If it's midpoint, it's pretty much redundant. If it's first/last trade, then I could see them being different values and having some value in being distinguished.

packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js
mrice32: Added tests for each of these cases!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: yeah, this is a little confusing. The floor makes the lookback longer because we're not computing a distance or difference, we're computing the _actual timestamp_. We would ceil the _distance_ or _difference_, but to create a timestamp that is further in the past, we want to floor since that creates the lower value. Does that makes sense?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: Done!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
daywiss: yea that makes sense, we want the smaller timestamp. for some reason i kept thinking it was time ago ( where larger number is further in past), rather than absolute time in past.

daywiss (approved):
lgtm

PR #2775, opened 2021-03-26
feat(liquidation-reserve-currency): add reserve currency liquidator to bots and refine smart contract

### LINKED ISSUE #2751: Ability for bots to only need to hold a single reserve currency versus  holding collateral and tokens for each EMP

chrismaree:
**Motivation**

issue #2751

**Summary**

This PR integrates the `SwapMintLiquidate` contract added in PR #2750 into the liquidator bot. Additionally, this PR modifies the original contract to consider any existing synthetics or collateral that the DSProxy might already have.

**Details**

The base of the implementation is built around a new module added to the bots: a `ProxyTransactionWrapper`. this wrapper abstracts the action of sending liquidation transactions away from the implementation thereby separating concerns between liquidation decision making and transaction sending. 

This module implements two main functions:

1. `getEffectiveSyntheticTokenBalance` which returns the balance of the DSProxy in synthetics + the total amount of synthetics the bot _could_ have, considering the collateral and reserve currencies based off a swap & mint transaction. This is used when computing WDF logic. If the bot is running in normal EOA mode then this simply returns the current synthetic balance.
2. `submitLiquidationTransaction` which sends a liquidation tx. This is either a normal EOA liquidation (as before) or is routed via the DSProxy liquidator.

The code change in this PR is pretty large. I've tried to identify key areas of interest and draw attention for the reviewer to places where the most complex/important logic is located.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

close #2751


packages/affiliates/payouts/2021-03-15_2021-03-22_0019.json
```diff
@@ -6,187 +6,45 @@
     "startTime": 1615849200000,
     "endTime": 1616454000000,
     "empWhitelist": [
-      [
```
chrismaree, line 9: unrelated the main PR. changes came due to running `yarn lint-fix`

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -2,6 +2,7 @@ pragma solidity ^0.6.0;
 pragma experimental ABIEncoderV2;
```
chrismaree, line 2: This contract was basically re-written with a bunch more logic within it. In particular, the reserve currency liquidator now considers any existing synthetic or colalteral tokens the DSproxy might have. For example, if we manually sent the DSProxy a bunch of synthetic tokens it should use these before minting extras.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBroker.sol
```diff
@@ -24,7 +24,7 @@ contract UniswapBroker {
      * @dev True price is expressed in the ratio of token A to token B.
      * @dev The caller must approve this contract to spend whichever token is intended to be swapped.
      * @param tradingAsEOA bool to indicate if the UniswapBroker is being called by a DSProxy or an EOA.
-     * @param uniswapRouter address of the uniswap router used to facilate trades.
```
chrismaree, line 27: fixed some typos.

packages/core/test/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.js
```diff
@@ -64,6 +64,19 @@ contract("ReserveTokenLiquidator", function(accounts) {
   const sponsor2 = accounts[2];
```
chrismaree, line 64: These tests were heavily extended to accommodate the new functionality.

packages/disputer/test/Disputer.js
```diff
@@ -73,7 +73,7 @@ const _setFundingRateAndAdvanceTime = async fundingRate => {
 
 // If the current version being executed is part of the `supportedVersions` array then return `it` to run the test.
 // Else, do nothing. Can be used exactly in place of a normal `it` to parameterize contract types and versions supported
-// for a given test.eg: versionedversionedIt([{ contractType: "any", contractVersion: "any" }])(["Perpetual-latest"])("test name", async function () { assert.isTrue(true) })
```
chrismaree, line 76: there was this common typo throughout the repo. unrelated to main PR.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -24,11 +24,6 @@ class DSProxyManager {
     this.dsProxyAddress = null;
 
     // Helper functions from web3.
```
chrismaree, line 26: unrelated to main code change. this utility had unused declerations.

packages/liquidator/index.js
```diff
@@ -196,9 +204,40 @@ async function run({
     const gasEstimator = new GasEstimator(logger);
     await gasEstimator.update();
 
+    const dsProxyManager = new DSProxyManager({
```
chrismaree, line 207: main change here is that we now need to instantiate the `dsProxyManager` and the `proxyTransaction` wrapper to enable. The latter is fed into the liquidator as a parameter.

packages/liquidator/index.js
```diff
@@ -337,7 +376,22 @@ async function Poll(callback) {
       startingBlock: process.env.STARTING_BLOCK_NUMBER,
       // Block number to search for events to. If set, acts to limit from where the monitor bot will search for events up
       // until. If either startingBlock or endingBlock is not sent, then the bot will search for event.
-      endingBlock: process.env.ENDING_BLOCK_NUMBER
+      endingBlock: process.env.ENDING_BLOCK_NUMBER,
+      // If enabled, the bot will funnel liquidations via a DSProxy which will be deployed on the bots behalf. This
```
chrismaree, line 380: how the DSProxy is used is configurable through envs, as per usual.

packages/liquidator/src/liquidator.js
```diff
@@ -205,7 +205,9 @@ class Liquidator {
   async liquidatePositions(maxTokensToLiquidateWei, liquidatorOverridePrice) {
     this.logger.debug({
       at: "Liquidator",
-      message: "Checking for liquidatable positions and preforming liquidations"
+      message: "Checking for liquidatable positions and preforming liquidations",
```
chrismaree, line 208: improved this log a bit by adding the params send in.

packages/liquidator/index.js
```diff
@@ -250,7 +289,7 @@ async function run({
           if (!isExpiredOrShutdown) {
             // Check for liquidatable positions and submit liquidations. Bounded by current synthetic balance and
             // considers override price if the user has specified one.
-            const currentSyntheticBalance = await syntheticToken.methods.balanceOf(accounts[0]).call();
```
chrismaree, line 253: note this change: previously the first parram fed into the `liquidatePositions` method was the current synthetic token balance of the liquidator. Now, it uses the `proxyTransactionWrapper`'s `getEffectiveSyntheticTokenBalance` method. This method considers: a) how many tokens the DSProxy already has b) how much collateral the DSProx has, and how many synths could be minted from this collateral and c) how much collateral the DSProxy has, how much collateral this could buy, and how many synthetics this could mint. i.e an "effective" balance that is available to the bot at the moment of liquidation.

packages/liquidator/src/liquidator.js
```diff
@@ -269,23 +271,22 @@ class Liquidator {
...
       const liquidationArgs = this.liquidationStrategy.processPosition({
-        // position is assumed to be undercollateralized
-        position,
-        // need to know our current balance to know how much to save for defense fund
-        syntheticTokenBalance,
-        currentBlockTime,
-        // this is required to create liquidation object
-        maxCollateralPerToken,
-        // maximum tokens we can liquidate in position
+        position, // position is assumed to be undercollateralized
```
chrismaree, line 284: styling nit. no need to use new lines here.

packages/liquidator/src/liquidator.js
```diff
@@ -310,79 +311,31 @@ class Liquidator {
...
+      this.logger.debug({
+        at: "Liquidator",
+        message: "Liquidating position",
+        position,
+        inputPrice: scaledPrice.toString(),
+        minLiquidationPrice: this.liquidationMinPrice,
+        maxLiquidationPrice: maxCollateralPerToken.toString(),
+        tokensToLiquidate: tokensToLiquidate.toString()
+      });
+      const logResult = await this.proxyTransactionWrapper.submitLiquidationTransaction(liquidationArgs);
```
chrismaree, line 323: this is the primary change to this file. previously, the `liquidator` would submit the liquidation inline here. see the code removed 👇 .

After this PR, the liquidator simply calls the `.proxyTransactionWrapper.submitLiquidationTransaction(liquidationArgs)` which handles all logic associated with the actual liquidation transaction.

packages/liquidator/src/liquidator.js
```diff
@@ -310,79 +311,31 @@ class Liquidator {
...
-        // Receipt without events
-        receipt = await ynatm.send({
-          sendTransactionFunction: gasPrice => liquidation.send({ ...txnConfig, nonce, gasPrice }),
-          minGasPrice,
-          maxGasPrice,
-          gasPriceScalingFunction,
-          delay: 60000 // Tries and doubles gasPrice every minute if tx hasn't gone through
-        });
-      } catch (error) {
+      if (logResult instanceof Error || !logResult)
```
chrismaree, line 325: log depending on the output.

packages/merkle-distributor/scripts/example.json
```diff
@@ -37,4 +37,4 @@
       "metaData": { "reason": ["Governance Votes KPI Options Round 1"] }
     }
   }
-}
```
chrismaree, line 40: change due to lint.

yarn.lock
```diff
@@ -1363,7 +1363,7 @@
   dependencies:
```
chrismaree, line 1363: not 100% sure why the diff here is so large. This simply came about after running a yarn install. Should I revert this file changes @UMAprotocol/eng ?

packages/liquidator/test/Liquidator.js
```diff
@@ -1365,8 +1395,7 @@ contract("Liquidator.js", function(accounts) {
               );
               await financialContract.setCurrentTime(nextTime);
 
-              // Liquidator only has enough balance to liquidate 1 position fully, will
-              // minimally liquidate the other
+              // Liquidator only has enough balance to liquidate 1 position fully, will minimally liquidate the other
```
chrismaree, line 1398: nit: line length can be wrapped.

packages/liquidator/test/Liquidator.js
```diff
@@ -1659,6 +1686,464 @@ contract("Liquidator.js", function(accounts) {
             }
           );
         });
+        describe("Liquidation via DSProxy", () => {
```
chrismaree, line 1689: pretty in depth tests. please let me know if there is an edge case I dont cover!

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+        { rawValue: this.maxReserverTokenSpent }, // maxReserverTokenSpent
+        { rawValue: liquidationArgs[1].rawValue }, // minCollateralPerTokenLiquidated
+        { rawValue: liquidationArgs[2].rawValue }, // maxCollateralPerTokenLiquidated. This number need to be >= the token price.
+        { rawValue: liquidationArgs[3].rawValue }, // maxTokensToLiquidate. This is how many tokens the positions has (liquidated debt).
+        liquidationArgs[4]
+      )
+      .encodeABI();
+    const callCode = this.ReserveCurrencyLiquidator.bytecode;
+
+    const dsProxyCallReturn = await this.dsProxyManager.callFunctionOnNewlyDeployedLibrary(callCode, callData);
```
chrismaree, line 261: note that this does not use `ynatm`. a separate PR should be added to implement this gas escalation within the `dsProxyManager`. I added a tasks to track this and will work on it as soon as this is done with code review: https://github.com/UMAprotocol/protocol/issues/2788

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+        this.web3.utils.isAddress(this.liquidatorReserveCurrencyAddress),
+        "Must provide a reserve currency address to use the proxy transaction wrapper!"
+      );
+    }
+
+    this.reserveToken = new web3.eth.Contract(getAbi("ExpandedERC20"), this.liquidatorReserveCurrencyAddress);
+    this.ReserveCurrencyLiquidator = getTruffleContract("ReserveCurrencyLiquidator", web3, "latest");
+  }
+
+  // TODO: wrap this into a common util.
```
chrismaree, line 110: This needs to be refactored. I did not put it in this PR as it is in a few other places. I added an issue to track this task here: https://github.com/UMAprotocol/protocol/issues/2789

packages/liquidator/src/liquidator.js
```diff
@@ -393,13 +346,29 @@ class Liquidator {
       message: "Checking for expired and disputed liquidations to withdraw rewards from"
     });
 
+    // In legacy versions of the EMP, withdrawing needs to be done by a party involved in the liquidation (i.e liquidator,
+    // sponsor or disputer).As the liquidator is the DSProxy, we would require the ability to send the withdrawal tx
+    // directly from the DSProxy to facilitate this.This functionality is not implemented as almost all legacy EMPs expired.
+    if (this.proxyTransactionWrapper.isUsingDsProxyToLiquidate && this.isLegacyEmpVersion) {
```
chrismaree, line 352: @UMAprotocol/eng reviewers: not this decision here. If the bot is using DSProxy and the contract is legacy then withdrawing rewards is disabled. I chose to do this to keep the overall code churn and feature addition as simple as possible, especially seeing that almost all legacy contracts have expired. If a liquidation happens on a legacy contract we can still liquidate. However, we will need to manually withdraw from the DSProxy.

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -39,44 +42,64 @@ contract ReserveCurrencyLiquidator {
...
+        // will factor this in. If the DSProxy has more collateral than the total amount required the purchased = 0.
+        FixedPoint.Unsigned memory collateralToBePurchased =
+            subOrZero(
+                totalCollateralRequired,
+                FixedPoint.Unsigned(IERC20(fc.collateralCurrency()).balanceOf(address(this)))
+            );
+
+        // 4.a. If there is some collateral to be purchased, execute a trade on uniswap to meet the shortfall.
+        // Note the path assumes a direct route from the reserve currency to the collateral currency.
+        if (collateralToBePurchased.isGreaterThan(FixedPoint.fromUnscaledUint(0))) {
```
nicholaspai, line 79: What if `fc.collateralCurrency() == reserveCurrency`?

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -39,44 +42,64 @@ contract ReserveCurrencyLiquidator {
         address financialContract,
         address reserveCurrency,
         address liquidatedSponsor,
-        FixedPoint.Unsigned calldata maxReserverTokenSpent,
+        FixedPoint.Unsigned calldata maxReserveTokenSpent,
         FixedPoint.Unsigned calldata minCollateralPerTokenLiquidated,
         FixedPoint.Unsigned calldata maxCollateralPerTokenLiquidated,
         FixedPoint.Unsigned calldata maxTokensToLiquidate,
         uint256 deadline
     ) public {
```
nicholaspai, line 50: I thought about reentrancy vulnerabilities here and I don't think there's any issue since this contract doesn't store any internal state, but just want to make sure you've thought about that too and perhaps you add a reentrancy guard if its a concern.

packages/liquidator/index.js
```diff
@@ -49,7 +51,8 @@ async function run({
   liquidatorConfig,
   liquidatorOverridePrice,
   startingBlock,
-  endingBlock
+  endingBlock,
+  proxyTransactionWrapperConfig
```
nicholaspai, line 55: Can you add comments about this param?

packages/liquidator/index.js
```diff
@@ -196,9 +202,37 @@ async function run({
...
+    let dsProxyManager;
+    if (proxyTransactionWrapperConfig?.isUsingDsProxyToLiquidate) {
+      dsProxyManager = new DSProxyManager({
+        logger,
+        web3,
+        gasEstimator,
+        account: accounts[0],
+        dsProxyFactoryAddress:
+          proxyTransactionWrapperConfig?.dsProxyFactoryAddress || getAddress("DSProxyFactory", networkId),
+        dsProxyFactoryAbi: getAbi("DSProxyFactory"),
```
nicholaspai, line 214: Since these two ABI's are hardcoded, I think it might be cleaner to call `getAbi()` from within the DXProxyManager. I do this with the `funding-rate-proposer/src/proposer.js` client, but I'm not strongly opionated.

packages/liquidator/index.js
```diff
@@ -196,9 +202,37 @@ async function run({
...
+        dsProxyAbi: getAbi("DSProxy")
+      });
+
+      await dsProxyManager.initializeDSProxy();
```
nicholaspai, line 218: What does this function do? Might be worth adding a comment about which side effects trigger

packages/liquidator/src/liquidator.js
```diff
@@ -269,23 +271,22 @@ class Liquidator {
 
       // Note: query the time again during each iteration to ensure the deadline is set reasonably.
       const currentBlockTime = this.financialContractClient.getLastUpdateTime();
-      const syntheticTokenBalance = this.toBN(await this.syntheticToken.methods.balanceOf(this.account).call());
+
+      // TODO: right now the `processPosition` method takes in maxTokensToLiquidateWei and syntheticTokenBalance. These
```
nicholaspai, line 275: Don't you think it could be useful to distinguish between the bot's current balance and the amount of balance its allocating to liquidate with? I.e. it allows the caller to set a limit on how many tokens to liquidate

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+    let truffleContractCreator = truffleContract(contractJsonObject);
+    truffleContractCreator.setProvider(web3.currentProvider);
+    return truffleContractCreator;
+  }
+
+  // Get the effective synthetic token balance. If the bot is executing in normal mode (liquidations sent from an EOA)
+  // then this is simply the token balance of the unlocked account. If the liquidator is using a DSProxy to liquidate,
+  // then consider the synthetics could be minted, + any synthetics the DSProxy already has.
+  async getEffectiveSyntheticTokenBalance() {
+    const syntheticTokenBallance = await this.syntheticToken.methods.balanceOf(this.account).call();
```
nicholaspai, line 121: nit: mispelled `balance`

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+   * */
+  constructor({
+    web3,
+    financialContract,
+    gasEstimator,
+    syntheticToken,
+    collateralToken,
+    account,
+    dsProxyManager = undefined,
+    isUsingDsProxyToLiquidate = false,
```
nicholaspai, line 38: Maybe i'm trying to over-refactor, but why can't we remove the `isUsingDsProxyToLiquidate` config boolean, and just check if `dsProxyManager == undefined`? It seems below on lines 94, 122,  and 182 that you are checking the former but I feel like you can also just check for `!dsProxymanager`

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+
+  // Main entry point for submitting a liquidation. If the bot is not using a DSProxy then simply send a normal EOA tx.
+  // If the bot is using a DSProxy then route the tx via it.
+  async submitLiquidationTransaction(liquidationArgs) {
+    // If the liquidator is not using a DSProxy, use the old method of liquidating
+    if (!this.isUsingDsProxyToLiquidate) return await this._executeLiquidationWithoutDsProxy(liquidationArgs);
+    else return await this._executeLiquidationWithDsProxy(liquidationArgs);
+  }
+
+  async _executeLiquidationWithoutDsProxy(liquidationArgs) {
```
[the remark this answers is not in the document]
nicholaspai: Maybe this isn't the PR to do this but we should probably refactor the `ynatm` retry logic into `common/src/ContractUtils.js:runTransaction()`, which currently manages transactions for the `funding-rate-proposer` and `optimistic-oracle-proposer`. This would remove the `estimateGas-send` code below.

packages/liquidator/src/proxyTransactionWrapper.js
nicholaspai: Can we refactor some of this into a generalized function in  `common/src/ContractUtils`?

nicholaspai (commented):
Tons of good code good job, left some comments for my first pass. Once you address I'll go through again.

packages/liquidator/src/proxyTransactionWrapper.js
chrismaree: yup! I will change the `dsProxyManager` to use this. 

However, it looks like your use of the `runTransaction` in the proposer does not use ynatm either. we should ideally modify the run transaction to use this so we consistently bump gas all over the place.

packages/liquidator/src/liquidator.js
chrismaree: @nicholaspai I think it's fine as there are basically no contracts running on the legacy version.

@mrice32 yup! we would need a very simple script to do this. I added a simple smart contract to this PR that we could use, if the need arises. it implements the correct interface for this contract version:

```
pragma solidity ^0.6.0;
pragma experimental ABIEncoderV2;

import "../../common/implementation/FixedPoint.sol";

// Simple contract used to withdraw liquidations using a DSProxy from legacy contracts (1.2.2 and below).
contract LiquidationWithdrawer {
    function withdrawLiquidation(
        address financialContractAddress,
        uint256 liquidationId,
        address sponsor
    ) public returns (IFinancialContract.RewardsData memory) {
        return IFinancialContract(financialContractAddress).withdrawLiquidation(liquidationId, sponsor);
    }
}

interface IFinancialContract {
    function withdrawLiquidation(uint256 liquidationId, address sponsor)
        external
        returns (FixedPoint.Unsigned memory amountWithdrawn);
}
```

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
```diff
@@ -39,44 +42,64 @@ contract ReserveCurrencyLiquidator {
...
+            subOrZero(maxTokensToLiquidate, FixedPoint.Unsigned(IERC20(fc.tokenCurrency()).balanceOf(address(this))));
 
-        // 4. Mint synthetics with collateral. Note we are minting at the GCR and minting the exact number liquidated.
-        TransferHelper.safeApprove(fc.collateralCurrency(), address(fc), totalCollateralNeeded.rawValue);
-        fc.create(collateralToMintAtGcr, maxTokensToLiquidate);
-
-        // 5. Liquidate position with newly minted synthetics.
+        // 2. Calculate how much collateral is needed to make up the token shortfall from minting new synthetics.
+        FixedPoint.Unsigned memory gcr = fc.pfc().div(fc.totalTokensOutstanding());
+        FixedPoint.Unsigned memory collateralToMintShortfall = tokenShortfall.mul(gcr);
```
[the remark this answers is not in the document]
chrismaree: this is a great point @nicholaspai! I dont think there is any risk to change this to divCeil other than there might be a bit of dust left over in the contract. this is a strictly better state than the contract having a shortfall of synthetics.

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: This is a great question. I'm going to write some unit tests to verify this but my understanding is that the trade will just not happen. might be better to have some branching logic

chrismaree: I added a test which validates this. Simply no trade happens and everything else continues on as it was before. no additional branching is needed; the `swapTokenForExactTokens` simply does nothing and returns :)

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
chrismaree: ye, not too worried about it. this contract is called from a DSProxy and so the context of state is the DSProxy storage. This contract should never hold funds either.

packages/liquidator/index.js
chrismaree: Ideally I dont want to modify the `DSProxyManager` in this PR though as I feel that'll be covering too much. I'll do it in a follow up PR where I refactor the `DSProxyManager` to use the `runTransaction`.

packages/liquidator/index.js
chrismaree: it will load in an existing DSProxy for the account EOA if one already exists or create a new one for the user. I added a comment

packages/liquidator/src/liquidator.js
chrismaree: ye, this is something I spent quite a lot of time thinking about. Previously, this parameter was simply the bots token balance. Within the `liquidatePositions` method it also read in the current token balance. Then, the WDF logic took in two parameters that were _the same_.

What I did here is implement this to be the _same_ as the previous implementation with the goal of not modifying key logic that we know to work within the WDF. In a separate PR we should refactor this. There were a few other things in how the WDF is called and the data it returns that I'd also like to change.

I felt it best to mimic the current logic as much as posible and leave these changes for later. do you agree?

packages/liquidator/src/proxyTransactionWrapper.js
chrismaree: you could do that. the reason I did it like this is that I want the user to explicitly define a `useDsProxyToLiquidate=true` within the `DSPROXY_CONFIG` and by doing it like this I can simply pass down the `proxyTransactionWrapperConfig` into the wrapper. if I modified it the way you described I'd need to destruct the config within the index and do some logic on it rather than just passing it around>

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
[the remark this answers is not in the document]
chrismaree: good point. done!

packages/liquidator/src/proxyTransactionWrapper.js
```diff
@@ -0,0 +1,288 @@
...
+  createContractObjectFromJson(contractJsonObject) {
+    let truffleContractCreator = truffleContract(contractJsonObject);
+    truffleContractCreator.setProvider(web3.currentProvider);
+    return truffleContractCreator;
+  }
+
+  // Get the effective synthetic token balance. If the bot is executing in normal mode (liquidations sent from an EOA)
+  // then this is simply the token balance of the unlocked account. If the liquidator is using a DSProxy to liquidate,
+  // then consider the synthetics could be minted, + any synthetics the DSProxy already has.
+  async getEffectiveSyntheticTokenBalance() {
```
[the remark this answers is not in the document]
chrismaree: great point and a good refactor. it would be posible to return say the "effective" balance from the smart contract and then call this as a view function. However, this contract is already at the limit of size and any extra returns cause it to bow over the limit :(

For now I think its ok to leave like this, if you ok with that?

packages/liquidator/src/liquidator.js
nicholaspai: Is this tested?

packages/core/contracts/bot-helpers/atomic-liquidator/ReserveCurrencyLiquidator.sol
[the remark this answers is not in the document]
nicholaspai: Agreed

packages/liquidator/src/liquidator.js
nicholaspai: Yes agreed, minimize churn for this PR.

packages/liquidator/src/proxyTransactionWrapper.js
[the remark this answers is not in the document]
nicholaspai: I'll work on this in a spearate PR

packages/liquidator/src/proxyTransactionWrapper.js
nicholaspai: I'm going to work on this in another PR

nicholaspai (commented):
Thanks for reviewing my comments, my only large comment remaining is to add tests to the LiquidationWithdrawer.sol contract

packages/liquidator/src/proxyTransactionWrapper.js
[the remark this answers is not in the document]
mrice32: Yeah, that's cool with me. Are you saying the library smart contract is at the bytecode limit? That's kinda hard to believe....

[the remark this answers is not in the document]
chrismaree: it hits a stack too deep error with any additional state vars or any return types.

mrice32 (approved):

PR #2777, opened 2021-03-26
fix(find-contract-version): Fix the perpetual bytecode hash

smb2796:
**Summary**

Updates the perpetual hash in `FindContractVersion`

**Testing**

- [ ]  New unit tests created
- [X]  Untested

**Issue(s)**

NA

### DIFF
--- a/packages/common/src/FindContractVersion.js
+++ b/packages/common/src/FindContractVersion.js
@@ -60,7 +60,7 @@ const versionMap = {
     contractType: "ExpiringMultiParty",
     contractVersion: "latest"
   },
-  "0x39dca3c82cfcba75d4355b45d546eedba21d3f9d6198324b8d951f33756a51c4": {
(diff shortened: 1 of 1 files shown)


packages/common/src/FindContractVersion.js
```diff
@@ -60,7 +60,7 @@ const versionMap = {
     contractType: "ExpiringMultiParty",
     contractVersion: "latest"
   },
-  "0x39dca3c82cfcba75d4355b45d546eedba21d3f9d6198324b8d951f33756a51c4": {
+  "0x238569485842107d2e938ff59c78841860b4dcd00d37be9859699f2c4ddbb3a0": {
```
pemulis, line 63: Do you really want to replace this hash or add an additional hash that will be recognized as `latest`, like with the EMPs above?

smb2796: Good point. Updated to not overwrite

pemulis (approved):

PR #2792, opened 2021-03-30
feat(price-feed): Proposed dVIX Price Feed Interface

ccashwell:
**Motivation**

ethVIX and iethVIX allow for the trading of model-free expected volatility of ETH. ethVIX is best understood as a long volatility position while iethVIX is a short volatility position. The ethVIX index is very similar to the Cboe VIX which tracks the expected volatility of the S&P 500. By trading ethVIX or iethVIX traders can take long or short positions against volatility, either as a speculative trading instrument itself or a hedge for other speculative trades.

Here are some examples of how traders might use ethVIX and iethVIX:

- In early March of 2020, 0xAlice believes that the global pandemic will greatly affect the price of ETH but she’s not sure if it will drop like the stock market or go up as a safe haven asset. She buys ethVIX for roughly 80 USDC per token. On Black Thursday (March 12th, 2020), the price of ETH drops significantly and the price of ethVIX moves to around 280 USDC. At expiration, 0xAlice redeems her ethVIX tokens while the price feed is still over 200 USDC. 

- In March of 2021, the price of ETH goes up quickly in a bull market. This volatility causes ethVIX to rise to roughly 120 USDC. iethVIX drops inversely to 80 USDC. 0xCarol believes that the price of ETH will soon stabilize. She buys iethVIX at roughly 90 USDC on March 12. At expiry on March 19 the price feed has moved to 97.50 USDC and she redeems her tokens for that price. 

**Summary**

We've implemented a price feed compatible with UMA's `PriceFeedInterface` as per the UMIP price feed submission guidelines.

**Details**

There are no dependency changes or implementation decisions in this PR.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -8,6 +8,7 @@ const { QuandlPriceFeed } = require("./QuandlPriceFeed");
 const { DefiPulsePriceFeed } = require("./DefiPulsePriceFeed");
 const { UniswapPriceFeed } = require("./UniswapPriceFeed");
 const { BalancerPriceFeed } = require("./BalancerPriceFeed");
+const { ETHVIXPriceFeed } = require("./EthVixPriceFeed");
 const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
 const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
 const { CoinMarketCapPriceFeed } = require("./CoinMarketCapPriceFeed");
@@ -375,6 +376,22 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       config.lastUpdateTime,
       config.priceFeedDecimals, // Defaults to 18 unless supplied. Informs how the feed should be scaled to match a DVM response.
       config.lookback
+    );
+  } else if (config.type === "ethvix") {
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating EthVixPriceFeed",
+      config
+    });
+
+    return new ETHVIXPriceFeed(
+      logger,
+      web3,
+      config.inverse,
+      networker,
+      getTime,
+      config.minTimeBetweenUpdates,
+      config.priceFeedDecimals
     );
   } else if (config.type === "invalid") {
     logger.debug({
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -1,30 +1,32 @@
 const assert = require("assert");
 const { ChainId, Token, Pair, TokenAmount } = require("@uniswap/sdk");
-const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
-const { FallBackPriceFeed } = require("./FallBackPriceFeed");
-const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
-const { ForexDailyPriceFeed } = require("./ForexDailyPriceFeed");
-const { QuandlPriceFeed } = require("./QuandlPriceFeed");
-const { DefiPulsePriceFeed } = require("./DefiPulsePriceFeed");
-const { UniswapPriceFeed } = require("./UniswapPriceFeed");
-const { BalancerPriceFeed } = require("./BalancerPriceFeed");
-const { ETHVIXPriceFeed } = require("./EthVixPriceFeed");
-const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
-const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
-const { CoinMarketCapPriceFeed } = require("./CoinMarketCapPriceFeed");
-const { CoinGeckoPriceFeed } = require("./CoinGeckoPriceFeed");
-const { TraderMadePriceFeed } = require("./TraderMadePriceFeed");
-const { PriceFeedMockScaled } = require("./PriceFeedMockScaled");
-const { InvalidPriceFeedMock } = require("./InvalidPriceFeedMock");
 const { defaultConfigs } = require("./DefaultPriceFeedConfigs");
 const { getTruffleContract } = require("@uma/core");
-const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
-const { VaultPriceFeed } = require("./VaultPriceFeed");
-const { LPPriceFeed } = require("./LPPriceFeed");
 const { BlockFinder } = require("./utils");
 const { getPrecisionForIdentifier, PublicNetworks } = require("@uma/common");
-const { FundingRateMultiplierPriceFeed } = require("./FundingRateMultiplierPriceFeed");
 const { multicallAddressMap } = require("../helpers/multicall");
+
+// Price feed interfaces (sorted alphabetically)
+const { BalancerPriceFeed } = require("./BalancerPriceFeed");
+const { BasketSpreadPriceFeed } = require("./BasketSpreadPriceFeed");
+const { CoinGeckoPriceFeed } = require("./CoinGeckoPriceFeed");
+const { CoinMarketCapPriceFeed } = require("./CoinMarketCapPriceFeed");
+const { CryptoWatchPriceFeed } = require("./CryptoWatchPriceFeed");
+const { DefiPulsePriceFeed } = require("./DefiPulsePriceFeed");
+const { DominationFinancePriceFeed } = require("./DominationFinancePriceFeed");
+const { ETHVIXPriceFeed } = require("./EthVixPriceFeed");
+const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
+const { FallBackPriceFeed } = require("./FallBackPriceFeed");
+const { ForexDailyPriceFeed } = require("./ForexDailyPriceFeed");
+const { FundingRateMultiplierPriceFeed } = require("./FundingRateMultiplierPriceFeed");
+const { InvalidPriceFeedMock } = require("./InvalidPriceFeedMock");
+const { LPPriceFeed } = require("./LPPriceFeed");
+const { MedianizerPriceFeed } = require("./MedianizerPriceFeed");
+const { PriceFeedMockScaled } = require("./PriceFeedMockScaled");
+const { QuandlPriceFeed } = require("./QuandlPriceFeed");
+const { TraderMadePriceFeed } = require("./TraderMadePriceFeed");
+const { UniswapPriceFeed } = require("./UniswapPriceFeed");
+const { VaultPriceFeed } = require("./VaultPriceFeed");
 
 // Global cache for block (promises) used by uniswap price feeds.
 const uniswapBlockCache = {};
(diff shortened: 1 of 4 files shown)


packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,114 @@
+const assert = require("assert");
+const moment = require("moment");
+const { parseFixed } = require("@uma/common");
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+
+// An implementation of PriceFeedInterface that uses the dVIX API to retrieve ethVIX prices.
+class ETHVIXPriceFeed extends PriceFeedInterface {
```
chrismaree, line 7: Would it be useful to parameterize the identifier the VIX is calculated on? For example, this feed is ETH specific. it might be useful to refactor this to have a separate param that is `ETH` or whatever you want to calculate VIX on.

This might be useful seeing you are planning on adding other vix identifiers in the future.

ccashwell: The main concern there is we're not entirely set on using this exact formula for future feeds. For instance, this feed relies on Deribit data and we feel an urgent need to move toward an entirely decentralized approach. For that reason we think keeping this explicitly the `ethVIX/iethVIX` feed rather than the `*VIX` price feed is more appropriate.

ccashwell: To add to that, we would leave it up to the community to determine whether we continue to support the feed based on this model-free approach which has more centralized reliance than we'd probably like to see in DeFi. It is possible we forever continue using this exact approach, but it's equally possible a new and more decentralized source of truth comes available.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,114 @@
...
+    //   ...
+    // ]
+
+    // 3. Get last result in the stack.
+    const mostRecent = response[response.length - 1];
+
+    // 4. Store results.
+    this.lastUpdateTime = currentTime;
+    this.historicalPrices = [...this.historicalPrices, ...response];
+    this.currentPrice = parseFixed(this.inverse ? mostRecent.iVix : mostRecent.vix, 6);
```
chrismaree, line 108: Here, you are scaling the price to 6 decimal points. Your UMIP defines the collateral decimals as 6 and this might be the reason why you are scaling this to 6. However, I think you should be scaling your price to _18_ decimals to be consistent with what the DVM would return from a price request (the price feed should mimic what the DVM returns).

The general design of price feeds and their associated scaling is to define the `priceFeedDecimals` within the price feed to match what the DVM would return in a price request. If a voter was voting on an ethVIX request the number returned would be scaled by 1e18.

To add a bit of context, financial contracts deploy a synthetic token with synthetic decimals matching the collateral's decimals. All price feeds are scaled by 1e18. For example, a synthetic deployed with USDC collateral would have 6 decimal places to match the collateral. This example synthetic would have a price scaled by 1e18.

The only time you should be scaling a price feed to something other than 18  decimals is when you are using a legacy UMA financial contract. Previously we would scale the price request to match the collateral, leaving the synthetic at 18 decimals.

packages/financial-templates-lib/test/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,174 @@
...
+        assert.equal(inversePriceFeed.getCurrentPrice().toString(), "142440000");
+      });
+    });
+
+    describe("Historical price discovery", () => {
+      it("can return the historical ethVIX price at a specific time", async () => {
+        const { timestamp, vix } = historicalResponse[1];
+        await priceFeed.update();
+        const historicalPrice = await priceFeed.getHistoricalPrice(moment.utc(timestamp).valueOf());
+        assert.equal(historicalPrice.toString(), parseFixed(vix, 6).toString());
```
chrismaree, line 73: all these cases where you are asserting values on the price feed would need to be changed to `1e18` scalled numbers. (i.e `toWei` from web3)

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
chrismaree: See here: https://github.com/UMAprotocol/protocol/blob/0eea3fbb610f74694c22ca36f6902faf3fa9092b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L60

for an example on how we do this in other feeds.

chrismaree (commented):
This feed is pretty great! good implementation, clean comments and tests. I have an issue with scaling that I raised within the review.

TO get this feed to work correctly with the reference UMA bots you will need to add some code to the `CreatePriceFeed` file. you can find it here: https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js

Your feed is most similar to the `DefiPulsePriceFeed` or the `DominationFinancePriceFeed`. I would recommend looking at their implementation of the `CreatePriceFeed.js` for reference. You should also add tests for this change the tests for `CreatePriceFeed`.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
ccashwell: This is understandable. Tbh it was a bit confusing in the docs why the precision was expected at 18 decimals, but it seems reasonable given this context. I'll make the changes shortly and push the updated branch for review.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
pemulis: I would suggest that the UMIP could be updated if the dVIX community chooses a different method for calculating VIX for ETH or for other assets.

This would be similar, in my opinion, to how UMIPs can be updated to reference different exchanges to get asset prices depending on changes in exchange volume or reliability. It's just more complicated, since the methodology changes may be substantially more than, "look at these other exchanges instead, and maybe take a median of three exchanges instead of a Uniswap TWAP," or whatever.

I don't think it's critical for this initial UMIP, but worth considering for a future UMIP that could add support for multiple asset VIX's through parameterization, especially as you evaluate how this methodology performs!

That future UMIP could include some description of how the methodology could change if the dVIX community decides there's a better way to calculate VIX for an asset, and that what you're relying on when using this price identifier in the general case is the dVIX community's judgement of how to calculate VIX, as modified by the DVM's judgement if someone thinks the dVIX community has gone rogue in some way.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,114 @@
...
+    //     "timestamp": "2021-03-24T15:00:00.000Z",
+    //     "iVix": 142.4470728085134,
+    //     "vix": 70.20151276427174,
+    //     ...
+    //   },
+    //   ...
+    // ]
+
+    // 3. Get last result in the stack.
+    const mostRecent = response[response.length - 1];
```
mrice32, line 103: Is it guaranteed that the response is sorted? Should you call `.sort` on the array to ensure that the response timestamps are increasing?

mrice32 (commented):
Looks good once @chrismaree's comments are resolved!

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
ccashwell: I think these are good points, perhaps the best thing to do is stick with this model for now and update the UMIP later if there are more true VIX-style indexes published. Can we defer this to a later update?

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
ccashwell: It is guaranteed on the API side but it's worth doing here for posterity.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.js
pemulis: For sure, I think that makes sense for now.

ccashwell:
I think we're all set here for another look @chrismaree

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -376,6 +379,22 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
+    logger.debug({
+      at: "createPriceFeed",
+      message: "Creating EthVixPriceFeed",
+      config
+    });
+
+    return new ETHVIXPriceFeed(
+      logger,
+      web3,
+      config.inverse,
```
nicholaspai, line 392: Is `inverse` a required param?

ccashwell: No, because if not provided it would be the long variant (ethVIX) by default. However I could see an argument for making it a required boolean field. What's your opinion?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -1,30 +1,33 @@
...
 const { getTruffleContract } = require("@uma/core");
-const { ExpressionPriceFeed, math, escapeSpecialCharacters } = require("./ExpressionPriceFeed");
-const { VaultPriceFeed } = require("./VaultPriceFeed");
-const { LPPriceFeed } = require("./LPPriceFeed");
 const { BlockFinder } = require("./utils");
 const { getPrecisionForIdentifier, PublicNetworks } = require("@uma/common");
-const { FundingRateMultiplierPriceFeed } = require("./FundingRateMultiplierPriceFeed");
 const { multicallAddressMap } = require("../helpers/multicall");
 
+// Price feed interfaces (sorted alphabetically)
```
chrismaree, line 9: nice.

packages/financial-templates-lib/test/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,185 @@
...
+
+  describe("After an initial update has been performed", () => {
+    beforeEach(async () => {
+      networker.getJsonReturns = new Array(3).fill(historicalResponse);
+    });
+
+    describe("Price discovery", () => {
+      it("can return the latest ethVIX price", async () => {
+        await priceFeed.update();
+        assert.equal(priceFeed.getCurrentPrice().toString(), "70200000000000000000");
```
chrismaree, line 59: nit: you might consider changing this to `web3.uti.toWei("70.2")`

packages/financial-templates-lib/test/price-feed/EthVixPriceFeed.js
```diff
@@ -0,0 +1,185 @@
...
+      });
+
+      it("can return the latest iethVIX price", async () => {
+        await inversePriceFeed.update();
+        assert.equal(inversePriceFeed.getCurrentPrice().toString(), "142440000000000000000");
+      });
```
chrismaree, line 65: same nit. please use `toWei`

chrismaree (approved):
this looks good to me! Thank you for implementing the scaling changes. I think we are pretty much good to merge once we get sign off from one other @UMAprotocol/eng 

I've initiated a run of your unit tests in circle CI as well.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
ccashwell: Yeah tbh this whole file is getting out of hand, this was the minimum viable improvement and anything more felt like an overstep 😅

pemulis (approved):

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
chrismaree: ye. we do need to refactor much of this. it's a bit of a mess right now.

chrismaree:
looks like you need to re-merge master into your branch to get your tests to pass correctly.

PR #2804, opened 2021-03-31
feat(run-transaction-helper): Move ynatm functionality into runTransaction helper

nicholaspai:

**Motivation**

The `optimistic-oracle-proposer` and `funding-rate-proposer` both use the `runTransaction` helper to simulate and send transactions, while the `liquidator` and `disputer` implement their own transaction lifecycle management logic and integrate the `ynatm` package which adds retry-gas-price-bump logic.

This PR adds the `ynatm` logic into the `runTransaction` helper and upgrades the `liquidator` and `disputer` to use `runTransaction`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/disputer/src/disputer.js
+++ b/packages/disputer/src/disputer.js
@@ -1,7 +1,7 @@
 const {
   PostWithdrawLiquidationRewardsStatusTranslations,
-  revertWrapper,
-  createObjectFromDefaultProps
+  createObjectFromDefaultProps,
+  runTransaction
 } = require("@uma/common");
 
 class Disputer {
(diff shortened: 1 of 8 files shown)


packages/disputer/src/disputer.js
```diff
@@ -175,65 +175,51 @@ class Disputer {
...
-      let receipt;
       try {
-        receipt = await dispute.send(txnConfig);
+        // Get successful transaction receipt and return value or error.
+        const transactionResult = await runTransaction({
+          transaction: dispute,
+          config: {
+            gasPrice: this.gasEstimator.getCurrentFastPrice(),
+            from: this.account,
+            nonce: await this.web3.eth.getTransactionCount(this.account)
```
mrice32, line 190: why are we explicitly providing the nonce here? Isn't this generally injected automatically or is this required by the ynatm?

mrice32 (approved):
LGTM! One minor question around the nonce thing

packages/disputer/src/disputer.js
nicholaspai: The nonce is passed through from the config passed into `ynatm` so I'm pretty sure `ynatm` ONLY works if you hardcode the `nonce` to prevent it from auto increasing between retries

nicholaspai: This is how the tests in `ynatm` use `nonce` + our liquidator implementation

chrismaree (approved):
This LGTM! Ideally I think we should leave the liquidation changes out until PR #2775 has been merged, as discussed IRL.

PR #2808, opened 2021-04-01
feat(action-wrappers): Add a simple contract and script to enable DSProxy token redemtion

chrismaree:
**Motivation**

This PR adds a script that enables a DSProxy to redeem tokens from a contract. This should be used after a swap/mint/liquidate transaction to close out a bots position.

The `WithdrawTokensFromDSProxy.ts` was also modified slightly after testing on mainnet.


nicholaspai (approved):

packages/trader/scripts/RedeemTokensWithDSProxy.ts
```diff
@@ -0,0 +1,79 @@
...
+  await dsProxyManager.initializeDSProxy(argv.dsProxyAddress, false);
+
+  const dsProxyAddress = dsProxyManager.getDSProxyAddress();
+  if (!dsProxyAddress) throw new Error("DSProxy Address was not found or parameterized");
+
+  // Create the callData as the encoded tx against the TokenRedeemer contract.
+  const TokenRedeemer = getTruffleContract("TokenRedeemer", web3);
+  const TokenRedeemerInstance = new web3.eth.Contract(TokenRedeemer.abi);
+  const callData = TokenRedeemerInstance.methods
+    .redeem(argv.financialContractAddress, { rawValue: argv.numTokens })
```
nicholaspai, line 55: Do you think its worth making user specify `numTokens` as an already to-wei'd value? For example in stead of passing in `10000000000000000000`, they can pass in `10` and we just perform `parseFixed(numTokens, syntheticTokenDecimals)`?

PR #2816, opened 2021-04-02
feat(liquidator-dsproxymanager) add run transaction to liquidator and DSProxy Manager

chrismaree:
**Motivation**

As part of updating the liquidator bot to use DSProxy for liquidations, we need to upgrade the bot to use the `runTransaction` module for sending transactions. This PR adds this logic to the liquidator and importantly the DSProxy manger. 

**Summary**

Updated the bot and the DSProxy manager to use `runTransaction`

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -139,16 +146,20 @@ class DSProxyManager {
...
     await this.gasEstimator.update();
-    const executeTransaction = await this.dsProxy.methods["execute(bytes,bytes)"](callCode, callData).send({
-      from: this.account,
-      gasPrice: this.gasEstimator.getCurrentFastPrice()
+    const executeTransaction = await runTransaction({
+      transaction: this.dsProxy.methods["execute(bytes,bytes)"](callCode, callData),
+      config: {
+        gasPrice: this.gasEstimator.getCurrentFastPrice(),
+        from: this.account,
+        nonce: await this.web3.eth.getTransactionCount(this.account)
```
nicholaspai, line 154: None of these `runTransaction` methods are getting called in parallel with each other right? Otherwise the `nonce` is going to be problematic. It doesn't look like an issue but just want to double check

nicholaspai (approved):
Looks great! Just want you to confirm that none of the `runTransactions` with nonce's passed in are getting called in parallel. I had to remove this in the `funding-rate-proposer` [for example](https://github.com/UMAprotocol/protocol/blob/master/packages/funding-rate-proposer/src/proposer.js#L238)

mrice32 (approved):

PR #2817, opened 2021-04-02
fix: call update on gas estimator before sending transactions

mrice32:
**Motivation**

Minor trader fix to ensure the gas price estimation is updated before sending transactions.

**Summary**

Just added a few update calls.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
+++ b/packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
@@ -90,6 +90,7 @@ class DSProxyManager {
         message: "No DSProxy found for EOA. Deploying new DSProxy",
         account: this.account
       });
+      await this.gasEstimator.update();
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):
Can we call this in the file that constructs the `DSProxyManager` and passes in the gasEstimator? This way we don't have to remember to `update` it before each method

mrice32:
> Can we call this in the file that constructs the `DSProxyManager` and passes in the gasEstimator? This way we don't have to remember to `update` it before each method 

We could, but then it never gets re-updated in a looping bot, right?

mrice32:
@nicholaspai this is somewhat moot anyway, because @chrismaree is going to overhaul the transaction management here anyway.

mrice32:
Going to go ahead and merge so we can get the bot up and running. Happy to change in a follow up!

PR #2818, opened 2021-04-02
fix(trader): pass along range trader config from env variables

mrice32:
**Motivation**

There is no way to provide the range trader config via env variables.

**Summary**

Adds the range trader config to the index.ts file and passes it along.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


nicholaspai (approved):
Nice catch. How did this pass tests before? Or because we hadn't tested `index.ts` properly?

mrice32:
> Nice catch. How did this pass tests before? Or because we hadn't tested `index.ts` properly?

Not sure. I'm guessing we don't test env variables.

PR #2822, opened 2021-04-04
fix(forex-daily-pricefeed): Fix endpoint

nicholaspai:
**Motivation**

[It appears that api.exchangerates.io was acquired recently and put behind a paywall](https://github.com/exchangeratesapi/exchangeratesapi/issues/117) so this PR replaces the endpoint with a recommended stable drop-in replacement (using the same ECB daily forex data) [exchangerate.host](https://exchangerate.host/#/#faq).

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -140,7 +140,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       uniswapBlockCache
     );
   } else if (config.type === "forexdaily") {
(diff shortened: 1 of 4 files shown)


packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -140,7 +140,7 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
       uniswapBlockCache
     );
   } else if (config.type === "forexdaily") {
-    const requiredFields = ["base", "symbol", "lookback", "forexDailyApiKey"];
```
nicholaspai, line 143: No longer need this key

pemulis (approved):

PR #2824, opened 2021-04-04
improve(funding-rate-proposer): Publishes pending funding rate when it has diverged significantly from pricefeed's current rate

nicholaspai:
**Motivation**

Currently the `funding-rate-proposer` does not try to publish funding rate proposals when there is a pending proposal, even if that pending proposal has expired its proposal liveness and is stale. When sponsor activity is light, the `funding-rate-proposer` risks blocking the update for a stale funding rate because `applyFundingRate` is not called. This becomes a problem when the pending funding rate diverges meaningfully from the funding-rate-identifier pricefeed's current price. This PR adds the `multicall` logic to the proposer bot so that the bot can detect when a proposal has expired. 

If after simulating `applyFundingRate`, the `proposalTime != 0`, then the pending proposal hasn't yet expired. If the `proposalTime == 0` but the pricefeed's price is too close to the publishable funding rate, then the bot won't do anything. Only if the proposal has expired AND it is significantly different from the pricefeed rate, will the bot propose a new rate (and simultaneously publish the pending rate). 

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/funding-rate-proposer/index.js
```diff
@@ -8,6 +8,7 @@ const {
   waitForLogger,
   delay,
   FinancialContractFactoryClient,
+  OptimisticOracleClient,
```
chrismaree, line 11: so now the `funding-rate-proposer` and the `optimisic-oracle` both import the `OptimisticOracleClient` and use it in somewhat diffrent ways. I was originally going to say that I think that this should stay within the domain of the `optimistic-oracle` but after more thought it does make sence to keep it here within the `funding-rate-proposer` as this bot's job is to monitor the funding rates of perps, not just OO enquired prices. 

Despite this small mixing of concerns, I think it makes sneak as a pattern.

Not really anything actionable here, just something I thought I'd raise. Thoughts?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -125,6 +133,18 @@ class FundingRateProposer {
     });
   }
 
+  async applyFundingRates() {
+    this.logger.debug({
+      at: "PerpetualProposer#applyFundingRates",
```
chrismaree, line 138: I feel like we should create a task to go back through all our old log messages and make sure this `#at` syntax is correctly applied. It really helps with debugging

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -278,12 +275,104 @@ class FundingRateProposer {
...
-        proposedRate: offchainFundingRate,
+        proposedRate: pricefeedPrice,
         currentRate: onchainFundingRate,
         allowedError: this.fundingRateErrorPercent
       });
     }
   }
+
+  // If there is a pending funding rate that can be published, AND the current offchain price differs from the pending
+  // rate more than some threshold, than publish the pending rate
```
chrismaree, line 286:
```suggestion
  // rate more than some threshold, than publish the pending rate.
```

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -278,12 +275,104 @@ class FundingRateProposer {
...
+  async _applyFundingRate(contractAddress) {
+    const cachedContract = this.contractCache[contractAddress];
+    const currentFundingRateData = cachedContract.state.currentFundingRateData;
+    const fundingRateIdentifier = this.hexToUtf8(currentFundingRateData.identifier);
+
+    // Get `getSettleableProposals` in which the proposer was this bot account. Return early if there are no
+    // pending proposals.
+    const publishableProposals = this.optimisticOracleClient.getSettleableProposals(this.account).filter(proposal => {
```
chrismaree, line 294: it only considers rates that are proposed by _this_ bot? can you comment on this?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -278,12 +275,104 @@ class FundingRateProposer {
...
+    const pricefeedPrice = await this._getPriceFeedAndPrice(fundingRateIdentifier);
+    if (!pricefeedPrice) return;
+    let shouldUpdateFundingRate = isDeviationOutsideErrorMargin(
+      this.toBN(publishableFundingRate), // ObservedValue
+      this.toBN(pricefeedPrice), // ExpectedValue
+      this.toBN(this.toWei("1")),
+      this.toBN(this.toWei(this.fundingRateErrorPercent.toString()))
+    );
+    if (shouldUpdateFundingRate) {
+      const apply = cachedContract.contract.methods.applyFundingRate();
```
chrismaree, line 318: nit: this is not a great var name

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -278,12 +275,104 @@ class FundingRateProposer {
...
+        return;
+      }
+    } else {
+      this.logger.debug({
+        at: "PerpetualProposer#applyFundingRate",
+        message: "Skipping apply because pending rate is within allowed margin of error",
+        fundingRateIdentifier,
+        pricefeedPrice: pricefeedPrice,
+        pendingRateToPublish: publishableFundingRate,
+        allowedError: this.fundingRateErrorPercent
```
chrismaree, line 371: does this need to be `.toString`ed? same above on line 325.

chrismaree (approved):
this is pretty awesome. good work @nicholaspai . I had a few small nits/comments/questions but otherwise this LGTM!

packages/funding-rate-proposer/index.js
nicholaspai: I think this highlights the problems with the `update() --> yourMethod()` pattern, where `update()` fetches unneccessary data for ALL OO requests/proposals/disputes whereas this bot only needs data coming from this specific bot.

However, it is nice that this bot doesn't have to determine which proposals have expired, instead it can make a simple `getSettleableProposals()` call to the OO client.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: yeah, you mean using the `#` should be more enforced?

packages/funding-rate-proposer/src/proposer.js
nicholaspai: I can generalize this actually, it is currently set like this because `getSettleableProposals` expects an `account` to be passed. I'll change the OO client to allow the caller to get proposals for ALL accounts

packages/funding-rate-proposer/src/proposer.js
nicholaspai: What would you propose changing it to? I feel that `shouldUpdateFundingRate` directly explains what its set to: if the deviation is outside the error threshold, the pending funding rate should be published. Is `shouldPublishFundingRate` better?

packages/funding-rate-proposer/src/proposer.js
nicholaspai: No its a `Number`

packages/funding-rate-proposer/index.js
chrismaree: ye, exactly. this is a pretty systemic issue with how we've built the client/implementor pattern. It would require a pretty drastic refactor to sort out and for now I think its' fine. Queries to Infura are cheap!

packages/funding-rate-proposer/src/proposer.js
chrismaree: yup, exactly.

packages/funding-rate-proposer/src/proposer.js
chrismaree: I see. that's probs fine. it's just that this might need to be called on proposals that were made by another bot, right? ideally it should be generalised but this can probs be put in the backlog for later.

packages/funding-rate-proposer/src/proposer.js
chrismaree: I was complaining about the `apply` var name. perhaps `applyFundingRateTransaction` or something?

packages/funding-rate-proposer/src/proposer.js
chrismaree: O I see. in the other places where this is used with `.toString()` it is being converted to a big number. all good in that case 👍

packages/funding-rate-proposer/src/proposer.js
nicholaspai: This is no longer relevant as I've refactored this PR to simulate the `applyFundingRate` method

nicholaspai:
@mrice32 @chrismaree based on our offline discussion about this PR implementation, I've changed the bot to use the `multicall` contract to simulate `applyFundingRate` before loading the contract's `fundingRate` data. This way the bot will be able to propose new funding rates even when the pending funding rate has expired, AS LONG AS the pending funding rate is different enough from the pricefeed's rate.

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -153,57 +157,34 @@ class FundingRateProposer {
...
-    // Set `offchainFundingRate` to correct precision by converting back and forth between the pricefeed's output
-    // precision:
-    offchainFundingRate = parseFixed(
-      Number(formatFixed(offchainFundingRate.toString(), priceFeed.getPriceFeedDecimals()))
-        .toFixed(this.precision)
-        .toString(),
-      priceFeed.getPriceFeedDecimals()
-    ).toString();
+    const pricefeedPrice = await this._getPriceFeedAndPrice(fundingRateIdentifier);
+    if (!pricefeedPrice) return;
```
mrice32, line 161: Why did we drop the error here?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -375,11 +357,35 @@ class FundingRateProposer {
...
-    ]);
+    let fundingRateData, configStoreAddress;
+    if (!this.multicallContractAddress) {
+      // If no multicontract address set, then read the contract's on-chain funding rate data. Note that this will not
+      // take into account any pending funding rates that will be published on the next contract interaction.
+      [fundingRateData, configStoreAddress] = await Promise.all([
+        perpetualContract.methods.fundingRate().call(),
+        perpetualContract.methods.configStore().call()
+      ]);
+    } else {
```
mrice32, line 368: Why can't we just assume that we always have a multicall address or explicitly require it?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -375,11 +357,35 @@ class FundingRateProposer {
...
+        callData: perpetualContract.methods.fundingRate().encodeABI()
+      };
+      const [outputs, _configStoreAddress] = await Promise.all([
+        aggregateTransactionsAndCall(this.multicallContractAddress, this.web3, [applyFundingRateCall, fundingRateCall]),
+        perpetualContract.methods.configStore().call()
+      ]);
+      configStoreAddress = _configStoreAddress;
+      // `returnData` is an array of decoded return data bytes corresponding to the transactions passed to
+      // the multicall aggregate method. Therefore, `fundingRate()`'s return output is the second element.
+      fundingRateData = outputs[1];
```
mrice32, line 386: nit: why not
```suggestion
     [[,fundingRateData], configStoreAddress] = await Promise.all([
        aggregateTransactionsAndCall(this.multicallContractAddress, this.web3, [applyFundingRateCall, fundingRateCall]),
        perpetualContract.methods.configStore().call()
      ]);
```

mrice32 (commented):
Looks good -- minor questions and comments.

packages/funding-rate-proposer/src/proposer.js
nicholaspai: I refactored the error into an internal helper function

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Oh I like asserting that `multicallContractAddress` is passed via construction I'll add

packages/financial-templates-lib/src/helpers/multicall.js
```diff
@@ -30,6 +29,24 @@ const aggregateTransactionsAndCall = async (multicallAddress, web3, transactions
   return returnData.map((data, i) => _decodeOutput(transactions[i].callData, data, web3));
 };
 
+const multicallAddressMap = {
```
nicholaspai, line 32: Copied the JSON from the npm package so that we can remove the npm dependency

mrice32 (approved):
LGTM!

PR #2826, opened 2021-04-05
feat(optimistic-oracle-proposer): Blacklist identifiers to skip post-expiry

nicholaspai:

**Motivation**

`uSTONKS` expires at the end of April and `uGAS` expires at the end of May. They are self-referential contracts that are using the latest version of the EMP. The bots currently are going to provide a price at expiry of the 2-hour TWAP instead of the final closing index value of uSTONKS. We need the bots to blacklist these two contracts so they do not propose a price. We assume that a manual price will occur instead.

The price request alert should still be emitted by the `optimistic-oracle-monitor` (i.e. the blacklist resides in the proposer bot, not the client shared between bot and monitor).

**Summary**

To check if an identifier is coming from an EMP post-expiry, we attempt to construct an EMP using the price request's `requester` address, which is the expired EMP address if the price request was an expiry one.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/PriceIdentifierUtils.js
+++ b/packages/common/src/PriceIdentifierUtils.js
@@ -25,8 +25,21 @@ const getPrecisionForIdentifier = identifier => {
   return IDENTIFIER_NON_18_PRECISION[identifier] ? IDENTIFIER_NON_18_PRECISION[identifier] : 18;
 };
 
+// The optimistic oracle proposer should skip proposing prices for these identifiers, for expired EMP contracts,
+// because they map to self-referential pricefeeds pre-expiry, but have different price resolution ogic post-expiry.
+// For example, please see [UMIP47](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md):
+// - "The type of price that the DVM will return is dependent on the timestamp the price request is made at. This
+//   timestamp is the expiry timestamp of the contract that is intended to use this price identifier, so the TWAP
+//   calculation is used pre-expiry and the closing index value of uSTONKS calculation is used at expiry.""
+const OPTIMISTIC_ORACLE_IGNORE_POST_EXPIRY = [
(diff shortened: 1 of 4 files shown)


packages/common/src/PriceIdentifierUtils.js
```diff
@@ -25,8 +25,21 @@ const getPrecisionForIdentifier = identifier => {
   return IDENTIFIER_NON_18_PRECISION[identifier] ? IDENTIFIER_NON_18_PRECISION[identifier] : 18;
 };
 
+// The optimistic oracle proposer should skip proposing prices for these identifiers, for expired EMP contracts,
+// because they map to self-referential pricefeeds pre-expiry, but have different price resolution ogic post-expiry.
+// For example, please see [UMIP47](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-47.md):
+// - "The type of price that the DVM will return is dependent on the timestamp the price request is made at. This
+//   timestamp is the expiry timestamp of the contract that is intended to use this price identifier, so the TWAP
+//   calculation is used pre-expiry and the closing index value of uSTONKS calculation is used at expiry.""
+const OPTIMISTIC_ORACLE_IGNORE_POST_EXPIRY = [
```
nicholaspai, line 34: @smb2796 can you confirm these are the identifiers we want to blacklist?

smb2796: Yep these are the only identifiers needed for now. They are used in [uSTONKS](https://etherscan.io/address/0x4F1424Cef6AcE40c0ae4fc64d74B734f1eAF153C) contract and the [uGAS-JUN21](https://etherscan.io/address/0x4E8d60A785c2636A63c5Bd47C7050d21266c8B43) contract.

mrice32 (approved):
LGTM!

chrismaree (approved):
nice! simple, clean code

PR #2828, opened 2021-04-05
improve(financial-contracts): Clean up comments

nicholaspai:

**Motivation**

Addressing some typographical errors:

- Line 77 says "an collateral token" instead of "a collateral token"
- Line 86 says "cover call" instead of "covered call"
- Misleading comment in Line 454 of the Liquidatable contract that states the sponsor and disputer reward percentages cannot (cumulatively) exceed zero but the actual restriction is that they cannot exceed 1.

### DIFF
--- a/packages/common/src/FindContractVersion.js
+++ b/packages/common/src/FindContractVersion.js
@@ -50,7 +50,7 @@ const versionMap = {
     contractType: "ExpiringMultiParty",
     contractVersion: "1.2.2"
   },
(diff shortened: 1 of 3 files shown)


pemulis (approved):
Looks good, but from the failing tests, it looks like `FindContractVersion` has got us again. cc @chrismaree

mrice32 (approved):
LGTM!

chrismaree (approved):

PR #2837, opened 2021-04-06
feat(call): allow anyone to set strike price in covered call library

pemulis:

**Motivation**

Updates financial product library to allow anyone to set the strike price for a newly deployed covered call. Previously, only the deployer of the _library_ contract can set the strike price for a covered call contract using the library, which limits the reusability of the library, as each deployer would need to deploy their own library.

There is a limited risk of an uneconomical griefing attack, where someone (for some reason) sets the strike price of a covered call contract before the deployer sets the strike price. There is no incentive to do so, but if it became a problem for some reason, deployment of the covered call and setting of the strike price could be done via an atomic transaction.

A different risk would be an improper deployment order of operations, where the deployer mints call options and uses them to seed an AMM pool before setting a strike price in the contract. That could create an economic incentive for someone else to set an incorrect strike price. (This risk is similar, conceptually, to the risk of misconfiguration of EMP parameters, like a too-low collateralization ratio.)

Documentation for covered call deployment should emphasize that the strike price should be set after deploying the covered call contract and before minting any options.

**Summary**

Removes `Ownable` and the `onlyOwner` modifier in the set strike price method.

**Details**

n/a

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/CoveredCallFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/CoveredCallFinancialProductLibrary.sol
@@ -1,7 +1,6 @@
 pragma solidity ^0.6.0;
 pragma experimental ABIEncoderV2;
 import "./FinancialProductLibrary.sol";
-import "@openzeppelin/contracts/access/Ownable.sol";
 import "../../../common/implementation/Lockable.sol";
 
 /**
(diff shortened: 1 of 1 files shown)


chrismaree (approved):

packages/core/contracts/financial-templates/common/financial-product-libraries/CoveredCallFinancialProductLibrary.sol
```diff
@@ -28,7 +27,6 @@ contract CoveredCallFinancialProductLibrary is FinancialProductLibrary, Ownable,
      */
     function setFinancialProductStrike(address financialProduct, FixedPoint.Unsigned memory strikePrice)
         public
```
nicholaspai, line 29: Can you change the `Note` which now improperly references how only the owner can set the strike? Also, if you can add a note like you did in the description about the recommended usage, that'd be great: i.e. caller should set strike after deployment but before depositing the synthetic token into a liquidity pool

nicholaspai (approved):

PR #2843, opened 2021-04-07
feat(gas-rebate): Enable 2Key wallet overrides

nicholaspai:
Signed-off-by: Nick Pai <nick@umaproject.org>

**Motivation**

Allow user to define a 2KEY_ADDRESS_OVERRIDE.json file to replace designated wallet recipients with hot wallets for gas rebate scripts

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/affiliates/gas-rebate/VoterGasRebate.js
+++ b/packages/affiliates/gas-rebate/VoterGasRebate.js
@@ -35,6 +35,9 @@ const argv = require("minimist")(process.argv.slice(), {
 const fs = require("fs");
 const path = require("path");
 
+// Locally stored list of DesignatedVoting wallets to substitute with their hot wallets:
+const addressesToReplace = require("./2KEY_ADDRESS_OVERRIDE") || {};
+
 const FindBlockAtTimestamp = require("../liquidity-mining/FindBlockAtTimeStamp");
 const { getAbi, getAddress } = require("@uma/core");
 const { getWeb3 } = require("@uma/common");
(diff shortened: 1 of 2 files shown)


packages/affiliates/gas-rebate/VoterGasRebate.js
```diff
@@ -35,6 +35,16 @@ const argv = require("minimist")(process.argv.slice(), {
...
 
+// Locally stored list of DesignatedVoting wallets to substitute with their hot wallets:
+let addressesToReplace = {};
+try {
+  if (fs.existsSync("./2KEY_ADDRESS_OVERRIDE.json")) {
+    addressesToReplace = require("./2KEY_ADDRESS_OVERRIDE.json");
+  }
+} catch (err) {
+  // Do nothing if file is not provided.
+}
```
mrice32, line 46: Why would this error if you're checking that the file exists?

packages/affiliates/gas-rebate/VoterGasRebate.js
```diff
@@ -35,6 +35,16 @@ const argv = require("minimist")(process.argv.slice(), {
 const fs = require("fs");
 const path = require("path");
 
+// Locally stored list of DesignatedVoting wallets to substitute with their hot wallets:
+let addressesToReplace = {};
```
mrice32, line 39: What is the layout of this mapping? address => ?

mrice32 (approved):
Looks great! Two minor comments.

packages/affiliates/gas-rebate/VoterGasRebate.js
nicholaspai: Good point, removed

packages/affiliates/gas-rebate/VoterGasRebate.js
nicholaspai: Its just address => bool. If I make it an array, then exporting it is a bit uglier (require("addressesToReplace).addressArray). I'm not strongly opinionated. WDYT is the best format?

mrice32: I think an array would be slightly more intuitive. Can the top level in the JSON file be an array?

[
  "0x1234",
  "0x1234"
]

nicholaspai: Oh wow didn't realize you could export an array in a json file. Done

PR #2851, opened 2021-04-12
feat(bot-strategy-runner): add generalised bot runner

chrismaree:
**Motivation**

As the number of bots within the ecosystem grows so we need more scalable techniques for running bots. This PR adds a generalised bot runner that can pull addresses from a whitelist (such as the dev mining white list) and use this to execute a series of bots.

**Summary**

The design implemented in this PR is meant to accommodate community members running bots at home or on a VPS as well as UMA's serverless bot running infrastructure. 

**Details**

This implementation contains a few key components:

1) `ConfigBuilder.ts` constructs configs to execute different kinds of bots (liquidator, disputer and monitor currently supported). Contains some helper functions for building white lists and fetching remote config files.
2) `LogCaptureTransport.ts` a custom Winston transport to capture logs produced by the strategy runner.
3) `BotEntryWrapper.ts` but runner that wraps the bots entry point, enabling generalized bot execution.
4) `index.ts` file containing all the logic to pull all the parts together and all concurrency logic.

**For details on how to run the strategy runner (and other descriptions of the configs) please see the README.md file contained in this PR.**

See [this](https://imgur.com/ZOSwlf0) gif for what the progress bar looks like 👯 

**Testing**

- [ ]  New unit tests created
- [X]  Untested


packages/bot-strategy-runner/README.md
```diff
@@ -0,0 +1,96 @@
+# @uma/bot-strategy-runner
```
chrismaree, line 1: this looks really nasty like this in preview. take a look at the file [here](https://github.com/UMAprotocol/protocol/blob/51884a2382f26856f9d9beb31af5351897362eaa/packages/bot-strategy-runner/README.md) for something a bit easier to read.

packages/bot-strategy-runner/ExampleConfig.json
```diff
@@ -0,0 +1,14 @@
+{
```
chrismaree, line 1: sample config that will pull directly from the global whitelist.

chrismaree:
@UMAprotocol/eng this package does not yet have any unit tests. Thus far I've been manually testing this against all main net bots.

.eslintrc.js
```diff
@@ -27,7 +27,8 @@ module.exports = {
         "plugin:@typescript-eslint/recommended"
       ],
       rules: {
-        "@typescript-eslint/no-var-requires": 0
+        "@typescript-eslint/no-var-requires": 0,
+        "@typescript-eslint/no-empty-function": 0
```
mrice32, line 31: Do we need to remove these rules globally or is it possible to just add an inline comment to ignore the rule where we violate it?

packages/bot-strategy-runner/README.md
```diff
@@ -0,0 +1,96 @@
...
+  "globalAddressWhitelist": [
+    "0x0f4e2a456aAfc0068a0718E3107B88d2e8f2bfEF",
+    "0xd9af2d7E4cF86aAfBCf688a47Bd6b95Da9F7c838",
+    "0xd60139B287De1408f8388f5f57fC114Fb4B03328"
+  ],
+  "commonConfig": {
+    "priceFeedConfig": {
+      "cryptowatchApiKey": "YOUR-CRYPTO-WATCH-API-KEY",
+    },
+  "liquidatorSettings": { "enableBotType": true },
```
mrice32, line 34: I think the preview is broken because there's a typo in the json. I think this should fix it.

```suggestion
    }
  },
  "liquidatorSettings": { "enableBotType": true },
```

packages/bot-strategy-runner/ExampleConfig.json
```diff
@@ -0,0 +1,14 @@
...
+  "globalAddressWhitelistUrls": [
+    "https://raw.githubusercontent.com/UMAprotocol/protocol/master/packages/affiliates/payouts/devmining-status.json"
+  ],
+  "globalAddressBlacklist": [
+    "0x516f595978D87B67401DaB7AfD8555c3d28a3af4",
```
nicholaspai, line 6: What are these addresses? Would be helpful as an example to explain whether they are EMP's or something else

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+import * as cliProgress from "cli-progress";
+import * as bluebird from "bluebird";
+
+import { Logger, delay } from "@uma/financial-templates-lib";
+
+import { strategyRunnerConfig, buildBotConfigs, buildGlobalWhitelist, mergeConfig } from "./ConfigBuilder";
+import { runBot } from "./BotEntryWrapper";
+
+// Defaults for bot execution. If the user does not define these then these settings will be used.
+const defaultPollingDelay = 120;
```
nicholaspai, line 14: Why not 5 mins?

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+const defaultStrategyTimeout = 60;
```
nicholaspai, line 15: Should this timeout be 2 mins? I feel like 60 mins is too short, while anything over 2 mins is definitely an error. Empirically speaking.

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+  for (;;) {
+    Logger.debug({
+      at: "BotStrategyRunner",
+      message: "Executing set of bots concurrently",
+      concurrency: strategyRunnerConfig.botConcurrency
+    });
+
+    progressBar.start(allBotsConfigs.length, 0);
+
+    // Execute all bots in a bluebird map with limited concurrency. Note than none of the `runBot` calls ever throw errors.
```
nicholaspai, line 60:
```suggestion
    // Execute all bots in a bluebird map with limited concurrency. Note that none of the `runBot` calls ever throw errors.
```

nicholaspai (commented):
Since these scripts are going to run bots in prod, similar to those in `serverless-orchestration`, should we have unit tests? I can see us wanting to test any of the scripts in `bot-strategy-runner/src` for example.

.eslintrc.js
chrismaree: I tried adding the inline rule for a few hours but could not get it to work as expected unfortunately. I agree with you that this is not ideal to have within the global rules. I will try again with the inline stuff and report back.

packages/bot-strategy-runner/ExampleConfig.json
chrismaree: These ones are either expired or we dont have price feeds for them. There are some contracts we run dev mining on that we dont run feeds on. that's what these addresses are for.

How do you recommend I add comments? json does not support comments.

packages/bot-strategy-runner/src/index.ts
chrismaree: can change it to that to be more consistant with the hub! I set it to this as the default polling delay within the bots is 2 mins.

packages/bot-strategy-runner/src/index.ts
chrismaree: ye, this was mainly to match to the bots implementation. I'm hoary to make this 2 miss and then set the polling delay in to 5 mins. will be easier to reason about I think

chrismaree:
> Since these scripts are going to run bots in prod, similar to those in `serverless-orchestration`, should we have unit tests? I can see us wanting to test any of the scripts in `bot-strategy-runner/src` for example.

sure. I can add unit tests for it but perhaps we can leave it for a separate PR as this is already pretty bloated and a lot of code. I dont think we will roll this into production just yet so it'll be ok to have untested code in master for a few days.

packages/bot-strategy-runner/src/index.ts
chrismaree: I've updated to use 5 mins for the polling delay and 2 mins for timeout to be more conservative and consistant with the hub.

packages/bot-strategy-runner/src/ConfigBuilder.ts
```diff
@@ -0,0 +1,203 @@
...
+        denominatorPriceFeedConfig: addressConfig.denominatorPriceFeedConfig
+      };
+      botConfigs.push(botConfig);
+    }
+  });
+  return botConfigs;
+}
+
+// Takes in two objects of any depth and merges them. In the case of a collision in keys the second object will take priority.
+export function mergeConfig(...args: any) {
```
daywiss, line 172: fyi lodash has several utilities for this, these look most closely to what you want:
- [merge](https://lodash.com/docs/4.17.15#merge)
- [defaultsDeep](https://lodash.com/docs/4.17.15#defaultsDeep)

this is tricky code and probably should have a unit test

packages/bot-strategy-runner/src/ConfigBuilder.ts
```diff
@@ -0,0 +1,203 @@
...
+  // iterate through all objects and deep merge them with target
+  for (let i = 0; i < args.length; i++) {
+    merger(args[i]);
+  }
+
+  return target;
+}
+
+// Takes in an object of any structure and returns the exact same object with all strings converted to check sum format.
+function replaceAddressCase(object: any) {
```
daywiss, line 199: this is cool

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    progressBar.start(allBotsConfigs.length, 0);
+
+    // Execute all bots in a bluebird map with limited concurrency. Note that none of the `runBot` calls ever throw errors.
+    // Rather, they indicate an error via an `error` key within the response. The promise.race between the `runBot` and
+    // _rejectAfterDelay bounds how long a given strategy can run for.
+    const executionResults = await (bluebird as any).map(
+      allBotsConfigs,
+      (botConfig: any) =>
+        Promise.all([
+          _updateProgressBar(botConfig), // As the promise progresses through bots update the progress bar.
```
daywiss, line 67: fancy

packages/bot-strategy-runner/src/ConfigBuilder.ts
```diff
@@ -0,0 +1,203 @@
...
+// files define in `globalAddressWhitelistUrls`.
+export async function buildGlobalWhitelist(config: strategyRunnerConfig) {
+  let whitelist = config.globalAddressWhitelist ? config.globalAddressWhitelist : [];
+
+  if (config.globalAddressWhitelistUrls) {
+    const responses = await Promise.all(config.globalAddressWhitelistUrls.map((url: string) => nodeFetch(url)));
+    const responseJson = await Promise.all(responses.map((response: any) => response.json()));
+    responseJson.forEach((contractAddressesResponse: any) => {
+      if (contractAddressesResponse.empWhitelist) whitelist = [...whitelist, ...contractAddressesResponse.empWhitelist];
+      else console.log("Global Whitelist file does not have the `empWhitelist` key or is malformed");
```
daywiss, line 48: why not throw error?

packages/bot-strategy-runner/src/LogCaptureTransport.ts
```diff
@@ -0,0 +1,32 @@
...
+  logStorage: any;
+  constructor(opts: any, logStorage: Array<any>) {
+    super(opts);
+    this.logStorage = logStorage;
+  }
+  async log(info: any, callback: any) {
+    try {
+      // eslint-disable-next-line prefer-const
+      let { timestamp, level, error, ...args } = info;
+      if (error) {
```
daywiss, line 15: is there a way the logger could not care about the data its logging?

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    urlConfig = await response.json(); // extract JSON from the http response
+    assert(!urlConfig.message && !urlConfig.error, `Could not fetch config! :${JSON.stringify(urlConfig)}`);
+  }
+  return mergeConfig(fileConfig, urlConfig);
+};
+
+// Returns a promise that is resolved after `seconds` delay. Used to limit how long a spoke can run for.
+const _rejectAfterDelay = (seconds: number | undefined, executionConfig: any) =>
+  new Promise((resolve, _) => {
+    setTimeout(resolve, (seconds ? seconds : defaultStrategyTimeout) * 1000, {
```
daywiss, line 153: this should be rejecting based on the function name, but i understand why you would rather resolve. we are seeing the logging system leaking into completely unrelated parts of the code.

daywiss (approved):
this is really cool, most of the complexity here is just wrangling the configs and logging, actually running the bots is really simple, which is great.  

this is not a blocker, but i guess my main concern here is error handling. It seems like there's a pattern of avoiding real errors. There is probably a good reason for why you do this though, maybe its just carry over from previous design or for the sake of clean logging, but I think things would be easier to debug if we embraced throwing often. Have a top level error handler which handles formatting the error into a log, rather than doing this deep in the code or requiring logger to understand context of the log.

packages/bot-strategy-runner/src/ConfigBuilder.ts
```diff
@@ -0,0 +1,203 @@
...
+  });
+  return botConfigs;
+}
+
+// Takes in two objects of any depth and merges them. In the case of a collision in keys the second object will take priority.
+export function mergeConfig(...args: any) {
+  const target: any = {}; // create a new object
+
+  // deep merge the object into the target object
+  const merger = (obj: any) => {
```
mrice32, line 176: nit: why not call this method deepMerge?

packages/bot-strategy-runner/src/ConfigBuilder.ts
```diff
@@ -0,0 +1,203 @@
...
+    merger(args[i]);
+  }
+
+  return target;
+}
+
+// Takes in an object of any structure and returns the exact same object with all strings converted to check sum format.
+function replaceAddressCase(object: any) {
+  const stringifiedObject = JSON.stringify(object);
+  const replacedStringifiedObject = stringifiedObject.replace(/0x[a-fA-F0-9]{40}/g, toChecksumAddress);
```
mrice32, line 201: I don't think this will cause any problems, but is it possible that this will mess with byte strings? I'm assuming that even if it does, it won't break anything. Just wanted to mention it.

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+import { strategyRunnerConfig, buildBotConfigs, buildGlobalWhitelist, mergeConfig } from "./ConfigBuilder";
+import { runBot } from "./BotEntryWrapper";
+
+// Defaults for bot execution. If the user does not define these then these settings will be used.
+const defaultPollingDelay = 300;
+const defaultStrategyTimeout = 120;
+const defaultBotConcurrency = 10;
+const defaultNetwork = "mainnet_mnemonic";
+
+// Global progress bar to show the status of the strategy runner during it's execution.
```
mrice32, line 19:
```suggestion
// Global progress bar to show the status of the strategy runner during its execution.
```

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+      message: "Executing set of bots concurrently",
+      concurrency: strategyRunnerConfig.botConcurrency
+    });
+
+    progressBar.start(allBotsConfigs.length, 0);
+
+    // Execute all bots in a bluebird map with limited concurrency. Note that none of the `runBot` calls ever throw errors.
+    // Rather, they indicate an error via an `error` key within the response. The promise.race between the `runBot` and
+    // _rejectAfterDelay bounds how long a given strategy can run for.
+    const executionResults = await (bluebird as any).map(
```
mrice32, line 63: Why do you have to do this cast? Do the bluebird types not contain the map function?

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+      allBotsConfigs,
+      (botConfig: any) =>
```
mrice32, line 65: Don't we have well-defined types for the bot configs?

packages/bot-strategy-runner/src/index.ts
```diff
@@ -0,0 +1,213 @@
...
+    strategyRunnerConfig: strategyRunnerConfig
+  });
+
+  // Generate a global whitelist of addresses that all enabled bots will run on.
+  const globalWhiteList = await buildGlobalWhitelist(strategyRunnerConfig);
+
+  // Construct bot configs for all enabled bots. This includes all bot specific overrides, whitelists, blacklists and any
+  // extra settings defined in the config. See the readme for full details on what is possible with this config.
+  const allBotsConfigs = await buildBotConfigs(globalWhiteList, strategyRunnerConfig);
+  console.log("allBotsConfigs", allBotsConfigs);
```
mrice32, line 41: Is this console.log supposed to be here?

packages/bot-strategy-runner/src/index.ts
mrice32: Agree, this is a bit confusing.

mrice32 (commented):
Looks great! Left a few minor comments.

One high-level question: what do you think we could do to bring down the number of `any` declarations in the code? Just convert all the other packages to ts or are they inevitable?

packages/bot-strategy-runner/src/ConfigBuilder.ts
chrismaree: ye, this is a better solution! thanks for pointing me towards them. I'll replace this custom method with the lodash one.

packages/bot-strategy-runner/src/ConfigBuilder.ts
chrismaree: ye, you right. will throw an error here.

chrismaree: well actually there was a reason for this. it is possible that the global whitelist is malformed but everything else is still fine. in that case the way it is structure here the bot will continue to work as expected, minus the ones from the whitelist.

However, this is pretty difficult to reason about so I think we should ideally throw an error like you prescribed. have updated accordingly.

packages/bot-strategy-runner/src/LogCaptureTransport.ts
chrismaree: so the reason it needs to do this is that the error object is of type `Error` and this does not log nicely at all. If you want to store this within a json object you cant easily read it or consume it within a nested structure (such as within a GCP log). This code here basically strips out all nasty punctuation and enables this to be treated as a simple string.

I'm curious how you'd prefer this to be consumed. very open to any alternative pattern!

packages/bot-strategy-runner/src/index.ts
chrismaree: ye, this is a great point and something I spent a few hours trying to find a better pattern on but ended up settling on this.

The main issue is if the promise rejects (or an error is thrown within the promise) then you _cant_ use bluebird promises. I tried out using node [async](https://caolan.github.io/async/v3/docs.html) and [p-limit](https://github.com/sindresorhus/p-limit) which all worked pretty well but their error handling was really poor. Node async did not have any pattern that was workable with this (that I could find). p-limit was better as you could use `promise.allSettled` which can handle errors via a `rejected` status but the code was more convoluted and even harder to read. We do something like this in the serverless hub. you can check it out here https://github.com/UMAprotocol/protocol/blob/48c290fd0600cbc35aad8c9193a31a1a692311de/packages/serverless-orchestration/src/ServerlessHub.js#L119

The pattern I ended up going for was to ensure that promises _never_ error or reject. For example, the bot-wrapper has a try-catch block wrapping any code that could error. Then, append a standard `error` prop to the response from the method and check for this after the fact. This standardizes things over all promises within this context.

My question is: how would you recommend dealing with rejected promises? what pattern would you follow if you want to handle concurrency like with a bluebird promise?

packages/bot-strategy-runner/src/ConfigBuilder.ts
chrismaree: ye, it _could_ but I think that this will only modify strings of length 40 starting with 0x. bytestrings tend to be longer at 64 chars (bytes32) with a 0x at the front.

packages/bot-strategy-runner/src/ConfigBuilder.ts
chrismaree: I've removed this method and am using lodash after Davids comment :)

packages/bot-strategy-runner/src/index.ts
chrismaree: otherwise I get this error:
image


I _have_ installed `"@types/bluebird"` so I'm not sure why this is the case. will do some more playing

chrismaree: ok I was able to address this by changing the import type. now imported as `import bluebird from "bluebird";` and it works as expected :)

packages/bot-strategy-runner/src/index.ts
chrismaree: yup! good point. I have imported them and used them in this mapping.

packages/bot-strategy-runner/src/index.ts
chrismaree: nope. have remove it!

chrismaree:
> Looks great! Left a few minor comments.
> 
> One high-level question: what do you think we could do to bring down the number of `any` declarations in the code? Just convert all the other packages to ts or are they inevitable?

Thanks for your review! I have addressed all your comments/questions and nits.

regarding your `any` comment you right that the main thing is just converted things in ts and declaring more interfaces.

@mrice32 what's your opinion on testing in this package?

packages/bot-strategy-runner/src/ConfigBuilder.ts
mrice32: I think it will touch any strings of length _greater than_ 40 hex characters (since the first 40 hex characters would satisfy the regex). Regardless, I'm not too worried about checksumming random byte strings > 40 characters -- just might look a little weird in the logs.

My only point was to raise the issue in case you thought this _could_ break something (like toChecksum erroring on non-addresses or common interpreters breaking if we paste in bytes with varied casings).

packages/bot-strategy-runner/src/index.ts
mrice32: Yeah, I hear you. Promise.allSettled not being available in bluebird is annoying and makes this inconvenient. I think your current implementation is reasonable!

mrice32:
> > Looks great! Left a few minor comments.
> > One high-level question: what do you think we could do to bring down the number of `any` declarations in the code? Just convert all the other packages to ts or are they inevitable?
> 
> Thanks for your review! I have addressed all your comments/questions and nits.
> 
> regarding your `any` comment you right that the main thing is just converted things in ts and declaring more interfaces.
> 
> @mrice32 what's your opinion on testing in this package?

@chrismaree is the question whether it's okay to not have tests for it? Or how we should structure tests?

As far as how you would structure a test, I think we could do limited testing with a sample config to prove:
1. That the runner is reading the config correctly.
2. That the runner acts as expected on error.

However, I'm not super opinionated. What are your thoughts?

mrice32 (approved):

chrismaree:
> > > Looks great! Left a few minor comments.
> > > One high-level question: what do you think we could do to bring down the number of `any` declarations in the code? Just convert all the other packages to ts or are they inevitable?
> > 
> > 
> > Thanks for your review! I have addressed all your comments/questions and nits.
> > regarding your `any` comment you right that the main thing is just converted things in ts and declaring more interfaces.
> > @mrice32 what's your opinion on testing in this package?
> 
> @chrismaree is the question whether it's okay to not have tests for it? Or how we should structure tests?
> 
> As far as how you would structure a test, I think we could do limited testing with a sample config to prove:
> 
> 1. That the runner is reading the config correctly.
> 2. That the runner acts as expected on error.
> 
> However, I'm not super opinionated. What are your thoughts?

It was mainly about the structure of the tests. I agree with your breakdown. I will leave this for a separate PR though as there is a fair amount of code already in this PR! I also think we should have a test that validates configs are built correctly in the right structure.

PR #2854, opened 2021-04-12
improve(funding-rate-proposer): update funding rates sequentially in order to set nonce

nicholaspai:

**Motivation**

`updateFundingRate` currently proposes funding rates to all Perpetuals in parallel, which means we cannot hardcode the `nonce` before passing transactions to the `ynatm` transaction manager. This leads `ynatm` to send ou transactions that bump gas but incorrectly increment the nonce, which causes unneccessary transaction reverts.

We need to update `ynatm` or implement an alternative transaction management feature that handles nonce management for parallel transactions. Until then, we need to submit transactions sequentially to `ynatm`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/funding-rate-proposer/src/proposer.js
+++ b/packages/funding-rate-proposer/src/proposer.js
@@ -126,9 +126,13 @@ class FundingRateProposer {
     // TODO: Should allow user to set an ROI-based metric to customize whether
     // bot strategically waits to submit funding rate proposals. Rewards
     // increase with time since last update.
-    await Promise.map(Object.keys(this.contractCache), contractAddress => {
-      return this._updateFundingRate(contractAddress, usePriceFeedTime);
-    });
+    // TODO: Note we update funding rates sequentially so that we can hardcode the nonce for each proposal passing
+    // transactions to the `ynatm` package. If these calls were submitted in parallel then we wouldn't be able to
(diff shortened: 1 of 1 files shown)


packages/funding-rate-proposer/src/proposer.js
```diff
@@ -126,9 +126,13 @@ class FundingRateProposer {
...
     // bot strategically waits to submit funding rate proposals. Rewards
     // increase with time since last update.
-    await Promise.map(Object.keys(this.contractCache), contractAddress => {
-      return this._updateFundingRate(contractAddress, usePriceFeedTime);
-    });
+    // TODO: Note we update funding rates sequentially so that we can hardcode the nonce for each proposal passing
+    // transactions to the `ynatm` package. If these calls were submitted in parallel then we wouldn't be able to
+    // hardcode the nonce, which could cause unintended reverts due to duplicate transactions. Once the `ynatm` package
+    // can handle nonce management, then we should update this logic to run in parallel.
+    for (let contractAddress of Object.keys(this.contractCache)) {
```
chrismaree, line 133: can this be `const`?

chrismaree (approved):
LGTM. better to do things sequentially right now

packages/funding-rate-proposer/src/proposer.js
nicholaspai: image

pemulis (approved):

PR #2855, opened 2021-04-12
fix(lm-payouts): address wrong week number in YD-BTC

chrismaree:
Last weeks LM had the wrong week number in the payout file due to a user error. The payouts recipients and amounts were correct, this was simply a semantic error.


evaldofelipe (approved):
LGTM

PR #2871, opened 2021-04-14
fix(trader & atomic liquidator): add wait for block to be mined util to address slow node updates

chrismaree:
**Motivation**

There have been a few scenarios where a state change happens on-chain (such as an atomic liquidator or a range trader trade) which are followed immediately by state reads. Infura often has not fully updated it's state of the chain, resulting in invalid logging. This PR adds a simple util function that blocks execution until a specific block number has been mined. This can be used to ensure that state has fully updated between state modification and state reads.

**Summary**

Added the util and implemented blocking code in the two places errors have been seen in production.

**Details**

Note that this issue is only present when working with the DSProxy. In normal cases, this error is not present as we normally execute a function call and read transaction events. For example, when executing a liquidation transaction we liquidate a position and then read the events from the transaction receipt. This does not work with the DSProxy as events are not preserved in a consumable way from the transaction. Rather, when dealing with the DSProxy, we execute the liquidation transaction which is then followed by an immediate query of `LiquidationCreated` events. If Infura has not fully updated this will not return the correct information. Equally, after preforming a DSProxy trade in the range trader the `spotPrice` immediately after the trade does not update accordingly due to a delay in Infura updates.

This PR mediates this by blocking the execution until the _next_ block is mined. An alternative solution would be to simply put a delay in the code but I felt that this technique is easier to reason about. I'm open to alternative solutions to this problem though!

**Testing**

- [ ]  New unit tests created
- [X] Untested


packages/common/src/ContractUtils.js
```diff
@@ -97,8 +99,19 @@ const runTransaction = async ({ transaction, config }) => {
...
-  revertWrapper,
-  runTransaction
+/**
+ * Blocking code until a specific block number is mined. Will re-fetch the current block number every 500ms. Useful when
+ * using methods called on contracts directly after state changes. Max blocking time should be ~ 15 seconds.
+ * @param {Object} web3 Provider from Truffle/node to connect to Ethereum network.
+ * @param {number} blockerBlockNumber block execution until this block number is mined.
+ */
+const blockUntilBlockMined = async (web3, blockerBlockNumber) => {
+  if (argv._.indexOf("test") != -1) return;
```
nicholaspai, line 109:
```suggestion
  if (argv._.indexOf("test") !== -1) return;
```

packages/common/src/ContractUtils.js
```diff
@@ -97,8 +99,19 @@ const runTransaction = async ({ transaction, config }) => {
...
+ * @param {Object} web3 Provider from Truffle/node to connect to Ethereum network.
+ * @param {number} blockerBlockNumber block execution until this block number is mined.
+ */
+const blockUntilBlockMined = async (web3, blockerBlockNumber) => {
+  if (argv._.indexOf("test") !== -1) return;
+  for (;;) {
+    const currentBlockNumber = await web3.eth.getBlockNumber();
+    if (currentBlockNumber >= blockerBlockNumber) break;
+    await new Promise(r => setTimeout(r, 500));
+  }
```
mrice32, line 114: Could this be done using: https://web3js.readthedocs.io/en/v1.2.11/web3-eth-subscribe.html#subscribe-newblockheaders? Or does that not work with http connections?

chrismaree: you could totally do it like that but I was not sure how to structure the method to _block_ until it had mined the associated block number. I tried using the `.on("data", function(blockHeader){...` syntax to toggle a flag and then the method would return but I could not get it to behave as expected. 

Perhaps you can think of a pattern to do this? I agree a subscription would be better than polling like I'm doing here.

nicholaspai: I think the for loop is a bit easier to reason about, because you `await` this entire method in `_executeLiquidationWithDsProxy`

nicholaspai (approved):
LGTM, nice fix!

packages/common/src/ContractUtils.js
chrismaree: yup, exactly. only trade off it is could be bit slower (up to half a second) and potentially has more web3 calls.

mrice32 (approved):
LGTM!

PR #2872, opened 2021-04-14
feat(version-management): Update hard coded latest package versions in the bots to use 2.0 packages

chrismaree:
**Motivation**

After releasing version 2.0 of UMA packages, we should remove all references to "latest" within unit tests and bot code to correctly refer to released code. This PR does this thought the repo.

**Details**

Note that some additional changes were made. In particular, any place that previously used `getTruffleContract` or `getAbi` with the `latest` version were reduced by removing the version flag. This was done as these methods default to latest and so no version is needed.


packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -41,10 +41,7 @@ async function claimRewards() {
       // This uses the VotingAncillaryInterfaceTesting to create a voting contract that has _only_ the ancillary
       // functions and events. This interface also includes almost all the regular voting methods that are unrelated
       // to the ancillary data change. Using this interface effectively avoids the overload problem.
-      return new web3.eth.Contract(
-        getAbi("VotingAncillaryInterfaceTesting", "latest"),
-        getAddress("Voting", networkId, version)
-      );
+      return new web3.eth.Contract(getAbi("VotingAncillaryInterfaceTesting"), getAddress("Voting", networkId, version));
```
chrismaree, line 44: this is the example of removing the redundant "latest" flag.

packages/core/scripts/local/ClaimAllRewards.js
```diff
@@ -90,7 +87,7 @@ async function claimRewards() {
           contract => toChecksumAddress(contract.options.address) === toChecksumAddress(contractAddress)
         );
         if (!voting) throw `Couldn't find voting contract for ${contractAddress}`;
-        const fullVotingContract = new web3.eth.Contract(getAbi("Voting", "latest"), voting.options.address);
+        const fullVotingContract = new web3.eth.Contract(getAbi("Voting"), voting.options.address);
```
chrismaree, line 90: this latest flag is redundant here.

packages/core/scripts/local/EncodeParams.js
```diff
@@ -49,8 +49,8 @@ const encodeParams = async callback => {
       timerAddress: await expiringMultiPartyCreator.timerAddress()
     };
 
-    // Inject constructor params neccessary for "latest" version of the EMPCreator:
-    if (abiVersion === "latest") {
+    // Inject constructor params neccessary for "2.0.1" version of the EMPCreator:
```
chrismaree, line 52: here (and in all the other bots) we can bound to this version to get anything EMP v2 or perp related.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
```diff
@@ -6,17 +6,15 @@ const { OptimisticOracleClient } = require("../../src/clients/OptimisticOracleCl
 const { interfaceName, advanceBlockAndSetTime } = require("@uma/common");
 const { getTruffleContract } = require("@uma/core");
 
-const CONTRACT_VERSION = "latest";
```
chrismaree, line 9: this one might be a bit controversial. IMHO it's redundant to define this. it's a fair bit cleaner to remove it wholesale. what do you think @UMAprotocol/eng ?

mrice32: @nicholaspai may have stronger opinions, but I'm totally fine relying on the default arg.

packages/financial-templates-lib/test/clients/FinancialContractClient.js
```diff
@@ -92,6 +92,7 @@ contract("FinancialContractClient.js", function(accounts) {
 
     // Import the tested versions of contracts. note that financialContract is either an ExpiringMultiParty or a
     // Perpetual depending on the current iteration version.
+    console.log(contractVersion);
```
mrice32, line 95: Extra console.log?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -32,11 +32,11 @@ const { VaultPriceFeed } = require("./VaultPriceFeed");
...
-  const Uniswap = getTruffleContract("Uniswap", web3, "latest");
-  const ERC20 = getTruffleContract("ExpandedERC20", web3, "latest");
-  const Balancer = getTruffleContract("Balancer", web3, "latest");
-  const VaultInterface = getTruffleContract("VaultInterface", web3, "latest");
-  const Perpetual = getTruffleContract("Perpetual", web3, "latest");
+  const Uniswap = getTruffleContract("Uniswap", web3);
+  const ERC20 = getTruffleContract("ExpandedERC20", web3);
+  const Balancer = getTruffleContract("Balancer", web3);
+  const VaultInterface = getTruffleContract("VaultInterface", web3);
+  const Perpetual = getTruffleContract("Perpetual", web3);
```
mrice32, line 39: Should we pin the perp here? Or should we use latest?

More generally, how did you decide where to pin vs use latest for contracts that _we've_ launched (the interfaces above should obviously use latest)?

Outside of this particular instance, it looks like we're pinning in the src, but using latest in the tests. Is that correct? That seems reasonable, just want to make sure I'm understanding correctly.

mrice32 (approved):
Looks great! Minor comments/questions

daywiss (approved):
lgtm

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
nicholaspai: I am indifferent. In the future if we want to move off of latest we can easily change this variable, so I'd lean slightly towards keeping it.

packages/common/test/ContractUtils.js
```diff
@@ -1,8 +1,8 @@
 // Script to test
 const ContractUtils = require("../src/ContractUtils");
 const { getTruffleContract, getAbi } = require("@uma/core");
-const ERC20 = getTruffleContract("BasicERC20", web3, "latest");
-const ERC20ABI = getAbi("BasicERC20", "latest");
```
nicholaspai, line 5: Does the `latest` flag cause the tests to unexpectedly use the wrong contract versions?

packages/financial-templates-lib/test/clients/FinancialContractClient.js
chrismaree: indeed! was double checking something. have removed

packages/common/test/ContractUtils.js
chrismaree: I dont think so as these kinds of contracts have never changed between versions.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
chrismaree: for now, lets stay on the side of defaulting unless tests explicitly execute on a specific version.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
chrismaree: we probably should pin the perp version at this point but I actually don't think it will matter as the downstream implementations are expecting any `Perpetual` that implements the right interface. As long as we don't shift the interface it wont break anything to keep it like this (or how the code was before).

My main decision was if things used `latest` within the `getTruffleContract` then the param is redundant. If `latest` was used within unit tests (in particular within a `versionedIt`) then it should be a locked version and I used the `2.0.1` release bytecode.

PR #2877, opened 2021-04-17
fix(liquidator): address issue in log produced resulting in Winston slack transport error

chrismaree:
**Motivation**

There is a small error in the log produced that is sent to Winston resulting in an error slack message. This PR should address this.

Faulty slack log:

image

The error produced is `Error:TypeError: Cannot read property 'length' of undefined`. The `undefined` in reference here is the `txnConfig`. This was not correctly set previously and is addressed in this PR.


nicholaspai (approved):
Wow this is a sneaky bug good catch Chris!

PR #2878, opened 2021-04-18
fix(FallBackPriceFeed): implement missing method that broke in production

chrismaree:
The Expression price feed does not fully implement the `PriceFeedInterface` resulting in errors in production. An example error is shown below:

image

This PR addresses this issue.

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
@@ -68,6 +68,11 @@ class FallBackPriceFeed extends PriceFeedInterface {
     }
   }
 
+  // return the longest lookback within all the fallback feeds.
+  getLookback() {
+    return Math.max(this.priceFeeds.map(feed => feed.historicalLookback));
+  }
+
   // Errors out if any price feed had a different number of decimals.
   getPriceFeedDecimals() {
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

packages/financial-templates-lib/src/price-feed/FallBackPriceFeed.js
```diff
@@ -68,6 +68,11 @@ class FallBackPriceFeed extends PriceFeedInterface {
     }
   }
 
+  // Return the longest lookback within all the fallback feeds.
```
mrice32, line 71: Shouldn't we match what the [medianizer](https://github.com/UMAprotocol/protocol/blob/24073fcd91cf959a7dd1e8156420bf32a7ea5080/packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.js#L111-L117) does and take the min instead of the max?

chrismaree: does shortest make the most sence? my reasoning is if any of the feeds within the FallBackFeed have a longer duration then we should take the longest one to maximise the changes of overlapping with the desired timestamp.

mrice32: Ah, good point. Agreed.

mrice32 (approved):
LGTM!

PR #2880, opened 2021-04-19
fix(blockfinder): Explicitly cast timestamp to Number

nicholaspai:
**Motivation**

Sometimes `timestamp` is passed in as a String via an upstream `getHistoricalPrice()` call, such as [here](https://github.com/UMAprotocol/protocol/blob/master/packages/core/scripts/local/getHistoricalPrice.js#L94) in the script that's designed to aid UMA voters. This bug currently causes the script to return a slightly different price than one that would be returned by the `optimistic-oracle-proposer` bot, which correctly inputs `timestamp` as a Number [here](https://github.com/UMAprotocol/protocol/blob/master/packages/optimistic-oracle/src/proposer.js#L265).

There might be other instances in the repo where a String `timestamp` is unexpectedly passed into this Blockfinder utility method expecting a number.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/utils.js
+++ b/packages/financial-templates-lib/src/price-feed/utils.js
@@ -199,7 +199,6 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
   async function findBlock(startBlock, endBlock, timestamp) {
     // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
     // timestamp.
-    timestamp = Number(timestamp);
     if (endBlock.timestamp === timestamp) return endBlock;
 
     // If there's no equality, but the blocks are adjacent, return the startBlock, since we want the returned block's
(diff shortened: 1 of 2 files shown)


chrismaree (approved):
we. need. typescript!

packages/financial-templates-lib/src/price-feed/utils.js
```diff
@@ -199,6 +199,7 @@ exports.BlockFinder = (requestBlock, blocks = []) => {
   async function findBlock(startBlock, endBlock, timestamp) {
     // In the case of equality, the endBlock is expected to be passed as the one whose timestamp === the requested
     // timestamp.
+    timestamp = Number(timestamp);
```
mrice32, line 202: Do we need this cast if it's already casted in getBlockForTimestamp? I don't think this method is externally callable (if I remember correctly).

mrice32 (approved):
@chrismaree agreed. IMO, it's an antipattern to allow inputs that can be different, coercable types. Otherwise, you end up having to cast _everywhere_ because you don't know if people are passing around strings or numbers at every level.

packages/financial-templates-lib/src/price-feed/utils.js
nicholaspai: Good point. I had added originally for completeness but will minimize the size of the PR and remove

PR #2881, opened 2021-04-19
fix(funding-rate-proposer): Use historical-price to propose funding rates

nicholaspai:
**Motivation**

Funding Rate Proposer bot uses the latest block timestamp as the request timestamp. Any disputes will reference this same timestamp, therefore the proposed price should be the historical price for this timestamp. Currently, the current price is used, but the current price can be off from the historical price at the request timestamp because of the lag between the last block time and the current (off-chain) system time.

The latest block timestamp is used as the request time instead of the current system time because the Optimistic Oracle prevents request timestamps from being >= the latest block time. See [this PR](https://github.com/UMAprotocol/protocol/pull/2687) for more details.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/funding-rate-proposer/src/proposer.js
+++ b/packages/funding-rate-proposer/src/proposer.js
@@ -163,8 +163,33 @@ class FundingRateProposer {
 
     // Assume pricefeed has been cached and updated prior to this function via the `update()` call.
     const priceFeed = this.priceFeedCache[fundingRateIdentifier];
-    const pricefeedPrice = await this._getPriceFeedAndPrice(fundingRateIdentifier);
-    if (!pricefeedPrice) return;
+    if (!priceFeed) {
+      this.logger.error({
+        at: "PerpetualProposer",
+        message: "Failed to create pricefeed for funding rate identifier",
+        fundingRateIdentifier
+      });
+      return null;
+    }
+    // Get hypothetical timestamp to use for a new price proposal:
+    // Unless `usePriceFeedTime=true`, use the latest block's timestamp as the request
+    // timestamp so that the contract does not interpret `requestTimestamp` as being in the future.
+    const requestTimestamp = usePriceFeedTime
+      ? priceFeed.getLastUpdateTime()
+      : (await this.web3.eth.getBlock("latest")).timestamp;
+    let _pricefeedPrice;
+    try {
+      _pricefeedPrice = (await priceFeed.getHistoricalPrice(Number(requestTimestamp))).toString();
+    } catch (error) {
+      this.logger.error({
+        at: "PerpetualProposer",
+        message: "Failed to query current price for funding rate identifier",
+        fundingRateIdentifier,
+        requestTimestamp
+      });
+      return;
+    }
+    const pricefeedPrice = this.formatPriceToPricefeedPrecision(_pricefeedPrice, priceFeed);
     let onchainFundingRate = currentFundingRateData.rate.toString();
 
     // Check that pricefeedPrice is within [configStore.minFundingRate, configStore.maxFundingRate]
@@ -195,11 +220,6 @@ class FundingRateProposer {
       this.toBN(this.toWei(this.fundingRateErrorPercent.toString()))
     );
     if (shouldUpdateFundingRate) {
-      // Unless `usePriceFeedTime=true`, use the latest block's timestamp as the request
-      // timestamp so that the contract does not interpret `requestTimestamp` as being in the future.
-      const requestTimestamp = usePriceFeedTime
-        ? priceFeed.getLastUpdateTime()
-        : (await this.web3.eth.getBlock("latest")).timestamp;
       const proposal = cachedContract.contract.methods.proposeFundingRate(
         { rawValue: pricefeedPrice },
         requestTimestamp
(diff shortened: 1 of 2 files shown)


packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
...
+        at: "PerpetualProposer",
+        message: "Failed to create pricefeed for funding rate identifier",
+        fundingRateIdentifier
+      });
+      return null;
+    }
+    // Get hypothetical timestamp to use for a new price proposal:
+    // Unless `usePriceFeedTime=true`, use the latest block's timestamp as the request
+    // timestamp so that the contract does not interpret `requestTimestamp` as being in the future.
+    const requestTimestamp = usePriceFeedTime
```
nicholaspai, line 177: This `requestTimestamp` logic is copied from below, line 198ish

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
 
     // Assume pricefeed has been cached and updated prior to this function via the `update()` call.
     const priceFeed = this.priceFeedCache[fundingRateIdentifier];
-    const pricefeedPrice = await this._getPriceFeedAndPrice(fundingRateIdentifier);
-    if (!pricefeedPrice) return;
+    if (!priceFeed) {
```
nicholaspai, line 166: This checking for `!priceFeed` is copied from below, line 397ish

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
...
+      return null;
+    }
+    // Get hypothetical timestamp to use for a new price proposal:
+    // Unless `usePriceFeedTime=true`, use the latest block's timestamp as the request
+    // timestamp so that the contract does not interpret `requestTimestamp` as being in the future.
+    const requestTimestamp = usePriceFeedTime
+      ? priceFeed.getLastUpdateTime()
+      : (await this.web3.eth.getBlock("latest")).timestamp;
+    let _pricefeedPrice;
+    try {
```
nicholaspai, line 181: This try-catch is the new logic in that it computes a historical (funding rate) price instead of the current price.

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
...
+      _pricefeedPrice = (await priceFeed.getHistoricalPrice(Number(requestTimestamp))).toString();
+    } catch (error) {
+      this.logger.error({
+        at: "PerpetualProposer",
```
mrice32, line 185: should we log the error obj/message?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
...
+    if (!priceFeed) {
+      this.logger.error({
+        at: "PerpetualProposer",
+        message: "Failed to create pricefeed for funding rate identifier",
+        fundingRateIdentifier
+      });
+      return null;
+    }
+    // Get hypothetical timestamp to use for a new price proposal:
+    // Unless `usePriceFeedTime=true`, use the latest block's timestamp as the request
```
mrice32, line 175: When would we want to use `usePriceFeedTime`?

packages/funding-rate-proposer/src/proposer.js
```diff
@@ -163,8 +163,33 @@ class FundingRateProposer {
...
+    // timestamp so that the contract does not interpret `requestTimestamp` as being in the future.
+    const requestTimestamp = usePriceFeedTime
+      ? priceFeed.getLastUpdateTime()
+      : (await this.web3.eth.getBlock("latest")).timestamp;
```
mrice32, line 179: Is it a problem if the latest block time is > the price feed's latest update time?

mrice32 (commented):

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Yes! Thanks for catching this

packages/funding-rate-proposer/src/proposer.js
nicholaspai: In test environments so we don't have to modify the block timing from JS. This is convenient for testing because we can manually set the pricefeed's timestamp easier than we can manually set the block time

packages/funding-rate-proposer/src/proposer.js
nicholaspai: Or we can do like the `optimistic-oracle-proposer` and just conservatively `update()` the priceFeed before fetching historical price. [reference](https://github.com/UMAprotocol/protocol/blob/master/packages/optimistic-oracle/src/proposer.js#L262)

packages/funding-rate-proposer/src/proposer.js
mrice32: Ah, makes sense!

packages/funding-rate-proposer/src/proposer.js
nicholaspai: I implemented the above ^

mrice32: Sounds good! This isn't guaranteed to have an update time > block time, but I think this is probably as close as we can reasonably get:
1. Price feeds usually throttle to ensure they don't update more than once per minute by default.
2. Block times can (but usually don't) be ahead of wall-clock time (they have a 90 second or so tolerance if I remember correctly).

mrice32 (approved):
LGTM!

packages/funding-rate-proposer/src/proposer.js
nicholaspai: yeah the other solution which I think is a bit over engineered for now is something like

```js
let requestTimestamp;

// Update pricefeed until its last update time is caught up with the latest network block, or
// break early after 1 minute and assume that's good enough (or throw).
setTimeout(function() {
    while(true) {
        await priceFeed.update();
        requestTimestamp = (await web3.eth.getBlock()).timestamp
        if (priceFeed.lastUpdateTime() >= requestTimestamp) break;
    }
},
60000
```

mrice32: Yeah, agree with your decision to do it the way you have it in the PR. LGTM!

mrice32 (approved):
LGTM!

packages/funding-rate-proposer/src/proposer.js
nicholaspai: But this wouldn't be useful unless the `minTimeBetweenUpdates` was respected, and given the expected iteration cadence of 5 minutes we don't have much timeout period to play with. This sort of method would make sense in a looping infrastructure

PR #2893, opened 2021-04-20
improve (CircleCI): Split lerna packages tests 

evaldofelipe:
Signed-off-by: evaldofelipe <contato@evaldofelipe.com>

**Motivation**

Improve the CI speed on the tests execution

**Summary**

This PR implement Lerna commands to check the packages with changes, and the dependencies, and circleCI tools and bash scripts for split these packages in separated jobs to increase the execution speed. The result is packages test only if necessary, in packages and their dependencies.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


.circleci/config.yml
```diff
@@ -86,7 +86,13 @@ jobs:
             mkdir truffle_workaround
             mv protocol truffle_workaround/
             cd truffle_workaround/protocol
-            yarn run test-concurrent
+            export PREV_SHA=$(git log HEAD~1 --pretty=format:%H -1)
```
mrice32, line 89: nit: do you need export?
```suggestion
            PREV_SHA=$(git log HEAD~1 --pretty=format:%H -1)
```

package.json
```diff
@@ -11,6 +11,7 @@
     "prettier": "prettier './**/*.js' './**/*.sol' './**/*.md' './**/*.ts'",
     "test": "lerna run --stream --concurrency=4 test",
     "test-concurrent": "lerna run --stream --concurrency=4 test --ignore @uma/serverless-orchestration --ignore @uma/affiliates && lerna run --stream --concurrency=1 test --scope @uma/serverless-orchestration --scope @uma/affiliates",
+    "test-concurrent-test": "lerna run --stream --concurrency=4 test --scope ${PACKAGES_CHANGES}",
```
mrice32, line 14: As discussed today, I think you could just use `yarn test --scope ${SOME_PACKAGE_NAME}`

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,65 @@
...
+EOF
+  done
+
+  printf "\n\nworkflows:\n  version: 2.1\n  build_and_test:\n    jobs:\n" >> $CI_CONFIG_FILE
+
+  for PACKAGE in "${PACKAGES_ARRAY[@]}"
+    do
+      cat <<EOF >> $CI_CONFIG_FILE
+      - test-${PACKAGE:5}:
+          context: api_keys
```
mrice32, line 61: As discussed today, we probably want to add a single job that requires all these jobs to complete so we can require that that job completes before merging.

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,65 @@
...
+  for PACKAGE in "${PACKAGES_ARRAY[@]}"
+    do
+      cat <<EOF >> $CI_CONFIG_FILE
+    test-${PACKAGE:5}:
+      docker:
+        - image: circleci/node:lts
+        - image: trufflesuite/ganache-cli
+          command: ganache-cli -i 1234 -l 9000000 -p 9545
+      working_directory: ~/protocol
+      resource_class: large
```
mrice32, line 42: nit: we can probably drop this line because we no longer need a large instance if we're not trying to run a bunch of test processes in parallel in the same job.

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,65 @@
...
+      steps:
+        - restore_cache:
+            key: protocol-completed-build-{{ .Environment.CIRCLE_SHA1 }}
+        - run:
+            name: Run tests
+            command: |
+              ./ci/truffle_workaround.sh
+              export PACKAGES_CHANGES=$PACKAGE
+              yarn run test-concurrent-test;
```
mrice32, line 51: As discussed, you can probably replace this with something a little simpler.
```suggestion
              yarn test --scope ${PACKAGE};
```

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,65 @@
...
+  echo "No packages for testing."
+  circleci-agent step halt;
+
+else
+
+  echo "Packages to execute:"
+  echo ${PACKAGES_ARRAY[*]}
+
+  echo "Ignored packages:"
+  echo ${PACKAGES_IGNORE[*]}
```
mrice32, line 29: As discussed today, we don't need to ignore any packages.

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,65 @@
...
+
+for IGNORE in ${PACKAGES_IGNORE[@]}
+do
+   PACKAGES_ARRAY=(${PACKAGES_TOTAL[@]/$PACKAGES_IGNORE})
+done
+
+if [ ${#PACKAGES_ARRAY[@]} -eq 0 ]; then
+
+  echo "No packages for testing."
+  circleci-agent step halt;
```
mrice32, line 21: I would still want to understand why this is necessary?

mrice32 (commented):
Looks great! A few comments that came out of our conversation earlier today.

ci/generate_lerna_config.sh
evaldofelipe: Fixed.

ci/generate_lerna_config.sh
evaldofelipe: Changed.

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,92 @@
...
+          command: ganache-cli -i 1234 -l 9000000 -p 9545
+      working_directory: ~/protocol
+      steps:
+        - restore_cache:
+            key: protocol-completed-build-{{ .Environment.CIRCLE_SHA1 }}
+        - run:
+            name: Run tests
+            command: |
+              ./ci/truffle_workaround.sh
+              export PACKAGES_CHANGES=$PACKAGE
```
mrice32, line 55: nit: can you drop this line? It looks like `PACKAGES_CHANGES` isn't used anywhere.

ci/check_lerna_packages.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+HASH=$(git merge-base HEAD master)
+yarn lerna ls --since ${HASH} --include-dependents > lerna_output
+if [[ $(cat lerna_output | grep @) ]]; then
+  cat lerna_output | grep @ > lerna_packages
```
mrice32, line 9: Why do you need to grep for `@`?

mrice32 (approved):
Looks good -- two minor comments!

evaldofelipe:
In reply to: https://github.com/UMAprotocol/protocol/pull/2893#discussion_r625131673 This was necessary to get all the packages from the Lerna output and avoid  unnecessary information. All packages start like `@uma/core` and we need this list to use in the next steps

.circleci/config.yml
```diff
@@ -1,4 +1,10 @@
-version: 2
+version: 2.1
+
+setup: true
+
```
chrismaree, line 4: nit: do we want to have blank lines in this markdown?

ci/check_lerna_packages.sh
chrismaree: I guess it's the symbol that precedes the lerna package? eg `@uma/core`

ci/generate_lerna_config.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
+
+set -o errexit
+set -o nounset
+
+PACKAGES_ARRAY=($(cat lerna_packages))
+CI_CONFIG_FILE=".circleci/lerna_config.yml"
+
+if [ ${#PACKAGES_ARRAY[@]} -eq 0 ]; then
```
chrismaree, line 9: this is pretty awesome.

ci/truffle_workaround.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Note: this is a workaround for a strnage truffle issue.
```
chrismaree, line 3:
```suggestion
# Note: this is a workaround for a strange truffle issue.
```

chrismaree (approved):
This is awesome! Great work on getting this out. it's a somewhat custom solution but I think it'll go a long way to helping us speed up our dev process.

image

mrice32:
@evaldofelipe we do have a few packages that don't start that way, but I think that's a separate issue we should fix.

PR #2894, opened 2021-04-20
feat(chainbridge): Demo script showing how cross-chain price request would work

nicholaspai:
**Motivation**

Uses GenericHandler from [chainbridge contracts](https://github.com/ChainSafe/chainbridge-solidity) that is compatible with the chainbridge CLI and SDK. Would enable us to conveniently pass messages between EVM networks assuming each network has a Bridge and Handler contract setup.  More details on architecture [here](https://chainbridge.chainsafe.io/chains/ethereum/).

The demo test `test/chainbridge/voting.js` showcases the following flow to request a price from Sidechain to Mainnet:
- Someone needs a price that is unavailable on the sidechain `Voting` contract, so they trigger the `deposit()` function on the sidechain `Bridge` contract.
- Off-chain relayer sees the deposit succeeded via a `Deposit` event. Relayer then triggers `executeProposal` on mainnet `Bridge` contract, which submits a `requestPrice()` transaction to the mainnet `Voting` contract.

Additionally, the flow to publish prices from Mainnet to Sidechain is:
- A price has been resolved on Mainnet. Someone triggers the `deposit()` function on the sidechain `Bridge` contract.
- Off-chain relayer sees the deposit succeeded via a `Deposit` event. Relayer then triggers `executeProposal` on sidechain `Bridge` contract, which submits a `pushPrice()` transaction to the sidechain `Voting` contract.

**Note for Reviewers**

All contract code is taken from [here](https://github.com/ChainSafe/chainbridge-solidity/tree/master/contracts) and the test `js` code is inspired by the `deposit` and `executeDeposit` tests [here](https://github.com/ChainSafe/chainbridge-solidity/tree/master/test/handlers/generic).

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/oracle/test/MockOracleAncillary.sol
```diff
@@ -44,6 +44,8 @@ contract MockOracleAncillary is OracleAncillaryInterface, Testable {
     mapping(bytes32 => mapping(uint256 => mapping(bytes => QueryIndex))) private queryIndices;
     QueryPoint[] private requestedPrices;
 
+    event PriceRequestAdded(uint256 indexed roundId, bytes32 indexed identifier, uint256 time);
```
nicholaspai, line 47: This event is added so that we can more easily test that the Handler is requesting a price correctly from this contract.

packages/core/contracts/chainbridge/Bridge.sol
```diff
@@ -0,0 +1,452 @@
+pragma solidity ^0.6.0;
```
mrice32, line 1: I really wish we didn't have to copy these contracts in, but that seems to be the only thing we can do since their package doesn't export the solidity, only the typescript bindings and ethers objects :/

mrice32 (approved):
I'm assuming that this is all copied? If so, LGTM!

packages/core/test/chainbridge/helpers.js
```diff
@@ -0,0 +1,126 @@
+/**
```
mrice32, line 1: Are these tests just testing the chainbridge code and/or replicating chainbridge tests? Do we need to test it if they already are? Or is this testing specific use cases that we have in mind?

nicholaspai: This file is copied from the chainbridge repo. The `voting.js` file is the only one that I created and it demonstrates how chainbridge would work with our Voting contract

mrice32: Are these tests for that file?

mrice32: Oh never mind, I see. Voting.js is a test.

mrice32 (approved):
LGTM!

mrice32 (approved):
Anything blocking us from merging this?

PR #2903, opened 2021-04-26
feat(L2): Implement BeaconOracle designed for cross-chain PriceRequest communication

nicholaspai:
**Motivation**

See #2894 for intended work flow, summarized below, with artistic rendering of the architecture [here](https://docs.google.com/presentation/d/1sGvKMknFAhDf-tfd6EkEOijGMrXCieC2hIycIObAVGs/edit#slide=id.gd5fbfb2d7f_0_70):

_Requesting a price from L2 to L1_
- [L2] `OptimisticOracle` requests price to `SinkOracle`
- [L2] `SinkOracle` stores price request and calls `deposit()` on `Bridge` contract with price request metadata
- [L2] `Bridge` performs sanity check on price request data and if successful, emits a `Deposit` event.
- [off-chain] Relayer bot EOA sees `Deposit` event emitted from L2 and calls `executeProposal` with the latest price request metadata on L1 `Bridge`.
- [L1] `Bridge` calls `GenericHandler.executeProposal` with price request metadata, which calls `SourceOracle.requestPrice()`
- [L1] `SourceOracle` ultimately makes request to DVM, `Voting.sol`.

_Publishing a price from L1 to L2_
- [L1] Some admin account copies price on `SourceOracle` from `DVM` via `publishPrice()`
- [L1] `SourceOracle` stores price resolution and calls `deposit()` on `Bridge` contract with price resolution metadata
- [L1] `Bridge` performs sanity check on price request data and if successful, emits a `Deposit` event.
- [off-chain] Relayer bot EOA sees `Deposit` event emitted from L1 and calls `executeProposal` with the latest price resolution metadata on L2 `Bridge`.
- [L2] `Bridge` calls `GenericHandler.executeProposal` with price resolution metadata, which calls `SinkOracle.publishPrice()`
- [L1] `SinkOracle` ultimately stores price resolution data which is now available to the `OptimisticOracle`

**Follow up PR**
Add unit tests for `BeaconOracle` implementations and test between Goerli<>Rinkeby.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
```diff
@@ -0,0 +1,35 @@
...
+
+    function requestPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public override {
+        _requestPrice(identifier, time, ancillaryData);
+
+        // TODO: Call Bridge.deposit() to intiate cross-chain price request.
+        // _getBridge().deposit(formattedMetadata);
```
nicholaspai, line 24: Here is where we'd want to create recreate this `javascript` code in Solidity:

```js
const encodedMetaDataProposal = Helpers.abiEncode(
      ["bytes32", "uint256", "bytes"],
      [padRight(identifier, 64), requestTime, ancillaryData]
    );
    const depositData = Helpers.createGenericDepositData(encodedMetaDataProposal);
    const depositTxn = await bridgeSidechain.deposit(chainId, votingResourceSidechainId, depositData, {
      from: depositerAddress
    });
```

packages/core/contracts/oracle/implementation/cross-chain/SourceOracle.sol
```diff
@@ -0,0 +1,37 @@
...
+    function pushPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData,
+        int256 price
+    ) public {
+        _pushPrice(identifier, time, ancillaryData, price);
+
+        // TODO: Call Bridge.deposit() to intiate cross-chain publishing of price request.
+        // _getBridge().deposit(formattedMetadata);
```
nicholaspai, line 35: And here is where we'd want to recreate this `javascript` code in Solidity to pass data to `pushPrice`:

```js
    const encodedMetaDataProposal = Helpers.abiEncode(
      ["bytes32", "uint256", "bytes", "int256"],
      [padRight(identifier, 64), requestTime, ancillaryData, requestPrice]
    );
    const depositData = Helpers.createGenericDepositData(encodedMetaDataProposal);
    const depositTxn = await bridgeMainnet.deposit(sidechainId, votingResourceId, depositData, {
      from: depositerAddress
    });
```

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint256 time;
+        bytes ancillaryData;
+    }
+
+    // Conceptually we want a (time, identifier, ancillaryData) -> price map.
+    mapping(bytes32 => mapping(uint256 => mapping(bytes => Price))) internal verifiedPrices;
+
+    // The mapping and array allow retrieving all the elements in a mapping and finding/deleting elements.
+    // Can we generalize this data structure?
+    mapping(bytes32 => mapping(uint256 => mapping(bytes => QueryIndex))) internal queryIndices;
```
mrice32, line 49: Would it be simpler if we did something similar to the Voting contract where we pre-hash these into a bytes32 to make a single mapping `mapping(bytes32 => Price/QueryIndex)`?

Here's the voting contract implementation: https://github.com/UMAprotocol/protocol/blob/eb55ff2941bd5fb587c5db992a6bd8a89672b626/packages/core/contracts/oracle/implementation/Voting.sol#L907-L913.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
```diff
@@ -0,0 +1,150 @@
...
+            emit PriceRequestAdded(msg.sender, identifier, time, ancillaryData);
+        }
+    }
+
+    /**
+     * @notice Pushes the verified price for a requested query.
+     * @dev Should be called by an off-chain relayer who saw a price resolved on a different network's Source or Sink
+     * Oracle.
+     */
+    function _pushPrice(
```
mrice32, line 98: A lot of this functionality seems like it was pulled from the mock oracle, but I think we can simplify if we don't need to have a method that returns all of the unresolved price requests (which we do have in the mock oracle, but don't seem to need here).

If you agree, I think we could drop the 3 data structures and just have a single mapping from the (identifier, timestamp, ancillary data) to a struct that stores:
1. The state of the request: (NeverRequested, Requested, Resolved).
2. The price.

This would also mean these methods get a lot simpler.

For request price, all we would need to do is check the current state of the request and update it.

For push price, we would just check the state and set the price.

I think both would end up being 4-ish lines if I'm thinking about it correctly.

Thoughts?

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
```diff
@@ -0,0 +1,35 @@
...
+ * @title Simple implementation of the OracleInterface that is intended to be deployed on non-Mainnet networks and used
+ * to communicate price request data cross-chain with a Source Oracle on Mainnet. An Admin can request prices from
+ * this oracle, which might ultimately request a price from a Mainnet Source Oracle and eventually the DVM.
+ * @dev Admins capable of making price requests to this contract should be OptimisticOracle contracts. This enables
+ * optimistic price resolution on non-Mainnet networks while also providing ultimate security by the DVM on Mainnet.
+ */
+contract SinkOracle is BeaconOracle {
+    constructor(address _finderAddress) public BeaconOracle(_finderAddress) {}
+
+    function requestPrice(
```
mrice32, line 16: Do we need to check the registry to ensure the request is coming from an approved contract (to avoid spam)?

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
mrice32: Which parts are hardcoded? Everything except for the depositData?

Can you point me to Helpers? I don't know exactly what it's doing to format the data.

Last question, what method are you encoding to be called on the other side? requestPrice?

mrice32 (commented):
Looks great! A few minor questions

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
```diff
@@ -0,0 +1,150 @@
...
+ * @title Simple implementation of the OracleInterface used to communicate price request data cross-chain between
+ * EVM networks. Can be extended either into a "Source" or "Sink" oracle. The intention is that the "Source" Oracle
+ * is the originator of price resolution data, secured by the DVM, and the "Sink" Oracle can both receive this price
+ * resolution data and request prices cross-chain to the "Source" Oracle. The "Sink" is designed to be deployed on
+ * non-Mainnet networks and the "Source" should be deployed on Mainnet.
+ */
+abstract contract BeaconOracle is OracleAncillaryInterface, MultiRole {
+    enum Roles {
+        Owner, // The owner manages the other roles.
+        Requester, // Can make price requests.
```
mrice32, line 21: On the side-chain, wouldn't we be pulling this from the registry (i.e. wouldn't it be all registered contracts)? Or are we planning a different permissioning architecture?

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
```diff
@@ -0,0 +1,150 @@
...
+        uint256 time,
+        bytes ancillaryData,
+        int256 price
+    );
+
+    /**
+     * @notice Constructor.
+     * @param _finderAddress finder to use to get addresses of DVM contracts.
+     */
+    constructor(address _finderAddress) public {
```
mrice32, line 69: Do we need to initialize the multirole roles in the constructor?

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
nicholaspai: So I think we would deploy this contract with the `chainId` and `votingResourceSidechainId` sort of hard coded (although we could probably auto detect the latter):

Look [here at this other PR](https://github.com/UMAprotocol/protocol/pull/2894/files#diff-1f7f92853cda43a48b00c8da51120b1704efe313f79842fb27281a558d74014bR124) for reference

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: Is that easier? I think we shoudl consider from a performance standpoint whether that's better than a double nested mapping

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: I agree let's simplify. I do think we should be able to return pending price requests for convenience however, so perhaps we make the price requests struct pblic?

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
nicholaspai: Hmm we might want to? But that's why I made `BeaconOracle` inherit MultiRole and just set a `Requester` role, but we could instead just re-use the Registry.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: Let's go with Registry for requests, however do you re-use the same list for price publishers? If not, then do you create another role?

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: Ah yes, good point.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: I would argue that since we don't need it on-chain, we could just use events to construct it. If we do want to be able to return a list, the data structures as you have them may make sense -- I think that list is the source of the complexity if I remember correctly.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: Performance-wise, I would think they're pretty close, but the difference would likely be negligible. In the nested mapping, you basically do:
```
value = storage_slot(hash(hash(hash(some_constant, identifier), time), ancillary_data))
```

In the _encodePriceRequest method, you would be doing:
```
value = storage_slot(hash(some_constant, hash(identifier, time, ancillary_data)))
```

Technically the only difference should be one hash, but you never know which one will end up being more expensive bytecode-wise after compilation.

However, I would argue that since the perf differences are likely negligible, we should go for readability. Whichever one you find to be more readable/understandable SGTM!

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: Yeah, not sure. Let me explain how I see it, and let me know if you agree:
- Sidechain publisher: bridge.
- Sidechain requester: registry approved contracts.
- Main chain publisher: none? It seems like _this_ contract should grab the price directly from the DVM?
- Main chain requester: bridge.

A few of these don't really fit into multirole, but some do. How do you think we should manage that? Maybe we should just have these methods be non-public in the parent contract and have the derived contracts manage the permissioning individually?

Thoughts?

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
mrice32: So this is how you might do this:
Call abi.encodeWithSignature("YOUR_SIGNATURE_HERE", argument1, argument2, ...) to generate the calldata onchain. Note: this would look like this for requestPrice (not verified):
```
bytes memory callData = abi.encodeWithSignature("requestPrice(bytes32,uint256,bytes)", identifier, timestamp, ancillaryData);
```

Note: to make this a little cheaper, you might want to compute the function selector offchain and just hardcode on-chain.
```
bytes4 selector = bytes4(keccak256(bytes("requestPrice(bytes32,uint256,bytes)"))); // Computed once or even hardcoded
...
bytes memory callData = abi.encodeWithSelector(selector, identifier, timestamp, ancillaryData);
```

Reference: https://docs.soliditylang.org/en/v0.6.12/units-and-global-variables.html#abi-encoding-and-decoding-functions.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: Yeah removing that list simplies things a lot

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: We can also make all publish and request methods callable only by registered contracts, and set the Handler (i.e. the Bridge in your example, but its technically the Handler) as a registered contract.

nicholaspai: But I do agree with your categorization except that `bridge` should be `generic handler`.

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
nicholaspai: I ended up going with the hash approach because I actually think its a bit more readable not not have a triple array

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
nicholaspai: Ended up not needing to specify the function selector in the contract, only need to pack the params

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: Yeah, agreed.

nicholaspai:
Tested these contracts on public networks:

(_SinkOracle on Rinkeby_)
(_SourceOracle on Goerli_)

1. Requesting price on `SinkOracle` emits `Deposit` event on local `Bridge`: [etherscan](https://rinkeby.etherscan.io/tx/0xe770d9ca5fb9e7c3cfbd4b2d3be5028a688d9c29700868f776a4a8f37c22fb4d)
2. Off-chain relayer saw `Deposit` event and executed proposal on `Bridge` on Goerli, which triggers price request on Goerli `SourceOracle` and `MockOracle`: [etherscan](https://goerli.etherscan.io/tx/0x40b059da55c1c286ea7d6b4bd4ecfdd1cc86d07d437880bdd81c5bb38754981f)
3. After Oracle resolves price request, publishing price on `SourceOracle` emits `Deposit` event on local `Bridge`: [etherscan](https://goerli.etherscan.io/tx/0x40b059da55c1c286ea7d6b4bd4ecfdd1cc86d07d437880bdd81c5bb38754981f)
4. Off-chain relayer saw `Deposit` event and executed proposal on `Bridge` on Rinkeby, which triggers price resolution on Rinkeby `SinkOracle`: [etherscan](https://rinkeby.etherscan.io/tx/0xab339c32b66e299b98ecfdcd6834b73f85d8f10eec0778e89fd39def16fe8293)

packages/core/contracts/chainbridge/SourceOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+/**
+ * @title Extension of BeaconOracle that is intended to be deployed on Mainnet to give financial
+ * contracts on non-Mainnet networks the ability to trigger cross-chain price requests to the Mainnet DVM. This contract
+ * is responsible for triggering price requests originating from non-Mainnet, and broadcasting resolved price data
+ * back to those networks.
+ * @dev The intended client of this contract is some off-chain bot watching for resolved price events on the DVM. Once
+ * that bot sees a price has resolved, it can call `publishPrice()` on this contract which will call the local Bridge
+ * contract to signal to an off-chain relayer to bridge a price request to another network.
+ */
+contract SourceOracle is BeaconOracle {
```
mrice32, line 23: Do we deploy one of these per L2? Or do we deploy only one of these and it would service all other chains? I believe the latter, right? In that case (this isn't necessarily a problem), we should just be aware that someone can forward price results to any other chain if they want, and those will fail to push on the other side (because they were never requested). Not sure if this is gracefully handled by the client.

packages/core/contracts/chainbridge/SourceOracle.sol
```diff
@@ -0,0 +1,124 @@
...
+     */
+    function publishPrice(
+        uint8 destinationChainID,
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData,
+        int256 price
+    ) public {
+        require(_getOracle().hasPrice(identifier, time, ancillaryData), "DVM has not resolved price");
+        require(_getOracle().getPrice(identifier, time, ancillaryData) == price, "DVM resolved different price");
```
mrice32, line 49: Why require the caller to provide a price? Why not just always publish the result of getPrice?

packages/core/contracts/chainbridge/BeaconOracle.sol
```diff
@@ -0,0 +1,150 @@
...
+
+/**
+ * @title Simple implementation of the OracleInterface used to communicate price request data cross-chain between
+ * EVM networks. Can be extended either into a "Source" or "Sink" oracle that specializes in making and resolving
+ * cross-chain price requests, respectivly. The "Source" Oracle is the originator or source of price resolution data
+ * and can only resolve prices already published by the DVM. The "Sink" Oracle receives the price resolution data
+ * from the Source Oracle and makes it available on non-Mainnet chains. The "Sink" Oracle can also be used to trigger
+ * price requests from the DVM on Mainnet.
+ */
+abstract contract BeaconOracle is OracleAncillaryInterface {
```
mrice32, line 18: This is way simpler! +1

mrice32 (commented):
Looks awesome! Minor comments on the contracts.

packages/core/contracts/chainbridge/SinkOracle.sol
```diff
@@ -0,0 +1,117 @@
...
+        // Call Bridge.deposit() to intiate cross-chain price request.
+        _getBridge().deposit(destinationChainID, getResourceId(), _formatMetadata(identifier, time, ancillaryData));
+    }
+
+    /**
+     * @notice This method will ultimately be called after `requestPrice` calls `Bridge.deposit()`, which will call
+     * `GenericHandler.deposit()` and ultimately this method.
+     * @dev This method should basically check that the `Bridge.deposit()` was triggered by a valid price request,
+     * specifically one that has not resolved yet and was called by a registered contract. Without this check,
+     * `Bridge.deposit()` could be called by anybody in order to emit excess Deposit events.
```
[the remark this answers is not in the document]
nicholaspai: Without the check that the price request was actually made `require(lookup.state == RequestState.Requested)`, then its possible that a non-registered contract or entity could call `Bridge.deposit()` in order to trigger a DVM price request. This check makes sure that the price request was actually made by a registered contract.

packages/core/contracts/chainbridge/SourceOracle.sol
nicholaspai: We deploy a `Source` on Mainnet (or any other EVM that has a DVM on it) which can make price requests to and fetch price resolution data from the DVM. And we deploy one `Sink` per L2 that wants to use the `Source` as the source of truth.

> we should just be aware that someone can forward price results to any other chain if they want, and those will fail to push on the other side (because they were never requested). Not sure if this is gracefully handled by the client.

Great point. We actually configure the off-chain relayer with a `config.json` listing out all networks it should service, so its not a problem if the `destinationChainId` doesn't point to a real network:

```
{
  "chains": [
    {
      "name": "Goerli",
      "type": "ethereum",
      "id": "0",
      "endpoint": "wss://goerli-light.eth.linkpool.io/ws",
      "from": "0x",
      "opts": {
        "bridge": "0x",
        "genericHandler": "0x"
      }
    },
    {
      "name": "Rinkeby",
      "type": "ethereum",
      "id": "1",
      "endpoint": "wss://rinkeby-light.eth.linkpool.io/ws",
      "from": "0xrelayer",
      "opts": {
        "bridge": "0x",
        "genericHandler": "0x"
      }
    }
  ]
}
```

packages/core/contracts/chainbridge/BeaconOracle.sol
nicholaspai: Than what?

packages/core/contracts/chainbridge/SourceOracle.sol
nicholaspai: Ah great catch.

packages/core/contracts/chainbridge/BeaconOracle.sol
mrice32: Sorry, I just meant the entire contract is way simpler without the need to be able to list all the requests (the refactor you did!).

packages/core/contracts/oracle/implementation/cross-chain/SinkOracle.sol
mrice32: Makes sense!

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: +1

packages/core/contracts/oracle/implementation/cross-chain/BeaconOracle.sol
mrice32: +1

packages/core/contracts/chainbridge/SourceOracle.sol
mrice32: What if we had 3 networks, two L2s (id 2 and 3) and 1 L1 (id 1)? What would happen if id 2 requests a price, and someone pushes the price to id 2 and id 3? What would happen when the bridge tries to push the price to id 3?

Also, what would happen if someone tried to publish it to id 1 (the same as the source chain)?

mrice32 (commented):
Looks great! One more comment.

packages/core/contracts/chainbridge/SourceOracle.sol
nicholaspai: > What would happen if id 2 requests a price, and someone pushes the price to id 2 and id 3? What would happen when the bridge tries to push the price to id 3?

The `publishPrice` on the id 3 `SinkOracle` would revert because there was no request pending.

> Also, what would happen if someone tried to publish it to id 1 (the same as the source chain)?

This has to happen on L1 to start the cross-chain communication. Someone can call `publishPrice` on L1 to trigger a `Bridge.deposit` call on L1. No one can directly call `publishPrice` on L2 except for the `GenericHandler` which would be called by `Bridge.deposit` on L2.

Do these answers make sense and do you agree with them?

packages/core/contracts/chainbridge/BeaconOracle.sol
```diff
@@ -0,0 +1,164 @@
...
+    /**
+     * @notice Returns resolved price for the request.
+     * @return int256 Price, or reverts if no resolved price for any reason.
+     */
+
+    function getPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public view override onlyRegisteredContract() returns (int256) {
```
mrice32, line 100: Just to double check -- we would never really use these functions on the source side right? Because the source can't request prices for itself, right?

mrice32 (approved):
LGTM!

packages/core/contracts/chainbridge/BeaconOracle.sol
nicholaspai: well I suppose you could call this from the source side, but it'd be only worst than calling the DVM directly.

mrice32 (approved):
LGTM!

PR #2913, opened 2021-04-27
fix(core,affiliates): log warning to console.error, make json piping …

### LINKED ISSUE #2912: json output between affiilates process broken due to console log in core
**Description**

FindContractVersions logs a warning if you have an out of date core build. This is logged to console.log, but this breaks output for affiliates piping. Should be logged to console.error instead.

daywiss:
…easier to find issues

**Motivation**

#2912 

**Summary**

Simply change warning log to console.error, so that it does not interfere with affiliates piping. In general warnings like this should log to error to prevent unexpected output to standard out.  Related change to affiliates pipe handler to log full output before throwing error, to ease debugging.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #2912

### DIFF
--- a/packages/affiliates/libs/affiliates/utils.js
+++ b/packages/affiliates/libs/affiliates/utils.js
@@ -362,7 +362,14 @@ function makeUnixPipe(cb, stdin = process.stdin) {
         return result + str;
       })
       // once string is final, we try to parse it as json. We assume that our caller is passing us a valid json string.
-      .map(x => JSON.parse(x))
+      .map(x => {
+        try {
+          return JSON.parse(x);
+        } catch (err) {
+          console.error(x);
+          throw err;
+        }
+      })
       // once we have the json object we call the callback passing it as a param
       .map(async x => cb(x))
       // we need to do this to extract the result of the promise in the stream.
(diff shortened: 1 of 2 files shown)


nicholaspai (approved):

packages/affiliates/libs/affiliates/utils.js
```diff
@@ -362,7 +362,14 @@ function makeUnixPipe(cb, stdin = process.stdin) {
         return result + str;
       })
       // once string is final, we try to parse it as json. We assume that our caller is passing us a valid json string.
-      .map(x => JSON.parse(x))
+      .map(x => {
+        try {
+          return JSON.parse(x);
+        } catch (err) {
+          console.error(x);
```
mrice32, line 369: nit: why are we logging this error and rethrowing? Seems unrelated to the point of this PR and it also seems a bit strange to log the error that you're bubbling up? Won't it get printed twice?

mrice32 (approved):
Looks good -- one minor comment!

packages/affiliates/libs/affiliates/utils.js
daywiss: its actually not logging the error, its logging the object which triggered the parsing error. The reason is because the error only gives us a line within the object to refer to, but its useless without being able to actually see the object!

PR #2918, opened 2021-04-28
feat(price-feeds): add uniswap v3 price feed

mrice32:
**Motivation**

We need a price feed to track prices on a uniswap v3 pool.

**Summary**

Adds a uniswap v3 price feed that mostly reuses logic from the v2 feed.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -23,6 +23,7 @@
     "winston-transport": "^4.3.0"
   },
   "devDependencies": {
+    "bignumber.js": "^9.0.1",
     "@tsconfig/node14": "^1.0.0",
     "typescript": "^4.1.3"
   },
--- a/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
@@ -146,7 +146,7 @@ class UniswapPriceFeed extends PriceFeedInterface {
       // By taking larger powers of 2, this doubles the lookback each time.
       fromBlock = Math.max(0, latestBlockNumber - lookbackBlocks * 2 ** i);
 
-      const newEvents = await this._getSortedSyncEvents(fromBlock, toBlock).then(newEvents => {
+      const newEvents = await this._getSortedEvents(fromBlock, toBlock).then(newEvents => {
         // Grabs the timestamps for all blocks, but avoids re-querying by .then-ing any cached blocks.
         return Promise.all(
           newEvents.map(event => {
@@ -160,7 +160,7 @@ class UniswapPriceFeed extends PriceFeedInterface {
             // Add a .then to the promise that sets the timestamp (and price) for this event after the promise resolves.
             return this.blocks[event.blockNumber].then(block => {
               event.timestamp = block.timestamp;
-              event.price = this._getPriceFromSyncEvent(event);
+              event.price = this._getPriceFromEvent(event);
               return event;
             });
           })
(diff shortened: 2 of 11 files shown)


packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -227,15 +244,47 @@ class UniswapPriceFeed extends PriceFeedInterface {
       return reserve1.mul(fixedPointAdjustment).div(reserve0);
     }
   }
+}
 
-  _computeTwap(eventsIn, startTime, endTime) {
-    const events = eventsIn.map(e => {
-      return [e.timestamp, e.price];
+class UniswapV3PriceFeed extends UniswapPriceFeed {
```
chrismaree, line 249: this is super clean. I like that it's also within the same file as most of the logic is re-used. very nice implementation!

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
```diff
@@ -227,15 +244,47 @@ class UniswapPriceFeed extends PriceFeedInterface {
...
+      parseFixed("1", this.invertPrice ? this.token1Precision : this.token0Precision).toString()
+    );
+
+    const X96Adjustment = this.toBN(2).pow(this.toBN(96));
+    const rawSqrtPrice = this.toBN(event.returnValues.sqrtPriceX96);
+    // This effectively computes the price by squaring the value, multiplying it up by our intended precision, then dividing out both X96 fixed point multipliers.
+    const nonInvertedPrice = rawSqrtPrice
+      .mul(rawSqrtPrice)
+      .mul(fixedPointAdjustment)
+      .div(X96Adjustment)
```
chrismaree, line 280: this double multiplication and division seems overly complex. why don't you just use `pow`? for power? In my experimentation on the range trader, I simply had a method like this
```
function decodePriceSqrt(priseSqrt) {
  return new bn(priseSqrt).dividedBy(new bn(2).pow(96)).pow(2);
}
```

chrismaree (approved):
This is pretty awesome stuff! Really clean implementation. I had a few nits and comments but overall looks great!

packages/financial-templates-lib/package.json
```diff
@@ -23,6 +23,7 @@
     "winston-transport": "^4.3.0"
   },
   "devDependencies": {
+    "bignumber.js": "^9.0.1",
```
nicholaspai, line 26: Is this an accidental import?

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
mrice32: I don't use bignumber.js (handles decimals) to avoid having to convert back and forth and to avoid adding an extra runtime dependency.

Without using it, we have to multiply up before dividing to preserve precision. However, we could switch the double multiplies with using pow, but I'm not sure how much simpler it would be. Here's a sample of that implementation:
```
rawSqrtPrice.pow(this.toBN(2)).mul(fixedPointAdjustment).div(this.toBN(2).pow(192));
```

Personally, I don't find this to be much simpler or any more concise (even if you were to sub in X96Adjustment instead of doing the pow). I find it a bit more difficult to decode what's going on. However, if you find something like this to be easier to read and interpret I'm happy to change, as I may be in the minority since I may be biased!

packages/financial-templates-lib/package.json
mrice32: Nope, bignumber.js was included as a dev dep so we could use it in the tests (I pulled a function from uniswap's tests where they use it to convert prices). Would you prefer that we drop it in favor of just using web3's native BN?

nicholaspai: nope just double checking

nicholaspai (approved):
Really like how you're making both Univ2 and Univ3 inherit from the same parent class, makes it a lot easier to reason about and maintain going forward.

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
chrismaree: ye, it's a good point. having to pull in big number.js is not ideal. I'm ok with leaving it how you've got it here to avoid that dependency within the price feed.

In the unit tests I've pulled it in but there is also a fair amount of this kind of computation there so I feel ok with having it in that context.

packages/financial-templates-lib/package.json
chrismaree: If we _do_ keep it then you can change the `decodePriceSqrt` computation :)

packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.js
mrice32: Yep, and same, I pulled it into dev deps for unit tests, but wanted to avoid having it installed with the package if it could be avoided.

PR #2920, opened 2021-04-28
fix: Alternate rounding implementation to address GCP logging issue

nicholaspai:
**Motivation**

The `round(x)` element of the expression used to construct the `ExpressionPriceFeed` introduced in [this PR](https://github.com/UMAprotocol/protocol/commit/22017d22d2ad3cc8c43c898b14fda47554012181) causes GCP logging to hang when [this code path](https://github.com/UMAprotocol/protocol/blob/master/packages/funding-rate-proposer/src/proposer.js#L147) in the `funding-rate-proposer` is entered. This is a point where the `ExpressionPriceFeed` has been constructed and updated, but no price is fetched. The GCP logging does NOT hang when either a funding rate is proposed, or the funding rate to propose is within some margin of error. This is strange behavior and we're not sure yet why this breaks, but removing the `round` from the expression empirically works locally.

This alternative implementation of rounding also causes GCP logging to succeed. It ensures that the same rounding is applied in all bots that push prices on-chain: `optimistic-oracle-proposer`, `funding-rate-proposer`, `liquidator`, and `disputer`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/helpers/math.js
+++ b/packages/financial-templates-lib/src/helpers/math.js
@@ -1,5 +1,7 @@
 const assert = require("assert");
 
+const { parseFixed, formatFixed, getRoundingForIdentifier } = require("@uma/common");
+
 // Takes in two big numbers and returns the error between them. using: δ = (observed - expected) / expected
 // For example an observed price of 1.2 with an expected price of 1.0 will return (1.2 - 1.0) / 1.0 = 0.20
 // This is equivalent of a 20 percent deviation between the numbers.
@@ -31,7 +33,26 @@ const isDeviationOutsideErrorMargin = (observedValueBN, expectedValueBN, scaling
   return deviationBN.abs().gt(errorMarginBN);
 };
 
+const formatPriceToPricefeedPrecision = (price, priceFeedDecimals, identifier) => {
+  if (!getRoundingForIdentifier(identifier)) {
+    return price.toString();
+  } else {
+    // Round `price` to custom number of decimals by converting back and forth between the pricefeed's
+    // configured precision:
+    return parseFixed(
+      // 1) `formatFixed` converts the price in wei to a floating point.
+      // 2) `toFixed` removes decimals beyond `this.precision` in the floating point.
+      // 3) `parseFixed` converts the floating point back into wei.
+      Number(formatFixed(price.toString(), priceFeedDecimals))
+        .toFixed(getRoundingForIdentifier(identifier))
+        .toString(),
+      priceFeedDecimals
+    ).toString();
+  }
+};
+
 module.exports = {
   calculateDeviationError,
-  isDeviationOutsideErrorMargin
+  isDeviationOutsideErrorMargin,
+  formatPriceToPricefeedPrecision
 };
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -859,7 +859,7 @@ const defaultConfigs = {
     type: "expression",
     expression: `
         ETHBTC_FV = ETH\\/BTC * PERP_FRM;
-        round(max(-0.00001, min(0.00001, (ETHBTC_FV - ETHBTC_PERP) / ETHBTC_FV / 86400)), 9)
+        max(-0.00001, min(0.00001, (ETHBTC_FV - ETHBTC_PERP) / ETHBTC_FV / 86400))
     `,
     lookback: 7200,
     minTimeBetweenUpdates: 60,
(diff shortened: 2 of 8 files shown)


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -859,7 +859,7 @@ const defaultConfigs = {
     type: "expression",
     expression: `
         ETHBTC_FV = ETH\\/BTC * PERP_FRM;
-        round(max(-0.00001, min(0.00001, (ETHBTC_FV - ETHBTC_PERP) / ETHBTC_FV / 86400)), 9)
```
nicholaspai, line 862: Removing this `round` fixes the GCP issue.

chrismaree: This is so wack. I dont understand how this could cause GCP to blow out like this. 

One explanation could be that when you've been testing this locally you are running it on your Mac within your dev environment. GCP is running it within a docker container. it is possible that there is some execution difference between the two that breaks this js math logic?

packages/financial-templates-lib/src/helpers/math.js
```diff
@@ -31,7 +33,26 @@ const isDeviationOutsideErrorMargin = (observedValueBN, expectedValueBN, scaling
   return deviationBN.abs().gt(errorMarginBN);
 };
 
+const formatPriceToPricefeedPrecision = (price, priceFeedDecimals, identifier) => {
```
chrismaree, line 36: this makes a lot of sense as a method and is something we should have actually done a while ago. Previously we had a few UMIP implementations that we were actually not conforming to as a result of not having something like this, such as within the liquidator/disputer. It only really makes a difference on the margin though (or in the case of the funding rate related bots!).

chrismaree (approved):
It blows my mind that this fixes things. really don't understand why this was breaking in production and these changes sorted it out 🤷 .

In any case, I think this is a good thing to add as we are more accurately implementing UMIPs by doing this kind of rounding for identifiers. 

Great work on debugging and solving this @nicholaspai !

mrice32:
> It blows my mind that this fixes things. really don't understand why this was breaking in production and these changes sorted it out 🤷 .
> 
> In any case, I think this is a good thing to add as we are more accurately implementing UMIPs by doing this kind of rounding for identifiers.
> 
> Great work on debugging and solving this @nicholaspai !

@nicholaspai awesome work on this, and very good work on finding and implementing a fix so quickly!!

However, I see this solution as a stop-gap rather than the ideal implementation of rounding.

I think manually maintaining a mapping and adding custom, repeated logic to each bot to handle rounding is more bloat and complexity than we should have just to round prices.

For all of our identifiers I think this should be implemented at the price feed level, so each bot would get it automatically, and it is specified in the same spot as the rest of the price feed (the default price feeds file) implementation. I don't really see rounding as a separate thing from the rest of the identifier spec.

I'm hoping that we can move this logic back into the price feed down the road!

mrice32 (approved):

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
nicholaspai: I've been running the spoke on GCP actually, and I can replicate the GCP logging bug unless I remove this `round`

nicholaspai:
> > It blows my mind that this fixes things. really don't understand why this was breaking in production and these changes sorted it out 🤷 .
> > In any case, I think this is a good thing to add as we are more accurately implementing UMIPs by doing this kind of rounding for identifiers.
> > Great work on debugging and solving this @nicholaspai !
> 
> @nicholaspai awesome work on this, and very good work on finding and implementing a fix so quickly!!
> 
> However, I see this solution as a stop-gap rather than the ideal implementation of rounding.
> 
> I think manually maintaining a mapping and adding custom, repeated logic to each bot to handle rounding is more bloat and complexity than we should have just to round prices (despite that being the only working solution we have).
> 
> For all of our identifiers I think this should be implemented at the price feed level, so each bot would get it automatically, and it is specified in the same spot as the rest of the price feed (the default price feeds file) implementation. I don't really see rounding as a separate thing from the rest of the identifier spec.
> 
> I'm hoping that we can move this logic back into the price feed down the road!
> 
> And obviously, this isn't a criticism of your implementation @nicholaspai -- just wanted to note that I think we're going with a less-ideal pattern to avoid this bug that we are unable to completely understand.
> 
> cc @chrismaree

I agree, which is why I'm hesitant to merge. Will give it 24 hours to try to debug more before we fallback to this implementation.

nicholaspai:
Closing in favor of #2922

PR #2923, opened 2021-04-30
feat(affiliates): uniswap v3 fetch state proof of concept

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/563/interface-with-uniswap-v3-to-get-lp-positions

**Summary**
This mainly shows how we can relevenat uniswap v3 liquidity mining state. pools, positions and ticks through events and contract queries.  This will be useful for determining reward values for liquidity mining.  This is very much a work in progress, but its a good checkpoint before the code gets too large. 

**Details**

You will notice this uses ethers, this is mainly becuase I was building typechain files from uniswap and it built to ethers, so all types were based on this. This could probably be switched to web3 with a bit of work.

Most of the logic is split between 3 files:
- `scripts/univ3` - this shows how to fetch events and state from contracts and put them into models
- `libs/uniswap/models` - these are uniswap v3 specific data, it includes `Positions`, `NftPositions`, `Ticks` and `Pools`. Most of the logic is the same across tables.
- `libs/uniswap/processors` - these are the event handlers which handle specific contract events and fill up the models: `NftEvents`,  `PoolEvents`, `PoolFactory`.

**Remaining Challenges**
- still need to be able to get this state for arbitrary block numbers 
- break down uniswap liquidity and sqrtPrice into virtual X/Y amounts
- use virtual liquidity to calcluate a USD value for either positions or pools
- might want to be able to compare pools to each other, this requires usd price normalization
- add a flexible weighting or scaling system so that we can experiment with per pool reward weighting

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/563/interface-with-uniswap-v3-to-get-lp-positions

### DIFF
--- a/packages/affiliates/libs/uniswap/models.ts
+++ b/packages/affiliates/libs/uniswap/models.ts
@@ -1,5 +1,5 @@
 import assert from "assert";
-import { exists } from "./utils";
+import { exists, getPositionKey } from "./utils";
 import { ethers } from "ethers";
 
 type CacheType = {
@@ -17,7 +17,7 @@ export const Cache = (table = new Map<string | number, CacheType>()) => {
     list
   };
 };
-type NftPosition = {
+export type NftPosition = {
   id?: string;
   tokenId: string;
   amount?: string;
@@ -74,12 +74,11 @@ export const NftPositions = (config: any, table: ReturnType<typeof Cache>) => {
     list: table.list
   };
 };
-type Position = {
+export type Position = {
   id?: string;
   operator: string;
   tickLower: string;
   tickUpper: string;
-  nonce?: string;
   token0?: string;
   token1?: string;
   liquidity?: string;
@@ -93,30 +92,7 @@ type Position = {
 // These positions are from core pool contracts and are indexed using hash of upper/lower/user address.
 export const Positions = (config: any, table: ReturnType<typeof Cache>) => {
   function makeId(data: Position) {
-    const encoded = ethers.utils.defaultAbiCoder.encode(
-      [
-        ethers.utils.ParamType.from({
-          indexed: true,
-          name: "owner",
-          type: "address"
-        }),
-        ethers.utils.ParamType.from({
-          indexed: true,
-          baseType: "int24",
-          name: "tickLower",
-          type: "int24"
-        }),
-        ethers.utils.ParamType.from({
-          indexed: true,
-          baseType: "int24",
-          name: "tickUpper",
-          type: "int24"
-        })
-      ],
-      [data.operator, data.tickLower, data.tickUpper]
-    );
-    // this is how positions are indexed in the v3 core contract
-    return ethers.utils.keccak256(encoded);
+    return getPositionKey(data.operator, data.tickLower, data.tickUpper);
   }
   async function create(data: Position) {
     const id = makeId(data);
@@ -155,7 +131,7 @@ export const Positions = (config: any, table: ReturnType<typeof Cache>) => {
   };
 };
 
-type GlobalState = {
+export type GlobalState = {
   id?: string;
   token0: string;
   token1: string;
@@ -213,25 +189,25 @@ export const Pools = (config: any, table: ReturnType<typeof Cache>) => {
   };
 };
 
-type TickInfo = {
-  //     // the total position liquidity that references this tick
+export type TickInfo = {
+  // the total position liquidity that references this tick
   liquidityGross?: string;
-  //     // amount of net liquidity added (subtracted) when tick is crossed from left to right (right to left),
+  // amount of net liquidity added (subtracted) when tick is crossed from left to right (right to left),
   liquidityNet?: string;
-  //     // fee growth per unit of liquidity on the _other_ side of this tick (relative to the current tick)
-  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  // fee growth per unit of liquidity on the _other_ side of this tick (relative to the current tick)
+  // only has relative meaning, not absolute — the value depends on when the tick is initialized
   feeGrowthOutside0X128?: string;
   feeGrowthOutside1X128?: string;
-  //     // the cumulative tick value on the other side of the tick
+  // the cumulative tick value on the other side of the tick
   tickCumulativeOutside?: string;
-  //     // the seconds per unit of liquidity on the _other_ side of this tick (relative to the current tick)
-  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  // the seconds per unit of liquidity on the _other_ side of this tick (relative to the current tick)
+  // only has relative meaning, not absolute — the value depends on when the tick is initialized
   secondsPerLiquidityOutsideX128?: string;
-  //     // the seconds spent on the other side of the tick (relative to the current tick)
-  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  // the seconds spent on the other side of the tick (relative to the current tick)
+  // only has relative meaning, not absolute — the value depends on when the tick is initialized
   secondsOutside?: string;
-  //     // true iff the tick is initialized, i.e. the value is exactly equivalent to the expression liquidityGross != 0
-  //     // these 8 bits are set to prevent fresh sstores when crossing newly initialized ticks
+  // true iff the tick is initialized, i.e. the value is exactly equivalent to the expression liquidityGross != 0
+  // these 8 bits are set to prevent fresh sstores when crossing newly initialized ticks
   initialized?: boolean;
   pool?: string;
   index?: string;
--- a/packages/affiliates/libs/uniswap/models.ts
+++ b/packages/affiliates/libs/uniswap/models.ts
@@ -1,0 +1,277 @@
+import assert from "assert";
+import { exists } from "./utils";
+import { ethers } from "ethers";
+
+type CacheType = {
+  [key: string]: string | number | boolean;
+};
+export const Cache = (table = new Map<string | number, CacheType>()) => {
+  function list() {
+    return [...table.values()];
+  }
+  return {
+    has: table.has.bind(table),
+    set: table.set.bind(table),
+    get: table.get.bind(table),
+    delete: table.delete.bind(table),
+    list
+  };
+};
+type NftPosition = {
+  id?: string;
+  tokenId: string;
+  amount?: string;
+  amount0?: string;
+  amount1?: string;
+  token0?: string;
+  token1?: string;
+  liquidity?: string;
+  feeGrowthInside0LastX128?: string;
+  feeGrowthInside1LastX128?: string;
+  tokensOwed0?: string;
+  tokensOwed1?: string;
+  pool?: string;
+};
+
+// Nft positions are gathered from the nft contract, and require slightly different indexing than the positions in pool contract
+export const NftPositions = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(data: NftPosition) {
+    return data.tokenId;
+  }
+  async function create(data: NftPosition) {
+    const id = makeId(data);
+    assert(!(await has(id)), "Position exists");
+    return set({ id, ...data });
+  }
+  async function set(data: NftPosition) {
+    assert(data.id, "Position requires id");
+    await table.set(data.id, { ...data });
+    return data;
+  }
+  async function get(id: string) {
+    assert(await table.has(id), "Position does not exist");
+    return table.get(id);
+  }
+  async function has(id: string) {
+    return table.has(id);
+  }
+  async function forEach(cb: (value: NftPosition, index: number, array: NftPosition[]) => void) {
+    const list: NftPosition[] = (await table.list()) as NftPosition[];
+    list.forEach(cb);
+  }
+  async function update(id: string, data: any = {}) {
+    const position = await get(id);
+    return set({ ...position, ...data });
+  }
+
+  return {
+    create,
+    set,
+    get,
+    has,
+    update,
+    forEach,
+    list: table.list
+  };
+};
+type Position = {
+  id?: string;
+  operator: string;
+  tickLower: string;
+  tickUpper: string;
+  nonce?: string;
+  token0?: string;
+  token1?: string;
+  liquidity?: string;
+  feeGrowthInside0LastX128?: string;
+  feeGrowthInside1LastX128?: string;
+  tokensOwed0?: string;
+  tokensOwed1?: string;
+  pool?: string;
+};
+
+// These positions are from core pool contracts and are indexed using hash of upper/lower/user address.
+export const Positions = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(data: Position) {
+    const encoded = ethers.utils.defaultAbiCoder.encode(
+      [
+        ethers.utils.ParamType.from({
+          indexed: true,
+          name: "owner",
+          type: "address"
+        }),
+        ethers.utils.ParamType.from({
+          indexed: true,
+          baseType: "int24",
+          name: "tickLower",
+          type: "int24"
+        }),
+        ethers.utils.ParamType.from({
+          indexed: true,
+          baseType: "int24",
+          name: "tickUpper",
+          type: "int24"
+        })
+      ],
+      [data.operator, data.tickLower, data.tickUpper]
+    );
+    // this is how positions are indexed in the v3 core contract
+    return ethers.utils.keccak256(encoded);
+  }
+  async function create(data: Position) {
+    const id = makeId(data);
+    assert(!(await has(id)), "Position exists");
+    return set({ id, ...data });
+  }
+  async function set(data: Position) {
+    assert(data.id, "Position requires id");
+    await table.set(data.id, { ...data });
+    return data;
+  }
+  async function get(id: string) {
+    assert(await table.has(id), "Position does not exist");
+    return table.get(id);
+  }
+  async function has(id: string) {
+    return table.has(id);
+  }
+  async function forEach(cb: (value: Position, index: number, array: Position[]) => void) {
+    const list: Position[] = (await table.list()) as Position[];
+    list.forEach(cb);
+  }
+  async function update(id: string, data: any = {}) {
+    const position = await get(id);
+    return set({ ...position, ...data });
+  }
+
+  return {
+    create,
+    set,
+    get,
+    has,
+    update,
+    forEach,
+    list: table.list
+  };
+};
+
+type GlobalState = {
+  id?: string;
+  token0: string;
+  token1: string;
+  fee: string;
+  sqrtPriceX96?: string;
+  tick?: number;
+  liquidity?: string;
+  feeProtocol?: number;
+  feeGrowthGlobal0X128?: string;
+  feeGrowthGlobal1X128?: string;
+  protocolFeesToken0?: string;
+  protocolFeesToken1?: string;
+  address?: string;
+};
+// pool id is tokenA, tokenB, fee
+export const Pools = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(state: GlobalState) {
+    const { token0, token1, fee } = state;
+    if (token0 < token1) {
+      return [token0, token1, fee].join("!");
+    }
+    return [token1, token0, fee].join("!");
+  }
+  async function create(state: GlobalState) {
+    const id = makeId(state);
+    assert(!(await has(id)), "State already exists:" + id);
+    return set({ id, ...state });
+  }
+  async function set(state: GlobalState) {
+    assert(state.id, "requires id");
+    await table.set(state.id, state);
+    return state;
+  }
+  async function get(id: string) {
+    assert(await has(id), "No such pool state");
+    return table.get(id);
+  }
+  async function has(id: string) {
+    return table.has(id);
+  }
+  async function update(id: string, data: any = {}) {
+    const pool = await get(id);
+    return set({ ...pool, ...data });
+  }
+  async function list() {
+    return table.list();
+  }
+  return {
+    create,
+    set,
+    get,
+    has,
+    update,
+    list
+  };
+};
+
+type TickInfo = {
+  //     // the total position liquidity that references this tick
+  liquidityGross?: string;
+  //     // amount of net liquidity added (subtracted) when tick is crossed from left to right (right to left),
+  liquidityNet?: string;
+  //     // fee growth per unit of liquidity on the _other_ side of this tick (relative to the current tick)
+  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  feeGrowthOutside0X128?: string;
+  feeGrowthOutside1X128?: string;
+  //     // the cumulative tick value on the other side of the tick
+  tickCumulativeOutside?: string;
+  //     // the seconds per unit of liquidity on the _other_ side of this tick (relative to the current tick)
+  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  secondsPerLiquidityOutsideX128?: string;
+  //     // the seconds spent on the other side of the tick (relative to the current tick)
+  //     // only has relative meaning, not absolute — the value depends on when the tick is initialized
+  secondsOutside?: string;
+  //     // true iff the tick is initialized, i.e. the value is exactly equivalent to the expression liquidityGross != 0
+  //     // these 8 bits are set to prevent fresh sstores when crossing newly initialized ticks
+  initialized?: boolean;
+  pool?: string;
+  index?: string;
+  id?: string;
+};
+export const Ticks = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(state: TickInfo) {
+    const { pool, index } = state;
+    return [pool, index].join("!");
+  }
+  async function create(state: TickInfo) {
+    const id = makeId(state);
+    assert(!(await has(id)), "tick already exists:" + id);
+    return set({ id, ...state });
+  }
+  async function set(state: TickInfo) {
+    assert(state.id, "requires id");
+    await table.set(state.id, state);
+    return state;
+  }
+  async function get(id: string) {
+    assert(await has(id), "No such tick state");
+    return table.get(id);
+  }
+  async function has(id: string) {
+    return table.has(id);
+  }
+  async function update(id: string, data: any = {}) {
+    const got = await get(id);
+    return set({ ...got, ...data });
+  }
+  async function list() {
+    return table.list();
+  }
+  return {
+    create,
+    set,
+    get,
+    has,
+    update,
+    list
+  };
+};
--- a/packages/affiliates/libs/uniswap/models.ts
+++ b/packages/affiliates/libs/uniswap/models.ts
@@ -1,21 +1,20 @@
 import assert from "assert";
 import { exists, getPositionKey } from "./utils";
-import { ethers } from "ethers";
 
-type Id = string | number;
-type HasId = {
+type Id = string;
+type MaybeId = {
   id?: Id;
 };
 interface MakeId<T> {
-  (obj: T): string;
+  (obj: T): Id;
 }
 
 // Generic table with common functions, store is a map now, but we could replace this with any kind of
 // data store with some minor modifications, as long as it has, set, get, delete, has and values
-export function Table<T extends HasId>(config: { makeId: MakeId<T>; type: string }, store: Map<Id, T>) {
+export function Table<T extends MaybeId>(config: { makeId: MakeId<T>; type: string }, store: Map<Id, T>) {
   const { makeId, type } = config;
   async function create(data: T) {
-    const id = makeId(data);
+    const id = exists(data.id) ? data.id : makeId(data);
     assert(!(await has(id)), `${type} exists`);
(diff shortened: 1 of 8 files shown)


chrismaree (commented):
This is pretty awesome! I really like the pattern you've set up here. it's _very_ expressive.

My main feedback, which should not block this PR, is that this almost seems _overly verbose. you've taken the strategy of defining pretty much every type we could ever need, even if the data is not needed right now. I dont really have a strong issue with this; however, it likely introduced a bunch more lines of code than what will actually be needed. What do you think?

packages/affiliates/scripts/univ3.js
```diff
@@ -0,0 +1,186 @@
...
+      quoterAddress: "0x539BF58f052dE91ae369dAd59f1ac6887dF39Bc5",
+      swapRouter: "0xbBca0fFBFE60F60071630A8c80bb6253dC9D6023",
+      nonfungibleTokenPositionDescriptorAddress: "0xc4b81504F9a2bd6a6f2617091FB01Efb38D119c8",
+      descriptorProxyAddress: "0xDbe2c61E85D06eaA6E7916049f38B93288BA30f3",
+      nonfungibleTokenPositionManagerAddress: "0xd3808aBF85aC69D2DBf53794DEa08e75222Ad9a1",
+      v3MigratorAddress: "0x9dF511178D1438065F7672379414F5C46D5B51b4"
+    }
+  ],
+  [
+    "rinkeby",
```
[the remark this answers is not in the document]
daywiss: when i started this, i didnt see any activity on kovan, and also theres a subgraph and activity on rinkeby, so i ended up doing my testing there.

packages/affiliates/libs/uniswap/processor.ts
```diff
@@ -0,0 +1,249 @@
...
+  const handlers: { [key: string]: (...args: any[]) => Promise<void> } = {
+    async Burn(
+      owner: string | null,
+      tickLower: BigNumberish | null,
+      tickUpper: BigNumberish | null,
+      amount: null,
+      amount0: null,
+      amount1: null
+    ) {
+      // nothing
```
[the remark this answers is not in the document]
daywiss: yea theres no point other than to document the events and parameters, i can clean this up

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+    create,
+    set,
+    get,
+    has,
+    update,
+    list
+  };
+};
+
+export type TickInfo = {
```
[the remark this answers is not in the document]
daywiss: the main concern with allowing BNs in here is that its not compatible between web3 numbers and ethers afaik.  slightly less important is that you cant serialize BigNumbers or BNs to a database and while not needed for this, you never know when that could be handy.  You are right it would be more performant to just choose one and go with it though since everything is in memory.

daywiss:
> My main feedback, which should not block this PR, is that this almost seems _overly verbose. you've taken the strategy of defining pretty much every type we could ever need, even if the data is not needed right now. I dont really have a strong issue with this; however, it likely introduced a bunch more lines of code than what will actually be needed. What do you think?

Initially I just wanted to have all the data and interfaces available to kind of understand things, and make sure I could retrieve it correctly. So yea you are right, most of this wont be needed.  I see a path now that cuts out like 80% of this code, and I will remove it accordingly.

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+
+type CacheType = {
+  [key: string]: string | number | boolean;
+};
+export const Cache = (table = new Map<string | number, CacheType>()) => {
+  function list() {
+    return [...table.values()];
+  }
+  return {
+    has: table.has.bind(table),
```
nicholaspai, line 13: I wasn't sure exactly how `bind` works but found this article about how `bind` loses some type safety, maybe just use slightly more verbose arrow functions instead? https://hamednourhani.gitbooks.io/typescript-book/content/docs/tips/bind.html

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+  feeGrowthInside1LastX128?: string;
+  tokensOwed0?: string;
+  tokensOwed1?: string;
+  pool?: string;
+};
+
+// Nft positions are gathered from the nft contract, and require slightly different indexing than the positions in pool contract
+export const NftPositions = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(data: NftPosition) {
+    return data.tokenId;
```
nicholaspai, line 39: Why do you have both an `id` and `tokenId` in the `NftPosition` type? `tokenId` is required and is used to construct a new `NftPosition`, so in this case `id` seems redundant?

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+    get,
+    has,
+    update,
+    forEach,
+    list: table.list
+  };
+};
+export type Position = {
+  id?: string;
+  operator: string;
```
nicholaspai, line 79: Same question above for `tableId vs id` but this time with `operator and id`. Maybe there's something about `typescript` i'm missing.

packages/affiliates/libs/uniswap/processor.ts
```diff
@@ -0,0 +1,249 @@
...
+    },
+    async DecreaseLiquidity(
+      tokenId: BigNumberish | null,
+      liquidity: BigNumberish | null,
+      amount0: BigNumberish | null,
+      amount1: BigNumberish | null
+    ) {
+      assert(exists(tokenId), "requires token id");
```
nicholaspai, line 37: Instead of `asserting`  that these params are not `null`, why can't you just remove the `| null` from the type?

packages/affiliates/libs/uniswap/processor.ts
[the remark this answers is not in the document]
nicholaspai: I lean towards removing unused interfaces. The benefit of having them is to demonstrate a template for future code features, but you already have some events that you DO parse like `Swap` that give enough of an example for how to process an event into a model

packages/affiliates/libs/uniswap/processor.ts
```diff
@@ -0,0 +1,249 @@
...
+          amount0: amount0.toString(),
+          amount1: amount1.toString()
+        });
+      }
+    },
+    async Transfer(from: string | null, to: string | null, tokenId: BigNumberish | null) {
+      // nothing
+    }
+  };
+  async function handleEvent(event: Event) {
```
nicholaspai, line 86: Since `handleEvent` is re-used, can you refactor into a common lib that each of `NftEvents`, `PoolEvents` and `PoolFactory` can use?

packages/affiliates/scripts/univ3.js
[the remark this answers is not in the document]
nicholaspai: I'm surprised these addresses aren't available from the @uniswap package

nicholaspai (approved):
Looks good to me and I see no risk to merging this as is. However I had some nit stylistic questions

packages/affiliates/libs/uniswap/models.ts
daywiss: interesting, yea sounds good

packages/affiliates/libs/uniswap/processor.ts
daywiss: This was actually generated by typechain based on the solidity code. I found it weird that these values could be null, and maybe in practice they would never be? There were instances where the generated typescript was just null for some reason, so i think maybe its not bullet proof and could just remove the nulls

packages/affiliates/libs/uniswap/models.ts
daywiss: you are right, i was keeping it as "id" more as a convention to say you can always trust .id to be the canonical way to lookup any model, where as tokenId would require context about the data.

packages/affiliates/scripts/univ3.js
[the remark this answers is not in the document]
daywiss: the only place i found this was in a readme in periphery https://github.com/Uniswap/uniswap-v3-periphery/blob/v1.0.0-beta.23/testnet-deploys.md

packages/affiliates/libs/uniswap/processor.ts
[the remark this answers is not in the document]
daywiss: yea good point, will remove them

packages/affiliates/libs/uniswap/models.ts
daywiss: position is slightly more complex, its a hash of operator, lower tick and upper tick. in this case you have to derive the key, which you can see in the utils file:
```js
export function getPositionKey(address: string, lowerTick: string, upperTick: string): string {
  return ethers.utils.keccak256(
    ethers.utils.solidityPack(["address", "int24", "int24"], [address, lowerTick, upperTick])
  );
}
```

I think in this case it makes sense to have an id

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
+import assert from "assert";
+import { exists, getPositionKey } from "./utils";
+import { ethers } from "ethers";
+
+type CacheType = {
+  [key: string]: string | number | boolean;
+};
+export const Cache = (table = new Map<string | number, CacheType>()) => {
```
mrice32, line 8: It seems like you're using this in places where you're storing a single, well-defined type in the table. Have you considered making this type a generic?

For a function:
```ts
export function Cache<KeyType, ValueType>(table = new Map<KeyType, ValueType>) {
  // build your return type based on KeyType and ValueType defined inside this function.
}
```

For a class (as suggested below):
```ts
class Cache<KeyType, ValueType> extends Map<KeyType, ValueType> {
  // Not sure if you'll have to tell ts the return type here.
  list() {
    return [...this.values()];
  }
}
```

mrice32 (commented):
A few comments to start. Will review more later today!

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -0,0 +1,33 @@
...
+  const num = BigInt(sqrtPriceX96) * BigInt(sqrtPriceX96);
+  const denom = Q192;
+  const price1 = BigInt(num) / BigInt(denom);
+  const price0 = 1n / price1;
+  return [price0.toString(), price1.toString()];
+}
+
+// check if a value is not null or undefined, useful for numbers which could be 0
+/* eslint-disable-next-line @typescript-eslint/ban-types */
+export function exists(value: any): value is {} {
```
mrice32, line 15: what is this syntax `value is {}`?

mrice32 (commented):
A few more comments! LG!

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+  function list() {
+    return [...table.values()];
+  }
```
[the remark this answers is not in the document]
daywiss: interesting... thats awesome, i will try it.

packages/affiliates/libs/uniswap/models.ts
daywiss: yea this is a really good suggestion. I have a version of this where i tried this and seems to be working

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,253 @@
...
+  pool?: string;
+  index?: string;
+  id?: string;
+};
+export const Ticks = (config: any, table: ReturnType<typeof Cache>) => {
+  function makeId(state: TickInfo) {
+    const { pool, index } = state;
+    return [pool, index].join("!");
+  }
+  async function create(state: TickInfo) {
```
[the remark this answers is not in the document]
daywiss: yea this is a really good suggestion. I have a version of this where i tried this and seems to be working

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -0,0 +1,33 @@
+import assert from "assert";
+import { ethers } from "ethers";
+// taken from uniswap subgraph, not sure if correct yet
+const Q192: BigInt = 2n ** 192n;
+export function sqrtPriceX96ToTokenPrices(sqrtPriceX96: BigInt): string[] {
+  const num = BigInt(sqrtPriceX96) * BigInt(sqrtPriceX96);
+  const denom = Q192;
+  const price1 = BigInt(num) / BigInt(denom);
+  const price0 = 1n / price1;
```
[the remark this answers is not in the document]
daywiss: youre right, i dont think this math works just yet. It was copied from uniswap but they were converting to floats and i was in the process of translating

packages/affiliates/libs/uniswap/utils.ts
daywiss: actually spent quite a bit of time researching how to represent a value which is not null and not undefined in typescript. This `{}` i think is the closest representation of this that i found, but for some reason is disallowed by default in eslint. 

from a [stackoverflow](https://stackoverflow.com/questions/60381643/what-does-mean-in-typescript):

> So {} imposes almost no constraints on its values; the only rule is that it can't be null or undefined. It is similar to the type object in this regard, except that object also forbids primitive values, while {} allows them

daywiss: and i just realized you are asking about `value is {}`, i dont remember actually why this had to be there, but i had trouble getting this function to compile without it. will look into it some more

daywiss: Heres why https://stackoverflow.com/questions/40081332/what-does-the-is-keyword-do-in-typescript

Basically im using it as a guard, and typescript wont be able to tell that its guarding againts null or undefined, and will keep erroring unless you use 'is'

packages/affiliates/libs/uniswap/models.ts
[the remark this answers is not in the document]
daywiss: ended up just using a map directly, much simpler

daywiss:
i've cleaned up event handlers based on nicks suggestions, and simplified the types based on matts suggestions. converted the script to use typescript as well. will go through and remove more code in next PR as needed based on chris's comments.

packages/affiliates/libs/uniswap/utils.ts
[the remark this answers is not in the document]
daywiss: im going to comment this out for now, and address it in next PR. might not need it at all.

packages/affiliates/libs/uniswap/utils.ts
mrice32: Very cool!

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
...
+  tokensOwed1?: string;
+  pool?: string;
+};
+
+// Nft positions are gathered from the nft contract, and require slightly different indexing than the positions in pool contract
+export const NftPositions = () => {
+  function makeId(data: NftPosition) {
+    return data.tokenId;
+  }
+  const store = new Map<Id, NftPosition>();
```
mrice32, line 78: nit: why construct this outside? Why not construct this within the Table constructor since it's the same every time?

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
...
+  liquidity?: string;
+  feeGrowthInside0LastX128?: string;
+  feeGrowthInside1LastX128?: string;
+  tokensOwed0?: string;
+  tokensOwed1?: string;
+  pool?: string;
+};
+
+// Nft positions are gathered from the nft contract, and require slightly different indexing than the positions in pool contract
+export const NftPositions = () => {
```
mrice32, line 74: If you overload the makeId function, you might be able to avoid this function entirely (although it may not be cleaner).
```ts
// 
function makeId(data: NftPosition, type: "NftPositions");
function makeId(data: Position, type: "Positions");
function makeId(data: any, type: string) {
  if (type === "NftPositions") {
    // Typescript should (?) know that data is an NftPosition.
  else if (type === "Positions") {
    // Same for Positions.
  }
```

https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads. Note: the implementation function is not considered a "valid" signature, meaning that you can't call it directly -- you have to use one of the overloads -- https://www.typescriptlang.org/docs/handbook/2/functions.html#overload-signatures-and-the-implementation-signature.

Not sure if you want to do it this way, but it might mean you can just define this overload and call it within the constructor of the Table and then not have to have any other code that's specific to these interfaces.

mrice32 (commented):
A few more comments!

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+}
+async function processPoolEvents(params: {
+  provider: any;
+  pool: Pool;
+  positions: ReturnType<typeof Positions>;
+  pools: ReturnType<typeof Pools>;
+}) {
+  const { pools, pool, provider, positions } = params;
+  assert(pool.id, "requires pool id");
+  assert(pool.address, "requires pool address");
```
mrice32, line 113: These should never be empty right? If they can, can we check them at ingestion time rather than at processing time? Then we could make the type _require_ these fields and have the typechecker guarantee this for us.

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  const poolHandler = PoolEvents({ positions, id: pool.id, pools });
+  const contract = new ethers.Contract(pool.address, UniswapV3Pool.abi, provider);
+  const events = await contract.queryFilter({});
+  await Promise.map(events, poolHandler);
+}
+async function getPositionState(params: { position: Position; provider: any; pool: Pool }) {
+  const { position, provider, pool } = params;
+  assert(pool.address, "requires pool address");
```
mrice32, line 121: Same comment as above.

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  assert(position.tickLower, "requires position tickLower");
+  if (BigInt(position.liquidity.toString()) === 0n) return false;
+  if (BigInt(tick.toString()) > BigInt(position.tickUpper.toString())) return false;
+  if (BigInt(tick.toString()) < BigInt(position.tickLower.toString())) return false;
+  return true;
+};
+
+async function getTickInfo(params: { pool: Pool; provider: any }) {
+  const { pool, provider } = params;
+  assert(pool.address, "requires pool address");
```
mrice32, line 141: Same comment as above. I think the pool type should just require this.

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  const contract = new ethers.Contract(pool.address, UniswapV3Pool.abi, provider);
+  return {
+    pool: pool.address,
+    ...convertValuesToString<Position>(await contract.positions(position.id))
+  };
+}
+
+const IsPositionActive = (tick: BigNumberish) => (position: Position) => {
+  assert(position.liquidity, "requires position liquidity");
```
mrice32, line 130: Same comment as the one for the Pool type. I think this type should just require these fields, and we can check their existence wherever we ingest.

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  assert(pool.address, "requires pool address");
+  const contract = new ethers.Contract(pool.address, UniswapV3Pool.abi, provider);
+  const slot0 = await contract.slot0();
+  const liquidity = await contract.liquidity();
+  return {
+    ...slot0,
+    liquidity: liquidity.toString()
+  };
+}
+async function processNftEvents(params: { provider: any; positions: ReturnType<typeof NftPositions> }) {
```
mrice32, line 85: 1. Is provider not a well-defined type in Ethers?

2. I thought you re-defined the NftPositions type so that we didn't have to write ReturnType... ? Did that not work?

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
+import assert from "assert";
+import { exists, getPositionKey } from "./utils";
+import { ethers } from "ethers";
+
+type Id = string | number;
```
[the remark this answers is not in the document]
daywiss: good point, should probably just use strings

packages/affiliates/libs/uniswap/models.ts
daywiss: id like to inject a different type of store in the table eventually.  Would eventually need a store interface independent of the JS Map that all stores could conform to, so right now id like to leave that part out of the table. Essentially the table can become agnostic to the store technology, and you would be able to plug in different store adapters.

packages/affiliates/libs/uniswap/models.ts
daywiss: interesting technique.... that is really cool. I think though in this case i want to keep the implementation of makeId encapsulated with the table, so they can be broken out as independent features.  will have to think more about how to use this though, i could see it being really useful in some situations.

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  await Promise.map(events, nftHandler);
+}
+async function getNftPositionState(params: { provider: any; position: NftPosition }) {
+  const { provider, position } = params;
+  const contract = new ethers.Contract(
+    getAddress(network, "nonfungibleTokenPositionManagerAddress"),
+    NFTPositionManager.abi,
+    provider
+  );
+  return convertValuesToString(await contract.positions(position.tokenId));
```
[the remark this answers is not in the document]
daywiss: dont know what you mean by "provide a here".  I think i need to change this line to `return convertValuesToString<NftPosition>(await contract.positions(position.tokenId));`  maybe thats what you mean

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
...
+    return set({ id, ...data });
+  }
+  async function set(data: T) {
+    assert(data.id, `${type} requires id`);
+    await store.set(data.id, { ...data });
+    return data;
+  }
+  async function get(id: string) {
+    assert(await store.has(id), `${type} does not exist`);
+    return store.get(id) as T;
```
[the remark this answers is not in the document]
daywiss: yes it should, but typescript doesnt recognize that there is a guard statement with the assert above it, and thinks it can possibly return null, which is an incompatible type with T, so this forces it to be T.  its actually quite annoying

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
...
+// data store with some minor modifications, as long as it has, set, get, delete, has and values
+export function Table<T extends HasId>(config: { makeId: MakeId<T>; type: string }, store: Map<Id, T>) {
+  const { makeId, type } = config;
+  async function create(data: T) {
+    const id = makeId(data);
+    assert(!(await has(id)), `${type} exists`);
+    return set({ id, ...data });
+  }
+  async function set(data: T) {
+    assert(data.id, `${type} requires id`);
```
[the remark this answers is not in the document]
daywiss: a couple reasons, and this touches on why not make `id` required. The user doesn't always know the id of the data when sending it to the table. Its the tables responsibility with `makeId` to determine how to identify the data and insert this into the store.   Because of this, at the top of the Table definition I allow generic types without an id: ` Table<T extends HasId>` where `HasId` has an optional id, since the user might submit raw data without an `.id` on it.

I tried your suggestion though, and when I union `T` and `{id:Id}` it complains that its incompatible with ` Argument of type 'T & { id?: string | undefined; }' ` 

maybe theres a better way to do this though, i really am just mainly hacking

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,169 @@
...
+}
+
+// Generic table with common functions, store is a map now, but we could replace this with any kind of
+// data store with some minor modifications, as long as it has, set, get, delete, has and values
+export function Table<T extends HasId>(config: { makeId: MakeId<T>; type: string }, store: Map<Id, T>) {
+  const { makeId, type } = config;
+  async function create(data: T) {
+    const id = makeId(data);
+    assert(!(await has(id)), `${type} exists`);
+    return set({ id, ...data });
```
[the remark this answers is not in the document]
daywiss: really good point

packages/affiliates/scripts/univ3.ts
```diff
@@ -0,0 +1,210 @@
...
+  pool: Pool;
+  positions: ReturnType<typeof Positions>;
+  pools: ReturnType<typeof Pools>;
+}) {
+  const { pools, pool, provider, positions } = params;
+  assert(pool.id, "requires pool id");
+  assert(pool.address, "requires pool address");
+  const poolHandler = PoolEvents({ positions, id: pool.id, pools });
+  const contract = new ethers.Contract(pool.address, UniswapV3Pool.abi, provider);
+  const events = await contract.queryFilter({});
```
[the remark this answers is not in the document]
daywiss: yea thats right, i think jason ran into this issue of how to set these filters correctly with the voter dapp, but i found this is the equivalent of "everything". will document.

packages/affiliates/scripts/univ3.ts
daywiss: This goes back to the earlier point around making id required. Would be difficult particularly for positions which have a composite id based on owner, lowertick and uppertick, to have the user always pre calculate id.  Because typescript sees that these are optional, it forces us to assert they exist.  This is also a problem with partial updates, where events might include some data about the position, but not others.

IMO this would be  a really pain to fix, it requires making 2 types for every model, one that is a `Partial`, which is what essentially the types are now, and one that is a `Full` which has stronger guarantees around what fields exist.  Then you as the user must know which type to use based on the situation. Maybe you can see another way, happy to entertain suggestions.

packages/affiliates/scripts/univ3.ts
daywiss: ive taken a look at pool.address, and i am able to make that required without any trouble, but that's only because we are never doing any updates to the data in the pool table.  It would be awkward to have to force `{address}` to be included with any partial update to the pool. but since some fields are already being required now, im going to make it required.  will see how it goes.

packages/affiliates/scripts/univ3.ts
daywiss: this one unfortunately can't be required, because the event which creates a pool does not have liquidity. So the flow goes like, PoolCreated event -> pool.create -> getPoolState -> pool.update.   The poolCreated event does not have liquidity.

Its possible to fix this with default values, though, which actually might allow for a lot more required keys.

packages/affiliates/scripts/univ3.ts
daywiss: i just realized typescript has a `Partial<T>` utility. going to try this and see if i can make all fields required

daywiss: partial doesnt really solve things since everything ends up being a partial type. though i was able to apply this to the update function

mrice32: As discussed offline, I think there are some solutions to make _everything_ partial except specific required fields:
1. 
```ts
type PartialT = Partial<T> & {
  requiredField1: string;
  requiredField2: string;
};
```

2. 

```ts
type PartialExcept<T, K extends keyof T> = Partial<Omit<T, K>> & Required<Pick<T, K>>;
type PartialT = PartialExcept<T, 'requiredField1' | 'requiredField2'>;
```

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,168 @@
...
+    const id = exists(data.id) ? data.id : makeId(data);
+    assert(!(await has(id)), `${type} exists`);
+    return set({ id, ...data });
+  }
+  async function set(data: T) {
+    assert(data.id, `${type} requires id`);
+    await store.set(data.id, { ...data });
+    return data;
+  }
+  async function get(id: string) {
```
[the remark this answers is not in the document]
daywiss: fixed both of these issues. The `set` function was fixed by changing `MaybeId` type to an interface rather than a type, then i could do `set(data: T & {id:Id})`.  Dont know why this works but it seems simplest.

also fixed a couple places where i was using string rather than Id

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -0,0 +1,181 @@
...
+// data store with some minor modifications, as long as it has, set, get, delete, has and values
+export function Table<T extends MaybeId>(config: { makeId: MakeId<T>; type: Id }, store: Map<Id, T>) {
+  const { makeId, type } = config;
+  async function create(data: T) {
+    const id = exists(data.id) ? data.id : makeId(data);
+    assert(!(await has(id)), `${type} exists`);
+    return set({ id, ...data });
+  }
+  async function set(data: T & { id: Id }) {
+    assert(exists(data.id), `${type} requires id`);
```
mrice32, line 22: nit: why do we need this now? Isn't Id required?

PR #970, opened 2020-02-25
Upgrade to Solidity 0.6

nicholaspai:
Changes to NPM dependencies:
- Truffle bumped to 5.1.14 from 5.1.4 ([>= 5.1.5 supports solc 0.6](https://github.com/trufflesuite/truffle/releases/tag/v5.1.5))
- OpenZeppelin bumped to [v3](https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v3.0.0-beta.0) which supports solc 0.6
- Prettier Plugin Solidity bumped to alpha.42 from alpha.36 ([>= alpha.40 seemingly supports **some of** solc 0.6](https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-alpha.40))
- Solidity coverage bumped to [0.7.2](https://github.com/sc-forks/solidity-coverage/issues/464) which addresses some solidity 0.6 changes
- [Patched `@drizzle/store` fork to account for 0.6 ABI changes and bumped version](https://github.com/UMAprotocol/drizzle/pull/1)

___
Had to `prettier-ignore` several solidity code blocks because the solidity plugin that we use for prettier does not support some of the breaking 0.6 changes. See #969 :
- ~~Inline assembly call in `Governor` that attempts to call built-in opcode `gas()` (previously `gas`)~~ Fixed by [this PR in alpha.42](https://github.com/prettier-solidity/prettier-plugin-solidity/pull/239)
- All `override` methods that implement `virtual` methods

___
Changes to contract implementations
- No longer can adjust an array's size by dynamically settings its length, for example:
```
 uint newLength = partyMap[parties[i]].derivatives.push(derivativeAddress);
 partyMap[parties[i]].derivativeIndex[derivativeAddress] = newLength - 1;
```
becomes
```
partyMap[parties[i]].derivatives.push(derivativeAddress);
uint newLength = partyMap[parties[i]].derivatives.length;
partyMap[parties[i]].derivativeIndex[derivativeAddress] = newLength - 1;
```
- `Array.push()` no longer returns a value

___
TODO:
- [Solidity Docgen](https://github.com/OpenZeppelin/solidity-docgen/pull/152) needs to be upgraded to work with solc 0.6
- Temporarily removed the following from CI:
     - Slither does not work with 0.6 #976 
- ~~Drizzle store seemingly incompatible with 0.6 ABI's, I raised an [issue](https://github.com/trufflesuite/drizzle/issues/74)~~

Fixes #815

Issue #459, opened 2019-05-28
Store testing - ERC20, final fees, etc.


rosalindstengle:
The Store should have additional tests for final fee calculation, payout in ERC20, and permissioning.

mrice32:
This is done.

PR #2924, opened 2021-04-30
feat(core): updated solidity version to 0.8.x

chrismaree:
**Motivation**

Update from 0.6.x to 0.8.x to: 
1) reduce gas costs, 
2) reduce bytecode size 
3) access new solidity features
4) stay up to date with the latest and greatest

**Summary**

All contracts across the repo have been updated. Additional stylistic changes between solidity versions were also changed.

**Details**

The following changes were made:
1) all instances of solidity `0.6.0` were replaced with `0.8.0`.
2) `pragma experimental ABIEncoderV2;` replaced with `pragma abicoder v2;`
3) Update OZ libraries to the latest version. Pretty much every contract moved in this process so all contract paths were also updated.
4) hard hat and truffle configs updated.
5) docs generation script updated.
6) constructor method accessibility removed to conform to new solidity specification.
7) some syntax changed in a few places. this was updated
8) some data structures needed to change. this was also updated.
9) ensured that `// SPDX-License-Identifier: AGPL-3.0-only` is on all contracts.
10) updated gas-specific tests that were hardcoded to 0.6.x gas values.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**


packages/core/contracts/bot-helpers/uniswap-broker/UniswapBroker.sol
```diff
@@ -156,16 +153,138 @@ contract UniswapBroker {
...
         (address token0, address token1) = sortTokens(tokenA, tokenB);
         pair = address(
-            uint256(
-                keccak256(
-                    abi.encodePacked(
-                        hex"ff",
-                        factory,
-                        keccak256(abi.encodePacked(token0, token1)),
-                        hex"96e8ac4277198ff8b6f785478aa9a39f403cb768dd02cbee326c3e7da348845f" // init code hash
+            uint160(
```
chrismaree, line 156: additional type casting required

packages/core/contracts/common/implementation/Testable.sol
```diff
@@ -46,7 +46,7 @@ abstract contract Testable {
         if (timerAddress != address(0x0)) {
             return Timer(timerAddress).getCurrentTime();
         } else {
-            return now; // solhint-disable-line not-rely-on-time
+            return block.timestamp; // solhint-disable-line not-rely-on-time
```
daywiss, line 49: curious what the difference is between block.timestamp and now, and also does it have any effects on tests

packages/core/contracts/financial-templates/common/WETH9.sol
```diff
@@ -80,7 +80,7 @@ contract WETH9 {
     ) public returns (bool) {
         require(balanceOf[src] >= wad);
 
-        if (src != msg.sender && allowance[src][msg.sender] != uint256(-1)) {
+        if (src != msg.sender && allowance[src][msg.sender] != type(uint256).max) {
```
daywiss, line 83: i just looked up some breaking changes with solidity .8, so this is the recommended way to get `uint256(-1)`, but i have some unrelated questions. 
- if no allowance is set, is `type(uint256).max` what the value defaults to?
- if the max allowance is set, does it differ from `type(uint256).max`

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBroker.sol
```diff
@@ -156,16 +153,138 @@ contract UniswapBroker {
...
+        // 512-bit multiply [prod1 prod0] = a * b
+        // Compute the product mod 2**256 and mod 2**256 - 1
+        // then use the Chinese Remainder Theorem to reconstruct
+        // the 512 bit result. The result is stored in two 256
+        // variables such that product = prod1 * 2**256 + prod0
+        uint256 prod0; // Least significant 256 bits of the product
+        uint256 prod1; // Most significant 256 bits of the product
+        assembly {
+            let mm := mulmod(a, b, not(0))
+            prod0 := mul(a, b)
```
daywiss, line 194: > Optional: If you use SafeMath or a similar library, change x.add(y) to x + y, x.mul(y) to x * y etc.

you probably know this but just in case

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -251,7 +251,7 @@ contract("MerkleDistributor.js", function(accounts) {
           },
           { from: rando }
         );
-        assertApproximate(87259, claimTx.receipt.gasUsed);
+        assertApproximate(87749, claimTx.receipt.gasUsed);
```
daywiss, line 254: gas went up slightly?

daywiss (approved):
looks good, i also briefly looked over breaking changes in .8, and this looked scary, hopefully compiler warns you if we have instances of this (doubtful):

> Exponentiation is right associative, i.e., the expression a**b**c is parsed as a**(b**c). Before 0.8.0, it was parsed as (a**b)**c.

package.json
```diff
@@ -51,6 +51,7 @@
     "prettier-plugin-solidity": "^1.0.0-beta.1",
```
nicholaspai, line 51: Seems like we should also bump `prettier-plugin-solidity` to 1.0.0-beta.7 which introduces a fix for breaking solidity 8 changes: https://github.com/prettier-solidity/prettier-plugin-solidity/releases/tag/v1.0.0-beta.7

(pull request #459)

package.json
```diff
@@ -51,6 +51,7 @@
...
     "pretty-quick": "^2.0.1",
     "secp256k1": "^3.7.1",
+    "solc-0.8": "npm:solc@^0.8.4",
     "solidity-docgen": "^0.5.3",
```
nicholaspai, line 55: do we not also have to bump solidity-docgen?

package.json
```diff
@@ -51,6 +51,7 @@
...
     "truffle": "^5.2.3",
```
nicholaspai, line 56: Doesn't seem like its needed yet but we should probably upgrade truffle as well, as it looks like there were some network performance features added since this update

nicholaspai (approved):
Thanks for making these changes, assuming CI passes LGTM. I raised some questions about other solidity-related dependencies that we should also consider bumping in the same PR. 

Additionally, I am a bit anxious that `drizzle` breaks.

See this big PR that I made a while ago that bumped us to Solidity 6 and also includes notes on other packages that I had to bump: #970

packages/core/contracts/common/implementation/Testable.sol
chrismaree: they are the same thing I believe. after version 7 the `block.timestamp` syntax is preferred.

packages/core/contracts/financial-templates/common/WETH9.sol
chrismaree: this is in the default WETH9 contract so it's not part of our implementation.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBroker.sol
chrismaree: yup! this is assembly taken from uniswap so for now will leave it the same :)

packages/core/test/merkle-distributor/MerkleDistributor.js
chrismaree: ye, looks like it! some went down though.

package.json
chrismaree: yup! as discussed offline though we should do this as a separate PR.

package.json
chrismaree: as discussed offline we will strip this from this PR with the goal of bringing back the dock gen in a later PR.

packages/core/contracts/bot-helpers/action-wrappers/LiquidationWithdrawer.sol
```diff
@@ -1,5 +1,6 @@
-pragma solidity ^0.6.0;
-pragma experimental ABIEncoderV2;
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+pragma abicoder v2;
```
mrice32, line 3: nit: any reason to keep these? Might be nice to just completely drop them given that it's the default now (could be confusing that files that _don't_ have this pragma still have the v2 encoder enabled anyway).

packages/core/contracts/common/implementation/Withdrawable.sol
```diff
@@ -23,7 +23,7 @@ abstract contract Withdrawable is MultiRole {
      * @notice Withdraws ETH from the contract.
      */
     function withdraw(uint256 amount) external onlyRoleHolder(roleId) {
-        Address.sendValue(msg.sender, amount);
+        Address.sendValue(payable(address(msg.sender)), amount);
```
mrice32, line 26: tiny nit: according to [this](https://docs.soliditylang.org/en/v0.8.4/080-breaking-changes.html#new-restrictions), you should be able to just do `payable(msg.sender)` rather than `payable(address(msg.sender))`.

packages/financial-templates-lib/truffle-config.js
```diff
@@ -1,4 +1,5 @@
 const path = require("path");
 const wkdir = path.dirname(require.resolve("@uma/core/package.json"));
+console.log("wkdir", wkdir);
```
mrice32, line 3: extra console log?

packages/serverless-orchestration/package.json
```diff
@@ -12,8 +12,7 @@
     "@awaitjs/express": "^0.3.0",
     "@google-cloud/datastore": "^6.0.0",
     "@google-cloud/storage": "^2.4.2",
-    "@openzeppelin/contracts": "3.0.0",
-    "@uma/core": "npm:@uma/core@1.2.0",
+    "@uma/core": "^2.2.0",
```
mrice32, line 15: Did you mean to do this upgrade in this PR?

mrice32 (approved):
LGTM! A few minor comments

packages/core/contracts/bot-helpers/action-wrappers/LiquidationWithdrawer.sol
chrismaree: I mainly left it like this just to be as explicit as possible but we can remove it to be more concise. I don't really mind either way.

packages/serverless-orchestration/package.json
chrismaree: yup Previously this was bound to this old core version. this meant that the tests in this package broke due to the version bumping.

packages/core/contracts/oracle/implementation/Voting.sol
```diff
@@ -277,13 +277,13 @@ contract Voting is
             // Price requests always go in the next round, so add 1 to the computed current round.
             uint256 nextRoundId = currentRoundId.add(1);
 
-            priceRequests[priceRequestId] = PriceRequest({
```
chrismaree, line 280: z

PR #2927, opened 2021-04-30
feat(core): add typescript types for core

mrice32:
**Motivation**

We would like to export typescript types for our contracts.

**Summary**

This adds typechain to the build process so that when core is built, a types directory is generated that includes an index.d.ts file to declare types for the index file, but it also includes a `contract-types` directory that is used to export typechain types in various formats (Ethers, Truffle, and Web3).

A few notes:
- The entire core package was not converted to ts, only the exported portions (index.ts and the contracts).
- I had to add the core index.ts file to prettier ignore because we need to upgrade prettier before it will allow us to use the `import type` directive in typescript.
- I've pre-released this in `@uma/core@2.3.0-alpha.1` so we can try these out in some of our frontend code.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/affiliates/payouts/devmining/2021-04-19_2021-04-26_0024.json
```diff
@@ -6,41 +6,41 @@
...
-      ["0x46f5E363e69798a74c8422BFb9EDB63e3FB0f08a","0x2CA781A4dcA1f37F52696238e70A004746068F3b","2.0.1"],
-      ["0xb40BA94747c59d076B3c189E3A031547492013da","0xd332f9Faf23Ba97d9f0f9EE9b3d8798C49Ff269B","2.0.1"],
-      ["0x312Ecf2854f73a3Ff616e3CDBC05E2Ff6A98d1f0","0xFc41Ad4FbAa1e6Ed9690E989121122fF6755C918","2.0.1"],
-      ["0x56BaBEcb3dCaC063697fE38AB745c10181c56fA6","0x7eb1e51351380E67F641B84545bf41eCdc10298E","2.0.1"],
-      ["0x10E018C01792705BefB7A757628C2947E38B9426","0xDAa953f01048253A201BA9B0bd0786575f9C2468","2.0.1"],
-      ["0xC9E6C106C65eDD67C83CC6e3bCd18bf8d2Ebf182","0x45Ea614a7Ea47Ec393BDA310F901702DB1347df9","2.0.1"],
-      ["0xCef85b352CCD7a446d94AEeeA02dD11622289954","0xb8E237CC8D2E81b7C9C9C4B5877200f85100fC87","2.0.1"],
-      ["0xd9af2d7E4cF86aAfBCf688a47Bd6b95Da9F7c838","0xa0dfF1F09a0DCEda7641b7eB07c3BDbe66a06C35","2.0.1"],
-      ["0x0f4e2a456aAfc0068a0718E3107B88d2e8f2bfEF","0xa0dfF1F09a0DCEda7641b7eB07c3BDbe66a06C35","2.0.1"]
+      ["0xaB3Aa2768Ba6c5876B2552a6F9b70E54aa256175", "0x45Ea614a7Ea47Ec393BDA310F901702DB1347df9", "1.2.2"],
```
mrice32, line 9: @daywiss @evaldofelipe Why are these files not formatted correctly? And how is this passing our ci lint job?

evaldofelipe: The rewards linter was added very recently. The next files will not have this issue.

.circleci/config.yml
```diff
@@ -27,6 +27,9 @@ jobs:
     steps:
       - restore_cache:
           key: protocol-{{ .Environment.CIRCLE_SHA1 }}
+      - run:
+          name: Install Prereqs
+          command: sudo apt-get install -y rsync
```
nicholaspai, line 32: What was this for?

packages/core/src/FindContractVersion.ts
```diff
@@ -16,16 +26,13 @@ try {
...
-  assert(contractAddress, "Contract address must be provided");
-
+export async function findContractVersion(contractAddress: string, web3: Web3) {
   // Note: there is an unknown issue in web3.js that means that the `getCode` syntax does not function correctly in
   // production. However, ethers has proven to work correctly in production. The code below is a patch to still enable
   // this module to work while we find a better long term solution for the web3.js issue. If running within unit tests
   // then the web3.js version is required as it is scope according to the unit test.
   let contractCode;
-  if (global.web3) {
+  if (((global as unknown) as { web3: Web3 | undefined }).web3) {
```
nicholaspai, line 35: This is pretty ugly, but I guess the global.web3 object could be different types depending on the provider?

.circleci/config.yml
mrice32: We need to rsync the typechain types files on top of the types generated by tsc. See the build process in core/package.json.

nicholaspai (approved):
This is a really productive zero-to-one PR for us migrating longer term to 100% typescript. So is the recommended process for adding future `.ts` files to `core` as follows:
- Compile `.ts` file to `core/dist/`
- Run `.js` from `dist`?

packages/serverless-orchestration/test/ServerlessHub.js
```diff
@@ -455,9 +455,9 @@ contract("ServerlessHub.js", function(accounts) {
     assert.isTrue(lastSpyLogIncludes(hubSpy, "error Command INVALID not found"));
     assert.isTrue(
       responseObject.output.errorOutputs["testServerlessMonitorError2"].execResponse.stderr.includes(
-        "Returned values aren't valid, did it run Out of Gas?"
+        "Contract code hash is null"
       )
     ); // invalid emp error
-    assert.isTrue(lastSpyLogIncludes(hubSpy, "Returned values aren't valid, did it run Out of Gas?"));
+    assert.isTrue(lastSpyLogIncludes(hubSpy, "Contract code hash is null"));
```
mrice32, line 461: @chrismaree not sure why I had to change these error messages. Looking at the monitor index.js, this seems right. Not sure how it previously produced the `Returned values aren't valid, did it run Out of Gas?` error since it first grabs the contract hash before making contract calls (I think),

.circleci/config.yml
nicholaspai: I only see that you added the `build` script for the `common` package, I don't see `rsync`?

packages/core/src/FindContractVersion.ts
mrice32: Yeah, the issue is that typescript doesn't know what properties global has, so we have to basically cast it before testing it. And it forces you to cast to unknown before the custom type :/.

mrice32:
> This is a really productive zero-to-one PR for us migrating longer term to 100% typescript. So is the recommended process for adding future `.ts` files to `core` as follows:
> 
> * Compile `.ts` file to `core/dist/`
> * Run `.js` from `dist`?

@nicholaspai that's the pattern I'm following, but definitely open to other suggestions!

.circleci/config.yml
mrice32: Which means we need to install the rsync package.

nicholaspai:
> > This is a really productive zero-to-one PR for us migrating longer term to 100% typescript. So is the recommended process for adding future `.ts` files to `core` as follows:
> > 
> > * Compile `.ts` file to `core/dist/`
> > * Run `.js` from `dist`?
> 
> @nicholaspai that's the pattern I'm following, but definitely open to other suggestions!

I think this is the lowest churn option and allows backwards compatibility.

packages/core/README.md
```diff
@@ -69,3 +69,60 @@ const governor = await governor.deployed();
 
 Note: this use case is not particularly common, but it is sometimes useful to have access to multiple abi versions
 side-by-side.
+
+## Typescript!
```
chrismaree, line 73: this is very useful!

packages/core/package.json
```diff
@@ -66,8 +73,10 @@
     "mocha-test": "mocha ./mocha-test --recursive",
     "load-addresses": "yarn run apply-registry",
     "clean": "rm -rf build",
-    "buildLatestHardhatVersionHashes": "yarn hardhat run ./src/BuildContractVersionHashes.js",
-    "build": "yarn truffle compile && yarn load-addresses && tsc && yarn buildLatestHardhatVersionHashes",
+    "build": "yarn truffle compile && yarn load-addresses && yarn generate-contract-types && yarn compile-ts && yarn buildLatestHardhatVersionHashes",
```
chrismaree, line 76: this is pretty sweet! hopefully it does not take too long to run

packages/serverless-orchestration/test/ServerlessHub.js
chrismaree: well the `Contract code hash is null` is the error you added to the `FindContractVersion` above with
```
  assert(contractCodeHash !== null, "Contract code hash is null");
```
right?

In any case this test is meant to validate that the hub correctly deals with an `0x0000000000000000000000000000000000000000` set as the contract address and the associated error produced within the spokes (and child process initialised by the spoke). Your change made to the `FindContractVersion` is catching _before_ it gets to the bot's initialisation, where it was previously rejecting. 

Strictly speaking, your change is _better_ than what it was doing before as it is erroring out when trying to fetch byte code.

i.e this is expected, given your code change and seems to be correct :)

package.json
```diff
@@ -15,8 +15,8 @@
     "load-addresses": "yarn workspace @uma/core load-addresses",
     "create-release": "lerna version --no-git-tag-version --conventional-commits --no-push",
     "publish-release": "./scripts/run_release.sh",
-    "qbuild": "yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*'",
-    "build": "yarn workspace @uma/core build && yarn lerna run build --stream",
+    "qbuild": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*' --ignore '@uma/common' --ignore '@uma/core'",
```
chrismaree, line 18: why do you need this diffrent syntax on these builds now?

packages/serverless-orchestration/package.json
```diff
@@ -41,7 +41,7 @@
     "/src/**/*.js"
   ],
   "scripts": {
-    "test": "yarn truffle test test test/*.js"
+    "test": "yarn truffle test test/*.js"
```
chrismaree, line 44: `yarn test test test test test`

TEST IT YARN!!

packages/core/README.md
```diff
@@ -69,3 +69,60 @@ const governor = await governor.deployed();
...
+
+### Web3
+
+Web3 types can be imported similarly to truffle. However, the import syntax is quite limited. There is no way to import
+all UMA web3 types from the same import. Each contract is specified in a separate file.
+
+```ts
+import type { Voting } from "@uma/core/contract-types/web3/Voting";
+
+const voting = new web3.eth.Contract(VOTING_ABI, VOTING_ADDRESS) as Voting;
```
chrismaree, line 127: this syntax is awesome. as next steps we should refactor the typescript bots (trader specifically) to take advantage of this.

package.json
```diff
@@ -29,6 +29,7 @@
   "devDependencies": {
     "@antora/cli": "^2.1.2",
     "@antora/site-generator-default": "^2.1.2",
+    "@typechain/ethers-v5": "^6.0.5",
```
chrismaree, line 32: bit of a pain that `typechain` needs ethers.

chrismaree (approved):
This is a pretty awesome start! Really like that we now have the typechain exports, extremely useful! I also like how this is a somewhat minimal change; you did not need to update _every_ file to ts and rather started with the most important parts.

As next steps do you see value in trying to convert the other `.js` files over to ts within the core package? this will be a bit painful as there are a lot of scripts but it is part of this transition anyway. 

Perhaps we should make a list of the ordering of next steps on package conversion and then can nock them off sprint by sprint?

packages/serverless-orchestration/package.json
mrice32: hahaha yeah, I saw this in trying to debug those tests and I was like 🙃

packages/core/README.md
mrice32: Yep! I tried this out in some of the other packages and it seemed to work great!

package.json
mrice32: This isn't ethers, it's just the ethers typechain plugin, so it should have a pretty small footprint (I hope!)

package.json
mrice32: Two points I noticed:
- We were re-running the core build because we were running it as a pre-step and then it was also being run as part of the second step, so I just removed the redundancy. Unrelated to the typechain change.
- Common is technically a dependency of core. The core `.ts` files now use types from common, so it needs to be built before core, so I needed to specifically run it first to preempt the core build. Later, I would like to figure out how to force lerna to figure out the right order automatically, but I think it currently doesn't work because we have some circular deps.

mrice32: Gonna go ahead and merge since this seems pretty minor, but happy to refactor in a follow up if you have a different suggestion!

PR #2928, opened 2021-05-03
feat(range-trader): add uniswap v3 support

chrismaree:

**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,631 @@
...
+                amountOutMinimum: 0,
+                sqrtPriceLimitX96: sqrtRatioTargetX96
+            })
+        );
+
+        return state.requiredInputAmount;
+    }
+}
+
+// The two methods below are taken almost verbatim from https://github.com/Uniswap/uniswap-v3-core/blob/main/contracts/libraries/TickBitmap.sol.
```
chrismaree, line 170: the rest of this smart contract is libraries and helpers taken from Uniswap and updated to accommodate solidity 0.8.0. Ideally, we would not need to have these in this file and could just import them. However, as we currently still use truffle we need _all_ contracts to be on the same version. _some_ of the solidity files from uniswap are bound to `0.7.x` versions and these are incompatible with our repo. 

once we get rid of truffle completely we will be able to ditch all this code and just import the libs :)

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,631 @@
...
+        // Now that denominator is an odd number, it has an inverse
+        // modulo 2**256 such that denominator * inv = 1 mod 2**256.
+        // Compute the inverse by starting with a seed that is correct
+        // correct for four bits. That is, denominator * inv = 1 mod 2**4
+        uint256 inv = (3 * denominator) ^ 2;
+        // Now use Newton-Raphson iteration to improve the precision.
+        // Thanks to Hensel's lifting lemma, this also works in modular
+        // arithmetic, doubling the correct bits in each step.
+        // NOTE: this is modified from the original Full math implementation to work with solidity 8 with the unchecked syntax.
+        unchecked {
```
chrismaree, line 592: this unchecked syntax was a pain to figure out. it had the most useless revert message ever before

package.json
```diff
@@ -48,7 +48,7 @@
     "lint-staged": "^10.1.3",
     "lodash.startcase": "^4.4.0",
     "prettier": "1.19.1",
-    "prettier-plugin-solidity": "^1.0.0-beta.1",
+    "prettier-plugin-solidity": "^1.0.0-beta.10",
```
chrismaree, line 51: I had to bump this version to correctly lint some new solidity 0.8.x code. in particular the `unchecked` syntax showed up as having errors without this.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,631 @@
...
+            }
+
+            // Finally, set the state price to the next price for the next iteration.
+            state.sqrtPriceX96 = step.sqrtPriceNextX96;
+            state.tick = step.tickNext;
+        }
+
+        // Based on the direction we are moving, set the input and output tokens.
+        address tokenIn = zeroForOne ? pool.token0() : pool.token1();
+        address tokenOut = zeroForOne ? pool.token1() : pool.token0();
```
mrice32, line 145:
```suggestion
        (address tokenIn, address tokenOut) = zeroForOne ? (pool.token0(), pool.token1()) : (pool.token1(), pool.token0());
```

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,631 @@
...
+
+            uint160 nextTickPriceX96 = TickMath.getSqrtRatioAtTick(step.tickNext);
+            uint256 inputAmountForStep;
+
+            // If zeroForOne is true, then we are moving the price UP. In this case we need to ensure that if the next
+            // tick price is more than the target price, we set the set the the next step price to the target price. This
+            // ensures that the price does not undershoot when the next tick is the last tick. Else, traverse the whole tick.
+            if (zeroForOne) {
+                step.sqrtPriceNextX96 = nextTickPriceX96 > sqrtRatioTargetX96 ? sqrtRatioTargetX96 : nextTickPriceX96;
+                inputAmountForStep = SqrtPriceMath.getAmount0Delta( // As we are trading token0 for token1, calculate the token0 input.
```
mrice32, line 104: Just curious, what is this function telling you? How much you need to put in to exhaust the tick? Or is it just telling you how much of token 0 is at the tick?

packages/core/test/bot-helpers/uniswap-broker/UniswapV3Helpers.js
```diff
@@ -0,0 +1,156 @@
+// This file contains a number of useful helper functions. They were either written specifically for these unit tests,
```
mrice32, line 1: Could you point out which ones are pulled from uniswap and which ones we wrote? Just so we know what to review?

mrice32 (commented):
Looks great! Made some minor code suggestions (feel free to ignore or comment back on any you don't like!)

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,631 @@
...
+ * @title UniswapBrokerV3
+ * @notice Trading contract used to arb uniswapV3 pairs to a desired "true" price. Intended use is to arb UMA perpetual
+ * synthetics that trade off peg. This implementation can ber used in conjunction with a DSProxy contract to atomically
+ * swap and move a uniswap market.
+ */
+
+contract UniswapBrokerV3 {
+    using SafeCast for uint256;
+    using LowGasSafeMath for uint256;
+    using LowGasSafeMath for int256;
```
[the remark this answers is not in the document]
chrismaree: we _could_ do that but right now everything in this file is using the `mul`/`div` syntax. dropping this would require a larger refactor. I'd say that for consistency with our other contract let's keep it in and then we can do a larger refactor where we remove all of this tuff in one go.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
chrismaree: it tells you the amount of token 0 you need to move the price from `step.sqrtPriceStartX96` to `step.sqrtPriceNextX96` for a given `state.liquidiy`. This method basically implements equation 5 from the PR description. it's not about exhausting a tick. rather, within a constant liquidity range, how much of token 0 you need to move the price.

packages/core/test/bot-helpers/uniswap-broker/UniswapV3Helpers.js
chrismaree: all tests that were pulled from uniswapv3 were modified to work without ethers so everything was adapted at some level. Regarding reviewing them I think you can basically assume that they all "work" as expected as they came from uniswap v3 (plus modification for web3) except for:
1) `getCurrentPrice`
2) `getTickFromPrice`
3) `replaceLibraryBindingReferenceInArtitifact`

which I added specifically for these tests.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,629 @@
...
+            // ensures that the price does not undershoot when the next tick is the last tick. Else, traverse the whole tick.
+            if (zeroForOne) {
+                step.sqrtPriceNextX96 = nextTickPriceX96 > sqrtRatioTargetX96 ? sqrtRatioTargetX96 : nextTickPriceX96;
+                inputAmountForStep = SqrtPriceMath.getAmount0Delta( // As we are trading token0 for token1, calculate the token0 input.
+                    step.sqrtPriceStartX96,
+                    step.sqrtPriceNextX96,
+                    state.liquidity,
+                    false
+                );
+                // Else, if zeroForOne is false, then we are moving the price UP. In this case we need to ensure that we
```
nicholaspai, line 110:
```suggestion
                // Else, if zeroForOne is false, then we are moving the price DOWN. In this case we need to ensure that we
```

packages/core/test/bot-helpers/uniswap-broker/UniswapBrokerV3.js
```diff
@@ -0,0 +1,266 @@
...
+        MAX_UINT_VAL, // max deadline.
+        { from: trader }
+      );
+
+      // check the price moved up to 12 correctly.
+      const postTradePrice = (await getCurrentPrice(poolAddress)).toNumber();
+      assert.equal(postTradePrice, 8.5);
+    });
+  });
+  describe("Multi-complex liquidity provider", () => {
```
chrismaree, line 215: see here for a complex set of tests with multiple liquidity providers.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,629 @@
...
+                    state.liquidity,
+                    false
+                );
+            }
+
+            // Add amount for this step to the total required input.
+            state.requiredInputAmount = state.requiredInputAmount.add(inputAmountForStep);
+
+            // If we have hit(or exceeded) our target price in the associate direction, then stop.
+            if (zeroForOne && state.sqrtPriceX96 <= sqrtRatioTargetX96) break;
```
nicholaspai, line 126: Where do you set `state.sqrtPriceX96` in this loop?

nicholaspai, line 126: Why don't you check these earlier at the beginning of the loop? Otherwise you might do some unneccessary computations  before you `break`

packages/core/test/bot-helpers/uniswap-broker/UniswapBrokerV3.js
nicholaspai, line 215: What happens if there is 0 liquidity for a certain price, but >0 liquidity below and above it? Does the Uniswap helper module account for that?

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
```diff
@@ -0,0 +1,629 @@
...
+
+        // Work out the direction we need to trade. If the current price is more than the target price then we are
+        // trading token0 for token1. Else, we are trading token1 for token0.
+        bool zeroForOne = sqrtPriceX96 >= sqrtRatioTargetX96;
+
+        // Fetch the current pool liquidity. Build a state object to store this information which can be re-used during
+        // tick traversal later on.
+        uint128 startingLiquidity = pool.liquidity();
+        SwapState memory state =
+            SwapState({ sqrtPriceX96: sqrtPriceX96, tick: tick, liquidity: pool.liquidity(), requiredInputAmount: 0 });
```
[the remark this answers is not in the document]
chrismaree: yup! even better I can just get rid of the `startingLiquidity` variable.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
chrismaree: good question. the reason is that on the _first_ loop you need to add in the initial amount _before_ updating the liquidity. Doing it in this ordering means that on the _initial_ liquidity, pulled in from `pool.liquidity`, we calculate the initial `requiredInputAmount` traversed before updating the liquditiy.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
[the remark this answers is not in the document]
mrice32: That makes sense. A one-time total refactor probably would be better.

As a slightly less invasive suggestion, could we use OZ's safe math implementation? I only suggest it bc it's cheaper gas-wise and more consistent with our other code.

packages/core/contracts/bot-helpers/uniswap-broker/UniswapBrokerV3.sol
nicholaspai: Ah interesting, you could try using a `do...while`? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/do...while

chrismaree: that would work, ye. The way this is set up though we actually dont want to break until after the `state.requiredInputAmount` has been updated  with the next `inputAmountForStep`.

i.e if we used a `do...while` the loop would exit out as soon as the `state.sqrtPriceX96` hits the `sqrtRatioTargetX96` and it would not increment the `requiredInputAmount`.

More over, there are two different break conditions depending on if `zeroForOne` is true (or not). This would be pretty messy to deal with in the `do...while` syntax. For these reason I think I want to leave it as it is now.

nicholaspai (approved):
LGTM! Forgot to approve it

PR #2941, opened 2021-05-05
improve(dependencies): upgrade prettier version

mrice32:
**Motivation**

To use newer typescript notation, we need a newer version of prettier (both eslint and prettier cannot be ignored).

**Summary**

Upgraded prettier and ran the formatter. Had to make a small manual fix due to a disagreement between eslint and prettier.

Note: newer versions of prettier have some different opinionated formats (like trailing commas and parens around arrow function args). We can disable some of these if we see fit. Feel free to comment if you don't like any of the new formatting.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A


packages/affiliates/scripts/example-logs.js
```diff
@@ -45,6 +45,4 @@ async function runTest() {
   // console.log(balances.getTokens().snapshot())
 }
 
-runTest()
```
chrismaree, line 48: glad it no longer wraps this kind of thing

package.json
```diff
@@ -15,8 +15,8 @@
     "load-addresses": "yarn workspace @uma/core load-addresses",
     "create-release": "lerna version --no-git-tag-version --conventional-commits --no-push",
     "publish-release": "./scripts/run_release.sh",
-    "qbuild": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*' --ignore '@uma/common' --ignore '@uma/core'",
-    "build": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '@uma/common' --ignore '@uma/core'",
+    "qbuild": "yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*'",
```
chrismaree, line 18: ha?

chrismaree (approved):
I scrolled through all the changes and they pretty much all seem reasonable. They are all pretty much adding an extra `,` after the last element within an object or adding `(x)` brackets around vars in functions. Few interesting changes but overall they all look good to me!

pemulis:
Looks like this will just need to resolve some merge conflicts.

mrice32:
Thanks for pointing this out @pemulis. Need to loop back on this!

.eslintrc.js
```diff
@@ -2,19 +2,19 @@ module.exports = {
...
     mocha: true,
-    es2020: true
+    es2020: true,
   },
   extends: ["plugin:prettier/recommended", "eslint:recommended"],
   plugins: ["prettier", "mocha"],
   rules: {
     "prettier/prettier": ["error"],
-    indent: ["error", 2, { SwitchCase: 1 }],
+    indent: 0, // avoid conflict with prettier's indent system
```
mrice32, line 11: Prettier and eslint have different implementations of indent, so even when they are configured similarly, they sometimes make conflicting decisions (encountered one in affiliates), so I turned off the eslint checker to allow the prettier one to take over.

PR #2942, opened 2021-05-05
feat(trader): add uniswapv3 broker to trader

chrismaree:
**Motivation**

AS part of adding the uniswap v3 support to the trader, the next step is to add the new smart contracts to the JS code. This PR does just that.

NOTE: the file diff in this PR looks larger than it actually is. This is primarily due to the a few files that were re-named and moved around. In particular, the uniswap v2 stuff was all modified so it could live alongside the v3 changes.

**Summary**

- Added logic to the `createExchangeAdapter` to enable the creation of a Uniswapv3 adapter.
- Added a new exchange adapter: `UniswapV3Trader` to interact with the v3 broker contracts.
- Moved all uniswap v2 broker stuff around and did a bit or re-naming so these can live side by side happily
- Added unit tests for v3 broker integration

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/trader/src/RangeTrader.ts
```diff
@@ -151,7 +152,7 @@ export class RangeTrader {
 
     // Get the post trade spot price to double check deviation error.
     await this.tokenPriceFeed.update();
-    const exchangeSpotPriceAfterTrade = this.tokenPriceFeed.getCurrentPrice();
+    const exchangeSpotPriceAfterTrade = this.tokenPriceFeed.getLastBlockPrice();
```
chrismaree, line 155: we should actually avoid the twap to show the exact spot price after the swap.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
```diff
@@ -13,7 +14,7 @@ async function createExchangeAdapter(
 ) {
   assert(config.type, "Exchange adapter must have a type. EG uniswap for a uniswap dex");
 
-  if (config.type === "uniswap") {
+  if (config.type === "uniswap-v2") {
```
chrismaree, line 17: this is a breaking change! we must update the current trader at the same time that this is rolled out.

packages/trader/test/UniswapV2Trader.ts
```diff
@@ -220,20 +216,21 @@ describe("index.js", function() {
...
     blockNumberBefore = await web3.eth.getBlockNumber();
     await rangeTrader.checkRangeMovementsAndTrade();
+    await tokenPriceFeed.update();
     assert.equal(await web3.eth.getBlockNumber(), blockNumberBefore + 1); // The block number should have been incremented by 1 as a trade was done
 
     // Validate that the correct log messages were produced.
-    assert.isTrue(spyLogIncludes(spy, -5, "Checking if the priceFeed error exceeds the threshold"));
-    assert.isTrue(spyLogIncludes(spy, -4, "The deviationError is greater than the threshold to execute a trade"));
-    assert.isTrue(spyLogIncludes(spy, -3, "Executing function on library"));
+    assert.isTrue(spyLogIncludes(spy, -6, "Checking if the priceFeed error exceeds the threshold"));
```
chrismaree, line 228: updated tests to work correctly with the new possible factory settings.

packages/trader/test/UniswapV3Trader.ts
```diff
@@ -0,0 +1,427 @@
+import { web3, assert } from "hardhat";
```
chrismaree, line 1: These tests are very similar to the `UniswapV2Trader` tests in that they take the contract through a number of different cases and ensures the bot moves the market as expected.

packages/trader/test/index.ts
```diff
@@ -3,18 +3,16 @@ import { web3, assert } from "hardhat";
 
 const { toWei, utf8ToHex, padRight } = web3.utils;
 const {
-  MAX_UINT_VAL,
```
chrismaree, line 6: these variables were not used.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
...
+import BigNumber from "bignumber.js";
+
+const bn = require("bignumber.js"); // Big number that comes with web3 does not support square root.
+bn.config({ EXPONENTIAL_AT: 999999, DECIMAL_PLACES: 40 });
+
+const { encodePriceSqrt } = require("@uma/common");
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
+
+class UniswapV3Trader implements InstanceType<typeof ExchangeAdapterInterface> {
```
mrice32, line 12: nit: why do we need to use `InstanceType<typeof ExchangeAdapterInterface>`? My guess is that this is because we're using JS style requires rather than typescript `import X from Y;` syntax.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
+import winston from "winston";
+import Web3 from "web3";
+import BigNumber from "bignumber.js";
+
+const bn = require("bignumber.js"); // Big number that comes with web3 does not support square root.
+bn.config({ EXPONENTIAL_AT: 999999, DECIMAL_PLACES: 40 });
+
+const { encodePriceSqrt } = require("@uma/common");
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
```
mrice32, line 10: Wherever we can, we should ideally be using `import X from Y;` syntax. If this isn't possible with core and common, it should at least be allowed for ExchangeAdapterInterface since that's a ts file.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
```diff
@@ -2,7 +2,8 @@ import winston from "winston";
 import Web3 from "web3";
 
 const assert = require("assert");
-const { UniswapTrader } = require("./UniswapTrader");
+const { UniswapV2Trader } = require("./UniswapV2Trader");
+const { UniswapV3Trader } = require("./UniswapV3Trader");
 
 async function createExchangeAdapter(
```
mrice32, line 8: Might be useful to define a return type for this function (`Promise<ExchangeAdapterInterface | null>` should probably work?) so it doesn't default to any.

packages/trader/test/UniswapV3Trader.ts
```diff
@@ -0,0 +1,427 @@
...
+  FeeAmount,
+  createContractObjectFromJson,
+  replaceLibraryBindingReferenceInArtitifact
+} = require("@uma/common");
+
+// Script to test
+const { RangeTrader } = require("../src/RangeTrader");
+
+// Helper scripts
+const { createExchangeAdapter } = require("../src/exchange-adapters/CreateExchangeAdapter");
```
mrice32, line 22: Same comments as elsewhere: we should import typescript files as `import X from Y;`. Otherwise, I think typescript has more trouble with it. Note: this may not work for core or common, but I think it's worth a try since they technically have type declarations.

packages/trader/test/UniswapV3Trader.ts
```diff
@@ -0,0 +1,427 @@
...
+const Token = getTruffleContract("ExpandedERC20", web3);
+const WETH9 = getTruffleContract("WETH9", web3);
+const DSProxyFactory = getTruffleContract("DSProxyFactory", web3, "latest");
+const DSProxy = getTruffleContract("DSProxy", web3, "latest");
+
+// Import all the uniswap related contracts.
+const SwapRouter = require("@uniswap/v3-periphery/artifacts/contracts/SwapRouter.sol/SwapRouter.json");
+const UniswapV3Factory = require("@uniswap/v3-core/artifacts/contracts/UniswapV3Factory.sol/UniswapV3Factory.json");
+const NonfungiblePositionManager = require("@uniswap/v3-periphery/artifacts/contracts/NonfungiblePositionManager.sol/NonfungiblePositionManager.json");
+const UniswapV3Pool = require("@uniswap/v3-core/artifacts/contracts/UniswapV3Pool.sol/UniswapV3Pool.json");
```
mrice32, line 33: I _think_ we can also use typescript imports for all json files (and get nice type autocomplete etc!) if we add the resolveJsonModule compiler option: https://mariusschulz.com/blog/importing-json-modules-in-typescript.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
...
+        this.dsProxyManager.getDSProxyAddress(), // to: the output of the trade will send the tokens to the DSProxy.
+        Number((await this.web3.eth.getBlock("latest")).timestamp) + this.tradeDeadline // Deadline in the future
+      )
+      .encodeABI();
+
+    try {
+      return await this.dsProxyManager.callFunctionOnNewlyDeployedLibrary(callCode, callData);
+    } catch (error) {
+      return error;
+    }
```
mrice32, line 55: Does this satisfy the declared return type of the interface (`void`)? What types are you returning here? Also, just out of curiosity, why are we returning the error rather than just throwing?

packages/trader/src/exchange-adapters/UniswapV2Trader.ts
```diff
@@ -6,9 +6,9 @@ const { MAX_UINT_VAL } = require("@uma/common");
 const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
 const { getTruffleContract } = require("@uma/core");
 
-class UniswapTrader implements InstanceType<typeof ExchangeAdapterInterface> {
+class UniswapV2Trader implements InstanceType<typeof ExchangeAdapterInterface> {
```
mrice32, line 9: +1 on all these name changes!

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
...
+const bn = require("bignumber.js"); // Big number that comes with web3 does not support square root.
+bn.config({ EXPONENTIAL_AT: 999999, DECIMAL_PLACES: 40 });
+
+const { encodePriceSqrt } = require("@uma/common");
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
+
+class UniswapV3Trader implements InstanceType<typeof ExchangeAdapterInterface> {
+  readonly tradeDeadline: number;
+  readonly UniswapV2Broker: any;
```
mrice32, line 14:
```suggestion
  readonly UniswapV3Broker: any;
```

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
...
+    readonly uniswapPoolAddress: string,
+    readonly uniswapRouterAddress: string,
+    readonly dsProxyManager: any
+  ) {
+    this.logger = logger;
+    this.web3 = web3;
+    this.uniswapPoolAddress = uniswapPoolAddress;
+    this.uniswapRouterAddress = uniswapRouterAddress;
+
+    this.dsProxyManager = dsProxyManager;
```
mrice32, line 29: nit: I think you can drop all of these. Tossing public, protected, private, or readonly in front of these constructor params automatically assigns them to instance vars. Reference: https://www.typescriptlang.org/docs/handbook/2/classes.html#parameter-properties.

```suggestion
```

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
```diff
@@ -0,0 +1,59 @@
...
+bn.config({ EXPONENTIAL_AT: 999999, DECIMAL_PLACES: 40 });
+
+const { encodePriceSqrt } = require("@uma/common");
+const ExchangeAdapterInterface = require("./ExchangeAdapterInterface");
+const { getTruffleContract } = require("@uma/core");
+
+class UniswapV3Trader implements InstanceType<typeof ExchangeAdapterInterface> {
+  readonly tradeDeadline: number;
+  readonly UniswapV2Broker: any;
+  uniswapPair: any;
```
mrice32, line 15: nit: is this variable being used?

mrice32 (commented):
Looks awesome! Super clean!

Some minor comments around typescript. Biggest one is preferring the `import X from Y;` (or `import { X } from Y;` syntax over `require`. I think the former will allow typescript to do a better job at resolving types.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: could be. will try to update to the correct syntax and see.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: yup! will update

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: it's not! looks like our linter is not working at detecting unused variables in ts.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: you right, it does not. This method will either return what `callFunctionOnNewlyDeployedLibrary` returns (which contains transaction information, tx config and return values from function calls on contract) OR returns an error. within the `RangeTrader` the return type is checked if it is an `instanceOfError`. if it is then error logic is handled in that layer.

An alternative pattern would be to just throw (or not even wrap in try catch here) and then put the try-catch logic one layer up within the `RangeTrader`. the advantage of this is that we dont need these `try-catch` blocks within the exchange-adapters, thereby placing the burden on the caller of this adapter. WDYT?

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: ye this is a problem in a few more places. will update it throught.

packages/trader/src/exchange-adapters/CreateExchangeAdapter.ts
chrismaree: good idea! done

packages/trader/test/UniswapV3Trader.ts
chrismaree: this did not work perfectly out the box:
image

after adding `"resolveJsonModule": true` to the ts config it worked! great call out!

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: this is a great point! I did not know you could strip all these out. I've updated all other files in the trader to follow this **MUCH** cleaner pattern.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: > Looks awesome! Super clean!
> 
> Some minor comments around typescript. Biggest one is preferring the `import X from Y;` (or `import { X } from Y;` syntax over `require`. I think the former will allow typescript to do a better job at resolving types.

@mrice32 Thank you for your feedback! I went back and refactored basically every file in the trader to follow the correct TS patterns. I also realized I had not updated the `hardhat.config.ts` as we'd discussed IRL which meant the unit tests were not correctly running in CI.

Please take a look again and let me know if you're happy :)

packages/trader/hardhat.config.ts
```diff
@@ -5,7 +5,7 @@ import { getHardhatConfig } from "@uma/common";
 import path from "path";
 
 const coreWkdir = path.dirname(require.resolve("@uma/core/package.json"));
-const packageWkdir = path.dirname(require.resolve("@uma/optimistic-oracle/package.json"));
+const packageWkdir = path.dirname(require.resolve("@uma/trader/package.json"));
```
chrismaree, line 8: this meant that trader tests were not running in CI 👀

packages/trader/src/RangeTrader.ts
chrismaree: right now the bot will create a default Uniswap feed for the synthetic provided. By doing it this way we don't need to do any configuration changes on the default feed created.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: I've updated the type in the interface to return `Promise<Error | { transactionHash: string; any: any }>`. In this way, the `RangeTrader` can deal with the error case by checking the type returned. the `RangeTrader` uses the `transactionHash` in logging. I added the `any:any` to enable us to pass extra props in the return object without any issues (additional props are passed back in the `runTransaction` Util PR, for example).

packages/trader/src/index.ts
```diff
@@ -1,12 +1,11 @@
 import winston from "winston";
 import Web3 from "web3";
-
+import retry from "async-retry";
 import { config } from "dotenv";
 
-import retry from "async-retry";
```
chrismaree, line 6: @mrice32 see the new imports

packages/trader/src/index.ts
```diff
@@ -92,7 +91,11 @@ export async function run(logger: winston.Logger, web3: Web3): Promise<void> {
       await retry(
         async () => {
           // Update the price feeds & gasEstimator.
-          await Promise.all([tokenPriceFeed.update(), referencePriceFeed.update(), gasEstimator.update()]);
+          await Promise.all([
+            (tokenPriceFeed as any).update(),
```
chrismaree, line 95: @mrice32 one slightly painful thing is that the type returned by the `createTokenPriceFeedForFinancialContract` does not return an appropriate type so I had to type cast this. do you see any better solution to this?

packages/trader/test/index.ts
```diff
@@ -39,25 +33,22 @@ describe("index.js", function() {
...
-  const MockOracle = getTruffleContract("MockOracle", web3);
-  const Token = getTruffleContract("ExpandedERC20", web3);
-  const SyntheticToken = getTruffleContract("SyntheticToken", web3);
-  const Timer = getTruffleContract("Timer", web3);
-  const UniswapMock = getTruffleContract("UniswapV2Mock", web3);
-  const Store = getTruffleContract("Store", web3);
-  const ConfigStore = getTruffleContract("ConfigStore", web3);
-  const OptimisticOracle = getTruffleContract("OptimisticOracle", web3);
-  const DSProxyFactory = getTruffleContract("DSProxyFactory", web3);
+  const FinancialContract = getTruffleContract("Perpetual", web3 as any);
```
chrismaree, line 39: changing the import of `web3` also mean that I had to do this type casting. not sure what solutions there are to this.

The error produced was as follows:
image

packages/trader/src/RangeTrader.ts
mrice32: But isn't it just as trivial to add an override config that contains a `twapLength: 0` parameter?

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
mrice32: That sounds great! How are the `any: any` parts used? Is it just for logging? What does the caller do with the `any` field? If it's just for logging, could we have a type like the following so you require a transaction hash and also allow any other string -> string fields.
```ts
Promise<Error | { transactionHash: string; [key: string]: string; }>
```

If it really could be anything, we can use `any`, but I think it's probably preferable to try and describe what shape the extra props are going to come in so they can be more effectively used by the calling code.

packages/trader/test/index.ts
mrice32: This is really annoying. Seems like it might be caused by different versions of web3 that are being imported here and in core.

I think to make it less annoying to deal with, you should just create a casted version at the top.

```ts
import { web3: uncastWeb3 } from "hardhat";
...

// This is a workaround because the core web3 type and local one are slightly different.
type CoreWeb3 = Parameters<typeof getTruffleContract>[1]; // Grabs the type of the second argument.
const web3 = uncastWeb3 as CoreWeb3; // Note: you could also use any here, but it would make web3 any everywhere in this file.
```

mrice32 (approved):
LGTM!

packages/trader/hardhat.config.ts
nicholaspai: oh wow, great catch

packages/trader/package.json
```diff
@@ -41,7 +42,7 @@
   "main": "dist/src/index.js",
   "scripts": {
     "build": "tsc -b",
-    "test": "hardhat test"
+    "test": "hardhat test  --network hardhat"
```
nicholaspai, line 45: do you need to specify --network hardhat? isn't that default?

packages/trader/src/RangeTrader.ts
chrismaree: this did not work as expected in unit tests. the feed would return no prices. in production, this is a bit irritating as it bloats the config more. Alternatively, we could get the `index` entry to inject this config but that seems more complex.

Do you see any downside with just using the `getLatestBlockTime`? this trader only works with uniswap right now (and therefore Uniswap feeds) so there will always be this method.

packages/trader/src/exchange-adapters/UniswapV3Trader.ts
chrismaree: I'm not exactly sure what the structure will be after the refactor in the `runTransaction` PR. this is why I left it as `any:any`. I think it could be both an `Error` or a string-keyed object describing transaction outputs.

packages/trader/test/index.ts
chrismaree: damn that's pretty messy. I'll update accordingly

packages/trader/package.json
chrismaree: no but this is how we do it in other tests :)

packages/trader/test/index.ts
chrismaree: this did not work :( I had type issues all over the place when trying this technique.

packages/trader/package.json
chrismaree: > actually this is not in other tests. I've removed it.

PR #2944, opened 2021-05-05
improve(L2): Add BeaconOracle unit tests

nicholaspai:
**Motivation**

Contracts introduced by #2903 need unit tests in addition to implemented e2e test.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/test/chainbridge/sinkOracle.js
```diff
@@ -0,0 +1,149 @@
...
+      "should have price after publish"
+    );
+    assert.equal(
+      (await sinkOracle.getPrice(testIdentifier, testRequestTime, testAncillary, { from: owner })).toString(),
+      testPrice,
+      "should not revert after publish"
+    );
+  });
+  it("formatMetadata", async function() {
+    const metadata = await sinkOracle.formatMetadata(chainID, testIdentifier, testRequestTime, testAncillary);
```
mrice32, line 141: Just curious, is this method only public so we can test it? Or is it needed for some other reason?

packages/core/test/chainbridge/sourceOracle.js
```diff
@@ -0,0 +1,190 @@
...
+          event.chainID.toString() === destinationChainID.toString() &&
+          hexToUtf8(event.identifier) === hexToUtf8(testIdentifier) &&
+          event.time.toString() === testRequestTime.toString() &&
+          event.ancillaryData.toLowerCase() === testAncillary.toLowerCase() &&
+          event.price.toString() === testPrice.toString()
+      );
+
+      // Deposit event will be emitted after successful Bridge.deposit() internal call if the resource ID is set up
+      // properly.
+      const internalTxn = await TruffleAssert.createTransactionResult(bridge, txn.tx);
```
mrice32, line 111: Never seen this syntax before... OOC what is this doing?

mrice32 (approved):
LGTM! Two non-blocking questions.

packages/core/test/chainbridge/sinkOracle.js
nicholaspai: Yeah its public primarily for testing, but its also convenient in case we need to call Bridge.deposit() directly, because the `metadata` return value can be passed directly as the third param to `deposit()`

PR #2940, opened 2021-05-04
feat(price-feed): added USDXIO price feed in DefaultPriceFeedConfigs.js

asadnaeem1:
**Motivation**

Default price feed config for USDXIO price identifier, [UIMP-71](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-71.md).

**Summary**

Default config for the USDXIO price identifier is created in the file DefaultPriceFeedConfigs.js.

**Details**

The config is of type expression. It uses uniswap pairs of XIO/ETH and ETH/USDT to calculate the price of USD in terms of XIO.

**Testing**

Check a box to describe how you tested these changes and list the steps for reviewers to test.

- [ ]  Ran end-to-end test, running the code as in production
- [ ]  New unit tests created
- [ ]  Existing tests adequate, no new tests required
- [ ]  All existing tests pass
- [x]  Untested

PR #2945, opened 2021-05-05
fix(price-feeds): allow for cryptowatch price inversion when twapLength is defined

smb2796:
**Motivation**

When a `twapLength` is defined for a cryptowatch price feed, `invertPrice` is not used. This PR allows for inverting cryptowatch prices while also using a twap.

To reproduce, define a non-zero `twapLength` config parameter for any cryptowatch default price feed config that also has an `invertPrice` value of true.

**Summary**

Inverts the twap returned from `_computeTwap()` when `invertPrice` is true.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

Tested locally by running a monitor bot, but did not add anything to the unit tests. Happy to do so if it is necessary.

**Issue(s)**

NA

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -287,8 +287,10 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       })
       .flat();
     const startTime = endTime - this.twapLength;
+    const twapPrice = computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
 
-    return computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
+    if (!this.invertPrice) return twapPrice;
+    else return this._invertPriceSafely(twapPrice);
   }
 
   _invertPriceSafely(priceBN) {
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -287,8 +287,10 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       })
       .flat();
     const startTime = endTime - this.twapLength;
+    const twapPrice = computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
 
-    return computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
+    if (!this.invertPrice) return twapPrice;
```
nicholaspai, line 292: Prob could make this a one liner: `return (this.invertPrice ? this._invertPriceSafely(twapPrice) : twapPrice)`

nicholaspai (commented):
Good catch, curious how you found this bug?

nicholaspai (approved):

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
smb2796: Yeah good point. Updated to this.

smb2796:
> Good catch, curious how you found this bug?

Thanks! Was causing issues when trying to write the default pf config in #2940

mrice32 (approved):
LGTM!

PR #2947, opened 2021-05-05
improve (CircleCI): Test packages with parallelism

evaldofelipe:
**Motivation**

Some packages are huge enough to spend more than 50 minutes to complete the tests. This PR use a circleCI feature called [parallelism](https://circleci.com/docs/2.0/parallelism-faster-jobs/) to split the biggest packages. 

**Testing**

- [ ]  New unit tests created
- [ ]  All existing tests `pass`
- [ ]  Untested


packages/financial-templates-lib/src/clients/FinancialContractClient.js
```diff
@@ -337,3 +337,5 @@ class FinancialContractClient {
 }
 
 module.exports = { FinancialContractClient };
+
+// Test comment for CI test proposes
```
nicholaspai, line 341: Should we keep this comment?

nicholaspai (commented):
Is CI still going to perform the optimization where only packages that were modified are tested?

Comparing against the last `master` commit that touched all tests, this does seem to speed up CI from 1 hour to 40 mins!

ci/tests/test-liquidator.sh
```diff
@@ -0,0 +1,32 @@
...
+            cd packages/liquidator
+            yarn mocha mocha-test
+      - run:
+          name: Run tests
+          command: |
+            ./ci/truffle_workaround.sh
+            pwd
+            cd packages/liquidator
+            echo $TESTS_GLOB
+            circleci tests glob "test/**/*.js" | circleci tests split > /tmp/test-files
```
mrice32, line 30: Very cool!

ci/tests/test-financial-templates-lib.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+TESTS_GLOB="$(echo 'circleci tests glob "test/**/*.js"')"
+TESTS_FILE="$(echo '$(cat /tmp/test-files)')"
```
mrice32, line 4: Interesting. Just curious, why did you decide to put this inside a variable up here rather than just writing it out in the command itself?

mrice32 (approved):
Looks great!

packages/financial-templates-lib/src/clients/FinancialContractClient.js
evaldofelipe: Removed the comment @nicholaspai, Thanks!

ci/tests/test-financial-templates-lib.sh
evaldofelipe: good catch! I've used for debug proses, and before split the test file on `ci/tests/`used  as variable over the same file, and stay as is on the new one. Fixed for now!

packages/financial-templates-lib/test-truffle/BalancerPriceFeed.js
```diff
@@ -1,9 +1,9 @@
 const { toWei, toBN } = web3.utils;
 const winston = require("winston");
 
-const { BalancerPriceFeed } = require("../../src/price-feed/BalancerPriceFeed");
```
chrismaree, line 4: pretty sure you did not mean to change these. likely want to revert them I'm guessing?

ci/generate_lerna_config.sh
```diff
@@ -5,84 +5,65 @@ set -o nounset
...
-EOF
+  /bin/bash $TESTS_PATH/test-required.sh >> $CI_CONFIG_FILE
 
 
   printf "\n\nworkflows:\n  version: 2.1\n  build_and_test:\n    jobs:\n" >> $CI_CONFIG_FILE
 
-  for PACKAGE in "${PACKAGES_ARRAY[@]}"
+  if [[ " ${PACKAGES_ARRAY[@]} " =~ " @uma/financial-templates-lib " ]]; then
+    REMOVE=@uma/financial-templates-lib
+    PACKAGES_ARRAY=( "${PACKAGES_ARRAY[@]/$REMOVE}" )
```
chrismaree, line 39: what is this magic exactly? can you explain a bit? perhaps a comment might help as well.

nicholaspai:
Any update on this PR @evaldofelipe ?

daywiss:
@evaldofelipe ping eng if you need help to resolve the conflicts with the code

packages/financial-templates-lib/test-truffle/BalancerPriceFeed.js
evaldofelipe: it's correct! it's because to create the parallelism on tests, was necessary split the test folder in `test` and `tests-truffle` to run with truffle and hardhat in different jobs, so the required files change the folder too.

ci/generate_lerna_config.sh
evaldofelipe: First we transform the packages edited in a list, and create an array with that information. That works until the parallelism appear. Now we need execute separated tests for each tool used to test a package. In this case, we search inside the array the `financial-templates-lib` package. If exists, we remove that package from the array and add `financial-templates-lib-hardhat` and `financial-templates-lib-truffle` that make sense to you? @chrismaree

PR #2949, opened 2021-05-05
feat(kpi): generalize kpi options library to set transformed price

pemulis:

**Motivation**

Generalizes the KPI options financial product library to allow an address to set a pre-expiry transformed price for an EMP using the library. Previously, the transformed price was hardcoded as `2`, which obviously does not apply to every use case. Similar to the covered call library, the transformed price should be set prior to adding collateral to the EMP, can be set by any address, and can only be set once.

**Summary**

Adds functionality to set, get, and use a transformed price passed in after the EMP contract is created, similar to setting the strike price in the covered call library.

**Details**

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
  * Post-expiry, the collateral requirement is left as 1 and the price is left unchanged.
  */
 contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable {
+    using FixedPoint for FixedPoint.Unsigned;
+
+    mapping(address => FixedPoint.Unsigned) financialProductTransformedPrices;
```
nicholaspai, line 16: think you should explicitly set this as a `public` var

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
...
+     * @param transformedPrice the price for the financial product to be used if the contract is pre-expiration.
+     * @dev Note: a) Any address can set price transformations b) The price can't be set to blank.
+     * c) A transformed price can only be set once to prevent the deployer from changing it after the fact.
+     * d)  financialProduct must expose an expirationTimestamp method.
+     */
+    function setFinancialProductTransformedPrice(address financialProduct, FixedPoint.Unsigned memory transformedPrice)
+        public
+        nonReentrant()
+    {
+        require(transformedPrice.isGreaterThan(0), "Cant set price of 0");
```
nicholaspai, line 30:
```suggestion
        require(transformedPrice.isGreaterThan(0), "Cannot set price to 0");
```

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
...
+        require(financialProductTransformedPrices[financialProduct].isEqual(0), "Price already set");
+        require(ExpiringContractInterface(financialProduct).expirationTimestamp() != 0, "Invalid EMP contract");
```
nicholaspai, line 32: Think you should move this external function call to the FinancialContract after the state change to stick to the [Check-Effects-Interaction pattern](https://fravoll.github.io/solidity-patterns/checks_effects_interactions.html). But curious what @chrismaree @mrice32 think too

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
...
+
+    /**
+     * @notice Returns the transformed price associated with a given financial product address.
+     * @param financialProduct address of the financial product.
+     * @return transformed price for the associated financial product.
+     */
+    function getTransformedPriceForFinancialProduct(address financialProduct)
+        public
+        view
+        nonReentrantView()
```
nicholaspai, line 44: Don't think you need this `nonReentrantView` since there is no external call

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
nicholaspai: Also, are we ok not allowing 0 prices? For example, this library would be bad for the funding rate price identifier whose value could realistically be 0.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
chrismaree: agreed

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
chrismaree: the original idea was it would prevent you from setting uninitialised prices but that is a good point @nicholaspai

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
...
 contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable {
+    using FixedPoint for FixedPoint.Unsigned;
+
+    mapping(address => FixedPoint.Unsigned) financialProductTransformedPrices;
+
+    /**
+     * @notice Enables any address to set the transformed price for an associated financial product.
+     * @param financialProduct address of the financial product.
+     * @param transformedPrice the price for the financial product to be used if the contract is pre-expiration.
+     * @dev Note: a) Any address can set price transformations b) The price can't be set to blank.
```
chrismaree, line 22: nit: ideally put the note above the params.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
chrismaree: interesting. I see this more as a _check_ part of the pattern as this is not actually an _interaction_. it's validating that the `financialProduct` is an EMP indirectly by ensuring it has this `expirationTimestamp` method` and is not actually interacting with the external contract.

More over, I personally prefer having all the requires at the top of the method. Lastly, this method has `nonReentrant` modifier so nothing negative can come from having it above the state modification.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
daywiss: is this function redundant if `financialProductTransformedPrices` map is public?

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,42 @@ import "../../../common/implementation/Lockable.sol";
...
+     * c) A transformed price can only be set once to prevent the deployer from changing it after the fact.
+     * d)  financialProduct must expose an expirationTimestamp method.
+     */
+    function setFinancialProductTransformedPrice(address financialProduct, FixedPoint.Unsigned memory transformedPrice)
```
daywiss, line 26: i dont really understand the full context of this change, but can you explain the consequences of allowing people to arbitrarily set this value?

daywiss (commented):
im going to leave approval to @nicholaspai and @chrismaree on this

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
nicholaspai: But couldn't you theoretically set a malicious `expirationTimestampI()` implementation that tries to do something else in addition to returning the expiration time?

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
nicholaspai: Agreed @daywiss

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
chrismaree: If this library had any token balance in it or something like that then I guess you are right from that perspective. I think that having the modifier is sufficient though, don't you?

evaldofelipe:
@pemulis Any update on this PR?

chrismaree:
@pemulis any updates on this PR?

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
pemulis: This would allow anyone to set the value for an EMP using this library, where before it was restricted to the deployer of the library (*not* the deployer of the EMP).

When this was restricted to the deployer of the library, that prevents anyone else from using the library without our help.

By opening it up to anyone, a deployer of a new EMP using this library can set a value after deployment but before putting any money in the EMP contract. It theoretically opens you up to an economically irrational griefing attack where someone sets a bogus value before you can set the real value, but we don't consider that a likely attack, and it could be mitigated by deploying the EMP and setting the transformed price in the same transaction if it actually started to happen.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
pemulis: I refactored this a bit to allow a price of 0, for cases where that is useful.

pemulis:
Made some updates based on everyone's feedback! Once this is in good shape, we can deploy it so that it's ready to use during the hackathon.

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
nicholaspai: yeah good call, forgot about `nonReentrant`

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -24,17 +51,18 @@ contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable
...
         if (requestTime < ExpiringContractInterface(msg.sender).expirationTimestamp()) {
-            return FixedPoint.fromUnscaledUint(2);
+            return transformedPrice.price;
         } else {
             return oraclePrice;
         }
     }
 
     /**
-     * @notice Returns a transformed collateral requirement that is set to be equivalent to 2 tokens pre-expiry.
```
nicholaspai, line 37: nice find

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -11,6 +11,33 @@ import "../../../common/implementation/Lockable.sol";
...
  */
 contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable {
+    using FixedPoint for FixedPoint.Unsigned;
+
+    struct TransformedPrice {
+        bool set;
+        FixedPoint.Unsigned price;
+    }
+
+    mapping(address => TransformedPrice) public financialProductTransformedPrices;
```
nicholaspai, line 21: @eng is there any reason to make this mapping private (and also permission `transformPrice` to only be called by contracts) in consideration of oracle parasitic usage? I'm guessing its not a problem, but want to highlight

packages/core/test/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.js
```diff
@@ -31,7 +31,7 @@ contract("KpiOptionsFinancialProductLibrary", function() {
     );
   });
   describe("price transformation", () => {
```
nicholaspai, line 33: I know this test script is copied from another FPLib contract but thoughts on adding a unit test to check if a revert occurs when the transformedPrice is NOT set yet?

nicholaspai (commented):
Thanks for responding to this John and great work! Please ping for a review once you see my comments

packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
chrismaree: I dont think that there is any parasitic usage on this. This is a transformation so it's not storing any price information that other contracts could use in any meaningful way. IMHO it's better to leave it public to ensure that we can call it to verify the state of set transformations.

chrismaree:
@pemulis this PR is still sitting :)

nicholaspai (approved):

chrismaree (approved):
@pemulis let's try to close this PR, please. There are still a few outstanding comments but once those have been resolved we should be good to go.

daywiss (approved):

packages/core/test/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.js
pemulis: Would this work in the non-1e18 case when the price is passed as a `FixedPoint.Unsigned`?

pemulis: @nicholaspai hmm, it actually looks like it should be reverting but isn't, since `setFinancialProductTransformedPrice` is never called. Looking into it.

pemulis:
Closing for now, but will reopen after hackathon kickoff to finalize the tests.

PR #2950, opened 2021-05-06
feat: Add hardhat deployment infrastructure

nicholaspai:
**Motivation**

Add example deployment, etherscan-verification, and post-deployment setup sprints, all using the native `hardhat` tools as much as possible!

**Summary**

Follow up features:
- Figure out best way to export deployed addresses, bytecode, and ABI similar to `core/networks/X.json`.
- Take full advantage of `hardhat-deploy` tags to make deploying different systems (test, prod, sink-oracle, etc.) possible in one line CLI commands.
- Make verification work on Matic, might require some custom logic because Matic has an Etherscan substitute. We can hopefully use [sourcify](https://github.com/wighawag/hardhat-deploy#5-hardhat-sourcify) to verify programatically on Matic.
- Merge `core/hardhat.config.js` with `common/HardhatConfig.js` which generally should serve the same role. It would be great if we could share `HardhatConfig.js` similar to how we use `TruffleConfig.js` across packages.
- Use `deployments` in tests.

**Details**

NOTE FOR REVIEWERS:
- The entrypoint for this PR is in `README.md` and `package.json` to see the new CLI scripts added.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/common/src/Constants.js
```diff
@@ -9,7 +9,8 @@ const interfaceName = {
   FundingRateStore: "FundingRateStore",
   OptimisticOracle: "OptimisticOracle",
   Bridge: "Bridge",
-  GenericHandler: "GenericHandler"
+  GenericHandler: "GenericHandler",
+  MockOracleAncillary: "Oracle"
```
nicholaspai, line 13: @mrice32 I don't think there's any risk to adding this mapping here, which is convenient for the `core/scripts/hardhat/finder.js` script but wanted to highlight it.

packages/core/deployments/goerli/solcInputs/87af3bddb96fa5cfcfc99d0e6fd39498.json
```diff
@@ -0,0 +1,416 @@
+{
```
mrice32, line 1: Do we have to commit all of these generated files to the repo? It seems like all we should really need are the deployed addresses, like we store in networks/?

nicholaspai: My thought originally was to share the solc standard input JSON so that others could verify contracts, but whoever deploys contracts should ideally just use their own locally generated JSON. I'm going to remove everything in `deployments` (and add to `.gitignore`) but I'm going to add commentary about a script that we can use to easily export the local `deployments` data (incl. address and ABI) to the `networks/` directory.

For example `yarn hardhat export --export-all /networks/hardhat/deployed.json` makes a dictionary of contract addresses and ABI for each network that the user has local deployments data for.

[More on the export command](https://hardhat.org/plugins/hardhat-deploy.html#exporting-deployments)

packages/common/src/Ethers.js
```diff
@@ -0,0 +1,20 @@
+const { hexlify, toUtf8Bytes } = require("ethers/utils");
+
+function stringToBytes32(text) {
```
mrice32, line 3: Just curious, is this copied from somewhere? I feel like I recognize the code

packages/core/deploy/001_deploy_finder.js
```diff
@@ -0,0 +1,14 @@
+const func = async function(hre) {
```
mrice32, line 1: i'm guessing we'll add to these when we add contracts to the setup (like financial contracts, collateral whitelists, etc)?

packages/core/deploy/008_mock_oracle.js
```diff
@@ -0,0 +1,33 @@
+const { stringToBytes32, interfaceName, ZERO_ADDRESS } = require("@uma/common");
```
mrice32, line 1: Is the mock oracle only expected to be deployed on a test L1? Do you think we should have a `prod` tag and a `test` tag or something so we can deploy the entire test setup or the entire prod setup? Also, are you allowed to provide multiple tags on deployment? If so, how do they mix (union or intersection of the tagged scripts?)?

packages/core/README.md
```diff
@@ -70,6 +70,44 @@ const governor = await governor.deployed();
...
+
+The following commands are implemented as [hardhat tasks](https://hardhat.org/guides/create-task.html) that make it easy to interact with deployed contracts via the CLI:
+
+`yarn hardhat register-deployer --network <NETWORK-NAME>`
+
+Registers the `deployer` account (as defined in the `namedAccounts` param in `hardhat.config.js`) with the deployed Registry for the network.
+
+`yarn hardhat setup-finder --registry --bridge --generichandler --network <NETWORK-NAME>`
+
+Sets specified contracts in the deployed Finder. More contracts available to be set can be found in the `core/scripts/hardhat/tasks/finder.js` script.
```
mrice32, line 109: Is the idea that most/all scripts that we currently have in scripts/ should be implemented as tasks in hardhat?

packages/core/networks/hardhat/deployed.json
```diff
@@ -0,0 +1,5765 @@
+{
```
mrice32, line 1: A few questions:
1. Is this pulled in by hardhat scripts or does that use the generated files under artifacts/?
2. Are the abis needed? If not, is there an easy way to drop them from this file to make it easier to read/interpret?

packages/core/scripts/hardhat/deployContracts.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+
+# The type of oracle system to set up, for example "beacon-l1" or "beacon-l2". Default is "dvm".
+ORACLE_TYPE=$1
```
mrice32, line 4: Could we just make higher-level tags (in addition to the existing ones), like beacon-l1, beacon-l2, dvm to avoid having to write this script? Or is that a misuse of tags?

packages/core/scripts/hardhat/tasks/registry.js
```diff
@@ -0,0 +1,33 @@
+const { task } = require("hardhat/config");
+const { RegistryRolesEnum } = require("@uma/common");
+const { waitForTxn } = require("./utils");
+
+task("register-deployer", "Register deployer with Registry capable of making price requests").setAction(async function(
```
mrice32, line 5: Do these automatically get registered as tasks because they're in the tasks directory?

packages/core/scripts/hardhat/tasks/registry.js
```diff
@@ -0,0 +1,33 @@
+const { task } = require("hardhat/config");
```
mrice32, line 1: I generally find the task syntax to be more pleasant than the truffle syntax. Do you prefer this to a normal node script? What do you think the tradeoffs are between this and a normal node script for this sort of script?

packages/core/scripts/hardhat/verifyDeployedContracts.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+NETWORK_NAME=$1
+if [ ! "$NETWORK_NAME" ]; 
+then 
+    echo "Must specify network name"
+    exit 1
+fi
+
+# The network to verify Etherscan contracts for
+yarn hardhat etherscan-verify --license AGPL-3.0 --force-license --network $NETWORK_NAME
```
mrice32, line 10: Very cool!

packages/common/src/Ethers.js
nicholaspai: Got it from this ethers [issue](https://github.com/ethers-io/ethers.js/issues/207#issuecomment-398691570)

packages/core/deploy/001_deploy_finder.js
nicholaspai: yup! For each contract we add extra functionality to run if the contract gets newly deployed.

packages/core/deploy/008_mock_oracle.js
nicholaspai: Ah yes we should totally use `prod` and `test` tags. I'm thinking we give each contract two tags at least, its name (for use as a dependency in other deployment scripts) and a `prod` tag. This way we can leave out the `prod` tag from the `MockOracle` so we can just run `yarn hardhat deploy --tags prod` to deploy all prod contracts

packages/core/README.md
nicholaspai: Yeah that's what my crazy thought is long term. They could be ported pretty easily as scripts, which would make the CLI invocation a lot less verbose. The main benefit it seems of using tasks is you get access to the hardhat runtime environment which has a lot of nifty objects: https://hardhat.org/advanced/hardhat-runtime-environment.html

packages/core/networks/hardhat/deployed.json
nicholaspai: This is not used in any hardhat scripts. This is generated by `yarn hardhat export --export-all /path/to/deployed.json`, which is run manually by the caller. This is so we can programmatically save deployed addresses from the local `deployments/` directory which does not get committed.

The ABI's are not needed, but `export` doesn't strip them out of the box.

nicholaspai: The main use case of `export` i think is to allow a front-end client to run the `export` command from CLI and easily pull in abi/addresses into their app. So maybe just don't commit this file for now

packages/core/scripts/hardhat/deployContracts.sh
nicholaspai: Yeah I just created `production`, `sink-oracle` and `source-oracle` tags for now. Will remove this script

packages/core/scripts/hardhat/tasks/registry.js
nicholaspai: because theyre imported in `hardhat.config.js`. Note the `require("/path/to/registry.js")` at the bottom

packages/core/scripts/hardhat/tasks/registry.js
nicholaspai: I agree! you also get CLI input args and params super easily along, and tasks can call each other easily as well: https://hardhat.org/guides/create-task.html

I think we should switch all of our scripts over to hardhat tasks because we can remove all of the logic that sets up parts of web3.

packages/core/scripts/hardhat/verifyDeployedContracts.sh
nicholaspai: We need this actually because our license tag `SPDX-License-Identifier: AGPL-3.0-only` is "not supported" by Etherscan?

packages/common/src/Ethers.js
mrice32: Ah yep! I think @Tulun used this some snippet in the voter dapp. May be useful (later) to pull this out into a separate lib.

packages/core/scripts/hardhat/verifyDeployedContracts.sh
mrice32: Weird, I thought it was listed on etherscan when doing manual deployment....

packages/core/scripts/hardhat/tasks/registry.js
mrice32: I think the big downside is coupling ourselves (again) to a specific framework. I can see arguments either way. But definitely agreed. Might be nice to avoid web3/provider setup, although, we could always write a utility function to do all that work for us (or just use getWeb3 :)).

packages/core/hardhat.config.js
```diff
@@ -1,17 +1,57 @@
...
+  etherscan: {
+    // Your API key for Etherscan
+    // Obtain one at https://etherscan.io/
+    apiKey: process.env.ETHERSCAN_API_KEY
   }
 };
 
+// Tasks: These tasks are conveniently available via the hardhat CLI: `yarn hardhat <TASK>`
+// Documentation on creating tasks here: https://hardhat.org/guides/create-task.html
+require("./scripts/hardhat/tasks/finder");
```
mrice32, line 54: (Nothing to change now)

One advantage to [hardhat](https://hardhat.org/guides/scripts.html) (or node) scripts rather than tasks is that we don't need to import them into the config. That matters for two reasons:
- Every time you add a "script" you have to add it to this long list of imports. If you forget, it doesn't work.
- Every time the config is loaded, node will need to parse (and execute the global-level stuff, like adding the task to hardhat's global list) these tasks, even when not used.

packages/core/scripts/hardhat/tasks/registry.js
mrice32: Makes sense. I think there are some disadvantages to this over hardhat scripts, but we can talk more about that later.

packages/core/scripts/hardhat/deployContracts.sh
mrice32: Awesome! Very cool. What does the production tag run? Both oracle deployments or just the source?

mrice32 (approved):
LGTM! Nothing blocking, but I left a few responses and comments that we can talk about more as we port more of core to hardhat!

packages/common/src/Ethers.js
nicholaspai: Yeah thats why I started this Ethers.js file to start collecting common ethers utils

packages/core/README.md
nicholaspai: Yep you can use HRE in js scripts

packages/core/scripts/hardhat/deployContracts.sh
nicholaspai: production runs everything, but I'm thinking now we should change the name to "dvm" and not deploy either sink or source. "dvm" would be Finder, Registry, IdWhitelist, etc.

packages/core/README.md
```diff
@@ -70,6 +70,48 @@ const governor = await governor.deployed();
 Note: this use case is not particularly common, but it is sometimes useful to have access to multiple abi versions
 side-by-side.
 
+## Deployment with Hardhat
+
+Here is a list of scripts you can execute:
+
+`yarn void:deploy`
+
+This will deploy your contracts on the in-memory hardhat network and exit, leaving no trace. Quickest way to ensure that deployments work as intended without consequences.
```
chrismaree, line 79: do you want to put the description _above_ the command? makes it a bit easier to read.

packages/core/README.md
```diff
@@ -70,6 +70,48 @@ const governor = await governor.deployed();
...
+
+`yarn hardhat deploy --network <NETWORK-NAME> --tags <TAGS>`
+
+Deploy all contracts to specified network. Requires a `CUSTOM_NODE_URL` HTTP(s) endpoint and a `MNEMONIC` to be set in environment. Available contract tags can be found in `/deploy` scripts, and available networks are found in the `networks` object within `hardhat.config.js`. Tags can be powerful, for example running `yarn hardhat deploy --tags Bridge` will only deploy the Bridge contract its dependencies (such as the Finder).
```
chrismaree, line 83: siiick

packages/core/deploy/007_identifier_whitelist.js
```diff
@@ -0,0 +1,29 @@
...
+
+  if (deployResult.newlyDeployed) {
+    const txn = await execute(
+      "Finder",
+      { from: deployer },
+      "changeImplementationAddress",
+      stringToBytes32(interfaceName.IdentifierWhitelist),
+      deployResult.address
+    );
+    log(
```
chrismaree, line 22: _some_ scripts have logs and others dont. should we make that more consistent?

packages/core/scripts/hardhat/tasks/finder.js
```diff
@@ -0,0 +1,45 @@
+const { task } = require("hardhat/config");
+const { waitForTxn } = require("./utils");
+const { stringToBytes32, interfaceName } = require("@uma/common");
+
+task("setup-finder", "Points Finder to DVM system contracts")
+  .addFlag("registry", "Use if you want to set Registry")
```
chrismaree, line 6: this is cool. will be very powerful for us to define all logic and scripts using these lags.

packages/core/scripts/hardhat/verifyDeployedContracts.sh
chrismaree, line 10: you don't need to tell this function what contracts you are verifying? does it just try _every_ contract? how does this work exactly.

chrismaree (approved):
This is pretty awesome! I _really_ like the tasks. I look forward to us being able to move everything over to using this cleaner, more useable syantx!

good work!

One additional question: did you get running node scripts working correctly with hardhat? i.e the equivalent of `yarn truffle run` using the `yarn hardhat run`.

packages/core/scripts/hardhat/tasks/finder.js
nicholaspai: Agreed, but see Matt's comment about coupling ourselves to another framework. I tend to agree with him, but maybe we should (1) make sure all current scripts are compatible with `yarn hardhat run` and (2) rather than replace all scripts with tasks, add some commonly used tasks that we might want to run via CLI!

packages/core/scripts/hardhat/verifyDeployedContracts.sh
nicholaspai: it just tries to verify every single contract stored in the local `deployments` folder. This is a description of the `etherscan-verify` task that the `hardhat-deploy` plugin adds to hardhat:

```
This plugin adds the etherscan-verify task to Hardhat.

This task will submit the contract source and other info of all deployed contracts to allow etherscan to verify and record the sources.
```

packages/core/hardhat.config.js
mrice32: That's fair. It's just something I think we should consider. I think tasks are meant for high-level custom commands (similar to hardhat compile or deploy, but not built in) that are more general. I don't think it was meant as a place to put you contract-related scripts, which is where `hardhat run` comes in.

PR #2951, opened 2021-05-06
feat(GasEstimator): update the default to use fast gas price

chrismaree:
**Motivation**

Right now the gas estimator returns the `SafeGasPrice` or `fast` gas price, depending on the API called. Ideally, we should always use the _fastest_ price returned. 

An example of where this matters is when proposing UMIPs. The script will use the recommended price and the gas price will move up a few GWEI and before you know it you are stuck with a pending transaction. At all times we should use the `fastest` recommendation to ensure transactions are mined as soon as posible. This should result in a general improvement in quality of life for proposers as well as indirectly improve the bots performance (and reduce how often we need to bump gas).

**Summary**

Update the default.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/helpers/GasEstimator.js
+++ b/packages/financial-templates-lib/src/helpers/GasEstimator.js
@@ -82,8 +82,8 @@ class GasEstimator {
     try {
       const response = await fetch(url);
       const json = await response.json();
-      if (json.fast) {
-        let price = json.fast;
+      if (json.fastest) {
+        let price = json.fastest;
         return price;
       } else {
         throw new Error("Etherchain API: bad json response");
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/helpers/GasEstimator.js
```diff
@@ -82,8 +82,8 @@ class GasEstimator {
     try {
       const response = await fetch(url);
       const json = await response.json();
-      if (json.fast) {
-        let price = json.fast;
+      if (json.fastest) {
```
nicholaspai, line 85: Nice!

nicholaspai (approved):
This is a prudent PR, good job for thinking about it. LGTM

PR #2953, opened 2021-05-06
feat(affiliates): add uniswap v3 single pool LM script

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/564/update-the-liquidity-mining-scripts-to-work-for-a-global-liquidity-mining-program-on-uniswap-v3

**Summary**

This adds very basic logic to run liquidity mining on a single pool over a block range. It outputs % share for each LP in the pool. This also has the ability to run off the most current chain state if block start/ends are omitted.

**Details**

**This calculator makes a couple assumptions that could be challenged**

- a positions liquidity is measured for a single tick, if they are spread from 1 - 100 ticks, at any point only 1/100th of the total liquidity will be considered active
- a positions active tick range is from [lowerTick,upperTick), ie upperTick is not inclusive, if current tick == upperTick the position will not be considered active. This implies that lowerTick and upperTick on a position cannot be equal, since it will not span a full tick

**Some things discovered**

- slot0 on a pool contains a `liquidity` value, this value does not represent active liquidity in the tick as i first assumed, but all active positions liquidity. If some positions span multiple ticks, this will be part of the global liquidity value.  This number cannot be used to normalize a single positions liquidity, since our design only considered liquidity within a tick.
- the nftPosition contract, which is meant to be a nice handle into positions, actually is hard to use because you cannot easily query positions for a single pool, therefore positions are seeded through mint events on the pool we care about. 

**Future work**
- historical queries have not been tested at all since we dont have archive capaibilities on testnet afaik
- more thorough tests would be nice, but not critical since the logic is not too complex
- This is a really simplistic calculator and very slow if considering historical queries from an archive node.  Would like to couple this with persisent ingestions pipeline to query position data on each block, and reference this data vs an archive node.
- Currently the app supports running on a single pool, we would like to create a script which can run on several pools at once 
- Once we can run on several pools, we might want price data to normalize liquidity sizes between them for reward weighting

This is output from a rinkby pool:
image

**Testing**

- [x]  New unit tests created
- [ ]  Untested


packages/affiliates/libs/uniswap/models.ts
```diff
@@ -40,7 +39,7 @@ export function Table<T extends MaybeId>(config: { makeId: MakeId<T>; type: Id }
     const got = await get(id);
     return set({ ...got, ...data });
   }
-  function entries() {
+  async function entries() {
```
daywiss, line 42: these functions are async to mimic the interface if the data was being queried from a database or remote source.

packages/affiliates/apps/UniswapV3LiquidityMining.ts
```diff
@@ -0,0 +1,65 @@
...
+
+const App = (env: ProcessEnv) => async (config: Config) => {
+  assert(env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL for eth provider");
+  const { poolAddress, rewards = 1, startBlock, endBlock, sampleRate = 1 } = config;
+
+  const provider = ethers.getDefaultProvider(env.CUSTOM_NODE_URL);
+  const poolClient = PoolClient(provider);
+  const mining = LiquidityMining({ poolClient });
+
+  let result;
```
mrice32, line 49: nit: do we need to add a type to result? I'm assuming it can't be inferred, right?

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,135 @@
...
+  | "descriptorProxyAddress"
+  | "nonfungibleTokenPositionManagerAddress"
+  | "v3MigratorAddress";
+
+type NetworkContracts = {
+  [key in ContractType]: string;
+};
+
+// Maybe this is available somewhere in uniswaps module, but not sure
+const networks = new Map<NetworkName, NetworkContracts>([
```
mrice32, line 30: Mainly just curious (nothing wrong with what you did here), but why did you choose to use a map rather than a normal object?
```ts
type Networks = {
  [key in NetworkName]: {
    [key in ContractType]: string
  }
};

const networks: Networks = {
  kovan: {
    v3CoreFactoryAddress: "0x7046f9311663DB8B7cf218BC7B6F3f17B0Ea1047",
    weth9Address: "0xd0A1E359811322d97991E03f863a0C30C2cF029C",
    ...
  },
  ...
};
```

Note: this _may_ (not sure) give you a guarantee that all lookups into this object return values that are well-defined (not null or undefined), so you won't need to assert it.

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,135 @@
...
+      v3MigratorAddress: "0x1988F2e49A72C4D73961C7f4Bb896819d3d2F6a3"
+    }
+  ]
+]);
+
+export function getAddress(network: NetworkName, contractName: ContractType) {
+  const addresses = networks.get(network);
+  // typescript complains that addresses object could be undefined even though the network type and Map is defined.
+  assert(addresses, "invalid network: " + network);
+  assert(addresses[contractName], "invalid contract name: " + contractName);
```
mrice32, line 85: As mentioned offline, I think if we do the typing of networks as suggested above, we can drop all of these asserts and make this just:
```ts
return networks[network][contractName];
```

packages/affiliates/libs/uniswap/contracts.ts
daywiss: i tend to use maps in JS because i can check for existince with `has`, and it will correctly return true for values like 0,, which you have to explicitly check for with an object. With typescript though I think you are on to something, I can go back to objects and have strong guarantees and more concise code.

packages/affiliates/apps/UniswapV3LiquidityMining.ts
daywiss: hmm typing result here would really only be for documentation purposes since its only being logged. strange that typescript did not complain about this, but it may have been a warning. I can def type it though.

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -32,3 +24,57 @@ export function convertValuesToString<T>(obj: { [k: string]: any }): T {
...
+
+// percent: numerator, denominator and scale factor
+export function percent(val: string | BigInt, sum: string | BigInt, scale: string | BigInt = 10n ** 18n) {
+  return (BigInt(val) * BigInt(scale)) / BigInt(sum);
+}
+
+export function percentShares(
+  contributions: { [key: string]: string } = {},
+  sum?: string | BigInt,
+  scale: string | BigInt = 10n ** 18n
```
mrice32, line 65: nit: why not make these either strings or BigInts. For all of these methods, can we just require the caller to coerce to BigInt rather than pass around these `|` types?

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -32,3 +24,57 @@ export function convertValuesToString<T>(obj: { [k: string]: any }): T {
...
+  if (BigInt(tick.toString()) >= BigInt(position.tickUpper.toString())) return false;
+  if (BigInt(tick.toString()) < BigInt(position.tickLower.toString())) return false;
+  if (BigInt(position.tickUpper) <= BigInt(position.tickLower)) return false;
+  return true;
+};
+
+// percent: numerator, denominator and scale factor
+export function percent(val: string | BigInt, sum: string | BigInt, scale: string | BigInt = 10n ** 18n) {
+  return (BigInt(val) * BigInt(scale)) / BigInt(sum);
+}
```
mrice32, line 60: Same question here: can we just require the caller to make it a BigInt rather than taking either?

packages/affiliates/apps/UniswapV3LiquidityMining.ts
```diff
@@ -0,0 +1,65 @@
...
+type Config = {
+  poolAddress: string;
+  rewards?: number;
+  startBlock?: number;
+  endBlock?: number;
+  sampleRate?: number;
+};
+
+const App = (env: ProcessEnv) => async (config: Config) => {
+  assert(env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL for eth provider");
```
nicholaspai, line 42: I thought you had to write `process.env` not just `env`?

packages/affiliates/apps/UniswapV3LiquidityMining.ts
```diff
@@ -0,0 +1,65 @@
...
+  const { poolAddress, rewards = 1, startBlock, endBlock, sampleRate = 1 } = config;
```
nicholaspai, line 43: Maybe set this default config object outside of this function and in all caps to make it more clear these are the defaults

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,135 @@
...
+  return addresses[contractName];
+}
+
+// Wraps some functionality to get state from a single pool, gets pool global state, individual positions
+// and basic state reconstruction from events.
+export function PoolClient(provider: ethers.providers.Provider) {
+  const abi = UniswapV3Pool.abi;
+  // gets the slot0 and liquidity from pool at a block number ( assuming archive node)
+  async function getPoolState(params: { blockNumber: number | string | undefined; address: string }) {
```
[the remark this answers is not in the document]
daywiss: hmmmm yea ok. ideally if i ethers exposed the type i would use that, but we can narrow it to just what we are currently using

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,135 @@
...
+    const { blockNumber, address } = params;
+    const contract = new ethers.Contract(address, abi, provider);
```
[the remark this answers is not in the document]
daywiss: i will attempt it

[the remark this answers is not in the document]
daywiss: i inspected their package and they dont provide any typechain artifacts only the abi and bytecode from what i can see.

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -32,3 +24,57 @@ export function convertValuesToString<T>(obj: { [k: string]: any }): T {
...
+// this is stronger than position, but compatible, since these fields have been validated to exist
+type ActivePosition = Position & {
+  liquidity: BigNumberish;
+  tickUpper: BigNumberish;
+  tickLower: BigNumberish;
+};
+
+// figure out if a position is active by looking a liquidity and combination of current tick and position bounds
+// this assume the upper tick is not inclusive, meaning if the upper tick matches the current tick, its not active
+export const IsPositionActive = (tick: BigNumberish) => (
```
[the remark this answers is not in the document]
daywiss: i usually do this if it something that is class-like, ie it has a "constructor" and returns an instance of something, in this case its a function with the "tick" value set on construction.  Not sure if that convention makes sense and willing to change it.  Here is another example of this: https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/clients/FinancialContractClient.js#L73

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -144,36 +118,50 @@ export const Pools = () => {
...
+  function getOrCreate(address: string) {
+    if (store.has(address)) return store.get(address);
+    return create(address);
+  }
+  function create(addr: string) {
+    assert(!store.has(addr), "Already has address");
+    store.set(addr, "0");
+    return "0";
+  }
+  function add(addr: string, amount: string) {
```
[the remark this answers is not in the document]
daywiss: chris asked a similar question on last pr, why not just use BNs. The main reason is to keep data within the table serializable without any extra steps, ie it can be written to DB without any further adapting, the second reason is because it think this should work with BN, BigNumber, BigInt, and strings are the most predictable way to represent large ints, so its easy to reason about, and im willing to trade that off for computation optimization.

packages/affiliates/libs/uniswap/utils.ts
daywiss: yea its a good point. i think id prefer to keep them as strings so the caller can reason about it easier

packages/affiliates/libs/uniswap/utils.ts
daywiss: going to make these strings in case the caller is using something like BN or BigNumber

packages/affiliates/apps/UniswapV3LiquidityMining.ts
daywiss: its being injected as a parameter

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,121 @@
...
+  // gets the slot0 and liquidity from pool at a block number ( assuming archive node)
+  async function getPoolState(params: { blockNumber: BlockNumber; address: string }) {
+    const { blockNumber, address } = params;
+    const contract = new ethers.Contract(address, abi, provider);
+    const slot0 = await contract.slot0({ blockTag: blockNumber });
+    const liquidity = await contract.liquidity({ blockTag: blockNumber });
+    return {
+      ...slot0,
+      address,
+      liquidity: liquidity.toString()
```
nicholaspai, line 92: So I understand: This `liquidity` value contains the "active" liquidity for some pool, that we measure any LP's tick range against to determine if they are being an active LP who should earn rewards?

nicholaspai (commented):
Meta-level comment, I think you're calculating LM rewards the correct way: rewarding active positions (where the [lower,upper) ticks engulf the active liquidity). And we're going to use the Uniswap's contract to define which ticks are active?

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,121 @@
...
+  | "v3MigratorAddress";
+
+type Networks = {
+  readonly [key in NetworkName]: {
+    readonly [key in ContractType]: string;
+  };
+};
+
+// Maybe this is available somewhere in uniswaps module, but not sure
+const networks: Networks = {
```
chrismaree, line 32: we should coordinate on _where_ we want to put these kinds of addresses. Ideally, we should have a file within `common` that maps all external addresses used through the repo. For example, right now the trader needs to know about `uniswap v2` router and finder and the new trader (for univ3) needs similar info with the `swapRouter` in particular. We can sync up on this later in the eng meeting.

packages/affiliates/libs/uniswap/models.ts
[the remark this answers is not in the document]
chrismaree: ye it is a good point and I think strings are fine for this reason. as a general rule we should just follow the same pattern all over the place and I think strings are the most consistent within the `affiliates package.

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -32,3 +24,58 @@ export function convertValuesToString<T>(obj: { [k: string]: any }): T {
...
+
+// this is stronger than position, but compatible, since these fields have been validated to exist
+type ActivePosition = Position & {
+  liquidity: BigNumberish;
+  tickUpper: BigNumberish;
+  tickLower: BigNumberish;
+};
+
+// figure out if a position is active by looking a liquidity and combination of current tick and position bounds
+// this assume the upper tick is not inclusive, meaning if the upper tick matches the current tick, its not active
```
chrismaree, line 43: is this assumption true? if the upper tick is the current tick I thought it _was_ active?

chrismaree: o, wait. it might just be inclusive with the lower tick 🤔 . in any case can you provide a bit more on this assumption?

packages/affiliates/libs/uniswap/liquidity.ts
```diff
@@ -0,0 +1,123 @@
...
+      poolState,
+      positions
+    };
+  }
+
+  // sums this blocks liquidity across all positions and returns a Balances table which contains the liquidity
+  async function addLiquidity(params: { positions: Position[]; poolState: Pool }, allLiquidity: Balances) {
+    const { poolState, positions } = params;
+    assert(exists(poolState.tick), "requires tick");
+    const activePositions = positions.filter(IsPositionActive(poolState.tick));
```
chrismaree, line 45: this is super clean.

packages/affiliates/libs/uniswap/contracts.ts
```diff
@@ -0,0 +1,121 @@
...
+    };
+  }
+
+  // gets state of position at a block number.
+  async function getPositionState(params: { position: Position; blockNumber: BlockNumber; address: string }) {
+    const { position, blockNumber, address } = params;
+    const contract = new ethers.Contract(address, abi, provider);
+    return {
+      pool: address,
+      ...convertValuesToString<Position>(await contract.positions(position.id, { blockTag: blockNumber }))
```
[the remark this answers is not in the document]
daywiss: This is a called a [generic type](https://www.typescriptlang.org/docs/handbook/2/generics.html). It basically allows the called function to have the context of the type "Position", without having to know anything about the type. You can then use that type within the function, in this case its used within the function to set the functions output type as Position.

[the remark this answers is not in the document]
daywiss: You can see similar things when you create some kind of structure, like a Map, which can take any type, but you could use a generic to say `new Map<string,Position>()` so now it knows the values will always return a `Position`

packages/affiliates/libs/uniswap/contracts.ts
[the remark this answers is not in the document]
mrice32: Ah, okay. Thought you pulled event abis from typechain. No worries.

packages/affiliates/libs/uniswap/utils.ts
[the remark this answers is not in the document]
mrice32: I see. Makes sense

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -144,36 +118,50 @@ export const Pools = () => {
...
+    const amountn = BigInt(amount);
+    assert(amountn >= 0n, "amount must be >= 0: " + amount);
+    const balance = getOrCreate(addr);
+    const result = (BigInt(balance) + amountn).toString();
+    store.set(addr, result);
+    total = total + amountn;
```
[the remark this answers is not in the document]
mrice32: Personally, I agree that BigInts are just better than BNs since you can use all the normal primitive syntax. This is especially convenient for comparison operators.

packages/affiliates/libs/uniswap/utils.ts
```diff
@@ -32,3 +24,58 @@ export function convertValuesToString<T>(obj: { [k: string]: any }): T {
...
+) {
+  const defaultSum =
+    sum ||
+    Object.values(contributions)
+      .reduce((sum, val) => {
+        return sum + BigInt(val);
+      }, 0n)
+      .toString();
+
+  if (defaultSum == "0") return {};
```
mrice32, line 75: Even though this is typescript, we should probably still default to `===`. https://stackoverflow.com/a/57126110
```suggestion
  if (defaultSum === "0") return {};
```

mrice32 (approved):

packages/affiliates/libs/uniswap/models.ts
```diff
@@ -144,36 +118,50 @@ export const Pools = () => {
...
+    return result;
   }
-  const store = new Map<Id, Tick>();
-  return Table<Tick>({ makeId, type: "Ticks" }, store);
+  function sub(addr: string, amount: string) {
+    const amountn = BigInt(amount);
+    assert(amountn >= 0n, "amount must be >= 0: " + amount);
```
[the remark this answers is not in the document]
daywiss: hmm yea technically it doesn't matter, but practically i want to assume we only add/sub with positive values to catch unexpected negative values going in.  If this table is used out of the context of LM, its a useful check to make sure untrusted input could never increase a balance by submitting a negative number.

PR #2955, opened 2021-05-07
fix(Trader): Post wargames changes

chrismaree:
**Motivation**

After testing out the range trader on kovan, there were a few small changes that needed to be added. In particular:
- Uniswap updated on of their events when broke the v3 price feed
- inverted price feed support is required when making trades.

**Summary**

Fixes bots after wargames on kovan

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/trader/src/RangeTrader.ts
```diff
@@ -165,6 +169,13 @@ export class RangeTrader {
       .mul(this.fixedPointAdjustment) // Scale the numerator before division
       .div(this.normalizePriceFeedDecimals(expectedValue));
   }
+
+  // TODO: there are a number of places through the repo that re-use this method. it should be refactored to a common util.
+  _invertPriceSafely(price: BigNumber) {
```
nicholaspai, line 174: +1 to this comment, maybe add it as a common util in this PR and we can refactor it elsewhere as we go?

packages/trader/src/RangeTrader.ts
```diff
@@ -125,6 +125,10 @@ export class RangeTrader {
       desiredPrice: this.formatDecimalString(this.normalizePriceFeedDecimals(desiredPrice))
     });
 
+    if (this.tokenPriceFeed.invertPrice === "true") {
```
nicholaspai, line 128: why not just `if (this.tokenPriceFeed.invertPrice)`

nicholaspai (approved):
Left some nit comments, LGTM

packages/trader/src/RangeTrader.ts
chrismaree: yup! there are a few places of refactoring I want to do like this. I think we have a task tracking it for next sprint

packages/trader/src/RangeTrader.ts
chrismaree: it's a string not a bool!

PR #2960, opened 2021-05-11
feat(common): Refactor HardhatConfig exported module

nicholaspai:
**Motivation**

#2950 introduced some `hardhat` architecture into `packages/core` to demonstrate how we could take advantage of `hardhat-deploy` to deploy and verify contracts more easily, as well as hardhat "tasks" to interact with the contracts via succinct CLI commands.

This PR refactors the logic implemented in `core/hardhat.config.js` into the exported `common/HardhatConfig.js` module so that all packages that use `hardhat` are getting the same hardhat "runtime environment". This means that running `hardhat` tasks from any package, such as root, now works!

This is also a first step in moving some of the `core/scripts` functionality into `common` such that they can be exported as helper modules to other packages. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/package.json
```diff
@@ -33,6 +33,7 @@
     "@ethersproject/bignumber": "^5.0.5",
     "@google-cloud/kms": "^0.3.0",
     "@google-cloud/storage": "^2.4.2",
+    "@nomiclabs/hardhat-ethers": "npm:hardhat-deploy-ethers@^0.3.0-beta.7",
```
chrismaree, line 36: why this?

hardhat.config.js
```diff
@@ -0,0 +1,18 @@
+// Note: this config is only here as a matter of convenience to allow all hardhat commands that can be run from core
```
chrismaree, line 1: this _might_ case issues. you now have effectively created nested hard hat packages. i.e the protocol is has a hardhat and the sub packages have hard hat. I _think_ that if you are within the child package (common/liquidator est.) then you'll use the child packages hard hat and if you are in root of the protocol then you'll use the `core` hard hat (as this one points to the `core` package. does that map to your understanding?

packages/common/src/HardhatConfig.js
```diff
@@ -1,202 +1,20 @@
...
+  // get access to the plugins as well.
   require("@nomiclabs/hardhat-truffle5");
-  require("@nomiclabs/hardhat-etherscan");
   require("hardhat-gas-reporter");
   require("@nomiclabs/hardhat-web3");
+  require("hardhat-deploy");
+  require("@nomiclabs/hardhat-ethers");
+
+  // Custom tasks to interact conveniently with smart contracts.
+  require("./hardhat/tasks");
```
chrismaree, line 13: +1 for putting tasks in a separate directory

packages/common/src/HardhatConfig.js
```diff
@@ -1,202 +1,20 @@
...
-        bre.config.networks.buidlerevm.blockGasLimit = 0x1fffffffffffff;
-        bre.config.networks.buidlerevm.gas = 12000000;
-
-        console.log(chalkPipe("bold.underline")("Running tests in debug mode"));
-      }
-
-      await runSuper(taskArgs);
-    });
-
-  task("etherscan-verification", "Verifies contract on etherscan")
```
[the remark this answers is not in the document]
chrismaree: yeeess! goodby broken unused code!

packages/common/src/HardhatConfig.js
```diff
@@ -225,12 +55,15 @@ function getHardhatConfig(configOverrides) {
       // Your API key for Etherscan
       // Obtain one at https://etherscan.io/
       apiKey: process.env.ETHERSCAN_API_KEY
+    },
+    namedAccounts: {
```
chrismaree, line 59: what does this do?

packages/core/README.md
```diff
@@ -72,11 +72,12 @@ side-by-side.
...
 ## Deployment with Hardhat
 
-Here is a list of scripts you can execute:
+Here is a list of scripts you can execute to take full advantage of `hardhat`'s succinct deployment and verification
+process:
 
 This will deploy your contracts on the in-memory hardhat network and exit, leaving no trace. Quickest way to ensure that deployments work as intended without consequences.
 
-`yarn void:deploy`
+`yarn hardhat eploy`
```
chrismaree, line 80: `eploy`?

chrismaree (approved):
This is a great refactor and much better to put this stuff in common. As we build out more tasks we'll be able to delete a lot of code within many unit tests (basically every test has a very similar `beforeEach` block that we can kill. 

Also, great to remove dead, unused code in these configs!

packages/common/package.json
nicholaspai: This is effecetively installing `hardhat-deploy-ethers` via the recommended installation details below. This adds an `ethers` object to the `hardhat runtime environment` which provides useful utilities such as `getContractAt`.

```
if you use ethers.js we recommend you also install hardhat-deploy-ethers which add extra features to access deployments as ethers contract.

Since hardhat-deploy-ethers is a fork of @nomiclabs/hardhat-ethers and that other plugin might have an hardcoded dependency on @nomiclabs/hardhat-ethers the best way to install hardhat-deploy-ethers and ensure compatibility is the following:

npm install --save-dev  @nomiclabs/hardhat-ethers@npm:hardhat-deploy-ethers ethers
Which means you then need to do require("@nomiclabs/hardhat-ethers") instead of require("hardhat-deploy-ethers") in your hardhat.config.js file.

More details on hardhat-deploy-ethers repo : https://github.com/wighawag/hardhat-deploy-ethers#readme
```

packages/common/src/HardhatConfig.js
chrismaree: I see! that's pretty awesome

PR #2961, opened 2021-05-11
feat(run-transaction): add multi EOA transaction runner to DSProxy bots

chrismaree:
**Motivation**

DSProxies can accommodate multiple privileged accounts by using `DSAuthority`. This lets n number of EOAs make calls to the DSProxy. This feature can be used to deal with nonce collisions in the bots by cycling threw EOAs that can interact with the DSProxy when a transaction is sent.

**Summary**

If enabled, the `runTransaction` method will check if the current EOA has any pending transactions that are still in the mempool. If it detects any then it will use the _next_ unlocked EOA. By adding these EOAs to `DSAuthority`, they can all act on behalf of the DSProxy, enabling multiple accounts to initiate on-chain actions.

**Details**

Finding a method to compute pending transactions was difficult. There is no native way to do this in web3. it turns out that Infura exposes a method that makes this pretty easy! It simply needs to be called from web3 using the `provider.send` syntax. See infra docs: https://infura.io/docs/ethereum/json-rpc/eth-getTransactionCount

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/common/test/ContractUtils.js
```diff
@@ -13,10 +13,10 @@ contract("ContractUtils.js", function(accounts) {
         const erc20Contract = new web3.eth.Contract(ERC20ABI, erc20.address);
         // Allowance is not set for accounts[0] so this should fail on .call()
         const transaction = erc20Contract.methods.transferFrom(accounts[0], accounts[1], "1");
-        const config = {
+        const transactionConfig = {
           from: accounts[0]
         };
-        await ContractUtils.runTransaction({ transaction, config });
+        await ContractUtils.runTransaction({ web3, transaction, transactionConfig });
```
chrismaree, line 19: prop name was changed from `config` to `transactionConfig` to be more explicit. Method now needs a `web3` instance to be passed in as well.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -10,7 +10,16 @@ class DSProxyManager {
...
-  constructor({ logger, web3, gasEstimator, account, dsProxyFactoryAddress, dsProxyFactoryAbi, dsProxyAbi }) {
+  constructor({
+    logger,
+    web3,
+    gasEstimator,
+    account,
+    dsProxyFactoryAddress,
+    dsProxyFactoryAbi,
+    dsProxyAbi,
+    availableAccounts = 1
```
chrismaree, line 21: only in the case of the `DSProxyManager` do we need `availableAccounts` as this only works with DSProxy enabled bots. If not set, defaults to 1 (first and only EOA).

packages/trader/src/index.ts
```diff
@@ -54,9 +55,10 @@ export async function run(logger: winston.Logger, web3: Web3): Promise<void> {
       web3,
       gasEstimator,
       account: accounts[0],
-      dsProxyFactoryAddress: config.dsProxyFactoryAddress || getAddress("DSProxyFactory", networkId),
+      dsProxyFactoryAddress: config.dsProxyConfig?.dsProxyFactoryAddress || getAddress("DSProxyFactory", networkId),
       dsProxyFactoryAbi: getAbi("DSProxyFactory"),
-      dsProxyAbi: getAbi("DSProxy")
+      dsProxyAbi: getAbi("DSProxy"),
+      availableAccounts: config.dsProxyConfig?.availableAccounts || 1
```
mrice32, line 61: nit: just curious -- what are the downsides to just assume all accounts loaded into web3 are available to send transactions? Backwards compatibility?

mrice32 (approved):
Looks awesome! only minor comments.

packages/trader/src/index.ts
chrismaree: if the other EOAs dont have permission to access the DSProxy then transactions will fail. we could update the `DsProxyManager` to figure out how many EOAs it has loaded that have permissions then dynamically update this value at execution time? WDYT?

packages/common/src/TransactionUtils.js
```diff
@@ -0,0 +1,128 @@
...
+/**
+ * Returns the number of pending transaction an account has. This method uses `provider.send` syntax. This undocumented
+ * web3 method lets you call direct jsonrpc methods on the provider. This method differs from the web3.js getTransactionCount
+ * by including mempool transactions. see https://infura.io/docs/ethereum/json-rpc/eth-getTransactionCount
+ * @param {*Object} web3.js object for making queries and accessing Ethereum related methods.
+ * @param {*string} account account to check.
+ * @returns number representing the number of transactions, including pending.
+ */
+const getPendingTransactionCount = async (web3, account) => {
+  const sendRpc = util.promisify(web3.currentProvider.send).bind(web3.currentProvider);
```
nicholaspai, line 102: So this will only work with infura providers? Wb alchemy for example, or other default nodes used in Polygon

packages/common/src/TransactionUtils.js
```diff
@@ -0,0 +1,128 @@
...
+ * using methods called on contracts directly after state changes. Max blocking time should be ~ 15 seconds.
+ * @param {Object} web3 Provider from Truffle/node to connect to Ethereum network.
+ * @param {number} blockerBlockNumber block execution until this block number is mined.
+ */
+const blockUntilBlockMined = async (web3, blockerBlockNumber) => {
+  if (argv._.indexOf("test") !== -1) return;
+  for (;;) {
+    const currentBlockNumber = await web3.eth.getBlockNumber();
+    if (currentBlockNumber >= blockerBlockNumber) break;
+    await new Promise(r => setTimeout(r, 500));
```
nicholaspai, line 124: Seems potentially useful to parameterize the delay time between `setTimeout`'s, and setting it by default to 500.

packages/disputer/src/disputer.js
```diff
@@ -204,7 +200,8 @@ class Disputer {
           message: "Position has been disputed!👮‍♂️",
           liquidation: disputeableLiquidation,
           disputeResult: logResult,
-          totalPaid: returnValue
+          totalPaid: returnValue.toString(),
+          transactionConfig
```
nicholaspai, line 204: Any particular reason to add back `transactionConfig` to logs or just for completeness? I removed because I recall it doesn't print particularly well and sometimes one of the keys in the `transactionConfig` would turn up null, causing the Logger to throw.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -146,20 +159,20 @@ class DSProxyManager {
...
     });
 
     this.logger.info({
       at: "DSProxyManager",
       message: "Executed function on a freshly deployed library, created in the same tx 🤗",
       callData,
-      tx: receipt.transactionHash
+      tx: receipt.transactionHash,
+      transactionConfig,
+      returnValue
```
nicholaspai, line 175: Do we need to `.toString` the `returnValue`?

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
```diff
@@ -115,21 +128,21 @@ class DSProxyManager {
...
 
     this.logger.info({
       at: "DSProxyManager",
       message: "Executed function on deployed library 📸",
       libraryAddress,
       callData,
-      tx: receipt.transactionHash
+      tx: receipt.transactionHash,
+      transactionConfig,
+      returnValue
```
nicholaspai, line 145: Same comment as below

packages/reporters/liquidation-reporter/excel-writer.ts
```diff
@@ -72,15 +72,18 @@ export function createExcelSheetFromLiquidationDrawDownData(drawDownData: any) {
...
     );
+    console.log("a");
 
     collateralData.activeFinancialContracts.forEach((financialContractInfo: any, index: number) => {
-      sheet.cell(4 + index, 7).string(financialContractInfo.contractAddress);
-      sheet.cell(4 + index, 8).number(Number(Number(financialContractInfo.collateralValueInUsd).toFixed(2)));
-      sheet.cell(4 + index, 9).string(financialContractInfo.contractPriceIdentifier);
-      sheet.cell(4 + index, 10).date(new Date(financialContractInfo.contractExpirationTime * 1000));
-      sheet.cell(4 + index, 11).number(financialContractInfo.collateralRequirement);
+      sheet.cell(4 + index, 7).string(financialContractInfo.contractAddress || "unknown address");
```
nicholaspai, line 78: Is all of the `liquidation-reporter` work indepenednt of the multi-EOA stuff? Just want to make sure I shouldn't consider these as related changes

nicholaspai (commented):
Awesome work! This nonce management is going to be SUPER powerful, great start to that feature.

packages/common/src/TransactionUtils.js
chrismaree: not actually 100% sure. I will test it out on cloud flare (our backup node) to double check. it is a normal web3 endpoint that should work with any web3 provider, in theory.

packages/common/src/TransactionUtils.js
chrismaree: can do. this util was already in the repo but I agree setting a param as default to 500 would be better.

packages/disputer/src/disputer.js
chrismaree: ye added just for better logging. I guess it's pretty useless seeing we can always get it from etherscan. as long as we have the tx hash I think we can remove this.

packages/reporters/liquidation-reporter/excel-writer.ts
chrismaree: o ye this should not have been committed. have reverted it. thanks for calling it out!

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: tests I did return a FixedPoint that did not need to be converted. but for consistency, I think we should `toString` it.

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.js
chrismaree: done

packages/disputer/src/disputer.js
chrismaree: Actually looking again I remember the main reason I did this. The `runTransaction` method will _modify_ the config passed in, such as changing the `fromAddress`, `nonce`, `usingOffSetDSProxyAccount` and injects the starting `gasPrice`. By having access to this config we can see if the offset was used and additional data that is not visible from etherscan.

Ideally, I think it should be included to help with understanding what's going on when bumping EOA accounts.

chrismaree: however, you are right that this will be pretty noisy... not sure exactly. WDYT?

evaldofelipe:
@chrismaree any updates on this PR?

packages/trader/src/index.ts
mrice32: Why would we load those EOAs into the bot if they didn't have access to the DSProxy?

packages/disputer/src/disputer.js
nicholaspai: These logs are relatively rare, might as well leave it in since we're modifying `transactionConfig` now

nicholaspai (approved):
Thanks for ereviewing my comments, lgtm

PR #2968, opened 2021-05-11
feat(kpi): example bDigg KPI implementation

smb2796:
Should not be merged in. PR is up simply for review.

This follows the same pattern as the previously used [KPI Options FPL](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol), except for it transforms the pre-expiry price to 0.001 instead of 2.

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
@@ -24,17 +24,17 @@ contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable 
         nonReentrantView()
         returns (FixedPoint.Unsigned memory)
     {
-        // If price request is made before expiry, return 2. Thus we can keep the contract 100% collateralized with
-        // each token backed 1:2 by collateral currency. Post-expiry, leave unchanged.
+        // If price request is made before expiry, return 0.001. Thus we can keep the contract 100% collateralized with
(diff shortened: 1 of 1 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/KpiOptionsFinancialProductLibrary.sol
```diff
@@ -24,17 +24,17 @@ contract KpiOptionsFinancialProductLibrary is FinancialProductLibrary, Lockable
         nonReentrantView()
         returns (FixedPoint.Unsigned memory)
     {
-        // If price request is made before expiry, return 2. Thus we can keep the contract 100% collateralized with
-        // each token backed 1:2 by collateral currency. Post-expiry, leave unchanged.
+        // If price request is made before expiry, return 0.001. Thus we can keep the contract 100% collateralized with
+        // each token backed 1:0.001 by collateral currency. Post-expiry, leave unchanged.
         if (requestTime < ExpiringContractInterface(msg.sender).expirationTimestamp()) {
-            return FixedPoint.fromUnscaledUint(2);
+            return FixedPoint.fromUnsigned(1000000000000000);
```
nicholaspai, line 30: Looks right

smb2796: I think it was wrong and should have been `FixedPoint.Unsigned(1000000000000000)` instead

PR #2969, opened 2021-05-11
feat(core): add governor source and sink

mrice32:
**Motivation**

Add a source and sink governor to relay governance transactions to L2s.

**Summary**

Added two contracts: GovernorSource and GovernorSink.

Note: still working on writing tests.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/chainbridge/SinkGovernor.sol
+++ b/packages/core/contracts/chainbridge/SinkGovernor.sol
@@ -1,0 +1,47 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../oracle/interfaces/FinderInterface.sol";
+import "../oracle/implementation/Constants.sol";
+
+/**
+ * @title Governor contract on L2 that receives governance actions from L1.
+ */
+contract SinkGovernor {
+    FinderInterface public finder;
+
+    event ExecutedGovernanceTransaction(address indexed to, uint256 value, bytes indexed data);
+
+    constructor(FinderInterface _finder) {
+        finder = _finder;
+    }
+
+    /**
+     * @notice This method will ultimately be called after a governance execution has been bridged cross-chain from Mainnet
+     * to this network via an off-chain relayer. The relayer will call `Bridge.executeProposal` on this local network,
+     * which call `GenericHandler.executeProposal()` and ultimately this method.
+     * @dev This method should send the arbitrary transaction emitted by the L1 governor on this chain.
+     */
+    function executeGovernance(
+        address to,
+        uint256 value,
+        bytes memory data
+    ) external {
+        require(msg.sender == finder.getImplementationAddress(OracleInterfaces.GenericHandler));
+
+        // Mostly copied from:
+        // solhint-disable-next-line max-line-length
+        // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
+        // solhint-disable-next-line no-inline-assembly
+
+        bool success;
+        assembly {
+            let inputData := add(data, 0x20)
+            let inputDataSize := mload(data)
+            success := call(gas(), to, value, inputData, inputDataSize, 0, 0)
+        }
+        require(success);
+
+        emit ExecutedGovernanceTransaction(to, value, data);
+    }
+}
--- a/packages/core/contracts/chainbridge/SinkGovernor.sol
+++ b/packages/core/contracts/chainbridge/SinkGovernor.sol
@@ -27,7 +27,10 @@ contract SinkGovernor {
         uint256 value,
         bytes memory data
     ) external {
-        require(msg.sender == finder.getImplementationAddress(OracleInterfaces.GenericHandler));
+        require(
+            msg.sender == finder.getImplementationAddress(OracleInterfaces.GenericHandler),
+            "Generic handler must call"
+        );
 
         // Mostly copied from:
         // solhint-disable-next-line max-line-length
@@ -39,7 +42,7 @@ contract SinkGovernor {
             let inputDataSize := mload(data)
             success := call(gas(), to, value, inputData, inputDataSize, 0, 0)
         }
-        require(success);
+        require(success, "Governance call failed");
 
         emit ExecutedGovernanceTransaction(to, value, data);
     }
--- a/packages/core/contracts/chainbridge/SinkGovernor.sol
+++ b/packages/core/contracts/chainbridge/SinkGovernor.sol
@@ -32,6 +32,7 @@ contract SinkGovernor {
             "Generic handler must call"
         );
 
+        // Note: this snippet of code is copied from Governor.sol.
         // Mostly copied from:
         // solhint-disable-next-line max-line-length
         // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
--- a/packages/core/contracts/chainbridge/SinkGovernor.sol
+++ b/packages/core/contracts/chainbridge/SinkGovernor.sol
@@ -33,7 +33,6 @@ contract SinkGovernor {
         // solhint-disable-next-line max-line-length
         // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
         // solhint-disable-next-line no-inline-assembly
-
         bool success;
         assembly {
             let inputData := add(data, 0x20)
--- a/packages/core/contracts/chainbridge/SourceGovernor.sol
+++ b/packages/core/contracts/chainbridge/SourceGovernor.sol
@@ -61,7 +61,7 @@ contract SourceGovernor is Ownable {
      * @notice Gets the resource id to send to the bridge.
      */
     function getResourceId() public view returns (bytes32) {
-        return keccak256(abi.encode(bytes32("Governor"), currentChainId));
+        return keccak256(abi.encode("Governor", currentChainId));
     }
 
     function _getBridge() internal view returns (IBridge) {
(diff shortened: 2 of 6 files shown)


yarn.lock
```diff
@@ -5969,7 +5969,7 @@
   resolved "https://registry.yarnpkg.com/@uniswap/v3-core/-/v3-core-1.0.0-rc.2.tgz#a1afb3253a7295bec6165ad1d960121e6851a576"
   integrity sha512-vsqkqAHPCKsVi0nWwWeX+mHnSTJ8ZUdu1zAVXB9Mj9A+aeBQGV9foRKs9ufDGJq7S1nqmz+7FdjSOcVoeiUqgQ==
 
-"@uniswap/v3-periphery@^1.0.0-beta.23":
+"@uniswap/v3-periphery@1.0.0-beta.23", "@uniswap/v3-periphery@^1.0.0-beta.23":
```
mrice32, line 5972: When I ran yarn, it created local changes.

packages/core/contracts/chainbridge/SinkGovernor.sol
```diff
@@ -0,0 +1,46 @@
...
+     * to this network via an off-chain relayer. The relayer will call `Bridge.executeProposal` on this local network,
+     * which call `GenericHandler.executeProposal()` and ultimately this method.
+     * @dev This method should send the arbitrary transaction emitted by the L1 governor on this chain.
+     */
+    function executeGovernance(
+        address to,
+        uint256 value,
+        bytes memory data
+    ) external {
+        require(msg.sender == finder.getImplementationAddress(OracleInterfaces.GenericHandler));
```
nicholaspai, line 30: why not add an error description to the require?

mrice32: I can. It seems like we had recently become more comfortable trading require messages for gas (and bytecode) efficiency, but I can add back for better visibility.

packages/core/contracts/chainbridge/SourceGovernor.sol
```diff
@@ -0,0 +1,84 @@
...
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "./IBridge.sol";
+
+/**
+ * @title Governor contract on L2 that receives governance actions from L1.
+ */
+contract SourceGovernor is Ownable {
+    FinderInterface public finder;
+    uint8 public currentChainId;
+    bytes32 internal currentRequestHash;
```
nicholaspai, line 15: So this means that we can only send one governance action at a time? I think this is a reasonable constraint, but I wonder if there is a reason why you didn't use an array of governance actions.

packages/core/contracts/chainbridge/SinkGovernor.sol
mrice32: Done.

packages/core/contracts/chainbridge/SinkGovernor.sol
```diff
@@ -0,0 +1,49 @@
...
+        uint256 value,
+        bytes memory data
+    ) external {
+        require(
+            msg.sender == finder.getImplementationAddress(OracleInterfaces.GenericHandler),
+            "Generic handler must call"
+        );
+
+        // Mostly copied from:
+        // solhint-disable-next-line max-line-length
```
nicholaspai, line 36: Is it worth highlighting that this is also copied from Governor? Just want to be extra verbose with our comments surrounding assembly code because its so hard to review

nicholaspai (approved):
Nice clean code, can you please add comments and detailed `require` error messages before merging? Tests are very simple too. Thoughts on adding an end-to-end test to the `e2e.js` test I have in core?

packages/core/contracts/chainbridge/SourceGovernor.sol
mrice32: 1. That's how the mainnet governor works (i.e. each transaction within a governance proposal must be executed independently). The reason we did this on mainnet was to ensure that you don't end up with block gas limit issues if you try to do too much in a single proposal (like the omnibus UMIP, for instance). 
2. Each transaction on the L2 would line up with one transaction sent by the governor on mainnet, so it's easier to track.
3. Simplicity of encoding/decoding (no arrays of arrays).

I'm definitely open to changing, but this seems simplest to start with. Thoughts?

packages/core/contracts/chainbridge/SinkGovernor.sol
mrice32: Done.

mrice32:
> Nice clean code, can you please add comments and detailed `require` error messages before merging? Tests are very simple too. Thoughts on adding an end-to-end test to the `e2e.js` test I have in core?

Will add an end to end test for sending a gov transaction. Only other case that may have been missed is an invalid deposit call. Will add the latter to the unit tests.

packages/core/contracts/chainbridge/SinkGovernor.sol
```diff
@@ -0,0 +1,50 @@
...
+        finder = _finder;
+    }
+
+    /**
+     * @notice This method will ultimately be called after a governance execution has been bridged cross-chain from Mainnet
+     * to this network via an off-chain relayer. The relayer will call `Bridge.executeProposal` on this local network,
+     * which call `GenericHandler.executeProposal()` and ultimately this method.
+     * @dev This method should send the arbitrary transaction emitted by the L1 governor on this chain.
+     */
+    function executeGovernance(
```
chrismaree, line 25: nit: nat spec not documented correctly.

chrismaree: actually this applies to all methods in this PR.

packages/core/contracts/chainbridge/SourceGovernor.sol
```diff
@@ -0,0 +1,87 @@
...
+        uint8 destinationChainId,
+        address to,
+        uint256 value,
+        bytes memory data
+    ) external onlyOwner {
+        require(currentRequestHash == bytes32(0), "Request hash already set");
+        currentRequestHash = _computeRequestHash(to, value, data);
+        _getBridge().deposit(destinationChainId, getResourceId(), _formatMetadata(to, value, data));
+        currentRequestHash = bytes32(0);
+        emit RelayedGovernanceRequest(destinationChainId, to, value, data);
```
chrismaree, line 44: would it be useful to emit the `currentRequestHash` as well?

packages/core/test/chainbridge/sourceGovernor.js
```diff
@@ -0,0 +1,108 @@
...
+    await erc20.addMember(1, owner);
+    await erc20.mint(sourceGovernor.address, web3.utils.toWei("1"));
+  });
+  it("construction", async function() {
+    assert.equal(await sourceGovernor.finder(), finder.address, "finder not set");
+  });
+  it("resource id", async function() {
+    assert.equal(await sourceGovernor.getResourceId(), getResourceId(chainID), "resource id not computed correctly");
+  });
+  it("relayGovernance", async function() {
```
chrismaree, line 78: nice. this test verifies arbitrary message passing with encoded functions.

chrismaree (approved):
This is pretty awesome! nice and simple. Obviously, it hides all the complexity in the bridge infra. the on-chain stuff is super simple!

chrismaree (approved):
This is pretty awesome! nice and simple. Obviously, it hides all the complexity in the bridge infra. the on-chain stuff is super simple!

packages/core/contracts/chainbridge/SourceGovernor.sol
nicholaspai: Let's keep it this way for simplicity of reasoning about!

packages/core/contracts/chainbridge/SourceGovernor.sol
mrice32: It's really only used internally to track the active request, so I'm not sure what use it would provide to an external observer. Thoughts?

mrice32:
@nicholaspai Added additional tests as you requested. PTAL.

packages/core/contracts/chainbridge/SourceGovernor.sol
nicholaspai: Yeah i don't see any use to storing it

packages/core/test/chainbridge/e2e.js
```diff
@@ -305,4 +346,40 @@ contract("GenericHandler - [UMA Cross-chain Voting]", async accounts => {
...
+    );
+
+    await bridgeSidechain.voteProposal(chainId, expectedDepositNonce, _resourceID, dataHash, { from: relayer1Address });
+    await bridgeSidechain.voteProposal(chainId, expectedDepositNonce, _resourceID, dataHash, { from: relayer2Address });
+    await bridgeSidechain.executeProposal(chainId, expectedDepositNonce, data, _resourceID, {
+      from: relayer1Address
+    });
+
+    assert.equal((await erc20.balanceOf(rando)).toString(), web3.utils.toWei("1"));
+    assert.equal((await erc20.balanceOf(sinkGovernor.address)).toString(), "0");
```
nicholaspai, line 383: This is sick to see in e2e

packages/core/contracts/chainbridge/SinkGovernor.sol
```diff
@@ -0,0 +1,50 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../oracle/interfaces/FinderInterface.sol";
+import "../oracle/implementation/Constants.sol";
+
+/**
+ * @title Governor contract on L2 that receives governance actions from L1.
+ */
+contract SinkGovernor {
```
nicholaspai, line 10: Thoughts on refactoring the logic in this contract into a shared lib between `oracle/Governor` and this contract? Just saying because it reuses logic from Governor.

nicholaspai: This is definitely non-blocking though, just a side comment

mrice32: I'll put together a change to do this, but I would prefer to do it in a follow-up just so OZ can review the change to the Governor independently and make sure it causes no functional changes.

PR #2970, opened 2021-05-12
feat(disputer): single reserve currency disputer + unit tests

chrismaree:
**Motivation**

As part of scaling up our infrastructure, it idea be great to not have to hold collateral for every financial contract within the disputer. To accommodate this, the disputer should be able to buy collateral on the open market when preforming disputes.

**Summary**

This PR ads a smart contract that can be called by a DSProxy to buy collateral on Uniswap which is then used to dispute a liquidation.

**Details**

Follows the same pattern that the `ReserveCurrencyLiquidator` used.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
```diff
@@ -0,0 +1,126 @@
...
+            );
+        }
+
+        // 6. Finally, submit the dispute.
+        TransferHelper.safeApprove(fc.collateralCurrency(), address(fc), totalCollateralRequired.rawValue);
+        fc.dispute(liquidationId, sponsor);
+    }
+
+    // Helper method to work around subtraction overflow in the case of: a - b with b > a.
+    function subOrZero(FixedPoint.Unsigned memory a, FixedPoint.Unsigned memory b)
```
nicholaspai, line 85: This seems like a method we should add to FixedPoint lib since we use it in many places now, wdyT?

nicholaspai (approved):
LGTM, I would have thought we could re-use some contract code from the ReserveLiquidator, wdyt?

mrice32:
@chrismaree is the title correct or am I misunderstanding? Looks like it's saying that this PR just adds unit tests, but it seems there's implementation work in this PR too!

packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
```diff
@@ -0,0 +1,126 @@
...
+     * @param maxReserveTokenSpent maximum number of reserve tokens to spend in the trade. Bounds slippage.
+     * @param deadline abort the trade and dispute if the transaction is mined after this timestamp.
+     **/
+    function swapDispute(
+        address uniswapRouter,
+        address financialContract,
+        address reserveCurrency,
+        uint256 liquidationId,
+        address sponsor,
+        FixedPoint.Unsigned calldata maxReserveTokenSpent,
```
mrice32, line 40: nit: why pass this in as a fixed point since it seems like we only use it when passing its rawValue into functions that take uint256? Put another way, we don't use any fixed point operations, so it might be simpler just to pass it in as a uint256. Thoughts?

packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
```diff
@@ -0,0 +1,126 @@
...
+
+import "../../common/implementation/FixedPoint.sol";
+
+/**
+ * @title ReserveCurrencyDisputer
+ * @notice Helper contract to enable a disputer to hold one reserver currency and dispute against any number of
+ * financial contracts. Is assumed to be called by a DSProxy which holds reserve currency.
+ */
+
+contract ReserveCurrencyDisputer {
```
mrice32, line 19: tiny nit: does this work out of the box for both uniswap v2 and v3? If not, should we name this something that implies it's v2-specific as we've done retroactively for some of the trader functionality?

mrice32 (approved):
LGTM! This is sweet!

packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
chrismaree: it's in the two range trader bots now :) totally could refactor this into Fixedpoint in the future.

chrismaree:
> @chrismaree is the title correct or am I misunderstanding? Looks like it's saying that this PR just adds unit tests, but it seems there's implementation work in this PR too!

The title was wrong! it is indeed implementation + tests :)

packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
chrismaree: it does not work with uniswap v3! this is basically the same as the single reserve currency liquidator in that it only works with v2.

it might be worth us considering how we want to consolidate having multiple implementations of things, such as this reserve liquidator/disputer/trader that operate over different market versions. The simplest solution might be to just _leave_ this as being uni v2 support only until uni v3 has grown enough. Once there is enough liquidity in all the pairs we care about on v3 we can simply upgrade and deprecate the old contracts so we don't need to support two sets. WDYT?

packages/core/contracts/bot-helpers/atomic-disputer/ReserveCurrencyDisputer.sol
chrismaree: good point. have changed it.

PR #2973, opened 2021-05-12
feat: Add polygon networks to truffle and hardhat configs

nicholaspai:

**Motivation**

Now that [infura](https://consensys.net/blog/press-release/infura-and-truffle-now-support-polygon/) supports Polygon, it makes adding the mainnet and testnet into our truffle config very easy.

**Summary**

Can now run truffle and hardhat commands like so:

- `yarn truffle migrate --network polygon-matic/mumbai`
- `yarn hardhat deploy --network matic/mumbai`

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/PublicNetworks.js
+++ b/packages/common/src/PublicNetworks.js
@@ -48,8 +48,7 @@ const PublicNetworks = {
     etherscan: "https://explorer-mainnet.maticvigil.com/",
     customTruffleConfig: {
       confirmations: 2,
-      timeoutBlocks: 200,
-      gasPrice: 1000000000 // 1 gWei
+      timeoutBlocks: 200
     }
   },
   80001: {
@@ -57,8 +56,7 @@ const PublicNetworks = {
     etherscan: "https://explorer-mumbai.maticvigil.com/",
     customTruffleConfig: {
       confirmations: 2,
-      timeoutBlocks: 200,
-      gasPrice: 1000000000 // 1 gWei
+      timeoutBlocks: 200
     }
   }
 };
--- a/packages/common/src/PublicNetworks.js
+++ b/packages/common/src/PublicNetworks.js
@@ -43,9 +43,23 @@ const PublicNetworks = {
     daiAddress: "0xbF7A7169562078c96f0eC1A8aFD6aE50f12e5A99",
     wethAddress: "0xd0A1E359811322d97991E03f863a0C30C2cF029C"
   },
+  137: {
+    name: "polygon-matic",
+    etherscan: "https://explorer-mainnet.maticvigil.com/",
+    customTruffleConfig: {
+      confirmations: 2,
+      timeoutBlocks: 200,
+      gasPrice: 1000000000 // 1 gWei
+    }
+  },
   80001: {
-    name: "mumbai",
-    etherscan: "https://explorer-mumbai.maticvigil.com/"
+    name: "polygon-mumbai",
+    etherscan: "https://explorer-mumbai.maticvigil.com/",
+    customTruffleConfig: {
+      confirmations: 2,
+      timeoutBlocks: 200,
+      gasPrice: 1000000000 // 1 gWei
+    }
   }
 };
 
--- a/packages/common/src/TruffleConfig.js
+++ b/packages/common/src/TruffleConfig.js
@@ -56,12 +56,12 @@ function getNodeUrl(networkName, useHttps = false) {
 // Adds a public network.
 // Note: All public networks can be accessed using keys from GCS using the ManagedSecretProvider or using a mnemonic in the
 // shell environment.
-function addPublicNetwork(networks, name, networkId) {
+function addPublicNetwork(networks, name, networkId, customTruffleConfig) {
   const options = {
     networkCheckTimeout: 10000,
     network_id: networkId,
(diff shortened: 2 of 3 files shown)


packages/common/src/TruffleConfig.js
```diff
@@ -176,6 +177,7 @@ addLocalNetwork(networks, "test");
 // Mainnet fork is just a local network with id 1 and a hardcoded gas limit because ganache has difficulty estimating gas on forks.
 // Note: this gas limit is the default ganache block gas limit.
 addLocalNetwork(networks, "mainnet-fork", { network_id: 1, gas: 6721975 });
+addLocalNetwork(networks, "polygon-matic-fork", { network_id: 137, gas: 6721975, gasPrice: 1000000000 });
```
mrice32, line 180: Just out of curiosity, why is a hardcoded gas price required for the fork network?

packages/common/src/PublicNetworks.js
```diff
@@ -43,9 +43,23 @@ const PublicNetworks = {
     daiAddress: "0xbF7A7169562078c96f0eC1A8aFD6aE50f12e5A99",
     wethAddress: "0xd0A1E359811322d97991E03f863a0C30C2cF029C"
   },
+  137: {
+    name: "polygon-matic",
+    etherscan: "https://explorer-mainnet.maticvigil.com/",
+    customTruffleConfig: {
+      confirmations: 2,
+      timeoutBlocks: 200,
+      gasPrice: 1000000000 // 1 gWei
```
mrice32, line 52: If someone provides the `--gasPrice` parameter, this hardcoded value overrides that value... shouldn't it be the other way around? Seems like we might want to unify how the default and override behavior happens across the different networks.

packages/common/src/PublicNetworks.js
```diff
@@ -43,9 +43,23 @@ const PublicNetworks = {
...
+  },
   80001: {
-    name: "mumbai",
-    etherscan: "https://explorer-mumbai.maticvigil.com/"
+    name: "polygon-mumbai",
+    etherscan: "https://explorer-mumbai.maticvigil.com/",
+    customTruffleConfig: {
+      confirmations: 2,
+      timeoutBlocks: 200,
+      gasPrice: 1000000000 // 1 gWei
```
mrice32, line 61: minor nit: have you thought about just importing web3 at the top and using it to get 1 GWEI so this is easier to read (we do this in the truffle config to interpret the user input).

```js
const Web3 = require("web3");

...
gasPrice: Web3.utils.toWei("1", "gwei");
```

mrice32 (commented):
Very nice! A few minor comments and nits.

packages/common/src/TruffleConfig.js
nicholaspai: Its not , we can remove!

nicholaspai: I left it there on principle

packages/common/src/PublicNetworks.js
nicholaspai: Hmm interesting, the problem is that the default gas price of 20 on ETH is too high for Polygon which can work with 1

packages/common/src/TruffleConfig.js
mrice32: Ah, I commented this before I saw the other hard coded gas prices.

packages/common/src/PublicNetworks.js
nicholaspai: I'll remove this gas price for now, let's just let truffle auto detect it

mrice32: It seems like the gas price is still hardcoded for mainnet. IMO, we should make this hardcoded gas price just be the default and allow the --gasPrice to override it in the main config, but definitely open to other ideas/solutions you have!

mrice32 (commented):
Just left one comment on the `--gasPrice` parameter thread! Otherwise, this LGTM!

mrice32 (approved):
LGTM!

packages/common/src/PublicNetworks.js
```diff
@@ -43,9 +43,21 @@ const PublicNetworks = {
...
+  },
   80001: {
-    name: "mumbai",
-    etherscan: "https://explorer-mumbai.maticvigil.com/"
+    name: "polygon-mumbai",
+    etherscan: "https://explorer-mumbai.maticvigil.com/",
+    customTruffleConfig: {
+      confirmations: 2,
+      timeoutBlocks: 200
+    }
```
daywiss, line 60: just to be clear, i think this is ok, but this is more a meta comment about how to organize data for multi chain support. it seems like both chainids and networkids tend to [correlate](https://chainid.network/), and keying only by network id probably is fine even across chains.  I dont know of any instances where 2 chains would have the same network id, but i am paranoid of this.  Do you guys have any opinions on this?

daywiss (approved):

packages/common/src/PublicNetworks.js
nicholaspai: This is a really great point david, we might want to consder storing a chainID in this mapping of networkID's --> network configs, but I don't see anything to do rn

PR #2974, opened 2021-05-13
feat(price-feed): add iFARMUSD and USDiFARM price feed configs

CryptJS13:
Hey all,

I added the requested price feed configs for iFARM to `DefaultPriceFeedConfigs.js`. As far as I understood this was all that was needed from this PR. Let me know if there's anything I am forgetting or if you find anything wrong in what I have provided.

The config follows the specs presented in [UIMP-89](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-89.md), but fixes a bug in the method used to get the `FARM_PER_SHARE` number.

### DIFF
--- a/packages/core/contracts/common/interfaces/HarvestVaultInterface.sol
+++ b/packages/core/contracts/common/interfaces/HarvestVaultInterface.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+/**
+ * @title Interface for Harvest-style vaults.
+ * @dev This only contains the methods/events that we use in our contracts or offchain infrastructure.
+ */
+abstract contract HarvestVaultInterface {
+    // Return the underlying token.
+    function underlying() external view virtual returns (IERC20);
+
+    // Gets the number of return tokens that a "share" of this vault is worth.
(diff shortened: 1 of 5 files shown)


smb2796 (commented):
Hey @CryptJS13, thanks for submitting!

Left a few general syntactical comments.

Also there is a compatibility issue here that needs to be fixed before this price feed can be used. The `VaultPriceFeed` was made to work with yearn v1 and badger vaults, and does not currently work with iFARM. Specifically there is no `token()` method for the iFARM contract, which is used [here](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/VaultPriceFeed.js#L92).

@mrice32 want to get your eyes on this since you are more familiar with the VaultPriceFeed. Does this seem like something we can accommodate? Looks like all of the behavior is the same, we would just need to switch the name of the read method to `underlying` instead.

CryptJS13:
@smb2796 Thanks for your review! I will work on the comments soon. As far as the 'VaultPriceFeed` goes, I missed the `token()` call earlier. Only saw the `getPricePerFullShare()` which works well. I guess we could make a separate `HarvestVaultPriceFeed` template to fix this of you're open to that.

CryptJS13:
I think I added everything that would be necessary to add the Harvest Vault config. Let me know what you guys think.

CryptJS13:
Alright, sorry guys, I made a bit of a mess here. Gonna fix it up and open a new PR. Sorry!

PR #2978, opened 2021-05-14
fix(affiliates-ci): removed private from package.json

chrismaree:
**Motivation**

If packages are set to `"private": true` then they are not picked up in our new Lerna CI configuration. This PR modifies this to ensure that affiliates is no longer private so that the test will run in CI.

**Summary**

Fixed affiliates CI.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/affiliates/package.json
+++ b/packages/affiliates/package.json
@@ -27,7 +27,6 @@
     "registry": "https://registry.npmjs.com/",
     "access": "public"
   },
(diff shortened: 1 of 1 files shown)


evaldofelipe:
LGTM! nice discovery 👍

daywiss (approved):

PR #2980, opened 2021-05-14
feat: Deploy Polygon mainnet and testnet contracts + update helper scripts

nicholaspai:
**Motivation**

New `137.json` and `80001.json` configs for Polygon deployed contracts. All contracts are verified and linked to the Finder. Deployer is registered with Registry to make price requests for now.

Additional small changes I will leave comments on individually.

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/Constants.js
```diff
@@ -6,11 +6,13 @@ const interfaceName = {
   Store: "Store",
   IdentifierWhitelist: "IdentifierWhitelist",
   CollateralWhitelist: "CollateralWhitelist",
+  AddressWhitelist: "CollateralWhitelist",
   FundingRateStore: "FundingRateStore",
   OptimisticOracle: "OptimisticOracle",
   Bridge: "Bridge",
   GenericHandler: "GenericHandler",
-  MockOracleAncillary: "Oracle"
+  MockOracleAncillary: "Oracle",
```
nicholaspai, line 14: These are added to make the `hardhat` task "setup-finder` (implemented by `common/src/hardhat/tasks/finder.js`) easier to write, doesn't pose any risk to add to this dict IIUC.

packages/common/src/HardhatConfig.js
```diff
@@ -61,6 +67,14 @@ function getHardhatConfig(configOverrides) {
     },
     namedAccounts: {
       deployer: 0
+    },
+    external: {
+      // This allows us to access the same deployed contract addresses that truffle can after running apply-registry,
+      deployments: {
```
nicholaspai, line 73: Adding this `deployments` section allows `deployments.get("X")` to read the address of `X` as written in `core/networks/NETWORK.json` after running `yarn apply-registry`. Apply registry sets addresses in `/core/build/contracts` which can now be read by hardhat.

Note that adding this deployments struct to the `path` object in `hardhat` doesn't work.

packages/common/src/hardhat/tasks/finder.js
```diff
@@ -0,0 +1,71 @@
...
+  .addFlag("generichandler", "Use if you want to set GenericHandler")
+  .addFlag("bridge", "Use if you want to set Bridge")
+  .addFlag("identifierwhitelist", "Use if you want to set IdentifierWhitelist")
+  .addFlag("addresswhitelist", "Use if you want to set AddressWhitelist")
+  .addFlag("financialcontractsadmin", "Use if you want to set FinancialContractsAdmin")
+  .addFlag("optimisticoracle", "Use if you want to set OptimisticOracle")
+  .addFlag("store", "Use if you want to set Store")
+  .addFlag("mockoracle", "Use if you want to set MockOracle as the Oracle")
+  .addFlag("sinkoracle", "Use if you want to set SinkOracle as the Oracle")
+  .addFlag("prod", "Set all production contracts in Finder (i.e. SinkOracle instead of MockOracle)")
```
nicholaspai, line 15: Running `yarn hardhat set-upfinder --prod --network <NETWORK>` sets up the Finder completely assuming that the corresponding networks.json file is configured properly.

packages/common/src/hardhat/tasks/finder.js
```diff
@@ -0,0 +1,83 @@
...
+  .addFlag("generichandler", "Use if you want to set GenericHandler")
+  .addFlag("bridge", "Use if you want to set Bridge")
+  .addFlag("identifierwhitelist", "Use if you want to set IdentifierWhitelist")
+  .addFlag("addresswhitelist", "Use if you want to set AddressWhitelist")
+  .addFlag("financialcontractsadmin", "Use if you want to set FinancialContractsAdmin")
+  .addFlag("optimisticoracle", "Use if you want to set OptimisticOracle")
+  .addFlag("store", "Use if you want to set Store")
+  .addFlag("mockoracle", "Use if you want to set MockOracle as the Oracle")
+  .addFlag("sinkoracle", "Use if you want to set SinkOracle as the Oracle")
+  .addFlag("prod", "Configure production setup in Finder")
```
nicholaspai, line 15: You can set up the Finder really easily now via `yarn hardhat setup-finder --prod --sinkoracle --network matic` for example.

packages/core/deploy/002_deploy_registry.js
```diff
@@ -1,30 +1,14 @@
...
   const { deployer } = await getNamedAccounts();
 
-  const deployResult = await deploy("Registry", {
+  await deploy("Registry", {
     from: deployer,
     args: [],
     log: true
   });
-
-  if (deployResult.newlyDeployed) {
```
nicholaspai, line 15: I removed all this Finder-pointing from the deployment scripts so that the deploy files wouldn't all be linked to the Finder. Practically, I was running each of these scripts individually and wanted to avoid having to deploy the Finder over and over again.

packages/common/src/HardhatConfig.js
chrismaree: nice. very useful that we can re-use these from all packages like this.

chrismaree (approved):
This is pretty awesome! I'm really liking all the scripts we are building up.  great work on making this fit within our existing patterns for dealing with contract addresses.

packages/common/src/HardhatConfig.js
```diff
@@ -1,4 +1,6 @@
 const { getNodeUrl, mnemonic } = require("./TruffleConfig");
+const path = require("path");
+const coreWkdir = path.dirname(require.resolve("@uma/core/package.json"));
```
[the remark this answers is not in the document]
nicholaspai: Doesn't `require.resolve` just return the filename, rather than importing the module?

It seems to work fine after removing `node_modules` from root and `common`. Would you expect otherwise?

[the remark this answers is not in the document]
mrice32: I think if you publish the package and someone installed `@uma/common`, require resolve would error out because there would be no `@uma/core` (since `@uma/common` doesn't install it).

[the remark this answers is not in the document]
mrice32: Have we considered doing something more similar to what we do in the truffle config where we export the config via a function and allow the caller to pass in the working directory?

https://github.com/UMAprotocol/protocol/blob/b53757cd6e131aaca285ebccae625c7f5df88c4e/packages/common/src/TruffleConfig.js#L193

mrice32 (approved):
Outside of my comment around the core dependency, this LGTM!

packages/common/src/HardhatConfig.js
```diff
@@ -49,6 +51,10 @@ function getHardhatConfig(configOverrides) {
       matic: {
         url: getNodeUrl("polygon-matic", true),
         accounts: { mnemonic }
+      },
+      mainnet: {
+        url: getNodeUrl("polygon-matic", true),
```
mrice32, line 56: nit: shouldn't this be "mainnet"?

PR #2981, opened 2021-05-15
fix(core): fix typescript exports

mrice32:
**Motivation**

The current typescript exports are broken for frontend imports because you can't do imports like `@uma/core/x/y/z`.

**Summary**

This change improves the types exported from the index file so they can be imported that way and includes the ethers factories.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -12,3 +12,4 @@ cache
 types
 contract-types
 packages/core/index.ts
+packages/core/browser.ts
--- a/packages/core/README.md
+++ b/packages/core/README.md
@@ -125,14 +125,14 @@ Note: this is a work in progress and the typescript API will likely change and i
 The best support is for Ethers contract types. To construct an Ethers contract in node, simply import from the ethers factories:
 
 ```ts
-import { EthersFactories } from "@uma/core";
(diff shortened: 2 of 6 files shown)


nicholaspai (approved):
LGTM but I'm a bit confused how are generally handling `web3` vs `ethers`. It seems that this PR is biased to support `ethers`, is this the general direction that we should be moving the protocol towards?

mrice32:
> LGTM but I'm a bit confused how are generally handling `web3` vs `ethers`. It seems that this PR is biased to support `ethers`, is this the general direction that we should be moving the protocol towards?

@nicholaspai That's only because typechain's ethers support is much better. There is no index file for web3 that we can re-export. There is one for truffle and ethers. Ethers also has factories in typechain that users can take advantage of.

In terms of typechain support ethers > truffle > web3. All should technically be usable in node, but due to some limitations of the create react app environment, importing files from a package directly (as typechain-web3 requires because it doesn't produce an index file) doesn't seem to be possible, at least in my experimentation.

Does that make sense?

PR #2982, opened 2021-05-17
fix(package.json): bump web3 to 1.3.5

chrismaree:
**Motivation**

Previously, the protocol repo was using web3 1.3.4. this generates irritating logs every time any script is run:
```
web3-shh package will be deprecated in version 1.3.5 and will no longer be supported.
web3-bzz package will be deprecated in version 1.3.5 and will no longer be supported.
```

These are produced in the bots as well, which is pretty useless.

To resolve this, we simply need to bump the web3 version to 1.3.5. as described [here](https://github.com/ChainSafe/web3.js/issues/3898).

**Summary**

Bumps web3 version to fix warning message.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/package.json
+++ b/package.json
@@ -54,7 +54,7 @@
     "secp256k1": "^3.7.1",
     "solc-0.8": "npm:solc@^0.8.4",
     "truffle": "^5.2.3",
-    "web3": "^1.3.4"
+    "web3": "^1.3.5"
   },
   "resolutions": {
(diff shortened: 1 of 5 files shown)


nicholaspai (approved):
Thanks for making this change!

mrice32 (approved):
LGTM! Agreed about that message being annoying.

PR #2986, opened 2021-05-18
feat(price-feed): Add default pf configs for USDDEXTF

dextfprotocol:
Motivation

The [USDDEXTF](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-91.md) price identifier needs price feed config.


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -951,6 +951,20 @@ const defaultConfigs = {
         invertPrice: true
       }
     }
+  },
+  USDDEXTF: {
+    type: "expression",
+    expression: "ETHDEXTF * USDETH",
+    lookback: 7200,
+    minTimeBetweenUpdates: 60,
+    twapLength: 3600,
```
smb2796, line 960: The UMIP specifies a 5 minute TWAP, so to meet UMIP specs `twapLength` should be `300`

smb2796, line 960: Also currently the UMIP specifies that the 5-min TWAP is only applied to the ETHDEXTF price and not the USDETH price. Specifying `twapLength` at this level applies it to both constituent price feeds. If following the specs of the UMIP, this should be moved into the `ETHDEXTF` customFeed below. 

If it seems like a better approach to apply the twap to both, the UMIP can be updated instead. What would you prefer?

smb2796 (commented):
Looks good overall! Just a couple of comments about the twap.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
dextfprotocol: Ok to apply `twapLength` to both constituent price feeds

dextfprotocol: File updated to use 300 secs

evaldofelipe:
@dextfprotocol The PR looks good! Could you please take a look on DCO status on required tests?

dextfprotocol:
I clicked on [DCO Details](https://github.com/UMAprotocol/protocol/pull/2986/checks?check_run_id=2626405479) and run the commands (`git rebase` and `git push`) a couple of times, but I got access denied. Eventually I realised my credential settings were using my personal account, so I switched to dextfprotocol. Then re-execute the 2 commands with success.
Anyway the DCO action is still marked with a red cross and appears as required.
What else can I do?

smb2796:
> > @dextfprotocol The PR looks good! Could you please take a look on DCO status on required tests?
> 
> I clicked on [DCO Details](https://github.com/UMAprotocol/protocol/pull/2986/checks?check_run_id=2626405479) and run the commands (`git rebase` and `git push`) a couple of times, but I got access denied. Eventually I realised my credential settings were using my personal account, so I switched to dextfprotocol. Then re-execute the 2 commands with success.
> Anyway the DCO action is still marked with a red cross and appears as required.
> What else can I do?

Hey @dextfprotocol. You should be able to correctly sign off with the instructions listed here https://github.com/UMAprotocol/protocol/pull/2986/checks.

Did you run these already, or did you only run `git rebase` and `git push`? If you ran the specific commands listed in that link and those didn't work, please let me know!

dextfprotocol:
Hi @smb2796.
Indeed, we executed the 2 commands `git rebase` and `git push` from the terminal (local folder where the code is).
In the page [https://github.com/UMAprotocol/protocol/pull/2986/checks](https://github.com/UMAprotocol/protocol/pull/2986/checks) I cannot see any other command other than `git rebase` and `git push`.

smb2796:
> Hi @smb2796.
> Indeed, we executed the 2 commands `git rebase` and `git push` from the terminal (local folder where the code is).
> In the page https://github.com/UMAprotocol/protocol/pull/2986/checks I cannot see any other command other than `git rebase` and `git push`.

Sorry for the delay @dextfprotocol! Thanks for bearing with me here.

To correctly sign off on your commits, you should:

Run:

```
git rebase HEAD~50 --signoff
```

Which will sign off on your commits, and then you should run:

```
git push --force-with-lease origin master
```

Let me know if either you have already run these without success, or you have any issues with these. Thanks!

dextfprotocol:
Hi @smb2796.
We run both commands successfully (see screenshot). Anyway the DCO error is still present :-(
terminal_logs

smb2796:
I wanted to also add a DEXTFUSD price feed config, so I just added USDDEXTF to that pr in #3017 and closed this one, which solves the issues here. Sorry about the delay here, but the pf config should be good to go!

PR #2996, opened 2021-05-19
feat(price-feed): Add iFARM price feed configs and HarvestVault template

CryptJS13:
Alright, back again!

I updated with the comments left for `DefaultPriceFeedConfig.js`.

I also added a new template for a `HarvestVaultPriceFeed`. It is very similar to the existing `VaultPriceFeed`, but it uses `underlying()` instead of `token()`. This should solve the problems found earlier.

Let me know what you guys think!


mrice32:
@CryptJS13 sorry for the delay -- reviewing now!

packages/financial-templates-lib/src/price-feed/HarvestVaultPriceFeed.js
```diff
@@ -0,0 +1,114 @@
+const { PriceFeedInterface } = require("./PriceFeedInterface");
+const { BlockFinder } = require("./utils");
+const { ConvertDecimals } = require("@uma/common");
+const assert = require("assert");
+
+class HarvestVaultPriceFeed extends PriceFeedInterface {
```
mrice32, line 6: Since this is mostly copied from the VaultPriceFeed, might I propose a way to make this a derived of that one?

```js
class VaultPriceFeedBase {
...
    const underlyingTokenAddress = await this._tokenTransaction().call();
...

  _tokenTransaction() {
    throw new Error("Must be implemented by derived class");
  }
}

// Note: we may rename this in the future to YearnV1Vault or something, but just for simplicity, we can keep the name the same for now.
class VaultPriceFeed extends VaultPriceFeedBase {
  _tokenTransaction() {
    return this.vault.methods.token();
  }
}

class HarvestVaultPriceFeed extends VaultPriceFeedBase {
  _tokenTransaction() {
    return this.vault.methods.underlying();
  }
}
```

You could even put them in the same file if you like. We do something similar in the UniswapPriceFeed!

The way you have it now also works, it's just difficult to tell what's new code and what has changed. From your description, the differences seem pretty small!

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -952,6 +952,42 @@ const defaultConfigs = {
...
+      FARM_PER_SHARE: { type: "harvestvault", address: "0x1571eD0bed4D987fe2b498DdBaE7DFA19519F651" },
+    },
+  },
+  USDiFARM: {
+    type: "expression",
+    expression: `
+        FARMUSD = FARMETH_UNISWAP * ETHUSD;
+        FARMUSD * FARM_PER_SHARE
+    `,
+    invertPrice: true,
```
mrice32, line 983: Invert price is a setting that is specific to the uniswap feed and the cryptowatch feeds I think, so putting it here will mean that the nested uniswap feed inverts the price, but the price per full share is not inverted if I'm reading this correctly. Is that what you're going for?

If you're looking for the inverse of the price feed you defined above, I think this might be simpler:
```js
  USDiFARM: {
    type: "expression",
    expression: "1 / iFARMUSD",
  }
```

Is this what you were going for?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -952,6 +952,42 @@ const defaultConfigs = {
...
+    medianizedFeeds: [
+        { type: "cryptowatch", exchange: "coinbase-pro", pair: "ethusd" },
+        { type: "cryptowatch", exchange: "binance", pair: "ethusdt" },
+        { type: "cryptowatch", exchange: "kraken", pair: "ethusd" },
+    ],
+  },
+  iFARMUSD: {
+    type: "expression",
+    expression: `
+        FARMUSD = FARMETH_UNISWAP * ETHUSD;
```
mrice32, line 967: nit: would you be able to avoid defining the `ETHUSD` feed by doing the following:
```js
FARMUSD = FARMETH_UNISWAP / USDETH
```
?

That is a little less readable, so maybe you'd prefer to do the following:
```js
ETHUSD = 1 / USDETH;
FARMUSD = FARMETH_UNISWAP * ETHUSD;
```

Or is there some other reason you'd prefer to define the `ETHUSD` differently from the way we define the `USDETH`?

mrice32 (commented):
This looks awesome!!! A few minor comments.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
CryptJS13: Will update

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
CryptJS13: TBH it was just copied from the XSUSHIUSD config. I will update it this way.

packages/financial-templates-lib/src/price-feed/HarvestVaultPriceFeed.js
CryptJS13: Will update

CryptJS13:
This update should fix every comment you made!

mrice32 (approved):
This looks great! Thanks for responding to my comments.

To get ci to pass, you probably want to do two things:

1. Run the linter over your files. We recently changed the linter versions, so you may need to reinstall packages just to be safe. You can do this by running this from the top-level:

```sh
yarn
yarn lint-fix
```

2. Follow the instructions here: https://github.com/UMAprotocol/protocol/pull/2996/checks?check_run_id=2628143131 to sign-off your commits and get DCO to pass. Let me know if you have problems with this and I can take a look and give specific instructions to make this work!

mrice32:
@evaldofelipe why are the tests not running?

CryptJS13:
> This looks great! Thanks for responding to my comments.
> 
> To get ci to pass, you probably want to do two things:
> 
> 1. Run the linter over your files. We recently changed the linter versions, so you may need to reinstall packages just to be safe. You can do this by running this from the top-level:
> 
> ```shell
> yarn
> yarn lint-fix
> ```
> 
> 1. Follow the instructions here: https://github.com/UMAprotocol/protocol/pull/2996/checks?check_run_id=2628143131 to sign-off your commits and get DCO to pass. Let me know if you have problems with this and I can take a look and give specific instructions to make this work!

I think I did everything that's necessary from my side, or am I missing something?

mrice32:
Merging this once ci passes.

PR #3001, opened 2021-05-20
fix(financial-templates-lib): make CW price feed error on missing data

mrice32:
**Motivation**

Cryptowatch price feed should throw an error if it doesn't have an OHLC data point for a requested historical price rather than trying to fill the gap.

**Summary**

Slightly modified how the feed searches through the data to find a price.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -106,30 +106,12 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return time < pricePeriod.closeTime;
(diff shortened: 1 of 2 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -106,30 +106,12 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return time < pricePeriod.closeTime;
+      return time < pricePeriod.closeTime && time >= pricePeriod.openTime;
     });
 
-    // If there is no match, that means that the time was past the last data point.
```
nicholaspai, line 112: Good catch not to default to `currentPrice` which could cause even worst errors.

nicholaspai (approved):

chrismaree (approved):
looks good to me. ideally erroring is a much better outcome than using a bad price.

PR #3003, opened 2021-05-20
fix(monitors): update vol monitor to work with recent changes in the CW price feed

mrice32:
**Motivation**

The recent change in the cryptowatch price feed means that requesting historical prices that are _very_ recent no longer works. This causes the vol monitor to throw warnings.

**Summary**

This makes the vol monitor use the current price rather than a recent historical price.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/monitors/src/SyntheticPegMonitor.js
+++ b/packages/monitors/src/SyntheticPegMonitor.js
@@ -310,14 +310,13 @@ class SyntheticPegMonitor {
     // record the minimum and maximum.
     const latestTime = pricefeed.getLastUpdateTime();
 
+    // @dev: This might mean that the current price reported is a bit after the volatility window, but the error
+    // should be small enough that it shouldn't impact the results. Furthermore, the price is not used in the vol
+    // computation (which depends on the min/max), it only is reported alongside it as a reference point.
+    const pricefeedLatestPrice = await pricefeed.getCurrentPrice();
+
     // `_calculateHistoricalVolatility` will throw an error if it does not return successfully.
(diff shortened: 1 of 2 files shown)


packages/monitors/src/SyntheticPegMonitor.js
```diff
@@ -310,15 +310,14 @@ class SyntheticPegMonitor {
     // record the minimum and maximum.
     const latestTime = pricefeed.getLastUpdateTime();
 
+    // @dev: This might mean that the current price reported is a bit after the volatility window, but the error
+    // should be small enough that it shouldn't impact the results. Furthermore, the price is not used in the vol
+    // computation (which depends on the min/max), it only is reported alongside it as a reference point.
+    const pricefeedLatestPrice = await pricefeed.getCurrentPrice();
```
nicholaspai, line 316: Should we `try-catch` the `getCurrentPrice`?

nicholaspai (approved):
Nice quick fix, only question is whether we should be even more paranoid and try-catch the `getCurrentPrice` call

packages/monitors/src/SyntheticPegMonitor.js
mrice32: I think getCurrentPrice isn't supposed to error. According to the docs, it returns null if it doesn't have a price. Will toss a try-catch in just to be safe though. I also just realized it doesn't need an await.

mrice32: Fixed!

PR #3007, opened 2021-05-21
feat(sdk): add sdk with some basic structure and code examples

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/913/create-basic-structure-and-logic-for-starting-the-sdk-api-will-be-using-the-sdk

**Summary**

This adds the sdk package generated through tsdx. Has some docs and code examples as well as high level structure of library. 

**Details**
See readmes for more detail on each folder and code module. Licenses probably incorrect, please advise on this as its automatic with tsdx.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/913/create-basic-structure-and-logic-for-starting-the-sdk-api-will-be-using-the-sdk


packages/sdk/src/libs/clients/registry/registry.ts
```diff
@@ -0,0 +1,55 @@
...
+import Registry from "@uma/core/build/contracts/Registry.json"
+import type { SignerOrProvider} from '../../index.d'
+import {EventFragment, Result} from '@ethersproject/abi'
+import { BlockTag} from "@ethersproject/abstract-provider";
+import {EventFilter,Event} from 'ethers'
+
+export interface TypedEvent<EventArgs extends Result> extends Event {
+  args: EventArgs;
+}
+type NewContractRegisteredEvent = TypedEvent< [string, string, string[]] & { contractAddress: string; creator: string; parties: string[] } >;
```
daywiss, line 12: this is copied from typechain outputs. I dont know how to actually import this, it would be really nice

daywiss:
having some issues with linting and prettify. will update this pr when its fixed but the logic and readmes should still be OK to review

packages/sdk/src/libs/clients/registry/registry.ts
```diff
@@ -0,0 +1,59 @@
+import assert from "assert";
+import { Registry__factory } from "@uma/core/contract-types/ethers";
+import Registry from "@uma/core/build/contracts/Registry.json";
+import type { SignerOrProvider } from "../../index.d";
+import { EventFragment, Result } from "@ethersproject/abi";
+import { BlockTag } from "@ethersproject/abstract-provider";
+import { EventFilter, Event } from "ethers";
+
+export interface TypedEvent<EventArgs extends Result> extends Event {
```
mrice32, line 9: This is really annoying. They really don't have great support for events.

Could we build something like this to make an auto-retrieval of the type? (reference: https://stackoverflow.com/a/62733441). Note: this is untested.

```ts

// We could put all of this in some utils file so it could be 
import type { TypedEventFilter } from "@uma/core/contract-types/ethers/commons";
import { Registry } from "@uma/core/contract-types/ethers/Registry";
// We may be able to make this slightly easier to use by allowing the caller to pass the ContractType and the
// name of the Event, but I'm not sure exactly how that syntax would work.
type GetEventType<EventFilterFunction> = ReturnType<EventFilter> extends TypedEventFilter<infer T> ? T : never; 


....

// There may be cleaner ways to access the properties on Registry.
type NewContractRegistered = GetEventType<Registry["filters"]["NewContractRegistered"]>;
```

packages/sdk/src/libs/tables/base/base.ts
```diff
@@ -0,0 +1,37 @@
+import assert from "assert";
+import { exists } from "../../utils";
+import type { Store, MakeId, MaybeId, HasId } from "../../index.d";
+
+export default function Table<I, D>(config: { makeId: MakeId<I, D>; type: string }, store: Store<I, D>) {
```
mrice32, line 5: I think the design of the store and table are quite nice. It think the types are pretty well articulated and organized. However, I don't quite understand why we need both. One seems to be slightly higher-level than the other, but is there any reason for that? Do we use the store independently?

One (admittedly unlikely issue depending on our ingestion architecture) with keeping them separate is that _sometimes_ you might want the higher-level functions to act in a highly consistent way (from a remote database perspective).

There are two examples of possible issues you might see with a remote database:

```ts
  async function create(data: D & MaybeId<I>) {
    // Doesn't exist.
    const id = exists(data.id) ? data.id : makeId(data);
    // Someone else creates it just after we checked it.
    assert(!(await has(id)), `${type} exists`);
    // We overwrite their newly created datum!!
    return set({ id, ...data });
  }
```

Similarly:
```ts
  async function update(id: I, data: Partial<D>) {
    const got = await get(id);
    // Data changes just after we retrieve the data for id.
    // We revert the other change because `got` is out of date by the time we call set!!
    return set({ ...got, ...data });
  }
```

If table is aware of the underlying database implementation, it can use the tools from that db to create consistent transactions: see the transaction object from google cloud datastore -- https://googleapis.dev/nodejs/datastore/latest/Transaction.html.

mrice32 (commented):
Looks great! Only minor comments!

packages/sdk/.github/workflows/main.yml
```diff
@@ -0,0 +1,32 @@
+name: CI
```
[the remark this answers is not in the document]
daywiss: im actually ok with just removing this, i think it must be a tsdx thing?

packages/sdk/README.md
```diff
@@ -0,0 +1,140 @@
...
+
+This builds to `/dist` and runs the project in watch mode so any edits you save inside `src` causes a rebuild to `/dist`.
+
+To do a one-off build, use `npm run build` or `yarn build`.
+
+To run tests, use `npm test` or `yarn test`.
+
+## Configuration
+
+Code quality is set up for you with `prettier`, `husky`, and `lint-staged`. Adjust the respective fields in `package.json` accordingly.
```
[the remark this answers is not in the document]
daywiss: im actually having a lot of trouble removing this stuff so the monorep can work correctly

packages/sdk/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "node": ">=10"
+  },
+  "scripts": {
+    "start": "tsdx watch",
+    "build": "tsdx build",
+    "test": "tsdx test",
+    "test:watch": "tsdx test --watch",
+    "eslint": "eslint './src/**/*.ts'",
+    "prettier": "prettier './**/*.md' './src/**/*.ts'",
+    "lint": "yarn prettier --write && yarn eslint --fix",
```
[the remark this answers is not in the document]
daywiss: this was actually added when i was hacking around to get linting working. the standard tsdx is `tsdx lint` which i could not get working.

packages/sdk/src/libs/clients/registry/registry.ts
daywiss: will play around with this and see if there is a way thanks

packages/sdk/src/libs/tables/base/base.ts
daywiss: yea these are great points. ill try to respond to each.
1. data consistency - the implicit assumption with these tables is that there should only ever be one process writing, though many can read. with multiple writers you run into the exact problem you are illustrating. a "process" can also be thought of an ordered stream of mutations (calls to table), as long as each mutation is ordered, the reads will be consistent within each function call.
2. database specific optimizations - you are right that there is a lot to take advantage of with google datastore, or any other db, im trying to avoid lock in by only using the most primitive CRUD style methods.  This is less performant, but im favoring flexibility over performance. Its not possible to avoid all DB specific calls, but the idea is to create a consistent interface in front of tech specific code.
3. store vs table: the store is technology specific but data agnostic, while the table is data specific but technology agnostic. The table acts as your interface to data. Its not perfect, but the idea is within your application you could easily switch or add new database "stores" without changing much if any business logic that references tables.  In my exp this flexibility unlocks a lot of power. If these things were combined you would find yourself repeating a lot more code than necessary.
4. using stores independently - you probably wouldnt use it independently unless you just wanted a very simple KV interface to a database. the interface though is what allows you to easily add new database adapters. You can imagine quickly adding a mongodb, redis, or even local storage adapter ( for frontend), since the interface is so simple, and all of a sudden the rest of your app can consume it with very minimal work outside the store.

packages/sdk/src/libs/clients/registry/registry.ts
daywiss: the main blocker here is that i cant figure out a way to get return type from a generic, theres some discussion here, but its hard to wrap my head around: https://stackoverflow.com/questions/50321419/typescript-returntype-of-generic-function

daywiss: actually i think i figured it out: `type GetEventType<F> = ReturnType<F extends Callable ? F : never>extends TypedEventFilter<infer T, infer S> ? TypedEvent<T & S & Result> : never;`

this is compiling, will give it a go

daywiss: it compiles and runs, but doesnt properly typecheck things. I tested by accessing some props which should have been invalid, but typescript did not throw an error.

packages/sdk/package.json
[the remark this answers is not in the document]
daywiss: because this package does not contain an eslint config or prettier config, i think it ends up using the repos. I'm going to leave this in for now, because husky does not seem to be working for me, and i have to run this in console.

packages/sdk/src/libs/clients/registry/README.md
```diff
@@ -0,0 +1,23 @@
...
+```js
+import {ethers} from 'ethers'
+import uma from '@uma/sdk'
+const {Registry} from uma.clients
+
+// assume you have a url injected from env
+const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL)
+
+// get the contract instance, address lookup happens based on network
+const contract = Registry.connect(provider,1)
```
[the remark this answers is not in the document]
daywiss: thanks for pointing this out. ive updated this to be more clear!

packages/sdk/package.json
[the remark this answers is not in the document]
mrice32: Have you tried reinstalling the packages?

Go to the root of the repo and run `yarn`.

packages/sdk/src/libs/tables/base/base.ts
mrice32: > 1. data consistency - the implicit assumption with these tables is that there should only ever be one process writing, though many can read. with multiple writers you run into the exact problem you are illustrating. a "process" can also be thought of an ordered stream of mutations (calls to table), as long as each mutation is ordered, the reads will be consistent within each function call.

This seems like a pretty big assumption we're making if this is the storage layer for all database interactions. I could imagine us violating this if, for instance, we manually need to modify the database outside of the ingestion pipeline.

> database specific optimizations

I'm less worried about optimizations, and more worried about correctness.

> using stores independently - you probably wouldnt use it independently unless you just wanted a very simple KV interface to a database. the interface though is what allows you to easily add new database adapters. You can imagine quickly adding a mongodb, redis, or even local storage adapter ( for frontend), since the interface is so simple, and all of a sudden the rest of your app can consume it with very minimal work outside the store.

Maybe a different question is whether we imagine other interfaces to the store other than Table? Put another way, is there a reason for the extra layer of abstraction since the table has fewer methods and could be simpler to implement.

daywiss: > his seems like a pretty big assumption we're making if this is the storage layer for all database interactions. I could imagine us violating this if, for instance, we manually need to modify the database outside of the ingestion pipeline.

this is a drawback of the design. in practice you can usually get away with mutating directly to database without issue, this is dangerous though in any system, particularly if you have a caching layer. the right way to do this is to expose privileged calls through your service to touch the database, this way you can validate data and guard against changes which should not be allowed, update caches and also avoid inconsistent state.  

> Maybe a different question is whether we imagine other interfaces to the store other than Table? Put another way, is there a reason for the extra layer of abstraction since the table has fewer methods and could be simpler to implement.

I think i might be missing some of the nuances to your thinking here.  Maybe what is missing is that the tables currently are just passthroughs of a store. Eventually tables could be more complicated than stores and have specific calls to change data ( you can imagine a balances table which validates withdrawals against user balance). This business logic can be unit tested with a js map store in memory, the assumption is then it will work with any other store without having to deploy a db for the tests.  In many cases someone can update the table logic without knowing anything specific about how to use a specific database. 

Hopefully this makes some sense, though would be happy to discuss offline as well!

packages/sdk/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
mrice32, line 1: nit: we should drop this license. There's a license at the base of the repo that is not MIT.

packages/sdk/src/libs/clients/registry/client.ts
```diff
@@ -0,0 +1,45 @@
...
+export function getAddress(network: Network): string {
+  const address = RegistryArtifacts?.networks?.[network]?.address;
+  assert(address, "no address found for network: " + network);
+  return address;
+}
+
+export type Instance = ReturnType<typeof Registry__factory.connect>;
+export function connect(address: string, provider: SignerOrProvider): Instance {
+  return Registry__factory.connect(address, provider);
+}
```
mrice32, line 24: Shouldn't `Instance` just be the `Registry` type that you import at the top? Can we drop the `Instance` type altogether in favor of `Registry`?

packages/sdk/src/libs/index.ts
```diff
@@ -0,0 +1,33 @@
...
+export interface Callable {
+  (...args: any[]): any;
+}
+// this convoluted type is meant to cast events to the types you need based on the contract and event name
+// example: type NewContractRegistered = GetEventType<Registry,"NewContractRegistered">;
+export type GetEventType<ContractType extends Contract, EventName extends string> = ReturnType<
+  ContractType["filters"][EventName] extends Callable ? ContractType["filters"][EventName] : never
+> extends TypedEventFilter<infer T, infer S>
+  ? TypedEvent<T & S extends Result ? T & S : never>
+  : never;
```
mrice32, line 33: 😮

packages/sdk/src/libs/stores/README.md
```diff
@@ -0,0 +1,45 @@
...
+- store.test.ts: Any tests you want to run
+
+### Store interface
+
+Stores should try to conform to the interface defined [here](./index.ts). This will allow [tables](../tables/README.md)
+to interoperate with the store seamlesly. See an example store [here](./js-map/store).
+
+### Design Tips
+
+- the name of the stores directory should reflect the underlying technology, ie: google-store, js-map, mongo, etc.
```
mrice32, line 42:
```suggestion
- the name of the store's directory should reflect the underlying technology, ie: google-store, js-map, mongo, etc.
```

packages/sdk/src/libs/stores/README.md
```diff
@@ -0,0 +1,45 @@
...
+- consider injecting your specific database client into the constructor of the store, this makes it possible to
+  mock this for testing later, and lets the user instanciate this outside your class.
```
mrice32, line 44:
```suggestion
  mock this for testing later, and lets the user instantiate this outside your class.
```

packages/sdk/src/libs/stores/README.md
```diff
@@ -0,0 +1,45 @@
...
+- the interface is modeled from a JS Map, so keep in mind this should simply act like a key value store
```
mrice32, line 45:
```suggestion
- the interface is modeled from a JS Map, so keep in mind this should simply act like a key value store.
```

packages/sdk/src/libs/stores/google-datastore/store.ts
```diff
@@ -0,0 +1,53 @@
+import type { Store } from "..";
+import type { Datastore, Key } from "@google-cloud/datastore";
+
+export default function <D>(kind: string, store: Datastore): Store<string, D> {
```
mrice32, line 4: Should we put some TODOs in here so we know that this is still a WIP?

packages/sdk/src/libs/tables/base/README.md
```diff
@@ -0,0 +1,39 @@
+# Base Table
+
+Contains basic functions you want in most of your tables, unless its a really special case.
```
mrice32, line 3: nit: I think we should require these types to adhere to the interface strictly.

packages/sdk/src/libs/clients/emp/client.ts
```diff
@@ -0,0 +1,144 @@
...
+
+export interface EventState {
+  // any address that created a position, regardless of if they have closed it
+  sponsors?: string[];
+  tokens?: Balances;
+  collateral?: Balances;
+  expired?: boolean;
+}
+
+type RequestTransferPositionExecuted = GetEventType<ExpiringMultiParty, "RequestTransferPositionExecuted">;
```
mrice32, line 22: Very cool!

mrice32 (approved):
Minor comments -- LGTM!

packages/sdk/src/libs/clients/registry/client.ts
daywiss: good call!

packages/sdk/src/libs/tables/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+See the [blocks table](./blocks/README.md) as an example.
+
+## Usage Considerations
+
+These tables are designed to work in such a way that data is queried and updated in the application layer
+rather than in the database layer. This is a subtle distinction, but in some usecases this may cause race conditions
+during times where many updates may be happening concurrently ( queried state within table functions may be inconsitent with database state).
+If this applies to your use case, you may need to write a custom table using the database driver directly or
+order your mutations such that they execute serially.
```
daywiss, line 38: i added this section based on our convo @mrice32

PR #3013, opened 2021-05-24
improve(beacon-oracles): validateDeposit should revert after first execution

nicholaspai:
**Motivation**

Currently, anyone can call `Bridge.deposit` directly to induce duplicate `Deposit` events. This is problematic because an offchain Relayer monitoring the `Bridge` is expected to submit transactions to bridge messages across EVM networks.

**Summary**

This PR adds two pending states to the BeaconOracle `PriceRequest` struct. The idea is that an initial call to `SinkOracle.requestPrice` or `SourceOracle.publishPrice` should mark the state as `pendingRequest` and `pendingResolve` respectively. These calls will call `Bridge.deposit` which is expected to subsequently call back to `SinkOracle.validateDeposit` and `SourceOracle.validateDeposit`. These `validateDeposits` should then mark the states as `Requested` and `Resolved`, or revert if the state is not a pending one.

This fix ensures that no one can call `Bridge.deposit` directly after a `Deposit` event was already emitted.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/chainbridge/BeaconOracle.sol
+++ b/packages/core/contracts/chainbridge/BeaconOracle.sol
@@ -83,6 +83,22 @@ abstract contract BeaconOracle {
     }
 
     /**
+     * @notice Derived contract needs call this method in order to advance state from PendingRequest --> Requested 
+     * before _publishPrice can be called.
+     */
+    function _finalizeRequest(
+        uint8 chainID,
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) internal {
+        bytes32 priceRequestId = _encodePriceRequest(chainID, identifier, time, ancillaryData);
+        Price storage lookup = prices[priceRequestId];
+        require(lookup.state == RequestState.PendingRequest, "Price has not been requested");
+        lookup.state = RequestState.Requested;
+    }
+
+    /**
      * @notice Publishes price for a requested query. Will revert if request hasn't been requested yet or has been
      * resolved already.
      */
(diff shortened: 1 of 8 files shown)


packages/core/test/chainbridge/beaconOracle.js
```diff
@@ -46,9 +46,6 @@ contract("BeaconOracle", async (accounts) => {
         event.time.toString() === testRequestTime.toString() &&
         event.ancillaryData.toLowerCase() === testAncillary.toLowerCase()
     );
-    // Requesting another price does not throw an error or emit a PriceRequested event:
-    const txn2 = await beaconOracle.requestPrice(testIdentifier, testRequestTime, testAncillary, { from: owner });
```
nicholaspai, line 50: This logic is tested in implementing `Sink` and `Source` contracts.

packages/core/contracts/chainbridge/BeaconOracle.sol
```diff
@@ -14,7 +14,7 @@ import "../oracle/implementation/Constants.sol";
  * price requests from the DVM on Mainnet.
  */
 abstract contract BeaconOracle {
-    enum RequestState { NeverRequested, Requested, Resolved }
+    enum RequestState { NeverRequested, PendingRequest, Requested, PendingResolve, Resolved }
```
mrice32, line 17: I'm assuming this contract is depending on derived contracts to move it from PendingRequest to Requested and PendingResolve to Resolved? Is that correct?

Might be a little confusing that this state is actively managed by base and derived contracts, but I can't think of another way to do it without creating another storage variable outside of RequestState....

mrice32 (approved):
LGTM!

packages/core/contracts/chainbridge/BeaconOracle.sol
nicholaspai: I see what you're saying. Do you think this would be more clear with (1) better comments, (2) internal methods like `_validateRequest` and `_validatePublish` which just advance the state, or (3) something else?

mrice32: I could imagine 2 being effective, yeah. That way all the logic is somewhat contained in the base class.

nicholaspai: Done, this removed some code from the derived Sink/Source oracles which is nice

mrice32 (approved):
LGTM!

PR #3016, opened 2021-05-25
feat(cfd): initial CFD stub and tests

chrismaree:
**Motivation**

This PR adds initial CFD smart contract and unit tests. A number of basic libraries were also added.

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+import "../../../common/implementation/FixedPoint.sol";
+
+abstract contract ContractForDifferenceFinancialProductLibrary {
+    using FixedPoint for FixedPoint.Unsigned;
+}
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
@@ -4,4 +4,8 @@ import "../../../common/implementation/FixedPoint.sol";
 
 abstract contract ContractForDifferenceFinancialProductLibrary {
     using FixedPoint for FixedPoint.Unsigned;
+
+    function expirationTokensForCollateral(int256 expiryPrice) public view returns (uint256) {
+        return 5e17;
+    }
 }
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
@@ -1,11 +1,15 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.8.0;
 import "../../../common/implementation/FixedPoint.sol";
+import "@openzeppelin/contracts/utils/math/SafeMath.sol";
+
+interface ExpiringContractInterface {
+    function expirationTimestamp() external view returns (uint256);
+}
 
 abstract contract ContractForDifferenceFinancialProductLibrary {
     using FixedPoint for FixedPoint.Unsigned;
+    using SafeMath for uint256;
 
-    function expirationTokensForCollateral(int256 expiryPrice) public view returns (uint256) {
-        return 5e17;
-    }
+    function expirationTokensForCollateral(uint256 expiryPrice) public view virtual returns (uint256);
 }
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
@@ -11,5 +11,5 @@ abstract contract ContractForDifferenceFinancialProductLibrary {
     using FixedPoint for FixedPoint.Unsigned;
     using SafeMath for uint256;
 
-    function expirationTokensForCollateral(uint256 expiryPrice) public view virtual returns (uint256);
+    function computeExpiraryTokensForCollateral(uint256 expiryPrice) public view virtual returns (uint256);
 }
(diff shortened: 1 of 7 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/ContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+import "../../../common/implementation/FixedPoint.sol";
+
+abstract contract ContractForDifferenceFinancialProductLibrary {
```
mrice32, line 5: Why is this abstract if the method is implemented? Should this be an interface?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+    modifier onlyOpenState() {
+        require(contractState == ContractState.Open, "Contract state is not OPEN");
+        _;
+    }
+
+    constructor(
+        uint32 _expirationTimestamp,
+        uint256 _collateralPerPair,
+        address _longTokenAddress,
+        address _shortTokenAddress,
```
mrice32, line 73: nit: can we just make these parameter types `ExpandedIERC20` for simplicity?

Same suggestion for the other address params here.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        collateralToken.safeTransferFrom(msg.sender, address(this), collateralUsed);
+
+        require(longToken.mint(msg.sender, tokensToMint));
+        require(shortToken.mint(msg.sender, tokensToMint));
+    }
+
+    function redeem(uint256 tokensToRedeem) public preExpiration() nonReentrant() returns (uint256 collateralReturned) {
+        require(longToken.burnFrom(msg.sender, tokensToRedeem));
+        require(shortToken.burnFrom(msg.sender, tokensToRedeem));
```
mrice32, line 106: This is so nice.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+
+        collateralToken.safeTransferFrom(msg.sender, address(this), collateralReturned);
+    }
+
+    function expire() public postExpiration() onlyOpenState() nonReentrant() {
+        _requestOraclePriceExpiration();
+        contractState = ContractState.ExpiredPriceRequested;
+    }
+
+    function settleExpired(uint256 longTokensToRedeem, uint256 shortTokensToRedeem)
```
mrice32, line 118: Thoughts on naming this settle?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+        longToken = ExpandedIERC20(_longTokenAddress);
+        shortToken = ExpandedIERC20(_shortTokenAddress);
+        collateralToken = IERC20(_collateralAddress);
+        priceIdentifier = _priceIdentifier;
+
+        financialProductLibrary = ContractForDifferenceFinancialProductLibrary(_financialProductLibrary);
+    }
+
+    function mint(uint256 tokensToMint) public preExpiration() nonReentrant() returns (uint256 collateralUsed) {
+        collateralUsed = FixedPoint.fromUnscaledUint(tokensToMint).mul(collateralPerPair).rawValue;
```
mrice32, line 96: nit: should we be scaling tokensToMint? Seems like we should just cast it to a FixedPoint.Unsigned?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+        require(longToken.burnFrom(msg.sender, longTokensToRedeem));
+        require(shortToken.burnFrom(msg.sender, shortTokensToRedeem));
+
+        // expirationTokensForCollateral is a number between 0 and 1e18. 0 means all collateral goes to short tokens and
+        // 1 means all collateral goes to the long token. Total collateral returned is the sum of payouts.
+        uint256 collateralPerToken = collateralPerPair.div(2);
+
+        uint256 longCollateralRedeemed =
+            FixedPoint
+                .fromUnscaledUint(longTokensToRedeem)
```
mrice32, line 143: Similar to above, I don't think we should be scaling token values, right?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+                .mul(collateralPerToken)
+                .mul(expirationTokensForCollateral)
+                .rawValue;
+        uint256 shortCollateralRedeemed =
+            FixedPoint
+                .fromUnscaledUint(shortTokensToRedeem)
```
mrice32, line 149: Same here.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -0,0 +1,183 @@
...
+        return IdentifierWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.IdentifierWhitelist));
+    }
+
+    function _getOptimisticOracle() internal view returns (OptimisticOracleInterface) {
+        return OptimisticOracleInterface(finder.getImplementationAddress(OracleInterfaces.OptimisticOracle));
+    }
+
+    function _getAncillaryData() internal view returns (bytes memory) {
+        return abi.encodePacked(address(longToken), address(shortToken));
+    }
```
mrice32, line 182: We don't need to do this in this PR, but do you think we should allow the deployer to customize the ancillary data?

mrice32 (commented):
This looks awesome! A few minor comments. As mentioned offline, I think we can move forward with merging this pretty quickly (even without tests).

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
chrismaree: sure. For now I left it like this just to follow the same pattern as the EMP. we totally should change this in the future.

mrice32 (approved):
LGTM!

PR #3018, opened 2021-05-25
feat(api): add api package with basic emp state 

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/914/get-all-emps-and-the-state-for-all-emps

**Summary**
This app allows you to run an API which collects information about EMPs from the block chain, and allows queries through
an restful endpoint.

This is an app architected as an always on monolithic server. Future updates may break this apart into ingestion / consumption services which can run independently, but for now this is the quickest implementation.

**Details**
High level diagram to help review
```
                                                                 +----------------------+
                                                                 |       SDK            |
                                                                 |   packages/sdk       |
                                                                 |   uses several libs  |
                                                                 |  Tables: emp, blocks |
                                +-------------------------+      |  Clients: emp,       |
                                |       API               |      |    registry          |
                                |  main app entry point   | SDK  |                      |
                                |   apps/api/app          +------+                      |
                                |                         |      +----------------------+
                                |    initializes SDK      |
             +------------------+-------------+-----------+--------------------+
             |                                |                                |
             |                                |                                |
             |                                |                                |
+------------v--------------+ +---------------v----------------+ +-------------v-----------+
|                           | |                                | |                         |
|        Block service      | |         Registry Service       | |       EMP service       |
|   stores seen blocks      | |  Gets registered emp addresses | | Fetches/stores emp state|
|   services/blocks         | |      services/emp-registry     | |    services/emp-state   |
|                           | |                                | |                         |
+-----------+---------------+ +--------------+-----------------+ +------------+------------+
            |                                |                                |
            |                                |                                |
            |                                |                                |
            |                 +--------------v----------+                     |
            |                 |                         |                     |
            |                 |     Action service      |                     |
            +---------------->|  request handlers       |<--------------------+
                              |     services/actions    |
                              |                         |
                              +-------------------------+
                              +-------------------------+
                              |                         |
                              |    Express Service      |
                              |   exposes calls to HTTP |
                              |   services/express      |
                              |                         |
                              +-----------+-------------+
                                          |
                                          |
                                          |
                             +------------v-------------+
                             |                          |
                             |     Action Client        |
                             |  simple fetch client to  |
                             |  call into server        |
                             |     utils/action-client  |
                             |                          |
                             +--------------------------+
```

This breaks code down into 3 main categories: apps, services and libraries.
1. **Apps** are found in the `src/apps` folder, theres currently only the api app.
2. **Libraries** are found in `src/libs` folder. Libraries are any code that is not an app or service.
3. **Services** are found in `src/services` folder. Services are like libraries with a particular interface that expect dependencies to be injected in. This is where all the business logic happens. Services are organized by ownership over particular data or functionality. Hopefully the names of the services are self documenting.

There are special services worth mentioning:
- express - this is a basic express server with an RPC style request point that allows POST data which conform to a particular schema
- actions - these are the request handlers in the form of a collection of functions attached to an object.

Example output when requesting emp data in postman:
image

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/914/get-all-emps-and-the-state-for-all-emps


packages/api/src/services/emp-state.ts
```diff
@@ -0,0 +1,132 @@
...
+type Instance = ReturnType<typeof Emp.connect>;
+export default (config: Json, libs: Libs) => {
+  const { registeredEmps, provider, emps } = libs;
+
+  async function calcGcr(params: { totalTokensOutstanding: string | null; totalPositionCollateral: string | null }) {
+    const { totalTokensOutstanding, totalPositionCollateral } = params;
+    assert(uma.utils.exists(totalTokensOutstanding), "requires total tokens outstanding");
+    assert(uma.utils.exists(totalPositionCollateral), "requires total position collateral");
+    if (BigNumber.from(totalTokensOutstanding).gt(0)) {
+      return utils.parseUnits(totalPositionCollateral).div(totalTokensOutstanding);
```
daywiss, line 19: hope this is correct. also is there an example where we calc gcr for tokens/collateral with varying decimals?

packages/api/src/libs/action-client.ts
```diff
@@ -0,0 +1,40 @@
+import fetch from "isomorphic-fetch";
+import type { Json } from "..";
+import assert from "assert";
+import join from "url-join";
+
+// This is a basic client to allow you to call into the server in the form of
+// client(action,arg1,arg2,...etc) => Promise<Json>.
+// see tests for usage.
+export default function Client(host: string) {
```
daywiss, line 9: @Gamaranto i added you to review just to see this client, which u can use as a reference implementation to talk to server.  currently no way to export this to you yet.

nicholaspai:
General question: why did you choose to implement this as something that could be run as a monolithic server, versus a package with individual exported classes/functions?

packages/api/src/apps/api/README.md
```diff
@@ -0,0 +1,56 @@
...
+
+Assuming all dependencies are installed and `.env` is configured:
+`npx ts-node src/start.ts api`
+
+## Usage
+
+All queries can be made to `http:localhost:${EXPRESS_PORT}` through a POST.
+Example: `curl -X POST localhost:8282/actions` will list availabe actions
+
+## Actions
```
chrismaree, line 26: this is really cool. I like the documentation you've got here.

packages/api/src/apps/api/app.ts
```diff
@@ -0,0 +1,65 @@
+import assert from "assert";
+import { tables } from "@uma/sdk";
```
chrismaree, line 2: yyeeess.

packages/api/src/services/emp-state.ts
chrismaree: hhmm. I dont think what you have here will for for all financial contracts. `utils. parseUnits` without a second parameter converts the input number to 1e18. Numerically, this will work fine if the collateral & synthetics are both 1e18 scaled as you'll basically be going collateral*1e18/debt which works fine if they are both 1e18 scaled.

To do this correctly I think you should pass the number of decimals in the collateral/synthetics as a second param into the `parseUnits` method. Note this assumes you are dealing with the _new_ emp where the collateral and synthetics decimals always match. If we want this to work with all old EMPs as well then you will need to do some more involved normalization. see https://github.com/UMAprotocol/protocol/blob/bb4dd0bff53733aad3ced3ce9e78c94a64aa8161/packages/financial-templates-lib/src/clients/FinancialContractClient.js#L318

However, I don't really think that's needed given almost all of these legacy contracts have expired!

packages/api/src/apps/api/app.ts
```diff
@@ -0,0 +1,65 @@
+import assert from "assert";
```
mrice32, line 1: It seems like we have 4 levels of nesting in this package folder structure, and the files seem to be a bit sparse. Curious why we need to specify api twice in the path -- that seems a bit redundant. In terms of moving through the repo, I think going for fewer levels of folder nesting is ideal.

One rule of thumb is that you should probably have around 10 items per level of nesting before you need to create a new level. If you have way less than that at a level, for example, then that should be the bottom level and there should be no folders below. If we need to restructure later to encompass more complexity or files, that seems preferable to starting with a complex, but sparse dir structure.

Thoughts?

Thoughts?

packages/api/src/index.ts
```diff
@@ -0,0 +1,27 @@
+import * as uma from "@uma/sdk";
```
mrice32, line 1: Is this file exporting types...?

packages/api/src/libs/action-client.ts
```diff
@@ -0,0 +1,40 @@
+import fetch from "isomorphic-fetch";
+import type { Json } from "..";
+import assert from "assert";
+import join from "url-join";
+
+// This is a basic client to allow you to call into the server in the form of
+// client(action,arg1,arg2,...etc) => Promise<Json>.
```
mrice32, line 7: We're building client functionality into the API package?

Should this go in a separate package? Seems like we'd ideally not require the API user to need to worry about what's going on behind the API to use it. For instance, if I were downloading this client package, it might be confusing for me to see a bunch of documentation around how to run the API.

packages/api/src/services/index.ts
```diff
@@ -0,0 +1,4 @@
+export { default as Blocks } from "./blocks";
+export { default as Emps } from "./emp-state";
+export { default as Registry } from "./emp-registry";
+export { default as Prices } from "./prices";
```
mrice32, line 4: nit: not a strong opinion, but curious why you chose to export these as defaults, and then name them a level up?

Another way you could do this is to give it a name at the implementation level:
```ts
export function SomeName = ...;
```

And then at the next level up:
```ts
export * from "./ModuleName";
```

or if you want to be specific:
```ts
export { SomeName as SomeOtherName } from "ModuleName";
```

I'm guessing you're aware of all this syntax, but curious why you chose to do default exports and then un-default them at the next level. Is that helpful for some specific use case? Or is it more readable?

mrice32 (commented):
This looks great!

One meta-comment about these PRs. These are really really big and somewhat hard to review. It's hard to put the pieces together when there's so much new stuff. I think the reviews could be way faster and more thorough (and we could catch any potential problems earlier) if we made smaller, more modular PRs. Your code is already structured to be quite modular and separated, so I think it would be really easy to just PR each of these pieces separately from one another.

packages/api/src/services/emp-state.ts
daywiss: ok thanks @chrismaree will update

daywiss:
> General question: why did you choose to implement this as something that could be run as a monolithic server, versus a package with individual exported classes/functions?

for simplicity at first, though the services are written in such a way they could be split up into separate apps with some minor changes. I think ultimately we want to have a persistent ingestion server which updates database tables, and then a server less http server which handles client requests.  its structured so we can get there eventually, but didnt want to do this as first iteration.

packages/api/src/services/index.ts
daywiss: yea its a great question, we talked offline but ill summarize:
- i see naming as separate problem from the implementation in these cases, the writer can just export a single function in most and pass naming onto a higher level process. 
- this creates a nice predictable template when adding new "modules", its almost to the point where you can copy a folder from a past implementation, and replace the logic with your own. no need to find and replace specific module names within the code.
- i like that using the default puts the focus on what's being exported as the main subject, and implies there's no need to export any other named things.

packages/api/src/apps/README.md
```diff
@@ -0,0 +1,18 @@
...
+- [api](./api/README.md): entry point to run all the api logic
+
+## Quickstart
+
+There maybe 1 or more available apps in the `src/apps` folder. To start the `api` app:
+
+Install dependencies: `yarn`
+
+To run `api`:
+`npx ts-node src/start api`
```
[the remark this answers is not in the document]
daywiss: yea technically we shouldnt use ts-node in production. but I'm having blockers to be able to run this with pure node that i'd like to punt until later. I dont fully understand the problem, but it has to do with building the SDK while having imports with nested references. Running node will crash, but ts-node can figure it out.

packages/api/src/apps/api/app.ts
daywiss: yea i could flatten the app folder with just the api app for now. I forsee though splitting this app into potentially 2 other apps, if not more. You can imagine an app which is ingesting price feed data, smart contract data, maybe data from Contentful or something else and putting them in database. Another express app could be running serverless which handles requests. These would fill up the app folder and each have their own readme.

packages/api/src/index.ts
daywiss: hmm this file probably isnt needed, looks like a mistake

packages/api/src/libs/action-client.ts
daywiss: i didnt have a good place to put this file, its for francesco to reference when interfacing with the API.  it shouldn't belong here, or be exported in the api package, but i also didnt think it fit into the SDK.

packages/api/src/services/prices.ts
```diff
@@ -0,0 +1,43 @@
...
+      empAddress,
+      { priceFeedDecimals: decimals, cryptowatchApiKey, tradermadeApiKey }
+    );
+
+    assert(priceFeed, "Create Reference price feed for emp returned an undefined value");
+    await priceFeed.update();
+    return Promise.all([priceFeed.getLastUpdateTime(), priceFeed.getCurrentPrice()]);
+  }
+  return { getCurrentPrice };
+};
```
chrismaree, line 43: I cant comment on the file below but why do we need an empty `tokens.ts` file?

daywiss:
sorry for size of PR, believe it or not tried to break this down but in hindsight could have been done more. will try to improve pr size in future

packages/api/src/index.ts
daywiss: ok took a look at this again. its not a mistake, im importing everything but just using the types currently. will change it to show types being imported.

chrismaree (approved):
LGTM! Great work on layout out this plumbing. I'm super excited to have this API in production!

packages/api/src/libs/action-client.ts
mrice32: Maybe put it in a non-exported examples folder so it's clear it's not production code?

packages/api/src/services/emp-state.ts
```diff
@@ -0,0 +1,162 @@
...
+import { BigNumber, utils } from "ethers";
+const { parseBytes32String } = utils;
+import { asyncValues, calcGcr } from "../libs/utils";
+import { Json, Libs } from "..";
+
+type Instance = uma.clients.emp.Instance;
+export default (config: Json, libs: Libs) => {
+  const { registeredEmps, provider, emps } = libs;
+
+  async function readEmpDynamicState(instance: Instance, address: string) {
```
mrice32, line 14: These make a _ton_ of requests at once. To get faster updates and decreased load on the node, I think we should probably use multicall to wrap all of this into a single on-chain query (and make this a standard pattern for situations like this).

Thoughts?

packages/api/src/services/index.ts
mrice32: makes sense!

mrice32 (approved):
LGTM!

As discussed offline, I think we should also begin (in later PRs) moving towards exporting traditional classes, just for ease of external use.

packages/api/src/services/emp-state.ts
daywiss: great idea, would love to try this

packages/api/src/libs/action-client.ts
daywiss: thats a good idea, i will make a folder like examples, and TODO to move it out there once we have a FE library

PR #3028, opened 2021-05-26
fix(Logger): address flaky error reporting

chrismaree:
**Motivation**

Sometimes the logger transport provides very useless errors, like this:
image

These are hard to debug.

The source of the problem is the `Logger` where we try and extract the `stack` from `Error` message. This PR modifies this to try and pull other parts of the log message to not return `null`.

**Summary**

I discovered this by looking in the spoke logs and saw this:
image
 
In the case of this error, there is no `stack`. this is why the error showed up as `null`. Rather, we should first try extract the `stack` and if this does not exist then go for the `message`. Further, we can try calling the `.toString()` method that every `Error` object has. Finally, if this does not exist produce an error string to inform that something went wrong.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/Logger.js
+++ b/packages/financial-templates-lib/src/logger/Logger.js
@@ -47,12 +47,11 @@ function errorStackTracerFormatter(logEntry) {
   return logEntry;
 }
 
-// Handle case where `error` is an array of errors and we want to display all
-// of the error stacks recursively. i.e. `error` is in the shape:
-// [[Error, Error], [Error], [Error, Error]]
+// Handle case where `error` is an array of errors and we want to display all of the error stacks recursively.
+// i.e. `error` is in the shape: [[Error, Error], [Error], [Error, Error]]
 function handleRecursiveErrorArray(error) {
   // If error is not an array, then just return the stack for there is no need to recurse further.
-  if (!Array.isArray(error)) return error.stack;
+  if (!Array.isArray(error)) return error.stack || error.message || error.toString();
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/logger/Logger.js
```diff
@@ -47,12 +47,11 @@ function errorStackTracerFormatter(logEntry) {
...
-// Handle case where `error` is an array of errors and we want to display all
-// of the error stacks recursively. i.e. `error` is in the shape:
-// [[Error, Error], [Error], [Error, Error]]
+// Handle case where `error` is an array of errors and we want to display all of the error stacks recursively.
+// i.e. `error` is in the shape: [[Error, Error], [Error], [Error, Error]]
 function handleRecursiveErrorArray(error) {
-  // If error is not an array, then just return the stack for there is no need to recurse further.
-  if (!Array.isArray(error)) return error.stack;
+  // If error is not an array, then just return error information for there is no need to recurse further.
+  if (!Array.isArray(error)) return error.stack || error.message || error.toString() || "could not extract error info";
```
nicholaspai, line 54: WIll this just match from left to right?

nicholaspai (approved):
LGTM

packages/financial-templates-lib/src/logger/Logger.js
chrismaree: yup! 
```
console.log("a"||"b")
a

console.log(null||"b")
b

console.log(null||undefined||"c")
c
```

PR #3030, opened 2021-05-26
feat(cfd): minor improvements to CFD suite

mrice32:
**Motivation**

Build CFD contract.

**Summary**

Made some minor improvements to the CFD suite of contracts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
@@ -1,28 +1,33 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.8.0;
-import "./ContractForDifferenceFinancialProductLibrary.sol";
 
+import "@openzeppelin/contracts/utils/math/SignedSafeMath.sol";
+
+import "./ContractForDifferenceFinancialProductLibrary.sol";
 import "../../../common/implementation/Lockable.sol";
 
 contract LinearContractForDiffrenceFinancialProductLibrary is ContractForDifferenceFinancialProductLibrary, Lockable {
     using FixedPoint for FixedPoint.Unsigned;
-    using SafeMath for uint256;
+    using SignedSafeMath for int256;
 
     struct LinearContractForDifferenceParameters {
-        uint256 upperBound;
-        uint256 lowerBound;
+        int256 upperBound;
+        int256 lowerBound;
     }
 
     mapping(address => LinearContractForDifferenceParameters) contractForDifferenceParameters;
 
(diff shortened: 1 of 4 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
```diff
@@ -1,43 +1,52 @@
...
+        int256 upperBound,
+        int256 lowerBound
     ) public nonReentrant() {
         require(ExpiringContractInterface(contractForDifferenceAddress).expirationTimestamp() != 0, "Invalid address");
         require(upperBound > lowerBound, "Invalid bounds");
-        require(contractForDifferenceParameters[contractForDifferenceAddress].upperBound > 0, "Parameters already set");
+
+        LinearContractForDifferenceParameters memory params =
+            contractForDifferenceParameters[contractForDifferenceAddress];
+        require(params.upperBound != 0 || params.lowerBound != 0, "Parameters already set");
```
chrismaree, line 30: hhmmm. this seems redundant. If `upperBound>lowerBound` then surely checking that the `upperBound>0` for the `contractForDifferenceAddress` is sufficient?

chrismaree: how you have it is more explicit and does not rely on any dependence between requires but still.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -29,9 +30,7 @@ import "../../oracle/implementation/Constants.sol";
 
 contract ContractForDifference is Testable, Lockable {
     using FixedPoint for FixedPoint.Unsigned;
-    using SafeMath for uint256;
     using SafeERC20 for IERC20;
-    using SafeERC20 for ExpandedIERC20;
```
chrismaree, line 34: why remove this?

mrice32: I don't think it's being used.

chrismaree (approved):
LGTM! Thanks for making the changes. I had 2 questions/nits.

packages/core/contracts/financial-templates/common/financial-product-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
mrice32: It's because they are both integers now. So we assume that at least one of them is nonzero. Technically they could both be negative, or one be zero and the other positive, etc. But it shouldn't be allowed for them both to be zero.

PR #3031, opened 2021-05-26
feat(DSProxy): add DSGuard contracts to enable multi-dsproxy ownership

chrismaree:
**Motivation**

To enable the DSPRoxy we use in liquidation to be accessible by multiple EOAs, we require a DSGuard & DSAuthority to set up permissions on who can and cant call methods on the DSProxy over and above the normal "owner" setup. This PR adds the associated smart contracts to facilitate this.

### DIFF
--- a/packages/core/contracts/common/implementation/dsproxy/DSGuardFactory.sol
+++ b/packages/core/contracts/common/implementation/dsproxy/DSGuardFactory.sol
@@ -1,4 +1,4 @@
-// This file was originally take from dapphub dsGuarg and modified. Original source code can be found
+// This file was originally take from dapphub DSGuard and modified. Original source code can be found
 // here: https://github.com/dapphub/ds-guard/blob/master/src/guard.sol
 // Changes are limited to updating the Solidity version and some stylistic modifications.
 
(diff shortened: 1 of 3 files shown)


packages/core/contracts/common/implementation/dsproxy/DSGuardFactory.sol
```diff
@@ -0,0 +1,147 @@
+// This file was originally take from dapphub DSGuard and modified. Original source code can be found
```
mrice32, line 1:
```suggestion
// This file was originally taken from dapphub DSGuard and modified. Original source code can be found
```

packages/core/contracts/common/implementation/dsproxy/DSProxyFactory.sol
```diff
@@ -1,11 +1,7 @@
-// This file was originally take from from Maker: DS Proxy Factory and modified. The original source code can be found
+// This file was originally take from the Maker DS Proxy Factory and modified. The original source code can be found
```
mrice32, line 1:
```suggestion
// This file was originally taken from the Maker DS Proxy Factory and modified. The original source code can be found
```

mrice32 (approved):

nicholaspai (approved):
Meta comment: Is there no way to import the DS-X contracts similar to how we import `@openzeppelin` contracts? Or is the canonical way to copy directly into our repo

PR #3034, opened 2021-05-26
improve(optimistic-oracle): Make bot compatible with OO on Polygon

nicholaspai:
**Motivation**

Bugs encountered while integration testing Polygon mainnet contracts. 

**Summary**

- Fixed `PublicNetworks` import in `FormattingUtils`
- Set `chainId` in transaction configs which seems to be required for the Polygon RPC I'm working with. This is a good feature to have for the future as [replay-protection](https://community.infura.io/t/polygon-rpc-only-replay-protected-eip-155-transactions-allowed-over-rpc/2997) becomes more popular, anyways.
- Refactor `OptimisticOracleEventClient` and `OptimisticOracleMonitor` to handle price requests that did not originate from a `financial-contract`. Currently we assume that the `requester` is a contract that has a `collateralCurrency` callable method, but this is not always going to be the case.
- Add flag to `optimistic-oracle/index.js` to set `SinkOracle` as the Oracle for any non-mainnet network. I think this is a fair assumption that there will only be 1 "Voting" contract, that lives on Mainnet, and every other Oracle contract will be a "SinkOracle"-esque implementation. (Although, we will likely have to change this code again for rollups, for example)

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/FormattingUtils.js
+++ b/packages/common/src/FormattingUtils.js
@@ -1,4 +1,4 @@
-const networkUtils = require("./PublicNetworks");
+const { PublicNetworks: networkUtils } = require("./PublicNetworks");
 
 const BigNumber = require("bignumber.js");
 const moment = require("moment");
--- a/packages/common/src/TransactionUtils.js
+++ b/packages/common/src/TransactionUtils.js
@@ -16,7 +16,7 @@ const runTransaction = async ({ web3, transaction, transactionConfig, availableA
   // Add chainId in case RPC enforces transactions to be replay-protected, (i.e. enforced in geth v1.10,
   // https://blog.ethereum.org/2021/03/03/geth-v1-10-0/).
   transactionConfig.chainId = await web3.eth.getChainId();
-  
+
   // Multiplier applied to Truffle's estimated gas limit for a transaction to send.
   const GAS_LIMIT_BUFFER = 1.25;
 
--- a/packages/common/src/TransactionUtils.js
+++ b/packages/common/src/TransactionUtils.js
@@ -13,6 +13,10 @@ const ynatm = require("@umaprotocol/ynatm");
  * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
  */
 const runTransaction = async ({ web3, transaction, transactionConfig, availableAccounts = 1 }) => {
+  // Add chainId in case RPC enforces transactions to be replay-protected, (i.e. enforced in geth v1.10,
+  // https://blog.ethereum.org/2021/03/03/geth-v1-10-0/).
+  transactionConfig.chainId = await web3.eth.getChainId();
+  
   // Multiplier applied to Truffle's estimated gas limit for a transaction to send.
   const GAS_LIMIT_BUFFER = 1.25;
 
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
@@ -136,6 +136,8 @@ class OptimisticOracleEventClient {
 
     // DisputePrice events.
     for (let event of disputePriceEventsObj) {
+      // The OptimisticOracle contract should ideally emit `currency` as part of this event, but alternatively we can
+      // query the currency address on-chain.
       const requestData = await this._getRequestData(
         event.returnValues.requester,
         event.returnValues.identifier,
@@ -158,6 +160,7 @@ class OptimisticOracleEventClient {
 
     // Settlement events.
     for (let event of settlementEventsObj) {
+      // See explanation above in disputeEventsObj loop.
       const requestData = await this._getRequestData(
         event.returnValues.requester,
         event.returnValues.identifier,
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
@@ -136,6 +136,12 @@ class OptimisticOracleEventClient {
 
     // DisputePrice events.
     for (let event of disputePriceEventsObj) {
+      const requestData = await this._getRequestData(
+        event.returnValues.requester, 
+        event.returnValues.identifier, 
+        event.returnValues.timestamp, 
+        event.returnValues.ancillaryData
+      );     
       this.disputePriceEvents.push({
         transactionHash: event.transactionHash,
         blockNumber: event.blockNumber,
@@ -146,11 +152,18 @@ class OptimisticOracleEventClient {
         timestamp: event.returnValues.timestamp,
         ancillaryData: event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x",
         proposedPrice: event.returnValues.proposedPrice,
+        currency: requestData.currency
       });
     }
 
     // Settlement events.
     for (let event of settlementEventsObj) {
+      const requestData = await this._getRequestData(
+        event.returnValues.requester, 
+        event.returnValues.identifier, 
+        event.returnValues.timestamp, 
+        event.returnValues.ancillaryData
+      );
       this.settlementEvents.push({
         transactionHash: event.transactionHash,
         blockNumber: event.blockNumber,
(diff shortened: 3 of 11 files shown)


packages/common/src/FormattingUtils.js
```diff
@@ -1,4 +1,4 @@
-const networkUtils = require("./PublicNetworks");
+const { PublicNetworks: networkUtils } = require("./PublicNetworks");
```
nicholaspai, line 1: Importing like this has been a bug for a while which has caused `createEtherscanLinkFromtx` to always return `etherscan.io` by default instead of the specific network block explorer url.

packages/common/src/TransactionUtils.js
```diff
@@ -13,6 +13,10 @@ const ynatm = require("@umaprotocol/ynatm");
  * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
  */
 const runTransaction = async ({ web3, transaction, transactionConfig, availableAccounts = 1 }) => {
+  // Add chainId in case RPC enforces transactions to be replay-protected, (i.e. enforced in geth v1.10,
```
nicholaspai, line 16: I raised this issue to the Infura forum [here](https://community.infura.io/t/polygon-rpc-only-replay-protected-eip-155-transactions-allowed-over-rpc/2997) but this is something we should be setting anyways according to [this](https://blog.ethereum.org/2021/03/03/geth-v1-10-0/)

packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
```diff
@@ -146,11 +152,18 @@ class OptimisticOracleEventClient {
         timestamp: event.returnValues.timestamp,
         ancillaryData: event.returnValues.ancillaryData ? event.returnValues.ancillaryData : "0x",
         proposedPrice: event.returnValues.proposedPrice,
+        currency: requestData.currency,
```
nicholaspai, line 155: It's probably an oversight that we don't emit `currency` as part of these events, but fortunately we do store it on-chain. I'm open to also parallelizing the `_getRequestData` call but I think its possibly over-optimizing since this monitor isn't expected to be fetching that many events per run, in production. WDYT? @UMAprotocol/eng

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -210,12 +210,10 @@ class OptimisticOracleContractMonitor {
     this.lastSettlementBlockNumber = this._getLastSeenBlockNumber(latestEvents);
   }
 
-  // Returns helper method for converting collateral token associated with financial contract to human readable form.
-  async _getCollateralDecimalsConverted(financialContractAddress) {
-    const financialContract = new this.web3.eth.Contract(getAbi("FeePayer"), financialContractAddress);
```
nicholaspai, line 215: Because we currently don't set the `currency` in the `event` object coming from the `OptimisticOracleEventClient`, we infer the currency address by assuming that the `requester` is a financial contract that we can call `collateralCurrency` on. This is a bad assumption so this change just seems like a generally good feature to have.

packages/optimistic-oracle/index.js
```diff
@@ -48,15 +47,18 @@ async function run({
...
+    // If bot is not running on mainnet, then assume that the network's "Oracle" in the Finder is a SinkOracle.
+    const isMainnet = Boolean((await web3.eth.getChainId()) === 1);
+
     // Create the OptimisticOracleClient to query on-chain information, GasEstimator to get latest gas prices and an
     // instance of the OO Proposer to respond to price requests and proposals.
     const optimisticOracleClient = new OptimisticOracleClient(
       logger,
       getAbi("OptimisticOracle"),
-      getAbi("Voting"),
+      isMainnet ? getAbi("Voting") : getAbi("SinkOracle"),
```
nicholaspai, line 58: We can probably change this to a mapping once we have other non-Voting Oracles, for rollup networks for example.

packages/common/src/TransactionUtils.js
chrismaree: when would these chainIds not match? web3 is initialized using the providers how could they break?

packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
chrismaree: ye, it totally should be part of the event. the way you have it here is a good workaround for now.

chrismaree: prehaps add a comment explaining this

packages/common/src/TransactionUtils.js
chrismaree: nvm. I see it's not about them mismatching but rather due to replay protection.

packages/financial-templates-lib/src/helpers/allowances.js
```diff
@@ -15,6 +15,9 @@ const setAllowance = async (web3, gasEstimator, ownerAddress, spenderAddress, cu
       tx: await collateralToken.methods.approve(spenderAddress, MAX_UINT_VAL).send({
         from: ownerAddress,
         gasPrice: gasEstimator.getCurrentFastPrice(),
+        // Note: Add chainId in case RPC enforces transactions to be replay-protected, (i.e. enforced in geth v1.10,
+        // https://blog.ethereum.org/2021/03/03/geth-v1-10-0/).
+        chainId: await web3.eth.getChainId(),
```
chrismaree, line 20: is there any way to enforce this on all transactions without needing to add it to every call like this?

chrismaree (approved):
pretty cool stuff! I'm curious if there is a way to always enforce the `chainId`. otherwise, we'd need to add this logic in a bunch more places I believe.

Really excited to be on L2!

packages/common/src/TransactionUtils.js
nicholaspai: Yeah its all about replay protection, and it looks like geth is trying to get more users to start specifying `chainId` going forward

mrice32: Interesting. I noticed this on our own node, and didn't know what to do about it. +1

packages/financial-templates-lib/src/clients/OptimisticOracleEventClient.js
mrice32: Yeah, there's a tradeoff because event data also drives up gas costs. I think your solution here is totally reasonable and agree on the comment suggestion.

packages/financial-templates-lib/src/helpers/allowances.js
nicholaspai: Yeah good call this is suboptimal. I decided to import the `runTransactions` utility and use it here.

mrice32: Yeah, web3 should definitely be doing this internally IMO. But it doesn't look like they've had any recent releases that address it. I wonder if we could wrap the object in some way to force it to tack this on.

Or maybe there's some sort of property we could set in web3?

https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#defaultcommon (although the docs aren't super clear on what this actually does...)

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: Agreed, this is way better.

packages/optimistic-oracle/index.js
```diff
@@ -143,6 +145,8 @@ async function Poll(callback) {
       optimisticOracleProposerConfig: process.env.OPTIMISTIC_ORACLE_PROPOSER_CONFIG
         ? JSON.parse(process.env.OPTIMISTIC_ORACLE_PROPOSER_CONFIG)
         : {},
+      // Type of "Oracle" set for this network's Finder, default is "Voting". Other possible types include "SinkOracle".
+      oracleType: process.env.ORACLE_TYPE ? process.env.ORACLE_TYPE : "Voting",
```
mrice32, line 149: Very nice!

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/helpers/allowances.js
nicholaspai: PTAL as its a bit of sweeping change across different packages

PR #3036, opened 2021-05-27
feat(Contract-ForDifference): Add more unit tests and add LinearCFDLib tests

chrismaree:
**Motivation**

Additional unit tests added for CFD contract and linear CFD unit tests. This PR also moves around a whole bunch of code by moving the previous FinancialTemplateLibraries into their own directory and creating a new `contract-for-diffrence-libraries` to keep things somewhat seperate.


packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -119,15 +119,17 @@ contract ContractForDifference is Testable, Lockable {
     ) Testable(_timerAddress) {
         finder = _finderAddress;
         require(_expirationTimestamp > getCurrentTime());
-        require(_getIdentifierWhitelist().isIdentifierSupported(_priceIdentifier));
-        require(address(_financialProductLibraryAddress) != address(0));
+        require(_getIdentifierWhitelist().isIdentifierSupported(_priceIdentifier), "Identifier not registered");
+        require(address(_getOptimisticOracle()) != address(0), "Invalid finder");
```
nicholaspai, line 123: Should the comment be "Invalid OptimisticOracle"?

packages/core/test/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/LinearContractForDiffrenceFinancialProductLibrary.js
```diff
@@ -0,0 +1,99 @@
+const { didContractThrow, ZERO_ADDRESS } = require("@uma/common");
```
nicholaspai, line 1: nit: This file name is mispelled, should be "Difference" not "DIffrence"

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
```diff
@@ -15,19 +15,22 @@ contract LinearContractForDiffrenceFinancialProductLibrary is ContractForDiffere
...
+        require(
+            ExpiringContractInterface(contractForDifferenceAddress).expirationTimestamp() != 0,
+            "Invalid CFD address"
+        );
         require(upperBound > lowerBound, "Invalid bounds");
 
         LinearContractForDifferenceParameters memory params =
             contractForDifferenceParameters[contractForDifferenceAddress];
-        require(params.upperBound != 0 || params.lowerBound != 0, "Parameters already set");
+        require(params.upperBound == 0 && params.lowerBound == 0, "Parameters already set");
```
mrice32, line 33: Ha, I'm so dumb. This is why testing is important :)

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -147,8 +149,8 @@ contract ContractForDifference is Testable, Lockable {
 
         collateralToken.safeTransferFrom(msg.sender, address(this), collateralUsed);
 
-        require(longToken.mint(msg.sender, tokensToCreate));
-        require(shortToken.mint(msg.sender, tokensToCreate));
+        require(longToken.mint(msg.sender, tokensToCreate), "Minting long failed");
+        require(shortToken.mint(msg.sender, tokensToCreate), "Minting short failed");
```
mrice32, line 153: Do we know if require messages add gas to the _execution_? If so, I'd be in favor of dropping these given that they are _so_ unlikely to happen, and we should be able to identify the issue via a trace if it happened on mainnet.

They are unlikely because:
- The ERC20 contract we're using reverts on failure and always returns true. So we'd need to pass in a different ERC20.
- Assuming the contract has the right permissions, minting failure should be nearly impossible.


I could even see us dropping the require entirely due to the fact that we know the implementation of the ERC20 being passed in (and that it always returns true).

Thoughts?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -198,8 +200,8 @@ contract ContractForDifference is Testable, Lockable {
             contractState = ContractState.ExpiredPriceReceived;
         }
 
-        require(longToken.burnFrom(msg.sender, longTokensToRedeem));
-        require(shortToken.burnFrom(msg.sender, shortTokensToRedeem));
+        require(longToken.burnFrom(msg.sender, longTokensToRedeem), "Redeeming long failed");
+        require(shortToken.burnFrom(msg.sender, shortTokensToRedeem), "Redeeming short failed");
```
mrice32, line 204: Similar point to the above. This will revert elsewhere, so is there a point to having the message (or even the require itself)?

mrice32 (commented):
Looks awesome!!! Some minor comments on the contract changes. Moving on to reviewing the tests.

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/LinearContractForDiffrenceFinancialProductLibrary.sol
```diff
@@ -4,7 +4,7 @@ pragma solidity ^0.8.0;
 import "@openzeppelin/contracts/utils/math/SignedSafeMath.sol";
 
 import "./ContractForDifferenceFinancialProductLibrary.sol";
-import "../../../common/implementation/Lockable.sol";
+import "../../../../common/implementation/Lockable.sol";
 
 contract LinearContractForDiffrenceFinancialProductLibrary is ContractForDifferenceFinancialProductLibrary, Lockable {
```
nicholaspai, line 9: Nit: you're going to hate this, but you mispelled `Difference` in the contract name....

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
chrismaree: nope. this tests that the _finder_ is set correctly as the call `_getOptimisticOracle` queries the `finder` for the `OracleInterfaces.OptimisticOracle`. This call will error if the finder is not set correctly.

packages/core/test/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/LinearContractForDiffrenceFinancialProductLibrary.js
chrismaree: thank you spell check nick 🤝

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
chrismaree: So I tested this out in transaction execution:
- With requires + messages `gasUsed` was `170563`
- With requires and no messages `gasUsed` was `170563`
- no requires at all `gasUsed` was `170539`

So net difference of 24 gas. a fraction of a fraction of the total transaction. I agree with you though that we can remove them entirely as they will revert upstream.

I've done this through and validates it works just the same via the existing unit tests.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
chrismaree: removed :)

chrismaree:
Options contracts and tests have been added to a follow on PR that will be ready for review as soon as this one is merged in: https://github.com/UMAprotocol/protocol/pull/3040

mrice32 (approved):

mrice32 (approved):
LGTM!!

PR #3037, opened 2021-05-27
improve(beacon-oracle): Audit Response for L2 infrastructure contracts

nicholaspai:
**Motivation**

Audit response to Pull Requests #2903 and #3013

These PRs use Chainbridge to send messages between different EVM-compatible chains. In particular, DVM price requests are passed from another chain to the Ethereum mainnet, where the standard resolution mechanism can occur. Once resolved, the price is pushed back to the "sink" chain where it can be utilized. This allows UMA financial contracts to be deployed in multiple different chains but still use the Ethereum mainnet DVM to resolve prices as required.

**Summary**

Contract changes:
- On both chains, the [`PriceRequestAdded` event](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/BeaconOracle.sol#L81) includes the `msg.sender` as the `requester`. However, in the `SourceOracle`, [this will be the `GenericHandler` contract](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/SourceOracle.sol#L88), not the address that requested the price. Similarly, the `SinkOracle` will emit the `GenericHandler` as the `pusher` in the `PushedPrice` event. In both cases, consider passing the original sender over the bridge so it can be emitted on the destination chain.
- Solution: remove `requester` and `pusher` param from `PriceRequestAdded` and `PushedPrice` events, respectively. `pusher` is not useful information and `requester` is not added to the event of the same name in the ` Voting` contract. Ultimately, `requester` is not super useful information, and will be captured by the `OptimisticOracle` assuming that the OO is the contract requesting prices from the `SinkOracle`.

Improvements to comments:
- The [`_requestPrice` function comment](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/BeaconOracle.sol#L69) says the function will revert if it's already been requested. In fact, it returns without doing anything.
- The `_formatMetadata` function comments on [the `SinkOracle`](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/SinkOracle.sol#L151-L152) and [the `SourceOracle`](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/SourceOracle.sol#L113-L114) claim the `uint256` length field takes up 64 bytes instead of 32.
- `SourceOracle` and `SinkOracle` are missing natspec comments.

The following comments pertain to the BeaconOracle:
- [Line 11](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/BeaconOracle.sol#L11) says "respectivly".
- [Line 27](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/BeaconOracle.sol#L27) excludes `chainId` from the list of encoded values
- [Line 68](https://github.com/UMAprotocol/protocol/blob/2814e8021294adb2578649ed57fb066e86bb5b73/packages/core/contracts/chainbridge/BeaconOracle.sol#L68) also excludes `chainId` and refers to the whole collection of values as a "pair"

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/chainbridge/BeaconOracle.sol
+++ b/packages/core/contracts/chainbridge/BeaconOracle.sol
@@ -31,14 +31,12 @@ abstract contract BeaconOracle {
     FinderInterface public finder;
 
     event PriceRequestAdded(
-        address indexed requester,
         uint8 indexed chainID,
         bytes32 indexed identifier,
         uint256 time,
         bytes ancillaryData
     );
     event PushedPrice(
-        address indexed pusher,
         uint8 indexed chainID,
         bytes32 indexed identifier,
         uint256 time,
@@ -78,7 +76,7 @@ abstract contract BeaconOracle {
         Price storage lookup = prices[priceRequestId];
         if (lookup.state == RequestState.NeverRequested) {
             lookup.state = RequestState.PendingRequest;
-            emit PriceRequestAdded(msg.sender, chainID, identifier, time, ancillaryData);
+            emit PriceRequestAdded(chainID, identifier, time, ancillaryData);
         }
     }
 
@@ -114,7 +112,7 @@ abstract contract BeaconOracle {
         require(lookup.state == RequestState.Requested, "Price request is not currently pending");
         lookup.price = price;
         lookup.state = RequestState.PendingResolve;
-        emit PushedPrice(msg.sender, chainID, identifier, time, ancillaryData, lookup.price);
+        emit PushedPrice(chainID, identifier, time, ancillaryData, lookup.price);
     }
 
     function _finalizePublish(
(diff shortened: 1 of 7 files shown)


mrice32 (approved):
LGTM!

PR #3040, opened 2021-05-28
feat(CFD): Add call options financial template library

chrismaree:
**Motivation**

A simple contract type built on CFDs is a basic call option. This PR adds the required library and unit tests to implement call options on top of the CFD base implementation.


packages/core/test/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/CallOptionContractForDifferenceFinancialProductLibrary.js
```diff
@@ -0,0 +1,89 @@
...
+      assert.equal(expiraryTokensForCollateral.toString(), toWei("0"));
+    });
+    it("Higher than strike correct value", async () => {
+      const expiraryTokensForCollateral = await callOptionCFDFPL.computeExpiraryTokensForCollateral.call(toWei("500"), {
+        from: expiringContractMock.address,
+      });
+      assert.equal(expiraryTokensForCollateral.toString(), toWei("0.2"));
+    });
+    it("Arbitary expirary price above strike should return correctly", async () => {
+      for (const price of [toWei("500"), toWei("600"), toWei("1000"), toWei("1500"), toWei("2000")]) {
```
mrice32, line 69: tiny nit: rather than replicating the computation off-chain, it might be preferable to have this be an array of objects with trivially verifiable output values:
`[{ input: toWei("500"), expectedOutput: toWei("1234") },...]`. If it's hard to create trivially verifiable values, keeping it the way it is could make sense. Totally up to you -- just wanted to mention it.

mrice32 (approved):
LGTM!

PR #3041, opened 2021-05-28
fix(bot-stratergy-runner): fix broken default setting

chrismaree:
**Motivation**

When running in cloud run bot settings sometimes need to be set to `0`. For example `pollingDelay` must be set to `0` to run in serverless mode. The way the code was set up before makes this config impossible due to how defaults are set. This PR addresses this.

### DIFF
--- a/packages/bot-strategy-runner/src/index.ts
+++ b/packages/bot-strategy-runner/src/index.ts
@@ -177,9 +177,9 @@ function _reduceLog(
 // Apply defaults to the strategyRunnerConfig.
 function _setConfigDefaults(config: strategyRunnerConfig) {
   config.botNetwork = config.botNetwork ? config.botNetwork : defaultNetwork;
-  config.strategyTimeout = config.strategyTimeout ? config.strategyTimeout : defaultStrategyTimeout;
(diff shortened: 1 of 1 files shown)


packages/bot-strategy-runner/src/index.ts
```diff
@@ -177,9 +177,9 @@ function _reduceLog(
 // Apply defaults to the strategyRunnerConfig.
 function _setConfigDefaults(config: strategyRunnerConfig) {
   config.botNetwork = config.botNetwork ? config.botNetwork : defaultNetwork;
-  config.strategyTimeout = config.strategyTimeout ? config.strategyTimeout : defaultStrategyTimeout;
-  config.botConcurrency = config.botConcurrency ? config.botConcurrency : defaultBotConcurrency;
-  config.pollingDelay = config.pollingDelay ? config.pollingDelay : defaultPollingDelay;
+  config.strategyTimeout = config.strategyTimeout != undefined ? config.strategyTimeout : defaultStrategyTimeout;
```
Gamaranto, line 180: minuscole nit, you can use `??` instead of the ternary here

PR #3046, opened 2021-06-01
feat: Add hardhat task to migrate identifier whitelist to new contract

nicholaspai:
**Motivation**

Helpful script to query all whitelisted identifiers on one contract and whitelist the entire list on another contract. Used script to deploy and set up new `IdentifierWhitelist` on Kovan network because existing contract ownership was corrupted.

Also removed two duplicate `deploy` scripts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/hardhat/tasks/identifiers.js
```diff
@@ -24,3 +24,68 @@ task("whitelist-identifiers", "Whitelist identifiers from JSON file")
...
+    addedIdentifierEvents.forEach((_event) => {
+      isIdentifierSupportedPromises.push(
+        oldWhitelist.methods.isIdentifierSupported(_event.returnValues.identifier).call()
+      );
+    });
+    (await Promise.all(isIdentifierSupportedPromises)).forEach((supported, i) => {
+      if (supported) {
+        identifiersToWhitelist.push(addedIdentifierEvents[i].returnValues.identifier);
+      }
+    });
```
mrice32, line 53: nit: I think this could be simplified a bit.

The first part seems more suited for a map `addedIdentifierevents.map(...` (so you don't have to declare and append).

You could also have the first map return identifier or null rather than true/false and then just filter out the nulls, which might be a bit simpler.

packages/common/src/hardhat/tasks/identifiers.js
```diff
@@ -24,3 +24,68 @@ task("whitelist-identifiers", "Whitelist identifiers from JSON file")
...
+
+    if (to) {
+      const newWhitelist = new web3.eth.Contract(IdentifierWhitelist.abi, to);
+      isIdentifierSupportedPromises = [];
+      identifiersToWhitelist.forEach((id) => {
+        isIdentifierSupportedPromises.push(newWhitelist.methods.isIdentifierSupported(id).call());
+      });
+
+      const isIdentifierSupportedResults = await Promise.all(isIdentifierSupportedPromises);
```
mrice32, line 70: nit: I think this can be simplified
```suggestion
      const isIdentifierSupportedResults = await Promise.all(identifiersToWhitelist.map(id => newWhitelist.methods.isIdentifierSupported(id).call()));
```

mrice32 (approved):
LGTM!

chrismaree (approved):

packages/common/src/hardhat/tasks/identifiers.js
nicholaspai: Done nice refactor

packages/common/src/hardhat/tasks/identifiers.js
nicholaspai: done

mrice32 (approved):
LGTM!

PR #3053, opened 2021-06-02
feat(CFD): Add range bond financial template library

chrismaree:
**Motivation**

Add range bonds Contract for difference financial product library and unit tests.
A range bond is defined as = Yield Dollar - Put Option + Call option.

 Numerically this is found using:        
- `N` = Notional of bond
- `P` = price of token
- `T` = number of tokens
- `R1` = Low Price Range
- `R2` = High Price Range
- long token redemption: `T = min(N/P,N/R1) + max((N/R2*(P-R2))/P,0)`

As the financial product library returns a number between 0 and 1 for CFD payouts, we can take the long token redemption value `T` and divide it by the CFDs `collateralPerPair` to find a number between 0 and 1 that represents the long/short payout `expiryPercentLong`.

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
@@ -70,8 +70,7 @@ contract RangeBondContractForDifferenceFinancialProductLibrary is
         // multiplications. However this introduces rounding issues. By keeping it in this form this is avoided.
         // NOTE: the ternary is used over Math.max for the second term to avoid negative numbers.
 
-        uint256 positiveExpiraryPrice = 0;
-        if (expiryPrice > 0) positiveExpiraryPrice = uint256(expiryPrice);
+        uint256 positiveExpiraryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
 
         FixedPoint.Unsigned memory longTokensRedeemed =
             FixedPoint
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
@@ -45,6 +45,11 @@ contract RangeBondContractForDifferenceFinancialProductLibrary is
             "Parameters already set"
         );
 
+        uint256 cfdCollateralPerPair = ContractForDifferenceInterface(contractForDifferenceAddress).collateralPerPair();
+
+        require(FixedPoint.Unsigned(cfdCollateralPerPair).mul(lowPriceRange).isEqual(bondNotional), "Bad params");
+        require(FixedPoint.Unsigned(bondNotional).div(highPriceRange).isLessThan(cfdCollateralPerPair), "Bad params");
+
         contractForDifferenceParameters[contractForDifferenceAddress] = RangeBondContractForDifferenceParameters({
             bondNotional: bondNotional,
             lowPriceRange: lowPriceRange,
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
@@ -19,7 +19,6 @@ contract RangeBondContractForDifferenceFinancialProductLibrary is
     using SignedSafeMath for int256;
 
     struct RangeBondContractForDifferenceParameters {
-        uint256 bondNotional;
         uint256 highPriceRange;
         uint256 lowPriceRange;
     }
@@ -28,7 +27,6 @@ contract RangeBondContractForDifferenceFinancialProductLibrary is
 
     function setContractForDifferenceParameters(
         address contractForDifferenceAddress,
-        uint256 bondNotional,
         uint256 highPriceRange,
         uint256 lowPriceRange
     ) public nonReentrant() {
@@ -40,20 +38,10 @@ contract RangeBondContractForDifferenceFinancialProductLibrary is
 
         RangeBondContractForDifferenceParameters memory params =
             contractForDifferenceParameters[contractForDifferenceAddress];
-        require(
-            params.bondNotional == 0 && params.lowPriceRange == 0 && params.highPriceRange == 0,
-            "Parameters already set"
-        );
-
-        uint256 cfdCollateralPerPair = ContractForDifferenceInterface(contractForDifferenceAddress).collateralPerPair();
-
-        require(FixedPoint.Unsigned(cfdCollateralPerPair).mul(lowPriceRange).isEqual(bondNotional), "Bad params");
-        require(FixedPoint.Unsigned(bondNotional).div(highPriceRange).isLessThan(cfdCollateralPerPair), "Bad params");
 
         contractForDifferenceParameters[contractForDifferenceAddress] = RangeBondContractForDifferenceParameters({
-            bondNotional: bondNotional,
-            lowPriceRange: lowPriceRange,
-            highPriceRange: highPriceRange
+            highPriceRange: highPriceRange,
+            lowPriceRange: lowPriceRange
         });
     }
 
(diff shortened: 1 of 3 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,100 @@
...
+        // This represents the value of the long token(range bond holder). This function's method must return a value
+        //  between 0 and 1 to be used as a collateralPerPair that allocates collateral between the short and long tokens.
+        // We can use the value of the long token to compute the relative distribution between long and short CFD tokens
+        // by simply computing longTokenRedeemed using equation above devided by the collateralPerPair of the CFD.
+        // NOTE: the equation's second term could be simplified slightly alebraically to have fewer devided and
+        // multiplications. However this introduces rounding issues. By keeping it in this form this is avoided.
+        // NOTE: the ternary is used over Math.max for the second term to avoid negative numbers.
+
+        uint256 positiveExpiraryPrice = 0;
+        if (expiryPrice > 0) positiveExpiraryPrice = uint256(expiryPrice);
```
mrice32, line 74: A ternary might make this a little simpler.
```suggestion
        uint256 positiveExpiraryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
```

mrice32 (commented):
Asked some questions offline -- will have a more substantive review once I get a better intuition for this concept!

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,99 @@
...
+        // This represents the value of the long token(range bond holder). This function's method must return a value
+        //  between 0 and 1 to be used as a collateralPerPair that allocates collateral between the short and long tokens.
+        // We can use the value of the long token to compute the relative distribution between long and short CFD tokens
+        // by simply computing longTokenRedeemed using equation above devided by the collateralPerPair of the CFD.
+        // NOTE: the equation's second term could be simplified slightly alebraically to have fewer devided and
+        // multiplications. However this introduces rounding issues. By keeping it in this form this is avoided.
+        // NOTE: the ternary is used over Math.max for the second term to avoid negative numbers.
+
+        uint256 positiveExpiraryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+
```
mrice32, line 74: In light of our offline conversation, here's an alternative implementation of the math (but please let me know if you disagree!).

This really depends on whether our prices are in the form USDCOLLATERAL or COLLATERALUSD. If we assume the more natural COLLATERALUSD, we have to invert everything (in this proposed implementation).

```
FixedPoint.Unsigned memory maxInvertedPrice = FixedPoint.fromUnscaledUint(1).div(FixedPoint.Unsigned(params.lowPriceRange));

FixedPoint.Unsigned memory minInvertedPrice = FixedPoint.fromUnscaledUint(1).div(FixedPoint.Unsigned(params.highPriceRange));

FixedPoint.Unsigned memory invertedPrice = FixedPoint.min(FixedPoint.max(FixedPoint.fromUnscaledUint(1).div(FixedPoint.Unsigned(expiryPrice)), minInvertedPrice), maxInvertedPrice);

return invertedPrice.div(maxInvertedPrice).rawValue;
```

Not sure if this is simpler, but I _think_ this would also work, and a bonus is that it shouldn't require the contract to know anything about the notional.

Thoughts?

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,99 @@
...
+        RangeBondContractForDifferenceParameters memory params =
+            contractForDifferenceParameters[contractForDifferenceAddress];
+        require(
+            params.bondNotional == 0 && params.lowPriceRange == 0 && params.highPriceRange == 0,
+            "Parameters already set"
+        );
+
+        contractForDifferenceParameters[contractForDifferenceAddress] = RangeBondContractForDifferenceParameters({
+            bondNotional: bondNotional,
```
nicholaspai, line 49: Does this mean we're ok if the bondNotional is set to 0?

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
nicholaspai: How did you strip out `N` from `min(N/P,N/R1) + max((N/R2*(P-R2))/P,0)`? Or did you convert `min(N/P,N/R1) + max((N/R2*(P-R2))/P,0)` to another formula?

chrismaree: so this is a great meta question and comment. In general, I'm of the opinion that we should try to mimic the "financial" formula as much as possible without introducing abstraction.

Regarding your particular equation, I'm pretty sure this does _not_ work. you _do_ need the notional. From the starting base equation from the Range bond doc of `T = min(N/P,N/R1) + max((N/R2*(P-R2))/P,0)` we can simplify to the equation you posted on slack as `T=min(maxTKN, max(minTKN, N * USDTKN))` which workers fine with an output `T` looking like this:
image


The formula you posed in your code block does not work correctly (assuming I correctly converted it to normal mathematical notation). it plots like this:

image

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
chrismaree: previously this would be possible. I've added some additional checks that would make this possible.

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
chrismaree: please see the comment below on the update to this.

chrismaree:
After much discussion offline we've opted for a simplified version of this equation that removes the need for the Notional term `N` from the equation. This can be seen in the screenshot below of [this](https://www.wolframalpha.com/input/?i=plot+y+%3D+%28min%28max%281%2F50%2C1%2Fx%29%2C1%2F10%29%29%2F%281%2F10%29++between+x+%3D+0+and+x+%3D+100) wolfram Alpha computation.

image

@mrice32 & @nicholaspai please take a look and let me know what you think!

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,80 @@
...
+        RangeBondContractForDifferenceParameters memory params = contractForDifferenceParameters[msg.sender];
+
+        // A range bond is defined as = Yield Dollar - Put Option + Call option. Numerically this is found using:
+        // N = Notional of bond (100)
+        // P = price of token
+        // T = number of tokens
+        // R1 = Low Price Range
+        // R2 = High Price Range
+        // T = min(N/P,N/R1) + max((N/R2*(P-R2))/P,0)
+        // T = [min(max(1/R2,1/P),1/R1)]/R1 [simplified form without N]
```
nicholaspai, line 58: is it possible to link the wolfram alpha function visualizer with the function set with query params? Would be useful to visualize

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
```diff
@@ -0,0 +1,80 @@
...
+        // This represents the value of the long token(range bond holder). This function's method must return a value
+        //  between 0 and 1 to be used as a collateralPerPair that allocates collateral between the short and long tokens.
+        // We can use the value of the long token to compute the relative distribution between long and short CFD tokens
+        // by simply computing longTokenRedeemed using equation above divided by the collateralPerPair of the CFD.
+
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
```
nicholaspai, line 64: Is there not a max function you can use?

mrice32: OZ has a Math library for min/max, but I think it only supports uints

packages/core/test/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.js
```diff
@@ -0,0 +1,156 @@
...
+
+        const term2 = BN.max(
+          toBN(bondNotional)
+            .div(toBN(highPriceRange))
+            .mul(fixedPointAdjustment)
+            .mul(toBN(price).sub(toBN(highPriceRange)))
+            .div(toBN(price))
+            .addn(1),
+          toBN("0")
+        );
```
mrice32, line 145: This is awesome, as it verifies that the implementation matches the spec equation. +1

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/common/financial-product-libraries/contract-for-difference-libraries/RangeBondContractForDifferenceFinancialProductLibrary.sol
mrice32: +1

chrismaree: do you mean link it in the comments of the smart contract or within the PR? the comment I posted above has wolframalpha plotting this function. we could shorten a URL and put it in the comments but I don't think it's really that useful.

PR #3054, opened 2021-06-02
feat(polygon-fx-tunnel): Add Oracle tunnel integrating with Polygon hosted Arbitrary Message Bridge

nicholaspai:
**Motivation**

We want to bridge price requests in such a way that we are not relying on any more trusted actors than all Polygon users inherently do. In other words, we are OK trusting the Polygon validator set to bridge arbitrary oracle messages and this is preferable to trusting the cohort of Relayers that are required to run the Chainbridge-esque messaging system.

This PR introduces contracts that extend the required `Fx Portal` contracts that sync state between L1 and L2 the same way that Polygon PoS works.

**Summary**

Reviewers guide:
- There is a LOT more code diff than needs to be reviewed.
- `contracts/chainbridge` contracts were moved to `contracts/l2/chainbridge` with no modifications
- `contracts/l2/lib` and `contracts/l2/tunnel` are copied from the [fx-portal](https://github.com/jdkanani/fx-portal) repo directly. The minor modifications are a comment at the top giving attribution to the source repo, and a change to Solidity v8, which requires some small tweaks. I'll highlight these changes in the comments.
- The contracts to review are `OracleBaseTunnel`, `OracleRootTunnel`, and `OracleChildTunnel`. `OracleBaseTunnel` is a simplified version of `contracts/chainbridge/BeaconOracle` without the `chainId` parameter. `OracleChildTunnel` and `OracleRootTunnel` are expected to be deployed on Polygon and Ethereum respectively. Following [this tutorial](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer/), they implement `FxBaseChildTunnel` and `FxBaseRootTunnel`.
- Deployment scripts and `hardhat task`s are added to aid with war games testing on Mumbai and Goerli.

**Details**

Follow up PR will add unit tests for the `OracleX` contracts.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/l2/polygon/OracleChildTunnel.sol
```diff
@@ -0,0 +1,109 @@
...
+        if (lookup.state != RequestState.NeverRequested) {
+            // Clients expect that `requestPrice` does not revert if a price is already requested, so return gracefully.
+            return;
+        } else {
+            _requestPrice(identifier, time, ancillaryData);
+
+            // Initiate cross-chain price request:
+            // TODO: Can we pack more information into this request? We could try to check if the requester is an
+            // OptimisticOracle and pull price request metadata from that?
+            _sendMessageToRoot(abi.encode(identifier, time, ancillaryData));
```
nicholaspai, line 54: Reinis brought up in the chainbridge UMIP review that we should ideally publish information such as fees and original requester address to Mainnet to aid DVM voters. However, there is no obvious way to do this because the `OptimisticOracle` expects to call `requestPrice(bytes32,uint25,bytes)` according to the `OracleInterface`, so we can't pack more information into this call without creating a new interface and updating the `OptimisticOracle`. 

We _could_ assume that the requester was an OO and query data from the caller of this `requestPrice` method. However, see the note in `OracleRootTunnel` how `_processMessageFromChild` will NOT emit an event on Ethereum because that internal function is expected to be called via a System call that does not send a transaction.

packages/common/src/hardhat/tasks/fxTunnel.js
```diff
@@ -0,0 +1,26 @@
+const { task, types } = require("hardhat/config");
+const MaticJs = require("@maticnetwork/maticjs")
+require("dotenv").config();
+
+// In order to receive a message on Ethereum from Polygon, `receiveMessage` must be called on the Root Tunnel contract
+// with a proof derived from the Polygon transaction hash that was checkpointed to Mainnet. 
+task("root-chain-manager-proof", "Generate proof needed to receive data from root chain manager")
```
nicholaspai, line 7: The logic in this task will eventually have to be encapsulated in some cron job/polling bot that should listen for `NewHeaderBlock` on the `RootChainProxy` contract deployed on Ethereum. (example deployment on Goerli [here](https://goerli.etherscan.io/address/0x2890ba17efe978480615e330ecb65333b880928e#code)). 

On each of these events, the bot should derive the proof for each transaction hash on Polygon that emitted a `MessageSent` for a contract that it is monitoring. For example, [this transaction](https://explorer-mumbai.maticvigil.com/tx/0xbb1a9589ff82229230a78ac4af0e029c5c422f3730658e3299adcf40205e1504/internal-transactions) on Mumbai emitted a `MessageSent` event and the bot should ideally be aware of this, and be ready to derive a proof for the transaction hash `0xbb1a9589ff82229230a78ac4af0e029c5c422f3730658e3299adcf40205e1504` after each `NewHeaderBlock` event is emitted from Ethereum.

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
```diff
@@ -0,0 +1,82 @@
...
+     */
+    function _publishPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData,
+        int256 price
+    ) internal {
+        bytes32 priceRequestId = _encodePriceRequest(identifier, time, ancillaryData);
+        Price storage lookup = prices[priceRequestId];
+        if (lookup.state == RequestState.Requested) {
```
nicholaspai, line 64: Note: Unlike in BeaconOracle, this _publishPrice method does NOT revert if called more than once. This is because we are not concerned about spam cross-chain messages.

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
```diff
@@ -0,0 +1,106 @@
...
+     * @param identifier Identifier of price request.
+     * @param time Timestamp of price request.
+     * @param ancillaryData extra data of price request.
+     */
+    function requestPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public override onlyRegisteredContract() {
+        // This implementation allows duplicate price requests to emit duplicate MessageSent events via
```
nicholaspai, line 44: Important to read this design decision. This is based on feedback from Polygon team and is less of a concern than `publishPrice` because of the `onlyRegisteredContract` modifier.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
```diff
@@ -0,0 +1,63 @@
...
+     * @param ancillaryData extra data of price request to resolve.
+     */
+    function publishPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public {
+        require(_getOracle().hasPrice(identifier, time, ancillaryData), "DVM has not resolved price");
+        int256 price = _getOracle().getPrice(identifier, time, ancillaryData);
+        // This implementation allows duplicate MessageSent events via _sendMessageToRoot. The child tunnel on the
```
nicholaspai, line 33: Important to read this design decision. This is based on feedback from Polygon team. The benefit of allowing duplicate calls to `_sendMessageToChild` is ensuring that we can resubmit any transactions that the bridge might miss. The downside is that anyone can call this method to spam the Polygon hosted tunnel. Thoughts?

nicholaspai: Perhaps we can impose a rate-limit within the contract.

packages/core/contracts/l2/polygon/lib/RLPReader.sol
```diff
@@ -0,0 +1,245 @@
...
+        copy(item.memPtr, ptr, item.len);
+        return result;
+    }
+
+    function toAddress(RLPItem memory item) internal pure returns (address) {
+        require(!isList(item), "RLPReader: DECODING_LIST_AS_ADDRESS");
+        // 1 byte for the length prefix
+        require(item.len == 21, "RLPReader: INVALID_ADDRESS_LENGTH");
+
+        return address(uint160(toUint(item)));
```
nicholaspai, line 84: This line was changed from `address(toUint(item))` due to this Solidity v8 breaking change:

> Let us use the notation T(S) to denote the explicit conversion T(x), where, T and S are types, and x is any arbitrary variable of type S. An example of such a disallowed conversion would be uint16(int8) since it changes both width (8 bits to 16 bits) and sign (signed integer to unsigned integer). In order to do the conversion, one has to go through an intermediate type. In the previous example, this would be uint16(uint8(int8)) or uint16(int16(int8)). Note that the two ways to convert will produce different results e.g., for -1. The following are some examples of conversions that are disallowed by this rule.

> address(uint) and uint(address): converting both type-category and width. Replace this by address(uint160(uint)) and uint(uint160(address)) respectively.

packages/core/contracts/l2/polygon/lib/RLPReader.sol
```diff
@@ -0,0 +1,245 @@
...
+     * @param src Pointer to source
+     * @param dest Pointer to destination
+     * @param len Amount of memory to copy from the source
+     */
+    function copy(
+        uint256 src,
+        uint256 dest,
+        uint256 len
+    ) private pure {
+        unchecked {
```
nicholaspai, line 223: This `unchecked {}` syntax was added after war games testing the contract and seeing that a call to `copy` was reverting within a `receiveMessage` EOA submitted-transaction. Not 100% sure on what the root problem is but h/t @mrice32 for the suggested solution:

> Arithmetic operations revert on underflow and overflow. You can use unchecked { ... } to use the previous wrapping behaviour.
> If you rely on wrapping arithmetic, surround each operation with unchecked { ... }.

[Source](https://docs.soliditylang.org/en/v0.8.4/080-breaking-changes.html#silent-changes-of-the-semantics).

packages/core/contracts/l2/polygon/lib/RLPReader.sol
nicholaspai: [Source](https://docs.soliditylang.org/en/v0.8.4/080-breaking-changes.html#silent-changes-of-the-semantics).

packages/core/networks/5.json
```diff
@@ -0,0 +1,19 @@
+[
+    {
+      "contractName": "OracleRootTunnel",
```
nicholaspai, line 3: Contracts I used for war games testing

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
chrismaree: would `ancillaryData` not help a bit with this? it would be possible to append more info to it without affecting the original intention of the data, no?

packages/common/src/hardhat/tasks/fxTunnel.js
chrismaree: does encapsulating this logic within a task make it more difficult for us to re-use later on when dealing with this within a bot? how would you propose this logic be re-use when we get to building out this bot? would it be better to put it in a JS script that can be re-used more easily? wdyt?

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
```diff
@@ -0,0 +1,82 @@
...
+    function _requestPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) internal {
+        bytes32 priceRequestId = _encodePriceRequest(identifier, time, ancillaryData);
+        Price storage lookup = prices[priceRequestId];
+        if (lookup.state == RequestState.NeverRequested) {
+            lookup.state = RequestState.Requested;
+            emit PriceRequestAdded(identifier, time, ancillaryData);
```
chrismaree, line 49: nit: it might be useful to also emit the `priceRequestId` here so the re-hashing of this does not need to happen offchain.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
chrismaree: well seeing the price that is pulled here comes directly from the oracle there is not a huge amount of risk IMHO. what would you be able to do by spamming this? as the L2 has no rate limiting this should not be a problem on L1.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
```diff
@@ -0,0 +1,63 @@
...
+        // sidechain will not have a problem handling duplicate price resolutions (it will just ignore them). This is
+        // potentially a fallback in case the automatic state sync to the sidechain is missing the `publishPrice`
+        // transaction for some reason. There is little risk in duplicating MessageSent emissions because the sidechain
+        // bridge does not impose any rate-limiting.
+        _publishPrice(identifier, time, ancillaryData, price);
+        // Initiate cross-chain price request:
+        // TODO: Should we pack more information into this request?
+        _sendMessageToChild(abi.encode(identifier, time, ancillaryData, price));
+    }
```
chrismaree, line 42: are there any events you would like to emit over and above that produced by `_publishPrice`?

packages/common/src/hardhat/tasks/fxTunnel.js
chrismaree: that sounds good. how do you see this task easily translating to a bot implementation? would the bot call the task? would it be a re-write into a more long-form bot-type implementation?

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
```diff
@@ -0,0 +1,106 @@
...
+     * that will be included in regular checkpoint of all sidechain transactions to mainnet.
+     * @param identifier Identifier of price request.
+     * @param time Timestamp of price request.
+     * @param ancillaryData extra data of price request.
+     */
+    function requestPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public override onlyRegisteredContract() {
```
chrismaree, line 43: are there any additional events that would be useful here over and above what `requestPrice` emits?

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
chrismaree, line 49: would it be useful to emit the `priceRequestId` as well? you can clearly compute it off-chain but there is min cost to emitting it here as well don't you think?

packages/core/contracts/l2/polygon/lib/Merkle.sol
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: MIT
```
chrismaree, line 1: seeing this code was pulled from another repo, do you have any information on if it was audited in their repo? did you need to change the code at all?

packages/core/contracts/l2/polygon/lib/RLPReader.sol
chrismaree: prehaps add a comment to make that clear that this is the only thing that changed in the contract

packages/core/contracts/l2/polygon/lib/RLPReader.sol
chrismaree: yup; I had a few similar cases in the uni3 broker when dealing with any numerical operation that could wrap within assembly.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
chrismaree: If the L2 does not have any rate limiting and this does not call into and UMA DVM contracts other than the `hasPrice` and `viewPrice` view methods then I don't really see any downside in this.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
```diff
@@ -0,0 +1,63 @@
...
+     * @param time Timestamp of price request to resolve.
+     * @param ancillaryData extra data of price request to resolve.
+     */
+    function publishPrice(
+        bytes32 identifier,
+        uint256 time,
+        bytes memory ancillaryData
+    ) public {
+        require(_getOracle().hasPrice(identifier, time, ancillaryData), "DVM has not resolved price");
+        int256 price = _getOracle().getPrice(identifier, time, ancillaryData);
```
chrismaree, line 32: would `getPrice` not revert if there was not a price meaning the `require` is redundant?

chrismaree (approved):
This is pretty awesome. It's super clean and cool to see how this is possible with the polygon base implementation.

Great work @nicholaspai!

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
nicholaspai: > would `ancillaryData` not help a bit with this? it would be possible to append more info to it without affecting the original intention of the data, no?

Really good point. We'd have to modify the OptimisticOracle implementation that we deploy to Polygon to stamp additional information like `finalFee`, `bond`, `requester` etc. But that's ok because we do need to deploy the OO again to Polygon!

We could create a new `OptimisticOracleL2` contract for this purpose. Thoughts?

packages/common/src/hardhat/tasks/fxTunnel.js
nicholaspai: I think I'd rewrite this task into a long-form bot implementation, likely deleting this task or making this task import a common script (that the bot also imports). For now I just kept this as a task for convenience.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
nicholaspai: You would just be able to spam the Polygon-hosted relayers, which doesn't seem like a big risk for us.

packages/core/contracts/l2/polygon/lib/RLPReader.sol
nicholaspai: Adding to top of RLPReader.sol

packages/core/contracts/l2/polygon/lib/RLPReader.sol
nicholaspai: Oh really @chrismaree did you have to add `unchecked{}` to the code?

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
nicholaspai: I'm good with that. Alternatively we could change _encodePriceRequest to be an external method?

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
nicholaspai: I don't see anything, do you?

packages/core/contracts/l2/polygon/lib/Merkle.sol
nicholaspai: They are undergoing audit right now so this code might have to change...

I have only made changes in RLPReader and I've added comments indicating that.

nicholaspai: I'll also add comments for contracts that I made no changes to.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
nicholaspai: True... I'll remove `hasPrice`

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
nicholaspai: @mrice32 curious what you think of tradeoffs here

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
nicholaspai: Or maybe we just make a follow up PR that adds a lot information to the `OptimisticOracle` stamped ancillary data, and deploy that new version to Polygon?

chrismaree: ye I think that's totally reasonable. when we were thinking through ancillary data the key pattern was this ability to "stamp" data from a number of sets the data is passed through. an L2 bridge is a great example of where this stamping would work really well!

packages/common/src/hardhat/tasks/fxTunnel.js
chrismaree: sgtm

packages/core/contracts/l2/polygon/lib/RLPReader.sol
chrismaree: yup. I had to do that in a few places. here for example:
https://github.com/UMAprotocol/protocol/blob/10381de6bd2a3b35fef8b9ed44c4318ac5b5dfc5/packages/core/contracts/bot-helpers/uniswap-broker/UniswapV3Broker.sol#L575

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
nicholaspai: Also just noticed that the OO does stamp in `requester` address so it could make sense to add extra data

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
nicholaspai: see [here](https://github.com/UMAprotocol/protocol/pull/3054#discussion_r645526603) for my response

packages/core/contracts/l2/polygon/OracleChildTunnel.sol
nicholaspai: @chrismaree @mrice32 here's a change to the `OptimisticOracle` that we should make to help voters with decoding the stamped ancillary data. Additionally, consider what other pieces of data we might want to stamp to the DVM: ([PR](https://github.com/UMAprotocol/protocol/pull/3061))

packages/core/contracts/l2/polygon/lib/RLPReader.sol
```diff
@@ -0,0 +1,251 @@
...
+                assembly {
+                    mstore(dest, mload(src))
+                }
+
+                src += WORD_SIZE;
+                dest += WORD_SIZE;
+            }
+
+            // left over bytes. Mask is used to remove unwanted bytes from the word
+            uint256 mask = 256**(WORD_SIZE - len) - 1;
```
mrice32, line 243: tiny nit: should we try wrapping _only_ this line since we think this is where the breakage is? That way, the rest still has overflow protection?

I'm okay either way -- just curious to hear your thoughts.

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
```diff
@@ -0,0 +1,82 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
mrice32, line 1: A few questions:
- Does polygon export this in a library of some kind that would allow us _not_ to copy it over?
- Do we need to intermingle our code with external code or should we separate it? Since this is becoming more common, could it make sense for us to have an "external" folder that includes code that we didn't write and/or aren't maintaining with READMEs for each section detailing where the code came from and a changelog of any modifications we made to fit our work.
- Does this code need to be audited? Or should we wall this off for the auditors?

packages/core/deploy/017_deploy_oracle_root_tunnel.js
```diff
@@ -0,0 +1,34 @@
...
+    Finder.address,
+  ];
+  await deploy("OracleRootTunnel", {
+    from: deployer,
+    args,
+    log: true,
+  });
+};
+module.exports = func;
+func.tags = ["OracleRootTunnel"];
```
mrice32, line 34: Should we have tags for what we need to deploy for _each_ l2 system?

For instance, we could have tags:
- `l1` for all l1 contracts.
- `l1-polygon` for just the l1 contracts that need to be deployed for polygon.
- `l1-chainbridge` for just the l1 contracts that need to be deployed for chainbridge.
- `l2-polygon` ...
- `l2-chainbridge` ...

Note: we don't have to do this in this PR -- just want to come up with some sort of coherent, but not overly-complex tagging system.

Thoughts?

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
mrice32: The only risk I could see is that they somehow blacklist _our_ contracts because of spam. But my guess is that they won't do that.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
```diff
@@ -0,0 +1,65 @@
...
+        // `getPrice` will revert if there is no price.
+        int256 price = _getOracle().getPrice(identifier, time, ancillaryData);
+        // This implementation allows duplicate MessageSent events via _sendMessageToRoot. The child tunnel on the
+        // sidechain will not have a problem handling duplicate price resolutions (it will just ignore them). This is
+        // potentially a fallback in case the automatic state sync to the sidechain is missing the `publishPrice`
+        // transaction for some reason. There is little risk in duplicating MessageSent emissions because the sidechain
+        // bridge does not impose any rate-limiting.
+        _publishPrice(identifier, time, ancillaryData, price);
+        // Initiate cross-chain price request:
+        // TODO: Should we pack more information into this request?
```
mrice32, line 41: What sort of information were you thinking?

packages/core/contracts/l2/polygon/tunnel/FxBaseRootTunnel.sol
```diff
@@ -0,0 +1,185 @@
+// SPDX-License-Identifier: MIT
```
mrice32, line 1: The architecture here is really interesting. Might be nice to have a README somewhere that explains how the state connection works. It seems that the architecture basically works this way:

- There's a root contract and a child contract.
- Contracts that want to communicate inherit from the root and child and specify one another as the root/child addresses.
- These inheriting contracts have some variant of "sendMessage" and "receiveMessage" to send and receive messages over the bridge.
- They call sendMessage to deliver bytes to the other side.
- They override a "process" method to actually receive and do stuff with the data received.

The way it seems to be set up, you would have each contract that wants to communicate replicate this same setup and effectively be a bridge in itself rather than sending messages to and from a bridge. This is a really convenient architecture for permissioning purposes (despite being a bit bloated) since contracts are initialized in pairs that have a very obvious permissioning structure (they only allow messages to and from each other).

Assuming this is all correct, I'm still wanting to figure out a way where we don't have to couple our implementations of the bridged contracts (like the oracle and governor bridges) to whatever type of bridge they're interacting with.

We could try to make all of our bridge implementations look like this sort of inheritance tree.

We have an abstract oracle and governor contracts that define an unimplemented _sendMessage(bytes) function that needs to be implemented by a derived class. They also implement a _receiveMessage(bytes) function.

Then there are bridge adapter contracts that you can inherit from. They are also abstract. They implement a _sendMessage(bytes) function, but they have an unimplemented _receiveMessage(bytes) function that needs to be implemented by a derived class. Basically the opposite of the above.

Then all we would (theoretically need to do) is have a derived contracts that look like this (literally no code other than combining base classes). Note that this is not tested, so I haven't verified that this type of inheritance actually fills in the missing abstract implementations as one would expect.:
```
contract PolygonOracleSink is PolygonChildBridge, OracleSink {}

contract PolygonGovernorSink is PolygonChildBridge, GovernorSink {}

contract OptimismOracleSink is OptimismChildBridge, OracleSink {}

contract OptimismGovernorSink is OptimismChildBridge, GovernorSink {}

...
```

Thoughts? Just trying to brainstorm ways to generalize a bit so the only thing we need to add on each new bridge is the bridging logic, itself.

mrice32 (commented):
This is super cool! Very nice work on figuring all of this out!

A few comments.

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
nicholaspai: - no not yet
- yes we definitely should have an external code section
- the FX contracts are currently getting audited, so I don't know. Technically they are unaudited and would be safer for us to audit them. But its a lot of code.

packages/core/deploy/017_deploy_oracle_root_tunnel.js
nicholaspai: Yeah we definitely should. I'll add `-polygon` tags for now.

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
nicholaspai: final fees maybe? i'm not sure tbh, but its an opportunity to add more data. But we should probably add it from the OptimisticOracle side, not this contract.

packages/core/contracts/l2/polygon/tunnel/FxBaseRootTunnel.sol
nicholaspai: I'll add a readme and re-ping you for review!

Thinking about this proposed architecture.

packages/core/contracts/l2/polygon/lib/RLPReader.sol
nicholaspai: Done!

packages/core/contracts/l2/polygon/tunnel/FxBaseRootTunnel.sol
nicholaspai: Added read me with diagram

packages/core/contracts/l2/polygon/OracleBaseTunnel.sol
chrismaree: I like this idea of having an external section. this is the case in a bunch of places!

packages/core/contracts/l2/polygon/OracleRootTunnel.sol
chrismaree: I dont think so. I think you hit it just wanted to think it through fully as always more painful to add them back later!

packages/core/contracts/l2/polygon/README.md
```diff
@@ -0,0 +1,42 @@
+# Polygon <> Ethereum State Transfer
```
chrismaree, line 1: 10/10 on this. pretty awesome stuff!

chrismaree: We should use this as a source of information for `docs.umaproject.org` when documenting the L2 compatibility with UMA.

packages/core/contracts/l2/polygon/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+This document describes the architecture of how arbitrary messages can be passed between Ethereum and Polygon. For a more detailed explanation from the Polygon official docs site, go [here](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer).
+
+# Two-way bridge between Root on Ethereum and Child on Polygon
+
+At a high level we are deploying "Root" and "Child" bridge contracts on both networks that communicate only with each other and the native [state sync](https://docs.matic.network/docs/contribute/state-sync/state-sync/) infrastructure that Polygon uses to pass data between the Ethereum and Polygon EVM's. Polygon uses "tunnel" to describe what [other](https://docs.tokenbridge.net/amb-bridge/about-amb-bridge) [relayer](https://forum.makerdao.com/t/announcing-the-optimism-dai-bridge-with-fast-withdrawals/6938) [systems](https://developer.offchainlabs.com/docs/inside_arbitrum#bridging) call "bridges".
+
+Diagram of oracle tunnel system can be found [here](https://docs.google.com/presentation/d/19KdCTYFzCOR3H-ns2K2HB5eiTrPIwKTakKT4Qkr7G0I/edit#slide=id.p).
```
chrismaree, line 9: do you want to export this and add it inline? markdown supports images!
```suggestion
Diagram of the oracle tunnel system can be seen below
![image](https://user-images.githubusercontent.com/12886084/121140379-115e8e80-c83a-11eb-89e9-27a694e20814.png)
```

chrismaree: I uploaded it for you :)

packages/core/deploy/005_deploy_sink_oracle.js
```diff
@@ -25,5 +25,5 @@ const func = async function (hre) {
   });
 };
 module.exports = func;
-func.tags = ["SinkOracle", "bridge-l2"];
+func.tags = ["SinkOracle", "l2-chainbridge"];
```
chrismaree, line 28: it's not called `chainbridge` any more, is it?

packages/core/deploy/018_deploy_oracle_child_tunnel.js
```diff
@@ -0,0 +1,28 @@
+// Grabbed from official Polygon docs
+// https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer/#pre-requisite
+const ADDRESSES_FOR_NETWORK = {
```
chrismaree, line 3: nit: I think this could be done in fewer lines without wrapping onto new lines like this.

packages/core/deploy/005_deploy_sink_oracle.js
nicholaspai: it is still yeah

packages/core/deploy/018_deploy_oracle_child_tunnel.js
nicholaspai: Good point, done

packages/core/contracts/polygon/README.md
```diff
@@ -0,0 +1,42 @@
+# Polygon <> Ethereum State Transfer
```
mrice32, line 1: This is awesome!

mrice32 (approved):
LGTM! Very nice

packages/core/contracts/polygon/OracleChildTunnel.sol
```diff
@@ -96,9 +97,37 @@ contract OracleChildTunnel is OracleBaseTunnel, OracleAncillaryInterface, FxBase
...
         require(lookup.state == RequestState.Resolved, "Price has not been resolved");
         return lookup.price;
     }
+
+    /**
+     * @dev We don't handle specifically the case where `ancillaryData` is not already readily translateable in utf8.
+     * For those cases, we assume that the client will be able to strip out the utf8-translateable part of the
+     * ancillary data that this contract stamps.
+     */
+    function _stampAncillaryData(bytes memory ancillaryData, address requester) internal view returns (bytes memory) {
```
nicholaspai, line 111: @chrismaree @mrice32 this the ancillary data I'm thinking of stamping. I don't think the child tunnel address or root tunnel address is neccessary to stamp since (1) the root tunnel will be the one requesting the price on L1 and (2) the child tunnel is linked with the root tunnel.
Thoughts?

nicholaspai:
@mrice32 @chrismaree I realized that we could easily refactor `AncillaryData.sol` to provide an internal method for the re-used logic that appends a <key:value> to the ancillary data where `value` is an address. PTAL at this [PR](https://github.com/UMAprotocol/protocol/pull/3054/commits/fd0d4551762886acf251f5947507e3a57549f853)

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -33,4 +33,28 @@ library AncillaryData {
...
+    /**
+     * @notice Adds "key:value" to `currentAncillaryData` where `value` is an address that first needs to be converted
+     * to utf8 bytes. For example, if `utf8(currentAncillaryData)="k1:v1"`, then this function will return 
+     * `utf8(k1:v1,key:value)`, and if `currentAncillaryData` is blank, then this will return `utf8(key:value)`.
+     * @param currentAncillaryData This bytes data should ideally be able to be utf8-decoded, but its OK if not. 
+     * @param key Again, this bytes data should ideally be able to be utf8-decoded, but its OK if not. 
+     * @param value An address to set as the value in the key:value pair to append to `currentAncillaryData`.
+     * @return Newly appended ancillary data.
+     */
+    function appendAddressKey(bytes memory currentAncillaryData, bytes memory key, address value) internal pure returns (bytes memory) {
```
mrice32: +1!

There may be ways we want to expand this in the future, but I think this is the perfect functionality that we need today!

packages/core/test/common/AncillaryData.js
```diff
@@ -0,0 +1,61 @@
+const { assert } = require("chai");
+
+const AncillaryDataTest = artifacts.require("AncillaryDataTest");
+const { utf8ToHex } = web3.utils
+
+contract("AncillaryData", function () {
```
mrice32, line 6: +1 -- tests are very thorough and test a lot of important use cases.

mrice32 (approved):
LGTM!

packages/core/contracts/polygon/OracleChildTunnel.sol
```diff
@@ -0,0 +1,117 @@
...
+    /**
+     * @dev We don't handle specifically the case where `ancillaryData` is not already readily translateable in utf8.
+     * For those cases, we assume that the client will be able to strip out the utf8-translateable part of the
+     * ancillary data that this contract stamps.
+     */
+    function _stampAncillaryData(bytes memory ancillaryData, address requester) internal pure returns (bytes memory) {
+        // Price requests that originate from this method, on Polygon, will ultimately be submitted to the DVM on
+        // Ethereum via the OracleRootTunnel. Therefore this contract should stamp its requester's address in the
+        // ancillary data so voters can conveniently track the requests path to the DVM.
+        return AncillaryData.appendAddressKey(ancillaryData, "childTunnelRequester", requester);
```
mrice32, line 115: Should we (in addition or as a replacement for the requester) stamp it with some sort of identifier for the chain (like the chain id if we have it)? The reason I mention this is that:
1. For a particular chain, it's not clear that a voter _needs_ to know which contract this request is coming from (this isn't provided on mainnet). 
2. A voter would likely be able to back this info out as long as they know which chain it's coming from. We should still provide it if it's needed for price resolution, but the requesting contract would probably provide it in that case?
3. Addresses aren't unique across chains, so this doesn't uniquely identify the requester. You can deploy the same contract at the same address on multiple chains (uniswap does this, for example).

Thoughts?

nicholaspai: added in [this commit](https://github.com/UMAprotocol/protocol/pull/3054/commits/bc713c39438ad40de375b3f1c2dab6d4f66da20f). As discussed offline, converting the uint `chainId` to utf8 bytes was not trivial FYI.

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -34,6 +34,32 @@ library AncillaryData {
         else return bytes1(uint8(b) + 0x57);
     }
 
+    /**
+     * @notice Converts a uint into a base-10, UTF-8 representation stored in a `string` type.
+     * @dev This method is based off of this code: https://stackoverflow.com/a/65707309.
+     */
+    function toUtf8String(uint256 x) internal pure returns (string memory) {
```
mrice32, line 41: Why not change this to `toUtf8Bytes` and return bytes?

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -54,6 +80,16 @@ library AncillaryData {
         } else {
             prefix = abi.encodePacked(key, ":");
         }
-        return abi.encodePacked(currentAncillaryData, prefix, AncillaryData.toUtf8Bytes(value));
+        return abi.encodePacked(currentAncillaryData, prefix, toUtf8Bytes(value));
+    }
+
+    function appendChainId(bytes memory currentAncillaryData) internal view returns (bytes memory) {
```
mrice32, line 86: Have you thought about just having a single overloaded method:
```
function appendKeyValue(
        bytes memory currentAncillaryData,
        bytes memory key,
        address value
    );

function appendKeyValue(bytes memory currentAncillaryData, bytes memory key, uint256 value) internal view returns (bytes memory);
```

And then the caller can choose to put pass in the chain id if they like... thoughts?

nicholaspai: I like that

nicholaspai: I ended up refactoring the methods to use a common `internal` method because I found that after overloading the methods, the `PerpetualLib` contract ended up being over the bytecode limit. Not sure what happened here, but PTAL at this [PR](https://github.com/UMAprotocol/protocol/pull/3054/commits/1f4c72d24de1d51a5a331f79ae6cc28bf7a48397)

mrice32 (approved):
LGTM!! Very nice

PR #3057, opened 2021-06-03
feat(core): add custom ancillary data to CFD

mrice32:
**Motivation**

Users should be able to specify custom ancillary data that the CFD will send along with the request.

**Summary**

Added custom ancillary data arg.

Removes the token address ancillary data that is currently provided.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/test/financial-templates/contract-for-difference/ContractForDifference.js
```diff
@@ -29,7 +29,7 @@ let collateralWhitelist;
 let identifierWhitelist;
 let optimisticOracle;
 let finder;
-let ancillaryData;
+let ancillaryData = web3.utf8ToHex("CUSTOM_DATA");
```
nicholaspai, line 32: nit: thoughts on starting to make test data resemble how we actually intend to use ancillary data in prod, such as {key:value, key2:value2, ...}

nicholaspai (approved):

packages/core/test/financial-templates/contract-for-difference/ContractForDifference.js
mrice32: +1 -- will fix this

mrice32: Done!

packages/core/test/financial-templates/contract-for-difference/ContractForDifference.js
```diff
@@ -96,6 +95,7 @@ contract("ContractForDifference", function (accounts) {
       collateralTokenAddress: collateralToken.address,
       finderAddress: finder.address,
       contractForDifferenceLibraryAddress: contractForDifferenceLibrary.address,
+      ancillaryData,
```
chrismaree, line 98: cool but should you not add a test also that this `ancillaryData` is included within the OO price requests?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -100,10 +102,14 @@ contract ContractForDifference is Testable, Lockable {
      * @param _expirationTimestamp unix timestamp of when the contract will expire.
      * @param _collateralPerPair how many units of collateral are required to mint one pair of synthetic tokens.
      * @param _priceIdentifier registered in the DVM for the synthetic.
-     * @param _longTokenAddress ERC20 token used as long in the CFD. Requires mint and burn rights granted to this contract.
-     * @param _shortTokenAddress ERC20 token used as short in the CFD. Requires mint and burn rights granted to this contract.
+     * @param _longTokenAddress ERC20 token used as long in the CFD. Requires mint and burn rights granted to this
```
chrismaree, line 105: feels like it would be better to just shorten this by 1 word rather than using a whole extra line. Can leave this for a subsequent cleanup PR though where we try to reduce any of these extra wrapped lines.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -132,6 +139,7 @@ contract ContractForDifference is Testable, Lockable {
         collateralToken = _collateralAddress;
 
         financialProductLibrary = _financialProductLibraryAddress;
+        customAncillaryData = _customAncillaryData;
```
chrismaree, line 142: you should probs bound the length of this. it would be possible to put an ancillaryData in here that the OO would reject. `ancillaryBytesLimit = 8192;` in the OO and this should be checked here as well or you could create a CFD that would deploy just fine but would reject when requesting a price.

chrismaree (approved):
looks good! had a few comments

packages/core/test/financial-templates/contract-for-difference/ContractForDifference.js
mrice32: I think this is fixed? I added a test case where I grab the request from the OO and use the ancillary data to look it up. Is that what you were looking for? PTAL

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
mrice32: Added a check for this. PTAL. It's a little more complicated than just checking that hardcoded limit because the OO adds data too. So we need to check the length of the stamped OO data.

chrismaree (approved):
LGTM! I have one outstanding nit on the comment length in the constructor natspec.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
mrice32: Done!

PR #3059, opened 2021-06-04
clean(bot-tests): remove repeated function from repo

chrismaree:
**Motivation**

There are a number of common places through the monorepo where the method `createContractObjectFromJson` are repeated over and over. This PR removes this.


nicholaspai (approved):
Niiiiice thanks for circling back on this refactor

PR #3060, opened 2021-06-04
fix(common-ContractUtils): Add web3 param injection to createContractObjectFromJson

chrismaree:
**Motivation**

`createContractObjectFromJson` requires a web3 instance to extract the current provider. This works fine when running in unit tests as the `web3` instance is global. however, when run in production, the `web3` instance is _not_ globle. This causes errors when running in production.

When running within the `proxytransaction` wrapper, for example, we would need to pass in the `web3` object.

This PR simply addresses this.

### DIFF
--- a/packages/common/src/UniswapV3Helpers.js
+++ b/packages/common/src/UniswapV3Helpers.js
@@ -22,8 +22,8 @@ function decodePriceSqrt(priseSqrt) {
 }
 
 // Fetch the decoded price from a uniswap pool from slot0.
-async function getCurrentPrice(poolAddress) {
-  const pool = await createContractObjectFromJson(UniswapV3Pool).at(poolAddress);
+async function getCurrentPrice(poolAddress, web3) {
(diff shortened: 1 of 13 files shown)


packages/common/src/UniswapV3Helpers.js
```diff
@@ -22,8 +22,8 @@ function decodePriceSqrt(priseSqrt) {
 }
 
 // Fetch the decoded price from a uniswap pool from slot0.
-async function getCurrentPrice(poolAddress) {
-  const pool = await createContractObjectFromJson(UniswapV3Pool).at(poolAddress);
+async function getCurrentPrice(poolAddress, web3) {
```
nicholaspai, line 25: This should be `Web3` right, since you import that on line 3?

nicholaspai (approved):
Nice fix, one small nit

packages/common/src/UniswapV3Helpers.js
chrismaree: no sir. this is web3 ass passed in from the caller of this method.

PR #3061, opened 2021-06-04
improve(ancillary-data): Standardize stamping of ancillary data by contracts

nicholaspai:
**Motivation**

(insert UMIP/Proposal page here) declares a standard for UMA integrators that ancillary data should be translateable to UTF8 as a key-value dictionary (delineated by commas) in order to be considered by DVM voters. This PR adds an AncillaryData library that contracts like `OptimisticOracle` and `FundingRateApplier` can use to add data to ancillary data.

**Summary**

What other data should we stamp into the `OptimisticOracle`'s ancillary data assuming that the OptimisticOracle will be deployed to Polygon and rollups?

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


mrice32:
@nicholaspai as discussed offline, encodePacked is purely being used to concatenate two byte arrays, not directly encode data. Each contract can choose to encode its data any way it chooses.

If you do regular encode, you end up with something that is wrapped at each level. The data structure would look roughly like:

```
optimisticOracleData: {
  callingContractStamp: {
    someNumber: 1,
    someOtherNumber: 2,
    callingContractStamp: {
      someAddress: "0x12345",
      someOtherAddress: "0x123456"
  },
  requesterAddress: "0x1234"
```


Each contract that "stamps" would effectively be nesting the existing stamps inside its encoding. This would need to be unwrapped through recursive calls to decode offchain.

Instead of this, we've basically been appending, which basically means each contract encodes its data and appends it to the back of the bytes array like `[callingContract1Stamp, callingContract2Stamp, OOStamp]`, so there's no nesting. The downside is that you need to be able to decode it in a systematic way, by knowing the types and length of any dynamic types or just imposing something like a global utf8 encoding that can easily be appended to.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -545,7 +546,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
      * @return the stampped ancillary bytes.
      */
     function stampAncillaryData(bytes memory ancillaryData, address requester) public pure returns (bytes memory) {
```
nicholaspai, line 548: I didn't want to change this public method name yet in case client bots call it directly. Will do later.

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -54,14 +54,13 @@ contract("FundingRateApplier", function (accounts) {
...
-    // Check that the ancillary data matches expectations.
-    // Note: hashing seems to be the only way to generate a tight packing offchain.
-    const expectedHash = web3.utils.soliditySha3(
-      { t: "address", v: collateral.address },
-      { t: "bytes", v: web3.utils.utf8ToHex("OptimisticOracle") },
-      { t: "address", v: fundingRateApplier.address }
-    );
-    assert.equal(web3.utils.soliditySha3({ t: "bytes", v: lastQuery.ancillaryData }), expectedHash);
+    // FundingRateApplier initially saves the synthetic token address to ancillary data:
+    const expectedFRAAncillaryData = utf8ToHex(`tokenAddress:${collateral.address.substr(2).toLowerCase()}`)
```
nicholaspai, line 58: Important to note here how client dapps will need to translate ancillary data

packages/core/test/oracle/OptimisticOracle.js
```diff
@@ -653,6 +653,16 @@ contract("OptimisticOracle", function (accounts) {
...
     await optimisticRequester.clearState();
 
+    // Check that OptimisticOracle stamped ancillary data as expected before sending to Oracle, and that we can decode
+    // it.
+    const priceRequests = await mockOracle.getPastEvents("PriceRequestAdded", { fromBlock: 0 });
+    assert.equal(priceRequests.length, 1, "should only be one price request escalated to MockOracle");
+    const stampedAncillaryData = priceRequests[0].returnValues.ancillaryData;
+    assert.equal(
+      hexToUtf8(stampedAncillaryData),
+      `${hexToUtf8(ancillaryData)},requester:${optimisticRequester.address.substr(2).toLowerCase()}`
```
nicholaspai, line 663: we'll probably want to add a common library method to parse ancilary data

packages/common/src/HardhatConfig.js
```diff
@@ -27,9 +27,9 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
     },
     networks: {
       hardhat: {
-        gas: 11500000,
-        blockGasLimit: 11500000,
-        allowUnlimitedContractSize: false,
+        gas: 12000000,
+        blockGasLimit: 12000000,
+        allowUnlimitedContractSize: true,
```
mrice32, line 32: nit: why do we need this? If this is breaking, doesn't it signal that the contract is too big to deploy on mainnet (which we would want to show up in tests)?

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+/**
+ * @title Library for encoding and decoding ancillary data for DVM price requests.
+ */
+library AncillaryData {
+    // TODO: Get source for this code, h/t Matt.
```
mrice32, line 8: https://ethereum.stackexchange.com/a/8447/47801

There's also a slightly different but more general implementation in the same thread: 

https://ethereum.stackexchange.com/a/58341/47801


Note: I think strings and bytes are roughly the same thing in solidity, but it might be worth checking that these implementations that were designed for strings also work for bytes (I believe the one you have, at least, does.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -651,8 +652,14 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
...
-    function _stampAncillaryData(bytes memory ancillaryData, address requester) internal pure returns (bytes memory) {
-        return abi.encodePacked(ancillaryData, "OptimisticOracle", requester);
+    function _appendToAncillaryData(bytes memory ancillaryData, address requester)
+        internal
+        pure
+        returns (bytes memory)
+    {
+        // Since this contract will be the one to formally submit DVM price requests, its useful for voters to know who
+        // the original requester was.
+        bytes memory prefix = ",requester:";
```
mrice32, line 662: How do we deal with commas? Some callers could pass in empty ancillary data -- is our format okay with an extra leading comma? Should we add a conditional? Or should we do something else?

Not super opinionated, just wanted to raise the issue.

packages/core/test/oracle/OptimisticOracle.js
mrice32: Agreed, that would be really nice. I don't think it would be too hard to write a parsing function that just dumps these key-value pairs into a js object. We can do this in a follow-up,

mrice32 (approved):
Looks great!

Thoughts on whether/when we should upgrade all these on-chain?

packages/common/src/HardhatConfig.js
nicholaspai: I actually think the main issue is that our gas limit is set too low and we should leave it undefined and let hardhat estimate it. The contract is well under the bytecode according to CalculateContractBytecode

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: We could check length of `bytes`:

```
        bytes memory prefix;
        if (ancillaryData.length > 0) {
            prefix = ",requester:";
        } else {
            prefix = "requester:";
        }
        return abi.encodePacked(ancillaryData, prefix, AncillaryData.toUtf8Bytes(requester));
```

packages/core/package.json
```diff
@@ -71,7 +71,7 @@
   ],
   "scripts": {
     "test": "yarn mocha-test && yarn hardhat-test && yarn truffle-test",
-    "truffle-test": "truffle test ./truffle-test/*",
+    "truffle-test": "truffle test ./truffle-test/scripts/*",
```
nicholaspai, line 74: This command has been wrong for a while, fixing it here.

packages/common/src/HardhatConfig.js
nicholaspai: Ended up removing this as I realize that `PerpetualLib` was actually over bytecode limit, a problem that I fixed by removing require messages.

mrice32: Sounds good!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Yep, that seems like a good solution!

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
```diff
@@ -134,16 +135,13 @@ abstract contract FundingRateApplier is EmergencyShutdownable, FeePayer {
         nonReentrant()
         returns (FixedPoint.Unsigned memory totalBond)
     {
-        require(fundingRate.proposalTime == 0, "Proposal in progress");
+        require(fundingRate.proposalTime == 0);
```
chrismaree, line 138: were these stripped out to save bytecode?

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -0,0 +1,31 @@
...
+ */
+library AncillaryData {
+    /**
+     * @notice Returns utf8-encoded address that can be read via web3.utils.hexToUtf8.
+     * Source: https://ethereum.stackexchange.com/questions/8346/convert-address-to-string/8447#8447
+     * @dev Will return address in all lower case characters and without the leading 0x.
+     * @param x address to encode.
+     * @return utf8 encoded address bytes.
+     */
+    function toUtf8Bytes(address x) internal pure returns (bytes memory) {
```
chrismaree, line 15: what is the gas costs associated with this method?

packages/core/test/financial-templates/common/FundingRateApplier.js
```diff
@@ -54,14 +54,15 @@ contract("FundingRateApplier", function (accounts) {
...
-      { t: "bytes", v: web3.utils.utf8ToHex("OptimisticOracle") },
-      { t: "address", v: fundingRateApplier.address }
-    );
-    assert.equal(web3.utils.soliditySha3({ t: "bytes", v: lastQuery.ancillaryData }), expectedHash);
+    // FundingRateApplier initially saves the synthetic token address to ancillary data:
+    const expectedFRAAncillaryData = utf8ToHex(`tokenAddress:${collateral.address.substr(2).toLowerCase()}`);
+
+    // OptimisticOracle should append its address:
+    const expectedAppendedAncillaryData = utf8ToHex(
+      `,optimisticOracleRequester:${fundingRateApplier.address.substr(2).toLowerCase()}`
```
chrismaree, line 62: nice. can see the `,` delimitation here.

packages/core/truffle-test/scripts/Voting.js
```diff
@@ -1,394 +0,0 @@
-const VotingScript = require("../../scripts/Voting.js");
```
chrismaree, line 1: why was this removed?

chrismaree (approved):
I had a few questions and nits but overall this is pretty sweet and a great way to standardize things.

To double-check: there are no risks with us updating this and it breaking any existing contracts, right? This is 100% backward compatible with everything else? I'm pretty sure this is the case but just wanted to double check :)

packages/core/contracts/financial-templates/common/FundingRateApplier.sol
nicholaspai: yes unfortunately.

packages/core/truffle-test/scripts/Voting.js
nicholaspai: It was a broken script  (Voting.js doesn't even exist in the repo anymore)

nicholaspai:
> I had a few questions and nits but overall this is pretty sweet and a great way to standardize things.
> 
> To double-check: there are no risks with us updating this and it breaking any existing contracts, right? This is 100% backward compatible with everything else? I'm pretty sure this is the case but just wanted to double check :)

I don't see how this could break existing contracts. No contracts assume anything about how `ancillaryData` is formatted, so redeploying the OO wouldn't change anything. Even the clients only call `optimisticOracle.stampAncillaryData()` whose interface does not change.

@mrice32 can you think of any unintentional breakages?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -651,8 +652,15 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
...
-    // Stamps the ancillary data blob with the optimistic oracle tag denoting what contract requested it.
     function _stampAncillaryData(bytes memory ancillaryData, address requester) internal pure returns (bytes memory) {
-        return abi.encodePacked(ancillaryData, "OptimisticOracle", requester);
+        // Since this contract will be the one to formally submit DVM price requests, its useful for voters to know who
+        // the original requester was.
+        bytes memory prefix;
+        if (ancillaryData.length > 0) {
+            prefix = ",optimisticOracleRequester:";
+        } else {
+            prefix = "optimisticOracleRequester:";
```
mrice32, line 662: Just to reduce gas, do you think we should shorten this to `ooRequester`?

mrice32 (approved):
LGTM! One minor comment on ancillary data length

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
nicholaspai: sure

PR #3062, opened 2021-06-07
feat(serverless-hub): Update to enable generalize bot runner to work in production

chrismaree:
**Motivation**

We have built a set of tools internally that enable us to run a series of bots using a "strategy runner" with minimal config. Using this, a bot runner can define a very succinct config that is expanded to run n number of bots.

Ideally, this setup should also work within our serverless infrastructure to simplify bot running internally.

This PR the strategy runner and some logic in the serverless hub to facilitate this. 

**Summary**

There are three main changes in this PR:
- Modified how the main winston `Transports` module works with google `trace-agent` to first check if the agent has started before initializing the module. If this is _not_ done running the strategy runner in production errors out on child process execution.
- Modified the bot strategy runner to support config objects added within stringified environment variables. This helps with running within production cloud run setups.
- Modified the serverlessHub to include a `commonConfig` param that enables the hub to define a set of common fields that should be injected into all spoke processes. This will enable common fields (such as whitelists for bot strategy runner) to be defined _once_ at the top of the config file and then re-used be each config.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/bot-strategy-runner/src/index.ts
```diff
@@ -118,13 +118,18 @@ const processExecutionOptions = async () => {
     .option("-fc, --fileConfig <path>", "input path to JSON config file.")
     .option("-uc, --urlConfig <path>", "url to JSON config hosted online. Private resources use access token")
     .option("-at, --accessToken <string>", "access token to access private configs online. EG private a repo")
+    .option("-env, --envConfig <bool>", "use a stringified config within your env set under `RUNNER_CONFIG")
     .parse(process.argv)
     .opts();
 
-  assert(options.fileConfig != undefined || options.urlConfig != undefined, "provide a file config or a URL config");
+  assert(
+    options.fileConfig != undefined || options.urlConfig != undefined || options.envConfig,
```
nicholaspai, line 126: why not write this as `assert(options.fileConfig || options.urlConfig || options.envConfig)`?

packages/financial-templates-lib/src/logger/Transports.js
```diff
@@ -18,15 +18,15 @@ function createTransports(transportsConfig = {}) {
   // If the logger is running in serverless mode then add the GCP winston transport and console transport.
   if ((transportsConfig.environment ?? process.env.ENVIRONMENT) == "serverless") {
     const { LoggingWinston } = require("@google-cloud/logging-winston");
-    require("@google-cloud/trace-agent").start();
+    if (!require("@google-cloud/trace-agent").get().enabled) require("@google-cloud/trace-agent").start();
```
daywiss, line 21: this library is imported mulitple times, is there a reason you dont just import once?

packages/bot-strategy-runner/src/index.ts
```diff
@@ -118,13 +118,18 @@ const processExecutionOptions = async () => {
     .option("-fc, --fileConfig <path>", "input path to JSON config file.")
     .option("-uc, --urlConfig <path>", "url to JSON config hosted online. Private resources use access token")
     .option("-at, --accessToken <string>", "access token to access private configs online. EG private a repo")
+    .option("-env, --envConfig <bool>", "use a stringified config within your env set under `RUNNER_CONFIG")
```
daywiss, line 121: why do u need to set this option explicltly, why not just merge the env RUNNER_CONFIG if it exists?

daywiss (approved):
lgtm

packages/bot-strategy-runner/src/index.ts
chrismaree: fair! have refactored

packages/bot-strategy-runner/src/index.ts
chrismaree: just to be more explicit.

chrismaree: but you are right. I'll modify this sligtly to let it pull this in by default without the user having to say they are using an env config.

packages/financial-templates-lib/src/logger/Transports.js
daywiss: i dont see an easy way to do that, unless its instanced at the top level and passed in as a dependency, but its not worth the effort.

packages/bot-strategy-runner/src/index.ts
daywiss: i could imagine you might not want to delete some complex env json, but maybe disable it instead so you can easily enable it later. might be a use case for that, its up to you.

PR #3065, opened 2021-06-07
feat(sdk): add in memory sorted kv store 

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1228/add-sorted-in-memory-key-value-store

**Summary**
This is a useful primitive to allow for sorted queries on an in memory store. Will use this to serve price data until google datastore integration. Allows for querying between timestamps, or "slicing" ( like array.slice) which is great for pagination.

**Details**
This code was pulled from some previously unpublished code and converted to typescript.  Added more thorough tests.  Behaves similarly to traditional sorted No sql databases where keys are sorted alphabetically when dealing with strings.

This adds additional functions to the store interface `slice` and `between`

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1228/add-sorted-in-memory-key-value-store

### DIFF
--- a/packages/sdk/src/libs/stores/google-datastore/store.ts
+++ b/packages/sdk/src/libs/stores/google-datastore/store.ts
@@ -1,8 +1,8 @@
-import type { Store } from "..";
+import type { SortedStore } from "..";
 import type { Datastore, Key } from "@google-cloud/datastore";
 
 // TODO: this is a work in progress which will be completed during datastore integration.
-export default function <D>(kind: string, store: Datastore): Store<string, D> {
+export default function <D>(kind: string, store: Datastore): SortedStore<string, D> {
   function makeKey(id: string): Key {
     return store.key([kind, id]);
   }
--- a/packages/sdk/src/libs/stores/google-datastore/store.ts
+++ b/packages/sdk/src/libs/stores/google-datastore/store.ts
@@ -44,6 +44,16 @@ export default function <D>(kind: string, store: Datastore): Store<string, D> {
   async function size() {
     return 0;
   }
+  // TODO: implement these functions
+  async function clear() {
+    throw new Error("clear is not implemented");
+  }
+  async function slice() {
+    return [];
+  }
+  async function between() {
+    return [];
+  }
 
   return {
     get,
@@ -54,5 +64,8 @@ export default function <D>(kind: string, store: Datastore): Store<string, D> {
     values,
     keys,
     size,
+    clear,
+    between,
+    slice,
   };
 }
--- a/packages/sdk/src/libs/stores/index.ts
+++ b/packages/sdk/src/libs/stores/index.ts
@@ -2,19 +2,24 @@ export { default as JsMap } from "./js-map";
 export { default as SortedJsMap } from "./sorted-js-map";
 export { default as GoogleDatastore } from "./google-datastore";
 
-// types
-export interface Store<I, D> {
-  get: (id: I) => Promise<D | undefined>;
-  set: (id: I, data: D) => Promise<void>;
-  delete: (id: I) => Promise<void>;
-  has: (id: I) => Promise<boolean>;
-  entries: () => Promise<Array<[I, D]>>;
-  values: () => Promise<Array<D>>;
-  keys: () => Promise<Array<I>>;
+// Base store interface which is inspired by JS map
+export interface Store<Id, Data> {
+  get: (id: Id) => Promise<Data | undefined>;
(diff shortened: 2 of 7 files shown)


packages/sdk/src/libs/stores/index.ts
```diff
@@ -11,4 +12,9 @@ export interface Store<I, D> {
   values: () => Promise<Array<D>>;
   keys: () => Promise<Array<I>>;
   size: () => Promise<number>;
+  clear: () => Promise<void>;
+  // these fields are for stores with sorted data
+  between: (a: I, b: I) => Promise<Array<D>>;
+  slice: (id: I, length: number) => Promise<Array<D>>;
+  driver?: any;
```
daywiss, line 19: added some additional functions on stores. any ones that cant be directly implemented by the store should just return empty array or throw error ( if typescript will compile)

nicholaspai (approved):
I like putting this structure PR up for review before implementation, as it will make the latter (presumably chunkier) review a lot smoother. I agree with this structure you have

packages/sdk/src/libs/stores/index.ts
```diff
@@ -11,4 +12,9 @@ export interface Store<I, D> {
   values: () => Promise<Array<D>>;
   keys: () => Promise<Array<I>>;
   size: () => Promise<number>;
+  clear: () => Promise<void>;
+  // these fields are for stores with sorted data
+  between: (a: I, b: I) => Promise<Array<D>>;
+  slice: (id: I, length: number) => Promise<Array<D>>;
```
mrice32, line 18: A few comments about these new functions:

- Does this mean that Ids need to have some sort of defined ordering? Should/can we augment the type to enforce this?
- If we have instances of this interface that _don't_ implement some of these (or error when you call them), should they be part of the same interface? It seems like this breaks the abstraction and requires the caller to be aware of what underlying store type they're using, right?

I could see us keeping the abstraction by defining different interfaces, so we can have users specify exactly what interface they need at compile time:
```ts
export interface BaseStore<I, D> {
  get: (id: I) => Promise<D | undefined>;
  entries: () => Promise<Array<[I, D]>>;
  values: () => Promise<Array<D>>;
  size: () => Promise<number>;
}

export interface KVStore<I, D> extends BaseStore<I, D> {
  ...
}

export interface OrderedStore<I, D> extends BaseStore<I, D> {
  ...
}
```

Or if you think every Ordered store can do everything a KV store can do (which isn't obviously true for things like random deletes), you could make OrderedStore extend KVStore.

Thoughts?

mrice32 (commented):
Looks awesome! One high level question about the expanded store interface.

packages/sdk/src/libs/stores/index.ts
daywiss: > Does this mean that Ids need to have some sort of defined ordering? Should/can we augment the type to enforce this?

as long as ids are strings or numbers, they can be sorted just fine. arrays or objects would not work, maybe this needs to be clarified.

> If we have instances of this interface that don't implement some of these (or error when you call them), should they be part of the same interface? It seems like this breaks the abstraction and requires the caller to be aware of what underlying store type they're using, right?

Yea you are right, i kind of mashed things into the same interface.  originally tried what you are suggesting, breaking up the interfaces, but thought it didn't gain much because i only needed the 2 additional functions.  but what you are suggesting i think is the right way to do this. I think i will refactor it based on this suggestion.

packages/sdk/src/libs/stores/google-datastore/store.ts
```diff
@@ -1,8 +1,8 @@
-import type { Store } from "..";
+import type { SortedStore } from "..";
```
chrismaree, line 1: I've not seen this kind of import syntax before. the `from ".."`, what does it mean exactly?

chrismaree (approved):
this looks really useful! is there a pre-built JS library that would let us do this?

daywiss:
> this looks really useful! is there a pre-built JS library that would let us do this?

thats a good question, im sure there is, but i didnt look. this one was kind of prebuilt though, scavenged it from an old project.  if you know of a good existing library though i would take a look.

packages/sdk/src/libs/stores/sorted-js-map/README.md
```diff
@@ -0,0 +1,46 @@
+# JS Map Store
```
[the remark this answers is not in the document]
daywiss: that title is wrong actually, will update

PR #3066, opened 2021-06-07
feat(sdk): add prices table using sorted map

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1226/add-sdk-price-table

**Summary**

Adds the price table which relies on the sorted KV store. This is so we can capture and query time series data. Also refactors some of the tables api to get rid of circular dependencies and make them more consistent.

This is a subtask of https://app.clubhouse.io/uma-project/story/930/create-api-to-get-current-prices-of-collateral-s

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1226/add-sdk-price-table


packages/sdk/src/libs/tables/blocks/js-map.ts
```diff
@@ -1,13 +1,9 @@
...
+import { JsMap } from "../generic";
 
 // personally dont like to have this as a named export, but cannot export both function and type as default
-export const JsMap = (type = "Block") => {
-  function makeId(data: Data) {
-    return data.number;
-  }
-
-  const table = GenericJsMap<number, Data>(type, makeId);
+export const Table = (type = "Block") => {
```
mrice32, line 5: ^would classes solve this exporting issue?

mrice32 (approved):
LGTM!

Minor side note: there's a lot of jumping around between files that are (right now) very minimal wrappers of other files in this code (the historical prices dir is an example). Not sure if we're expecting a lot of these files to become more meaty, but if they're expected to stay below 20 lines or so, it could make sense to consolidate a little just to avoid having to jump around so much to follow the code and inheritance/composition paths.

packages/sdk/src/libs/tables/blocks/js-map.ts
daywiss: yes it would. i wanted to do a test with classes, but realized once you change any dependency to a class, the effects ripple through all the code if you want to use inheritance. was thinking we might want to take an hour and pair program to attempt a conversion!

daywiss:
> Minor side note: there's a lot of jumping around between files that are (right now) very minimal wrappers of other files in this code (the historical prices dir is an example). Not sure if we're expecting a lot of these files to become more meaty, but if they're expected to stay below 20 lines or so, it could make sense to consolidate a little just to avoid having to jump around so much to follow the code and inheritance/composition paths.

hmm yea i do have some justification for this, but perhaps we can strategize on this if you think its a problem, would like to hear your thoughts.

Heres an example of expanding the features of the [price table in an upcoming pr](https://github.com/UMAprotocol/protocol/blob/3a321ef5abdd6b015584631acc71868944973012/packages/sdk/src/libs/tables/historical-prices/sorted-js-map.ts). you can see how there are helper functions to simplify querying. 

you can see an [example here of a helper function for emps to add sponsors](https://github.com/UMAprotocol/protocol/blob/3a321ef5abdd6b015584631acc71868944973012/packages/sdk/src/libs/tables/emps/js-map.ts)

I expect more instances like this as the functionality grows.

PR #3068, opened 2021-06-07
feat(core): add all truffle migrations to hardhat deploy

mrice32:
**Motivation**

As a part of the process of moving to hardhat, all truffle migrations need to be converted to hardhat deploy scripts.

**Summary**

Notes: 
- These scripts are sometimes simpler than their truffle counterparts, as they do not include other setup steps.
- Hardhat deploy includes a "live" parameter that's part of the network config. This is true by default on most networks, but is false by default on localhost or hardhat, which is precisely what we want. We use this to determine whether to deploy the Timer.
- I reordered the deployment scripts: DVM -> financial contracts -> Bridges. The ordering generally shouldn't matter, but I just did this to make it feel a bit more organized where things that were alike were next to one another.

Next steps:
- Add setup scripts/tasks to allow deployers to easily set these contracts up for production or test use.
- Add test fixtures to specifically set these contracts up for different test usages.
- Convert tests to using these rather than truffle migrations and drop truffle completely in core.
- Drop truffle artifacts in favor of hardhat artifacts for committing addresses and releasing bytecode/abi/addresses via npm.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/deploy/002_deploy_timer.js
```diff
@@ -0,0 +1,20 @@
+const func = async function (hre) {
+  const { deployments, getNamedAccounts, network } = hre;
+  const { deploy } = deployments;
+
+  const { deployer } = await getNamedAccounts();
+  const { live } = network;
+
+  if (live === undefined) throw new Error("Network has no live parameter");
```
mrice32, line 8: Leaving this here in case some configuration change in hardhat causes this to be undefined. We would want to catch that to avoid accidentally deploying a Timer in prod.

packages/common/src/PublicNetworks.js
```diff
@@ -3,6 +3,7 @@
 // ID as the network (i.e. Rinkeby will return 4 as the chainId), but some networks with chainID's > 255 need to
 // override the default behavior because their network ID is too high.
 const BRIDGE_CHAIN_ID = {
+  1337: 253,
```
mrice32, line 6: Had to add this to make deployments work for localhost ganache networks.

nicholaspai: nice thanks!

packages/core/deploy/002_deploy_timer.js
nicholaspai: I agree with leaving this here

packages/core/deploy/006_deploy_voting.js
```diff
@@ -0,0 +1,41 @@
...
+  const gatPercentage = { rawValue: web3.utils.toWei("0.05", "ether") };
+
+  // Set the inflation rate.
+  const inflationRate = { rawValue: web3.utils.toWei("0.0005", "ether") };
+
+  // Set the rewards expiration timeout.
+  const rewardsExpirationTimeout = 60 * 60 * 24 * 14; // Two weeks.
+
+  // Set phase length to one day.
+  const phaseLength = "86400";
```
nicholaspai, line 23: What could be kind of useful is detecting whether we are on a production network or not, and setting phase length to 10 mins or shorter if not!

nicholaspai (approved):
Very nice, thanks for setting this up. LGTM

packages/core/deploy/006_deploy_voting.js
mrice32: Interesting..... I might add that

packages/core/deploy/002_deploy_timer.js
```diff
@@ -0,0 +1,20 @@
...
+
+  // If live === false, don't deploy a timer.
+  if (live === false) {
+    await deploy("Timer", {
+      from: deployer,
+      args: [],
```
chrismaree, line 14: I'm not super familiar with hard had deploy but do you have to provide empty arguments like this? would it work fine without this param?

packages/core/deploy/002_deploy_timer.js
```diff
@@ -0,0 +1,20 @@
...
+      log: true,
```
chrismaree, line 15: equally, seeing all of them have `log:true`, is there a way to set them all to default to `log:true` without defining it over and over again?

packages/core/deploy/016_deploy_perpetual_creator.js
```diff
@@ -0,0 +1,24 @@
...
+    args: [Finder.address, TokenFactory.address, Timer.address],
+    libraries: {
+      PerpetualLib: PerpetualLib.address,
+    },
+    log: true,
+  });
+};
+module.exports = func;
+func.tags = ["PerpetualCreator", "perpetual"];
+func.dependencies = ["Finder", "TokenFactory", "Timer"];
```
chrismaree, line 24: we'll need to add a script for `017_deploy_contract_for_difference_creator.js` once that contract has been merged in.

packages/core/deploy/015_deploy_emp_creator.js
```diff
@@ -20,4 +21,4 @@ const func = async function (hre) {
 };
 module.exports = func;
 func.tags = ["ExpiringMultiPartyCreator", "emp"];
-func.dependencies = ["Finder", "TokenFactory"];
+func.dependencies = ["Finder", "TokenFactory", "Timer"];
```
chrismaree, line 24: does this also not have a dependency on `dvm`? I guess as long as there is a `Finder` it will deploy fine but a `Finder` with no `dvm` is not very useful. This is likely what you were referring to yesterday though where you can deploy a contract ecosystem without configuring the finder correctly and that this is a reasonable assumption to let us simplify these scripts.

chrismaree (approved):
these scripts are super slick! great work.

packages/core/deploy/015_deploy_emp_creator.js
mrice32: I think our philosophy (or at least the one I'm going with :)) is only specifying dependencies that are _explicitly_ required to be passed in in the constructor. For instance, you may not need the entire dvm to be deployed, you may actually deploy the mock oracle in its place.

This is a weakly held opinion, but in my mind, this is part of keeping the process as light as possible.

Thoughts?

packages/core/deploy/016_deploy_perpetual_creator.js
mrice32: Yep!

packages/core/deploy/012_deploy_optimistic_oracle.js
```diff
@@ -8,13 +8,14 @@ const func = async function (hre) {
...
     from: deployer,
-    args: [defaultLiveness, Finder.address, ZERO_ADDRESS],
+    args: [defaultLiveness, Finder.address, Timer.address],
     log: true,
   });
 };
 module.exports = func;
 func.tags = ["OptimisticOracle", "dvm"];
-func.dependencies = ["Finder"];
+func.dependencies = ["Finder", "Timer"];
```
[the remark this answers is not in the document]
mrice32: Agreed, it's really nice. You can basically say "I need an Optimistic Oracle", and hardhat just figures out what needs to be deployed to make that happen (and will reuse existing deployments if they exist!).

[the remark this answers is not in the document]
mrice32: I think this will make upgrades really really simple.

packages/core/deploy/002_deploy_timer.js
mrice32: Will check the docs.

mrice32: Good point. You don't. It seems to default to empty. Removed where empty.

packages/core/deploy/002_deploy_timer.js
mrice32: It defaults to false, I don't see any info about how to make it default to true :/

packages/core/deploy/015_deploy_emp_creator.js
mrice32: Going to go ahead and merge, but yeah, this is somewhat opinionated around keeping things minimal as we discussed yesterday. Let me know if you'd like anything to change in a follow-up!

packages/core/deploy/006_deploy_voting.js
mrice32: As discussed offline, I think we should wait on a follow-up for this feature.

PR #3075, opened 2021-06-08
feat(contractForDiffrence): CFD creator contract and unit tests

chrismaree:
**Motivation**

As part of building a CFD financial contract, we require a contract factory that will be used to deploy CFD instances. This PR adds this factory contract and unit tests.

Note that the `contracts/common/implementation/AddressWhitelist.sol` contract was modified to use a `contracts/common/interfaces/AddressWhitelistInterface.sol` contract to decrease the bytecode required when querying the `AddressWhitelist` from downstream contracts.


packages/core/contracts/common/interfaces/AddressWhitelistInterface.sol
```diff
@@ -0,0 +1,12 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+interface AddressWhitelistInterface {
+    function addToWhitelist(address newElement) external;
```
nicholaspai, line 5: do you not need to label these as `virtual`?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
```diff
@@ -0,0 +1,132 @@
...
+        require(bytes(syntheticName).length != 0, "Missing synthetic name");
+        require(bytes(syntheticSymbol).length != 0, "Missing synthetic symbol");
+        TokenFactory tf = TokenFactory(tokenFactoryAddress);
+
+        // If the collateral token does not have a `decimals()` method, then a default precision of 18 will be
+        // applied to the newly created synthetic token.
+        uint8 collateralDecimals = _getSyntheticDecimals(collateralAddress);
+        ExpandedIERC20 longToken =
+            tf.createToken(
+                string(abi.encodePacked(syntheticName, "Long Token")),
```
nicholaspai, line 79: Should there be a leading space in front of "Long Token": " Long Token"?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
```diff
@@ -0,0 +1,132 @@
...
+                string(abi.encodePacked("l", syntheticSymbol)),
+                collateralDecimals
+            );
+        ExpandedIERC20 shortToken =
+            tf.createToken(
+                string(abi.encodePacked(syntheticName, "Short Token")),
```
nicholaspai, line 85: Same comment as above

packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
```diff
@@ -0,0 +1,132 @@
...
+                string(abi.encodePacked("s", syntheticSymbol)),
```
nicholaspai, line 86: Should the short and long tokens share some common symbol to indicate that they are related? This might be too wordy and ugly, but maybe something like `cfdsX` and `cfdlX`. But obviously better name than that haha

nicholaspai (approved):
LGTM. On a very separate note, but while reviewing I thought of this, once [this PR about ancillary data](https://github.com/UMAprotocol/protocol/pull/3061) is merged in, we should probably change how the `ContractForDifference.sol` implements `_getAncillaryData`.

Thoughts?

packages/core/contracts/common/interfaces/AddressWhitelistInterface.sol
chrismaree: ye good point. done!

packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
chrismaree: ye! true true.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
chrismaree: well, whatever your `syntheticSymbol` and `syntheticName` is will be common between them. is that not sufficient?

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -123,10 +123,11 @@ contract ContractForDifference is Testable, Lockable {
         address _timerAddress
     ) Testable(_timerAddress) {
         finder = _finderAddress;
-        require(_expirationTimestamp > getCurrentTime());
+        require(_expirationTimestamp > getCurrentTime(), "Expiration time stamp in past");
```
mrice32, line 126:
```suggestion
        require(_expirationTimestamp > getCurrentTime(), "Expiration timestamp in past");
```

packages/core/test/financial-templates/contract-for-difference/ContractForDifferenceCreator.js
```diff
@@ -0,0 +1,197 @@
...
+    // Force each test to start with a simulated time that's synced to the startTimestamp.
+    await timer.setCurrentTime(startTimestamp);
+
+    collateralToken = await Token.new("Wrapped Ether", "WETH", 18, { from: deployer });
+    await collateralToken.addMember(1, deployer, { from: deployer });
+    await collateralToken.mint(sponsor, toWei("1000"), { from: deployer });
+
+    await collateralWhitelist.addToWhitelist(collateralToken.address);
+
+    optimisticOracle = await OptimisticOracle.new(optimisticOracleLiveness, finder.address, timer.address);
```
mrice32, line 65: Just out of curiosity, why do we need to create a new OO? Isn't this provided by migrations?

mrice32 (approved):
LGTM!

packages/core/test/financial-templates/contract-for-difference/ContractForDifferenceCreator.js
chrismaree: this is true. this mistake is actually repeated in a few CFD unit tests. I will update all these in a separate cleanup PR.

PR #3079, opened 2021-06-08
feat(api): add historical price queries for collateral addresses 

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/932/create-api-to-get-historical-prices-of-collaterals

**Summary**

This adds a historical prices table in the sdk, which uses  a sorted map under the hood. Api uses this to construct historical price tables for all collateral currencies and keeps it up to date every N blocks as specified by `UPDATE_BLOCKS`

**Details**
Readme is updated with api calls:
- `historicalPricesByAddress(address:string,start:number=0,end:number=Date.now(),currency:"usd"="usd") => PriceSample[]`
- `sliceHistoricalPricesByAddress(address:string,start:number=0,length:number=1,currency:"usd"="usd") => PriceSample[]`

You can also check the postman collection of calls: https://www.getpostman.com/collections/e99f4e09e2443cb31ef4

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/932/create-api-to-get-historical-prices-of-collaterals

### DIFF
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -1,6 +1,6 @@
 import assert from "assert";
 import * as uma from "@uma/sdk";
-import { Json, Actions, Libs, CurrencySymbol } from "..";
+import { Json, Actions, Libs, CurrencySymbol, PriceSample } from "..";
 
 const { exists } = uma.utils;
 
@@ -41,6 +41,32 @@ export function Handlers(config: Json, libs: Libs): Actions {
       const priceSample = libs.prices[currency].latest[address];
       assert(exists(priceSample), "No price for address: " + address);
       return priceSample;
+    },
+    async historicalPricesByAddress(
+      address: string,
+      start = 0,
+      end: number = Date.now(),
+      currency: "usd" = "usd"
+    ): Promise<PriceSample[]> {
+      assert(start >= 0, "requires a start value >= 0");
+      assert(exists(libs.prices[currency]), "invalid currency type: " + currency);
+      assert(exists(libs.prices[currency].history[address]), "no prices for address" + address);
+      const prices = await libs.prices[currency].history[address].betweenByTimestamp(start, end);
+      // convert this to tuple to save bytes.
+      return prices.map(({ price, timestamp }) => [timestamp, price]);
+    },
+    async sliceHistoricalPricesByAddress(
+      address: string,
+      start = 0,
+      length = 1,
+      currency: "usd" = "usd"
+    ): Promise<PriceSample[]> {
+      assert(start >= 0, "requires a start value >= 0");
+      assert(exists(libs.prices[currency]), "invalid currency type: " + currency);
+      assert(exists(libs.prices[currency].history[address]), "no prices for address" + address);
+      const prices = await libs.prices[currency].history[address].sliceByTimestamp(start, length);
+      // convert this to tuple to save bytes.
+      return prices.map(({ price, timestamp }) => [timestamp, price]);
     },
   };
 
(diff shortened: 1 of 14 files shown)


packages/api/src/services/prices.ts
```diff
@@ -11,26 +12,84 @@ export default function (config: Config, libs: Libs) {
   assert(coingecko, "requires coingecko library");
   assert(prices[currency], `requires prices.${currency}`);
 
-  async function updatePrice(address: string) {
+  // if we have a new emp address, this will create a new price table structure to store historical price data
```
mrice32, line 15: Are we exclusively relying on coingecko to give us the market price of the synthetic tokens?

Are we expecting to use this to do things like compute collateralization ratio, etc?

These seem like shaky assumptions that created problems in the mining scripts. Should we not use our own price feed libraries to get more accurate market and notional prices?
What happens if coingecko doesn't list a token?

mrice32 (commented):
Just one high level comment!

packages/api/src/services/prices.ts
daywiss: this is only for collaterals, synthetic tokens is a future pr and will need to use bot price feeds!

mrice32: I see. That makes sense, then. Although it's not entirely clear if CG lists prices for all approved collateral tokens, I think this is probably a good starting point!

As a side note: long term, it might be preferable to use (as a fallback or replacement) 1inch or other DEX aggregator to grab prices (although there are tradeoffs there too).

mrice32: Also, why do we need collateral prices?

packages/api/src/services/actions.ts
```diff
@@ -1,6 +1,6 @@
 import assert from "assert";
 import * as uma from "@uma/sdk";
-import { Json, Actions, Libs, CurrencySymbol } from "..";
+import { Json, Actions, Libs, CurrencySymbol, PriceSample } from "..";
```
mrice32, line 3: Is an action a particular service? I'm having a little trouble understanding what's supposed to go in the services directory and what the action service is? I would think of each service as a concrete subset of endpoints on the api or something?

packages/sdk/src/libs/coingecko/coingecko.test.ts
```diff
@@ -2,7 +2,7 @@ import assert from "assert";
 import Coingecko from ".";
 
 // this requires e2e testing, should only test manually for now
-describe.skip("coingecko", function () {
+describe("coingecko", function () {
```
mrice32, line 5: How are we walling off manual/e2e tests from the other tests in ci?

Side note: are the compiled test files currently being exported to npm?

mrice32 (approved):
LGTM!

packages/api/src/services/prices.ts
daywiss: this is to calculate tvl per emp and over all emps in usd.  a lot of this overlaps with the work chris did with the merkle tree tvl calculations.

packages/sdk/src/libs/coingecko/coingecko.test.ts
daywiss: hm that should be skipped actually. i have meant to skip all non unit tests.  i dont have a great solution for how to deal wiht e2e style tests. 
tests are not compiled with the production build afaik!

packages/api/src/services/actions.ts
daywiss: the action service has the business logic to query tables in response to api calls. Express service exposes the actions to http. All other services fill various tables by pulling in data from outside sources. I consider services things which could live as their own process if they need to, there's no particular functionality prescribed to them, though they should be kind of specialized.

PR #3081, opened 2021-06-09
feat(gas-rebate): Rebate 10

nicholaspai:

**Summary**

One small change to make sure that UMA-ETH prices cut off after the 18th decimal so that we can parse them with `toWei`.


packages/affiliates/gas-rebate/VoterGasRebate.js
```diff
@@ -629,7 +629,8 @@ async function getHistoricalUmaEthPrice(dailyPrices) {
   for (let i = 0; i < dailyPrices.length; i++) {
     dailyUmaPrices.push({
       timestamp: dailyPrices[i].timestamp,
-      avgPx: umaEthPrices[i].toString(),
+      avgPx: umaEthPrices[i].toString().substr(0, 18), // Cut off any digits beyond 18th decimal so that we we can parse
```
daywiss, line 632: not a huge deal, but you can use `umaEthPrices[i].toFixed(18)` as well if this is a number and not a BN or string

daywiss (approved):

packages/affiliates/gas-rebate/VoterGasRebate.js
nicholaspai: Aha! thanks

PR #3067, opened 2021-06-07
feat(api): add latest prices for collateral addresses

daywiss:
**Motivation**
https://app.clubhouse.io/uma-project/story/930/create-api-to-get-current-prices-of-collateral-s
**Summary**
Cleanup some sdk stuff and make latest prices for collaterals available from the api. These prices are updated every X blocks, as specified by the environment variable `UPDATE_BLOCKS`.

**Details**

Adds api routes:

- ` allLatestPrices(currency='usd') => {[address:string]:{price:string,timestamp:number}}`
- ` latestPriceByAddress(address:string,currency:'usd') => {price:string,timestamp:number}`

when running locally, you can use postman to query:
https://www.getpostman.com/collections/e99f4e09e2443cb31ef4

**Testing**

Check a box to describe how you tested these changes and list the steps for reviewers to test.

- [x]  Ran end-to-end test, running the code as in production
- [ ]  New unit tests created
- [ ]  Existing tests adequate, no new tests required
- [ ]  All existing tests pass
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/930/create-api-to-get-current-prices-of-collateral-s

PR #3082, opened 2021-06-10
feat(polygon-amb): Add Unit Tests for Oracle Tunnel contracts

nicholaspai:
**Motivation**

Contracts were missing unit tests due to size of #3067 .

**Summary**

`e2e.js` is a great test to understand to see how the `OracleChildTunnel` and `OracleRootTunnel` would integrate in production.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+const FxChild = artifacts.require("FxChildMock");
+const FxRoot = artifacts.require("FxRootMock");
+const OracleChildTunnel = artifacts.require("OracleChildTunnel");
+const OracleRootTunnel = artifacts.require("OracleRootTunnelMock");
+const Finder = artifacts.require("Finder");
+const Registry = artifacts.require("Registry");
+const MockOracle = artifacts.require("MockOracleAncillary");
+const IdentifierWhitelist = artifacts.require("IdentifierWhitelist");
+
+contract("Polygon <> Ethereum Tunnel: End-to-End Test", async (accounts) => {
```
nicholaspai, line 17: @UMAprotocol/eng this is the key unit test to understand integration between Polygon <> Ethereum tunnels

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+        hexToUtf8(event.identifier) === hexToUtf8(testIdentifier) &&
+        event.time.toString() === testTimestamp.toString() &&
+        event.ancillaryData.toLowerCase() === expectedStampedAncillaryData.toLowerCase() &&
+        event.price.toString() === testPrice
+    );
+
+    // Until price is resolved on Child, hasPrice and getPrice should return false and revert, respectively.
+    // Note: the ancillary data input into hasPrice and getPrice is the original, pre-stamped ancillary data, which
+    // means that the original requester does not know about the ancillary data stamping that took place behind the
+    // scene.
```
mrice32, line 169: Should this comment imply that this is expected to succeed here, not fail?

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+const { assert } = require("chai");
+const TruffleAssert = require("truffle-assertions");
+
+const { utf8ToHex, toWei, hexToUtf8 } = web3.utils;
+
+const StateSync = artifacts.require("StateSyncMock");
+const FxChild = artifacts.require("FxChildMock");
+const FxRoot = artifacts.require("FxRootMock");
+const OracleChildTunnel = artifacts.require("OracleChildTunnel");
+const OracleRootTunnel = artifacts.require("OracleRootTunnelMock");
```
mrice32, line 11: Why are we using a mock for the root tunnel but the real contract for the child tunnel? Just curious, not suggesting we shouldn't!

mrice32 (commented):
Looks awesome!

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+
+  before(async function () {
+    finder = await Finder.deployed();
+    identifierWhitelist = await IdentifierWhitelist.deployed();
+    registry = await Registry.deployed();
+    mockOracle = await MockOracle.new(finder.address, ZERO_ADDRESS);
+
+    await finder.changeImplementationAddress(utf8ToHex(interfaceName.Registry), registry.address);
+    await finder.changeImplementationAddress(utf8ToHex(interfaceName.Oracle), mockOracle.address);
+    await finder.changeImplementationAddress(utf8ToHex(interfaceName.IdentifierWhitelist), identifierWhitelist.address);
```
chrismaree, line 50: if you are importing these contracts from the truffle artifacts using the `.deployed` syntax why do you need to `changeImplementationAddress` on them? are they not registered from the migrations like in other tests in core?

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+
+    // Set up Oracle tunnel system:
+    oracleChild = await OracleChildTunnel.new(fxChild.address, finder.address);
+    oracleRoot = await OracleRootTunnel.new(checkpointManager, fxRoot.address, finder.address);
+    await oracleChild.setFxRootTunnel(oracleRoot.address);
+    await oracleRoot.setFxChildTunnel(oracleChild.address);
+    await registry.registerContract([], oracleRoot.address, { from: owner });
+
+    expectedStampedAncillaryData = utf8ToHex(
+      `${hexToUtf8(testAncillaryData)},childRequester:${owner.substr(2).toLowerCase()},childChainId:31337`
```
chrismaree, line 73: prehaps put childChain as a variable?

Also, it would be nice for you to show in a comment the exact expected form. you _can_ re-construct it but there is a fair amount of nested calls here so a comment might help a little bit.

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+  });
+  it("request price from Polygon to Ethereum, resolve price from Ethereum to Polygon", async function () {
+    // Only registered caller can call.
+    assert(
+      await didContractThrow(
+        oracleChild.requestPrice(testIdentifier, testTimestamp, testAncillaryData, { from: rando })
+      )
+    );
+
+    // Should emit MessageSent event with ABI encoded requestPrice parameters..
```
chrismaree, line 84:
```suggestion
    // Should emit MessageSent event with ABI encoded requestPrice parameters.
```

packages/core/test/polygon/e2e.js
```diff
@@ -0,0 +1,176 @@
...
+      [
+        oracleRoot.address,
+        oracleChild.address,
+        web3.eth.abi.encodeParameters(
+          ["bytes32", "uint256", "bytes", "int256"],
+          [testIdentifier, testTimestamp, expectedStampedAncillaryData, testPrice]
+        ),
+      ]
+    );
+    TruffleAssert.eventEmitted(
```
chrismaree, line 138: +1 this is awesome

chrismaree (approved):
This is really awesome. it helped a fair amount with my understanding to see the e2e tests!

packages/core/test/polygon/e2e.js
nicholaspai: Ah I'll add a comment. Because the `RootTunnel` has an internal `_processMessageFromChild` that we want to test directly, whereas the corresponding `_processMessageFromRoot` on the `ChildTunnel` gets tested via `FxChildMock.onStateReceive`.

packages/core/test/polygon/e2e.js
nicholaspai: Good catch!

PR #3085, opened 2021-06-10
feat(api): add token names and decimals to returned emp state

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1291/need-erc20-token-data-in-api

**Summary**

Previous versions did not properly fetch token decimals or token names for the emp. This adds this ability, plus dedicated storage for related erc20 tokens.

**Details**

This adds a new pattern called `queries` which joins data between tables. This is a utility file which can be re-used between services, found in `src/libs/queries.ts`

new routes are added to query erc20 information for a single address or all known addresses

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1291/need-erc20-token-data-in-api

### DIFF
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -1,23 +1,23 @@
 // allow for more complex queries, joins and shared queries between services
-import { Libs } from "..";
+import type { AppState } from "..";
 import uma from "@uma/sdk";
 import { calcGcr } from "./utils";
 import assert from "assert";
 import bluebird from "bluebird";
 
-export default (libs: Libs) => {
+export default (appState: AppState) => {
   async function getAnyEmp(empAddress: string) {
-    if (await libs.emps.active.has(empAddress)) {
-      return libs.emps.active.get(empAddress);
+    if (await appState.emps.active.has(empAddress)) {
+      return appState.emps.active.get(empAddress);
     }
-    return libs.emps.expired.get(empAddress);
+    return appState.emps.expired.get(empAddress);
   }
   // joins emp with token state and gcr
   async function getFullEmpState(empState: uma.tables.emps.Data) {
     assert(empState.tokenCurrency, "requires tokenCurrency");
     assert(empState.collateralCurrency, "requires collateralCurrency");
-    const token = await libs.erc20s.get(empState.tokenCurrency);
-    const collateral = await libs.erc20s.get(empState.collateralCurrency);
+    const token = await appState.erc20s.get(empState.tokenCurrency);
+    const collateral = await appState.erc20s.get(empState.collateralCurrency);
 
     const state = {
       ...empState,
@@ -34,11 +34,11 @@ export default (libs: Libs) => {
   }
 
   async function listActiveEmps() {
-    const emps = libs.emps.active.values();
+    const emps = appState.emps.active.values();
     return bluebird.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
   }
   async function listExpiredEmps() {
-    const emps = libs.emps.expired.values();
+    const emps = appState.emps.expired.values();
     return bluebird.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
   }
 
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -5,7 +5,9 @@ import { calcGcr } from "./utils";
 import assert from "assert";
 import bluebird from "bluebird";
 
-export default (appState: AppState) => {
+type Dependencies = Pick<AppState, "erc20s" | "emps">;
+
+export default (appState: Dependencies) => {
   async function getAnyEmp(empAddress: string) {
     if (await appState.emps.active.has(empAddress)) {
       return appState.emps.active.get(empAddress);
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -1,0 +1,51 @@
+// allow for more complex queries, joins and shared queries between services
+import { Libs } from "..";
+import uma from "@uma/sdk";
+import { calcGcr } from "./utils";
+import assert from "assert";
+import bluebird from "bluebird";
+
+export default (libs: Libs) => {
+  async function getAnyEmp(empAddress: string) {
+    if (await libs.emps.active.has(empAddress)) {
+      return libs.emps.active.get(empAddress);
+    }
+    return libs.emps.expired.get(empAddress);
+  }
+  // joins emp with token state and gcr
+  async function getFullEmpState(empState: uma.tables.emps.Data) {
+    assert(empState.tokenCurrency, "requires tokenCurrency");
+    assert(empState.collateralCurrency, "requires collateralCurrency");
+    const token = await libs.erc20s.get(empState.tokenCurrency);
+    const collateral = await libs.erc20s.get(empState.collateralCurrency);
+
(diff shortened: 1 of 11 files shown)


packages/api/src/services/erc20-state.ts
```diff
@@ -0,0 +1,40 @@
+import { clients } from "@uma/sdk";
+import { Libs, Json } from "..";
+import { asyncValues } from "../libs/utils";
+type Config = Json;
+export default function (config: Config, libs: Libs) {
+  const { provider, erc20s, collateralAddresses, syntheticAddresses } = libs;
```
mrice32, line 6: I'm a little confused at the naming... why are the addresses to be tracked passed in as a lib? Seems like an odd name?

Also, as a secondary note: the Libs type seems to be quite expansive -- is libs a global object that's expected to be static and shared? Or should the requirements for each of these methods be down-specified to only the variables they actually use/need? Just seems odd to couple all the signatures of somewhat unrelated functions.

packages/api/src/services/erc20-state.ts
```diff
@@ -0,0 +1,40 @@
...
+    });
+  }
+  async function updateToken(address: string) {
+    if (await erc20s.has(address)) return;
+    const state = await getTokenState(address);
+    return erc20s.upsert(address, state);
+  }
+  async function updateTokens(addresses: string[]) {
+    return Promise.allSettled(addresses.map(updateToken));
+  }
```
mrice32, line 24: Is it the case that this function will not throw, but updateToken will? Are these meant to be private or public functions? As public functions, it could be confusing if these very similar functions have such different error behavior.

packages/api/src/services/erc20-state.ts
```diff
@@ -0,0 +1,40 @@
+import { clients } from "@uma/sdk";
+import { Libs, Json } from "..";
+import { asyncValues } from "../libs/utils";
+type Config = Json;
+export default function (config: Config, libs: Libs) {
+  const { provider, erc20s, collateralAddresses, syntheticAddresses } = libs;
+  async function getTokenState(address: string) {
+    const instance = clients.erc20.connect(address, provider);
+    return asyncValues({
```
mrice32, line 9: I'm assuming this is meant to only be used as an internal function (might be helpful to have a way to signal that)? if not, one might expect this to pull from the cache... but it seems this is meant to update the cache. An external caller might find this confusing since it _appears_ to be a getter.

packages/api/src/libs/queries.ts
```diff
@@ -0,0 +1,51 @@
+// allow for more complex queries, joins and shared queries between services
+import { Libs } from "..";
+import uma from "@uma/sdk";
+import { calcGcr } from "./utils";
+import assert from "assert";
+import bluebird from "bluebird";
+
+export default (libs: Libs) => {
```
mrice32, line 8: Similar question/comment about using the expansive type `Libs` here.

mrice32 (commented):
Looks good! A few minor questions/comments.

packages/api/src/services/erc20-state.ts
daywiss: good point, the naming sucks here, the 'libs' really represent state shared across the app and between services. Some service may update state, while another might query state.  For simplicity the entire state is composed into a single object defined by the `Libs` type. The state can be tables, sets, objects, whatever is the best way to represent the data. 

I am very open to renaming this, `State` seems too general, but maybe it could work, maybe `SharedState`, but i dont know

packages/api/src/services/erc20-state.ts
daywiss: thats right, this wont throw, its also meant to be internal. I'm adopting the allSettled pattern to resolve all promises even if some happen to throw. The errors do get logged though in the update function to make sure things dont silently fail.  The api for services arent really meant to be used like a library, although they technically could. I will update all calls to clarify which ones are meant to be public though.

packages/api/src/services/erc20-state.ts
daywiss: yes you are right, its meant to be internal, its technically is a getter though , it just returns the queried token state, the update happens in the `updateToken` function, which is also internal.

packages/api/src/libs/queries.ts
daywiss: on the expansiveness of this variable: its mainly a simplification/convention that I've bundled all mutable/query-able state into this variable. Breaking this apart per service is possible, but I don't see the benefit of this since everything is running in the same process. 

 If and when these services are broken out into their own processes it would probably be necessary to change this signature per process.  Would be happy to discuss this more though if you have some ideas.

packages/api/src/services/erc20-state.ts
daywiss: im going to rename this to `AppState` for now

packages/api/src/libs/queries.ts
mrice32: My main argument against it is that it kinda breaks modularity. if you wanted to test or reuse one of these components, you're required to initialize this libs variable with a lot of state that isn't actually needed by the function.

As a second point, from a readability perspective, it makes it a bit opaque as to what's being used by the class. Rather than being able to quickly glance at the signature, you have to read through the implementation.

My naive alternative is just to list the values that you do need. `{ a, b, c }: {a: TypeA, b: TypeB, c: TypeC}`. Since typescript uses duct-typing, this is completely compatible with the top-level libs type, it just allows you to be more specific about what you need/expect, and typescript will verify that the libs type that's being passed in matches. The downside is a bit of verbosity.

packages/api/src/services/erc20-state.ts
mrice32: If these are internal and the error behavior is commented, I think this is totally fine. the main concern was that one method just looked like the plural of the other, but they behave quite differently.

As a side note, I continue to feel like the lack of a distinction (even a superficial one, like a preceding underscore!) between private and public methods can be really confusing because it's unclear what's meant for public consumption. It's one major advantage of typescript (specifically typescript) classes, they have different declared visibilities that require code to make this clear.

packages/api/src/services/erc20-state.ts
daywiss: i see what you are saying though, its ambiguous where the source of the "get" is, yea thats a good point

mrice32 (approved):
Looks good! I have a comment about the libs/appState variable, but otherwise, LGTM!

packages/api/src/libs/queries.ts
daywiss: yea i could see that, basically have each service define its dependencies, while keeping the top level type the same. will give it a try

packages/api/src/services/erc20-state.ts
daywiss: hmm you are right, its hard to see whats public vs private, but i dont really understand why it matters. If we were just to treat all the functions as public, it doesn't materially affect anything, its just a matter of the developer understanding what the calls do.

packages/api/src/libs/queries.ts
daywiss: added something like this to each thing that uses the app state:
image

mrice32: Yep, that works! It's not 100% obvious what the types of those variables is, but it is clear which parts the function cares about!

mrice32 (approved):
LGTM!

PR #3086, opened 2021-06-10
feat(oo): adds optimistic oracle integration tutorial script

pemulis:
**Motivation**

We want to have a smart contract integration demo for the Optimistic Oracle to help developers understand how to use it.

**Summary**

Using the `DepositBox` tutorial as a base, this PR adds an `OptimisticDepositBox` contract and demo script that shows a simple example integration with the Optimistic Oracle.

**Details**

There are also some comments added to the `OptimisticOracleInterface`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

n/a


packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
```diff
@@ -0,0 +1,290 @@
...
+    function deposit(uint256 collateralAmount) public nonReentrant() {
+        require(collateralAmount > 0, "Invalid collateral amount");
+        OptimisticDepositBoxData storage depositBoxData = depositBoxes[msg.sender];
+        if (depositBoxData.collateral == 0) {
+            emit NewOptimisticDepositBox(msg.sender);
+        }
+
+        // Increase the individual deposit box and global collateral balance by collateral amount.
+        depositBoxData.collateral = depositBoxData.collateral.add(collateralAmount);
+        require(depositBoxData.collateral > 0, "Collateral not added");
```
mrice32, line 141: Is it possible for this to fail?

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
```diff
@@ -0,0 +1,290 @@
...
+            "Invalid withdraw request"
+        );
+
+        // Get the resolved price or revert.
+        uint256 exchangeRate = _getOraclePrice(depositBoxData.withdrawalRequestTimestamp);
+
+        // Calculate denomated amount of collateral based on resolved exchange rate.
+        // Example 1: User wants to withdraw $1000 of ETH, exchange rate is $2000/ETH, therefore user to receive 0.5 ETH.
+        // Example 2: User wants to withdraw $2500 of ETH, exchange rate is $2000/ETH, therefore user to receive 1.25 ETH.
+        uint256 denominatedAmountToWithdraw = depositBoxData.withdrawalRequestAmount.div(exchangeRate);
```
mrice32, line 198: What's the scaling of these numbers?

Is the exchange rate scaled by 1e18? If so, do we need to multiply by 1e18 before dividing (similar to how the FixedPoint library works)?

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
```diff
@@ -0,0 +1,290 @@
...
+
+        // If withdrawal request amount is > collateral, then withdraw the full collateral amount.
+        if (denominatedAmountToWithdraw > depositBoxData.collateral) {
```
mrice32, line 201: Should this be >=? Otherwise, I think if the withdrawal amount is equal to the collateral, the collateral will be 0 after, but we will have never emitted the Ended... event.

mrice32 (approved):
Looks great! A few minor comments

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
pemulis: The script passes in the exchange rate with no scaling (i.e., 2000 is 2000, not 2000000000000000000000), which works within the context of the demo script and contract but I guess would fail in practice if the exchange rate was scaled by 1e18.

Do you think 1e18 scaling should be baked into this demo, though, or just a note that in a real contract you would probably need to deal with scaling and point to examples? Scaling seems like a side topic that might distract from the main point of the tutorial. I'm inclined to just add a comment about scaling but could see an argument for going the other way.

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
pemulis: Good point; fixing.

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
pemulis: Hmm, I don't think so. I'll remove the check.

packages/core/contracts/financial-templates/demo/OptimisticDepositBox.sol
pemulis: I think I'll go ahead and use the FixedPoint library here.

mrice32: FixedPoint.Unsigned(depositBoxData.withdrawalRequestAmount).div(FixedPoint.Unsigned(exchangeRate)).rawValue;

pemulis: Thanks!

PR #3087, opened 2021-06-10
feat(api): add tvl endpoints for each contract 

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1245/obtain-the-tvl-of-each-contract-on-uma

**Summary**
Adds a statistics service to calculate tvl (will have other stats in future) based on prices emp state and erc20 token state (for decimals).
Exposes this through a couple api endpoints `getEmpStats` and `listEmpStats`

**Details**
image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1245/obtain-the-tvl-of-each-contract-on-uma

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -112,3 +112,11 @@ that call uses the collateral address, not the emp address.
 
 Returns latest synthetic price for an emp address. This differs from collateral price requests, as
 that call uses the collateral address, not the emp address.
+
+### getEmpStats(address: string, currency: "usd" = "usd") => StatData
+
+Get only tvl (and other future statistics) for a single emp.
+
+#### listEmpStats(currency: "usd" = "usd") => StatData[]
+
+Get all tvl (and other future statistics) for all known emps.
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -11,7 +11,7 @@ type Config = undefined;
 
 export function Handlers(config: Config, appState: Dependencies): Actions {
   const queries = Queries(appState);
-  const { registeredEmps, erc20s, collateralAddresses, syntheticAddresses, prices, synthPrices } = appState;
+  const { registeredEmps, erc20s, collateralAddresses, syntheticAddresses, prices, synthPrices,stats  } = appState;
 
   const actions: Actions = {
     echo(...args: Json[]) {
@@ -88,6 +88,17 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       const results = await prices[currency].history[address].sliceByTimestamp(start, length);
       // convert this to tuple to save bytes.
       return results.map(({ price, timestamp }) => [timestamp, price]);
+    },
+    async getEmpStats(address: string, currency: CurrencySymbol = "usd") {
+      assert(address, "requires address");
+      assert(currency, "requires currency");
+      assert(stats[currency], "No stats for currency: " + currency);
+      return stats[currency].latest.get(address);
+    },
+    async listEmpStats(currency: CurrencySymbol = "usd") {
+      assert(currency, "requires currency");
+      assert(stats[currency], "No stats for currency: " + currency);
+      return stats[currency].latest.values();
     },
   };
 
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -11,7 +11,7 @@ type Config = undefined;
 
 export function Handlers(config: Config, appState: Dependencies): Actions {
   const queries = Queries(appState);
-  const { registeredEmps, erc20s, collateralAddresses, syntheticAddresses, prices, synthPrices,stats  } = appState;
+  const { registeredEmps, erc20s, collateralAddresses, syntheticAddresses, prices, synthPrices, stats } = appState;
 
   const actions: Actions = {
     echo(...args: Json[]) {
--- a/packages/api/src/services/emp-stats.ts
+++ b/packages/api/src/services/emp-stats.ts
@@ -21,10 +21,14 @@ export default (config: Config, appState: Dependencies) => {
     const fullState = await queries.getFullEmpState(emp);
 
     assert(uma.utils.exists(fullState.collateralCurrency), "Emp requires collateralCurrency: " + address);
+
+    // PriceSample is type [ timestamp:number, price:string]
     const priceSample: PriceSample = prices[currency].latest[fullState.collateralCurrency];
     assert(uma.utils.exists(priceSample), "No latest price found for emp: " + address);
+
     const price = priceSample[1];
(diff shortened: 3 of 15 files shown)


packages/api/src/apps/api/README.md
```diff
@@ -96,3 +96,11 @@ Get name, decimals of an erc20 by address.
 ### getAllErc20Info() => Erc20Data[]
 
 List all known erc20s, collateral or synthetic and all known information about them
+
+### getEmpStats(address: string, currency: "usd" = "usd") => StatData
```
pemulis, line 100: Do you think we'll do something similar with LSPs?

packages/api/src/services/actions.ts
```diff
@@ -79,6 +79,17 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       // convert this to tuple to save bytes.
       return results.map(({ price, timestamp }) => [timestamp, price]);
     },
+    async getEmpStats(address: string, currency: CurrencySymbol = "usd") {
+      assert(address, "requires address");
+      assert(currency, "requires currency");
```
pemulis, line 84: I'm guessing these checks could never fail since `"usd"` is assigned above, but they're left in the code in case the default assignment is removed?

pemulis (approved):

packages/api/src/apps/api/README.md
daywiss: what is a LSP?

packages/api/src/services/actions.ts
daywiss: thats a good point, its possible for the frontend caller to override the default by passing null explicitly ( or any falsy value), its an edge case but this would catch it

packages/api/src/apps/api/README.md
daywiss: oh the CFD contracts, yea totally we could add stats about it

packages/api/src/services/actions.ts
```diff
@@ -79,6 +79,17 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       // convert this to tuple to save bytes.
       return results.map(({ price, timestamp }) => [timestamp, price]);
     },
+    async getEmpStats(address: string, currency: CurrencySymbol = "usd") {
```
chrismaree, line 82: what is the convention with documenting these kinds of functions with a natspec type thing? in other places in the protocol repo we document the function's intention, parameters, and return type. that might be useful here seeing this is an exported function that will be called from a lot of places. I think there are comment formatting types that will expose the comments when importing modules making working with packages much easier as your IDE will show you these definition comments import. thoughts?

chrismaree: why does this function need to be async? is the `latest.get(address)` async?

packages/api/src/services/actions.ts
```diff
@@ -79,6 +79,17 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
...
+      return stats[currency].latest.get(address);
+    },
+    async listEmpStats(currency: CurrencySymbol = "usd") {
+      assert(currency, "requires currency");
+      assert(stats[currency], "No stats for currency: " + currency);
+      return stats[currency].latest.values();
+    },
   };
 
   // list all available actions
```
chrismaree, line 95: would it be useful to have a function that updates the state and then returns? i.e I want the _latest_ stat for an emp.

chrismaree, line 95: would it be useful to have a method where you can feed in an array of EMP addresses and it returns all TVL/stats for that given array?

packages/api/src/services/emp-stats.ts
```diff
@@ -0,0 +1,55 @@
...
+  const queries = Queries(appState);
+  async function updateTvl(address: string) {
+    const emp = await queries.getAnyEmp(address);
+    // the full state has collateral decimals, pulled from erc20 state
+    const fullState = await queries.getFullEmpState(emp);
+
+    assert(uma.utils.exists(fullState.collateralCurrency), "Emp requires collateralCurrency: " + address);
+    const priceSample: PriceSample = prices[currency].latest[fullState.collateralCurrency];
+    assert(uma.utils.exists(priceSample), "No latest price found for emp: " + address);
+    const price = priceSample[1];
```
chrismaree, line 26: why use the 1st index and not the 0th index? or is the index 1 the price and index 0 the timestamp? in either case a comment might be useful

chrismaree (approved):
This is cool. I have some questions about how this scales to retrieving many EMPs and how updates work. but overall I like the pattern.

packages/api/src/services/actions.ts
daywiss: > why does this function need to be async? is the latest.get(address) async?

I guess it technically doesn't have to be, its just for consistency and i dont think it hurts anything. `latest.get(address)` is async but the parent function doesnt need to be.

daywiss: > what is the convention with documenting these kinds of functions with a natspec type thing? in other places in the protocol repo we document the function's intention, parameters, and return type.

yea this is tricky, these are meant to be functions exposed through some transport, in this case express. A user of the api would not necessarily have access to comments in this form. So from that perspective the comments would not be that useful and would be costly to maintain. This is what's great about graphql, because it gives a really useful discovery mechanism for api calls out of the box. 

What ive done in this case is documented all calls in the `api/README.md` and created example calls through [postman](https://www.getpostman.com/collections/e99f4e09e2443cb31ef4). though its not as thorough or formal as a natspec, but i feel like the readme / example calls are good enough for someone using the api. 

Would really love to be able to somehow compile the typescript into docs which could be put on a page or served through the api, like graphql.  Happy to get more discussion around this though!

packages/api/src/services/actions.ts
daywiss: stats are a unique case currently where all the calculations come from our internal state, and doing an update would be pretty cheap and require no third party calls.  Its not really a safe assumption to make in general though (an unprivileged user could spam third party calls), so I would propose that we try to maintain the latest state in the table itself through another more controlled process.  Currently this is done through polling every X seconds. 

At the moment its not really necessary to be super realtime, but if we were dealing with bots, or something that required the performance, i would change updating this table to be event based, and listen for changes on values which could change the stats object, this way we queries would always be up to date without needing to force an update on query.

daywiss: you read my mind, this is actually in a [future PR](https://github.com/UMAprotocol/protocol/pull/3090)

packages/api/src/services/emp-stats.ts
daywiss: yea i will comment, the `PriceSample` type is `[timestamp,price]`

PR #3090, opened 2021-06-11
feat(api): add total and partial tvl calls 

daywiss:
**Motivation**
https://app.clubhouse.io/uma-project/story/915/add-total-combined-tvl-to-api

**Summary**

Adds calls to get tvl based on all known emps, or a list of emp addresses through the `tvl` action.

**Details**

Adds a tvl endpoint `tvl(addresses?: string[], currency: CurrencySymbol = "usd") => string `
This tvl is currently returning a value of around 121m which is in line with https://monitor.simpleid.xyz/d/x4CYPILGk/uma?orgId=1&refresh=5m

This also adds some timing information just to see how long certain requests take for local profiling using `console.time`

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/915/add-total-combined-tvl-to-api

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -117,7 +117,7 @@ that call uses the collateral address, not the emp address.
 
 Get only tvl (and other future statistics) for a single emp.
 
-#### listEmpStats(currency: "usd" = "usd") => StatData[]
+### listEmpStats(currency: "usd" = "usd") => StatData[]
 
 Get all tvl (and other future statistics) for all known emps.
 
@@ -131,3 +131,7 @@ ascending. This is useful for charting betwen two dates. Note non synth historic
 Get a range of historical synthetic prices by emp address between start and count of price samples, sorted by timestamp
 ascending. This is useful for paginating data when you have X elements per page, and you know the last element seen.
 Note non synth historical prices are specified by erc20 address.
+
+### tvl(addresses?: string[], currency: CurrencySymbol = "usd") => string
+
+Returns tvl of all supplied addresses in USD 18 decimals. If no addresses supplied, returns TVL of all known contracts.
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -9,6 +9,7 @@ import Express from "../../services/express";
 import Actions from "../../services/actions";
 import { ProcessEnv, AppState } from "../..";
 import { empStats } from "../../tables";
+import { uniqueId } from "lodash";
 
 async function run(env: ProcessEnv) {
   assert(env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL");
@@ -98,25 +99,46 @@ async function run(env: ProcessEnv) {
   // expose calls through express
   await Express({ port: Number(env.EXPRESS_PORT) }, actions);
 
-  // main update loop, update every block
-  provider.on("block", (blockNumber: number) => {
+  // break all state updates by block events into a cleaner function
+  async function updateByBlock(blockNumber: number) {
+    await services.blocks.handleNewBlock(blockNumber);
     // dont do update if this number or blocks hasnt passed
-    services.blocks.handleNewBlock(blockNumber).catch(console.error);
     if (blockNumber - appState.lastBlockUpdate >= updateBlocks) {
-      services.registry(appState.lastBlock, blockNumber).catch(console.error);
-      services.emps(appState.lastBlock, blockNumber).catch(console.error);
+      // update everyting
+      await services.registry(appState.lastBlock, blockNumber);
+      await services.emps(appState.lastBlock, blockNumber);
+      await services.erc20s.update();
+      await services.empStats.update();
+
       appState.lastBlockUpdate = blockNumber;
-      services.erc20s.update().catch(console.error);
-      services.empStats.update().catch(console.error);
     }
     appState.lastBlock = blockNumber;
-    services.blocks.cleanBlocks(oldestBlock).catch(console.error);
+    await services.blocks.cleanBlocks(oldestBlock);
+  }
+
+  // main update loop, update every block
+  provider.on("block", (blockNumber: number) => {
+    // this timeid/time stuff just logs how long the update takes
+    const timeid = uniqueId("App State Updated ");
+    console.time(timeid);
+    updateByBlock(blockNumber)
+      .catch(console.error)
+      .finally(() => console.timeEnd(timeid));
   });
+
+  // separate out price updates into a different loop to query every few minutes
+  async function updatePrices() {
+    await services.collateralPrices.update();
+    await services.syntheticPrices.update();
+  }
 
   // coingeckos prices don't update very fast, so set it on an interval every few minutes
   setInterval(() => {
-    services.collateralPrices.update().catch(console.error);
-    services.syntheticPrices.update().catch(console.error);
+    const timeid = uniqueId("Prices Updated ");
+    console.time(timeid);
+    updatePrices()
+      .catch(console.error)
+      .finally(() => console.timeEnd(timeid));
   }, 5 * 60 * 1000);
 }
 
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -118,12 +118,7 @@ async function run(env: ProcessEnv) {
 
   // main update loop, update every block
   provider.on("block", (blockNumber: number) => {
-    // this timeid/time stuff just logs how long the update takes
-    const timeid = uniqueId("App State Updated ");
-    console.time(timeid);
-    updateByBlock(blockNumber)
-      .catch(console.error)
-      .finally(() => console.timeEnd(timeid));
+    updateByBlock(blockNumber).catch(console.error);
   });
 
   // separate out price updates into a different loop to query every few minutes
@@ -134,11 +129,7 @@ async function run(env: ProcessEnv) {
 
   // coingeckos prices don't update very fast, so set it on an interval every few minutes
   setInterval(() => {
-    const timeid = uniqueId("Prices Updated ");
-    console.time(timeid);
-    updatePrices()
-      .catch(console.error)
-      .finally(() => console.timeEnd(timeid));
+    updatePrices().catch(console.error);
   }, 5 * 60 * 1000);
 }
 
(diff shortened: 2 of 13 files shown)


packages/api/src/apps/api/app.ts
```diff
@@ -97,25 +98,46 @@ async function run(env: ProcessEnv) {
...
     appState.lastBlock = blockNumber;
-    services.blocks.cleanBlocks(oldestBlock).catch(console.error);
+    await services.blocks.cleanBlocks(oldestBlock);
+  }
+
+  // main update loop, update every block
+  provider.on("block", (blockNumber: number) => {
+    // this timeid/time stuff just logs how long the update takes
+    const timeid = uniqueId("App State Updated ");
+    console.time(timeid);
```
chrismaree, line 122: was this left over?

packages/api/src/services/express.ts
```diff
@@ -28,9 +28,14 @@ export default async (config: Config, actions: Actions) => {
 
   app.post("/:action", (req: Request, res: Response, next: NextFunction) => {
     const action = req?.params?.action;
+
+    // show how long calls take in ms
+    const timeid = lodash.uniqueId(`Action: ${action} `);
+    console.time(timeid);
```
chrismaree, line 34: also this?

packages/api/src/apps/api/README.md
```diff
@@ -117,6 +117,10 @@ that call uses the collateral address, not the emp address.
...
 Get only tvl (and other future statistics) for a single emp.
 
-#### listEmpStats(currency: "usd" = "usd") => StatData[]
+### listEmpStats(currency: "usd" = "usd") => StatData[]
 
 Get all tvl (and other future statistics) for all known emps.
+
+### tvl(addresses?: string[], currency: CurrencySymbol = "usd") => string
+
+Returns tvl of all supplied addresses in USD 18 decimals. If no addresses supplied, returns TVL of all known contracts.
```
chrismaree, line 126: can you explain what the process is of new contracts becoming "known"? is it about a contract being requested at some point and then it is cached into the future?

packages/api/src/apps/api/app.ts
daywiss: i kind of left this in as some logging to profile the app as it runs, its definitely not necessary and can remove it

packages/api/src/apps/api/README.md
daywiss: This is all based off the registry contract. Theres a service that pools the contract for events and will add emp addresses as they get added, so technically this will be handled in somewhat real time. 

You can see [here](https://github.com/UMAprotocol/protocol/blob/master/packages/api/src/services/emp-registry.ts#L18), the update function on the registry service simply just adds emp addresses to the set of known emps.

mrice32: With the OO, what happens when contracts aren't registered with the Registry? For instance, the LSP will not be registered.

packages/api/src/apps/api/app.ts
```diff
@@ -97,25 +98,46 @@ async function run(env: ProcessEnv) {
...
-      services.erc20s.update().catch(console.error);
-      services.empStats.update().catch(console.error);
     }
     appState.lastBlock = blockNumber;
-    services.blocks.cleanBlocks(oldestBlock).catch(console.error);
+    await services.blocks.cleanBlocks(oldestBlock);
+  }
+
+  // main update loop, update every block
+  provider.on("block", (blockNumber: number) => {
```
mrice32, line 119: Are we assuming that this is the update function? Is there a way to generalize this so it works inside a cron job? We don't need to do this now, but I think we should be mindful of architecting this in a way that traps us into running this in a long-running process.

For instance, I could imagine us wanting to have different trigger types:
1. New blocks/internal timer
2. An http post (we might separate the post calls for price updates vs blockchain updates if they need to run at diff frequencies).
3. A single-run script that executes an update and terminates.

I'm not saying we should _build_ these today. We're just building this API in a pretty generalized way, so we should be sure to add generalizations for use cases that we think are probable.

packages/api/src/apps/api/app.ts
```diff
@@ -97,25 +98,46 @@ async function run(env: ProcessEnv) {
   // expose calls through express
   await Express({ port: Number(env.EXPRESS_PORT) }, actions);
 
-  // main update loop, update every block
-  provider.on("block", (blockNumber: number) => {
+  // break all state updates by block events into a cleaner function
+  async function updateByBlock(blockNumber: number) {
```
mrice32, line 102: One thing I realized about this architecture (that I didn't realize before) is that it seems that we're designing this for a single process that does both ingestion and exposes the API. This seems to be specifically designed for a single-process and not generalizable for a multi-process system backed by a DB (which is our long term vision if I understand correctly).

Do you think it makes sense to separate out the API piece and the ingestion piece into distinct files/functions? This would allow us to run each by themselves and _also_ have a file that pulls both in and runs them together for an in-memory store. We could also do this all in one file and allow this to be customized via CLI params or env variables, but what's your vision for how we can run these parts separately?

mrice32 (commented):
Looks good! A few high-level architectural questions.

packages/api/src/apps/api/README.md
daywiss: The registry service handles adding emps from registry events, if there are new types of contracts, i would just add another service which specializes in those contracts, however they are discovered.  I might need to be sold a bit on the factory approach, i dont immediately see the value of it, but it sounds like something suited more in the SDK.

for one offs or custom addresses, this could be handled through priveleged calls in the api, which adds addresses to a table, or through envs/configs, CMS or git repo. what ever makes the most sense.

packages/api/src/apps/api/app.ts
daywiss: This particular file is like the top level composition of all the lego blocks (services, tables, timers).  Each service right now has a fairly unopinionated `update` function which can be called by anything, so you could add private http calls to update, add a new app which just calls update and exits, or change the interval on which they are called.  Once we have a shared data store, we can start to get more sophisticated with the design.

packages/api/src/apps/api/app.ts
daywiss: i think this lends itself quite well to a multi process shared db architecture, with some minor updates. Assume the "appState" is based all on shared DB tables, each service only depends on these tables for state, some write to them, others read. This means that you can break out each service into its own application, inject those same tables and thing should behave identically to the monolith.

mrice32: Totally agree that the lower-level components are agnostic. I was more talking about this file. Definitely something to worry about in a later PR, but it might be nice to have a way to compose the trigger, update process, and query handler so you can mix and match since those are things we probably want to vary across different environments. 

This could be especially cool for spinning up a local instance to test against in dapps. The dapps could even spin something like this up in process!

packages/api/src/apps/api/app.ts
mrice32: Yeah, totally agree on your point about the services being unopinionated.

I guess my main point is that this file seems to couple some pretty different pieces of functionality: the way the update is triggered, the way those updates are stored, and the way that the store is queried.

mrice32 (approved):
LGTM! I think we can talk more about the architecture stuff in later PRs.

packages/api/src/apps/api/app.ts
daywiss: cool, just to touch on this point and your other comment, i see the app files as bespoke, ie they are opinioned/custom compositions of lower level components. There's a couple schools of though on adding further features, one is to add switches within the app component with configuration through params (envs) to enable/disable features.  The other would be to create a completely new app for each type of feature. In practice i tend to lean towards the latter approach, though this depends on what needs to be added.

For example, we may want a serverless style of ingestion to db. Rather than have a serverless mode in the existing app, just make a new app which does this and call it `serverless-updater` or something.  This is mainly just a style thing but there are benefits as i find its less risky than changing working code, and the separation keep code simpler in the long run.

mrice32: (These comments are not blockers for merge, so please feel free to merge whenever!)

Agreed, you could definitely build multiple apps or have one app that was customized. I assumed you preferred the latter since this one is named app.ts (implying that there's only one).

The main point I'm making is that it seems like there's less modularity at this level than at the lower levels, but this is a level that we know we want to generalize. There's some distinct code in here that's clearly a class of "ingester" that handles grabbing new data. There's another set of distinct code that is involved in serving data using actions. Lastly, there seems to be a global state object initialized in-line and is a storage layer for state.

These all feel like distinct responsibilities to me, but I don't have super strong opinions. I just imagine we'll end up rewriting much of this code in other apps.

mrice32: (These comments are not blockers for merge, so please feel free to merge whenever!)

Agreed, you could definitely build multiple apps or have one app that was customized. I assumed you preferred the latter since this one is named app.ts (implying that there's only one).

The main point I'm making is that it seems like there's less modularity at this level than at the lower levels, but this is a level that we know we want to generalize. There's some distinct code in here that's clearly a class of "ingester" that handles grabbing new data. There's another set of distinct code that is involved in serving data using actions. Lastly, there seems to be a global state object initialized in-line and is a storage layer for state.

These all feel like distinct responsibilities to me, but I don't have super strong opinions. I just imagine we'll end up rewriting much of this code in other apps.

mrice32: (These comments are not blockers for merge, so please feel free to merge whenever!)

Agreed, you could definitely build multiple apps or have one app that was customized. I assumed you preferred the latter since this one is named app.ts (implying that there's only one).

The main point I'm making is that it seems like there's less modularity at this level than at the lower levels, but this is a level that we know we want to generalize. There's some distinct code in here that's clearly a class of "ingester" that handles grabbing new data. There's another set of distinct code that is involved in serving data using actions. Lastly, there seems to be a global state object initialized in-line and is a storage layer for state.

These all feel like distinct responsibilities to me, but I don't have super strong opinions. I just imagine we'll end up rewriting much of this code in other apps.

mrice32 (commented):
>

packages/api/src/apps/api/app.ts
daywiss: removed these to avoid confusion

PR #3091, opened 2021-06-11
feat(api): add ability to obtain a time series of tvl 

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1301/ability-to-obtain-a-time-series-of-tvl-of-all-contracts-on-uma-and-for-each-contract

**Summary**

Adds historical emp stats (tvl, and eventually others), and allows you to query between timestamps.

**Details**

adds the route `getEmpStatsBetween(address: string, start = 0, end: number = Date.now(), currency: CurrencySymbol = "usd")`

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1301/ability-to-obtain-a-time-series-of-tvl-of-all-contracts-on-uma-and-for-each-contract

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -85,9 +85,9 @@ Returns all known latest prices for collateral addresses. Synthetic prices not y
 
 Returns latest price for a particular collateral address.
 
-### historicalPricesByAddress(address:string,start:number=0,end:number=Date.now(),currency:"usd"="usd") => PriceSample[]
+### historicalPricesByAddress(address:string,start:number=0,end:number=Math.floor(Date.now() / 1000),currency:"usd"="usd") => PriceSample[]
 
-Get a range of historical prices between start and end in MS, sorted by timestamp ascending from an erc20 token address.
+Get a range of historical prices between start and end in seconds, sorted by timestamp ascending from an erc20 token address.
 This is useful for charting betwen two dates.
 
 ### sliceHistoricalPricesByAddress(address:string,start:number=0,length:number=1,currency:"usd"="usd") => PriceSample[]
@@ -121,14 +121,14 @@ Get only tvl (and other future statistics) for a single emp.
 
 Get all tvl (and other future statistics) for all known emps.
 
-### historicalSynthPricesByAddress(empAddress:string,start:number=0,end:number=Date.now()) => PriceSample[]
+### historicalSynthPricesByAddress(empAddress:string,start:number=0,end:number=Math.floor(Date.now() / 1000)) => PriceSample[]
 
-Get a range of historical prices synthetic prices by emp address between start and end in MS, sorted by timestamp
+Get a range of historical prices synthetic prices by emp address between start and end in unix seconds, sorted by timestamp
 ascending. This is useful for charting betwen two dates. Note non synth historical prices are specified by erc20 address.
 
 ### sliceHistoricalSynthPricesByAddress(address:string,start:number=0,length:number=1) => PriceSample[]
 
-Get a range of historical synthetic prices by emp address between start and count of price samples, sorted by timestamp
+Get a range of historical synthetic prices by emp address between start (in unix seconds) and count of price samples, sorted by timestamp
 ascending. This is useful for paginating data when you have X elements per page, and you know the last element seen.
 Note non synth historical prices are specified by erc20 address.
 
@@ -136,6 +136,6 @@ Note non synth historical prices are specified by erc20 address.
 
 Returns tvl of all supplied addresses in USD 18 decimals. If no addresses supplied, returns TVL of all known contracts.
 
-### getEmpStatsBetween(address: string, start = 0, end: number = Date.now(), currency: CurrencySymbol = "usd") => StatData[]
+### getEmpStatsBetween(address: string, start = 0, end: number = Math.floor(Date.now() / 1000), currency: CurrencySymbol = "usd") => StatData[]
 
-Returns all stats between timestamps at the highest resolution for a particular emp address.
+Returns all stats between timestamps (in seconds) at the highest resolution for a particular emp address.
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -135,3 +135,7 @@ Note non synth historical prices are specified by erc20 address.
 ### tvl(addresses?: string[], currency: CurrencySymbol = "usd") => string
 
 Returns tvl of all supplied addresses in USD 18 decimals. If no addresses supplied, returns TVL of all known contracts.
+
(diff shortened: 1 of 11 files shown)


packages/api/src/apps/api/README.md
```diff
@@ -135,3 +135,7 @@ Note non synth historical prices are specified by erc20 address.
 ### tvl(addresses?: string[], currency: CurrencySymbol = "usd") => string
 
 Returns tvl of all supplied addresses in USD 18 decimals. If no addresses supplied, returns TVL of all known contracts.
+
+### getEmpStatsBetween(address: string, start = 0, end: number = Date.now(), currency: CurrencySymbol = "usd") => StatData[]
```
chrismaree, line 139: would having a way to define the granularity be useful? for example, you _could_ have time-series data every second for the last year. requesting this would be an extremely taxing query.  if you could specify the granularity you could group things into say 10 mins bucket. 

by default, you could configure it to try an always return `n` number of data points where the granularity is dynamic depending on the time range requested. for example, you could say the query always returns 100 data points irrespective of the date range if no `granularity value is sent along.

This is not something that needs to be added in this PR, just something I wanted to bring up/ask about.

chrismaree: As a second point, `Date.now()` is a milliseconds js timestamp. would it make sense to refactor this to take in a Unix timestamp? the rest of the repo works on Unix time so it might be more consistent to query things by seconds rather than milliseconds.

packages/api/src/services/emp-stats.ts
```diff
@@ -15,7 +18,42 @@ export default (config: Config, appState: Dependencies) => {
...
+  async function hasStatsHistory(address: string, timestamp: number) {
+    return getOrCreateHistoryTable(address).has(timestamp);
+  }
+  async function createStatHistory(address: string, data: EmpHistoryStat) {
+    return getOrCreateHistoryTable(address).create(data);
+  }
+
+  async function updateHistory(address: string) {
+    const stat = await getLatestStats(address);
+    assert(uma.utils.exists(stat.timestamp), "stats require timestamp");
```
chrismaree, line 41: a question for my understanding: say one of these `asserts` fails. how does this manifest to the caller of the API? do they get a `500` internal server error with this assert message? where is this logic processed?

packages/api/src/apps/api/README.md
daywiss: > would having a way to define the granularity be useful? for example, you could have time-series data every second for the last year. requesting this would be an extremely taxing query. if you could specify the granularity you could group things into say 10 mins bucket.

I can definitely see this being useful and necessary for certain types of charts.  Will talk to tom about adding this as a feature. 

> As a second point, Date.now() is a milliseconds js timestamp. would it make sense to refactor this to take in a Unix timestamp? the rest of the repo works on Unix time so it might be more consistent to query things by seconds rather than milliseconds.

Good catch, currently all timestamps are stored in seconds internally. Its a bug and im querying things in MS once it hits the backend. Will need to fix this. 

As for the end user, i'm not sure what the right answer is in S vs MS.  To me its more intuitive to make queries with MS since most date manipulation classes give you this naturally, but I'm open to using seconds, just feels like an extra hoop for someone to jump through to get this.

packages/api/src/services/emp-stats.ts
daywiss: so theres 2 categories of errors here
1. most services are responsible for ingestion, so errors in this particular service will just be logged to console.error and are not tied to client requests.  I am using the same pattern in pretty much every ingestion service for logging errors, [here](https://github.com/daywiss/protocol/blob/feature/ch1301/ability-to-obtain-a-time-series-of-tvl-of/packages/api/src/services/emp-stats.ts#L83)
2. the actions/express service are responsible for responding to client requests, any errors thrown here get sent back to the user with a status 500 and the error message.  You can see [here where express calls the actions,](https://github.com/daywiss/protocol/blob/feature/ch1301/ability-to-obtain-a-time-series-of-tvl-of/packages/api/src/services/express.ts#L32) catches it and passes it through.  Eventually its handled by [express error handler](https://github.com/daywiss/protocol/blob/feature/ch1301/ability-to-obtain-a-time-series-of-tvl-of/packages/api/src/services/express.ts#L44)

packages/api/src/apps/api/README.md
chrismaree: IMHO seconds is more natural but I do get your point that in js land everything is ms.

packages/api/src/services/emp-stats.ts
chrismaree: makes sence! at some point replacing the `console.log` with an emit handler might be useful so we can capture the logs in GCP stackdriver

chrismaree (approved):

CarelessDoge (approved):

packages/api/src/apps/api/README.md
daywiss: i've changed it to seconds and we will see how it goes

packages/api/src/services/emp-state.ts
```diff
@@ -155,9 +155,8 @@ export default (config: Config, appState: Dependencies) => {
   }
 
   async function update(startBlock?: number | "latest", endBlock?: number) {
-    await Promise.map(Array.from(await registeredEmps.values()), (address: string) =>
-      updateOne(address, startBlock, endBlock)
-    );
+    const addresses = Array.from(await registeredEmps.values());
+    await Promise.mapSeries(addresses, (address: string) => updateOne(address, startBlock, endBlock));
```
mrice32, line 159: Why do we run these in series rather than in parallel?

packages/api/src/services/emp-stats.ts
```diff
@@ -15,7 +18,42 @@ export default (config: Config, appState: Dependencies) => {
...
+
+  async function getLatestStats(address: string) {
+    return stats[currency].latest.getOrCreate(address);
+  }
+  function getOrCreateHistoryTable(address: string) {
+    let table = stats[currency].history[address];
+    if (table) return table;
+    table = empStatsHistory.SortedJsMap("Emp Stat History for " + address);
+    stats[currency].history[address] = table;
+    return table;
```
mrice32, line 30:
```suggestion
    if (!stats[currency].history[address]) stats[currency].history[address] = empStatsHistory.SortedJsMap("Emp Stat History for " + address);
    return stats[currency].history[address];
```

mrice32 (approved):
LGTM!

packages/api/src/services/emp-state.ts
daywiss: its a soft way to slow down requests, considering theres like 100 emps, i didnt want to blast infura. But maybe that isnt necessary

PR #3092, opened 2021-06-11
improve: Nit changes made after integration testing OptimisticOracle + Oracle Tunnel system on Mumbai <> Goerli

nicholaspai:

**Motivation**

- Updated `OptimisticOracle`, `OracleChildTunnel`, and `OracleRootTunnel` deployments to reflect latest ancillary data logic.
- fixed typo in comment in `OptimisticOracle.sol`
- added `stampAncillaryData` read only method to `OracleChildTunnel.sol` for convenience
- fixed optimistic oracle deployment script
- updated `packages/optimistic-oracle/index.js` run script comments with instructions on how to run bot on sidechain where `OracleChildTunnel` is deployed as the Oracle
- 
**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/polygon/OracleChildTunnel.sol
+++ b/packages/core/contracts/polygon/OracleChildTunnel.sol
@@ -104,6 +104,16 @@ contract OracleChildTunnel is OracleBaseTunnel, OracleAncillaryInterface, FxBase
     }
 
     /**
+     * @notice Generates stamped ancillary data in the format that it would be used in the case of a price request.
+     * @param ancillaryData ancillary data of the price being requested.
+     * @param requester sender of the initial price request.
+     * @return the stamped ancillary bytes.
+     */
+    function stampAncillaryData(bytes memory ancillaryData, address requester) public view returns (bytes memory) {
+        return _stampAncillaryData(ancillaryData, requester);
+    }
+
+    /**
      * @dev We don't handle specifically the case where `ancillaryData` is not already readily translateable in utf8.
      * For those cases, we assume that the client will be able to strip out the utf8-translateable part of the
      * ancillary data that this contract stamps.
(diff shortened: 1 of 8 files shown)


packages/core/contracts/polygon/OracleChildTunnel.sol
```diff
@@ -103,6 +103,16 @@ contract OracleChildTunnel is OracleBaseTunnel, OracleAncillaryInterface, FxBase
         return lookup.price;
     }
 
+    /**
+     * @notice Generates stamped ancillary data in the format that it would be used in the case of a price request.
+     * @param ancillaryData ancillary data of the price being requested.
+     * @param requester sender of the initial price request.
+     * @return the stamped ancillary bytes.
+     */
+    function stampAncillaryData(bytes memory ancillaryData, address requester) public view returns (bytes memory) {
```
nicholaspai, line 112: I don't think we need to audit this method since its view-only

mrice32: I agree, but if another contract were to depend on the validity of the output, it might be audit-worthy. Probably won't be too hard to audit, though since they'll already be reviewing _stampAncillaryData. In my opinion, we should just toss it over to them just to be safe.

packages/core/deploy/012_deploy_optimistic_oracle.js
```diff
@@ -8,7 +8,7 @@ const func = async function (hre) {
   // 2 hours.
   const defaultLiveness = 7200;
   const Finder = await deployments.get("Finder");
-  const Timer = (await deployments.getOrNull("Timer")) || ZERO_ADDRESS;
+  const Timer = (await deployments.getOrNull("Timer")) || { address: ZERO_ADDRESS };
```
mrice32, line 11: Ah, nice catch! Sorry about that

mrice32 (approved):
LGTM!

PR #3095, opened 2021-06-14
refactor(serverless-hub): Support spokes running on multiple chains

nicholaspai:
**Motivation**

This PR enables one hub to be used for Spokes running on different chains. Previously the assumption was that the hub and its spokes were running on the same chain.

**Summary**

The main complexity is ensuring that the block cache does not skip blocks between bots running on the same chain, which is why I need to run multiple passes of the `botConfig` list and precompute last seen blocks/latest head blocks before executing the spokes.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -361,7 +361,6 @@ async function _saveQueriedBlockNumber(configIdentifier, blockNumbersForChain) {
         await datastore.save(dataBlob); // Overwrites the entire entity
       } else if (hubConfig.saveQueriedBlock == "localStorage") {
         Object.keys(blockNumbersForChain).forEach((chainId) => {
-          // TODO: Is it OK if this key name changes when in `localStorage` mode?
           process.env[`lastQueriedBlockNumber-${chainId}-${configIdentifier}`] =
             blockNumbersForChain[chainId].latestBlockNumber;
         });
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -49,6 +49,8 @@ const defaultHubConfig = {
   rejectSpokeDelay: 120, // 2 min.
 };
 
+const DEFAULT_GCP_DATASTORE_CHAINID = 1; // Spoke is assumed to be running on Ethereum Mainnet by default.
+
 hub.post("/", async (req, res) => {
   try {
     logger.debug({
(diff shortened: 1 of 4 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -64,29 +66,98 @@ hub.post("/", async (req, res) => {
...
+      // Fetch last seen block for this chain:
+      let lastQueriedBlockNumber = await _getLastQueriedBlockNumber(req.body.configFile, chainId);
+
+      // Next, get the head block for the chosen chain, which we'll use to override the last queried block number
+      // stored in GCP at the end of this hub execution.
+      let latestBlockNumber;
+
+      // If bot wants to use a non-default chain, then it must also specify a `CUSTOM_NODE_URL`, which we'll use to
+      // fetch the head block for the chain.
+      const spokeCustomNodeUrl = configObject[botName]?.environmentVariables?.CUSTOM_NODE_URL;
```
nicholaspai, line 107: This bit of code makes a subtle assumption that the `CUSTOM_NODE_URL` and the `gcpDataStoreChainId` are matched. I don't love making this assumption, but I'm not sure how else to dynamically grab a node URL for the chosen chainId without using the bot's custom set node URL.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -269,23 +341,30 @@ const _fetchConfig = async (bucket, file) => {
...
-            blockNumber,
-          },
+          data: latestBlockNumbersForChain,
         };
-        await datastore.save(dataBlob); // Saves the entity
+        await datastore.save(dataBlob); // Overwrites the entire entity
       } else if (hubConfig.saveQueriedBlock == "localStorage") {
-        process.env[`lastQueriedBlockNumber-${configIdentifier}`] = blockNumber;
+        Object.keys(blockNumbersForChain).forEach((chainId) => {
+          // TODO: Is it OK if this key name changes when in `localStorage` mode?
```
chrismaree, line 364: should not matter.

chrismaree (approved):
This is pretty great code. we'll commentted, good implementation.

packages/serverless-orchestration/src/ServerlessHub.js
mrice32: Hmmm, why do we even need the `gcpDataStoreChainId`? Could we:
1. Check for a CUSTOM_NODE_URL -- if one doesn't exist use the hub's CUSTOM_NODE_URL.
2. Get the chain id for that node. (web3.eth.getChainId)
3. Get the block number for that node.
4. Store them.

If you want to avoid excess queries and overwriting, you could construct a local cache of node url to chain id and block number, but that's just an optimization.

Thoughts?

mrice32 (commented):
Looks great! One comment

nicholaspai:
h/t @mrice32 for the suggestion, removed the need for the `gcpChainId` to be stored in the bot config

chrismaree (approved):

mrice32 (approved):
LGTM!

PR #3101, opened 2021-06-15
fix(cw-pricefeed): Fix getCurrentPrice when twapLength and invertPrice is defined

smb2796:
**Motivation**

There is a bug in the CryptowatchPriceFeed `getCurrentPrice` method, where an incorrectly inverted value is returned if `twapLength` and `invertPrice` is defined.

This is because the price is being inverted twice. Once in `_computeTwap()` and once in `getCurrentPrice()`.

**Summary**

Changes `getCurrentPrice()` to only invert if twapLength is not defined.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

NA

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -67,7 +67,7 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
 
   getCurrentPrice() {
     if (!this.twapLength && this.invertPrice) {
-      // The price should only be inverted if invertPrice is true and twapLength is not defined./
+      // The price should only be inverted if invertPrice is true and twapLength is not defined.
       // If twapLength is defined and invertPrice is true, the price will be inverted in _computeTwap().
       return this._invertPriceSafely(this.currentPrice);
     } else {
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -66,7 +66,11 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
   }
 
   getCurrentPrice() {
-    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+    return this.twapLength
+      ? this.currentPrice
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -66,7 +66,11 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
   }
 
   getCurrentPrice() {
-    return this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice;
+    return this.twapLength
+      ? this.currentPrice
+      : this.invertPrice
+      ? this._invertPriceSafely(this.currentPrice)
```
nicholaspai, line 72: I think this is more readable if you put parantheses around the `(this.invertPrice ? this._invertPriceSafely(this.currentPrice) : this.currentPrice)` part!

nicholaspai (approved):
Great catch, one small comment on readability

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
smb2796: Hm I had this, but linting seems to disagree.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -66,7 +66,11 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
...
+      : this.currentPrice;
```
mrice32, line 73: I think nested ternaries are a little difficult to read. Do you think this would be a little easier to understand?
```js
if (!this.twapLength && this.invertPrice) {
  // Maybe add a comment here as to why we need to do this for this case?
  return this._invertPriceSafely(this.currentPrice);
} else {
  return this.currentPrice;
}
```
  
Thoughts? Do you agree with this logic?

mrice32 (commented):
Looks good! One minor comment

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
smb2796: I do agree with the logic and that this is more readable. Updated to use this instead!

PR #3107, opened 2021-06-15
feat(api): add synthetic prices per emp

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/920/create-api-to-get-current-prices-of-synths-and-total-number-of-synths-on-uma

**Summary**

Uses the synthetic price getter from the affiliates package, but modifies it to only get current prices. Latest prices are fetched every 5 minutes and served through an api endpoint

**Details**

Synthetic price fetching happens on a timer, every 5 minutes.
2 api routes were added to query for prices:
- allLatestSynthPrices()
- latestSynthPriceByAddress(empAddress)

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/api/src/libs/synthPrices.test.ts
```diff
@@ -0,0 +1,27 @@
+require("dotenv").config();
+import assert from "assert";
+import SynthPrices from "./synthPrices";
+import Web3 from "web3";
+
+// this is an integration test, dont run in ci
+describe.skip("synthPrices", function () {
```
mrice32, line 7: Is there a way to tag tests in mocha so that we can run them with a flag or something? Just curious. Otherwise, it'll get pretty tedious to try and run all the integration tests, so much so that it might make sense to isolate them in a particular location so we can run them with a particular command.

packages/api/src/libs/synthPrices.ts
```diff
@@ -0,0 +1,52 @@
...
+  defipulseApiKey?: string;
+  priceFeedDecimals?: number;
+};
+type Dependencies = {
+  web3: Web3;
+};
+// cannot get type of Pricefeed, would be great if we could
+type PriceFeed = any;
+
+export default (config: Config = {}, { web3 }: Dependencies) => {
```
mrice32, line 19: nit: what's a dependency?  Is it another name for Lib? Or is it like an import that we've added state to? How does it differ from the stuff in appState?

mrice32 (approved):
LGTM! A few minor comments.

packages/api/src/libs/synthPrices.test.ts
daywiss: good question, will have to look into that, but its a good idea to be able to run these independently

packages/api/src/libs/synthPrices.ts
daywiss: in my head i think of it like dependency injection, usually stateful classes, or things that could be mocked that are not just JSON. it might even be better to add the logger to the dependencies, not sure. I think i'll just simplify this to just web3.

packages/api/src/libs/synthPrices.test.ts
daywiss: ok so i've changed integration/e2e tests to be identified by filename, so `synthPrices.test.ts` is now `synthPrices.e2e.ts`.  `yarn test` will not run these, only files with `.test.ts`, `yarn e2e` will run any tests that are `.e2e.ts`.  the naming here is not the best, kind of lumping all non unit tests into `e2e`, but essentially tests which depend on external systems and could be non deterministic.

PR #3108, opened 2021-06-15
feat(api): api to get historical prices of synths

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/923/create-api-to-get-historical-prices-of-synths

**Summary**

This copies the pattern from previous historical collateral prices and applies it to synthetics.  it does this by copying over the latest price into a time series table which can queried from the api. 

**Details**
creates 2 new api routes, which are documented in the readme
- historicalSynthPricesByAddress
- sliceHistoricalSynthPricesByAddress

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/923/create-api-to-get-historical-prices-of-synths

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -44,6 +44,7 @@ async function run(env: ProcessEnv) {
     },
     synthPrices: {
       latest: {},
+      history: {},
     },
     erc20s: tables.erc20s.JsMap(),
     stats: {
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -43,6 +43,9 @@ export type AppState = {
     latest: {
       [empAddress: string]: PriceSample;
     };
+    history: {
+      [empAddress: string]: uma.tables.historicalPrices.SortedJsMap;
+    };
   };
   erc20s: uma.tables.erc20s.JsMap;
   stats: {
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -100,6 +100,26 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       assert(stats[currency], "No stats for currency: " + currency);
       return stats[currency].latest.values();
     },
+    async historicalSynthPricesByAddress(
+      empAddress: string,
+      start = 0,
+      end: number = Date.now()
+    ): Promise<PriceSample[]> {
+      assert(empAddress, "requires emp address");
+      assert(start >= 0, "requires a start value >= 0");
+      assert(exists(synthPrices.history[empAddress]), "No synthetic prices for emp address: " + empAddress);
+      const results = await synthPrices.history[empAddress].betweenByTimestamp(start, end);
+      // convert this to tuple to save bytes.
+      return results.map(({ price, timestamp }) => [timestamp, price]);
+    },
+    async sliceHistoricalSynthPricesByAddress(empAddress: string, start = 0, length = 1): Promise<PriceSample[]> {
+      assert(empAddress, "requires emp address");
+      assert(start >= 0, "requires a start value >= 0");
+      assert(exists(synthPrices.history[empAddress]), "No synthetic prices for emp address: " + empAddress);
+      const results = await synthPrices.history[empAddress].sliceByTimestamp(start, length);
+      // convert this to tuple to save bytes.
+      return results.map(({ price, timestamp }) => [timestamp, price]);
+    },
   };
 
   // list all available actions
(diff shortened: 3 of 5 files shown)


packages/api/src/services/actions.ts
```diff
@@ -89,6 +89,26 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
...
+    ): Promise<PriceSample[]> {
+      assert(empAddress, "requires emp address");
+      assert(start >= 0, "requires a start value >= 0");
+      assert(exists(synthPrices.history[empAddress]), "No synthetic prices for emp address: " + empAddress);
+      const results = await synthPrices.history[empAddress].betweenByTimestamp(start, end);
+      // convert this to tuple to save bytes.
+      return results.map(({ price, timestamp }) => [timestamp, price]);
+    },
+    async sliceHistoricalSynthPricesByAddress(empAddress: string, start = 0, length = 1): Promise<PriceSample[]> {
+      assert(empAddress, "requires emp address");
```
mrice32, line 105: What is this asserting? Isn't its existence guaranteed by the typechecker? Or are you specifically asserting that it's not an empty string?

packages/api/src/services/actions.ts
```diff
@@ -89,6 +89,26 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       // convert this to tuple to save bytes.
       return results.map(({ price, timestamp }) => [timestamp, price]);
     },
+    async historicalSynthPricesByAddress(
+      empAddress: string,
+      start = 0,
+      end: number = Date.now()
+    ): Promise<PriceSample[]> {
+      assert(empAddress, "requires emp address");
```
mrice32, line 97: Same question.

packages/api/src/index.ts
```diff
@@ -40,6 +40,9 @@ export type AppState = {
     latest: {
       [empAddress: string]: PriceSample;
     };
+    history: {
+      [empAddress: string]: uma.tables.historicalPrices.SortedJsMap;
```
mrice32, line 44: I just realized that I'm unfamiliar with this syntax.

What's the difference between:
```ts
history: {
  [empAddress: string]: uma.tables.historicalPrices.SortedJsMap;
}
```

and 

```ts
history: {
  [key: string]: uma.tables.historicalPrices.SortedJsMap;
}
```

Are `key` and `empAddress` just annotations essentially?

mrice32 (commented):
Looks awesome! A few really minor questions/comments

packages/api/src/services/actions.ts
daywiss: all the action calls can be incorrectly called since these are exposed by express and called by a third party, so more checks need to be added here to give the client feedback.

packages/api/src/index.ts
daywiss: yea i think these are just labels, ive called it empAddress for clarity!

mrice32: That's cool! We should do this everywhere!

packages/api/src/services/actions.ts
mrice32: I see, so this _is_ the API. That makes sense!

One question on that note: it seems like the caller needs to specify these parameters in the correct order in the body of their request? Would it be less error-prone to use named params and just pass the body object directly?

```ts 
async sliceHistoricalSynthPricesByAddress({ empAddress: string, start = 0, length = 1 }): Promise<PriceSample[]> {
```

We don't have to do this now, but I think it would be a good quality of life improvement for the API users. Thoughts?

mrice32 (approved):
LGTM!

packages/api/src/services/actions.ts
daywiss: absolutely could, the main tradeoff with it is slightly more verbose calls on the client side and more data over the wire, but it may allow a more intuitive api.  in the past i have done the calls in the way you suggest, so i am open to that.

packages/api/src/apps/api/app.ts
```diff
@@ -44,6 +44,7 @@ async function run(env: ProcessEnv) {
     },
     synthPrices: {
       latest: {},
+      history: {},
```
chrismaree, line 47: what happens if the API is restarted? the history will be cleared out, right? how does the path to adding in a DB work with this exactly? I know you've probs explained it before in a previous PR but I'd just like to understand a bit better.

daywiss: Currently everything is in memory, so it will lose history. The history table here is set up to use what im calling a "store", which can be implemented with a google datastore ( currently using in memory store). This shouldn't change any of the logic in the api package, as the store conforms to a fixed interface.  So the idea is that we can implement a new store and gain persistence without having to change much else with business logic.

PR #3112, opened 2021-06-16
fix(serverlesshub): reduce logging noise

chrismaree:
**Motivation**

The most recent changes to the serverless hub made the logs **very** noisy. This PR reduces the noise by moving where logs are produced and opting for a more aggregate logging approach.

Before this PR: 
image

After this PR the logs are far more redacted: 
image

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -121,8 +121,8 @@ hub.post("/", async (req, res) => {
 
       // Store block number data for this chain ID which we'll use to update the GCP cache later.
       blockNumbersForChain[chainId] = {
-        lastQueriedBlockNumber,
-        latestBlockNumber,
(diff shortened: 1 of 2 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -121,17 +121,16 @@ hub.post("/", async (req, res) => {
 
       // Store block number data for this chain ID which we'll use to update the GCP cache later.
       blockNumbersForChain[chainId] = {
-        lastQueriedBlockNumber,
-        latestBlockNumber,
+        lastQueriedBlockNumber: Number(lastQueriedBlockNumber),
```
chrismaree, line 124: this type was sometimes a string when I tested it. thought it best to ensure it's a number. @nicholaspai let me know if you have a problem with this

nicholaspai (approved):
LGTM this is a nice consolidation

PR #3118, opened 2021-06-16
feat(core): add custom fee set by deployer for LSP

mrice32:
**Motivation**

We want deployers of the LSP contract to be able to incentivize price proposers.

**Summary**

Add a customizable proposal reward that deployers can preload into the contract.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
+++ b/packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
@@ -110,6 +110,7 @@ contract ContractForDifference is Testable, Lockable {
      * @param _financialProductLibrary Contract providing settlement payout logic.
      * @param _customAncillaryData Custom ancillary data to be passed along with the price request. If not needed, this
      *                             should be left as a 0-length bytes array.
+     * @param _prepaidProposerReward Preloaded reward to incentivize settlement price proposals.
      * @param _timerAddress Contract that stores the current time in a testing environment. Set to 0x0 in production.
      */
     constructor(
--- a/packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
+++ b/packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
@@ -63,6 +63,8 @@ contract ContractForDifference is Testable, Lockable {
 
     bytes public customAncillaryData;
 
+    uint256 prepaidProposerReward;
+
     /****************************************
      *                EVENTS                *
      ****************************************/
@@ -120,6 +122,7 @@ contract ContractForDifference is Testable, Lockable {
         FinderInterface _finder,
         ContractForDifferenceFinancialProductLibrary _financialProductLibrary,
         bytes memory _customAncillaryData,
+        uint256 _prepaidProposerReward,
         address _timerAddress
     ) Testable(_timerAddress) {
         finder = _finder;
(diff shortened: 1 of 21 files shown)


packages/core/contracts/financial-templates/contract-for-difference/ContractForDiffrenceCreator.sol
```diff
@@ -99,9 +104,14 @@ contract ContractForDifferenceCreator is Testable, Lockable {
                 finder,
                 financialProductLibrary,
                 customAncillaryData,
+                prepaidProposerReward,
                 timerAddress
             );
 
+        // Move prepaid proposer reward from the deployer to the newly deployed contract.
+        if (prepaidProposerReward > 0)
+            collateralToken.safeTransferFrom(msg.sender, address(cfd), prepaidProposerReward);
```
chrismaree, line 113: nit: you might as well put this below the typecasting of the `address(cfd)` to avoid the extra casting.

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
```diff
@@ -145,6 +149,7 @@ contract ContractForDifference is Testable, Lockable {
             "Ancillary Data too long"
         );
         customAncillaryData = _customAncillaryData;
+        prepaidProposerReward = _prepaidProposerReward;
```
chrismaree, line 152: if this contract is created directly (not via the factory) it is possible to make settlement revert if the LSP does not have the balance of the reward to pay on the `_requestOraclePriceExpiration()` method.

I don't really think this is an issue because:
1) people will use the factor almost always
2) the contract is not broken in this case as an external EOA can always send enough tokens to the contract to let settlement continue.

chrismaree (approved):
LGTM!

packages/core/contracts/financial-templates/contract-for-difference/ContractForDifference.sol
mrice32: Yes, yes, and yes :).

Those were my exact thoughts on it. I thought about trying to predict the address and pre-approve the tokens, but the code to predict deployment addresses in solidity is complex. And people should always use the factory. We have a similar issue if someone were to create an EMP without a factory. it would look fine, but when a dispute gets sent, it fails. I think we can make the same assumption here.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -276,8 +281,15 @@ contract LongShortPair is Testable, Lockable {
     function _requestOraclePriceExpiration() internal {
         OptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
 
-        // For now, we add no fees the the OO and set the reward to 0.
-        optimisticOracle.requestPrice(priceIdentifier, expirationTimestamp, customAncillaryData, collateralToken, 0);
+        // Use the prepaidProposerReward as the proposer reward.
+        collateralToken.safeApprove(address(optimisticOracle), prepaidProposerReward);
```
nicholaspai, line 285: Is there any potential savings in checking `prepaidProposerReward > 0` and if not, then no need to call `safeApprove`?

mrice32: Good point! Fixed.

PR #3119, opened 2021-06-17
improve(common): Upgrade hardhat packages

nicholaspai:
**Motivation**

Packages updated with notes on specific features we want included:

- `@nomiclabs/hardhat-etherscan`: [Adds support for verifying polygonscan contracts](https://github.com/nomiclabs/hardhat/releases/tag/hardhat-etherscan-v2.1.3)
- `hardhat`: [Adds Solidity v8 + Berlin support](https://github.com/nomiclabs/hardhat/releases)
- `hardhat-deploy`
- `hardhat-gas-reporter`

**Testing**

- [ ]  New unit tests created
- [X]  Untested


chrismaree (approved):
nice and simple.

mrice32 (approved):
Sweeeet! If this works, this should kill all of our verification issues!

nicholaspai:
> Sweeeet! If this works, this should kill all of our verification issues!

Not mumbai unfortunately

packages/api/src/services/express.ts
```diff
@@ -20,8 +20,6 @@ export default async (config: Config, actions: Actions) => {
   app.use(bodyParser.json({ limit: "1mb" }));
   app.use(bodyParser.urlencoded({ extended: true }));
 
-  app.options("*", cors());
```
nicholaspai, line 23: @daywiss I had to fix this to fix a TS compile error similar to [this issue](https://stackoverflow.com/questions/59167258/no-overload-matches-this-call-in-cors-typescript).

The solution was found when someone pointed out the correct way to use `cors` is like so.

[cors docs](https://expressjs.com/en/resources/middleware/cors.html)

packages/bot-strategy-runner/test/ConfigBuilder.ts
```diff
@@ -35,7 +35,7 @@ const minimumRequiredConfigKeys = [
   "pollingDelay",
 ];
 describe("buildGlobalWhitelist", async function () {
-  it("Correctly fetches global whitelist", async function () {
+  it("Correctly fetches global whitelist", async function (this: any) {
```
nicholaspai, line 38: @chrismaree not sure why this tsc error wasn't caught before but i needed to fix this

package.json
```diff
@@ -44,6 +44,7 @@
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
     "ganache-cli": "^6.12.2",
+    "hardhat": "^2.4.0",
```
nicholaspai, line 47: Moving this to here as I realized the `protocol/node_modules/.bin/hardhat` was pointing to `protocol/packages/cloud-functions/node_modules/.bin/hardhat` which was causing an error when trying to run `hardhat x` commands from within `core`

packages/api/src/services/express.ts
daywiss: Thanks nick!

daywiss (approved):
Just reviewed the cors change, it matches the original intent, thanks

packages/core/test/merkle-distributor/MerkleDistributor.js
```diff
@@ -251,7 +251,7 @@ contract("MerkleDistributor.js", function (accounts) {
           },
           { from: rando }
         );
-        assertApproximate(87749, claimTx.receipt.gasUsed);
+        assertApproximate(92049, claimTx.receipt.gasUsed);
```
mrice32, line 254: This is a bit odd -- why do you think these changed so much with a change in _hardhat_ version (not compiler version)?

packages/common/package.json
```diff
@@ -45,9 +45,8 @@
     "chalk-pipe": "^3.0.0",
     "dotenv": "^6.2.0",
     "eth-crypto": "^1.7.0",
-    "hardhat": "^2.0.4",
```
mrice32, line 48: I think we need to keep hardhat as a production dependency since it's required by files that are exported by the package. Otherwise, anyone who installs the package wouldn't automatically install hardhat and any `require("hardhat")` calls would fail on import.

mrice32 (commented):
Looks good. Two minor comments.

packages/core/test/merkle-distributor/MerkleDistributor.js
nicholaspai: `hardhat@2.2` introduced support for `Berlin` as the default `hardfork`!

packages/common/package.json
nicholaspai: hmmm, do you think that's why some tests are failing? I'll try that

packages/bot-strategy-runner/test/ConfigBuilder.ts
chrismaree: that's strange...not sure why. it must be related to your package change?

chrismaree (approved):

packages/common/package.json
nicholaspai: Looks like a more recent `hardhat` change forces us to specify `transactionConfig.chainId` as a hex number instead of an int. Trying to find the PR that changed this

PR #3120, opened 2021-06-17
feat(hardhat-deploy-scripts): add hardhat deploy scripts for LSP on test nets

chrismaree:
**Motivation**

In setting up the LSP on testnets we require some scripts to deploy the LSP and the associated libraries.

**Summary**

Two deployment scripts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/core/deploy/027_deploy_long_short_pair_creator.js
```diff
@@ -0,0 +1,20 @@
...
+
+  const { deployer } = await getNamedAccounts();
+
+  const Finder = await deployments.get("Finder");
+  const TokenFactory = await deployments.get("TokenFactory");
+  const Timer = await deployments.getOrNull("Timer");
+
+  await deploy("LongShortPairCreator", {
+    from: deployer,
+    args: [Finder.address, TokenFactory.address, Timer.address || ZERO_ADDRESS],
```
mrice32, line 14: Won't this break?

If `Timer` is null, won't we get an error when trying to access `null.address`?

```
> null.address
Uncaught TypeError: Cannot read property 'address' of null
```

If you look in some of the other files, we do it slightly differently to avoid this issue:
```js
const Timer = (await deployments.getOrNull("Timer")) || { address: ZERO_ADDRESS };

...
args: [Finder.address, TokenFactory.address, Timer.address],
...
```

packages/core/deploy/028_deploy_long_shor_pair_library.js
```diff
@@ -0,0 +1,13 @@
+const func = async function (hre) {
+  const { deployments, getNamedAccounts } = hre;
+  const { deploy } = deployments;
+
+  const { deployer } = await getNamedAccounts();
+
+  await deploy("BinaryOptionLongShortPairFinancialProductLibrary", { from: deployer, log: true });
+  await deploy("LinearLongShortPairFinancialProductLibrary", { from: deployer, log: true });
+  await deploy("CoveredCallLongShortPairFinancialProductLibrary", { from: deployer, log: true });
+  await deploy("RangeBondLongShortPairFinancialProductLibrary", { from: deployer, log: true });
```
mrice32, line 10: Just out of curiosity, why did we decide to put these all in the same file?

I only ask because we generally have contracts 1:1 with deployment files so that all contracts can be tagged and deployed independently of one another. I'm not strongly opposed to putting these in the same file for brevity, but it also might be nice to replace one or two of these without having to deploy the entire bunch.

mrice32 (commented):
Looks good! Two comments/questions

packages/core/deploy/027_deploy_long_short_pair_creator.js
chrismaree: hhmm. the code example you put there would work fine as you pulling out the address. the code I had won't break per say but it is just undefined and then defaults to the `ZERO_ADDRESS`.

I have changed it to map to the implementation in the other scripts like you described.

packages/core/deploy/028_deploy_long_shor_pair_library.js
chrismaree: main logic was just to deploy them all onto the chain the deployment script is run on.

I do totally see what you are saying about not always needing all of them. that is very true. we could split this file up into multiple `func` that have each lib? Not sure what structure would work best given we need all these libs on mainnet/testnets.

nicholaspai: I'm good with deploying these all within the same network. Two nits:

- this file name has a typo, missing "t" in "shor(t)".
- We should rename file to be "libraries" instead of "library" to make it more obvious what its doing

chrismaree: thanks nick. have fixed both points!

mrice32 (commented):
Only remaining comment is [this](https://github.com/UMAprotocol/protocol/pull/3120#discussion_r655604586)!

packages/core/deploy/028_deploy_long_shor_pair_library.js
chrismaree: I see your point. one contract per file would create 3 additional files, which might be ok but it seems a bit verbose. Otherwise, we could use some flags in the script where it by default will deploy all of them or you could add a param to specify how many you/which ones you want to deploy? however, I know that hardhat does not like unrecognized arguments in the command line. wdyt?

mrice32: There may be ways to organize these deployment files into directories, but overall, I think file bloat isn't a big deal if it makes the deployment experience better.

I agree with your analysis of params -- hardhat doesn't let you add them outside of a task (I think). I also think that params in individual files mean that people have to dig to figure out how to do what they want to do.

Overall, even with bloat, I would argue that 1 file per contract is the best setup for the user.

Thoughts?

mrice32 (commented):
https://github.com/UMAprotocol/protocol/pull/3120#discussion_r657983946

nicholaspai:
@chrismaree before merging, these file names need to be updated given that the latest `deploy` script now uses the prefix `029_x`

packages/core/deploy/028_deploy_long_shor_pair_library.js
chrismaree: I have updated accordingly. LMK what you think

nicholaspai (approved):
LGTM

PR #3121, opened 2021-06-17
fix(common): explicitly require the hardhat etherscan verification plugin

mrice32:
**Motivation**

The [hardhat docs](https://hardhat.org/plugins/nomiclabs-hardhat-etherscan.html) say you need to require the plugin in the config. We don't currently do this in our config.

Right now: `hardhat etherscan-verify` seems to work. However, this command seems to be slightly different from `hardhat verify`, which doesn't work unless the plugin is added to the config.

Not sure why this is this is the case or what the deep differences are, but we should do as recommended and put it in the config.

**Summary**

Added the require to the config.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/common/src/HardhatConfig.js
+++ b/packages/common/src/HardhatConfig.js
@@ -8,6 +8,7 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
   require("hardhat-gas-reporter");
   require("@nomiclabs/hardhat-web3");
   require("hardhat-deploy");
+  require("@nomiclabs/hardhat-etherscan");
 
   // Custom tasks to interact conveniently with smart contracts.
   require("./hardhat/tasks");


nicholaspai:
So `@nomiclabs/hardhat-etherscan` ships as a plugin from the `hardhat` team, while `etherscan-verify` is a plugin that is added with the `hardhat-deploy` package. So this PR will correctly add the `hardhat verify` task

nicholaspai (approved):

mrice32:
> So `@nomiclabs/hardhat-etherscan` ships as a plugin from the `hardhat` team, while `etherscan-verify` is a plugin that is added with the `hardhat-deploy` package. So this PR will correctly add the `hardhat verify` task
> 
> [hardhat-deploy](https://github.com/wighawag/hardhat-deploy#4-hardhat-etherscan-verify)

Ah, very helpful. Thanks for clarifying. So `etherscan-verify` will work for most contracts, but we'll need `verify` for contracts that are deployed by factories since you can manually pass the address and constructor params.

PR #3122, opened 2021-06-17
feat(api): 1405 add tvm calculations per emp 

daywiss:
**Motivation**
https://app.clubhouse.io/uma-project/story/1405/obtain-the-tvm-of-each-contract-on-uma
**Summary**

This adds tvm calculation which are very similar to tvl calculations, and updates the emp stats with this new data.

image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1405/obtain-the-tvm-of-each-contract-on-uma

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -149,3 +149,7 @@ Returns tvl of all supplied addresses in USD 18 decimals. If no addresses suppli
 ### getEmpStatsBetween(address: string, start = 0, end: number = Math.floor(Date.now() / 1000), currency: CurrencySymbol = "usd") => StatData[]
 
 Returns all stats between timestamps (in seconds) at the highest resolution for a particular emp address.
+
+### sliceHistoricalEmpStats(empAddress: string, start = 0, length = 1, currency:CurrencySymbol='usd') => StatData[]
+
+Returns all stats for emp starting at timestamp in seconds and a length number of samples after that ascending.
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -104,10 +104,15 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       if (addresses == null || addresses.length == 0) return queries.totalTvl(currency);
       return queries.sumTvl(addresses, currency);
     },
-    async getEmpStatsBetween(address: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
+    async getEmpStatsBetween(empAddress: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
       assert(stats[currency], "Invalid currency type: " + currency);
-      assert(stats[currency].history[address], "Invalid emp address: " + address);
-      return stats[currency].history[address].between(start, end);
+      assert(stats[currency].history[empAddress], "Invalid emp address: " + empAddress);
+      return stats[currency].history[empAddress].between(start, end);
+    },
+    async sliceHistoricalEmpStats(empAddress: string, start = 0, length = 1, currency: CurrencySymbol = "usd") {
+      assert(stats[currency], "Invalid currency type: " + currency);
+      assert(stats[currency].history[empAddress], "Invalid emp address: " + empAddress);
+      return stats[currency].history[empAddress].slice(start, length);
     },
   };
 
--- a/packages/api/src/services/actions.ts
+++ b/packages/api/src/services/actions.ts
@@ -73,6 +73,7 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     },
     async sliceHistoricalSynthPrices(empAddress: string, start = 0, length = 1): Promise<PriceSample[]> {
       assert(empAddress, "requires an empAddress");
+      assert(length < 1000, "length must be less than 1000 samples");
       const emp = await queries.getAnyEmp(empAddress);
       assert(exists(emp.tokenCurrency), "EMP does not have token currency address");
       return queries.sliceHistoricalPricesByTokenAddress(emp.tokenCurrency, start, length);
@@ -85,6 +86,7 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     },
     async sliceHistoricalCollateralPrices(empAddress: string, start = 0, length = 1): Promise<PriceSample[]> {
       assert(empAddress, "requires an empAddress");
+      assert(length < 1000, "length must be less than 1000 samples");
       const emp = await queries.getAnyEmp(empAddress);
       assert(exists(emp.collateralCurrency), "EMP does not have token currency address");
       return queries.sliceHistoricalPricesByTokenAddress(emp.collateralCurrency, start, length);
@@ -112,6 +114,7 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     async sliceHistoricalEmpStats(empAddress: string, start = 0, length = 1, currency: CurrencySymbol = "usd") {
       assert(stats[currency], "Invalid currency type: " + currency);
       assert(stats[currency].history[empAddress], "Invalid emp address: " + empAddress);
+      assert(length < 1000, "length must be less than 1000 samples");
       return stats[currency].history[empAddress].slice(start, length);
     },
   };
--- a/packages/api/src/services/emp-stats.ts
+++ b/packages/api/src/services/emp-stats.ts
@@ -50,9 +50,7 @@ export default (config: Config, appState: Dependencies) => {
     // PriceSample is type [ timestamp:number, price:string]
     const priceSample: PriceSample = synthPrices.latest[empAddress];
     assert(uma.utils.exists(priceSample), "No latest synthetic price found for emp: " + empAddress);
-
-    const price = priceSample[1];
-    assert(uma.utils.exists(price), "Invalid latest sythetic price found on emp: " + empAddress);
+    const [, price] = priceSample;
     return price;
   }
 
@@ -61,7 +59,6 @@ export default (config: Config, appState: Dependencies) => {
     const priceSample: PriceSample = prices[currency].latest[currencyAddress];
     assert(uma.utils.exists(priceSample), "No latest price found for emp: " + empAddress);
     const [, price] = priceSample;
-    assert(uma.utils.exists(price), "Invalid latest price found on emp: " + empAddress);
     return price;
   }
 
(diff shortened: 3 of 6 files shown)


packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
+    const emp = await getFullEmpState(empAddress);
+    update.timestamp = nowS();
+
+    if (uma.utils.exists(emp.collateralCurrency)) {
+      update.collateralPrice = await getPriceFromTable(empAddress, emp.collateralCurrency).catch(() => "0");
+      update.tvl = calcTvl(update.collateralPrice, emp).toString();
+    }
+
+    if (uma.utils.exists(emp.tokenCurrency)) {
+      update.syntheticPrice = await getPriceFromTable(empAddress, emp.tokenCurrency).catch(() => "0");
```
mrice32, line 85: Instead of providing 0 for prices we don't have, should we have a more well-defined `null` or `undefined` value that can go into our DB? Or did you consider just not updating these if they fail?

Totally not something that's worth a major refactor -- this just might be ambiguous for things where the price actually is zero (expired, worthless option, for instance), but we may never hit that since a lot of exchanges don't allow a price of 0.

packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
+    // the full state has collateral decimals, pulled from erc20 state
+    return queries.getFullEmpState(emp);
+  }
+  async function updateStats(empAddress: string) {
+    const update: empStats.Data = { address: empAddress };
+
+    const emp = await getFullEmpState(empAddress);
+    update.timestamp = nowS();
+
+    if (uma.utils.exists(emp.collateralCurrency)) {
```
mrice32, line 79: When would this be false? Same question for tokenCurrency below? Unless I'm missing something, that would signal that something's broken?

mrice32 (commented):
Looks good! A few minor comments

packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
-      timestamp: stat.timestamp,
-    };
-    return createStatHistory(address, update);
+  async function updateHistory(empAddress: string) {
+    const stat = await getLatestStats(empAddress);
+    assert(uma.utils.exists(stat.timestamp), "stats require timestamp for emp: " + empAddress);
+    if (await hasStatsHistory(empAddress, stat.timestamp)) return stat;
+    // remove id from the latest stats, since this gets overriden with history stats by timestamp
+    const { id, ...historyStat } = stat;
+    return createStatHistory(empAddress, historyStat as EmpHistoryStat);
```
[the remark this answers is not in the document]
daywiss: this is really tricky, had issues getting this to work. Heres what lead to this: 
There are 2 data structures, one that represents the latest stats, the other represents the historical stat. Timestamp is optional on the first, but required on the second ( address is key in the latest table but timestamp is the id in historical table).  For some reason, typescript doesnt care that i validate timestamp exists, it still complains that its the wrong type, so basically i have to validate and then cast, telling typescript its now the right type.  Heres an example of it failing:

image

somehow it still thinks timestamp can be undefined:
image

packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
 
+  async function getPriceFromTable(empAddress: string, currencyAddress: string) {
     // PriceSample is type [ timestamp:number, price:string]
-    const priceSample: PriceSample = prices[currency].latest[fullState.collateralCurrency];
-    assert(uma.utils.exists(priceSample), "No latest price found for emp: " + address);
-    const [timestamp, price] = priceSample;
-    assert(uma.utils.exists(price), "Invalid latest price found on emp: " + address);
+    const priceSample: PriceSample = prices[currency].latest[currencyAddress];
+    assert(uma.utils.exists(priceSample), "No latest price found for emp: " + empAddress);
+    const [, price] = priceSample;
```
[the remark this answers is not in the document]
daywiss: hm i can make that consistent

packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
+    assert(uma.utils.exists(price), "Invalid latest price found on emp: " + empAddress);
```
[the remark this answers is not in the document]
daywiss: i dont think i can the check is probably redundant, will remove

packages/api/src/services/emp-stats.ts
daywiss: i really only do it so i can just fall through to the tvl/tvm calculation without an extra check or causing errors. Actually I realized working on future PRs that a lot of this logic needs to be refactored, but that is another issue. 

I'm ok with leaving things null, we will just have to be more careful with dealing with that both on backend and frontend calls.

daywiss: i think the type currently accepts undefined, so i can test setting these to undefined

packages/api/src/services/emp-stats.ts
daywiss: theres 2 ways i can think of this is undefined. 

1. The the state for emps is not consistent, its possible to have state come  in over time, so if this happens to run before an emp is fully fetched, it could be undefined. This is not super likely because we wait for emp state to fetch before this is ever run, but you never know. 
2. More likely is that currently the system cant differentiate between non emp contracts deployed through the registry. It assumes all registry contract events are EMPs, but actually some of these contracts are like governance contracts, or even early test emp contracts which dont have this collateral address.  So this is iterating over some contract addresses that are not actually EMPs, which cause problems. 

Dont' really have a good solution for #2 without someway to whitelist/blacklist addresses or filter them based on some kind of criteria

packages/api/src/services/emp-stats.ts
```diff
@@ -34,51 +33,76 @@ export default (config: Config, appState: Dependencies) => {
...
-    const fullState = await queries.getFullEmpState(emp);
+  // synthetic prices return price in wei, due to the way the bot price feeds work
+  async function getSyntheticPriceFromTable(empAddress: string) {
+    // PriceSample is type [ timestamp:number, price:string]
+    const priceSample: PriceSample = synthPrices.latest[empAddress];
+    assert(uma.utils.exists(priceSample), "No latest synthetic price found for emp: " + empAddress);
 
-    assert(uma.utils.exists(fullState.collateralCurrency), "Emp requires collateralCurrency: " + address);
+    const price = priceSample[1];
+    assert(uma.utils.exists(price), "Invalid latest sythetic price found on emp: " + empAddress);
```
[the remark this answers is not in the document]
daywiss: will remove

packages/api/src/apps/api/README.md
```diff
@@ -149,3 +149,7 @@ Returns tvl of all supplied addresses in USD 18 decimals. If no addresses suppli
 ### getEmpStatsBetween(address: string, start = 0, end: number = Math.floor(Date.now() / 1000), currency: CurrencySymbol = "usd") => StatData[]
 
 Returns all stats between timestamps (in seconds) at the highest resolution for a particular emp address.
+
+### sliceHistoricalEmpStats(empAddress: string, start = 0, length = 1, currency:CurrencySymbol='usd') => StatData[]
```
chrismaree, line 153: this is pretty awesome

packages/api/src/services/actions.ts
```diff
@@ -104,10 +104,15 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
...
     },
-    async getEmpStatsBetween(address: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
+    async getEmpStatsBetween(empAddress: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
       assert(stats[currency], "Invalid currency type: " + currency);
-      assert(stats[currency].history[address], "Invalid emp address: " + address);
-      return stats[currency].history[address].between(start, end);
+      assert(stats[currency].history[empAddress], "Invalid emp address: " + empAddress);
+      return stats[currency].history[empAddress].between(start, end);
+    },
+    async sliceHistoricalEmpStats(empAddress: string, start = 0, length = 1, currency: CurrencySymbol = "usd") {
```
chrismaree, line 112: is there any bounding you think would be reasonable on this? for example a max length or a min start? is there a case where you can overload this method somehow causing slow replies? what happens if the length is more than the data available?

chrismaree (approved):

packages/api/src/services/actions.ts
daywiss: yea thats a good call. there definitely needs to be some bounding, and we have to be careful about how we construct queries as well on the backend to avoid slow downs. I can add some preliminary limits, though we most likely will want to do this more thoroughly api is more stable.

packages/api/src/services/emp-stats.ts
mrice32: If 1 is possible, that seems a little scary (not necessarily for this case, but just in genneral). Is there a way to architecturally guarantee consistency or at least check for it?

2 makes sense. I think we'll eventually move to using events rather than the registry which would make this go away. If we weren't going to do that, we could try to infer the contract type upstream to avoid ingesting incomplete information.

mrice32 (approved):
LGTM!

packages/api/src/services/emp-stats.ts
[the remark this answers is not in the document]
daywiss: I think this was mainly to make it easy to construct the latest stats object partially, given only an address.  Becuase we have invalid EMPs we might not be able to get any "stats" for those addresses (including timestamp). 

I might be able to make timestamp required but at this point though might be moot since i am getting rid of this object in a future PR.

packages/api/src/services/emp-stats.ts
daywiss: > If 1 is possible, that seems a little scary (not just for this case, but just in general), especially when querying from databases that are keyed by time. Is there a way to architecturally guarantee consistency or at least check for it?

I've attempted to do this architecturally in the `api/app.ts` startup sequence. The order in which services get started is important, so we try to run an update to get all the emp state as one of the first routines, and this blocks any other services that would depend on this data from initializing.  Its not fool proof though, as an error fetching state from network can fail, leaving a null field in the emp state, so imo its best to try and be tolerant to inconsistent state.

PR #3123, opened 2021-06-18
Add pluggable MNEMONIC and GCKMS_KEYS to hardhat

mrice32:
**Motivation**

When using hardhat locally, we want to be able to use GCKMS keys as we do with truffle console.

**Summary**

Adds a plugin to inject GCKMS keys into hardhat. Because hardhat doesn't allow unrecognized command line parameters, the key names need to be passed in using an env variable called `GCKMS_KEYS=bot1,bot2,bot3`.

Note: this plugin works by initiating an async method inside a plugin, so it's not guaranteed to finish before the provider is created. In practice this seems to work for `hardhat console` which is primarily what we care about. However, running scripts or production applications inside hardhat using this plugin should be considered untested.

In general, we should stick with `getWeb3`-style providers in production and scripts to avoid being dependent on a particular framework.

This was the only way I could find to make this work, but if you have ideas to make this more resilient, please comment!

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/common/src/gckms/GckmsConfig.js
```diff
@@ -79,32 +65,48 @@ function getDefaultStaticConfig() {
...
+function getGckmsConfig(keys = arrayify(argv.keys), network = argv.network) {
+  let configOverride = {};
+
+  // If there is no env variable providing the config, attempt to pull it from a file.
+  // TODO: this is kinda hacky. We should refactor this to only take in the config using one method.
+  if (process.env.GCKMS_CONFIG) {
+    // If the env variable is present, just take that json.
+    configOverride = JSON.parse(process.env.GCKMS_CONFIG);
+  } else {
+    // Import the .GckmsOverride.js file if it exists.
```
chrismaree, line 77: this is backward compatible to still work with how we do things on truffle, right? clearly does not matter as we plan on removing Truffle in full but in the interim will things we used before still function?

packages/common/src/gckms/KeyInjectorPlugin.js
```diff
@@ -0,0 +1,41 @@
...
+      const name = client.cryptoKeyPath(config.projectId, config.locationId, config.keyRingId, config.cryptoKeyId);
+      const [result] = await client.decrypt({ name, ciphertext });
+      return "0x" + Buffer.from(result.plaintext, "base64").toString().trim();
+    })
+  );
+}
+
+extendConfig((config) => {
+  if (GCKMS_KEYS && MNEMONIC)
+    throw new HardhatPluginError("gckms::KeyInjectorPlugin", "Cannot provide both GCKMS_KEYS and MNEMONIC");
```
chrismaree, line 30: nice

packages/common/src/gckms/KeyInjectorPlugin.js
```diff
@@ -0,0 +1,41 @@
...
+      // Send the request to decrypt the downloaded file.
+      const client = new kms.KeyManagementServiceClient();
+      const name = client.cryptoKeyPath(config.projectId, config.locationId, config.keyRingId, config.cryptoKeyId);
+      const [result] = await client.decrypt({ name, ciphertext });
+      return "0x" + Buffer.from(result.plaintext, "base64").toString().trim();
+    })
+  );
+}
+
+extendConfig((config) => {
```
chrismaree, line 28: to double-check: this is the synchronous call that you were talking about offline, where and the part that cant be async?

packages/common/src/gckms/KeyInjectorPlugin.js
```diff
@@ -0,0 +1,41 @@
...
+  if (GCKMS_KEYS && MNEMONIC)
+    throw new HardhatPluginError("gckms::KeyInjectorPlugin", "Cannot provide both GCKMS_KEYS and MNEMONIC");
+
+  if (GCKMS_KEYS) {
+    const keyNameArray = GCKMS_KEYS.split(",");
+    const gckmsConfigs = getGckmsConfig(keyNameArray);
+    retrieveGckmsKeys(gckmsConfigs).then((keys) => {
```
chrismaree, line 35: yet you are calling the async method. this `.then` syntax means that _when_ the promise returns you will replace the network's keys?

chrismaree (approved):
Looks good! I know how painful this was to set up. I have a few comments/questions to just understand the code flow a bit better. otherwise, lgtm!

packages/common/src/gckms/GckmsConfig.js
```diff
@@ -79,32 +65,48 @@ function getDefaultStaticConfig() {
...
-function getNetworkName() {
-  if (isPublicNetwork(argv.network || "")) {
-    // Take everything before the underscore:
-    // mainnet_gckms -> mainnet.
-    return argv.network.split("_")[0];
+function getGckmsConfig(keys = arrayify(argv.keys), network = argv.network) {
+  let configOverride = {};
+
+  // If there is no env variable providing the config, attempt to pull it from a file.
+  // TODO: this is kinda hacky. We should refactor this to only take in the config using one method.
```
nicholaspai, line 72: > We should refactor this to only take in the config using one method.

You mean either via env variable or .GckmsOverride.js? If so, then I agree. Using env variable would be more consistent with the rest of our repo, but the env variable is very hard to parse and manipulate unlike the `.js` file.

Maybe @chrismaree or @daywiss has a better method of editing `.env` variables?

packages/common/src/gckms/KeyInjectorPlugin.js
```diff
@@ -0,0 +1,41 @@
+const kms = require("@google-cloud/kms");
+const { Storage } = require("@google-cloud/storage");
+const { extendConfig } = require("hardhat/config");
+const { HardhatPluginError } = require("hardhat/plugins");
+const { getGckmsConfig } = require("./GckmsConfig");
+
+const { GCKMS_KEYS, MNEMONIC } = process.env;
+
+async function retrieveGckmsKeys(gckmsConfigs) {
```
nicholaspai, line 9: So this is getting access to the key in `key management system` and using that to decrypt the file stored in `data storage` that can be found @ `keyMaterialBucket: cipherTextFile`?

A comment above this function would be useful for future maintainers

packages/common/src/gckms/GckmsConfig.js
mrice32: Sorry, this code was just moved, not changed :). I didn't add this comment. I tried to keep this implementation as close to the original as possible to maintain backwards compatibility.

packages/common/src/gckms/KeyInjectorPlugin.js
mrice32: Yeah, it's basically a more streamlined version of what's in `ManagedSecretProvider.js`. I may just combine the implementations so it's not repeated. I originally avoided that to minimize changes in production code, but 🤷 .

packages/common/src/gckms/KeyInjectorPlugin.js
mrice32: Yep, exactly.

packages/common/src/gckms/KeyInjectorPlugin.js
mrice32: Yep, exactly. As far as I can tell, there's no way to make this async. Another setup that would work just as well, but may be a bit less explicit would be to make the entire method async and use `await` syntax throughout. This would work, but not because hardhat supports it, just because it happens to finish before hardhat needs the result.

packages/common/src/gckms/GckmsConfig.js
mrice32: I have tested it with `truffle console` and it seems to still work!

packages/common/src/gckms/GckmsConfig.js
mrice32: Specifically on this issue, we use both in different places (the file locally and the env variable on GCP). We can definitely deprecate one, but I think that's separate from this PR.

The file works nicely locally because you can download it and forget it.

We have no good way of writing a file to cloud run, so env variables are more convenient. We could, however, host this file somewhere and have GCP download it, but that might be a little dangerous security-wise. Definitely more to think about there.

packages/common/src/gckms/GckmsConfig.js
mrice32: The vast majority of this code remains unchanged -- it was just moved to a function.

packages/common/src/gckms/KeyInjectorPlugin.js
mrice32: Done!

mrice32:
@chrismaree @nicholaspai I refactored the ManagedSecretProvider to use the same code as the hardhat plugin. This means that we should have a unified set of code to grab gckms keys regardless of purpose.

packages/common/src/gckms/GckmsConfig.js
chrismaree: yup. just wanted to double check :)

packages/common/src/gckms/GckmsConfig.js
chrismaree: ye IMHO having both is really useful and as @mrice32 pointed out envs work _much_ better in the context of cloud run.

chrismaree:
> @chrismaree @nicholaspai I refactored the ManagedSecretProvider to use the same code as the hardhat plugin. This means that we should have a unified set of code to grab gckms keys regardless of purpose.

awesome! looks great.

As an aside, your CI failed on the SDK package. I've re-run this for you

mrice32:
> > @chrismaree @nicholaspai I refactored the ManagedSecretProvider to use the same code as the hardhat plugin. This means that we should have a unified set of code to grab gckms keys regardless of purpose.
> 
> awesome! looks great.
> 
> As an aside, your CI failed on the SDK package. I've re-run this for you

Not sure why it's failing. Merged master to see if that helps 🤷

PR #3125, opened 2021-06-18
feat(LSP): single tx mint-sell & single tx mint-lp

chrismaree:

**Motivation**

Adds helper contract to batch LSP transactions with AMM transactions including:
- mint LSP tokens and sell all long/short into short/long tokens.
- mint LSP tokens and deposit them all into an LP position in the correct ratio

**Summary**

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested


nicholaspai:
General comment: let's simplify this PR and just implement "mint and sell", and I'll follow up with a subsequent "mint and LP" PR.

Thoughts?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,90 @@
...
+     * @param router Contract to call to exchange long and short tokens.
+     * @param collateralToMintWith Amount of collateral to deposit and borrow long and short tokens against.
+     * @param swapPath `Router.swapExactTokensForTokens` param: path with which to swap token to sell for the other.
+     * @param deadline `Router.swapExactTokensForTokens` param: time before transaction must be mined.
+     */
+    function atomicMintSellOneSide(
+        bool tradingAsEOA,
+        bool sellLong,
+        LongShortPair longShortPair,
+        IUniswapV2Router01 router, /* TODO: Should we allow `router` to be any exchange, such as a Matcha multihop? */
```
nicholaspai, line 41: I think this OK to not build for this version, but might be worth building into future version. Idea would be to allow caller to specify a `router`, which could be any contract, and then pass in `bytes swapData`, which could be called like: `router.call(swapData)`

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,90 @@
...
+            );
+
+        TransferHelper.safeApprove(
+            address(longShortPair.collateralToken()),
+            address(longShortPair),
+            collateralToMintWith
+        );
+
+        // 1) Deposit collateral into LSP and mint long and short tokens.
+        // TODO: So this assumes we are minting collateral --> long/short tokens in a one-to-one ratio?
```
nicholaspai, line 66: @chrismaree why do you this assumption that `LSP.collateralPerPair = 1`?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,90 @@
...
+        // i.e. The `LSP.collateralPerPair=1`
+        longShortPair.create(collateralToMintWith);
+
+        // 2) Determine which token we are selling and convert it all into the other.
+        IERC20 soldToken = IERC20(sellLong ? longShortPair.shortToken() : longShortPair.longToken());
+        // TODO: Should we make this check?
+        // require(soldToken.balanceOf(address(this)) > 0, "Missing balance of token to sell");
```
nicholaspai, line 73: Will `swapExactTokensForTokens` revert if the amount to sell is 0? If not, then its probably worth reverting early

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,90 @@
...
+import "../../common/implementation/FixedPoint.sol";
+
+import "../../financial-templates/long-short-pair/LongShortPair.sol";
+
+/**
+ * @title LspUniswapV2Broker
+ * @notice Helper contract to facilitate batched LSP and UniswapV2 transactions, including Mint+Sell and Mint+LP.
+ */
+contract LspUniswapV2Broker {
+    using FixedPoint for FixedPoint.Unsigned;
```
nicholaspai, line 18: I removed `SafeMath`. We can remove all `SafeMaths` in a future (massive) PR but might as well do it when we have a chance to do it.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,90 @@
...
+     * @param sellLong If True, converts all long tokens into short, else the opposite.
+     * @param longShortPair LSP contract address to mint position on.
+     * @param router Contract to call to exchange long and short tokens.
+     * @param collateralToMintWith Amount of collateral to deposit and borrow long and short tokens against.
+     * @param swapPath `Router.swapExactTokensForTokens` param: path with which to swap token to sell for the other.
+     * @param deadline `Router.swapExactTokensForTokens` param: time before transaction must be mined.
+     */
+    function atomicMintSellOneSide(
+        bool tradingAsEOA,
+        bool sellLong,
```
nicholaspai, line 39: I renamed this var from `tradeLong` to `sellLong` because its more consistent with the variable name `IERC20 soldToken` below. This makes it much clearer IMO which token we're selling.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: ye this will save bytecode and gas. save for later in other files

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: ye for sure. bit more complex like that though ;)

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: hhhmmm. it does not assume that 🤔 . it simply mints equal number of long and short (which is how the LPS works) and makes no assumption about the `collateralPerPair`. The `collateralPerPair` could be _any_ value

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: it will I believe. however, there are not exactly any downsides if you try call this method and mint 0 with `collateralToMintWith=0`

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: it looks like you actually added a test for this with `// Cannot mint with 0 collateral` which will sufficiently validate this I belive.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: it looks like you actually added a test for this with `// Cannot mint with 0 collateral` which will sufficiently validate this I believe.

nicholaspai: Yeah I'm more so thinking about the situation where `create` somehow creates 0 tokens, but I suppose not reverting is not an issue

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: I'm leaving the TODO there to remind us

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: You're right, my mistake!

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: This [PR](https://github.com/UMAprotocol/protocol/pull/3132) now prevents this edge case where you could create 0 tokens and try to swap 0 tokens.

nicholaspai (approved):
@chrismaree PTAL at my most recent changes and then LGTM

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,86 @@
...
+        LongShortPair longShortPair,
+        IUniswapV2Router01 router, /* TODO: Should we allow `router` to be any exchange, such as a Matcha multihop? */
+        uint256 collateralToMintWith,
```
mrice32, line 42: nit: would `amountCollateral` or just `collateral` be clearer? When I read this I thought it was an address :)

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,86 @@
...
+                msg.sender,
+                address(this),
+                collateralToMintWith
+            );
+
+        TransferHelper.safeApprove(
+            address(longShortPair.collateralToken()),
+            address(longShortPair),
+            collateralToMintWith
+        );
```
mrice32, line 63: nit: why do we use uniswap's safe transfer libraries rather than OZ (which is what we generally use elsewhere)? Is there an advantage to theirs?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,86 @@
...
+        uint256 deadline
+    ) public {
+        require(address(longShortPair) != address(0), "Invalid long short pair");
+        require(address(router) != address(0), "Invalid router");
+        require(collateralToMintWith != 0, "Collateral to mint with");
+
+        // 0) Pull collateral from caller if necessary and approve LSP to spend it.
+        if (tradingAsEOA)
+            TransferHelper.safeTransferFrom(
+                address(longShortPair.collateralToken()),
```
mrice32, line 53: Can we pull the collateralToken into memory so we don't have to call this function multiple times?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -0,0 +1,86 @@
...
+
+        // 2) Determine which token we are selling and convert it all into the other.
+        IERC20 soldToken = IERC20(sellLong ? longShortPair.shortToken() : longShortPair.longToken());
+        TransferHelper.safeApprove(address(soldToken), address(router), soldToken.balanceOf(address(this)));
+        require(swapPath[0] == address(soldToken), "Sold token != 0th swapPath");
+        router.swapExactTokensForTokens(
+            soldToken.balanceOf(address(this)), // sell all of the sold tokens held by the contract.
+            0,
+            swapPath,
+            address(this),
```
mrice32, line 76: Should we make this msg.sender if trading as EOA? Otherwise, when are we moving the result of the trade into the user's wallet?

mrice32 (commented):
Looks awesome! A few comments

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: sure. opted for `amountCollateral`

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: sure.done.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: OZ's `SafeERC20` works by overriding some of the behavior of `IERC20` or `ExpandedIERC20` (depending on the import) to ensure that token actions will throw on failure, rather than returning false. In my work previously with the DSProxy scripts I found the uniswap method `TransferHelper` to use _less_ gas and bytecode than the OZ implementation as you don't need to instantiate an instance of the contract you are calling into. Granted, you could do the same thing with the OZ libs by declaring it inline then calling methods on it but I found that even then the uniswap implementation used less. 

The pattern that I've gone for with these scripts is any code within the `proxy-scripts` directory (which includes things to help using a DSProxy or code like this file) uses the Uniswap helper and the rest of the repo uses the OZ implementation. 

I'm happy to refactor these to all use the OZ implementation but we might hit bytecode issues again in these scripts if they are on the margin.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: look below. it checks if it is an EOA then it sends all of the `purchasedToken` (the one the user purchased) to the user.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: I'm in favor of keeping Uniswap's transfer helper for consistency within this directory

nicholaspai (approved):
LGTM

PR #3129, opened 2021-06-21
feat(LSP): Validate all logic works as expected with non-18 decimal collateral

chrismaree:
**Motivation**

The LSP currently contains no unit tests to validate that it behaves correctly with non-18 decimal collateral. Many partners want to collateralized this product in USDC and so it's only prudent to write some tests to validate this behavior works exactly as expected.

**Summary**

Updates the LSP tests to validate USDC works with full lifecycle.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


nicholaspai (approved):
Nice tests LGTM

PR #3130, opened 2021-06-21
fix(LSP): Prevent parameter override on Range Bond Lib

chrismaree:
**Motivation**

There was a small but substantial bug in the `RangeBondLongShortPairFinancialProductLibrary.sol` that enabled the bond's params to be overridden. Clearly, this is undesirable behavior as it could lead to a malicious actor affecting settlement differently from the original contract parameterization. This PR fixes this bug as well as the tests that were showing up false positives on this test case.

**Summary**

Updated the contracts and tests to address a bug in an LSP contract.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
@@ -63,6 +63,7 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
         require(highPriceRange > lowPriceRange, "Invalid bounds");
 
         RangeBondLongShortPairParameters memory params = longShortPairParameters[LongShortPair];
+        require(params.highPriceRange == 0 && params.lowPriceRange == 0, "Parameters already set");
 
         longShortPairParameters[LongShortPair] = RangeBondLongShortPairParameters({
             highPriceRange: highPriceRange,
--- a/packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.js
+++ b/packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.js
@@ -34,22 +34,12 @@ contract("RangeBondLongShortPairFinancialProductLibrary", function () {
       assert.equal(setParams.highPriceRange.toString(), highPriceRange);
     });
     it("Can not re-use existing LSP contract address", async () => {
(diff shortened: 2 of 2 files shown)


packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.js
```diff
@@ -34,35 +34,20 @@ contract("RangeBondLongShortPairFinancialProductLibrary", function () {
...
-
-        highPriceRange,
-        lowPriceRange
-      );
+      await rangeBondLSPFPL.setLongShortPairParameters(lspMock.address, highPriceRange, lowPriceRange);
 
       // Second attempt should revert.
       assert(
         await didContractThrow(
-          rangeBondLSPFPL.setLongShortPairParameters(
```
chrismaree, line 47: this test was generating false positives! the parameterization of `lowPriceRange` first then `highPriceRange` will revert due to the other conditions that validate the correctness of the parameters, not the overriding of parameters.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
```diff
@@ -63,6 +63,7 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
         require(highPriceRange > lowPriceRange, "Invalid bounds");
 
         RangeBondLongShortPairParameters memory params = longShortPairParameters[LongShortPair];
+        require(params.highPriceRange == 0 && params.lowPriceRange == 0, "Parameters already set");
```
chrismaree, line 66: this is silly as we already had declared the `params` object to validate this but did not have the require!

nicholaspai: What if someone sets the `lowPriceRrange = 0`?

chrismaree: well `highPriceRange > lowPriceRange` from the requirements on line 63 so it would not be possible for them both to be 0. you _could_ set `lowPriceRrange = 0` but by definition `highPriceRange` could then also not be zero and this requirement would fail. i.e the only case where this require passes is when there is nothing set for this value.

do you agree?

nicholaspai: good point, I agree.

nicholaspai (approved):
LGTM

mrice32 (approved):
LGTM!

PR #3136, opened 2021-06-21
fix(core): fix polygon deployent scripts for test/local environments

mrice32:
**Motivation**

Polygon bridge contracts fail to deploy on networks other than those with hardcoded addresses in the deployment scripts (including local networks needed for testing).

**Summary**

Mocks were added to deployment process as a fallback in case these hardcoded addresses aren't present. These mocks are tagged with `test` and their names so they won't be deployed as a part of a typical l2 polygon deployment unless specifically added.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


nicholaspai (approved):
LGTM

PR #3141, opened 2021-06-22
feat(LSP-Broker): Add ability to mint and LP in one transaction

chrismaree:
**Motivation**

As part of adding LSP helpers, we require the ability so mint and LP tokens within one transaction. This PR adds this logic to the `LspUniswapV2Broker` contract and unit tests.

**Summary**

This implementation works by effectively computing how many tokens need to be bought/sold between the minted long and short tokens such that the remaining token balance in the smart contract is equal to the pool ratio _post_ trade (i.e considering the impact of the trade on the pool ratios).

See the implementation and code comments for more details on how this works.

Some other commenting and test changes were also made for consistency.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -19,23 +21,55 @@ contract LspUniswapV2Broker {
 
     /**
      * @notice Mint long and short tokens and deposit them all into a UniV2 Pool.
+
```
nicholaspai, line 24: Can you add a @dev comment to this and the `mintAndSell` function about how the caller needs to have approved this contract so spend `amountCollateral` of the collateral token?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -19,23 +21,55 @@ contract LspUniswapV2Broker {
...
+            TransferHelper.safeTransferFrom(address(collateralToken), msg.sender, address(this), amountCollateral);
+
+        TransferHelper.safeApprove(address(collateralToken), address(longShortPair), amountCollateral);
+
+        // 1) Deposit collateral into LSP and mint long and short tokens.
+        longShortPair.create(amountCollateral);
+
+        // 2) Send long and short tokens to the pair address and call the low-level mint function. This amounts to
+        // single asset deposit where the one side will be sold for the other to match the pool ratio.
+        IUniswapV2Pair pair = pairFor(router.factory(), longToken, shortToken);
```
nicholaspai, line 55: Would this work if the `longToken` and `shortToken` param order were reversed? Just curious

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: sure.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: yes. `pairFor` uses `sortTokens` which means it'll work irrespective of order.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: we should do that on `atomicMintSellOneSide` as well

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -19,23 +21,55 @@ contract LspUniswapV2Broker {
...
+        TransferHelper.safeTransfer(address(longToken), address(pair), longToken.balanceOf(address(this)));
+        TransferHelper.safeTransfer(address(shortToken), address(pair), shortToken.balanceOf(address(this)));
+        pair.mint(callingAsEOA ? msg.sender : address(this));
```
nicholaspai, line 58: Does `mint` guarantee that it uses 100% of the two tokens to create LP tokens? I.e. there is no dust?

chrismaree: I'm pretty sure it does, ye

packages/core/test/proxy-scripts/lsp-broker/LspUniswapV2Broker.js
```diff
@@ -496,4 +491,191 @@ contract("LspUniswapV2Broker", function (accounts) {
...
+      const lpTokensPerCollateral = (await lpToken.balanceOf(deployer))
+        .mul(toBN(toWei("1"))) // scalding factor
+        .div(toBN(toWei("100000"))) // number of collateral units used in minting
+        .addn(1); // offset to fix rounding error
+
+      await lspUniswapV2Broker.atomicMintAddLiquidity(
+        true, // tradingAsEOA. true as calling from an EOA (not DSProxy).
+        longShortPair.address, // longShortPair. address to mint tokens against.
+        router.address, // router. uniswap v2 router to execute trades
+        toWei("1000"), // collateralToMintWith. we will use 1000 units of collateral to mint 1000 long and 1000 short tokens.
```
nicholaspai, line 620: can you try one test where the amount to mint is > 50% of the pool, to test what happens with the low level `UniswapPair.mint` call when slippage is really high? So say minting with 5,000 collateral? or even 50,000 collateral? (I don't know if that would even work?)

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
  * @notice Helper contract to facilitate batched LSP and UniswapV2 transactions, including Mint+Sell and Mint+LP.
  */
 contract LspUniswapV2Broker {
+    using FixedPoint for FixedPoint.Signed;
     using FixedPoint for FixedPoint.Unsigned;
 
     /**
      * @notice Mint long and short tokens and deposit them all into a UniV2 Pool.
+     * @dev The caller of this method needs to approve `amountCollateral` collateral to be spent by this contract.
+     * @param callingAsEOA If True, caller has balance of collateral and expects to receive back all long/short tokens.
```
nicholaspai, line 27:
```suggestion
     * @param callingAsEOA If True, caller has balance of collateral and expects to receive back all LP plus long/short tokens.
```

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+     * @param longShortPair LSP contract address to mint position on.
+     * @param router Contract to call to exchange long and short tokens.
+     * @param amountCollateral Amount of collateral to deposit and mint long and short tokens against.
```
nicholaspai, line 30: missing `deadline` param comment

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        require(amountCollateral != 0, "Collateral to mint with");
+
+        IERC20 collateralToken = IERC20(longShortPair.collateralToken());
+        IERC20 longToken = IERC20(longShortPair.longToken());
+        IERC20 shortToken = IERC20(longShortPair.shortToken());
+
+        // 0) Pull collateral from caller if necessary and approve LSP to spend it.
+        if (callingAsEOA)
+            TransferHelper.safeTransferFrom(address(collateralToken), msg.sender, address(this), amountCollateral);
+
```
nicholaspai, line 50:
```suggestion
```

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        TransferHelper.safeApprove(address(collateralToken), address(longShortPair), amountCollateral);
+
+        // 1) Deposit collateral into LSP and mint long and short tokens.
+        uint256 tokensToCreate =
+            FixedPoint.fromUnscaledUint(amountCollateral).div(longShortPair.collateralPerPair()).rawValue;
+
```
nicholaspai, line 56:
```suggestion
```

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: I like grouping together actions that fall under the same comment. So the `transferFrom` and the `approve` both execute the comment: `0) Pull collateral from caller if necessary and approve LSP to spend it.`

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        longShortPair.create(tokensToCreate);
+
+        require(
```
nicholaspai, line 59: How does this `require` condition ever get reached?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+            longToken.balanceOf(address(this)) == shortToken.balanceOf(address(this)) &&
+                longToken.balanceOf(address(this)) == tokensToCreate,
+            "Create invariant failed"
+        );
+
+        {
+            // 2) Calculate if we need to sell the long token for the short or the short token for the long. Calculate
+            // the total trade size to bring the balance of long/short in this contract to equal the pool ratio after
+            // the trade. Note this computation must consider the impact of the trade with the resultant pool ratios.
```
nicholaspai, line 68:
```suggestion
            // the trade. Note this computation must consider how the resultant pool ratios is impacted by the trade.
```

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+                // know the exact input number of tokens that we need to sell (short tokens) and can use the router's
+                // swapExactTokensForTokens with the exact input spesificed.
+                path[0] = address(shortToken);
+                path[1] = address(longToken);
+                TransferHelper.safeApprove(address(shortToken), address(router), tradeSize);
+                router.swapExactTokensForTokens(tradeSize, 0, path, address(this), deadline);
+            }
+        }
+
+        // 3) Add lequidity to the pool via the router. Approve both long and short tokens with the full account balance
```
nicholaspai, line 100:
```suggestion
        // 3) Add liquidity to the pool via the router. Approve both long and short tokens with the full account balance
```

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+            // 4) Finally, if the caller is an EOA, send back the LP tokens from minting + any dust that was left over.
+            // The dust is in the long and short tokens and will be left over as a result of small rounding errors in
+            // the calculation on how many long/short should be bought/sold to meet the pool ratio before minting.
+            if (callingAsEOA) {
+                // Send the LP tokens back to the minter.
+                IERC20 LPToken = IERC20(address(pairFor(router.factory(), address(longToken), address(shortToken))));
+                TransferHelper.safeTransfer(address(LPToken), msg.sender, LPToken.balanceOf(address(this)));
+
+                // Send any dust left over back to the minter.
+                TransferHelper.safeTransfer(address(longToken), msg.sender, longToken.balanceOf(address(this)));
```
nicholaspai, line 124: Is there any gas savings if we can catch early if either `token.balanceOf(address(this)) == 0` and return early without calling `safeTransfer`??

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        router.addLiquidity(
+            address(longToken),
+            address(shortToken),
+            longToken.balanceOf(address(this)),
+            shortToken.balanceOf(address(this)),
+            0,
+            0,
+            address(this),
+            deadline
+        );
```
nicholaspai, line 113: I think after this step, we should run a `require(lpToken.balanceOf(address(this)) > minLpTokens` where `uint256 minLpTokens` is a function input that the user specifies to ensure that trade slippage does not result in fewer LP tokens than would normally be mintable with the long-short tokens.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+    // hold t_l=m+△l long tokens and t_s=m-△s after a trade of s for l tokens. The ratio t_l/t_s must equal the pool
+    // ratio after the trade. The pool ratio, considering the trade can be expressed as (R_a-△a)(R_b+△bλ)=k for a
+    // trade of △a for △b and a swap fee of λ=(1-swapFee). Using this, with a bit of algebra and simplification, we can
+    // solve for the resultant ratio to be (m+△a)/(m-△b)=(R_a-△a)/(R_b+△b). i.e the ratio of mint+trade must equal the
+    // ratio of token A and B in the pool after the trade has concluded. We also know that △a can be solved for as
+    // △a=(△bλR_a)/(R_b+△bλ) by manipulating a known uniswap equation. Using this equation and the expression of ratios
+    // we can solve simaltaniously for △b. I.e how many short tokens do we need to sell such that the ratio of long and
+    // short after the trade equal the pool ratios. Numerically, this works out to this form:
+    // △b=(sqrt(R_b(R_a+m)(R_a(R_b(λ^2+2λ+1)+4λm)+R_b(λ^2-2λ+1)m))+R_b(-λ-1)(R_a+m))/(2λ(R_a+m)). For how this was solved
+    // from the algebraic solution above see https://bit.ly/3jn7AF6 on wolfram alpha showing the derivation.
```
nicholaspai, line 142: +1 nice comment

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+            "Create invariant failed"
+        );
+
+        {
+            // 2) Calculate if we need to sell the long token for the short or the short token for the long. Calculate
+            // the total trade size to bring the balance of long/short in this contract to equal the pool ratio after
+            // the trade. Note this computation must consider the impact of the trade with the resultant pool ratios.
+            bool aToB;
+            uint256 tradeSize;
+            (uint256 reserveA, uint256 reserveB) =
```
nicholaspai, line 71: What if the rare case (but would happen right after a pool was initialized) where `reserveA == reserveB`? Could you skip the `computeSwapToMintAtPoolRatio` step and go straight to `addLiquidity`?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        FixedPoint.Signed memory denominator = num(2).mul(lambda()).mul(ra.add(m));
+
+        FixedPoint.Signed memory tradeSize = numerator.div(denominator);
+
+        // If the trade size is negative then we are traversing the equation backwards. In this case the contract must
+        // swap between swapTokensForExactTokens and swapExactTokensForTokens methods as we are always solving for △b.
+        // In other words, we are always solving for how many short tokens are needed to be bought/sold. If the number
+        // is positive then we are buying them. if it is negative then we are selling them. In the case we are selling
+        // them (negative) then we swap the polarity and apply the swap fee to the trade.
+        bool aToB = tradeSize.isLessThan(0);
```
nicholaspai, line 170: Its awesome that this function returns positive/negative depending on the direction you need, and the magnitude is always correct!

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+        {
+            // 4) Finally, if the caller is an EOA, send back the LP tokens from minting + any dust that was left over.
+            // The dust is in the long and short tokens and will be left over as a result of small rounding errors in
+            // the calculation on how many long/short should be bought/sold to meet the pool ratio before minting.
+            if (callingAsEOA) {
+                // Send the LP tokens back to the minter.
+                IERC20 LPToken = IERC20(address(pairFor(router.factory(), address(longToken), address(shortToken))));
+                TransferHelper.safeTransfer(address(LPToken), msg.sender, LPToken.balanceOf(address(this)));
+
+                // Send any dust left over back to the minter.
```
nicholaspai, line 123: Can we bound or describe in more detail how `dust` actually is created? I.e. how does the algebra in `computeSwapToMintAtPoolRatio` result in any dust and why can't we get the dust to 0? 

No need to numerically bound the dust, but would be useful to know what the general source of error is (i.e. fees, unquantifiable slippage, etc.)

packages/core/test/proxy-scripts/lsp-broker/LspUniswapV2Broker.js
```diff
@@ -496,4 +467,395 @@ contract("LspUniswapV2Broker", function (accounts) {
...
+      // each side. This verifies that the number returned is correctly proportional to that deposited.
+      const walletRatio = longTraderBalance.mul(adjustment).div(shortTraderBalance);
+      const poolRatio = poolLongBalPostAdd.mul(adjustment).div(poolShortBalPostAdd);
+
+      assert.isTrue(walletRatio.addn(10).gte(poolRatio) && walletRatio.subn(10).lte(poolRatio));
+    });
+    it("Can correctly mint and LP in one transaction with pool in unequal ratio with long<short", async function () {
+      // Add 100000 long and 1000000 short. this makes the price 100000/1000000 long/short. In other words every short is
+      // worth 0.01 long. When LPing we will be market selling longs for shorts to reach the appropriate ratio.
+      await addLiquidityToPool(longToken, shortToken, toWei("100000"), toWei("1000000"));
```
nicholaspai, line 619: Can you make the long side initial reserves = 1,000, the amount of longs you mint initially? I'm curious what this "max slippage" situation would look like.

nicholaspai: Oh nevermind I see you do this test below!

nicholaspai (commented):
This is AWESOME work! Nice job figuring out the numerical solution, empirically it looks great. Not at all sure how it works under the hood, but i'll think about situations which would maximize the amount of dust left over.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
```diff
@@ -15,27 +18,197 @@ import "../../financial-templates/long-short-pair/LongShortPair.sol";
...
+    // following logic: Assume m tokens are minted in equal proportion of long and short. This contract will therefore
+    // hold t_l=m+△l long tokens and t_s=m-△s after a trade of s for l tokens. The ratio t_l/t_s must equal the pool
+    // ratio after the trade. The pool ratio, considering the trade can be expressed as (R_a-△a)(R_b+△bλ)=k for a
+    // trade of △a for △b and a swap fee of λ=(1-swapFee). Using this, with a bit of algebra and simplification, we can
+    // solve for the resultant ratio to be (m+△a)/(m-△b)=(R_a-△a)/(R_b+△b). i.e the ratio of mint+trade must equal the
+    // ratio of token A and B in the pool after the trade has concluded. We also know that △a can be solved for as
+    // △a=(△bλR_a)/(R_b+△bλ) by manipulating a known uniswap equation. Using this equation and the expression of ratios
+    // we can solve simaltaniously for △b. I.e how many short tokens do we need to sell such that the ratio of long and
+    // short after the trade equal the pool ratios. Numerically, this works out to this form:
+    // △b=(sqrt(R_b(R_a+m)(R_a(R_b(λ^2+2λ+1)+4λm)+R_b(λ^2-2λ+1)m))+R_b(-λ-1)(R_a+m))/(2λ(R_a+m)). For how this was solved
```
mrice32, line 141: Maybe I'm missing the subtlety here, but after reading your comment, could the same thing be computed and mostly just reuse the logic we've written elsewhere to trade to a particular price? This way could be just as complicated, so even if this is correct, I'm not necessarily saying we should change anything. Just seemed a little simpler in my head.

Note that the following is just pseudocode that is under-representative of the complexity.
```
// We know that all the minted tokens are going to end up in the
// uniswap pool regardless, so the final reserves _must_ be the following.
long_reserves_final = long_minted + long_reserves_initial;
short_reserves_final = short_minted + short_reserves_final;

// Since we know that these reserves are going to be where we end up,
// we can divide them to get the price.
// Because adding liquidity doesn't change the price,
// we know this is the price we will mint at.
desired_price = long_reserves_final / short_reserves_final;

tradeToPrice(desired_price); // What the trader DSProxy code does.
router.mint(balanceOf(short_token), balanceOf(long_token)); // Mint using all of our tokens.
```

mrice32 (commented):
One high-level comment. This is super cool!!

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: What you put down is pretty much the same reasoning I used to get to this form. The form I presented is numerically the same logic but has been manipulated to simplify the equations so they will fit in solidity. I'm pretty sure you will arrive at an equally complex form if you do it this way (if not more complex). We should chat about it more IRL.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: it would be marginal I think. Plus the code complexity is properly not warranted for the saving seeing this is running on L2 in most cases. WDYT?

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: ye, not a bad idea. will add it in + associated unit tests.

chrismaree: I added this in plus unit tets to validate it. I also added tests to validate `Invalid params` as we have with the swap logic

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: if that happens then `tradeSize=0` and the if statement a few lines below where the trading happens will be skipped.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: I added some more comments to this effect.

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
nicholaspai: I would think the code complexity would actually go down though if you exit early

packages/core/contracts/proxy-scripts/lsp-broker/LspUniswapV2Broker.sol
chrismaree: I tried implementing your technique @mrice32. it worked out to be _much_ easier to reason about and had a lot simpler math. _However_ the dust leftover is measurably worse. On a simple setup with a pool with 1000000/100000 long/short (10:1) ratio has less than 100 wei dust in long with my computation for a mint of 1000 units of collateral. Your method has about 1.3e18 in long dust (this is still less than 0.3% of the total, which would be the fee amount).

I spent a bunch of time trying to re-work this to introduce additional fees into the computation to get it to track correctly with your technique but could never get the dust anywhere close to my computation.

I think this is due to the fees being ignored at some level in this computation. However, I’m not convinced by this as the fees will end up in the respective pools, irrespective of the swap size so that’s a bit confusing.

In any case, for now, I think we should stick with my implementation as it works and has low dust, despite its high complexity. You agree?

mrice32: Agreed!

mrice32 (approved):
LGTM!

nicholaspai (approved):
Let's go with your implementation over matt's simpler but seemingly less correct one

PR #3142, opened 2021-06-22
improve(core): Update deploy-emp script to enhance block explorer verification

nicholaspai:
**Motivation**

- Script included some superfluous params in the ABI-encoded construction params, making verification hard.
- Add `chainId` to transaction config to work with (Polygon) endpoints that enforce transaction-replay-protection

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/scripts/collateral-umip/1_Propose.js
+++ b/packages/core/scripts/collateral-umip/1_Propose.js
@@ -31,7 +31,7 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 async function runExport() {
   console.log("Running Upgrade🔥");
 
-  let netId = await web3.eth.net.getId();
+  const netId = await web3.eth.net.getId();
   console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
--- a/packages/core/scripts/collateral-umip/1_Propose.js
+++ b/packages/core/scripts/collateral-umip/1_Propose.js
@@ -30,14 +30,16 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  console.log("Connected to network id", await web3.eth.net.getId());
+
+  let netId = await web3.eth.net.getId();
+  console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
     winston.createLogger({
       silent: true,
     }),
     60, // Time between updates.
-    100 // Default gas price.
+    netId
   );
 
   if (!argv.collateral || !argv.fee) {
--- a/packages/core/scripts/creator-umip/1_Propose.js
+++ b/packages/core/scripts/creator-umip/1_Propose.js
@@ -17,14 +17,15 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  console.log("Connected to network id", await web3.eth.net.getId());
+  let netId = await web3.eth.net.getId();
+  console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
     winston.createLogger({
       silent: true,
     }),
     60, // Time between updates.
-    100 // Default gas price.
+    netId
   );
 
   // The proposal will add this new contract creator to the Registry.
--- a/packages/core/scripts/creator-umip/1_Propose.js
+++ b/packages/core/scripts/creator-umip/1_Propose.js
@@ -17,7 +17,7 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  let netId = await web3.eth.net.getId();
+  const netId = await web3.eth.net.getId();
   console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
--- a/packages/core/scripts/identifier-umip/1_Propose.js
+++ b/packages/core/scripts/identifier-umip/1_Propose.js
@@ -24,7 +24,7 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 async function runExport() {
   console.log("Running Upgrade🔥");
 
-  let netId = await web3.eth.net.getId();
+  const netId = await web3.eth.net.getId();
   console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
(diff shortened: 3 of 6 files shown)


packages/core/scripts/local/DeployEMP.js
```diff
@@ -153,7 +160,7 @@ const deployEMP = async (callback) => {
     // Done!
     console.log(`Created a new EMP @ ${emp.address} with the configuration:`);
     console.log(`Deployer address @ ${deployer}`);
-    console.log("Encoded EMP Parameters", encodedParameters);
+    console.log("Encoded EMP Parameters", encodedParameters.slice(2));
```
nicholaspai, line 163: ABI encoded params usually do not include leading 0x

packages/core/scripts/local/DeployEMP.js
```diff
@@ -127,17 +123,28 @@ const deployEMP = async (callback) => {
       };
     }
 
-    let _emp = await expiringMultiPartyCreator.createExpiringMultiParty.call(constructorParams, { from: deployer });
-    await expiringMultiPartyCreator.createExpiringMultiParty(constructorParams, { from: deployer });
+    if (!argv.gasprice) throw new Error("Specify --gasprice to deploy with");
```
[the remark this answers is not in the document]
chrismaree: a number of other scripts use the gasEstimator. that's likely the best solution don't you think?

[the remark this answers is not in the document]
nicholaspai: I don't think they do actually, only the bots use the gas estimator?

packages/core/scripts/local/DeployEMP.js
```diff
@@ -21,12 +21,13 @@
  *
  * Example: yarn truffle exec ./packages/core/scripts/local/DeployEMP.js --network test --test true --identifier ETH/BTC --cversion 1.1.0
```
pemulis, line 22: You could drop the `--cversion` flag in the example or set it to the default version of `1.2.2`.

pemulis (approved):
LGTM! This should be simpler. One small note about the example script in the comments.

packages/core/scripts/local/DeployEMP.js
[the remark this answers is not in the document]
mrice32: We added it to a few newer scripts. The script that executes governance proposals after they're approved and the one that claims/estimates rewards for any address both use it.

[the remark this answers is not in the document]
nicholaspai: Ah just realized that some of the `GasEstimator` constructions needed to change now that we support multiple network ID's. Requested your reviews again on [this commit](https://github.com/UMAprotocol/protocol/pull/3142/commits/463422a43ebd47dd31f33e720ff56c536e3fdb2e)

packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -30,14 +30,16 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  console.log("Connected to network id", await web3.eth.net.getId());
+
+  let netId = await web3.eth.net.getId();
```
mrice32, line 34:
```suggestion
  const netId = await web3.eth.net.getId();
```

packages/core/scripts/creator-umip/1_Propose.js
```diff
@@ -17,14 +17,15 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  console.log("Connected to network id", await web3.eth.net.getId());
+  let netId = await web3.eth.net.getId();
```
mrice32, line 20:
```suggestion
  const netId = await web3.eth.net.getId();
```

packages/core/scripts/identifier-umip/1_Propose.js
```diff
@@ -23,14 +23,16 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
 
 async function runExport() {
   console.log("Running Upgrade🔥");
-  console.log("Connected to network id", await web3.eth.net.getId());
+
+  let netId = await web3.eth.net.getId();
```
mrice32, line 27:
```suggestion
  const netId = await web3.eth.net.getId();
```

packages/core/scripts/local/DeployEMP.js
```diff
@@ -127,17 +125,35 @@ const deployEMP = async (callback) => {
...
+      await web3.eth.net.getId()
+    );
+    await gasEstimator.update();
+    const transactionParams = {
+      gas: 12000000, // 12MM is very high. Set this lower if you only have < 2 ETH or so in your wallet.
+      gasPrice: gasEstimator.getCurrentFastPrice(),
+      from: deployer,
+      chainId: await web3.eth.getChainId(),
+    };
+    let _emp = await expiringMultiPartyCreator.createExpiringMultiParty.call(constructorParams, transactionParams);
```
mrice32, line 142:
```suggestion
    const _emp = await expiringMultiPartyCreator.createExpiringMultiParty.call(constructorParams, transactionParams);
```

packages/core/scripts/collateral-umip/1_Propose.js
```diff
@@ -30,14 +30,16 @@ const proposerWallet = "0x2bAaA41d155ad8a4126184950B31F50A1513cE25";
...
+  console.log("Connected to network id", netId);
 
   const gasEstimator = new GasEstimator(
     winston.createLogger({
       silent: true,
     }),
     60, // Time between updates.
-    100 // Default gas price.
+    netId
```
mrice32, line 42: What's the tradeoff between using network id vs chain id for stuff like this? A lot of other code (like hardhat deployments) use chain id.

mrice32 (approved):
Looks good!

packages/core/scripts/collateral-umip/1_Propose.js
nicholaspai: Technically [network ID is for node-node communication while chain ID is for transaction signatures](https://besu.hyperledger.org/en/stable/Concepts/NetworkID-And-ChainID/). So the Gas Estimator stores a mapping of networks to gas station API's so I think its proper in this case to use the `networkId` to specify the "network" to use.  While the [transaction helper](https://github.com/UMAprotocol/protocol/blob/master/packages/common/src/TransactionUtils.js#L18) fetches and injects the chain ID when submitting transactions.

mrice32: Hmmmm, interesting. I see what you're saying wrt how these are used in practice, but just to play devil's advocate, if we wanted to run bots on ETC, it and ETH have the same network id (because they used to be the same network!), but different chain ids. So we would be unable to differentiate when trying to estimate gas. I think the ETC/ETH split was the reason they created chainids if I remember correctly.

We're not planning to roll out to ETC anytime soon, so I don't think this is a real concern. I was mainly curious how you were thinking about it.

mrice32 (approved):
LGTM!

PR #3147, opened 2021-06-22
feat(cryptowatch): Add historical timestamp buffer

nicholaspai:
**Motivation**

Enables CryptoWatchPriceFeed to add a buffer around price periods when matching a historical timestamp.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
@@ -21,6 +21,9 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
    * @param {Bool} invertPrice Indicates if prices should be inverted before returned.
    * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
    * @param {Number} ohlcPeriod Number of seconds interval between ohlc prices requested from cryptowatch.
+   * @param {Number} twapLength Number of seconds to use for TWAP window when computing prices.
+   * @param {Number} historicalTimestampBuffer Number of seconds +/- beyond a price period's open and close window
+   * that determines whether a historical timestamp falls "within" that price period.
    */
   constructor(
     logger,
@@ -35,7 +38,9 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
     invertPrice,
     priceFeedDecimals = 18,
     ohlcPeriod = 60, // One minute is CryptoWatch's most granular option.
-    twapLength = 0 // No TWAP by default.
+    twapLength = 0, // No TWAP by default.
+    historicalTimestampBuffer = 0 // Buffer of 0 means that historical timestamp must fall strictly within a price
+    // period's open and close time.
   ) {
     super();
     this.logger = logger;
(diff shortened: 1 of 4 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -105,31 +111,42 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
...
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return time < pricePeriod.closeTime && time >= pricePeriod.openTime;
+      return (
+        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
+        time >= pricePeriod.openTime - this.historicalTimestampBuffer
+      );
     });
 
+    // TODO: Add comments why `match` could be undefined, but the following code within the `if` statement
```
nicholaspai, line 127: @UMAprotocol/eng any idea why the below `match === undefined` situation could occur? The matching logic looks exactly the same as in above except for the `reverse()` call.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
```diff
@@ -759,6 +760,7 @@ const defaultConfigs = {
   OCEANUSD: {
     type: "medianizer",
     minTimeBetweenUpdates: 60,
+    historicalTimestampBuffer: 60,
```
nicholaspai, line 763: Should we add to other configs? or more slowly roll this out.

chrismaree: I would want to apply it to all financial products at once. however is this _enough_ of a buffer for ocean? it's a bit difficult when there is no volume at all in some of these pairs

nicholaspai: yeah, perhaps instead of updating every single config, we set the default buffer to 60, and then set ocean's to even wider since we've had problems with it specifically?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
mrice32: `match === undefined` happens when no candle is found where the timestamp falls in the range above.

Inside the if statement, there are two separate finds.

The first searches the array in reverse to find the last candle (or first in the reversed array) that falls before the timestamp in question. This is only half of the condition in the `find` above.

The second searches the array forwards to find the first candle that falls after the timestamp in question. Once again, this is only half of the inequality above.

I think the code formatting can be a little confusing, as it might make it look like these are one `find` call rather than two separate ones.

mrice32 (approved):
Looks great!!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js
nicholaspai: Ah I see all the logic inside the `if` statement is just aimed to provide the caller with more details

PR #3152, opened 2021-06-23
fix(LSP): Audit nits and cleanup

chrismaree:
**Motivation**

OZ identified a number of small issues during their audit. Most of these are typographical or semantic with major functionality issues identified. This PR implements the required changes to correct all errors identified in the audit.

**Summary**

Fixes the LSP contract post-OZ audit. All comments rased from OZ (as can be seen below) have been implemented in this PR.

**The OZ feedback on this set of contracts was as follows**:
The Long-Short-Pair template was developed and modified over several pull requests. Instead of reviewing the PRs individually, we reviewed the completed files. In particular, we reviewed the files in the [`long-short-pair` folder](https://github.com/UMAprotocol/protocol/tree/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair) and the [`long-short-pair-libraries` folder](https://github.com/UMAprotocol/protocol/tree/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries) at commit `b508f536ddfd94a79f93f633142bdc868b73461c`.

Our most important observation is that [the `setLongShortPairParameters` function of the `RangeBondLongShortPairFinancialProductLibrary` contract](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol#L56) allows anyone to overwrite the parameters at any point in time to maliciously change the payout structure.

The libraries are also inconsistent about whether the `computeExpiryTokensForCollateral` function should ensure the parameters have been set. In particular, the `BinaryOptionLongShortPairFinancialProductLibrary` [requires the strike](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol#L56) to have been set, but the other libraries do not perform the equivalent check. Consider ensuring the configuration parameters have been set in all libraries before calculating the appropriate collateral distribution.

We understand that the UMA team intend to include `nonReentrant` and `nonReentrantView` modifiers on all external and public functions. However, the modifier is missing from [the `create`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L165) and [the `getPositionTokens`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L264) functions of the `LongShortPair` contract. Moreover, it's missing from the `view` functions on all the libraries. Consider including the modifiers where appropriate.

The [`redeem` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L182) is restricted by the `preExpiration` modifier. However, if a user [settles](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L201) an equal number of long and short tokens, it will have the same effect (with a different event emitted) as token redemption, but it requires the user to wait for the price to resolve. Consider allowing users to call the `redeem` function at all times.

The [`_getSyntheticDecimals` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L138) is described and named as a `private` function, but it is `public`. Consider making it private.

The `createLongShortPair` function of the `LongShortPairCreator` contract [grants Minter and Burner roles](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L118-L124) for the synthetic tokens to the new `LongShortPair` contract. However, the creator contract still retains those roles. Consider renouncing them to reduce the attack surface.

The `create` function of the `LongShortPair` contract discards the return value when [minting synthetic tokens](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L170-L171). Similarly, the `settle` function discards the return value when [burning the tokens](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L221-L222). Consider requiring that the `mint` and `burnFrom` functions returns `true`.

The `computeExpiryTokensForCollateral` function of the `RangeBondLongShortPairFinancialProductLibrary` uses [a complicated closed-form expression](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol#L82) to determine how much collateral to assign to the long position. Instead, consider enumerating the cases explicitly so it is easier to reason about. This would reduce to something like:
```solidity
// (collateral tokens * lowPriceRange) is the notional value of the bond
// below the range, the collateral is worth less than this amount
// the long position is entitled to whatever is left
if(expiryPrice <= lowPriceRange) return 1;
// within the range, the long position is entitled to the notional value, 
// which is equal to the following fraction of collateral
if(expiryPrice <= highPriceRange) return lowPriceRange / expiryPrice;
// above the range, the long position is entitled to a fixed number of tokens
return lowPriceRange / highPriceRange;
```

The [`PositionSettled` event](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L75-L80), contains a typographical error in the second argument's identifier. It is presently `colllateralReturned`, when it should be `collateralReturned`. The subtle error could lead to issues with off-chain systems parsing the event details.

In `LongShortPairCreator.sol` there are no events emitted for the [creation of the `longToken`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L84-L89) or [the `shortToken`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L90-L95), nor are those addresses part of the [emitted `CreatedLongShortPair` event](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L126). Consider emitting events that contain the addresses of the new tokens to facilitate tracking.

The following functions have incomplete or incorrect Natural Specification comments. Consider correcting them:

- The [`createLongShortPair` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L66) of the `LongShortPairCreator` contract is missing its `@return` comment.
- The [`settle` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L201) incorrectly lists `collateralReturned` as a parameter (in addition to the returned value).

Some functions could benefit from renaming. These are our suggestions:

- The [`_getAddressWhitelist` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L299) of the `LongShortPair` contract could be renamed to `_getCollateralWhitelist`.
- The [`computeExpiryTokensForCollateral` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LongShortPairFinancialProductLibrary.sol#L11) of the `LongShortPairFinancialProductLibrary` contract could be renamed to `longPercentageCollateralAtExpiry` or similar.

The [`createLongShortPair` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L113) of the `LongShortPairCreator` contract and the [`create` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L168) of the `LongShortPair` contract both transfer funds on behalf of the message sender. The contracts implicitly assume that they have been granted an appropriate allowance. Consider documenting this in the function comments.

There are some misleading comments in the code base:

- The parameter setters in the library functions, for example the [`setLongShortPairParameters` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol#L36) in `BinaryOptionLongShortPairFinancialProductLibrary.sol`, refer to the non-existent `financialProduct` parameter instead of `LongShortPair`.
- The code base is inconsistent about using `1e18` to represent `100%`. For example, the [`expiryPercentLong` comment](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L50-L51) switches between `1e18` and `1` in the same comment. This is [repeated later in the contract](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L224-L225). Similarly, the `computeExpiryTokensForCollateral` functions in all the libraries [claim the function returns a number between 0 and 1](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol#L49), instead of between 0 and 1e18.
- The [`redeem` function's `@notice` comment](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L177) says `tokensToCreate` instead of `tokensToRedeem`. This comment would also be clearer if it started with "Redeem" instead of "Return".
- In the [`Testable` contract](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/common/implementation/Testable.sol) there is an [inline comment that says "If the contract is being run on the test network, then `timerAddress` will be the 0x0 address."](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/common/implementation/Testable.sol#L10) However, it appears that this comment is erroneous. The `onlyIfTest` modifier [requires that `timerAddress != address(0x0)`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/common/implementation/Testable.sol#L27). There are also other comments in the codebase that agree with the modifier: ["_timerAddress ... Set to 0x0 in production."](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L114). The comment in `Testable` should be modified to agree with the implementation to avoid potential confusion around such a critically important value.
- In `Timer.sol` the `@notice` docstring for the [`getCurrentTime` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/common/implementation/Timer.sol#L28) is misleading. It states, "Otherwise, it will return the block timestamp." However, the function is only capable of returning the stored value of `currentTime` and has no logic to conditionally return the block timestamp.

There are some unused imports in the code base:

- [`SafeMath.sol`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L4) in `LongShortPair.sol`
- [`IERC20Standard.sol`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L16) in `LongShortPair.sol`
- [`ContractCreator.sol`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L7) in `LongShortPairCreator.sol`
- [`AddressWhitelist.sol`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L9) in `LongShortPairCreator.sol`
- [`SyntheticToken.sol`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L12) in `LongShortPairCreator.sol`

Lastly, there are some typographical errors in the code base:

- [Line 55](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L55) of `LongShortPairCreator.sol` says "appended" instead of "prepended".
- [Line 57](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L57) of `LongShortPairCreator.sol` repeats the word "as".
- [Line 106](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L106) of `LongShortPair.sol` says "Requires mint and burn needed by this contract", which is ungrammatical.
- [Line 19](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol#L19) of `LinearLongShortPairFinancialProductLibrary.sol` has the extra word "be".
- Lines [35](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol#L35) and [41](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol#L41) of `LinearLongShortPairFinancialProductLibrary.sol` have the extra word "price".
- [Line 42](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol#L42) of `LinearLongShortPairFinancialProductLibrary.sol` has the extra word "a".
- [Line 63](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol#L63) of `LinearLongShortPairFinancialProductLibrary.sol` says "are" instead of "is".
- [Line 26](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/common/implementation/Lockable.sol#L26) of `Lockable.sol` reads "... and make it call a" when it should read "... and **making** it call a".

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/common/implementation/Lockable.sol
+++ b/packages/core/contracts/common/implementation/Lockable.sol
@@ -10,21 +10,18 @@ contract Lockable {
     bool private _notEntered;
 
     constructor() {
-        // Storing an initial non-zero value makes deployment a bit more
-        // expensive, but in exchange the refund on every call to nonReentrant
-        // will be lower in amount. Since refunds are capped to a percetange of
-        // the total transaction's gas, it is best to keep them low in cases
-        // like this one, to increase the likelihood of the full refund coming
-        // into effect.
+        // Storing an initial non-zero value makes deployment a bit more expensive, but in exchange the refund on every
+        // call to nonReentrant will be lower in amount. Since refunds are capped to a percentage of the total
+        // transaction's gas, it is best to keep them low in cases like this one, to increase the likelihood of the full
+        // refund coming into effect.
         _notEntered = true;
     }
 
     /**
      * @dev Prevents a contract from calling itself, directly or indirectly.
-     * Calling a `nonReentrant` function from another `nonReentrant`
-     * function is not supported. It is possible to prevent this from happening
-     * by making the `nonReentrant` function external, and make it call a
-     * `private` function that does the actual work.
+     * Calling a `nonReentrant` function from another `nonReentrant` function is not supported. It is possible to
+     * prevent this from happening by making the `nonReentrant` function external, and making it call a `private`
+     * function that does the actual state modification.
      */
     modifier nonReentrant() {
         _preEntranceCheck();
@@ -42,8 +39,9 @@ contract Lockable {
     }
 
     // Internal methods are used to avoid copying the require statement's bytecode to every `nonReentrant()` method.
-    // On entry into a function, `_preEntranceCheck()` should always be called to check if the function is being re-entered.
-    // Then, if the function modifies state, it should call `_postEntranceSet()`, perform its logic, and then call `_postEntranceReset()`.
+    // On entry into a function, `_preEntranceCheck()` should always be called to check if the function is being
+    // re-entered. Then, if the function modifies state, it should call `_postEntranceSet()`, perform its logic, and
+    // then call `_postEntranceReset()`.
     // View-only methods can simply call `_preEntranceCheck()` to make sure that it is not being re-entered.
     function _preEntranceCheck() internal view {
         // On the first call to nonReentrant, _notEntered will be true
--- a/packages/core/contracts/common/implementation/Testable.sol
+++ b/packages/core/contracts/common/implementation/Testable.sol
@@ -7,7 +7,7 @@ import "./Timer.sol";
  * @title Base class that provides time overrides, but only if being run in test mode.
  */
 abstract contract Testable {
-    // If the contract is being run on the test network, then `timerAddress` will be the 0x0 address.
+    // If the contract is being run in production, then `timerAddress` will be the 0x0 address.
     // Note: this variable should be set on construction and never modified.
     address public timerAddress;
 
--- a/packages/core/contracts/common/implementation/Timer.sol
+++ b/packages/core/contracts/common/implementation/Timer.sol
@@ -21,8 +21,7 @@ contract Timer {
     }
 
     /**
-     * @notice Gets the current time. Will return the last time set in `setCurrentTime` if running in test mode.
-     * Otherwise, it will return the block timestamp.
+     * @notice Gets the currentTime variable set in the Timer.
      * @return uint256 for the current Testable timestamp.
      */
     function getCurrentTime() public view returns (uint256) {
--- a/packages/core/contracts/financial-templates/common/TokenFactory.sol
+++ b/packages/core/contracts/financial-templates/common/TokenFactory.sol
@@ -24,8 +24,6 @@ contract TokenFactory is Lockable {
         uint8 tokenDecimals
     ) external nonReentrant() returns (ExpandedIERC20 newToken) {
         SyntheticToken mintableToken = new SyntheticToken(tokenName, tokenSymbol, tokenDecimals);
-        mintableToken.addMinter(msg.sender);
-        mintableToken.addBurner(msg.sender);
         mintableToken.resetOwner(msg.sender);
         newToken = ExpandedIERC20(address(mintableToken));
     }
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol
@@ -28,30 +28,30 @@ contract BinaryOptionLongShortPairFinancialProductLibrary is LongShortPairFinanc
 
     /**
      * @notice Enables any address to set the strike price for an associated binary option.
-     * @param LongShortPair address of the LSP.
+     * @param longShortPair address of the LSP.
      * @param strikePrice the strike price for the binary option.
      * @dev Note: a) Any address can set the initial strike price b) A strike can be 0.
      * c) A strike price can only be set once to prevent the deployer from changing the strike after the fact.
      * d) For safety, a strike price should be set before depositing any synthetic tokens in a liquidity pool.
-     * e) financialProduct must expose an expirationTimestamp method to validate it is correctly deployed.
+     * e) longShortPair must expose an expirationTimestamp method to validate it is correctly deployed.
      */
-    function setLongShortPairParameters(address LongShortPair, int256 strikePrice) public nonReentrant() {
-        require(ExpiringContractInterface(LongShortPair).expirationTimestamp() != 0, "Invalid LSP address");
-        require(!longShortPairParameters[LongShortPair].isSet, "Parameters already set");
+    function setLongShortPairParameters(address longShortPair, int256 strikePrice) public nonReentrant() {
+        require(ExpiringContractInterface(longShortPair).expirationTimestamp() != 0, "Invalid LSP address");
+        require(!longShortPairParameters[longShortPair].isSet, "Parameters already set");
 
-        longShortPairParameters[LongShortPair] = BinaryLongShortPairParameters({
+        longShortPairParameters[longShortPair] = BinaryLongShortPairParameters({
             isSet: true,
             strikePrice: strikePrice
         });
     }
 
     /**
-     * @notice Returns a number between 0 and 1 to indicate how much collateral each long and short token are entitled
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token are entitled
      * to per collateralPerPair.
      * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
      * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
      */
-    function computeExpiryTokensForCollateral(int256 expiryPrice)
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
         public
         view
         override
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
@@ -16,7 +16,7 @@ import "../../../../common/implementation/Lockable.sol";
  * expiryPercentLong = 1. if the price is lower than the lower bound then expiryPercentLong = 0. For example, consider
  * a linear LSP on the price of ETH collateralized in USDC with an upperBound = 4000 and lowerBound = 2000 with a
  * collateralPerPair of 1000 (i.e each pair of long and shorts is worth 1000 USDC). At settlement the expiryPercentLong
- * would equal 1 (each long worth 1000 and short worth 0) if ETH price was > 4000 and it would be equal 0 if < 2000
+ * would equal 1 (each long worth 1000 and short worth 0) if ETH price was > 4000 and it would equal 0 if < 2000
  * (each long is worthless and each short is worth 1000). If between the two (say 3500) then expiryPercentLong
  * = (3500 - 2000) / (4000 - 2000) = 0.75. Therefore each long is worth 750 and each short is worth 250.
  */
@@ -32,40 +32,40 @@ contract LinearLongShortPairFinancialProductLibrary is LongShortPairFinancialPro
     mapping(address => LinearLongShortPairParameters) public longShortPairParameters;
 
     /**
-     * @notice Enables any address to set the parameters price for an associated financial product.
-     * @param LongShortPair address of the LSP contract.
+     * @notice Enables any address to set the parameters for an associated financial product.
+     * @param longShortPair address of the LSP contract.
      * @param upperBound the upper price that the linear LSP will operate within.
      * @param lowerBound the lower price that the linear LSP will operate within.
      * @dev Note: a) Any address can set these parameters b) existing LSP parameters for address not set.
      * c) upperBound > lowerBound.
-     * d) parameters price can only be set once to prevent the deployer from changing the parameters after the fact.
-     * e) For safety, a parameters should be set before depositing any synthetic tokens in a liquidity pool.
-     * f) financialProduct must expose an expirationTimestamp method to validate it is correctly deployed.
+     * d) parameters can only be set once to prevent the deployer from changing the parameters after the fact.
+     * e) For safety, parameters should be set before depositing any synthetic tokens in a liquidity pool.
+     * f) longShortPair must expose an expirationTimestamp method to validate it is correctly deployed.
      */
     function setLongShortPairParameters(
-        address LongShortPair,
+        address longShortPair,
         int256 upperBound,
         int256 lowerBound
     ) public nonReentrant() {
-        require(ExpiringContractInterface(LongShortPair).expirationTimestamp() != 0, "Invalid LSP address");
+        require(ExpiringContractInterface(longShortPair).expirationTimestamp() != 0, "Invalid LSP address");
         require(upperBound > lowerBound, "Invalid bounds");
 
-        LinearLongShortPairParameters memory params = longShortPairParameters[LongShortPair];
+        LinearLongShortPairParameters memory params = longShortPairParameters[longShortPair];
         require(params.upperBound == 0 && params.lowerBound == 0, "Parameters already set");
 
-        longShortPairParameters[LongShortPair] = LinearLongShortPairParameters({
+        longShortPairParameters[longShortPair] = LinearLongShortPairParameters({
             upperBound: upperBound,
             lowerBound: lowerBound
         });
     }
 
     /**
-     * @notice Returns a number between 0 and 1 to indicate how much collateral each long and short token are entitled
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token is entitled
      * to per collateralPerPair.
      * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
      * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
      */
-    function computeExpiryTokensForCollateral(int256 expiryPrice)
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
         public
         view
         override
@@ -73,6 +73,7 @@ contract LinearLongShortPairFinancialProductLibrary is LongShortPairFinancialPro
         returns (uint256)
     {
         LinearLongShortPairParameters memory params = longShortPairParameters[msg.sender];
+        require(params.upperBound != 0 && params.lowerBound != 0, "Params not set for calling LSP");
 
         if (expiryPrice >= params.upperBound) return FixedPoint.fromUnscaledUint(1).rawValue;
 
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
@@ -73,7 +73,7 @@ contract LinearLongShortPairFinancialProductLibrary is LongShortPairFinancialPro
         returns (uint256)
     {
         LinearLongShortPairParameters memory params = longShortPairParameters[msg.sender];
-        require(params.upperBound != 0 && params.lowerBound != 0, "Params not set for calling LSP");
+        require(!(params.upperBound == 0 && params.lowerBound == 0), "Params not set for calling LSP");
 
         if (expiryPrice >= params.upperBound) return FixedPoint.fromUnscaledUint(1).rawValue;
 
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
@@ -73,7 +73,7 @@ contract LinearLongShortPairFinancialProductLibrary is LongShortPairFinancialPro
         returns (uint256)
     {
         LinearLongShortPairParameters memory params = longShortPairParameters[msg.sender];
-        require(!(params.upperBound == 0 && params.lowerBound == 0), "Params not set for calling LSP");
+        require(params.upperBound != 0 || params.lowerBound != 0, "Params not set for calling LSP");
 
         if (expiryPrice >= params.upperBound) return FixedPoint.fromUnscaledUint(1).rawValue;
 
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
@@ -85,7 +85,7 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
         returns (uint256)
     {
         RangeBondLongShortPairParameters memory params = longShortPairParameters[msg.sender];
-        require(params.highPriceRange != 0 && params.lowPriceRange != 0, "Params not set for calling LSP");
+        require(!(params.highPriceRange == 0 && params.lowPriceRange == 0), "Params not set for calling LSP");
 
         // This function's returns a value between 0 and 1e18 to be used in conjunction with the LSP collateralPerPair
         // that allocates collateral between the short and long tokens on expiry. This can be simplified by considering
@@ -101,6 +101,6 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
         if (positiveExpiryPrice <= params.highPriceRange)
             return FixedPoint.Unsigned(params.lowPriceRange).div(FixedPoint.Unsigned(positiveExpiryPrice)).rawValue;
         // 3) above the range, the long position is entitled to a fixed number of tokens.
-        else return FixedPoint.Unsigned(params.lowPriceRange).div(FixedPoint.Unsigned(params.highPriceRange)).rawValue;
+        return FixedPoint.Unsigned(params.lowPriceRange).div(FixedPoint.Unsigned(params.highPriceRange)).rawValue;
     }
 }
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
@@ -85,7 +85,7 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
         returns (uint256)
     {
         RangeBondLongShortPairParameters memory params = longShortPairParameters[msg.sender];
-        require(!(params.highPriceRange == 0 && params.lowPriceRange == 0), "Params not set for calling LSP");
+        require(params.highPriceRange != 0 || params.lowPriceRange != 0, "Params not set for calling LSP");
 
         // This function's returns a value between 0 and 1e18 to be used in conjunction with the LSP collateralPerPair
         // that allocates collateral between the short and long tokens on expiry. This can be simplified by considering
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
@@ -42,7 +42,7 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
 
     /**
      * @notice Enables any address to set the parameters price for an associated financial product.
-     * @param LongShortPair address of the LSP contract.
+     * @param longShortPair address of the LSP contract.
      * @param highPriceRange high price range after which the payout transforms from a yield dollar to a call option.
      * @param lowPriceRange low price range below which the payout transforms from a yield dollar to a short put option.
      * @dev between highPriceRange and lowPriceRange the contract will payout a fixed amount of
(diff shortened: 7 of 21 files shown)


packages/core/contracts/common/implementation/Lockable.sol
```diff
@@ -10,21 +10,18 @@ contract Lockable {
     bool private _notEntered;
 
     constructor() {
-        // Storing an initial non-zero value makes deployment a bit more
```
chrismaree, line 13: this was not related to the audit but the comment line length did not follow our standard.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol
```diff
@@ -28,30 +28,30 @@ contract BinaryOptionLongShortPairFinancialProductLibrary is LongShortPairFinanc
 
     /**
      * @notice Enables any address to set the strike price for an associated binary option.
-     * @param LongShortPair address of the LSP.
```
chrismaree, line 31: not audit-related. I noticed this inconsistency in all the libraries.

CarelessDoge (approved):

packages/core/contracts/financial-templates/common/TokenFactory.sol
```diff
@@ -24,8 +24,6 @@ contract TokenFactory is Lockable {
         uint8 tokenDecimals
     ) external nonReentrant() returns (ExpandedIERC20 newToken) {
         SyntheticToken mintableToken = new SyntheticToken(tokenName, tokenSymbol, tokenDecimals);
-        mintableToken.addMinter(msg.sender);
```
nicholaspai, line 27: This change should be tested in [TokenFactory.js](https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/financial-templates/TokenFactory.js), [EMPCreator.js](https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/financial-templates/expiring-multiparty/ExpiringMultiPartyCreator.js) and the [PerpCreator.js](https://github.com/UMAprotocol/protocol/blob/master/packages/core/test/financial-templates/perpetual-multiparty/PerpetualCreator.js) unit test files.

nicholaspai (commented):
High level comment: To aid reviewers, can you copy the suggestions from auditors into the PR description?

CarelessDoge (approved):

packages/core/contracts/financial-templates/common/TokenFactory.sol
chrismaree, line 27: @eng please take a look at this change. this came about after discussing with Nikesh that there is no point really to assign minter and burner to the caller if the caller has the owner role. The owner can set the mint and burn as needed. doing it this way is more conservative as to who we assign these roles to and makes it harder to leave unintended permissions, as was the case in the original factory.

packages/core/contracts/common/implementation/Testable.sol
```diff
@@ -7,7 +7,7 @@ import "./Timer.sol";
  * @title Base class that provides time overrides, but only if being run in test mode.
  */
 abstract contract Testable {
-    // If the contract is being run on the test network, then `timerAddress` will be the 0x0 address.
```
chrismaree, line 10: OZ comment:
```
In the Testable contract there is an inline comment that says "If the contract is being run on the test network, then timerAddress will be the 0x0 address." However, it appears that this comment is erroneous. The onlyIfTest modifier requires that timerAddress != address(0x0). There are also other comments in the codebase that agree with the modifier: "_timerAddress ... Set to 0x0 in production.". The comment in Testable should be modified to agree with the implementation to avoid potential confusion around such a critically important value.
```

packages/core/contracts/common/implementation/Timer.sol
```diff
@@ -21,8 +21,7 @@ contract Timer {
     }
 
     /**
-     * @notice Gets the current time. Will return the last time set in `setCurrentTime` if running in test mode.
-     * Otherwise, it will return the block timestamp.
+     * @notice Gets the currentTime variable set in the Timer.
```
chrismaree, line 24: OZ comment:
```
In Timer.sol the @notice docstring for the getCurrentTime function is misleading. It states, "Otherwise, it will return the block timestamp." However, the function is only capable of returning the stored value of currentTime and has no logic to conditionally return the block timestamp.

File: packages/core/contracts/financial-templates/common/TokenFactory.sol (hunk 3 above)
Comment (chrismaree) [reply packages/core/contracts/financial-templates/common/TokenFactory.sol:27]: done

File: packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol [hunk 6]
```diff
@@ -28,30 +28,30 @@ contract BinaryOptionLongShortPairFinancialProductLibrary is LongShortPairFinanc
...
     }
 
     /**
-     * @notice Returns a number between 0 and 1 to indicate how much collateral each long and short token are entitled
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token are entitled
      * to per collateralPerPair.
      * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
      * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
      */
-    function computeExpiryTokensForCollateral(int256 expiryPrice)
```
chrismaree, line 54: OZ comment on function renaming:
```
The computeExpiryTokensForCollateral function of the LongShortPairFinancialProductLibrary contract could be renamed to longPercentageCollateralAtExpiry or similar.
```

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
```diff
@@ -51,59 +51,56 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
...
         public
         view
         override
         nonReentrantView()
         returns (uint256)
     {
         RangeBondLongShortPairParameters memory params = longShortPairParameters[msg.sender];
+        require(params.highPriceRange != 0 && params.lowPriceRange != 0, "Params not set for calling LSP");
 
-        // expiryPercentLong=[min(max(1/R2,1/P),1/R1)]/(1/R1)
```
chrismaree, line 89: OZ comment on simplifying some logic in this computation:
The `computeExpiryTokensForCollateral` function of the `RangeBondLongShortPairFinancialProductLibrary` uses [a complicated closed-form expression](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol#L82) to determine how much collateral to assign to the long position. Instead, consider enumerating the cases explicitly so it is easier to reason about. This would reduce to something like:
```solidity
// (collateral tokens * lowPriceRange) is the notional value of the bond
// below the range, the collateral is worth less than this amount
// the long position is entitled to whatever is left
if(expiryPrice <= lowPriceRange) return 1;
// within the range, the long position is entitled to the notional value, 
// which is equal to the following fraction of collateral
if(expiryPrice <= highPriceRange) return lowPriceRange / expiryPrice;
// above the range, the long position is entitled to a fixed number of tokens
return lowPriceRange / highPriceRange;
```

packages/core/contracts/common/implementation/Timer.sol
nicholaspai: Oh wow, great catch OZ.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -220,17 +212,17 @@ contract LongShortPair is Testable, Lockable {
...
             contractState = ContractState.ExpiredPriceReceived;
         }
 
-        longToken.burnFrom(msg.sender, longTokensToRedeem);
-        shortToken.burnFrom(msg.sender, shortTokensToRedeem);
+        require(longToken.burnFrom(msg.sender, longTokensToRedeem));
+        require(shortToken.burnFrom(msg.sender, shortTokensToRedeem));
 
-        // expiryPercentLong is a number between 0 and 1e18. 0 means all collateral goes to short tokens and
-        // 1 means all collateral goes to the long token. Total collateral returned is the sum of payouts.
```
chrismaree, line 233: OZ comment on misleading comments:
- The code base is inconsistent about using `1e18` to represent `100%`. For example, the [`expiryPercentLong` comment](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L50-L51) switches between `1e18` and `1` in the same comment. This is [repeated later in the contract](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L224-L225). Similarly, the `computeExpiryTokensForCollateral` functions in all the libraries [claim the function returns a number between 0 and 1](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.sol#L49), instead of between 0 and 1e18.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -304,7 +296,7 @@ contract LongShortPair is Testable, Lockable {
         return IdentifierWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.IdentifierWhitelist));
     }
 
-    function _getAddressWhitelist() internal view returns (AddressWhitelistInterface) {
```
chrismaree, line 307: OZ comments on naming:

The [`_getAddressWhitelist` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L299) of the `LongShortPair` contract could be renamed to `_getCollateralWhitelist`.

packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol
```diff
@@ -47,19 +49,21 @@ contract LongShortPairCreator is Testable, Lockable {
...
 
     /**
+     * @notice Creates a longShortPair contract and associated long and short tokens.
+     * @dev The caller must approve this contract to transfer `prepaidProposerReward` amount of collateral.
      * @param expirationTimestamp unix timestamp of when the contract will expire.
      * @param collateralPerPair how many units of collateral are required to mint one pair of synthetic tokens.
      * @param priceIdentifier registered in the DVM for the synthetic.
      * @param syntheticName Name of the synthetic tokens to be created. The long tokens will have "Long Token" appended
      *     to the end and the short token will "Short Token" appended to the end to distinguish within the LSP's tokens.
-     * @param syntheticSymbol Symbol of the synthetic tokens to be created. The long tokens will have "l" appended
```
chrismaree, line 55: OZ typographical comments

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
nicholaspai: +1 for including this comment. Wasn't entirely sure at first what was going on but i figured you just refactored for readability

packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol
```diff
@@ -123,15 +127,15 @@ contract LongShortPairCreator is Testable, Lockable {
         shortToken.addBurner(lspAddress);
         shortToken.resetOwner(lspAddress);
 
-        emit CreatedLongShortPair(lspAddress, msg.sender);
+        emit CreatedLongShortPair(lspAddress, msg.sender, address(longToken), address(shortToken));
```
chrismaree, line 131: OZ comment on events

In `LongShortPairCreator.sol` there are no events emitted for the [creation of the `longToken`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L84-L89) or [the `shortToken`](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L90-L95), nor are those addresses part of the [emitted `CreatedLongShortPair` event](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L126). Consider emitting events that contain the addresses of the new tokens to facilitate tracking.

packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol
```diff
@@ -123,15 +127,15 @@ contract LongShortPairCreator is Testable, Lockable {
...
 
         return lspAddress;
     }
 
     // IERC20Standard.decimals() will revert if the collateral contract has not implemented the decimals() method,
     // which is possible since the method is only an OPTIONAL method in the ERC20 standard:
     // https://eips.ethereum.org/EIPS/eip-20#methods.
-    function _getSyntheticDecimals(IERC20Standard _collateralToken) public view returns (uint8 decimals) {
```
chrismaree, line 134: OZ comment on decimals method:

The [`_getSyntheticDecimals` function](https://github.com/UMAprotocol/protocol/blob/b508f536ddfd94a79f93f633142bdc868b73461c/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L138) is described and named as a `private` function, but it is `public`. Consider making it private.

packages/core/test/financial-templates/TokenFactory.js
```diff
@@ -38,13 +39,12 @@ contract("TokenFactory", function (accounts) {
     });
     const token = await Token.at(tokenAddress);
 
-    // Creator should be only minter
```
chrismaree, line 41: update to tests to match the new `TokenFactory` role allocation.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
```diff
@@ -32,47 +32,48 @@ contract LinearLongShortPairFinancialProductLibrary is LongShortPairFinancialPro
...
-    function computeExpiryTokensForCollateral(int256 expiryPrice)
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
         public
         view
         override
         nonReentrantView()
         returns (uint256)
     {
         LinearLongShortPairParameters memory params = longShortPairParameters[msg.sender];
+        require(params.upperBound != 0 && params.lowerBound != 0, "Params not set for calling LSP");
```
nicholaspai, line 76: Technically, the `lowerBound` can be 0, as long as the `upperBound` is > 0 right? The correct check should be `!(params.upperBound == 0 && params.lowerBound == 0)` I think.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
```diff
@@ -51,59 +51,56 @@ contract RangeBondLongShortPairFinancialProductLibrary is LongShortPairFinancial
...
-    function computeExpiryTokensForCollateral(int256 expiryPrice)
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
         public
         view
         override
         nonReentrantView()
         returns (uint256)
     {
         RangeBondLongShortPairParameters memory params = longShortPairParameters[msg.sender];
+        require(params.highPriceRange != 0 && params.lowPriceRange != 0, "Params not set for calling LSP");
```
nicholaspai, line 88: Same comment [here](https://github.com/UMAprotocol/protocol/pull/3152/files#r657266212)

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -72,12 +70,7 @@ contract LongShortPair is Testable, Lockable {
     event TokensCreated(address indexed sponsor, uint256 indexed collateralUsed, uint256 indexed tokensMinted);
     event TokensRedeemed(address indexed sponsor, uint256 indexed collateralReturned, uint256 indexed tokensRedeemed);
     event ContractExpired(address indexed caller);
-    event PositionSettled(
-        address indexed sponsor,
-        uint256 colllateralReturned,
-        uint256 longTokens,
-        uint256 shortTokens
-    );
+    event PositionSettled(address indexed sponsor, uint256 collateralReturned, uint256 longTokens, uint256 shortTokens);
```
mrice32, line 73: What changed here?

mrice32 (commented):
Looks awesome! A few comments around how we're using fixed point in the library.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/RangeBondLongShortPairFinancialProductLibrary.sol
chrismaree: 100%. This form is the same that I showed at the end of the LSP presentation a few weeks ago where you transform the equation into a piecewise function that looks something like this:
image

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
chrismaree: @nicholaspai your require is the same as the implementation, no?

@mrice32 that prevents the lower bound from being 0 which is a valid state!

chrismaree: However, I'm happy to go with your syntax @nicholaspai. I've updated accordingly.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: `colllateralReturned`->`collateralReturned` (typo!)

mrice32: Ah, thanks!

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/LinearLongShortPairFinancialProductLibrary.sol
nicholaspai: @chrismaree your require would return false if the lowerBound = 0 which is a valid state. 

However I suppose this is a non-problem because somewhere else you require upperBound > lowerBound. 

Wdyt

chrismaree: nope, it would not. the way it's set up now would return true if `lowerBound=0`, assuming `upperBound !=0`. The expression (like Matt pointed out) `params.upperBound != 0 || params.lowerBound != 0` is equivalent to what I had before which is `params.upperBound != 0 && params.lowerBound != 0` which basically will _only_ return false if both are 0. 

If the lower bound is 0 it's fine assuming the upper bound is not also 0.

do you agree?

PR #3153, opened 2021-06-23
feat(api): add usd price conversions to synthetic prices

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1492/convert-synthetic-prices-to-usd

**Summary**

In order to calculate TVM, we need all synthetics priced in USD.  Previously we only stored the native synthetic value, which may or may not be valued against usd. PR adds the feature as well as refactors some api calls. 

This adds a conversion utility which converts to usd based on synthetic price and collateral price.  USD synthetic prices are stored in the prices table, keyed by token address. API is refactored to make querying this data a bit easier. 

**Details**
This refactors and updates docs for several api endpoints and all prices are returned as USD in wei. The main pattern here is that it allows you to query by erc20 address or emp address when specifying either collateral or synthetic:

**latest prices**
- latestPriceByTokenAddress: price by erc20
- latestCollateralPrice: collateral price by emp
- latestSyntheticPrice: synth price by emp

**historical prices between** 
- historicalPricesByTokenAddress: historical price by erc20 between dates
- historicalSynthPrices: synth historical price by emp between dates
- historicalCollateralPrices: collateral historical price by emp between dates

**historical prices slice**
- sliceHistoricalPricesByTokenAddress: historical price by erc20 by start date and length
- sliceHistoricalSynthPrices: synth historical price by emp by start date and length
- sliceHistoricalCollateralPrices: collateral historical price by emp by start date and length

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1492/convert-synthetic-prices-to-usd

### DIFF
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -1,13 +1,51 @@
+import assert from "assert";
 // allow for more complex queries, joins and shared queries between services
-import type { AppState, CurrencySymbol } from "..";
-import uma from "@uma/sdk";
+import type { AppState, CurrencySymbol, PriceSample } from "..";
+import * as uma from "@uma/sdk";
 import { calcGcr } from "./utils";
-import Promise from "bluebird";
+import bluebird from "bluebird";
 import { BigNumber } from "ethers";
 
-type Dependencies = Pick<AppState, "erc20s" | "emps" | "stats" | "registeredEmps">;
+const { exists } = uma.utils;
+type Dependencies = Pick<AppState, "erc20s" | "emps" | "stats" | "registeredEmps" | "prices">;
 
 export default (appState: Dependencies) => {
+  const { prices } = appState;
+
+  async function historicalPricesByTokenAddress(
+    address: string,
+    start = 0,
+    end: number = Date.now(),
+    currency: CurrencySymbol = "usd"
+  ): Promise<PriceSample[]> {
+    assert(start >= 0, "requires a start value >= 0");
+    assert(exists(prices[currency]), "invalid currency type: " + currency);
+    assert(exists(prices[currency].history[address]), "no prices for address" + address);
+    const results = await prices[currency].history[address].betweenByTimestamp(start, end);
+    // convert this to tuple to save bytes.
+    return results.map(({ price, timestamp }) => [timestamp, price]);
+  }
+
+  async function sliceHistoricalPricesByTokenAddress(
+    address: string,
+    start = 0,
+    length = 1,
+    currency: "usd" = "usd"
+  ): Promise<PriceSample[]> {
+    assert(start >= 0, "requires a start value >= 0");
+    assert(exists(prices[currency]), "invalid currency type: " + currency);
+    assert(exists(prices[currency].history[address]), "no prices for address" + address);
+    const results = await prices[currency].history[address].sliceByTimestamp(start, length);
+    // convert this to tuple to save bytes.
+    return results.map(({ price, timestamp }) => [timestamp, price]);
+  }
+  async function latestPriceByTokenAddress(address: string, currency: CurrencySymbol = "usd") {
+    assert(address, "requires an erc20 token address");
+    assert(exists(prices[currency]), "invalid currency type: " + currency);
+    const priceSample = prices[currency].latest[address];
+    assert(exists(priceSample), "No price for address: " + address);
+    return priceSample;
+  }
   async function getAnyEmp(empAddress: string) {
     if (await appState.emps.active.has(empAddress)) {
       return appState.emps.active.get(empAddress);
@@ -42,15 +80,15 @@ export default (appState: Dependencies) => {
 
   async function listActiveEmps() {
     const emps = appState.emps.active.values();
-    return Promise.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
+    return bluebird.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
   }
   async function listExpiredEmps() {
     const emps = appState.emps.expired.values();
-    return Promise.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
+    return bluebird.map(emps, (emp) => getFullEmpState(emp).catch(() => emp));
   }
 
   async function sumTvl(addresses: string[], currency: CurrencySymbol = "usd") {
-    const tvl = await Promise.reduce(
+    const tvl = await bluebird.reduce(
       addresses,
       async (sum, address) => {
         const stats = await appState.stats[currency].latest.getOrCreate(address);
@@ -72,5 +110,8 @@ export default (appState: Dependencies) => {
     listExpiredEmps,
     totalTvl,
     sumTvl,
+    latestPriceByTokenAddress,
+    historicalPricesByTokenAddress,
+    sliceHistoricalPricesByTokenAddress,
   };
 };
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -30,7 +30,7 @@ export default (appState: Dependencies) => {
     address: string,
     start = 0,
     length = 1,
-    currency: "usd" = "usd"
+    currency: CurrencySymbol = "usd"
   ): Promise<PriceSample[]> {
     assert(start >= 0, "requires a start value >= 0");
     assert(exists(prices[currency]), "invalid currency type: " + currency);
--- a/packages/api/src/libs/utils.test.ts
+++ b/packages/api/src/libs/utils.test.ts
@@ -1,5 +1,8 @@
 import * as utils from "./utils";
 import assert from "assert";
+import { ethers } from "ethers";
+
+const { parseUnits } = ethers.utils;
 
 describe("utils", function () {
   it("asyncValues", async function () {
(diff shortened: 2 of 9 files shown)


packages/api/src/libs/utils.ts
```diff
@@ -40,12 +45,34 @@ export function calcGcr(
...
   assert(uma.utils.exists(totalPositionCollateral), "requires total position collateral");
   assert(uma.utils.exists(collateralDecimals), "requires collateralDecimals");
   const normalizedCollateral = uma.utils.ConvertDecimals(collateralDecimals, 18)(totalPositionCollateral);
-  return parseUnits(price).mul(normalizedCollateral).div(parseUnits("1"));
+  return BigNumber.from(price).mul(normalizedCollateral).div(ETH);
+}
+
+// collateralPrice is 1 collateral in USD, assumes price in wei
+// syntheticPrice is 1 synthetic to collateral, assumes price in wei
+export function calcSyntheticValue(
```
daywiss, line 61: this function needs some more eyes on it. This basically takes a synthetic price as returned from a bot price feed and a collateral price priced in USD, and comes up with a synthetic price in USD.  This was adapted from the affiliates logic. I guess the underlying assumption here is that a synthetic price as returned from the bot will always be against its collateral.

packages/api/src/libs/queries.ts
```diff
@@ -1,13 +1,51 @@
+import assert from "assert";
 // allow for more complex queries, joins and shared queries between services
-import type { AppState, CurrencySymbol } from "..";
-import uma from "@uma/sdk";
+import type { AppState, CurrencySymbol, PriceSample } from "..";
+import * as uma from "@uma/sdk";
 import { calcGcr } from "./utils";
-import Promise from "bluebird";
+import bluebird from "bluebird";
```
daywiss, line 6: typescript doesnt like when this is named "Promise", i've found there are type incompatibilities

packages/api/src/services/actions.ts
```diff
@@ -47,48 +47,47 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
...
-      currency: "usd" = "usd"
-    ): Promise<PriceSample[]> {
-      assert(start >= 0, "requires a start value >= 0");
-      assert(exists(prices[currency]), "invalid currency type: " + currency);
-      assert(exists(prices[currency].history[address]), "no prices for address" + address);
-      const results = await prices[currency].history[address].sliceByTimestamp(start, length);
-      // convert this to tuple to save bytes.
-      return results.map(({ price, timestamp }) => [timestamp, price]);
+    // get prices by token address
+    latestPriceByTokenAddress: queries.latestPriceByTokenAddress,
```
daywiss, line 51: these calls were refactored to be more clear on specifying token address vs emp address when getting a price. The calls essentially allow someone to query latest price or price  history by either emp (collateral or synthetic) or the actual erc20 address of the token

packages/api/src/services/synthetic-prices.ts
```diff
@@ -49,24 +58,99 @@ export default function (config: Config, appState: Dependencies) {
     return Promise.allSettled(addresses.map(updatePriceHistory));
   }
 
-  async function updateLatestPrice(empAddress: string) {
+  // updates raw synth price, in relation to collateral
+  async function updateLatestSynthPrice(empAddress: string) {
```
daywiss, line 62: this update adds another type of synthetic price, originally it was just the price pulled from the bot price feed, but in order to make this useful, we also have to convert it to USD. Thats what this PR adds, is the raw price conversion to usd.

packages/api/src/libs/utils.test.ts
```diff
@@ -60,11 +63,19 @@ describe("utils", function () {
   });
   it("calculateTvl", function () {
     const emp = {
-      totalPositionCollateral: "1000000000",
+      totalPositionCollateral: parseUnits("100", 8).toString(),
```
nicholaspai, line 66: tip: can also grab `parseFixed` from `@uma/common` which is based on another implementation from the [`ethers` project](https://github.com/ethers-io/ethers.js/tree/master/packages/bignumber)

packages/api/src/services/actions.ts
nicholaspai: Its a bit confusing that the function name is `latestSyntheticPrice` but you can query the EMP's collateral price?

nicholaspai: Nice, this name is much clearer

packages/api/src/services/collateral-prices.ts
```diff
@@ -4,6 +4,8 @@ import { AppState, CurrencySymbol } from "..";
 type Config = {
   currency?: CurrencySymbol;
 };
+import { parseUnits } from "../libs/utils";
```
nicholaspai, line 7: Same comment as above about `@uma/common`'s `parseFixed` method.

nicholaspai (commented):
Looks great, have some nits

packages/api/src/services/synthetic-prices.ts
```diff
@@ -49,24 +59,101 @@ export default function (config: Config, appState: Dependencies) {
...
+
+    // use the most recent timestamp to index this price
+    const timestamp = Math.max(collateralTimestamp, synthTimestamp);
+    prices[currency].latest[emp.tokenCurrency] = [timestamp, price.toString()];
+    return [timestamp, price];
+  }
+
   async function updateLatestPrices(empAddresses: string[]) {
-    return Promise.allSettled(empAddresses.map(updateLatestPrice));
+    // slow down updates by running them serially to prevent rate limits, conform to Promise.allSettled
```
mrice32, line 124: Rather than making each component serialize requests (but only its requests), would it be better to have a global manager for coingecko requests (and any others that are rate sensitive)? The manager can receive inbound requests and ensure a global limit on the number of coingecko requests, for instance.

The reason I suggest this is two fold:
1. Guaranteeing cross-component synchronization. Is there only one place where we talk to coingecko? What happens if the ingestion pipelines starts a second update before the first finishes? All-in-all, it seems tricky to try to put a global rate limit by having each component run the low-level requests that they issue in series unless we're somehow guaranteeing that only one component talks to CG at any one time.
2. This type of throttling depends on the API's response speed. If the API responds quickly, we could go over our per-second quota, which makes this strategy somewhat unpredictable.

packages/api/src/libs/utils.ts
```diff
@@ -40,14 +45,19 @@ export function calcGcr(
...
+  // price needs to be in wei
   price: string,
   emp: Pick<uma.tables.emps.Data, "totalPositionCollateral" | "collateralDecimals">
 ) {
   const { totalPositionCollateral, collateralDecimals } = emp;
   assert(uma.utils.exists(totalPositionCollateral), "requires total position collateral");
   assert(uma.utils.exists(collateralDecimals), "requires collateralDecimals");
   const normalizedCollateral = uma.utils.ConvertDecimals(collateralDecimals, 18)(totalPositionCollateral);
-  return parseUnits(price).mul(normalizedCollateral).div(parseUnits("1"));
+  return BigNumber.from(price).mul(normalizedCollateral).div(ETH);
```
mrice32, line 56: nit: why the name ETH? Might be a little confusing since it seems unrelated to the currency ETH? Thoughts on `SCALING_CONSTANT` or `SCALING_MULTIPLIER` instead?

mrice32 (commented):
Looks good! A few nits and a high-level comment

packages/api/src/services/synthetic-prices.ts
daywiss: yea i could see something like this, essentially a wrapper around the coingecko library which automatically rate limits requests.  If the coingecko api returns data on how many requests we have left, that could make it even nicer, as we could throttle calls such that we make them as fast as we can until the reset.  It would be challenging though to generalize this for any type of request.

> Guaranteeing cross-component synchronization. Is there only one place where we talk to coingecko? What happens if the ingestion pipelines starts a second update before the first finishes?  

Currently there is only a single service which depends on coingecko, the `collateral-prices` service. This is by design to avoid some of the problems you are illustrating. There is definitely a disconnect between the interval at which updates happen, and how long they could take, which is something that could be improved.  We can guarantee non overlapping intervals by creating a looping function which begins the timer once all processing is complete.

So what i think i am getting out of this is have a more generalized coingecko rate limiter and be more careful around the frequency of updates to insure they dont overlap. is that right?

packages/api/src/services/collateral-prices.ts
daywiss: cool thanks, i just tried to do this, but im getting typescript errors. Apparently the version in common is not the same as what im using in this project, and there are changes to the api.  Going to leave it as is for now to avoid having to refactor a bunch of stuff related to this!

packages/api/src/services/synthetic-prices.ts
daywiss: i ended up adding that async loop function in `apps/api/app.ts`. The idea behind this is that all price updates must finish running before the timer for the next iteration will start.  As for coingecko price rate limit, i looked into this, the api doesnt really return requests remaining or anything like that, but i can make a card for generalizing rate limiting for coingecko and beyond

mrice32: Hmmmm, do you think the top-level looping limit really is necessary if the rate limit problem is fixed? I might argue it's unnecessary complexity as long as we aim to solve the underlying rate limiting issue.

Personally, I think we should add a card, but have no top-level delay. Thoughts?

daywiss: I dont see how we can not have a looping query to coingecko, regardless of if we have native rate limiting.  If the argument is you move the looping to a different spot, that is something we could discuss, though personally i really like controlling all loops at the top level.  lets chat some more on this, it sounds like you have a particular design in mind.

mrice32 (approved):

PR #3154, opened 2021-06-24
feat(common): Initial PR adding Optimism plugin to Hardhat config

nicholaspai:
**Motivation**

The main purpose of this PR is to kickoff Optimism integration into our Hardhat infrastructure. The neccessary steps for full hardhat support (a prerequisite for implementing contracts to deploy to the OVM) are:
1) Can compile OVM-compatible contracts
2) Can test against local Optimism node
3) Can deploy to OVM
4) Can run Optimism node in circle CI

This PR implements (1) and (2) above.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


package.json
```diff
@@ -43,6 +43,7 @@
     "eslint-plugin-prettier": "^3.4.0",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^3.0.0",
+    "ethereum-waffle": "^3.4.0",
```
nicholaspai, line 46: Although technically `ethereum-waffle` is a full Ethereum (compile, deploy, test) toolkit, we import it here because it actually adds some great ["chai matchers"](https://ethereum-waffle.readthedocs.io/en/latest/matchers.html#) which make tests really readable. I added it here for demonstration purposes because the [optimism-tutorial](https://github.com/OpenZeppelin/openzeppelin-contracts) uses as an auxillary package to `hardhat` which they use to compile and deploy.

packages/common/package.json
```diff
@@ -30,12 +30,15 @@
     "browser.js"
   ],
   "dependencies": {
+    "@eth-optimism/hardhat-ovm": "^0.2.2",
     "@ethersproject/bignumber": "^5.0.5",
     "@google-cloud/kms": "^0.3.0",
     "@google-cloud/storage": "^2.4.2",
     "@maticnetwork/maticjs": "^2.0.40",
+    "@nomiclabs/hardhat-ethers": "npm:hardhat-deploy-ethers@^0.3.0-beta.10",
```
nicholaspai, line 38: I've had trouble getting `getAbi` or `getArtifact` to work when running scripts on `optimism`:

```
TypeError: artifacts.require is not a function
      at getTruffleContractTest (dist/index.js:108:21)
      at getAbiTest (dist/index.js:144:29)
      at Context.<anonymous> (test/optimism/Finder.js:26:41)
      at processImmediate (internal/timers.js:461:21)
```

nicholaspai: Perhaps this is because `compile --network optimism` builds contracts in `artifacts-ovm` and caches compile information in `cache-ovm`

packages/common/src/HardhatConfig.js
```diff
@@ -67,6 +73,28 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
         url: getNodeUrl("mainnet", true),
         accounts: { mnemonic },
       },
+      optimism: {
+        url: "http://127.0.0.1:8545",
```
nicholaspai, line 77: `optimism` needs to run against a local node. The best way to run a node currently is to pull a docker container. Instructions [here](https://github.com/ethereum-optimism/optimism#development-quick-start).

We'll need to think about how to integrate this into our CI test. Note that building the docker container took me ~10 minutes so we need to think of a different solution than building it on each run.

packages/common/src/HardhatConfig.js
```diff
@@ -67,6 +73,28 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
+        accounts: { mnemonic },
+        // This sets the gas price to 0 for all transactions on L2. We do this because account balances are not yet
+        // automatically initiated with an ETH balance.
+        gasPrice: 0,
+        // This sets the network as using the ovm and ensure contract will be compiled against that.
+        ovm: true,
+        ignoreRxList: [
```
nicholaspai, line 84: I ignore every single directory right now because the Optimism-solc fork does not work exactly the same as solc. Certain Solidity code cannot be compiled so easily, such as inline assembly. Additionally, the logs do not make it clear which contracts are failing to compile, so for demonstration purposes I ignore every directory except the `optimism` one.

Here's an example compile error:

```
yarn hardhat compile --network optimism --show-stack-traces
yarn run v1.22.10
$ /Users/nicholaspai/UMA/protocol/node_modules/.bin/hardhat compile --network optimism --show-stack-traces

Compiling 191 files with 0.8.4
OVM Compiler Error (insert "// @unsupported: ovm" if you don't want this file to be compiled for the OVM):
 InternalCompilerError: Assembly exception for bytecode


Error HH600: Compilation failed
```

packages/core/.gitignore
```diff
@@ -1,5 +1,7 @@
 cache/
+cache-ovm/
 artifacts/
+artifacts-ovm/
```
nicholaspai, line 4: Contracts are built to new folders when compiled with `--network optimism`, which is probably why `getAbi`, `artifacts.require(...)`, and `.deployed()` all fail right now when testing a script on `--network optimism`

packages/core/contracts/optimism/FinderFlattened.sol
```diff
@@ -0,0 +1,156 @@
+/**
+ * @notice This contract was flattened so that it no longer imports directly or indirectly contracts that cannot be
```
nicholaspai, line 2: This is just a flattened Finder.sol that strips out external contract imports so that I can compile it successfully.

packages/core/test/optimism/Finder.js
```diff
@@ -0,0 +1,57 @@
+/* External Imports */
+const { web3, ethers } = require("hardhat");
```
nicholaspai, line 2: demonstration script that can be run against `--network optimism` or `--network hardhat`. Things to look at:
- `waffle`: The chai matchers like `expect(tx).to.be.revertedWith("error message")` come with waffle, which is pretty cool.

package.json
nicholaspai: Question is whether we should adopt `waffle` into our repo, so this PR demonstrates its potential usage.

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -0,0 +1,33 @@
+const path = require("path");
+
+// This file is mostly taken from the modified `compile` task file written by Synthetix: https://github.com/Synthetixio/synthetix
+
+const { internalTask } = require("hardhat/config");
+const { TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS } = require("hardhat/builtin-tasks/task-names");
+
+const CONTRACTS_DIR = path.resolve(__dirname, "../../../../core/contracts");
+
+// This overrides a buidler internal task, which is part of its compile task's lifecycle.
```
mrice32, line 10:
```suggestion
// This overrides a hardhat internal task, which is part of its compile task's lifecycle.
```

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -0,0 +1,33 @@
...
+// This allows us to filter on whitelisted, OVM-compatible contracts from the compilation list,
+// which are entries in extended `compileWhitelist` configuration.
+internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network }, runSuper) => {
```
mrice32, line 13: Is this supposed to run just for optimism compilations or will this run for all of our compilations? If just optimism, where is that specified?

packages/common/package.json
```diff
@@ -30,12 +30,15 @@
...
     "@nomiclabs/hardhat-etherscan": "^2.1.3",
     "@nomiclabs/hardhat-truffle5": "^2.0.0",
+    "@nomiclabs/hardhat-waffle": "^2.0.1",
```
mrice32, line 41: Do you see any reason why hardhat-related dependencies need to be exported in the package? Could we move them all down to dev dependencies. Waffle especially seems like a dev dependency since it's only used in tests.

I only mention this because every dependency we add here gets built into every frontend or node app that uses our contracts.

packages/core/contracts/optimism/external/openzeppelin/contracts/access/Ownable.sol
```diff
@@ -0,0 +1,69 @@
+// SPDX-License-Identifier: MIT
```
mrice32, line 1: If we have the ability to whitelist directories for the OVM, why can't we just whitelist the OZ node modules directories/contracts that we want rather than adding a second set of OZ contracts in the repo?

packages/core/test/optimism/Finder.js
mrice32: This is cool!

package.json
mrice32: Does this mean we need to use ethers wherever we use waffle?

If so, does this mean we need to support ethers everywhere? Not sure how much overhead it would add to trying to support both ethers and web3.

nicholaspai: Yeah it seems waffle + ethers are linked. I think i'm going to remove waffle and just use web3

packages/common/src/hardhat/tasks/compile.js
nicholaspai: I see no reason why this couldn't be used for any network, which is how its currently configured.

packages/core/package.json
```diff
@@ -76,7 +76,7 @@
     "mocha-test": "mocha ./mocha-test --recursive",
     "load-addresses": "yarn run apply-registry",
     "clean": "rm -rf build",
-    "build": "yarn truffle compile && yarn load-addresses && yarn generate-contract-types && yarn compile-ts && yarn buildLatestHardhatVersionHashes",
+    "build": "yarn truffle compile && yarn load-addresses && yarn generate-contract-types && yarn compile-ts && yarn buildLatestHardhatVersionHashes && yarn hardhat compile --network optimism",
```
nicholaspai, line 79: Added the `compile --network optimism` task to `build`. We don't test this (as this would require running a local optimism node, which is not easy to do yet since the only way is to run 6 docker containers), but worth just compiling for now.

packages/core/test/ovm/Finder.js
```diff
@@ -0,0 +1,58 @@
+/* External Imports */
+const { web3 } = require("hardhat");
+const { didContractThrow } = require("@uma/common");
+const { utf8ToHex, hexToUtf8, toChecksumAddress, randomHex, padRight } = web3.utils;
+const Finder = require("../../artifacts-ovm/contracts/ovm/oracle/implementation/Finder.sol/Finder.json");
```
nicholaspai, line 5: Ideally can use `getAbi` here

packages/core/contracts/optimism/external/openzeppelin/contracts/access/Ownable.sol
nicholaspai: Great point, I refactored a bit will tag you

packages/common/package.json
nicholaspai: I think we can move them to devDependencies...

packages/common/src/HardhatConfig.js
```diff
@@ -27,12 +28,28 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
+    ovm: {
+      solcVersion: solcVersion,
+    },
     networks: {
       hardhat: {
         gas: 11500000,
         blockGasLimit: 11500000,
         allowUnlimitedContractSize: false,
         timeout: 1800000,
+        testBlacklist: ["ovm"],
```
nicholaspai, line 40: No need to run `test/ovm` tests for normal `yarn hardhat test` task

packages/common/package.json
nicholaspai: Done

packages/core/contracts/optimism/external/openzeppelin/contracts/access/Ownable.sol
nicholaspai: Done, no new contracts are added in this PR now

packages/core/test/ovm/Finder.js
nicholaspai: Ended up using `hardhat.artifacts.readArtifactsSync` which is the canonical hardhat way to get artifacts. I realize that all the methods exported by `core/index.ts` are designed for truffle tests. 

There is no obvious truffle plugin for Optimism, so when running bots eventually like the `optimistic-oracle-proposer` we'll need to consider whether to run them as `truffle` or `hardhat` scripts.

packages/core/test/ovm/Finder.js
```diff
@@ -0,0 +1,58 @@
+/* External Imports */
+const { web3, artifacts } = require("hardhat");
+const { didContractThrow } = require("@uma/common");
+const { utf8ToHex, hexToUtf8, toChecksumAddress, randomHex, padRight } = web3.utils;
+const Finder = artifacts.readArtifactSync("Finder");
```
chrismaree, line 5: this is new syntax. why did you import like this? is this the workaround for the OVM compiled contracts?

package.json
```diff
@@ -18,7 +18,9 @@
     "qbuild": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*' --ignore '@uma/common' --ignore '@uma/core'",
     "build": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '@uma/common' --ignore '@uma/core'",
     "clean": "yarn lerna run clean --stream",
-    "clean-packages": "yarn lerna clean --yes && rm -rf node_modules"
+    "clean-packages": "yarn lerna clean --yes && rm -rf node_modules",
+    "optimism-up": "./scripts/ovm/run-optimism.sh",
+    "optimism-down": "cd optimism/ops && docker-compose down -v"
```
chrismaree, line 23: why do you have the `-v` in this command? does this not just print the version?

packages/core/test/ovm/Finder.js
nicholaspai: Yes because `getAbi` doesn't know to look in the `artifacts-ovm` directory. More generally, `getAbi` looks in the `build/contracts` dir which is where Truffle places compiled contracts. 

Since there is not Optimism plugin for Truffle, you'll notice that this script is NOT a truffle script (i.e. it does not start with `contract("Finder", function(accounts){...})`, so its run by hardhat. Therefore there is an injected `hardhat runtime environment` which makes it trivial to read artifacts like so.

I actually think if we're going to be running more tests with hardhat, we should move to use syntax like this.

package.json
nicholaspai:
```
    -v, --volumes           Remove named volumes declared in the `volumes`
                            section of the Compose file and anonymous volumes
                            attached to containers.
```

So sanity check to remove any volumes created, but i don't think there are any. I can remove to be clean

packages/core/test/ovm/Finder.js
mrice32: @chrismaree I discussed this with @nicholaspai offline. I'm working on standardizing the syntax for a "hardhat test" (i.e. one that doesn't require the truffle plugin). In that vein, I will be sending out a PR with these practices/plugins written out, so we can align around them and @nicholaspai can use them here.

packages/common/src/HardhatConfig.js
```diff
@@ -27,12 +28,32 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
+        url: "http://127.0.0.1:8545",
+        accounts: { mnemonic: "test test test test test test test test test test test junk" },
+        // This sets the gas price to 0 for all transactions on L2. We do this because account balances are not yet
+        // automatically initiated with an ETH balance.
+        gasPrice: 0,
+        // This sets the network as using the ovm and ensure contract will be compiled against that.
+        ovm: true,
+        // We use custom logic to only compile contracts within the listed directories, as opposed to choosing which
+        // ones to ignore, because there are more contracts to ignore than to include.
+        compileWhitelist: [
```
nicholaspai, line 52: Contracts are limited to those that don't have libraries, which the Optimism-solc compiler doesn't quite handle yet: [issue](https://github.com/ethereum-optimism/solidity/issues/29)

packages/common/src/HardhatConfig.js
```diff
@@ -9,10 +9,11 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
   require("@nomiclabs/hardhat-web3");
   require("hardhat-deploy");
   require("@nomiclabs/hardhat-etherscan");
+  require("@eth-optimism/hardhat-ovm");
   require("./gckms/KeyInjectorPlugin");
 
   // Custom tasks to interact conveniently with smart contracts.
-  require("./hardhat/tasks");
+  require("./hardhat");
```
mrice32, line 16: Is this comment still correct (does this import more than just tasks)?

packages/common/src/HardhatConfig.js
```diff
@@ -27,12 +28,32 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
+    },
     networks: {
       hardhat: {
         gas: 11500000,
         blockGasLimit: 11500000,
         allowUnlimitedContractSize: false,
         timeout: 1800000,
+        testBlacklist: ["ovm"],
+      },
+      optimism: {
```
mrice32, line 42: Should this somehow specify that this is optimism running locally? This isn't the public optimism network, right?

packages/common/src/hardhat/tasks/compile.js
mrice32: Gotcha, makes sense.

packages/common/src/hardhat/index.js
```diff
@@ -0,0 +1,6 @@
+const fs = require("fs");
+const path = require("path");
+
+["tasks"].forEach((folder) =>
```
mrice32, line 4: I think this syntax is great for this use case, since hardhat will not be exported to frontends importing this package.

We should be aware of where we use it, thought, as it isn't compatible with frontends.

packages/common/src/hardhat/tasks/test.js
```diff
@@ -0,0 +1,35 @@
+const path = require("path");
+const { internalTask } = require("hardhat/config");
+const { TASK_TEST_GET_TEST_FILES } = require("hardhat/builtin-tasks/task-names");
+
+const TESTS_DIR = path.resolve(__dirname, "../../../../core/test");
+
+// This overrides a hardhat internal task, which is part of its test task's lifecycle.
```
mrice32, line 7: So the idea is to blacklist certain test files for the optimism test network, specifically?

packages/core/test/ovm/Finder.js
mrice32: See https://github.com/UMAprotocol/protocol/pull/3180.

packages/core/test/ovm/Finder.js
```diff
@@ -0,0 +1,58 @@
+/* External Imports */
```
mrice32, line 1: Why is there a separate folder for tests for optimism? I thought we were going to use the same tests for common contracts and just blacklist certain directories that weren't OVM compatible?

mrice32 (commented):
Once we get #3180 in and can move the test over to using that syntax, I think we should be good to go!

packages/core/test/ovm/Finder.js
chrismaree: I'm guessing the point of this was to prove we can run _some_ tests in this context. Once #3180 has been merged hopefully we can move to the blacklist syntax.

packages/common/src/HardhatConfig.js
nicholaspai: yeah its just local, for now I'm going to leave it as is because there aren't many dependencies. When we add in the public networks i'll make the change. WDYT?

packages/common/src/hardhat/index.js
nicholaspai: Ah, because `fs` and `path` don't exist on frontends?

packages/common/src/hardhat/tasks/test.js
nicholaspai: Other way around: blacklist `optimism/` tests for other networks. But now that I think about it, a test `whitelist` might make more sense so we can cherry pick which tests to run in the Optimism environment.

nicholaspai: Or, using the `blacklist` to block non-OVM compatible tests. I'll go with this solution as you suggested.

packages/core/test/oracle/Finder.js
```diff
@@ -12,15 +12,19 @@ describe("Finder", function () {
...
 
+  let runningOptimism;
   before(async function () {
     accounts = await web3.eth.getAccounts();
     [owner, user] = accounts;
-    await runDefaultFixture(hre);
+
+    // If connected to Optimism network, then we can't compile some of the other contracts required in the default
+    // test fixture, so we'll skip them for now.
+    runningOptimism = (await web3.eth.net.getId()) === 420;
```
nicholaspai, line 22: @chrismaree to demonstrate this test against the `optimism` network, we need to skip the default test fixture for this special case. We can remove this once solidity 0.8 is fully supported

packages/common/src/HardhatConfig.js
```diff
@@ -50,6 +52,19 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
...
+        url: "http://127.0.0.1:8545",
+        accounts: { mnemonic: "test test test test test test test test test test test junk" },
+        // This sets the gas price to 0 for all transactions on L2. We do this because account balances are not yet
+        // automatically initiated with an ETH balance.
+        gasPrice: 0,
+        // This sets the network as using the ovm and ensure contract will be compiled against that.
+        ovm: true,
+        // We use custom logic to only compile contracts within the listed directories, as opposed to choosing which
+        // ones to ignore, because there are more contracts to ignore than to include.
+        compileWhitelist: ["oracle/implementation/Finder.sol"],
```
chrismaree, line 65: looks good. for now we just make _one_ working compilation and can expand later.

packages/core/test/oracle/Finder.js
chrismaree: does that not mean this fixture is not required within this test, in general?

chrismaree (approved):
looks good! like the minimal setup for now that we can update later on.

packages/core/test/oracle/Finder.js
nicholaspai: Good point. I'll remove it

PR #3161, opened 2021-06-29
fix(api): use api.domination.finance endpoint (#3160)

### LINKED ISSUE #3160: Use `api.domination.finance` endpoint for historical price resolutions
**Why**

Currently the `DominationFinancePriceFeed` used two different endpoints, one being the `api.domination.finance` for the current price and `live.domination.finance` for the historical price. Instead only the (more stable and unprocessed data) `api.domination.finance` end point should be used.

**How**

The URL will simply be changed in the `DominationFinancePriceFeed` class. The JSON format returned is identical for the existing `live.domination.finance` API.

**Resolution**

The new endpoint is being used instead.

**Difficulty Score [1-10]**

1: <=5 minutes

jaxrtech:
Signed-off-by: Josh Bowden <josbow@gmail.com>

**Motivation**

Uses `api.domination.finance` endpoint exclusively for all API requests. See #3160.

**Summary**

The API URL was changed from `live.domination.finance` to `api.domination.finance` for the historical price feed.

The `api` endpoint does not do any processing of the pricing data and simply returns the raw historical data from CoinGecko. 
The `live` endpoint is primarily used to power our website's chart and has some data preprocessing for grouping by ticks, etc. 
Importantly, at the moment the `api` endpoint is significantly more stable than the `live` endpoint, primarily due to the fact that the `api` endpoint is not stateful.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #3160.

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/DominationFinancePriceFeed.js
@@ -58,7 +58,7 @@ class DominationFinancePriceFeed extends PriceFeedInterface {
   }
 
   get _historicalPricesUrl() {
-    return `https://live.domination.finance/api/v0/history/${this.pair}?tick=${this.tickPeriod}s&range=${this.lookback}s`;
+    return `https://api.domination.finance/api/v0/price/${this.pair}/history?tick=${this.tickPeriod}s&range=${this.lookback}s`;
   }
 
   getCurrentPrice() {


jaxrtech:
@smb2796, my apologies. I realize this should have been submitted quite a while ago, but a simple change. I wanted to make sure that this gets switched over before the DVM price request tomorrow at expiration. This will prevent the same issue/confusion from the previous expiring synths. Thanks.

nicholaspai (approved):

smb2796:
> @smb2796, my apologies. I realize this should have been submitted quite a while ago, but is a simple change. The API formats returned are the same, so only the URL needed to change.
> 
> I wanted to make sure that this gets switched over before the DVM price request tomorrow at expiration. This will prevent the same issues/confusion from the previous expiring synths we had before. Thanks.

Thanks Josh. It looks to me like the range parameter doesn't actually filter for a period of time in the api url. Is this expected?

smb2796 (approved):
Thanks for fixing @jaxrtech. This looks good to me

PR #3162, opened 2021-06-29
feat(price-id-utils): Ignore uTVL_KPI_UMA for OO proposals

smb2796:
**Motivation**

uTVL_KPI_UMA does not have a price feed and thus needs to be ignored by OO proposal bots.

**Summary**

Add to OPTIMISTIC_ORACLE_IGNORE_POST_EXPIRY.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

**Issue(s)**
NA


packages/common/src/PriceIdentifierUtils.js
```diff
@@ -36,6 +36,7 @@ const OPTIMISTIC_ORACLE_IGNORE_POST_EXPIRY = [
   "uSTONKS_APR21",
   "GASETH-TWAP-1Mx1M",
   "uSTONKS_JUN21",
+  "uTVL_KPI_UMA",
```
nicholaspai, line 39: So, this should be ignored post-expiry, but uses some other pricefeed pre-expiry, like a dex?

smb2796: Nope, should be ignored at all times. Switched lists

nicholaspai (approved):

PR #3167, opened 2021-06-29
fix(LSP-creator): update create params to add more deployment options

chrismaree:
**Motivation**

LSP creators should be able to define the long and short token names without the creator dictating the pattern used. this PR updates this accordingly.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


mrice32 (approved):
So nice that we only have to touch the creator!!

PR #3171, opened 2021-06-30
improve(api): 1533 refactor emp stats to have a table per stat type

daywiss:
**Motivation**

https://app.clubhouse.io/uma-project/story/1533/refactor-emp-stats-to-break-apart-tvl-tvm-and-other-stats-from-the-same-blob

**Summary**

Refactors the emp stats table to take a single value per table. Updates public api calls to reflect this change, stats no longer return a bundle of stats which include tvl and tvm, but rather you must explicitly call for tvl or tvm directly.  

**Details**

This mainly changes the emp-stats service, the emp-stats-history table structure and app state.  This change essentially changes the 1 historical stat table per emp, and now creates a single table per statistic. TVL and TVM get their own tables, and host multiple emp time series data. Queries on these tables are updated to reflect this.  This allows us now to have separate timestamps for each stat.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1533/refactor-emp-stats-to-break-apart-tvl-tvm-and-other-stats-from-the-same-blob

### DIFF
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -50,9 +50,13 @@ export type AppState = {
   erc20s: uma.tables.erc20s.JsMap;
   stats: {
     usd: {
-      latest: empStats.JsMap;
+      latest: {
+        tvl: empStats.JsMap;
+        tvm: empStats.JsMap;
+      };
       history: {
-        [address: string]: empStatsHistory.SortedJsMap;
+        tvl: empStatsHistory.SortedJsMap;
+        tvm: empStatsHistory.SortedJsMap;
       };
     };
   };
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -99,8 +99,12 @@ export default (appState: Dependencies) => {
     const tvl = await bluebird.reduce(
       addresses,
       async (sum, address) => {
-        const stats = await appState.stats[currency].latest.getOrCreate(address);
-        return sum.add(stats.tvl || "0");
+        try {
+          const stats = await appState.stats[currency].latest.tvl.get(address);
+          return sum.add(stats.value || "0");
+        } catch (err) {
+          return sum;
+        }
       },
       BigNumber.from("0")
     );
@@ -110,6 +114,25 @@ export default (appState: Dependencies) => {
     const addresses = Array.from(appState.registeredEmps.values());
     return sumTvl(addresses, currency);
   }
+  async function sumTvm(addresses: string[], currency: CurrencySymbol = "usd") {
+    const tvm = await bluebird.reduce(
+      addresses,
+      async (sum, address) => {
+        try {
+          const stats = await appState.stats[currency].latest.tvm.get(address);
+          return sum.add(stats.value || "0");
+        } catch (err) {
+          return sum;
+        }
+      },
+      BigNumber.from("0")
+    );
+    return tvm.toString();
+  }
+  async function totalTvm(currency: CurrencySymbol = "usd") {
+    const addresses = Array.from(appState.registeredEmps.values());
+    return sumTvm(addresses, currency);
+  }
 
   return {
     getFullEmpState,
@@ -118,6 +141,8 @@ export default (appState: Dependencies) => {
     listExpiredEmps,
     totalTvl,
     sumTvl,
+    totalTvm,
+    sumTvm,
     latestPriceByTokenAddress,
     historicalPricesByTokenAddress,
     sliceHistoricalPricesByTokenAddress,
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -96,18 +96,21 @@ export default (appState: Dependencies) => {
   }
 
   async function sumTvl(addresses: string[], currency: CurrencySymbol = "usd") {
-    const tvl = await bluebird.reduce(
-      addresses,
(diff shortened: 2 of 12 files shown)


packages/api/src/services/actions.ts
```diff
@@ -71,51 +72,58 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
...
-      assert(stats[currency].history[empAddress], "Invalid emp address: " + empAddress);
       assert(length < 1000, "length must be less than 1000 samples");
-      return stats[currency].history[empAddress].slice(start, length);
+      return stats[currency].history.tvm.sliceByAddress(empAddress, start, length);
+    },
+    async listTvls(currency: CurrencySymbol = "usd") {
+      return appState.stats[currency].latest.tvl.values();
+    },
+    async listTvms(currency: CurrencySymbol = "usd") {
+      return appState.stats[currency].latest.tvm.values();
```
daywiss, line 126: there are now separate but nearly identical calls to query tvl/tvm stats both for historical and latest values

packages/api/src/services/emp-stats.ts
```diff
@@ -1,132 +1,155 @@
 import assert from "assert";
 import * as uma from "@uma/sdk";
+export { BigNumber, utils } from "ethers";
```
daywiss, line 3: this file is responsible for calculating tvl/tvm from prices and emp states. Its seen a large refactor to break up the monolithic stats object, which contained both tvl/tvm together, into separate tables for these values. this is because a single timestamp was not sufficient to describe when tvl/tvm are updated.

packages/api/src/index.ts
```diff
@@ -50,9 +50,13 @@ export type AppState = {
...
     usd: {
-      latest: empStats.JsMap;
+      latest: {
+        tvl: empStats.JsMap;
+        tvm: empStats.JsMap;
+      };
       history: {
-        [address: string]: empStatsHistory.SortedJsMap;
+        tvl: empStatsHistory.SortedJsMap;
+        tvm: empStatsHistory.SortedJsMap;
```
daywiss, line 59: tvl and tvm are given their own tables now. these tables can store data for multiple emps.

packages/api/src/services/actions.ts
```diff
@@ -71,51 +72,58 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       assert(exists(emp.tokenCurrency), "EMP does not have token currency address");
       return queries.historicalPricesByTokenAddress(emp.tokenCurrency, start, end);
     },
+    async historicalCollateralPrices(empAddress: string, start = 0, end: number = Date.now()): Promise<PriceSample[]> {
+      assert(empAddress, "requires an empAddress");
+      const emp = await queries.getAnyEmp(empAddress);
+      assert(exists(emp.collateralCurrency), "EMP does not have token currency address");
+      return queries.historicalPricesByTokenAddress(emp.collateralCurrency, start, end);
+    },
```
daywiss, line 80: this function was moved and not changed, just so i could line it up with the readme.

packages/api/src/libs/queries.ts
```diff
@@ -102,6 +106,25 @@ export default (appState: Dependencies) => {
     const addresses = Array.from(appState.registeredEmps.values());
     return sumTvl(addresses, currency);
   }
+  async function sumTvm(addresses: string[], currency: CurrencySymbol = "usd") {
+    const tvm = await bluebird.reduce(
```
chrismaree, line 110: does the `bluebird.reduce` send these calls in parallel? else this could be pretty slow for 100+ calls, no?

chrismaree (approved):

packages/api/src/libs/queries.ts
daywiss: the bottleneck here would be pulling the data from the database. right now its in memory so its probably fine, but in the case its from a remote db, then there are some tricks we could do to speed it up, but ultimately we are limited by long it takes to get the last tvm value.

mrice32: Agreed. But does the reduce issue the requests all at once? Or does it wait on `n-1` before processing `n`? If that's the case, I think we should use something more like a Promise.all wrapping a map and then reduce the result.

packages/api/src/services/emp-stats.ts
```diff
@@ -1,132 +1,155 @@
...
+    const update = {
+      value,
+      timestamp: priceSample[0],
+    };
+    return stats[currency].latest.tvm.upsert(emp.address, update);
   }
-
   // update all stats based on array of emp addresses
-  async function updateAllStats(addresses: string[]) {
+  async function updateAllTvl(addresses: string[]) {
```
mrice32, line 97: minor side note: there seems to be a decent amount of repeated code in this class. Not necessarily a problem. Just wanted to note it.

mrice32 (approved):
LGTM!

packages/api/src/services/emp-stats.ts
daywiss: yea it eventually could be factored to share a single function under the hood, maybe. prefer to leave it due to time crunch

packages/api/src/libs/queries.ts
daywiss: it does not request all at once. i will refactor based on promise.all approach

PR #3180, opened 2021-07-01
feat(common): add hardhat plugin to enhance web3 ease of use

mrice32:
**Motivation**

Rather than requiring tests to manually pull in hardhat deployments and create web3 instantiations, we should have a plugin that provides similar artifact functionality to what truffle provides. Namely an equivalent to `artifacts.require`, `.new()`, `.at()`, `.deployed()`, and other artifact information, like `abi` or `bytecode`.

**Summary**

This creates a hardhat plugin that adds:
- `getContract` (`artifacts.require`).
- The return value from `getContract` has `.new()`, `.at()`, `.deployed()`, and other artifact information that comes from the hardhat artifacts, like `abi`, `bytecode`, etc.

All of this is truffle free! It all points to hardhat artifacts and deployments.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/common/src/hardhat/plugins/ExtendedWeb3.js
```diff
@@ -0,0 +1,84 @@
+// Pluging notes:
```
nicholaspai, line 1:
```suggestion
// Plugin notes:
```

packages/common/src/hardhat/plugins/ExtendedWeb3.js
```diff
@@ -0,0 +1,84 @@
...
+// The plugin adds 4 methods to the hre: getContract, findEvent, assertEventEmitted, assertEventNotEmitted.
+// It requires the web3 plugin to work. All contract instantiations are web3, not truffle.
```
nicholaspai, line 3:
```suggestion
// It requires the web3 hardhat plugin to work. All contract instantiations are web3, not truffle.
```

packages/common/src/hardhat/plugins/ExtendedWeb3.js
```diff
@@ -0,0 +1,84 @@
...
+const { extendEnvironment } = require("hardhat/config");
+
+extendEnvironment((hre) => {
+  hre._artifactCache = {};
+  hre.getContract = (name) => {
+    if (!hre._artifactCache[name]) hre._artifactCache[name] = hre.artifacts.readArtifactSync(name);
+    const artifact = hre._artifactCache[name];
+
+    const deployed = async () => {
+      const deployment = await hre.deployments.get(name);
```
nicholaspai, line 39: This should read from both `networks/X.json` and the `deployments` folder right?

mrice32: It should do whatever hardhat deploy is configured to do!

packages/common/src/hardhat/plugins/ExtendedWeb3.js
```diff
@@ -0,0 +1,84 @@
...
+
+    return {
+      ...artifact,
+      deployed,
+      new: newProp,
+      at,
+    };
+  };
+
+  hre.findEvent = async (txnResult, contract, eventName, fn = () => true) => {
```
nicholaspai, line 56: What is the expected shape of this `txnResult`? Since we're not building Truffle objects, I'm guessing its not like [this TransactionResult defined in truffle-assertions](https://github.com/rkalis/truffle-assertions#truffleassertcreatetransactionresultcontract-transactionhash)?

nicholaspai (commented):
This looks great, really smart to add to this to the `hre`! Can you include one usage of these new tasks in a unit test file as a sanity check?

packages/common/src/hardhat/plugins/ExtendedWeb3.js
mrice32: It is not. It is the web3 transaction result.

```
const txnResult = someContract.methods.someMethod(arg1, arg2, ...).send({ from: accounts[0] });
const event = await findEvent(txnResult, someContract, "MyEvent", someFunction);
```

mrice32:
> This looks great, really smart to add to this to the `hre`! Can you include one usage of these new tasks in a unit test file as a sanity check?

I've been moving a lot of tests over from our existing test suite in a separate branch. I can drop one of those test files in here as an example.

nicholaspai (approved):
Wow this is so easy to integrate into tests LGTM

PR #3182, opened 2021-07-01
feat(sdk): 1574 add multicall client and helper

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1574/add-multicall-sdk-client

**Summary**

This adds a basic multi-call client which can be eventually used in the api.  Currently only supports read requests, not write (transactions).

**Details**

This adds a contract client and a small helper function which will help encode/decode requests responses to the multicall contract.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1574/add-multicall-sdk-client

### DIFF
--- a/packages/sdk/src/libs/clients/multicall/client.ts
+++ b/packages/sdk/src/libs/clients/multicall/client.ts
@@ -1,0 +1,8 @@
+import { Multicall__factory, Multicall } from "@uma/core/contract-types/ethers";
+import type { SignerOrProvider } from "../..";
+
+export type Instance = Multicall;
+
+export function connect(address: string, provider: SignerOrProvider): Instance {
+  return Multicall__factory.connect(address, provider);
+}
--- a/packages/sdk/src/libs/multicall.ts
+++ b/packages/sdk/src/libs/multicall.ts
@@ -1,0 +1,56 @@
+import { multicall } from "./clients";
+import { Contract } from "ethers";
+import type { SignerOrProvider } from "..";
+import { zip } from "lodash";
+
+export type Call = [string, any[]] | [string];
+
+export type Request = {
+  contractInstance: Contract;
+  call: Call;
+};
+export type EncodedResponse = string;
+export type EncodedRequest = {
+  target: string;
+  callData: string;
+};
+export default class Multicall {
+  private multicallClient: multicall.Instance;
+  private requests: Request[];
+
+  constructor(address: string, provider: SignerOrProvider) {
+    this.multicallClient = multicall.connect(address, provider);
+    this.requests = [];
+  }
+
+  public add(contractInstance: Contract, ...call: Call) {
+    this.requests.push({ contractInstance, call });
+  }
+  public batch(contractInstance: Contract, calls: Call[]) {
+    calls.forEach((call: Call) => {
+      this.add(contractInstance, ...call);
+    });
+  }
+  private encodeRequest(request: Request) {
+    const { contractInstance, call } = request;
+    return {
+      target: contractInstance.address,
+      callData: contractInstance.interface.encodeFunctionData(call[0], call[1]),
+    };
+  }
+
+  private decodeResponse(request: Request, response: EncodedResponse) {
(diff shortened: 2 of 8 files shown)


packages/sdk/src/libs/clients/multicall/client.ts
```diff
@@ -0,0 +1,8 @@
+import { Multicall__factory, Multicall } from "@uma/core/contract-types/ethers";
+import type { SignerOrProvider } from "../..";
+
+export type Instance = Multicall;
+
+export function connect(address: string, provider: SignerOrProvider): Instance {
+  return Multicall__factory.connect(address, provider);
+}
```
mrice32, line 8: Any reason not just to write this in-line with your multicall class? Seems like the extra file isn't necessary since it's just wrapping the `@uma/core` factory import, and it would be a one-liner to call connect elsewhere.

mrice32 (commented):
Looks great! A few minor comments. The long comment I left on the mulicall class structure is just a note -- no changes are requested/required!

chrismaree (approved):

packages/sdk/src/libs/clients/multicall/client.ts
daywiss: yea you are right, it could live there. I thought maybe exposing it here in the same pattern set by previous clients, it could be used in the raw form if someone wanted to. All other clients in the sdk expose a static connect function which basically just does this lookup to the typechain implementation.

packages/sdk/src/libs/multicall.ts
```diff
@@ -0,0 +1,56 @@
...
+      callData: contractInstance.interface.encodeFunctionData(call[0], call[1]),
+    };
+  }
+
+  private decodeResponse(request: Request, response: EncodedResponse) {
+    const { contractInstance, call } = request;
+    return contractInstance.interface.decodeFunctionResult(call[0], response);
+  }
+
+  public async read(_requests: Request[] = this.requests) {
```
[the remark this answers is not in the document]
daywiss: great observation, this is something i was really struggling with during the design of this class. i dont like how the state sticks around, but figured if you needed to clear it you could make a new instance.  I think the chaining api you illustrate could work really well here, i didnt think of that.  I'm going to try it!

packages/sdk/src/libs/multicall.ts
```diff
@@ -0,0 +1,102 @@
...
+    return zipped.map(([request, response]) => {
+      if (request && response) return this.decodeResponse(request, response);
+      throw new Error("Unable to decode contract response");
+    });
+  }
+}
+
+// Multicall class that exposes public functions to the user and recursively chains itself.  Acts immutable
+// if you store reference to the parent intsance. Children will contain mutated state.
+export class Multicall {
```
mrice32, line 73: This implementation is very cool!

The big question I have is why we need the Core class since it seems to be transient? It seems like we could just dump all of that functionality into this class, right?

One suspicion I have for why you might've done it this way is because you want to expose information to the other Multicall instance that you don't want to expose to the user. Note: you can use private in a fairly clever way to achieve this without the other class.

Let's say you have two instances of Multicall. Each one has a private method `foo()` and private data `bar`. The following setup is totally allowed because different instances of the same class can access each others' private information:
```ts
export class Multicall {
  private number: bar;
  
  private foo() {
    console.log("called a private method!");
  }

  public passOtherMultiCall(otherMulticall: Multicall) {
    otherMultiCall.foo();
    console.log("some private data", othermulticall.bar);
  }
}
```

See https://www.typescriptlang.org/docs/handbook/2/classes.html#cross-instance-private-access for more info.

I think this means you could wrap all the mutation functionality into a private method or two within this class and call it from the other instance when needed. Thoughts?

mrice32 (commented):
Looks awesome! One minor comment on the multicall class structure!

mrice32 (approved):
LGTM!

mrice32 (approved):
This is really cool! +1

PR #3184, opened 2021-07-02
feat(LSP): Add LSP pair name, OO liveness and OO bond settings

chrismaree:
**Motivation**

The LSP contract needs three additional pieces of functionality:
- The notion of a `pairName` to name an LPS contract. This is useful in front ends to display information about the contract.
- The ability for the deployer to set a custom OO liveness via a `optimisticOracleLivenessTime` param. This is important with complex price requests that might need more than the normal 2 hours to facilitate price proposal verification.
- The ability for the deployer to set a custom OO bond via a `optimisticOracleProposerBond` param. This lets the deployer create stronger security guarantees around price proposals as there is now a cost associated with an invalid proposal.

**Summary**

Updated contracts and unit tests to add a) `pairName`, b) `optimisticOracleLivenessTime` and c) `optimisticOracleProposerBond`.

**Details**

The addition of these parameters put the contract constructor stack depth above Solidity limit. To work around this, the constructor and factory methods were refactored to take in a struct. This is the same pattern the EMP and Perp uses to work around these limits.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
+++ b/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
@@ -34,8 +34,26 @@ contract LongShortPair is Testable, Lockable {
      *  LONG SHORT PAIR DATA STRUCTURES  *
      *********************************************/
 
+    // Define the contract's constructor parameters as a struct to enable more variables to be specified.
+    struct ConstructorParams {
+        string pairName; // Name of the long short pair contract.
+        uint64 expirationTimestamp; // Unix timestamp of when the contract will expire.
+        uint256 collateralPerPair; // How many units of collateral are required to mint one pair of synthetic tokens.
+        bytes32 priceIdentifier; // Price identifier, registered in the DVM for the long short pair.
+        ExpandedIERC20 longToken; // Token used as long in the LSP. Mint and burn rights needed by this contract.
+        ExpandedIERC20 shortToken; // Token used as short in the LSP. Mint and burn rights needed by this contract.
+        IERC20 collateralToken;
+        LongShortPairFinancialProductLibrary financialProductLibrary;
+        bytes customAncillaryData;
+        uint256 prepaidProposerReward;
+        uint256 optimisticOracleLivenessTime;
+        uint256 optimisticOracleProposerBond;
+    }
+
     enum ContractState { Open, ExpiredPriceRequested, ExpiredPriceReceived }
     ContractState public contractState;
+
+    string public pairName;
 
     uint64 public expirationTimestamp;
 
@@ -59,9 +77,11 @@ contract LongShortPair is Testable, Lockable {
 
     LongShortPairFinancialProductLibrary public financialProductLibrary;
 
+    // Optimistic oracle customization parameters.
     bytes public customAncillaryData;
-
     uint256 public prepaidProposerReward;
+    uint256 public optimisticOracleLivenessTime;
+    uint256 public optimisticOracleProposerBond;
 
     /****************************************
      *                EVENTS                *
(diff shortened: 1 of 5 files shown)


packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -93,57 +113,45 @@ contract LongShortPair is Testable, Lockable {
...
-        require(_getIdentifierWhitelist().isIdentifierSupported(_priceIdentifier), "Identifier not registered");
+        require(params.expirationTimestamp > getCurrentTime(), "Expiration timestamp in past");
+        require(params.collateralPerPair > 0, "Collateral per pair cannot be 0");
+        require(_getIdentifierWhitelist().isIdentifierSupported(params.priceIdentifier), "Identifier not registered");
         require(address(_getOptimisticOracle()) != address(0), "Invalid finder");
-        require(address(_financialProductLibrary) != address(0), "Invalid FinancialProductLibrary");
-        require(_getCollateralWhitelist().isOnWhitelist(address(_collateralToken)), "Collateral not whitelisted");
+        require(address(params.financialProductLibrary) != address(0), "Invalid FinancialProductLibrary");
+        require(_getCollateralWhitelist().isOnWhitelist(address(params.collateralToken)), "Collateral not whitelisted");
+        require(params.optimisticOracleLivenessTime > 0, "OO liveness cannot be 0");
```
nicholaspai, line 132: liveness also has to be less than 5200 weeks according to the OptimisticOracle

nicholaspai (approved):
LGTM, nice and simple fix. One nit comment!

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: good catch! fixed

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -93,57 +113,46 @@ contract LongShortPair is Testable, Lockable {
...
      * @param _timerAddress Contract that stores the current time in a testing environment. Set to 0x0 in production.
      */
     constructor(
-        uint64 _expirationTimestamp,
-        uint256 _collateralPerPair,
-        bytes32 _priceIdentifier,
-        ExpandedIERC20 _longToken,
-        ExpandedIERC20 _shortToken,
-        IERC20 _collateralToken,
+        ConstructorParams memory params,
```
mrice32, line 121: nit: why exclude finder and timer?

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -59,9 +77,11 @@ contract LongShortPair is Testable, Lockable {
 
     LongShortPairFinancialProductLibrary public financialProductLibrary;
 
+    // Optimistic oracle customization parameters.
     bytes public customAncillaryData;
-
     uint256 public prepaidProposerReward;
+    uint256 public optimisticOracleLivenessTime;
```
mrice32, line 83: nit: not sure how much stack space we need to save, but if we only need to save like one variable, we could create a struct called OptimisticOracleParams that just contains the reward, liveness, and bond. That might get around the stack limit without creating so many changes. We don't have to do this, as it may be less clean, but it could reduce the lines changed so this is easier to get audited.

packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol
```diff
@@ -51,64 +68,46 @@ contract LongShortPairCreator is Testable, Lockable {
...
-     * @param collateralPerPair how many units of collateral are required to mint one pair of synthetic tokens.
-     * @param priceIdentifier registered in the DVM for the synthetic.
-     * @param longSynthName Name of the long synthetic tokens to be created.
-     * @param longSynthSymbol Symbol of the long synthetic tokens to be created.
-     * @param shortSynthName Name of the short synthetic tokens to be created.
-     * @param shortSynthSymbol Symbol of the short synthetic tokens to be created.
-     * @param collateralToken ERC20 token used as collateral in the LSP.
-     * @param financialProductLibrary Contract providing settlement payout logic.
-     * @param customAncillaryData Custom ancillary data to be passed along with the price request. If not needed, this
+     * @param params Constructor params used to initialize the LSP. Key-valued object with the following structure:
```
mrice32, line 71: nit: in the other struct, you put the comments on the struct itself rather than on the method. In this case, you put the comments on the method? Should we make these consistent?

mrice32 (approved):
LGTM!

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: I though it made sence to put the LSP specific contracts in the struct and leave the finder and timer out as the latter two are “supporting” information, where the others are parameterization

packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol
chrismaree: I added them to this method as this is method like this as this is the primary interface people will be calling whereas the constructor for the LSP is more of an internal function as we don't want people deploying LPSs directly ideally.

I'll add the comment there as well for consistency :)

mrice32 (approved):
LGTM!

PR #3201, opened 2021-07-07
fix(networks file): fix broken LSP creator address

chrismaree:
**Motivation**

While helping devx with some deployment info, I realized there was a mistake in the mainnet networks file. the LSP creator address does not point to the correct contract and rather points to a `Finder` contract which must have been deployed by mistake when creating the LSP deployment.

**Summary**

Fixes a mistake in the networks file.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/core/networks/1.json
+++ b/packages/core/networks/1.json
@@ -93,7 +93,7 @@
   },
   {
     "contractName": "LongShortPairCreator",
-    "address": "0x5a8d408B7908E4F00A2f5460B621dbF2DDce8042"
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

EdsonAlcala:
Don't you also need to create a new package version? Because the current one (2.4.0) doesn't have any networks in the artifacts.

nicholaspai:
> Don't you also need to create a new package version? Because the current one (2.4.0) doesn't have any networks in the artifacts.

Yup #3200 will release it after this PR is merged

PR #3202, opened 2021-07-07
feat(core): refactor core tests to no longer use truffle

mrice32:
**Motivation**

The ultimate goal is to move off of truffle entirely. To get there, the largest barrier is the large amount of test code relying on truffle abstractions.

**Summary**

This PR removes all (with one very small exception) truffle abstractions from the core tests.

Note: this is a very large PR, but since the changes are essentially the same process applied throughout the tests, I figured it might be just as convenient to review as one large change rather than broken up into many small ones.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/test/chainbridge/beaconOracle.js
```diff
@@ -22,8 +23,13 @@ contract("BeaconOracle", async (accounts) => {
   const testRequestTime = 123;
   const testPrice = "6";
 
-  beforeEach(async function () {
+  before(async () => {
+    [owner] = accounts;
```
nicholaspai, line 27: This order seems off, shouldn't `accounts = await web3.eth.getAccounts()` precede `[owner] = accounts`?

packages/core/test/truffle-fixture.js
```diff
@@ -1,37 +0,0 @@
-const deployFinder = require("../migrations/2_deploy_finder.js");
```
nicholaspai, line 1: +100

nicholaspai (approved):
Wow thanks for doing the ugly work, i pointed out any typos.

packages/core/test/chainbridge/beaconOracle.js
mrice32: Yep! This is broken. Fixing this and will push soon.

chrismaree (approved):
good catch!

packages/api/package.json
```diff
@@ -35,7 +35,7 @@
   "scripts": {
     "build": "tsc",
     "build:watch": "tsc --watch",
-    "test": "ts-mocha --recursive=true ./src/libs/**/*.test.ts ./src/tables/**/*.test.ts",
```
chrismaree, line 38: curious that this got through the linter in CI

packages/common/src/HardhatConfig.js
```diff
@@ -1,10 +1,10 @@
 const { getNodeUrl, mnemonic } = require("./TruffleConfig");
 const path = require("path");
 
-function getHardhatConfig(configOverrides, workingDir = "./") {
+function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = true) {
   // Hard hat plugins. These are imported inside `getHardhatConfig` so that other packages importing this function
```
chrismaree, line 5:
```suggestion
  // Hardhat plugins. These are imported inside `getHardhatConfig` so that other packages importing this function
```

packages/core/scripts/EmergencyShutdown.js
```diff
@@ -25,7 +23,9 @@ async function run(account, derivative) {
 
 async function runScript(callback) {
   const account = (await web3.eth.getAccounts())[0];
-  await run(account, argv.derivative);
+  const adminAbi = artifacts.require("FinancialContractsAdmin").abi;
```
chrismaree, line 26: why pass these in like this rather than just keeping them as consts at the top of the file? just trying to understand the general pattern for future files/changes.

packages/core/test/chainbridge/e2e.js
```diff
@@ -88,7 +88,10 @@ contract("GenericHandler - [UMA Cross-chain Communication]", async (accounts) =>
   let votingResourceId;
   let governanceResourceId;
 
-  beforeEach(async () => {
+  before(async () => {
+    accounts = await web3.eth.getAccounts();
```
chrismaree, line 92: this change in every file is pretty painful!

packages/core/test/financial-templates/common/financial-product-libraries/expiring-multiparty-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.js
```diff
@@ -1,98 +1,110 @@
...
-  beforeEach(async () => {
+  before(async () => {
+    await runDefaultFixture(hre);
+    accounts = await hre.web3.eth.getAccounts();
     timer = await Timer.deployed();
 
-    expirationTime = (await timer.getCurrentTime()) + 100; // use 100 seconds in the future as the expiration time.
-    identifierTransformationFPL = await PreExpirationIdentifierTransformationFinancialProductLibrary.new();
+    expirationTime = (await timer.methods.getCurrentTime().call()) + 100; // use 100 seconds in the future as the expiration time.
+    identifierTransformationFPL = await PreExpirationIdentifierTransformationFinancialProductLibrary.new().send({
```
chrismaree, line 34: if you _did not_ put the `{ from: accounts[0] }` here would it default to using this account or what would have happened? does it error out in that case?

packages/core/test/financial-templates/common/financial-product-libraries/expiring-multiparty-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.js
```diff
@@ -1,98 +1,110 @@
...
 
     // Calling the transformation function as a mocked emp caller should also work.
     assert.equal(
       hexToUtf8(
-        await expiringMultiParty.transformPriceIdentifier.call((await expiringMultiParty.getCurrentTime()).toString(), {
-          from: expiringMultiParty.address,
-        })
+        await expiringMultiParty.methods
+          .transformPriceIdentifier((await expiringMultiParty.methods.getCurrentTime().call()).toString())
+          .call({ from: expiringMultiParty.options.address })
```
chrismaree, line 84: it's a bit painful that the `address` prop is now within the contract instances `options` section

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.js
```diff
@@ -1,79 +1,108 @@
...
     it("Can not re-use existing LSP contract address", async () => {
-      await binaryLSPFPL.setLongShortPairParameters(expiringContractMock.address, strikePrice);
+      await binaryLSPFPL.methods
+        .setLongShortPairParameters(expiringContractMock.options.address, strikePrice)
+        .send({ from: accounts[0] });
 
       // Second attempt should revert.
       assert(
-        await didContractThrow(binaryLSPFPL.setLongShortPairParameters(expiringContractMock.address, strikePrice))
+        await didContractThrow(
```
chrismaree, line 55: Seeing all these changes in clude the `{ from:accounts[0] }` I'm going to assume this is a strict requirement for this. is there no way to tell the contract instance to default to a particular account? it would make these kinds of sets cleaner. I guess it is more explicitly to define every "from" address every time a transaction is sent.

chrismaree: main downside is this becomes pretty repetitive and the syntax is more verbose on every function call

chrismaree (approved):
This is a HUGE amount of work! The changes seem reasonable. it's primarily:
- Changing how we do imports
- changing how we initialize addresses in tests
- changing how we call methods on contracts
- changing how you fetch the address of a deployed contract.

The two things I think we could improve on ideally are:
1) It would be great if the contract instance would default to `accounts[0]` so you don't need to provide this in the `send` or `call` every time. However, there is an argument that this pattern _is_ better as it is more explicit.
2) it would be much cleaner if we did not need to fetch contract addresses via the `options` prop on the contract instance. this seems pretty unnasasary.


Also, having to call methods via the `.methods` interface is also a bit clinky but there is not really any workaround for that unless we start merging contract props in pretty nasty ways.

I'm curious: did this speed up the test time at all?

chrismaree:
it looks like the build failed due to timeout. will re-run this to see if it helps else will need to investigate why

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.js
mrice32: I think you can do it (because web3 instances have a default account field), but you'd need to modify web3 before the tests start using it (although I'm not 100% sure how you do this) since this doesn't happen by default. We could make a plug-in to do it if we want to have that for future tests.

packages/core/test/financial-templates/common/financial-product-libraries/expiring-multiparty-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.js
mrice32: It does error, yeah.

packages/core/test/financial-templates/common/financial-product-libraries/expiring-multiparty-libraries/PreExpirationIdentifierTransformationFinancialProductLibrary.js
mrice32: Yeah, not sure what we can do about that. I think there's a `._address` private field we could use, but I'm not sure that's recommended.

mrice32:
> it looks like the build failed due to timeout. will re-run this to see if it helps else will need to investigate why

@chrismaree I think I figured out what's going on. The removal of the truffle plugin broke the contract version hashes script that runs on build. I'll get this fixed up.

packages/api/package.json
mrice32: No idea :/

packages/core/scripts/EmergencyShutdown.js
mrice32: This is just to both keep the script truffle compatible (we haven't moved the scripts over yet), but still allow the function to be tested by the hardhat tests. This will likely be rewritten when we move the scripts over.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.js
mrice32: I do think explicitness is nice. In the tests, we are pretty inconsistent about denoting the account. Sometimes we define an owner and specifically call things from the owner. Other times, we rely on the truffle default.

I can try to include a default in web3 in a later PR and we can use whatever pattern we want for new files/tests!

mrice32:
> This is a HUGE amount of work! The changes seem reasonable. it's primarily:
> 
> * Changing how we do imports
> * changing how we initialize addresses in tests
> * changing how we call methods on contracts
> * changing how you fetch the address of a deployed contract.
> 
> The two things I think we could improve on ideally are:
> 
> 1. It would be great if the contract instance would default to `accounts[0]` so you don't need to provide this in the `send` or `call` every time. However, there is an argument that this pattern _is_ better as it is more explicit.

We can do this. I would prefer not to try and modify any code already changed, but I can look into how to do this at the hardhat level.

> 2. it would be much cleaner if we did not need to fetch contract addresses via the `options` prop on the contract instance. this seems pretty unnasasary.

According to the docs, that's the _right_ way to do it. Since we know the name of that prop, we could technically modify the contract abstraction, but I kinda feel like it isn't worth it. The fewer "weird" things in our codebase, the better IMO.

> Also, having to call methods via the `.methods` interface is also a bit clinky but there is not really any workaround for that unless we start merging contract props in pretty nasty ways.

Agreed. It's clunky. I think the only way around this is to either 1) use ethers in future code (I don't think we should ever try to change the existing code again) or 2) construct some contract abstraction ourselves.

> I'm curious: did this speed up the test time at all?

I think so, but I'm not sure. The obvious places where it sped things up was in tests that actually don't need migrations. I was able to drop the runDefaultFixture, which means there are no setup transactions being run or reset.

Note: one good reason to use vanilla libs and not do too much customization in hardhat is that these customizations don't apply in production because we don't run in a hardhat environment. When these aren't consistent, testing is hard to reason about.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/BinaryOptionLongShortPairFinancialProductLibrary.js
nicholaspai: I also favor explicitness its more readable

PR #3203, opened 2021-07-07
feat(api): 0x api client to fetch prices based on token name

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/1620/create-matcha-api-client

**Summary**

This adds a small client class for matcha (0x) trade routing system. We will need this to fetch market prices on token pairs. This works with both symbols and token addresses.

**Details**

This is designed as a class in case we want to add additional api calls to matcha, its easily extendable.  The errors returned from this api are not really standard, so theres a significant amount of code to parse this out into a standard error message.

**Testing**

- [x] New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1620/create-matcha-api-client

### DIFF
--- a/packages/api/src/libs/zrx.e2e.ts
+++ b/packages/api/src/libs/zrx.e2e.ts
@@ -1,0 +1,31 @@
+import assert from "assert";
+import Client from "./zrx";
+
+const baseUrl = "https://api.0x.org/";
+describe("0x client", function () {
+  let client: Client;
+  it("should init", function () {
+    client = new Client(baseUrl);
+    assert.ok(client);
+  });
+  it("should get a price", async function () {
+    const params = {
+      sellToken: "WETH",
+      // show we can use addresses as well
+      buyToken: "0x6b175474e89094c44da98b954eedeac495271d0f", // DAI
+      sellAmount: "1000000000000000000",
+    };
+    const result = await client.price(params);
+    assert.ok(result.price);
+  });
+  it("should have a human readable error", async function () {
+    const params = {
+      sellToken: "WETH",
+      buyToken: "DAI",
+      // missing sellAmount or buyAmount
+    };
+    const errorMessage =
+      "Bad Request(400): Validation Failed: should have required property 'sellAmount', should have required property 'buyAmount', should match exactly one schema in oneOf";
+    assert.equal(await client.price(params).catch((e) => e.message), errorMessage);
+  });
+});
--- a/packages/api/src/libs/zrx.ts
+++ b/packages/api/src/libs/zrx.ts
@@ -1,0 +1,68 @@
+import Axios, { AxiosInstance } from "axios";
+
+type ValidationError = {
+  field: string;
+  code: number;
+  reason: string;
+};
+
+// reverse engineered type from api response
+type Response = {
+  statusText: string;
+  status: string;
+  data: {
+    reason: string;
+    validationErrors: ValidationError[];
+  };
+};
+
+// describes valid axios config, might be a way to get this from axios
+type AxiosConfig = {
+  method: string;
+  url: string;
+  params: { [key: string]: any };
+};
+
+// 0x api client, maintains raw responses when possible, parses errors to be human readable.
+export default class Client {
+  private axios: AxiosInstance;
+  constructor(baseURL: string) {
+    this.axios = Axios.create({ baseURL });
+  }
+  // this api returns schema validation errors which are nice, but not readable in an error message
+  private stringifyValidationErrors(validationErrors: ValidationError[] = []) {
+    if (validationErrors.length == 0) return "";
+    return validationErrors.map(({ reason }) => reason).join(", ");
+  }
+  // errors returned from this api are rich, this function stringifies the complex object so it can be returned as a message
+  private stringifyErrorResponse(response: Response) {
+    let errorMessage = `${response.statusText}(${response.status})`;
+    const reason = response.data?.reason;
+    const validationErrorString = this.stringifyValidationErrors(response.data?.validationErrors);
+    errorMessage += `: ${reason}`;
+    if (validationErrorString.length) {
+      errorMessage += `: ${validationErrorString}`;
+    }
+    return errorMessage;
+  }
+  // general call function which does some common parsing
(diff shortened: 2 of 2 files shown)


packages/api/src/libs/zrx.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+  it("should have a human readable error", async function () {
+    const params = {
+      sellToken: "WETH",
+      buyToken: "DAI",
+      // missing sellAmount or buyAmount
+    };
+    const errorMessage =
+      "Bad Request(400): Validation Failed: should have required property 'sellAmount', should have required property 'buyAmount', should match exactly one schema in oneOf";
+    let plan = 1;
```
nicholaspai, line 30: What does `plan` stand for?

packages/api/src/libs/zrx.e2e.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+  it("should get a price", async function () {
+    const params = {
+      sellToken: "WETH",
+      // show we can use addresses as well
+      buyToken:'0x6b175474e89094c44da98b954eedeac495271d0f', // DAI
+      sellAmount: "1000000000000000000",
+    };
+    const result = await client.price(params);
+    console.log(result)
```
nicholaspai, line 19: leftover `console.log`?

packages/api/src/libs/zrx.e2e.ts
```diff
@@ -0,0 +1,39 @@
+import assert from "assert";
+import Client from "./zrx";
+
+const baseUrl = "https://api.0x.org/";
```
nicholaspai, line 4: How do you feel about having real external dependencies, such as the `api.0x.org` server, in our unit tests that are run in CI? Is there a way to mock this?

nicholaspai (approved):
LGTM, straightforward PR. Only concern is about the external dependency on the 0x server in the unit tests.

packages/api/src/libs/zrx.e2e.ts
daywiss: plan represents how many tests we plan to make. so i count down every time i expect an assert to hit. just to make sure nothing gets skipped over. plan should be 0 at the end of the test.

packages/api/src/libs/zrx.e2e.ts
daywiss: yea this is awkward. i've named these tests "e2e" to denote they should not be run in ci, and `yarn test` will not run this test.  you have to run `yarn e2e` in order for this test to run.  not ideal and open to suggestions here

mrice32: I like the setup you have @daywiss

packages/api/src/libs/zrx.e2e.ts
mrice32: numExpectedErrors might be clearer.

Also, using a .catch on the promise that returns the error object might be easier to test against than counting errors using a try catch.

```js
assert.equal(await foo.bar().catch(e => e.message), "some message");
```

packages/api/src/libs/zrx.ts
```diff
@@ -0,0 +1,72 @@
...
+      const response = await this.axios(config);
+      return response.data;
+    } catch (err) {
+      if (err.response == null) throw err;
+      throw new Error(this.stringifyErrorResponse(err.response));
+    }
+  }
+  // get a price quote: swap/v1/price?sellToken=WETH&buyToken=DAI&sellAmount=1000000000000000000
+  // this call requires one of sellAmount or buyAmount, but not both or neither
+  public price(params: { sellToken: string; buyToken: string; sellAmount?: string; buyAmount?: string }) {
```
mrice32, line 64: nit: why is the public method at the bottom?

packages/api/src/libs/zrx.ts
```diff
@@ -0,0 +1,72 @@
...
+    try {
+      const response = await this.axios(config);
+      return response.data;
+    } catch (err) {
+      if (err.response == null) throw err;
+      throw new Error(this.stringifyErrorResponse(err.response));
+    }
+  }
+  // get a price quote: swap/v1/price?sellToken=WETH&buyToken=DAI&sellAmount=1000000000000000000
+  // this call requires one of sellAmount or buyAmount, but not both or neither
```
mrice32, line 63: Should we check this? Or should we just let the api fail? (Either way seems reasonable IMO)

packages/api/src/libs/zrx.ts
```diff
@@ -0,0 +1,72 @@
...
+export default class Client {
+  private axios: AxiosInstance;
+  constructor(baseURL: string) {
+    this.axios = Axios.create({ baseURL });
+  }
+  // this api returns schema validation errors which are nice, but not readable in an error message
+  private stringifyValidationErrors(validationErrors: ValidationError[] = []) {
+    if (validationErrors.length == 0) return "";
+    return validationErrors
+      .map(({ reason }) => {
```
mrice32, line 36: Nit: you can drop the braces to make this more concise.

mrice32 (approved):
LGTM!

packages/api/src/libs/zrx.ts
daywiss: open to suggestions here, but their api errors are very informative, and will tell you why your request failed and what parameters are wrong. It might be possible to do this in typescript for a single call, but would be difficult to do if you wanted to add more calls without being intimately familiar with their api.

packages/api/src/libs/zrx.ts
daywiss: i dont really have any particular reason, i think i just typically put utility type functions (private) at the top of the file and business logic (public) at the bottom. not really opinionated on this if you have a preference

packages/api/src/libs/zrx.ts
daywiss: also tbh im not sure how to typescript something that can be one or the other, but not both or none

packages/api/src/libs/zrx.e2e.ts
daywiss: nice i like that

PR #3206, opened 2021-07-08
feat(api): 1523 add market prices to emp state and call to list known market prices

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1523/connect-aggregator-api-to-our-api-or-sdk-for-pulling-in-the-best-price-the-token-is-trading-across-dexs

**Summary**

Adds infrastructure to fetch latest market price from 0x api, and keep it relatively up to date. This then gets added to emp state when queried as `tokenMarketPrice`. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1523/connect-aggregator-api-to-our-api-or-sdk-for-pulling-in-the-best-price-the-token-is-trading-across-dexs

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -18,7 +18,7 @@ tradermadeApiKey=
 quandlApiKey=
 defipulseApiKey=
 
-# zrx price feed gets us market prices on tokens
+# 0x price feed gets us market prices on tokens
 zrxBaseUrl=https://api.0x.org
 ```
 
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -17,6 +17,9 @@ cryptowatchApiKey=
 tradermadeApiKey=
 quandlApiKey=
 defipulseApiKey=
+
+# zrx price feed gets us market prices on tokens
+zrxBaseUrl=https://api.0x.org
 ```
 
 ## Starting
@@ -92,6 +95,10 @@ Returns all known active collateral or synthetic addresses
 ### allLatestPrices(currency='usd') => {[address:string]:[timestamp:number,price:string]
 
 Returns all known latest prices for synthetic or collateral addresses in wei.
+
+### allLatestMarketPrices() => {[address:string]:[timestamp:number,price:string]
+
+Returns all latest known market prices (based on matcha/0x routing) for synthetic or collateral addresses in wei.
 
 ### latestPriceByTokenAddress(tokenAddress:string,currency:'usd') => [timestamp:number,price:string]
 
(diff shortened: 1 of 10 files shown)


packages/api/src/apps/api/README.md
```diff
@@ -17,6 +17,9 @@ cryptowatchApiKey=
 tradermadeApiKey=
 quandlApiKey=
 defipulseApiKey=
+
+# zrx price feed gets us market prices on tokens
```
chrismaree, line 21:
```suggestion
# 0x price feed gets us market prices on tokens
```

packages/api/src/libs/queries.ts
```diff
@@ -51,6 +54,13 @@ export default (appState: Dependencies) => {
     // [timestamp, price], returns just price
     return synthPrices.latest[empAddress][1];
   }
+  async function getLatestMarketPrice(address: string, currency: "usdc" = "usdc") {
```
chrismaree, line 57: how do decimals work with this exactly? is it always returned in the decimals of the `currency`?

packages/api/src/services/market-prices.ts
```diff
@@ -0,0 +1,55 @@
...
+type Dependencies = Pick<AppState, "zrx" | "marketPrices" | "collateralAddresses" | "syntheticAddresses">;
+
+// market prices are pulled from the 0x matcha api
+export default function (config: Config, appState: Dependencies) {
+  // these prices will be quoted against usdc by default, but can be specified as address or symbol
+  const { zrx, marketPrices, collateralAddresses, syntheticAddresses } = appState;
+  // this is hardcoded for now since it differs from the standard currency symbol usd
+  const currency = "usdc";
+
+  // does not do any queries, just a helper to mutate the latest price table
```
chrismaree, line 16: I might be missing something that came in a previous PR, but where does the fetching actually happen?

chrismaree (approved):

nicholaspai (approved):
Optional nit: rename zrx to 0x which is the name of the protocol, instead of the name of the protocol token

packages/api/src/libs/queries.ts
daywiss: prices are returned as a float and on return i convert it to we

packages/api/src/services/market-prices.ts
daywiss: yea, there was a previous pr to this https://github.com/UMAprotocol/protocol/pull/3203.  This uses the 0x api

daywiss:
> Optional nit: rename zrx to 0x which is the name of the protocol, instead of the name of the protocol token

The name of the project is annoying becuase js variables can't start with a number, which is why i ended up using zrx. Open to alternatives though @nicholaspai

packages/api/src/libs/queries.ts
chrismaree: right buy what are the decimals? i.e lets say the price is 1000 USDC per eth. is it represented as "1000" or is it "1000e6"  using the decimals for USDC?

chrismaree (approved):

PR #3207, opened 2021-07-08
fix(LSP): Continuous audit commenting and require modifications

chrismaree:
**Motivation**

A few small final comments were left on the LSP contracts after the addition of the OO liveness time, OO proposal bond and pair name. In particular, these comments were as follows:

This PR introduces a new `pairName` property of `LongShortPair` financial contracts and allows the creator to set custom values for the optimistic oracle's liveness and bond. It also refactors the `LongShortPair` and `LongShortPairCreator` contracts to use an initialization struct instead of a large number of individual parameters.

Our comments are:

- The `createLongShortPair` [`@param` comment](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L71) does not list the `pairName` field.
- The `pairName` is not validated to have non-zero length.
- The `optimisticOracleLivenessTime` is described as a "timer" (instead of a "time"), in its [definition](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L49), the [`LongShortPair` constructor comments](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L129) and the [`createLongShortPair` function comments](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPairCreator.sol#L84).
- Our assumption was that the `expirationTimestamp` is [only 64 bits long](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L60) so that it can share a storage location with the [`contractState` variable](https://github.com/UMAprotocol/protocol/blob/4eaa2e1c23065503336e61bf16916cefbab046bc/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L56). However, the new `pairName` variable splits them into separate storage slots. Consider defining `pairName` after the `expirationTimestamp`, or including a comment explaining why the non-stardard `uint64` type is used.

**Summary**

This PR implements the required changes as required in the audit.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
+++ b/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
@@ -55,9 +55,9 @@ contract LongShortPair is Testable, Lockable {
     enum ContractState { Open, ExpiredPriceRequested, ExpiredPriceReceived }
     ContractState public contractState;
 
-    string public pairName;
-
     uint64 public expirationTimestamp;
+
+    string public pairName;
 
     // Amount of collateral a pair of tokens is always redeemable for.
     uint256 public collateralPerPair;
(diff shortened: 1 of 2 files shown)


packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -55,10 +55,10 @@ contract LongShortPair is Testable, Lockable {
     enum ContractState { Open, ExpiredPriceRequested, ExpiredPriceReceived }
     ContractState public contractState;
 
-    string public pairName;
-
     uint64 public expirationTimestamp;
 
+    string public pairName;
```
nicholaspai, line 60: I know the suggestion was to not split up `contractState` and `expirationTimeestamp` but for the sake of future maintainers, I'd like to still include a @dev comment here explaining why `pairName` needs to follow `expirationTimestamp` which is a uint64

nicholaspai (approved):
Looks like you addressed all changes suggested to #3184, just one nit about an additional comment!

mrice32 (commented):
Looks great!

mrice32 (approved):

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
mrice32: Good point! +1

PR #3211, opened 2021-07-09
feat(LSP): Add rounding protection and additional unit tests

chrismaree:
**Motivation**

After some Kovan testing by the UMA team, it became apparent there are some cases where rounding can result in dust minting for free. This PR fixes this case and adds unit tests to validate the solution.

**Summary**

Update implementation and tests to better deal with rounding errors.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
+++ b/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
@@ -179,7 +179,7 @@ contract LongShortPair is Testable, Lockable {
      * @return collateralUsed total collateral used to mint the synthetics.
      */
     function create(uint256 tokensToCreate) public preExpiration() nonReentrant() returns (uint256 collateralUsed) {
-        collateralUsed = FixedPoint.Unsigned(tokensToCreate).mul(FixedPoint.Unsigned(collateralPerPair)).rawValue;
+        collateralUsed = FixedPoint.Unsigned(tokensToCreate).mulCeil(FixedPoint.Unsigned(collateralPerPair)).rawValue;
 
         collateralToken.safeTransferFrom(msg.sender, address(this), collateralUsed);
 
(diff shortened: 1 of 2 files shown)


packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -179,7 +179,9 @@ contract LongShortPair is Testable, Lockable {
      * @return collateralUsed total collateral used to mint the synthetics.
      */
     function create(uint256 tokensToCreate) public preExpiration() nonReentrant() returns (uint256 collateralUsed) {
-        collateralUsed = FixedPoint.Unsigned(tokensToCreate).mul(FixedPoint.Unsigned(collateralPerPair)).rawValue;
+        // Note the use of mulCeil to prevent small collateralPerPair causing rounding of collateralUsed to 0 enabling
+        // callers to mint dust LSP tokens without paying any collateral.
+        collateralUsed = FixedPoint.Unsigned(tokensToCreate).mulCeil(FixedPoint.Unsigned(collateralPerPair)).rawValue;
```
nicholaspai, line 184: No need to change the mul to mulCeil in redeem right? Because in all situations it's ok if the contract has more collateral than expected. I.e. no one can steal collateral

nicholaspai (approved):
Nice fix and find @smb2796 , this looks like the right fix and I like the tests for redeem and settle. Do we not have to change any math in settle and redeem?

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
mrice32: Agree. I also looked through the settlement logic and it looks like it always rounds down too.

mrice32 (approved):
LGTM! The tests are very thorough, +1.

PR #3214, opened 2021-07-09
feat(common): add compiler overrides to allow most contracts to be gas optimized

mrice32:
**Motivation**

We can't compile with high values for runs (for user gas optimization) because we have a few large contracts that require low runs values to be deployable.

**Summary**

Use hardhat options to compile most contracts with a high runs value, but leave the large contracts with a runs value of 199.

Note: this makes the compilation process take a little longer.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -2,7 +2,7 @@ module.exports = {
   env: {
     node: true,
     mocha: true,
-    es2020: true,
+    es2021: true,
   },
   extends: ["plugin:prettier/recommended", "eslint:recommended"],
   plugins: ["prettier", "mocha"],
--- a/packages/common/src/HardhatConfig.js
+++ b/packages/common/src/HardhatConfig.js
@@ -53,6 +53,7 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
         "contracts/financial-templates/perpetual-multiparty/PerpetualLiquidatable.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
         "contracts/financial-templates/expiring-multiparty/Liquidatable.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
         "contracts/oracle/implementation/Voting.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
+        "contracts/oracle/implementation/test/VotingTest.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
       },
     },
     networks: {
--- a/packages/common/src/HardhatConfig.js
+++ b/packages/common/src/HardhatConfig.js
@@ -20,15 +20,48 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
   // Solc version defined here so etherscan-verification has access to it
   const solcVersion = "0.8.4";
 
+  // This is a list of the contracts too large to compile at runs = 999999.
+  // Compilation settings are overridden for these to allow them to compile without going over the bytecode limit.
+  const bigContracts = [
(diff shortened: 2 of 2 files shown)


packages/common/src/HardhatConfig.js
```diff
@@ -20,14 +20,38 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
-        optimizer: {
-          enabled: true,
-          runs: 199,
+      compilers: [
+        {
+          version: solcVersion,
+          settings: {
+            optimizer: {
+              enabled: true,
+              runs: 1000000,
```
nicholaspai, line 43: Curious why not use 1_000_000 which is more readable?

packages/common/src/HardhatConfig.js
```diff
@@ -93,6 +117,8 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
       },
     },
   };
+
+  console.log(defaultConfig.solidity.overrides);
```
nicholaspai, line 121: Leftover console.log

nicholaspai (commented):
Just one comment on changing default runs from 199 to 200

packages/common/src/HardhatConfig.js
```diff
@@ -20,15 +20,48 @@ function getHardhatConfig(configOverrides, workingDir = "./") {
...
+  const overrides = Object.fromEntries(
+    bigContracts.map((name) => {
+      return [
+        name,
+        {
+          version: solcVersion,
+          settings: {
+            optimizer: {
+              enabled: true,
+              runs: 199,
```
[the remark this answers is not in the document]
nicholaspai: This doesn't show up anymore because the comment was made on older code. Tagging you for visibility @mrice32

chrismaree (approved):
this is pretty awesome. +100 for hardhat. 

There are a few other places we can simplify/remove things now that we don't need to deal with truffles pain points.

packages/common/src/HardhatConfig.js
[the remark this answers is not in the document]
mrice32: It shouldn't matter if we're verifying through hardhat since it sends these settings to etherscan, but I can try just to make things more standard.

packages/common/src/HardhatConfig.js
mrice32: VSCode kept saying it couldn't parse it. I'll try adding it and then running to see if VSCode is just wrong.

packages/common/src/HardhatConfig.js
mrice32: Nice catch!

packages/common/src/HardhatConfig.js
nicholaspai: Ok not a big deal

packages/common/src/HardhatConfig.js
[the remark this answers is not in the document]
nicholaspai: yeah I think 200 is preferable to 199 in a vacuum

.eslintrc.js
```diff
@@ -2,7 +2,7 @@ module.exports = {
   env: {
     node: true,
     mocha: true,
-    es2020: true,
+    es2021: true,
```
mrice32, line 5: I had to change this to use the `_` syntax in numbers.

nicholaspai (approved):
LGTM assuming all tests pass!

PR #3220, opened 2021-07-12
improve(optimistic-oracle): Remove dependencies on @uma/core and truffle

nicholaspai:

**Motivation**

We'd like to make a push to remove all package dependencies on core and truffle. This means each package should:
- Not import `@uma/core`
- Use the `hardhat` test framework in unit test files and specifically use its `getContract` and `web3` modules to mirror how `@uma/core` unit test implementations
- Use `@uma/contracts-node` in production files to import contract ABI's and deployed addresses. This package does not instantiate `truffle` contracts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/optimistic-oracle/test/index.js
```diff
@@ -5,39 +5,39 @@ const Main = require("../index.js");
...
-    finder = await Finder.new();
-    timer = await Timer.new();
-    mockOracle = await MockOracle.new(finder.address, timer.address);
-
+    const accounts = await web3.eth.getAccounts();
+    await runDefaultFixture(hre);
     // Deploy a new OptimisticOracle.
-    optimisticOracle = await OptimisticOracle.new("120", finder.address, timer.address);
+    const finder = await Finder.deployed();
+    optimisticOracle = await OptimisticOracle.deployed();
```
nicholaspai, line 35: Had to change this to `.deployed` instead of `.new()` because otherwise the test on line fails. It looks like the bot sees the `OptimisticOracle` address as the one matching `deployments.get("OptimisticOracle")` which is doesn't seem to change even after I run `OptimisticOracle.new()`.

Instead, it just seems to use the address set when all the `deploy/` scripts are run in the test fixture.

nicholaspai:
Pausing this PR until we can get a new contract deployed via `hardhat-contract.new()` to be read by `@uma/core:getAddress`. Currently this doesn't work because `getAddress` uses truffle artifacts under the hood instead of hardhat deployments

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
```diff
@@ -817,7 +818,7 @@ async function createReferencePriceFeedForFinancialContract(
 
 function getFinancialContractIdentifierAtAddress(web3, financialContractAddress) {
   try {
-    const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, "1.2.0");
```
nicholaspai, line 820: Is it OK to just use default latest version from `hre.getContract`?

mrice32: I think so. But if not, we'll need to import a pinned version of core to get them.

mrice32 (approved):
This looks awesome! Very nice work!

chrismaree (approved):
LGTM! nice job

PR #3222, opened 2021-07-13
feat(linter): proposal to minimize object sizing

chrismaree:
**Motivation**

This PR introduces a somewhat opinionated refactor to all js within the repo. Basically prettier does not always try to minimize JS objects, as can be seen in their docs https://prettier.io/docs/en/rationale.html#multi-line-objects. Rather, they keep objects spread over multiple lines. This can make code more verbose/take up more space than it really needs to and a lot of the time makes things look pretty messy. 

For example this code:

```
defaultCreationParams = {
      ...defaultCreationParams,
      collateralAddress: collateral.address,
    };
```
could equally be represented with this code:
```
defaultCreationParams = { ...defaultCreationParams, collateralAddress: collateral.address };
```

**Summary**

Introduces a sweeping change to how we format objects.

**Details**

This change is not possible with prettier on it's own and rather relies on a combination of prettier and eslint. In particular, the eslint `object-curly-newline` (documented [here](https://eslint.org/docs/rules/object-curly-newline)) can enable this kind or redaction. This is used along with the previous prettier config to ensure the exact same styling as before, just over fewer lines.

This change also makes some modifications to how long running function parameters look.

before:
```
await bpToken1.mint(shareHolders[0], toWei("2"), {
        from: contractCreator,
      });
```

after
```
await bpToken1.mint(shareHolders[0], toWei("2"), { from: contractCreator });
```

in another case it will wrap like this:

```
const receipt = await newWhitelist.methods.addSupportedIdentifier(identifiersToWhitelist[i]).send({
            from: deployer,
          });
```

to:
```
const receipt = await newWhitelist.methods
    .addSupportedIdentifier(identifiersToWhitelist[i])
    .send({ from: deployer });

```

I feel that both cases are more readable better looking.

I'm very curious to see what @UMAprotocol/eng thinks about this proposal!

**Testing**

- [ ]  New unit tests created
- [X]  Untested


.eslintrc.js
```diff
@@ -15,6 +11,7 @@ module.exports = {
     "spaced-comment": ["error", "always", { exceptions: ["-", "+"] }],
     "mocha/no-exclusive-tests": "error",
     "no-console": 0,
+    "object-curly-newline": ["off", "never"],
```
chrismaree, line 14: this is the ONLY config change I made to the linter settings

daywiss (commented):
I agree in many cases this makes code more readable. In rare cases when you have a lot of props, or want to set it up to easily add/remove props, it makes sense to put them on new lines.  does this change imply that all objects get flattened no matter how many props?  I'm not super opinionated on this, but in general i think it makes sense for the developer to decide one way or the other.

packages/common/src/HardhatConfig.js
```diff
@@ -21,67 +21,24 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
...
-      },
-      mainnet: {
-        chainId: 1,
-        url: getNodeUrl("mainnet", true),
-        accounts: { mnemonic },
-      },
-    },
-    mocha: {
-      timeout: 1800000,
+      hardhat: { gas: 11500000, blockGasLimit: 11500000, allowUnlimitedContractSize: false, timeout: 1800000 },
```
nicholaspai, line 26: This is a nice file diff

mrice32 (approved):

chrismaree:
> I agree in many cases this makes code more readable. In rare cases when you have a lot of props, or want to set it up to easily add/remove props, it makes sense to put them on new lines. does this change imply that all objects get flattened no matter how many props? I'm not super opinionated on this, but in general i think it makes sense for the developer to decide one way or the other.

Basically what this will do is try to is try keep it on as few lines as possible. if the object can be expressed on fewer lines, it will do that.

if the object is big and has a lot of props (like a config file) it will then span over multiple lines but still trying to keep it as short as posible.

daywiss (approved):

PR #3223, opened 2021-07-13
feat(api): 1680 cache global tvl rather than compute it for every query

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/1680/add-latest-global-tvl-value-and-expose-in-api

**Summary**

Theres an existing call to calculate tvl based on a number of emp addresses. This just calls it on an interval and caches it in the latest tvl table under the key "global".  This is a prerequisite to adding global tvl history, which we currently do not have, we only have tvl per EMP.

**Details**

No api changes were necessary, but when a user requests global tvl, it returns the cached version now rather than doing the math for the request. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/1680/add-latest-global-tvl-value-and-expose-in-api

### DIFF
--- a/packages/api/src/libs/queries.ts
+++ b/packages/api/src/libs/queries.ts
@@ -154,6 +154,11 @@ export default (appState: Dependencies) => {
     const addresses = Array.from(appState.registeredEmps.values());
     return sumTvm(addresses, currency);
   }
+  async function getGlobalTvl(currency: CurrencySymbol = "usd") {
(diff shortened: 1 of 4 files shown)


chrismaree (approved):
LGTM!

mrice32 (approved):
Looks great!

PR #3229, opened 2021-07-14
feat(core): add tasks to manage artifacts and deployments

mrice32:
**Motivation**

Core is currently dependent on truffle for its index file. We should migrate away from this dependency.

We have no way to load our core artifact addresses into hardhat.

**Summary**

This creates two new packages that are meant for consumption by node and frontend applications, respectively.

These packages have no runtime dependencies to remain lightweight.

`contracts-node`:
- `hardhat generate-contracts-node` generates the index file for this package.
- Build scripts and typescript config ensure that all dependencies living within core are copied into the dist folder to ensure users have them at runtime.
- Artifacts are resolved when requested to so node doesn't have to load all the artifacts into memory at startup. It lazily loads the modules that are requested.
- Functions to get artifacts look like `getAbi(contractName)` or `getBytecode(contractName)`.

`contracts-frontend`:
- `hardhat generate-contracts-frontend` generates the index file for this package.
- - Build scripts and typescript config ensure that all dependencies living within core are copied into the dist folder to ensure users have them at runtime.
- Artifacts are all directly imported at load time to allow build processes, like webpack to completely drop artifacts that aren't used.
- Functions to get artifacts look like `get[contractName]Abi()` or `get[contractName]Bytecode()` to ensure unused functions and artifacts get dropped from bundles.

A few other additions:
- Adds a new hardhat script called `load-addresses` that loads the networks/ files into the hardhat deployments folder for use by hardhat deploy. This is now part of the build process in core. We were able to drop the external field from the hardhat config because these addresses are readily available in deployments/ now.
- Typechain types are now generated through hardhat. The output should be the same as before.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/common/src/hardhat/tasks/artifacts.js
```diff
@@ -0,0 +1,206 @@
...
+    // do nothing.
+  }
+}
+
+async function getArtifactPathList(hre, relativeTo) {
+  const artifactPaths = await hre.artifacts.getArtifactPaths();
+
+  // Generate a unique list of artifacts and paths to them. Unique is necessary because there are some redundantly
+  // named contracts.
+  return uniqBy(
```
chrismaree, line 20: nice use of lodash.

packages/common/src/hardhat/tasks/artifacts.js
```diff
@@ -0,0 +1,206 @@
...
+  );
+}
+
+function getCorePath(hre, relativeTo) {
+  const artifactPath = hre.config.paths.artifacts ? path.join(hre.config.paths.artifacts, "../") : "./";
+  return `./${path.relative(relativeTo, artifactPath)}`;
+}
+
+function getAddressesMap(hre) {
+  // Generate a map of name => chain id => address.
```
chrismaree, line 35: how does this work if there are multiple named deployments for one contract. i.e multiple LSP factories, all equally "valid"? based off this I'm guessing it just has the latest but in general how would we expand this to accommodate more flexibility in this regard?

packages/common/src/hardhat/tasks/artifacts.js
```diff
@@ -0,0 +1,206 @@
...
+      if (!addresses[name]) {
+        addresses[name] = {};
+      }
+      addresses[name][chainId] = address;
+    }
+  }
+  return addresses;
+}
+
+task("generate-contracts-frontend", "Generate typescipt for the contracts-frontend package")
```
chrismaree, line 56: in this format, if I understand correctly, the artifacts still contain the bytecode of the contract. is this something we'd like to consider removing? I guess however that if you use the `typechain` exports then we'll end up tree-shaking away all the artifacts that contain bytecode so it amounts to just having the interfaces anyway.

packages/common/src/hardhat/tasks/artifacts.js
```diff
@@ -0,0 +1,206 @@
...
+      if (fs.existsSync(`typechain/ethers/factories/${contractName}__factory.ts`))
+        fs.appendFileSync(out, `  ${contractName}__factory as ${contractName}Ethers__factory,\n`);
+    });
+    fs.appendFileSync(out, '} from "../typechain/ethers";\n');
+
+    // Write abi and bytecode for the browser file.
+    // Note: the idea behind writing the functions this way is to make them as optimized as possible for tree-shaking
+    // to remove any unused json files. In modern versions of webpack, this should allow absolutely _no_ artifact
+    // information that isn't needed to be pulled in.
+    // artifacts.forEach(({ contractName, relativePath }) =>
```
chrismaree, line 84: is this meant to be commented out like this?

packages/contracts-frontend/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@uma/contracts-frontend",
+  "version": "0.1.0",
+  "description": "UMA smart contracts and unit tests",
+  "devDependencies": {
+    "@uma/core": "^2.5.0",
```
chrismaree, line 6: so the magic here, really, is that this is a `devDependency`, right? If this code is generated dynamically like this do you really need to include these dependencies anyways? 

I guess you do seeing you've got the logic for generating ts and the build within this package.json rather than within core or common so that does make sense. if you placed the generation logic in one of _those_ packages then we would not need either of these dependencies, right?

However, it might make more sense the way you have it now as this'll let you use core logic within unit tests/integration test from front-end dependents but can shake away all unused stuff in production.

packages/core/networks/1_args.json
```diff
@@ -1,8 +0,0 @@
-{
```
chrismaree, line 1: finally :)

chrismaree (commented):
This is pretty awesome stuff! I left a bunch of comments/questions, mainly for my own understanding. Thanks for putting all the time and effort into this!

packages/common/src/hardhat/tasks/artifacts.js
mrice32: If you intend to have multiple deployments of a contract, you can use the new field "deploymentName" in the config to denote the name of the deployment if it is different from the contract name (it defaults to contractName if deploymentName isn't there). Then, that address is accessed using that name rather than the contract name. That's why I elected to use the more general term "name" in the getAddress function.

packages/common/src/hardhat/tasks/artifacts.js
mrice32: Yeah, the intention is that we make bytecode accessible but tree-shake it away when it isn't used.

I spent some time testing this and it looks like the tree shaking works as expected. I did have to copy parts of the artifacts into the ts source to get that to work, though.

packages/common/src/hardhat/tasks/artifacts.js
mrice32: Oops, will remove! Thanks for pointing that out!

packages/contracts-frontend/package.json
mrice32: Yeah, we need to explicitly provide the deps here to tell lerna that those packages must be built before this one. Otherwise the artifacts might not exist when we start building. ethers is also required because the typescript compilation fails if you don't have it I think.

packages/core/networks/1_args.json
mrice32: :)

packages/common/src/hardhat/tasks/artifacts.js
mrice32: Yeah, this is an interesting case. Truffle has been warning us for a while that we have some duplicate contract names. This is mostly for interfaces that we declare within contracts, like the FPLs, so they aren't contracts we really care about accessing here. So this was just to avoid generating duplicate functions.

packages/contracts-frontend/package.json
mrice32: But to your initial question, yes. The dev dependencies mean that none of these should get packaged up in the final bundle.

packages/contracts-node/README.md
```diff
@@ -0,0 +1,71 @@
+# @uma/contracts-node
```
mrice32, line 1: @daywiss @nicholaspai @chrismaree PTAL at the node-centric README. This should be slightly better than the old import style. One big advantage is that the typechain types no longer require you to dig into the package directories as we've done in the past. You should be able to import them directly from the `@uma/contracts-node` package.

There is also no versioning anymore. To get versioning, you'll need to alias old versions of core or this package.

packages/contracts-frontend/README.md
```diff
@@ -0,0 +1,73 @@
+# @uma/contracts-frontend
```
mrice32, line 1: @Gamaranto @Tulun PTAL at this readme. This package should be strictly better than `@uma/core`. This package should be fully frontend compatible -- you should be able to `import { getVotingAbi, getVotingAddress, getVotingBytecode } from "@uma/contracts-frontend";` etc from a frontend.

The biggest change with this package is bundle size optimization. This package has 0 dependencies. The index file is over 2MBs because it contains all of the artifacts. However, unused functions get dropped from that file through webpack tree shaking. So, assuming you're only a few contracts, the package should end up taking up <10-20kb per contract factory that you import (according to my testing on large contracts, like Voting).

Another advantage is that you should now be able to pull in addresses by chain id from our repo: `getVotingAddress(chainId)`. This is something that I don't think you were able to do before.

The last nicety is that you can now import ethers or web3 contract types and factories a bit more easily:

```ts
import type { VotingEthers, VotingWeb3 } from "@uma/contracts-frontend";
import { VotingEthers__factory } from "@uma/contracts-frontend";
```

packages/common/src/hardhat/tasks/artifacts.js
chrismaree: makes sence!

packages/contracts-frontend/README.md
```diff
@@ -0,0 +1,73 @@
...
+const votingAbi = getAbiVotingAbi()
+
+// Gets the bytecode string for the voting contract.
+// This should usually only be required if you need to deploy a voting contract (rare).
+const votingBytecode = getVotingBytecode()
+
+// Gets the address of the voting contract for chain id 1 (eth mainnet).
+// If there is no single canonical deployment of a particular contract on the network provided, this will fail.
+const chainId = 1
+const votingAddress = getVotingAddress(chainId)
```
chrismaree, line 31: would it be useful to change this to default return mainnet if no `chainId` is provided?

packages/contracts-node/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+```ts
+import type { VotingWeb3 } from "@uma/contracts-node"
+import { getAbi, getAddress } from "@uma/contracts-node"
+
+const VOTING_ABI = getAbi("Voting")
+const NETWORK_ID = 1
+const VOTING_ADDRESS = getAddress("Voting", NETWORK_ID)
+
+const voting = new web3.eth.Contract(VOTING_ABI, VOTING_ADDRESS) as VotingWeb3
```
chrismaree, line 70: perhaps it;'s useful to show how to deploy an instance from this? it would look something like this:
```
let voting = await new web3Instance.eth.Contract(getAbi("Voting"), undefined)
    .deploy({ data: getBytecode("Voting"), arguments: props })
    .send({ from: account });
```

chrismaree: or prehaps rather than showing how to deploy a contract just link to the web3 documentation here: https://web3js.readthedocs.io/en/v1.4.0/web3-eth-contract.html?highlight=deploy#contract-deploy

packages/contracts-node/README.md
```diff
@@ -0,0 +1,71 @@
...
+```
+
+In addition to artifact support, typescript types (and sometimes factories) are available for ethers and web3.
+
+### Ethers
+
+The best support is for Ethers contract types. To construct an Ethers contract, import one of the ethers factories:
+
+```ts
+import { VotingEthers__factory, getAddress } from "@uma/contracts-node" // Factory to create ethers instance.
```
daywiss, line 48: this looks fine to me, id like to bring up potentially using types from typechain outside of the factory and contract. Don't know though if you are under the mercy of typechain, but ideally types for Events, and in lieu of that, access to the common.ts file which can be used for type inference. currently i have to do this: `"@uma/core/types/contract-types/ethers/commons"`.  This is also important if we want event types on the FE as well

daywiss (approved):
this is some crazy black magic, cant really comment on the artifact tasks but lgtm

packages/common/src/hardhat/tasks/artifacts.js
mrice32: The existing duplicates aren't contracts we'd really use outside of solidity. They only get exported because solidity has no way to construct an interface that isn't exported.

Open to ideas for how to represent duplicates in typescript, but if we want a simple name-based export strategy, I think we have to remove duplicates to get it to compile.

packages/contracts-node/README.md
mrice32: Will add this!

packages/contracts-frontend/README.md
mrice32: This is an interesting idea. As we move toward having multiple "production" networks (like polygon), it seems a little scary to let this silently default. Thoughts?

packages/contracts-node/README.md
mrice32: Doing blanket exports is tricky to avoid any sort of naming conflicts (and I played with namespaces, but also ran into problems there).

I looked at the Ethers contract files that typechain produced, and it appears as if the Event types don't exist outside of the contract class, itself (each contract and factory file has exactly one export).

I can export the the general event types from common under *Ethers aliases so they don't conflict with anything else.

Web3 does have some specific event types, but those are really tricky since they would need to be exported using a `*` since we don't know the events ahead of time. But since event names are certainly repeated across contracts, there will be collisions. I tried to manage this with namespaces, but ran into tons of ts compilation issues. Since nobody is using these web3 event types right now, I decided to leave that issue for another day.

Thoughts on me just adding the common.ts exports with the `Ethers` suffix?

packages/common/src/hardhat/tasks/artifacts.js
mrice32: Oh! One last point on this. Duplicate contract names aren't supported by typechain in general, I don't think. They create one flat dir that contains files called `{contractName}.ts`, so it's impossible for them to represent duplicate contract names.

packages/contracts-node/package.json
```diff
@@ -0,0 +1,39 @@
...
+    "url": "git+https://github.com/UMAprotocol/protocol.git"
+  },
+  "files": [
+    "dist/**/*"
+  ],
+  "main": "dist/contracts-node/generated/index.js",
+  "module": "dist/contracts-node/generated/index.js",
+  "types": "dist/contracts-node/generated/index.d.ts",
+  "sideEffects": false,
+  "side-effects": false,
```
nicholaspai, line 28: nit: what are `side-effects` and `sideEffects`?

nicholaspai (commented):
Really great code modifying the `fs` system. I need to spend more time understanding the new logic inside `artifacts.js` but the resultant interface looks great. One question: what happens in the following situation. Assume we're running the following code on the local `hardhat` network:

```js
const deployed = await MyContract.deployed();
assert.isTrue(deployed.options.address === getAddress("MyContract")
```

Initially, the `hardhat` deploy scripts deploy `MyContract` and we can correctly read its address on the local  network. Now, let's say we want to deploy a new contract:

```js
const newlyDeployed = await MyContract.new();
```

At this point, does `getAddress("MyContract") === newlyDeployed.options.address`?

I ask because  this specific feature is blocking #3220 from easily migrating its tests from `truffle` to `hardhat` and specifically this [comment](https://github.com/UMAprotocol/protocol/pull/3220#discussion_r668318071).

packages/contracts-frontend/README.md
Gamaranto: Awesome work! The API is almost identical to the previous one we used and tree shaking was a much needed change!

mrice32:
> Really great code modifying the `fs` system. I need to spend more time understanding the new logic inside `artifacts.js` but the resultant interface looks great. One question: what happens in the following situation. Assume we're running the following code on the local `hardhat` network:
> 
> ```js
> const deployed = await MyContract.deployed();
> assert.isTrue(deployed.options.address === getAddress("MyContract")
> ```
> 
> Initially, the `hardhat` deploy scripts deploy `MyContract` and we can correctly read its address on the local network. Now, let's say we want to deploy a new contract:
> 
> ```js
> const newlyDeployed = await MyContract.new();
> ```
> 
> At this point, does `getAddress("MyContract") === newlyDeployed.options.address`?
> 
> I ask because this specific feature is blocking #3220 from easily migrating its tests from `truffle` to `hardhat` and specifically this [comment](https://github.com/UMAprotocol/protocol/pull/3220#discussion_r668318071).

This is a great question. You need to use a different deployment mechanism to ensure the deployment gets _saved_ (this is what happens in our fixture: https://github.com/UMAprotocol/protocol/blob/master/packages/core/deploy/001_deploy_finder.js).

```js
// Deploy a new contract
await hre.deployments.deploy("ContractName", { args: YOUR_ARGUMENTS, from: FROM_ADDRESS });

// Alternatively, you can deploy and save the deployment manually (I think!)
const newlyDeployed = await MyContract.new();
hre.deployments.save("ContractName", { address: newlyDeployed.options.address, abi: MyContract.abi });
assert(getAddress("MyContract", await web3.eth.getChainId()) === newlyDeployed.options.address);
```

nicholaspai:
> > Really great code modifying the `fs` system. I need to spend more time understanding the new logic inside `artifacts.js` but the resultant interface looks great. One question: what happens in the following situation. Assume we're running the following code on the local `hardhat` network:
> > ```js
> > const deployed = await MyContract.deployed();
> > assert.isTrue(deployed.options.address === getAddress("MyContract")
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Initially, the `hardhat` deploy scripts deploy `MyContract` and we can correctly read its address on the local network. Now, let's say we want to deploy a new contract:
> > ```js
> > const newlyDeployed = await MyContract.new();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > At this point, does `getAddress("MyContract") === newlyDeployed.options.address`?
> > I ask because this specific feature is blocking #3220 from easily migrating its tests from `truffle` to `hardhat` and specifically this [comment](https://github.com/UMAprotocol/protocol/pull/3220#discussion_r668318071).
> 
> This is a great question. You need to use a different deployment mechanism to ensure the deployment gets _saved_ (this is what happens in fixtures: https://github.com/UMAprotocol/protocol/blob/master/packages/core/deploy/001_deploy_finder.js).
> 
> ```js
> // Deploy a new contract
> await hre.deployments.deploy("ContractName", { args: YOUR_ARGUMENTS, from: FROM_ADDRESS });
> 
> // Alternatively, you can deploy and save the deployment manually (I think!)
> const newlyDeployed = await MyContract.new();
> hre.deployments.save("ContractName", { address: newlyDeployed.options.address, abi: MyContract.abi });
> assert(getAddress("MyContract", await web3.eth.getChainId()) === newlyDeployed.options.address);
> ```

Can you add this to the README's? Super helpful feature

packages/contracts-node/README.md
daywiss: yea i am good with that since thats all i use currently

mrice32:
> > > Really great code modifying the `fs` system. I need to spend more time understanding the new logic inside `artifacts.js` but the resultant interface looks great. One question: what happens in the following situation. Assume we're running the following code on the local `hardhat` network:
> > > ```js
> > > const deployed = await MyContract.deployed();
> > > assert.isTrue(deployed.options.address === getAddress("MyContract")
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Initially, the `hardhat` deploy scripts deploy `MyContract` and we can correctly read its address on the local network. Now, let's say we want to deploy a new contract:
> > > ```js
> > > const newlyDeployed = await MyContract.new();
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > At this point, does `getAddress("MyContract") === newlyDeployed.options.address`?
> > > I ask because this specific feature is blocking #3220 from easily migrating its tests from `truffle` to `hardhat` and specifically this [comment](https://github.com/UMAprotocol/protocol/pull/3220#discussion_r668318071).
> > 
> > 
> > This is a great question. You need to use a different deployment mechanism to ensure the deployment gets _saved_ (this is what happens in fixtures: https://github.com/UMAprotocol/protocol/blob/master/packages/core/deploy/001_deploy_finder.js).
> > ```js
> > // Deploy a new contract
> > await hre.deployments.deploy("ContractName", { args: YOUR_ARGUMENTS, from: FROM_ADDRESS });
> > 
> > // Alternatively, you can deploy and save the deployment manually (I think!)
> > const newlyDeployed = await MyContract.new();
> > hre.deployments.save("ContractName", { address: newlyDeployed.options.address, abi: MyContract.abi });
> > assert(getAddress("MyContract", await web3.eth.getChainId()) === newlyDeployed.options.address);
> > ```
> 
> Can you add this to the README's? Super helpful feature

@nicholaspai will do!

mrice32:
@nicholaspai @chrismaree @daywiss I think I've responded to all of your comments. PTAL.

nicholaspai (approved):
LGTM

daywiss:
> @nicholaspai @chrismaree @daywiss I think I've responded to all of your comments. PTAL.

@mrice32 One small ask if you could update the readme to show how to import common from ethers. otherwise it looks good to me

packages/contracts-frontend/README.md
chrismaree: good point. In that case explicitly requiring the user to define the net id sounds better!

chrismaree (approved):
This is super. Thanks for all the hard work put into this PR and for addressing all the comments! excited to have these lightweight packages!

One reminding question I have: how does this technique play with how we do releases? Is it sufficient to keep the current release process (we release _all_ packages at the same time when doing a release) or would it make sense to have the ability to just bump the version in these packages? More generally, how do you see the update -> release life cycle happening with these packages? does this have to run in lockstep with `core` or can they be out of sync?

mrice32:
> This is super. Thanks for all the hard work put into this PR and for addressing all the comments! excited to have these lightweight packages!
> 
> One reminding question I have: how does this technique play with how we do releases? Is it sufficient to keep the current release process (we release _all_ packages at the same time when doing a release) or would it make sense to have the ability to just bump the version in these packages? More generally, how do you see the update -> release life cycle happening with these packages? does this have to run in lockstep with `core` or can they be out of sync?

These packages do not depend on core at runtime, so they can be released without a core release, but our release process still is tricky to use unless you release everything at the same time.

Another issue is that other packages may grow to depend on these packages rather than core. In that case, these packages might need to be released whenever those dependent ones are.

PR #3231, opened 2021-07-14
feat(api): 1445 add lsp state ingestion service

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1445/ability-to-pull-in-token-information-for-lsp

**Summary**
Adds 2 lsp ingestion services. One that monitors the creator contract for newly launched tokens, and another which uses multicall to query lsp state. The state gets put into global app state. Api queries will be in the next PR.

**Details**

This does not expose lsp data through the API yet, just ingestion and puts into tables for later querying.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1445/ability-to-pull-in-token-information-for-lsp

### DIFF
--- a/packages/api/package.json
+++ b/packages/api/package.json
@@ -22,7 +22,7 @@
     "@uma/sdk": "^0.3.0",
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
-    "ethers": "^5.1.4",
+    "ethers": "^5.1.3",
     "express": "^4.17.1",
     "isomorphic-fetch": "^3.0.0",
     "lodash": "^4.17.21",
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -16,7 +16,7 @@ async function run(env: ProcessEnv) {
   assert(env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL");
   assert(env.EXPRESS_PORT, "requires EXPRESS_PORT");
   assert(env.zrxBaseUrl, "requires zrxBaseUrl");
-  assert(env.multicallAddress, "requires multicallAddress");
+  assert(env.MULTI_CALL_ADDRESS, "requires MULTI_CALL_ADDRESS");
 
   const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL);
 
@@ -79,7 +79,7 @@ async function run(env: ProcessEnv) {
     // lsp related props. could be its own state object
     longAddresses: new Set<string>(),
     shortAddresses: new Set<string>(),
-    multicall: new Multicall(env.multicallAddress, provider),
+    multicall: new Multicall(env.MULTI_CALL_ADDRESS, provider),
     lsps: {
       active: lsps.JsMap("Active LSP"),
       expired: lsps.JsMap("Expired LSP"),
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -3,19 +3,20 @@ import Web3 from "web3";
 import { ethers } from "ethers";
 import moment from "moment";
 
-import { tables, Coingecko, utils } from "@uma/sdk";
+import { tables, Coingecko, utils, Multicall } from "@uma/sdk";
 
 import * as Services from "../../services";
 import Express from "../../services/express";
 import Actions from "../../services/actions";
 import { ProcessEnv, AppState } from "../..";
-import { empStats, empStatsHistory } from "../../tables";
+import { empStats, empStatsHistory, lsps } from "../../tables";
 import Zrx from "../../libs/zrx";
 
 async function run(env: ProcessEnv) {
   assert(env.CUSTOM_NODE_URL, "requires CUSTOM_NODE_URL");
   assert(env.EXPRESS_PORT, "requires EXPRESS_PORT");
   assert(env.zrxBaseUrl, "requires zrxBaseUrl");
+  assert(env.multicallAddress, "requires multicallAddress");
 
   const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL);
 
@@ -72,9 +73,19 @@ async function run(env: ProcessEnv) {
     lastBlock: 0,
     lastBlockUpdate: 0,
     registeredEmps: new Set<string>(),
+    registeredLsps: new Set<string>(),
     collateralAddresses: new Set<string>(),
     syntheticAddresses: new Set<string>(),
+    // lsp related props. could be its own state object
+    longAddresses: new Set<string>(),
+    shortAddresses: new Set<string>(),
+    multicall: new Multicall(env.multicallAddress, provider),
+    lsps: {
+      active: lsps.JsMap("Active LSP"),
+      expired: lsps.JsMap("Expired LSP"),
+    },
   };
+
   // services for ingesting data
   const services = {
     // these services can optionally be configured with a config object, but currently they are undefined or have defaults
@@ -94,6 +105,8 @@ async function run(env: ProcessEnv) {
     erc20s: Services.Erc20s(undefined, appState),
     empStats: Services.EmpStats({}, appState),
     marketPrices: Services.MarketPrices(undefined, appState),
+    lspCreator: Services.LspCreator({}, appState),
+    lsps: Services.Lsps(undefined, appState),
   };
 
   // services consuming data
@@ -101,9 +114,17 @@ async function run(env: ProcessEnv) {
 
   // warm caches
   await services.registry();
-  console.log("Got all emp addresses");
+  console.log("Got all EMP addresses");
+
+  await services.lspCreator.update();
+  console.log("Got all LSP addresses");
+
   await services.emps();
-  console.log("Updated emp state");
+  console.log("Updated EMP state");
+
+  await services.lsps.update();
+  console.log("Updated LSP state");
+
   await services.erc20s.update();
   console.log("Updated tokens");
 
@@ -138,7 +159,9 @@ async function run(env: ProcessEnv) {
     if (blockNumber - appState.lastBlockUpdate >= updateBlocks) {
       // update everyting
       await services.registry(appState.lastBlock, blockNumber);
+      await services.lspCreator.update(appState.lastBlock, blockNumber);
       await services.emps(appState.lastBlock, blockNumber);
+      await services.lsps.update();
       await services.erc20s.update();
       await services.empStats.update();
 
--- a/packages/api/src/index.ts
+++ b/packages/api/src/index.ts
@@ -1,6 +1,6 @@
 import uma from "@uma/sdk";
 import { ethers } from "ethers";
-import type { empStats, empStatsHistory } from "./tables";
+import type { empStats, empStatsHistory, lsps } from "./tables";
 import type Zrx from "./libs/zrx";
 
 export type Currencies = "usd";
@@ -30,6 +30,10 @@ export type AppState = {
   emps: {
     active: uma.tables.emps.JsMap;
     expired: uma.tables.emps.JsMap;
+  };
+  lsps: {
+    active: lsps.JsMap;
+    expired: lsps.JsMap;
   };
   prices: {
     usd: {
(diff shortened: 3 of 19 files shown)


packages/api/package.json
```diff
@@ -22,7 +22,7 @@
     "@uma/sdk": "^0.3.0",
     "body-parser": "^1.19.0",
     "cors": "^2.8.5",
-    "ethers": "^5.1.4",
+    "ethers": "^5.1.3",
```
daywiss, line 25: this needed to match cores version

packages/api/src/apps/api/app.ts
```diff
@@ -72,9 +73,19 @@ async function run(env: ProcessEnv) {
...
     collateralAddresses: new Set<string>(),
     syntheticAddresses: new Set<string>(),
+    // lsp related props. could be its own state object
+    longAddresses: new Set<string>(),
+    shortAddresses: new Set<string>(),
+    multicall: new Multicall(env.multicallAddress, provider),
+    lsps: {
+      active: lsps.JsMap("Active LSP"),
+      expired: lsps.JsMap("Expired LSP"),
+    },
```
daywiss, line 86: this is the global state, a bunch of new tables and data structures related to lsps are created. the new services require this as a dependency

packages/api/src/apps/api/app.ts
```diff
@@ -94,16 +105,26 @@ async function run(env: ProcessEnv) {
     erc20s: Services.Erc20s(undefined, appState),
     empStats: Services.EmpStats({}, appState),
     marketPrices: Services.MarketPrices(undefined, appState),
+    lspCreator: Services.LspCreator({}, appState),
+    lsps: Services.Lsps(undefined, appState),
```
daywiss, line 109: instantiating the 2 new services

packages/api/src/index.ts
```diff
@@ -70,10 +74,14 @@ export type AppState = {
...
+  registeredLsps: Set<string>;
   provider: Provider;
   web3: Web3;
   lastBlock: number;
   lastBlockUpdate: number;
   collateralAddresses: Set<string>;
   syntheticAddresses: Set<string>;
+  longAddresses: Set<string>;
+  shortAddresses: Set<string>;
+  multicall: uma.Multicall;
```
daywiss, line 86: typescript definition for updated app state

packages/api/src/libs/utils.ts
```diff
@@ -83,3 +83,49 @@ export function msToS(ms: number) {
...
+  return Object.fromEntries(
+    lodash.zip(calls, results).map((zipped) => {
+      const [method, result] = zipped;
+      if (method == null) return [];
+      if (result == null) return [];
+      const [key, map] = method;
+      return [key, map(result)];
+    })
+  );
+};
```
daywiss, line 131: this file was update with multicall helper functions. the multicall library in SDK is really bare bones, these helpers take care of parsing results to string/numbers and putting the results back into a key/value object. the multicall library doesnt do this and just returns a raw array of results.

packages/api/src/services/lsp-creator.ts
```diff
@@ -0,0 +1,33 @@
+import { clients } from "@uma/sdk";
```
daywiss, line 1: this file just uses the lsp-creator client and watches for new deployments, adds them to the registeredLsps set

packages/api/src/services/lsp-state.e2e.ts
```diff
@@ -0,0 +1,81 @@
+require("dotenv").config();
```
daywiss, line 1: this file is an end to end test which tests the lsp state service, just to make sure we get data back we expect. uses mainnet deployed contract.

packages/api/src/services/lsp-state.ts
```diff
@@ -0,0 +1,127 @@
+import * as uma from "@uma/sdk";
```
daywiss, line 1: most of the lsp state fetching logic. heavily uses multicall and is structured similarly to the emp-state service.

packages/api/src/services/lsp-state.ts
```diff
@@ -0,0 +1,127 @@
...
+    ["customAncillaryData", toString],
+    ["prepaidProposerReward", toString],
+  ];
+
+  const dynamicProps: [string, (x: any) => any][] = [
+    ["expiryPrice", toString],
+    ["expiryPercentLong", toString],
+    ["expirationTimestamp", toNumber],
+    ["contractState", toNumber],
+  ];
```
daywiss, line 34: this is an optimization. we define the properties we consider static and dynamic, so that we only need to call static functions the first time we see a new contract, where dynamic props will get called every time update runs.

packages/api/src/services/lsp-state.ts
```diff
@@ -0,0 +1,127 @@
...
+
+  async function batchRead(calls: [string, (x: any) => any][], instance: Instance, address: string) {
+    const result = await BatchRead(multicall)(calls, instance);
+    return {
+      address,
+      updated: nowS(),
+      ...result,
+    };
+  }
+  async function updateLsp(address: string, startBlock?: number | "latest", endBlock?: number) {
```
daywiss, line 44: this contains complex logic to optimize how often we call infura based on things like, have we seen this contract before, was it expired or active, or transitioning from expired to active.

packages/api/src/tables/lsps/js-map.ts
```diff
@@ -0,0 +1,21 @@
+import * as uma from "@uma/sdk";
```
daywiss, line 1: custom table for holding lsp contract state

packages/api/src/services/lsp-state.e2e.ts
```diff
@@ -0,0 +1,81 @@
...
+  "lsps" | "registeredLsps" | "provider" | "collateralAddresses" | "shortAddresses" | "longAddresses" | "multicall"
+>;
+
+const registeredContracts = ["0x372802d8A2D69bB43872a1AABe2bd403a0FafA1F"];
+
+describe("lsp-state service", function () {
+  let appState: Dependencies;
+  before(async function () {
+    assert(process.env.CUSTOM_NODE_URL);
+    assert(process.env.multicallAddress);
```
chrismaree, line 23: would it make sense to have this set as `MULTI_CALL_ADDRESS`?

packages/api/src/services/lsp-state.e2e.ts
```diff
@@ -0,0 +1,81 @@
...
+  AppState,
+  "lsps" | "registeredLsps" | "provider" | "collateralAddresses" | "shortAddresses" | "longAddresses" | "multicall"
+>;
+
+const registeredContracts = ["0x372802d8A2D69bB43872a1AABe2bd403a0FafA1F"];
+
+describe("lsp-state service", function () {
+  let appState: Dependencies;
+  before(async function () {
+    assert(process.env.CUSTOM_NODE_URL);
```
chrismaree, line 22: are these tests intended to run in CI?

packages/api/src/services/lsp-state.e2e.ts
```diff
@@ -0,0 +1,81 @@
...
+    assert.ok(result.expiryPrice);
+    assert.ok(result.expiryPercentLong);
+    assert.ok(result.expirationTimestamp);
+    assert.ok(result.contractState >= 0);
+  });
+  it("readStaticState", async function () {
+    const [address] = registeredContracts;
+    const service = Service(undefined, appState);
+    const instance = await uma.clients.lsp.connect(address, appState.provider);
+    const result = await service.utils.batchRead(service.utils.staticProps, instance, address);
```
chrismaree, line 58: these tests do do any validation on the _data_, and just check they are not null/are ok. do you see any merit in validating the correctness of the data returned?

packages/api/src/services/lsp-state.ts
```diff
@@ -0,0 +1,129 @@
...
+  AppState,
+  "lsps" | "registeredLsps" | "provider" | "collateralAddresses" | "shortAddresses" | "longAddresses" | "multicall"
+>;
+
+export default (config: Config, appState: Dependencies) => {
+  const { lsps, registeredLsps, provider, collateralAddresses, shortAddresses, longAddresses, multicall } = appState;
+
+  // default props we want to query on contract
+  const staticProps: [string, (x: any) => any][] = [
+    // TODO: deal with this when early contracts dont have pairName
```
chrismaree, line 18: I think there will only ever be one contract that does not have `pairName` which is the current mainnet range bond. all future ones will have this prop

packages/api/src/services/lsp-state.ts
```diff
@@ -0,0 +1,129 @@
...
+    ["longToken", toString],
+    ["shortToken", toString],
+    ["finder", toString],
+    ["financialProductLibrary", toString],
+    ["customAncillaryData", toString],
+    ["prepaidProposerReward", toString],
+  ];
+
+  const dynamicProps: [string, (x: any) => any][] = [
+    ["expiryPrice", toString],
```
chrismaree, line 32: how is `expiryPrice` and `expirationTimestamp` dynamic? once it is set it wont change, unlike contract state or `expiryPercentLong` which can change during the life of an LSP.

chrismaree: also, there is a state variable `expiryPrice` which might be useful to store

chrismaree (approved):
looks pretty awesome! great work on getting this plumbing set up. I had a few comments and questions but overall this loops pretty slick

packages/api/src/services/lsp-state.e2e.ts
daywiss: yea i dont mind, not super opinionated on the names here

packages/api/src/services/lsp-state.e2e.ts
daywiss: nope, any test with `e2e` in the filename means it wont run in ci, its run with `yarn e2e`

packages/api/src/services/lsp-state.e2e.ts
daywiss: thats interesting, havent thought of that.  i might need convincing, like if theres a particular scenario where we would need to do this for a piece of data. my thoughts in general are that we want as dumb a pass through to the contract as possible  and assume we can trust the values from it.  otherwise though i think it could be fairly easy to add validation on a per prop basis if you think we need it.

packages/api/src/services/lsp-state.ts
daywiss: ok cool. I was scoping out the lsp creator to see if other contracts have launched, but didnt see any others. I think it will be safe to enable this, but it was causing the e2e tests to fail currently.

packages/api/src/services/lsp-state.ts
daywiss: im considering static props set on contract creation, and dont change after that. I will move those to be static, good call. Just want to double check on `expiryPrice`, it doesnt seem like something that would be set at start of contract. 

> also, there is a state variable expiryPrice which might be useful to store

did you mean `expiryPercentLong`? it doesnt look like i have that one

packages/api/src/services/lsp-state.e2e.ts
daywiss: oh i just realized this is in the test, to validate the values we get back. hmm yea we could, though some of the dynamic props we might not want to for the test

packages/api/src/services/lsp-state.ts
daywiss: oh nm i see it

packages/api/src/services/lsp-state.e2e.ts
daywiss: added checks against expected value

packages/api/package.json
nicholaspai: How'd you figure this out?

packages/api/src/services/lsp-state.e2e.ts
nicholaspai: that would match the rest of the repo's style, to use all caps

nicholaspai (approved):

PR #3237, opened 2021-07-18
feat(lib): add SimpleSuccessToken and CappedYieldDollar FPLs

hlambur:
**Motivation**

Implements a FinancialProductLibrary to create Success Tokens, as introduced in this blog post: https://medium.com/uma-project/success-tokens-an-incentive-aligned-way-for-vc-funds-to-invest-in-daos-1e8b8244f2f4

**Summary**

Modified the CoveredCall FPL to include a base payout of 0.5% to longs (with the additional 0.5% payout determined by the value of the embedded call option).

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

n/a


nicholaspai (commented):
high level comment: we can't merge this in without unit tests

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -7,19 +7,24 @@ import "./LongShortPairFinancialProductLibrary.sol";
...
+ * collateralPerPair.
+ * - If the price is more than $20 then each long is worth 0.5 collateralPerPair plus 0.5 times the fraction of
+ * collateralPerPair that was in the money, and each short is worth the remaining collateralPerPair.
+ * - Say settlement price is $30.  Then expiryPercentLong = 0.5 + (0.5 * (30 - 20) / 30) = 0.6667. 
+ * If the collateralPerPair is 2, that means the long payout is 0.6667*2 = 1.3333 $SUSHI, which at a settlement
+ * price of $30 is worth $40. This is equivalent to the value of 1 $SUSHI plus the value of the $20 strike
+ * embedded call.
  */
-contract CoveredCallLongShortPairFinancialProductLibrary is LongShortPairFinancialProductLibrary, Lockable {
+contract SuccessTokenLongShortPairFinancialProductLibrary is LongShortPairFinancialProductLibrary, Lockable {
```
pemulis, line 27: Since we're working on a generalized success token library that could accept parameterized base payouts and custom variable payouts, I think we should rename this to something like `SushiSuccessTokenLongShortPairFinancialProductLibrary`, if we plan to merge and deploy this for a one-off use case, and save the `SuccessToken` name for the generalized version.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -7,19 +7,24 @@ import "./LongShortPairFinancialProductLibrary.sol";
...
+ * For example, consider a covered call option collateralized in SUSHI, with a strike a price of $20, 
+ * and collateralPerPair of 2.
+ * - If the price is less than $20 then the each long is worth 0.5 collateralPerPair and each short is worth 0.5
+ * collateralPerPair.
+ * - If the price is more than $20 then each long is worth 0.5 collateralPerPair plus 0.5 times the fraction of
+ * collateralPerPair that was in the money, and each short is worth the remaining collateralPerPair.
+ * - Say settlement price is $30.  Then expiryPercentLong = 0.5 + (0.5 * (30 - 20) / 30) = 0.6667. 
+ * If the collateralPerPair is 2, that means the long payout is 0.6667*2 = 1.3333 $SUSHI, which at a settlement
+ * price of $30 is worth $40. This is equivalent to the value of 1 $SUSHI plus the value of the $20 strike
+ * embedded call.
```
pemulis, line 25: Great description.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -58,16 +63,23 @@ contract CoveredCallLongShortPairFinancialProductLibrary is LongShortPairFinanci
...
+        // eg if SUSHI is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667% which if the
+        // collateralPerPair is 2, is worth 1.3333 $SUSHI, which is worth $40 if 1 $SUSHI is worth $30.
+        // This return value is strictly < 1, tending to 1 as the expiryPrice tends to infinity.
         return
-            (FixedPoint.Unsigned(uint256(expiryPrice)).sub(FixedPoint.Unsigned(contractStrikePrice)))
-                .div(FixedPoint.Unsigned(uint256(expiryPrice)))
-                .rawValue;
+            (FixedPoint.fromUnscaledUint(0.5).add(
+                FixedPoint.fromUnscaledUint(0.5).mul(
+                    FixedPoint.Unsigned(uint256(expiryPrice)).sub(FixedPoint.Unsigned(contractStrikePrice)))
```
pemulis, line 79: This math looks right, and just noting that in the generalized version we would replace lines 79-80 with the value returned by calling `percentageLongCollateralAtExpiry` on another financial product library (in this case, the covered call).

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -58,16 +63,23 @@ contract CoveredCallLongShortPairFinancialProductLibrary is LongShortPairFinanci
...
+        // Else, token expires to be worth the 0.5 of the collateral plus 0.5 * the fraction of a collateral token 
+        // that's in the money. 
+        // eg if SUSHI is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667% which if the
+        // collateralPerPair is 2, is worth 1.3333 $SUSHI, which is worth $40 if 1 $SUSHI is worth $30.
+        // This return value is strictly < 1, tending to 1 as the expiryPrice tends to infinity.
         return
-            (FixedPoint.Unsigned(uint256(expiryPrice)).sub(FixedPoint.Unsigned(contractStrikePrice)))
-                .div(FixedPoint.Unsigned(uint256(expiryPrice)))
-                .rawValue;
+            (FixedPoint.fromUnscaledUint(0.5).add(
```
pemulis, line 77: In the generalized version, this `0.5` will be a parameter set after deployment, not necessarily 50%.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -58,16 +63,23 @@ contract CoveredCallLongShortPairFinancialProductLibrary is LongShortPairFinanci
...
+                FixedPoint.fromUnscaledUint(0.5).mul(
```
pemulis, line 78: In the generalized version, the `0.5` here will be equal to `1` minus the parameterized `base payout`. If the base payout is 25%, for instance, the number here will be `.75`, which is `1 - 0.25`.

pemulis:
I'm going to update this to fix the build issues.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: Also, the decimal handling is a bit off, but working on a fix.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: Sidenote: Decimal handling needs to be fixed, so `0.5` is really `500000000000000000`, etc.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SmaugBondLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,77 @@
...
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/utils/math/SignedSafeMath.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+import "./LongShortPairFinancialProductLibrary.sol";
+import "../../../../common/implementation/Lockable.sol";
+
+/**
+ * @title Smaug Bond Long Short Pair Financial Product Library
```
smb2796, line 11: Can we name this something less specific and more descriptive than Smaug Bond? Maybe `YieldDollarLongShortPairFinancialProductLibrary`?

pemulis: Sure, I'll go with `YieldDollar` for now unless we have something better, since the embedded put option makes it not quite a yield dollar past the low range.

pemulis: Actually, on second thought, I'll do `YieldDollarMinusPut` to be totally explicit. Can do a global find and replace if we come up with something less clunky.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SmaugBondLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,77 @@
...
+        require(contractLowPriceRange != 0, "Params not set for calling LSP");
+
+        // This function returns a value between 0 and 1e18 to be used in conjunction with the LSP collateralPerPair
+        // that allocates collateral between the short and long tokens on expiry. This can be simplified by considering
+        // the price in two discrete ranges: 1) below the low price range, 2) above the low price range.
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+
+        // For expiry prices below lower bound, return expiryPercentLong = 1 (full position)
+        if (positiveExpiryPrice <= contractLowPriceRange) return FixedPoint.fromUnscaledUint(1).rawValue;
+        // For expiry prices above lower bound
```
smb2796: Might be good to include a payout example here.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SmaugBondLongShortPairFinancialProductLibrary.sol
smb2796: Sounds good

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SushiSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+ * - If the price is less than $20 then the each long is worth 0.5 collateralPerPair and each short is worth 0.5
+ * collateralPerPair.
+ * - If the price is more than $20 then each long is worth 0.5 collateralPerPair plus 0.5 times the fraction of
+ * collateralPerPair that was in the money, and each short is worth the remaining collateralPerPair.
+ * - Say settlement price is $30.  Then expiryPercentLong = 0.5 + (0.5 * (30 - 20) / 30) = 0.6667.
+ * If the collateralPerPair is 2, that means the long payout is 0.6667*2 = 1.3333 $SUSHI, which at a settlement
+ * price of $30 is worth $40. This is equivalent to the value of 1 $SUSHI plus the value of the $20 strike
+ * embedded call.
+ */
+contract SuccessTokenLongShortPairFinancialProductLibrary is LongShortPairFinancialProductLibrary, Lockable {
```
smb2796, line 27: File name should be updated to reflect the contract name

pemulis: good catch

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,87 @@
...
+        require(contractLowPriceRange != 0, "Params not set for calling LSP");
+
+        // This function returns a value between 0 and 1e18 to be used in conjunction with the LSP collateralPerPair
+        // that allocates collateral between the short and long tokens on expiry. This can be simplified by considering
+        // the price in two discrete ranges: 1) below the low price range, 2) above the low price range.
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+
+        // For expiry prices below lower bound, return expiryPercentLong = 1 (full position)
+        if (positiveExpiryPrice <= contractLowPriceRange) return FixedPoint.fromUnscaledUint(1).rawValue;
+        /**
```
chrismaree, line 80: nit: this commenting style is wrong here. use the `//` notation

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,104 @@
...
+
+// Tested Contract
+const CappedYieldDollarLongShortPairFinancialProductLibrary = getContract(
+  "CappedYieldDollarLongShortPairFinancialProductLibrary"
+);
+
+const LongShortPairMock = getContract("LongShortPairMock");
+
+const { toWei, toBN } = web3.utils;
+// const bondNotional = toBN(toWei("100"));
```
chrismaree, line 15: nit: can likely remove this

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,104 @@
...
+      // If the price is above the low price range then the payout is simply that of a yield dollar. i.e every
+      // long token is worth the bond notional of 100. At a price of 20 we are between the bounds. Each long should be worth
+      // 100 so there should be 100/20=5 UMA per long token. As each collateralPerPair is worth 10, expiryPercentLong should
+      // be 10/5=0.5, thereby allocating half to the long and half to the short.
+      const expiryTokensForCollateral1 = await cappedYieldDollarLSPFPL.methods
+        .percentageLongCollateralAtExpiry(toWei("20"))
+        .call({ from: lspMock.options.address });
+      assert.equal(expiryTokensForCollateral1.toString(), toWei("0.5"));
+
+      // Equally, at a price of 40 each long should still be worth 100 so there should be 100/40=2.5 UMA per long. As
```
chrismaree, line 96: this test does not do what your comments say it does. you say to enter 40 as the price and check the value is 0.25 from the lib such that value equals 100 but the test below uses 20 and checks for 0.5

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+ * @title Simple Success Token Long Short Pair Financial Product Library.
+ * @notice Adds settlement logic to create success token LSPs. A success token pays out 50% of collateral as a
+ * floor, with the remaining 50% functioning like an embedded covered call.
+ * If the settlement is below the strike price then longs are worth 50% of collateral.
+ * If the settlement is above the strike then the payout is equal to:
+ * 0.5 + (0.5 * (expiryPrice - strikePrice) / expiryPrice)
+ * For example, consider a covered call option collateralized in SUSHI, with a strike a price of $20,
+ * and collateralPerPair of 2.
+ * - If the price is less than $20 then the each long is worth 0.5 collateralPerPair and each short is worth 0.5
+ * collateralPerPair.
```
chrismaree, line 19:
```suggestion
 * collateralPerPair. i.e each long is worth 1 SUSHI (call expires worthless).
```

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+ * If the collateralPerPair is 2, that means the long payout is 0.6667*2 = 1.3333 $SUSHI, which at a settlement
+ * price of $30 is worth $40. This is equivalent to the value of 1 $SUSHI plus the value of the $20 strike
+ * embedded call.
+ */
+contract SimpleSuccessTokenLongShortPairFinancialProductLibrary is LongShortPairFinancialProductLibrary, Lockable {
+    using FixedPoint for FixedPoint.Unsigned;
+    using SafeMath for uint256;
+
+    mapping(address => uint256) public longShortPairStrikePrices;
+    uint256 basePercentage = 500000000000000000; // 0.5 with 18 decimals
```
chrismaree, line 32: would be pretty trivial to set custom `basePercentage` for each LSP address by modifying the `longShortPairStrikePrices` to be a struct containing both the `strike` and the `basePercentage`. thoughts?

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+        returns (uint256)
+    {
+        uint256 contractStrikePrice = longShortPairStrikePrices[msg.sender];
+        require(contractStrikePrice != 0, "Params not set for calling LSP");
+
+        // If the expiry price is less than the strike price then the long options expire worthless (out of the money).
+        // In this case, return of value of 50% (half of collateral goes to long)
+        // Note we do not consider negative expiry prices in this call option implementation.
+        if (expiryPrice < 0 || uint256(expiryPrice) < contractStrikePrice)
+            return FixedPoint.Unsigned(basePercentage).rawValue;
```
chrismaree, line 71: seeing you stored the base percentage as a uint256 you can simply return `basePercentage` without any type casting.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+        nonReentrantView()
+        returns (uint256)
+    {
+        uint256 contractStrikePrice = longShortPairStrikePrices[msg.sender];
+        require(contractStrikePrice != 0, "Params not set for calling LSP");
+
+        // If the expiry price is less than the strike price then the long options expire worthless (out of the money).
+        // In this case, return of value of 50% (half of collateral goes to long)
+        // Note we do not consider negative expiry prices in this call option implementation.
+        if (expiryPrice < 0 || uint256(expiryPrice) < contractStrikePrice)
```
chrismaree, line 70: this could be problematic if the `expiryPrice` is negative. consider adding in a separate state variable that ensures this is >=0. something like:
```
        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;

```

chrismaree, line 70: also, would you not want the case where it is `<=` the strike price to return the `basePercenage`? In the case that `expiryPrice == contractStrikePrice` then the term below with the computation will return exactly `basePercenage`, no?

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,99 @@
...
+        .setLongShortPairParameters(lspMock.options.address, strikePrice)
+        .send({ from: accounts[0] });
+    });
+    it("Lower than strike should return 0.5", async () => {
+      const expiryTokensForCollateral = await simpleSuccessTokenLSPFPL.methods
+        .percentageLongCollateralAtExpiry(toWei("300"))
+        .call({ from: lspMock.options.address });
+      assert.equal(expiryTokensForCollateral.toString(), toWei("0.5"));
+    });
+    it("Higher than strike correct value", async () => {
```
chrismaree, line 85: a few additional tests and comments might be useful here. I know this is really simple but just to be consistent with the other libraries.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.js
pemulis: Good catch, will fix that comment.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: I agree and plan to do this in the generalized version, in addition to making the variable payout logic a parameter. It's hardcoded in this simple version just for simplicity.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: That makes sense. It would be the same result but more work to go through the variable payout math if the strike price equals the expiry price.

pemulis:
Thanks for the comments, everyone! Will be addressing them over the next couple of days.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.js
pemulis: I found a couple of additional errors in these comments and will fix them, too.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.js
pemulis: I added some more comments to explain what's going on in here. If you think additional tests are needed, let me know.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,85 @@
...
+        lowPriceRanges[longShortPair] = lowPriceRange;
+    }
+
+    /**
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token are
+     * entitled to per collateralPerPair.
+     * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
+     * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
+     */
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
```
nicholaspai, line 63: why is `expiryPrice` allowed to be an `int` but `lowPriceRange` is a `uint`?

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SimpleSuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,91 @@
...
+        longShortPairStrikePrices[longShortPair] = strikePrice;
+    }
+
+    /**
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token are entitled
+     * to per collateralPerPair.
+     * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
+     * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
+     */
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
```
nicholaspai, line 57: Same question as before about `int` vs `uint`

nicholaspai (commented):
Looks great and thanks for adding tests, just one comment about type of `expiryPrice`

pemulis:
> Looks great and thanks for adding tests, just one comment about type of `expiryPrice`

Will address the comments, thanks!

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/CappedYieldDollarLongShortPairFinancialProductLibrary.sol
pemulis: Because it overrides this function in `LongShortPairFinancialProductLibrary`, which expects an `int`.

```
abstract contract LongShortPairFinancialProductLibrary {
    function percentageLongCollateralAtExpiry(int256 expiryPrice) public view virtual returns (uint256);
}
```

nicholaspai (approved):
Thanks for addressing my comments, LGTM. I also checked for potential rounding errors and see that both implementations of `percentageLongCollateralAtExpiry` could introduce only negative error (i.e. the collateral % is lower than expected). Just FYI

PR #3234, opened 2021-07-15
feat: Add cross-chain governance to UMIP collateral and identifier scripts

nicholaspai:
**Motivation**

Admin proposals should be able to contain transactions that whitelist collateral, set final fees, and whitelist identifiers on the Polygon contracts. Requests are sent through the `GovernorRootTunnel`.

This creates a new `admin-proposals` directory within the `core/` package, and all UMIP related scripts should eventually (1) be converted to `hardhat` scripts instead of `truffle` scripts and (2) moved to this folder. We should get rid of `collateral-umip/` and `identifier-umip/` for example.

**Summary**

New `node.js` files are created. Here's an example run where I simulate passing [this UMIP](https://github.com/UMAprotocol/UMIPs/pull/364/files) which contains both Polygon and Ethereum collaterals to whitelist:

1. Run `yarn hardhat node --fork https://mainnet.infura.io/v3/API-KEY --no-deploy --port 9545` in one window. Note that we use `hardhat` forking because the event querying is better than `ganache`, its more consistent with how our repo tooling is migrating to `hardhat`, and we need event querying to verify that Cross-chain governance proposals are sent correctly. This fork runs locally and is available on `--network localhost`.
2. Whitelist 6 new collateral on Ethereum and 1 new Polygon contract: `HARDHAT_NETWORK=localhost node ./packages/core/scripts/admin-proposals/collateral.js --collateral 0x3472a5a71965499acd81997a54bba8d852c6e53d,0x383518188c0c6d7730d91b2c03a03c837814a899,0x875773784af8135ea0ef43b5a374aad105c5d39e,0x6810e776880c02933d47db1b9fc05908e5386b96,0xbEa98c05eEAe2f3bC8c3565Db7551Eb738c8CCAb,,0x0cec1a9154ff802e7934fc916ed7ca50bde6844e --fee 50,1,100,2.4,2000,230,40 --polygon ,,,,,0x580a84c73811e1839f75d86d75d88cca0c241ff4, `

```sh
🚸 Connected to a local node, attempting to impersonate accounts on forked network 🚸
┌──────────┬──────────────────────────────────────────────┐
│ (index)  │                    Values                    │
├──────────┼──────────────────────────────────────────────┤
│ deployer │ '0x2bAaA41d155ad8a4126184950B31F50A1513cE25' │
└──────────┴──────────────────────────────────────────────┘
⛽️ Current fast gas price for Ethereum: 23 gwei

ℹ️  Relayer infrastructure for Polygon transactions:
[...]

PR #3243, opened 2021-07-20
feat(core): Add register-contract and add-contract-creator admin proposal scripts

nicholaspai:
**Motivation**

Script design modeled after #3234 to enable cross-chain governance.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/core/scripts/admin-proposals/setup.sh
```diff
@@ -0,0 +1,6 @@
+#!/usr/bin/env bash
```
mrice32, line 1: nit: maybe we should call this and setupFork something more specific to impersonating accounts? Otherwise, it's unclear what this script is and in what cases it needs to be run.

packages/core/scripts/admin-proposals/registerContract.js
```diff
@@ -0,0 +1,306 @@
...
+const assert = require("assert");
+const { GasEstimator } = require("@uma/financial-templates-lib");
+const Web3 = require("web3");
+const winston = require("winston");
+const { RegistryRolesEnum, interfaceName } = require("@uma/common");
+const { _getContractAddressByName, _setupWeb3 } = require("./utils");
+const { REQUIRED_SIGNER_ADDRESSES } = require("./constants");
+const argv = require("minimist")(process.argv.slice(), {
+  string: [
+    // address to register on Ethereum
```
mrice32, line 42: Is this required if polygon is omitted?

mrice32 (approved):
LGTM!

packages/core/scripts/admin-proposals/registerContract.js
nicholaspai: yes will add a comment.

PR #3244, opened 2021-07-21
fix(api,sdk): allow build to run with node by fixing imports

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1775/bug-replace-ts-node-with-node-when-running-api

**Summary**
This fixes some sdk imports which caused api to crash when running the built version. 

**Details**
Previously we could only run the api with ts-node. This fixes the runtime errors so that it runs from build. The hope is we can remove another possible cause of the API hitting 100% cpu usage intermittently.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1775/bug-replace-ts-node-with-node-when-running-api

### DIFF
--- a/packages/sdk/src/libs/clients/emp/client.ts
+++ b/packages/sdk/src/libs/clients/emp/client.ts
@@ -1,11 +1,13 @@
-import { ExpiringMultiParty__factory, ExpiringMultiParty } from "@uma/core/contract-types/ethers";
+import { EthersContracts } from "@uma/core";
 import type { SignerOrProvider, GetEventType } from "../..";
 import { Event } from "ethers";
 import { Balances } from "../../utils";
 
-export type Instance = ExpiringMultiParty;
+export type Instance = EthersContracts.ExpiringMultiParty;
+const Factory = EthersContracts.ExpiringMultiParty__factory;
+
 export function connect(address: string, provider: SignerOrProvider): Instance {
-  return ExpiringMultiParty__factory.connect(address, provider);
+  return Factory.connect(address, provider);
 }
 
 export interface EventState {
--- a/packages/sdk/src/libs/index.ts
+++ b/packages/sdk/src/libs/index.ts
@@ -6,7 +6,7 @@ export { default as Coingecko } from "./coingecko";
 export { default as Multicall } from "./multicall";
 
 // types
-import type { TypedEventFilter, TypedEvent } from "@uma/core/contract-types/ethers/commons";
+import type { TypedEventFilter, TypedEvent } from "@uma/core/types/contract-types/ethers/commons";
 import { Contract } from "ethers";
 import { Result } from "@ethersproject/abi";
 import { Signer } from "@ethersproject/abstract-signer";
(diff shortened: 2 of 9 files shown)


packages/sdk/src/libs/clients/emp/client.ts
```diff
@@ -1,11 +1,13 @@
-import { ExpiringMultiParty__factory, ExpiringMultiParty } from "@uma/core/contract-types/ethers";
+import { EthersContracts } from "@uma/core";
```
daywiss, line 1: this was the main issue. i was importing from the wrong directory

packages/sdk/src/libs/index.ts
```diff
@@ -6,7 +6,7 @@ export { default as Coingecko } from "./coingecko";
 export { default as Multicall } from "./multicall";
 
 // types
-import type { TypedEventFilter, TypedEvent } from "@uma/core/contract-types/ethers/commons";
+import type { TypedEventFilter, TypedEvent } from "@uma/core/types/contract-types/ethers/commons";
```
daywiss, line 9: this was also the wrong directory. took me a while to realize this

packages/sdk/src/libs/clients/emp/client.ts
```diff
@@ -1,11 +1,13 @@
...
 import type { SignerOrProvider, GetEventType } from "../..";
 import { Event } from "ethers";
 import { Balances } from "../../utils";
 
-export type Instance = ExpiringMultiParty;
+export type Instance = EthersContracts.ExpiringMultiParty;
+const Factory = EthersContracts.ExpiringMultiParty__factory;
```
daywiss, line 7: @mrice32 i know you wont like this 😀 but having common naming between all client files makes it easy to 1. template new clients and 2. refactor across all clients, like with this pr

mrice32: haha, the reason I don't like it is that it seems like it's designed to make copy and pasting easier, but then if the code is exactly the same, why can't it just be a generic?

Either way, I'm not strongly opinionated :)

Side note: are these cosmetic changes related to the compilation issues? Or are they just purely cosmetic?

packages/sdk/src/libs/index.ts
mrice32: Yeah, these import paths in core are very odd. The newer packages will fix these oddities. Sorry for the weirdness. Didn't realize using types like this would cause such strange paths, as I misunderstood typescript resolution.

mrice32 (approved):
LGTM!

packages/sdk/src/libs/clients/emp/client.ts
daywiss: thats a good point, i didnt think of that. you might be on to something

daywiss: also these are mainly cosmetic changes. the fix really was just with changing the imports

PR #3245, opened 2021-07-21
feat(api): expose various lsp state calls and refactor api slightly

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1714/expose-lsp-state-through-api

**Summary**

This exposes LSP state queried through ingestion services to the express api.

**Details**
There was some refactoring done to how the API works, but it should remain backward compatible. Specific contract calls to the api are now separated by "channels", so the path prepends `/emp/` or `/lsp/` to the action for instance `localhost:8282/emp/actions` - to list all actions on the channel.  The root channel `/` and `/emp` are identical for backward compatibility to existing clients, but this is deprecated and will probably change in the future.

Several new API calls were added specifically for the LSP which mirror many of the EMP calls. The README summarizes these.

There is a new [postman link ](https://www.getpostman.com/collections/9ce0fe57a483b5163fbf) which can be imported with the updated calls.

The action-client example has also been updated to show how to use the new channel system.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/1714/expose-lsp-state-through-api

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -63,128 +63,170 @@ This is a non exhaustive list of actions, for the full list see [action source](
 The following actions can be called to the server with `curl -X POST localhost:8282/${action}`.
 Parameters are attached as JSON to the request body in the follow format: `"[...arguments]"`. Arguments should be attached to the body as an array.
 
-### actions() => string[]
+## Channels
+
+Channels are a way to group actions into related sets. Currently channels are used to separate queries
+across contract types, but they can theoretically group any action calls together. To call an action on
+a channel, use the channel name as part of the path. `curl -X POST localhost:8282/${channel}/${action}`.
+Current channels are
+
+- emp - all emp calls
+- lsp - all lsp calls
+
+## Action Client
+
+See [../../examples/action-client.ts](Action Client Code) for an example of how to interface with the api from the client side.
+
+## Emp Api
+
+All calls here are on the emp channel.
+
+### emp/actions() => string[]
 
 Returns a list of all actions
 
-### echo(...args:Json[]) => args
+### emp/echo(...args:Json[]) => args
 
 Echos back any parameters you send to the server. For testing connectivity and client correctness.
 
-### listEmpAddresses() => string[]
+### emp/listEmpAddresses() => string[]
 
 Returns all known emp addresses for this servers network, which defaults to mainnet.
 
-### lastBlock() => number
+### emp/lastBlock() => number
 
 Returns the last block number seen.
 
-### listActiveEmps() => Emps[]
+### emp/listActiveEmps() => Emps[]
 
 Returns all known active emps data. Conforms to the EMP type defined in the uma sdk: `uma.tables.emp.Data`.
 
-### listExpiredEmps() => Emps[]
+### emp/listExpiredEmps() => Emps[]
 
 Returns all known expired emps data. Conforms to the EMP type defined in the uma sdk: `uma.tables.emp.Data`.
 
-### getEmpState(address: string) => EmpState
+### emp/getEmpState(address: string) => EmpState
 
 Get all known state of any emp by its address.
 
-### getErc20Info(address:string) => Erc20Data
+### emp/getErc20Info(address:string) => Erc20Data
 
 Get name, decimals of an erc20 by address.
 
-### getAllErc20Info() => Erc20Data[]
+### emp/getAllErc20Info() => Erc20Data[]
 
 List all known erc20s, collateral or synthetic and all known information about them
 
-### collateralAddresses() => string[]
+### emp/collateralAddresses() => string[]
 
-### syntheticAddresses() => string[]
+### emp/syntheticAddresses() => string[]
 
 Returns all known active collateral or synthetic addresses
 
-### allLatestPrices(currency='usd') => {[address:string]:[timestamp:number,price:string]
+### emp/allLatestPrices(currency='usd') => {[address:string]:[timestamp:number,price:string]
 
 Returns all known latest prices for synthetic or collateral addresses in wei.
 
-### allLatestMarketPrices() => {[address:string]:[timestamp:number,price:string]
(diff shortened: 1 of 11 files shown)


packages/api/src/apps/api/README.md
```diff
@@ -63,128 +63,170 @@ This is a non exhaustive list of actions, for the full list see [action source](
...
 
 Returns a list of all actions
 
-### echo(...args:Json[]) => args
+### emp/echo(...args:Json[]) => args
 
 Echos back any parameters you send to the server. For testing connectivity and client correctness.
 
-### listEmpAddresses() => string[]
+### emp/listEmpAddresses() => string[]
```
mrice32, line 92: Does this mean that the user has to figure out which type of contract a particular address is to get information on it? Is there a reason why we can't just offer all calls for any address and have an error message if that call/action isn't available for that action?

While this may be more elegant from a design/implementation perspective, it seems harder for the user or frontend to navigate because there are 1) more functions and 2) they need to have a deeper understanding of the contract that they're querying to actually make a query.

I think an illustrative example is etherscan -- a user shows up with an address and etherscan gives them curated information based on what it knows about that address without the user needing to request that information specifically.

Also, is there a way to ask the API for the type of a contract or is the expectation that the frontend figure that out?

mrice32 (commented):
Looks awesome! One high-level comment about the separation of the emp api and the lsp api. We might want to have a discussion offline about this.

packages/api/src/apps/api/README.md
daywiss: yea i totally get this comment. currently yes, it requires specific knowledge to call the right endpoint.  this is just a first step though towards the feature you are describing. I plan to use the root directory for "global" style calls which would include searching for contracts across types, as well as any kind of stats which combine these or other types together, like tvl. 

I didnt want to do that all in this pr since i wanted to try and maintain backward compatibility for now (root path points to the same calls as previous version), and it makes sense as a separate feature from this.

mrice32: If this PR is only to add the capability to build calls from root that can be on any contract type, then why do we need to expose these channels to the user at all?

mrice32 (commented):
Annoying that github doesn't let you just leave a single comment as a review 🙃. https://github.com/UMAprotocol/protocol/pull/3245#discussion_r674743236

packages/api/src/apps/api/README.md
daywiss: hmm good question. This is a higher level design discussion.  I ideally want to unblock FE work and defer this design decision.  This PR will allow FE to start integrating under the assumption they first know if an address is an LSP or EMP.  The only thing needed later in the global scope would be some calls to inform what type of contract a particular address is  and pass the address to the correct contract template. 

There's an alternative design i think, which is maybe what you are referring to, where the global scope contains all the calls for each contract type, and automatically switches based on the address.  Its not my favorite design, but will definitely take input from FE based on what they would prefer.  I think this decision can be deferred somewhat because this will at least  this will unblock some FE development and does not rule out either design.

mrice32 (approved):
As discussed offline, we may want to spec these api endpoints out ahead of time, but the code LGTM!

packages/api/src/services/express-channels.ts
```diff
@@ -0,0 +1,67 @@
...
+import { Json, BaseConfig } from "..";
+import { Profile } from "../libs/utils";
+
+interface Config extends BaseConfig {
+  port: number;
+  path?: string;
+}
+// represents the actions service, a single function which maps to different callable functions
+type Actions = (action: string, ...args: Json[]) => Promise<Json>;
+type Channel = [string, Actions];
```
chrismaree, line 16: this is cool. I like this ideas for building up dynamic paths

packages/api/src/apps/api/README.md
chrismaree: I agree that in an ideal world we don't have levels like this. a user should be able to query any address, just like on etherscan, and from this find if it is: a) an LSP, b) a EMP c) an LSP long or short token, d) and EMP synthetic f) even an LSP FNP! there should be no differentiation at the query level. 

but, from a design perspective, it does make sense to separate these into separate logical containers to keep things somewhat modular. I guess it is fine to defer this design discussion until all the functionality is built out, assuming you feel confident that you can easily pull all these end points together through some aggregator end point?

Unblocking the FE is good but not if it means incurring technical/design debt.

chrismaree (approved):

packages/api/src/apps/api/README.md
daywiss: > but, from a design perspective, it does make sense to separate these into separate logical containers to keep things somewhat modular. I guess it is fine to defer this design discussion until all the functionality is built out, assuming you feel confident that you can easily pull all these end points together through some aggregator end point?

thanks for discussion @chrismaree , also talked with @mrice32 a bit more about this offline. I think there is some implicit knowledge on my end which ends up making this particular design a bit suspect to others, including previous discussions around wanting to be able to query across contracts. Heres why I think this is appropriate, even if its eventually deprecated in favor of a unified api:

- discussed this design with @Gamaranto 1 on 1 and got his go ahead with this as a first pass, though did not document this publicly and probably should have
- from our discussion he also assured me that its also very easy to adapt his api client to changes in the API, which in my mind means that exposing standalone LSP api now is better than waiting until the more complex combined api, so the LSP logic can be vetted sooner (this is a big assumption and should probably run this again by francesco as it may not be true)
- moving contracts types to specific channels opens up the root channel for unified queries across all contracts, which was a big motivator for this change
- api calls contain little to no business logic, meaning from a development perspective, are almost free, the underlying business logic will not change to make a unified api
- unified api requires more logic in the api layer to allow switching between contract types, which i feel adds another new feature on an already new unvetted feature (LSPs) and makes sense as a separate PR.

Long story short though, the unified API is still on the table and will be discussing this in next infra meeting. The extra work to add contract specific api routes are minor (vs just not exposing them at all), so i see it as adding some flexibility for very little cost.

PR #3247, opened 2021-07-21
feat(periphery): Create new package for scripts that interact with core contracts

nicholaspai:
**Motivation**

To simplify `core`, this PR introduces a new package "periphery" that will house all scripts that designed to be run as one-off interactions with deployed core contracts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/periphery/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@uma/periphery",
+  "version": "1.0.0",
+  "description": "UMA Core Peripheral Scripts",
+  "dependencies": {
```
nicholaspai, line 5: @mrice32 @chrismaree should any or all of these be `devDeps`? I'm not really sure how to think about it. Also, maybe none of these even need to be deps if they are imported by the root `package.json`

packages/periphery/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+```sh
+./packages/periphery/setupFork.sh
+```
+
+3. Run your script against the local mainnet fork
+
+```sh
+node YOUR_FILE.js --network mainnet-fork
+```
```
chrismaree, line 41: very nice readme

packages/periphery/package.json
chrismaree: I think it makes sense to have these as `dependencies`. if there were any tests in this package, that had their own packages to facilitate the tests, then they would be dev dependencies.

packages/periphery/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@uma/periphery",
```
chrismaree, line 2: just wanted to ask about the name of this package. `periphery` does make sense but `admin-scripts` or `governance-scripts` might be more descriptive? not too convinced the ones I proposed are better but just wanted to discuss it.

nicholaspai: I was thinking this package would include other helper scripts like `AdvanceVotePhase` and `CreatePriceRequest` which are both better described as `voting-scripts`. WDYT?

chrismaree: `voting-scripts` is better imo than `periphery`. it's certainly more descriptive to what the packing contains.

nicholaspai: but not all of these will be voting scripts, some will be interacting with the LSP/financial contracts right? Like there might be one that creates a sponsor position. Do you think there should be a separate package for financial-contract-scripts?

I'm just imagining that every single script that interacts with any contract in `core` would go into `periphery`. And they're all closely related in that they need a similar `hardhat` setup and even a connection to a mainnet fork.

packages/periphery/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+The caller might find it useful to run these scripts against a local (hardhat) [Mainnet fork](https://hardhat.org/guides/mainnet-forking.html) in order to test that the contract state is modified as expected. The local node should be set to run on port `9545`.
+
+Scripts can then be run as `node` scripts connected to the local node by passing in the `--network mainnet-fork` flag.
+
+After the developer has tested against the local Mainnet fork, they can use the same script connected to a public network. Usually this would involve passing a network name like `mainnet_mnemonic` or `mainnet_gckms` to the `--network` flag to the script and other supporting flags like `--keys`.
+
+## How to run
+
+All scripts are designed to be framework-agnostic (i.e. not wed to `hardhat` or `truffle`) and run as `node` scripts.
```
mrice32, line 15: +1 this is great!

packages/periphery/package.json
mrice32: Is this package expected to be installed via npm at all? If not, we could set it to private so it doesn't get published.

I agree with @chrismaree. If the scripts are considered "production" code for this package, then their deps should be deps, and not devdeps.

packages/periphery/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "winston": "^3.2.1",
+    "web3": "^1.3.5"
+  },
+  "devDependencies": {},
+  "homepage": "https://umaproject.org",
+  "license": "AGPL-3.0-or-later",
+  "publishConfig": {
+    "registry": "https://registry.npmjs.com/",
+    "access": "public"
+  },
```
mrice32, line 22: If you want to set the package to private (so it isn't published to npm), you can replace this field with the following:
```
  "private": true,
```

mrice32 (approved):
LGTM!

packages/periphery/package.json
nicholaspai: hmm i could see it being published to npm because some of the scripts or utils might be helpful elsewhere. Are there downsides to publishing an additional package?

packages/periphery/package.json
chrismaree: ye see your point. how about `periphery-scripts` :)? or just `scripts`!

chrismaree (approved):
looks good! just want to decide on if we'd like to change this package name

PR #3249, opened 2021-07-22
feat(Insured-Bridge): initial contract interfaces & smocking

chrismaree:
**Motivation**

As part of the initial design and implementation of the insured bridge, we need to define the initial contract interfaces and smocking setup for testing.

**Summary**

This PR adds the following:
- Initial interfaces, as discussed in internal design process, for:
	- L1 Withdrawal contract
	- L2 deposit contract
- Initial tests for L2 showing how to use [smock](https://github.com/ethereum-optimism/optimism/blob/develop/packages/smock/README.md) to mock L1<->L2 communication in testing to enable isolated network testing within the HRE.

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/common/src/HardhatConfig.js
```diff
@@ -42,7 +46,7 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
         "contracts/oracle/implementation/test/VotingTest.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
       },
     },
-    ovm: { solcVersion: "0.8.4-broken_alpha" },
+    ovm: { solcVersion: "0.7.6" },
```
chrismaree, line 49: I change this to let us use the OVM contracts, such as the imports like `import "@eth-optimism/...`

packages/core/contracts/insured-bridge/implementation/L1_BridgePool.sol
```diff
@@ -0,0 +1,4 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.7.6;
+
+contract BridgePool {}
```
chrismaree, line 4: I created this stub as it's likely the L1 contracts will have two parts:
1) a router that all users interact with
2) a pool that contains the funds for a particular bridging action.

packages/core/test/insured-bridge/L2_DepositBox.js
```diff
@@ -0,0 +1,52 @@
+const hre = require("hardhat");
```
chrismaree, line 1: this initial test file shows how: a) we can continue to write tests in web3 b) we can _also_ use smockit tests to mock L1<->L2 communication.

packages/core/test/insured-bridge/L2_DepositBox.js
```diff
@@ -0,0 +1,52 @@
...
+  describe("Box Ownership logic", () => {
+    it("Transfer ownership", async () => {
+      // Owner should start out as the set owner.
+      assert.equal(await depositBox.methods.l1Owner().call(), l1Owner);
+
+      // Trying to transfer ownership from non-cross-domain owner should fail.
+      assert(await didContractThrow(depositBox.methods.transferL1Owner(user1).send({ from: rando })));
+
+      // Trying to call correctly via the L2 message impersonator, but from the wrong xDomainMessageSender should revert.
+      l2CrossDomainMessengerMock.smocked.xDomainMessageSender.will.return.with(() => rando);
```
chrismaree, line 40: this is the only smocked logic done here where we are simply setting the `xDomainMessageSender` will return a particular value. this could easily be done with a solidity mock.

In future, we can check that specific calls `l2CrossDomainMessengerMock.smocked.sendMessage.calls[0]` contain things like `_target` and `_message` to see that specific calls happened to the bridge. this would _not_ be possible with a solidity mock.

packages/common/package.json
```diff
@@ -51,13 +51,15 @@
...
-    "web3": "^1.3.5"
+    "web3": "^1.3.5",
+    "ethers": "^5.1.3"
   },
   "devDependencies": {
     "@eth-optimism/hardhat-ovm": "^0.2.2",
     "@nomiclabs/hardhat-etherscan": "^2.1.4",
     "@nomiclabs/hardhat-truffle5": "^2.0.0",
     "@nomiclabs/hardhat-web3": "^2.0.0",
+    "@nomiclabs/hardhat-ethers": "^2.0.2",
```
nicholaspai, line 62: Since we import `hardhat-deploy`, we should install `@nomiclabs/hardhat-ethers` the [hardhat-deploy recommended way](https://github.com/wighawag/hardhat-deploy#npm-install-hardhat-deploy). This arises from the fact that `hardhat-deploy` forks `hardhat-ethers` to add some extra features to the `deploy` task.

packages/common/src/HardhatConfig.js
```diff
@@ -30,7 +31,10 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
 
   const defaultConfig = {
     solidity: {
-      compilers: [{ version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 } } }],
+      compilers: [
+        { version: solcVersion, settings: { optimizer: { enabled: true, runs: 1000000 } } },
+        { version: "0.7.6", settings: { optimizer: { enabled: true, runs: 1000000 } } },
```
nicholaspai, line 36: Do you need to set 0.7.6 in `compilers`? I thought you only had to set it in the `ovm` value. [docs](https://community.optimism.io/docs/resources/hardhat.html#migrate-the-sample-app-to-optimistic-ethereum)

packages/common/src/HardhatConfig.js
```diff
@@ -62,7 +66,7 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
         ovm: true,
         // We use custom logic to only compile contracts within the listed directories, as opposed to choosing which
         // ones to ignore, because there are more contracts to ignore than to include.
-        compileWhitelist: ["oracle/implementation/Finder.sol"],
+        compileWhitelist: ["oracle/implementation/Finder.sol", "insured-bridge/implementation/L2_BridgeDepositBox.sol"],
```
nicholaspai, line 69: Should we remove `Finder` from the whitelist now since the Finder contract require 0.8?

packages/core/contracts/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,69 @@
...
+import "@eth-optimism/contracts/libraries/bridge/OVM_CrossDomainEnabled.sol";
+
+import "./L2_BridgeDepositBox.sol";
+
+contract BridgeRouter is OVM_CrossDomainEnabled {
+    address public l1Owner;
+    address public l2DepositBox;
+
+    event AddedL2DepositContract(address l2Contract);
+    event AddedL2DepositContract(address originToken, address destinationToken);
```
nicholaspai, line 13: Is this event misnamed? Same as above

packages/core/contracts/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,69 @@
...
+        address recipient,
+        address originToken,
+        address destinationToken,
+        address relayer,
+        uint256 amount,
+        uint256 realizedFee,
+        uint256 maxFee
+    );
+    event LPDepositCollateral(address token, uint256 amount, uint256 lpTokensMinted, address caller);
+    event L2TransferSpedUp(uint256 transactionId, address fastRelayer);
```
nicholaspai, line 25:
```suggestion
    event L2TransferSpeedUp(uint256 transactionId, address fastRelayer);
```

nit: event name IMO should be SpeedUp instead of SpedUp

packages/core/package.json
```diff
@@ -19,6 +20,8 @@
     "decimal.js": "^10.2.1"
   },
   "devDependencies": {
+    "@eth-optimism/smock": "^1.1.9",
+    "@nomiclabs/hardhat-ethers": "^2.0.2",
```
nicholaspai, line 24: See my prior note about `hardhat-ethers`, but also: do we need to import `hardhat-ethers` here if its imported by common and injected into the `HardhatConfig.js` file?

packages/common/src/HardhatConfig.js
chrismaree: yup _but_ when writing L2 tests (not e2e tests) we will run them in the normal HRE so we need our normal hardhat compiler to _also_ support this version to run in the HRE.

packages/common/src/HardhatConfig.js
chrismaree: ye, can do.

packages/core/test/insured-bridge/L2_DepositBox.js
nicholaspai: This is pretty awesome and makes this impersonation very succinct

packages/core/test/insured-bridge/helpers/SmockitHelper.js
```diff
@@ -0,0 +1,18 @@
+const { getContractDefinition } = require("@eth-optimism/contracts");
```
nicholaspai, line 1: Random question: why isn't there a github repo linked to this npm package? All the contract file paths such as `@eth-optimism/contracts/libraries/bridge/OVM_CrossDomainEnabled.sol` don't exist in what I thought was the optimism [contracts repo](https://github.com/ethereum-optimism/optimism/tree/develop/packages/contracts)

nicholaspai (approved):
Smockit is awesome, really good job setting up the scaffolding. I have only some nits and one question.

packages/common/package.json
chrismaree: done. I removed the previous dependency then ran:

```
yarn workspace @uma/common add @nomiclabs/hardhat-ethers@npm:hardhat-deploy-ethers ethers --dev
```

and updated the import.

packages/core/test/insured-bridge/helpers/SmockitHelper.js
chrismaree: no idea 🤷

packages/core/package.json
chrismaree: good point. will test to double check.

packages/core/contracts/insured-bridge/implementation/L1_BridgeRouter.sol
chrismaree: good catch. was meant to be `WhitelistedToken`

packages/common/src/HardhatConfig.js
nicholaspai: aha

packages/common/src/HardhatConfig.js
```diff
@@ -5,10 +5,12 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
...
   // get access to the plugins as well.
   if (includeTruffle) require("@nomiclabs/hardhat-truffle5");
-  require("hardhat-gas-reporter");
   require("@nomiclabs/hardhat-web3");
-  require("hardhat-deploy");
   require("@nomiclabs/hardhat-etherscan");
+  require("@nomiclabs/hardhat-ethers");
+  require("hardhat-deploy");
+  require("hardhat-gas-reporter");
+  require("hardhat-deploy");
```
nicholaspai, line 13: duplicate import

chrismaree: fixed.

PR #3251, opened 2021-07-22
feat(api): 1490 add lsp tvl calculations, per contract, totals and history

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1490/ability-to-pull-in-tvl-for-each-lsp

**Summary**

This refactors stats and queries to accommodate LSP TVL stats as well as adds some new API calls on the LSP channel. 

**Details**

This adds several TVL calls, including historical and TVL backfill for LSP. Readme is updated with calls.

With slightly more refactoring, LSP and EMP stats could potentially share a lot of the same code, but I was afraid this would be too many unecessary changes, so instead business logic is copied and scoped to a specific contract type. 

Future work will need to combine both EMP and LSP TVL and expose in global, or unified stats

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1490/ability-to-pull-in-tvl-for-each-lsp

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -65,14 +65,36 @@ export default async (env: ProcessEnv) => {
     },
     erc20s: tables.erc20s.JsMap(),
     stats: {
-      usd: {
-        latest: {
-          tvm: empStats.JsMap("Latest Tvm"),
-          tvl: empStats.JsMap("Latest Tvl"),
+      emp: {
+        usd: {
+          latest: {
+            tvm: empStats.JsMap("Latest Tvm"),
+            tvl: empStats.JsMap("Latest Tvl"),
+          },
+          history: {
+            tvm: empStatsHistory.SortedJsMap("Tvm History"),
+            tvl: empStatsHistory.SortedJsMap("Tvl History"),
+          },
         },
-        history: {
-          tvm: empStatsHistory.SortedJsMap("Tvm History"),
-          tvl: empStatsHistory.SortedJsMap("Tvl History"),
+      },
+      lsp: {
+        usd: {
+          latest: {
+            tvl: empStats.JsMap("Latest Tvl"),
+          },
+          history: {
+            tvl: empStatsHistory.SortedJsMap("Tvl History"),
+          },
+        },
+      },
+      global: {
+        usd: {
+          latest: {
+            tvl: empStats.JsMap("Latest Tvl"),
+          },
+          history: {
+            tvl: empStatsHistory.SortedJsMap("Tvl History"),
+          },
         },
       },
     },
@@ -110,10 +132,11 @@ export default async (env: ProcessEnv) => {
       appState
     ),
     erc20s: Services.Erc20s({ debug }, appState),
-    empStats: Services.EmpStats({ debug }, appState),
+    empStats: Services.stats.Emp({ debug }, appState),
     marketPrices: Services.MarketPrices({ debug }, appState),
     lspCreator: Services.MultiLspCreator({ debug, addresses: lspCreatorAddresses }, appState),
     lsps: Services.LspState({ debug }, appState),
+    lspStats: Services.stats.Lsp({ debug }, appState),
   };
 
   // warm caches
@@ -139,6 +162,9 @@ export default async (env: ProcessEnv) => {
     console.log("Updated Collateral Prices Backfill");
     await services.empStats.backfill();
     console.log("Updated EMP Backfill");
+
+    await services.lspStats.backfill();
+    console.log("Updated LSP Backfill");
   }
 
   await services.collateralPrices.update();
@@ -149,6 +175,9 @@ export default async (env: ProcessEnv) => {
 
   await services.empStats.update();
   console.log("Updated EMP Stats");
+
+  await services.lspStats.update();
+  console.log("Updated LSP Stats");
 
   await services.marketPrices.update();
   console.log("Updated Market Prices");
@@ -196,6 +225,7 @@ export default async (env: ProcessEnv) => {
     await services.syntheticPrices.update();
     await services.marketPrices.update();
     await services.empStats.update();
+    await services.lspStats.update();
   }
 
   // coingeckos prices don't update very fast, so set it on an interval every few minutes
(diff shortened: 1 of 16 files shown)


packages/api/src/apps/api/app.ts
```diff
@@ -64,14 +64,36 @@ export default async (env: ProcessEnv) => {
     },
     erc20s: tables.erc20s.JsMap(),
     stats: {
-      usd: {
-        latest: {
-          tvm: empStats.JsMap("Latest Tvm"),
-          tvl: empStats.JsMap("Latest Tvl"),
+      emp: {
```
daywiss, line 67: stats information is broken out per contract type, with global state meant to be the combination of each type. global state will come in future PR though its stubbed out here.

packages/api/src/services/actions/emp.ts
```diff
@@ -118,35 +118,35 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       return queries.sumTvm(addresses, currency);
     },
     async tvlHistoryBetween(empAddress: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
-      assert(stats[currency], "Invalid currency type: " + currency);
-      return stats[currency].history.tvl.betweenByAddress(empAddress, start, end);
+      assert(stats.emp[currency], "Invalid currency type: " + currency);
```
daywiss, line 121: no logic has changed, just accessor paths changed in this file

packages/api/src/services/actions/lsp.ts
```diff
@@ -36,6 +38,35 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     async shortAddresses() {
       return Array.from(shortAddresses.values());
     },
+    async listTvls(currency: CurrencySymbol = "usd") {
```
daywiss, line 41: these are all the new tvl calls for lsp only. Copied over from the EMP calls for the most part.  These calls are only scoped to lsps.  future work will need to combine emp and lsp tvl

packages/api/src/services/lsp-state.e2e.ts
```diff
@@ -40,11 +40,25 @@ describe("lsp-state service", function () {
...
     assert.ok(service);
   });
+  it("get collateral balance", async function () {
+    const collateralAddress = "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828";
+    const service = Service({}, appState);
+    const [address] = registeredContracts;
+    const result = await service.utils.getErc20BalanceOf(collateralAddress, address);
+    assert.ok(result);
+  });
+  it("getPositionCollateral", async function () {
```
daywiss, line 50: this was needed to be added to lsp state in order to calculate TVL. explicitly calls erc20 for balance

packages/api/src/services/stats/emp.ts
```diff
@@ -1,9 +1,9 @@
 import assert from "assert";
 import * as uma from "@uma/sdk";
 import { BigNumber } from "ethers";
-import { Currencies, AppState, PriceSample, BaseConfig } from "..";
-import { calcTvl, calcTvm, nowS } from "../libs/utils";
-import Queries from "../libs/queries";
+import { Currencies, AppState, PriceSample, BaseConfig } from "../..";
+import { calcTvl, calcTvm, nowS } from "../../libs/utils";
+import * as Queries from "../../libs/queries";
```
daywiss, line 6: again no logic really changed, just the pathing to various tables

packages/api/src/services/stats/lsp.ts
```diff
@@ -0,0 +1,165 @@
+import assert from "assert";
```
daywiss, line 1: this calculates the stats (tvl) for LSP.  This is almost verbatim from the EMP stats service with unnecessary calls removed. Theres so much code duplication and this is a good candidate to refactor to try and combine the logic for both contract types. Requires some more time and thought to do this, and probably would need to break out each stat type into its own file (tvl, tvm)

packages/api/src/libs/queries/lsp.ts
```diff
@@ -0,0 +1,85 @@
+import assert from "assert";
```
daywiss, line 1: this is a little utility library which is mostly copied from queries/emp but tailored to the LSP. mainly just joins state across tables, and sums tvls for now.  used as a utility in the actions (api) layer and some services

packages/api/src/services/lsp-state.e2e.ts
chrismaree: ye, somewhat different than the EMP for this kind of call

packages/api/src/services/lsp-state.ts
```diff
@@ -43,6 +44,28 @@ export default (config: Config, appState: Dependencies) => {
...
+    return batchRead(dynamicProps, instance, address);
+  }
+  // this is not ideal, but contract does not have a way to query this directly. this is a dyanmic value which
+  // depends first on gettin the static value 'collateralToken'.
+  async function getPositionCollateral(instance: Instance, address: string) {
+    const collateralToken = await instance.collateralToken();
+    return getErc20BalanceOf(collateralToken, address);
+  }
+  // get token state based on contract
+  async function getErc20BalanceOf(erc20Address: string, userAddress: string) {
```
chrismaree, line 61: this does not consider the decimals of the collateral. is this something to add?

packages/api/src/services/stats/lsp.ts
chrismaree: ye, for sure would be ideal to join these somehow in the future

chrismaree (approved):

packages/api/src/services/lsp-state.ts
daywiss: good point let me look into this

daywiss: ok this is probably not ideal, but its not broken. currently the only place this is used is to calculate TVL. and that calculation does take into account collateral decimals when its calculated in stats. This might be confusing if someone wanted to read the value from the emp state but they will have access to collateralDecimals on the emp state when they query.

Maybe ideally everything gets stored in 18 dec but not sure if that change is worth it just yet

mrice32 (approved):
LGTM! Agree with you that we need to figure out a way to abstract away the parts that are going to be the same across these two (and potentially other) contracts.

PR #3255, opened 2021-07-27
feat(insured-bridge): end to end testing stub

chrismaree:
**Motivation**

As part of building out the insured bridge, we require end-to-end tests to validate the full lifecycle of tokens moving from L1 to L2. 

This PR adds a basic setup for e2e using optimism's canonical token bridge. this tests out the life cycle from sending a token from L1->L2, checking it is minted correctly and then sending it back to L1.

**Summary**

Adds e2e tests, helpers and tools to be used later down the line to write UMA e2e tests for optimism.

to run these tests you'll need to use mocha. Run the following:
```
mocha --config ./test-e2e/insured-bridge/.mocharc-e2e.js
```

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/test-e2e/insured-bridge/e2e_InsuredBridge.js
```diff
@@ -0,0 +1,118 @@
...
+  createLocalEthersFactory,
+  createOptimismEthersFactory,
+  getProviders,
+} = require("./helpers/MultiProviderHelper");
+
+const { DEFAULT_ADMIN_KEY } = require("./helpers/OptimismConstants");
+
+// Create contract factories using L1 and L2 artifacts
+const factory__L1_ERC20 = createLocalEthersFactory("ExpandedERC20");
+const factory__L2_ERC20 = createOptimismEthersFactory("L2StandardERC20", true);
```
nicholaspai, line 19: OOC, why couldn't you use the same ExpandedERC20 ABI on L2?

packages/core/test-e2e/insured-bridge/e2e_InsuredBridge.js
```diff
@@ -0,0 +1,118 @@
...
+      assert.equal((await L1_ERC20.balanceOf(l1Wallet.address)).toString(), "0");
+      assert.equal((await L2_ERC20.balanceOf(l2Wallet.address)).toString(), "1234");
+    });
+    it("Can send tokens back from L2 to L1", async () => {
+      // Burn the tokens on L2 and ask the L1 contract to unlock on our behalf.
+      const withdrawTx = await L2StandardBridge.withdraw(L2_ERC20.address, 1234, 0, "0x");
+      await withdrawTx.wait();
+
+      // Wait for the message to be relayed to L1. Check addresses in the logs match the L1 token and bridge.
+      const [msgHash2] = await watcher.getMessageHashesFromL2Tx(withdrawTx.hash);
```
nicholaspai, line 108: Is this transfer instant because its on a test node?

packages/core/test-e2e/insured-bridge/helpers/MultiProviderHelper.js
```diff
@@ -0,0 +1,67 @@
...
+function findArtifactFromPaths(contractName, artifactsPath) {
+  const allArtifactsPaths = getAllFiles(artifactsPath);
+  const desiredArtifactPaths = allArtifactsPaths.filter((a) => a.endsWith(`/${contractName}.json`));
+
+  if (desiredArtifactPaths.length !== 1)
+    throw new Error(`Couldn't find desired artifact or found too many for ${contractName}`);
+  return JSON.parse(fs.readFileSync(desiredArtifactPaths[0], "utf-8"));
+}
+
+function getOptimismArtifact(contractName, ovm = false) {
```
nicholaspai, line 35: So this returns all artifacts that come with the eth-optimism npm package? This is kind of ugly and will be a pain to maintain as relative paths move around but I can't see a better way unless `eth-optimism` exports a helper

nicholaspai: Actually, why not add the required eth-optimism contracts to our `core/contracts/external` directory?

nicholaspai (approved):
This looks great as a stub so I won't block you but have some questions about implementations of helpers

packages/core/test-e2e/insured-bridge/e2e_InsuredBridge.js
chrismaree: few points:
- 1 this does not return an ABI, it returns ABI+bytecode
- the bytecode for the L2 token is compiled to work on optimism (not L1) and so it wont work on between networks.
- the `L2StandardERC20` has permissions to be `mint` and `burn` from an `onlyL2Bridge` modifier. i.e it the bridge contract on L2 can create and destroy tokens on L2.
- the `L2StandardERC20` implements `supportsInterface` which the bridge uses to validate the correctness of the token.
- in production, the bridge will use `OVM_L2StandardTokenFactory` to mint new instances of this token. By us deploying like this we get the exact same contract as production would use on L2.

packages/core/test-e2e/insured-bridge/e2e_InsuredBridge.js
chrismaree: yup

packages/core/test-e2e/insured-bridge/helpers/MultiProviderHelper.js
chrismaree: it just returns the one artifact of the particular `contractName` you provided., not all artifacts. it will search through all `artifacts` the package has though.

I agree it's not ideal. ideally, we should use the `getContractFactory` that the optimism `@eth-optimism/contracts` package exports. However, this does not work consistently for some reason and I have a bunch of problems actually using this function in these tests. Hopefully they fix this in the future (or add other ways of getting these artefacts) and then we can remove this logic and rather use their methods for this kind of call.

chrismaree: Regarding putting it in `core/contracts/external`, that is possible but IMHO it's a bit messy. we don't need those artifacts anywhere in the repo and it's actually a fair amount of code that we'd have to copy over.

packages/common/src/HardhatConfig.js
```diff
@@ -29,6 +29,9 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
     settings: { optimizer: { enabled: true, runs: 200 } },
   };
 
+  // Some tests in the /test directory should not be tested using hardhat. Define these tests to ignore.
+  const testBlacklist = ["insured-bridge/e2e"];
```
mrice32, line 33: @daywiss employed a strategy that I thought was really effective for designing "integration" tests in-line with other tests -- by using an extension: `*.e2e.js`. These get ignored in the normal test command.

Nothing to change now, but it might be ideal to have an easy rule to distinguish integration tests from normal hardhat tests (maybe e2e file path or extension) and have a separate pre-built command to run just those integration tests (if there are different sub-types that need diff environments, then a way to run each type).

packages/common/src/hardhat/tasks/test.js
```diff
@@ -31,5 +29,18 @@ internalTask(TASK_TEST_GET_TEST_FILES, async (_, { config, network }, runSuper)
     });
   }
 
+  const blacklist = config.networks[network.name].testBlacklist;
```
mrice32, line 32: Cool! This means we have a lot of control over how these are managed/processed.

packages/core/.mocharc.js
```diff
@@ -0,0 +1,15 @@
...
+process.on("unhandledRejection", (reason, promise) => {
+  console.error("Unhandled Rejection during test execution:", promise, "reason:", reason);
+  process.exit(1);
+});
+
+module.exports = {
+  require: ["ts-node/register/transpile-only"],
+  extension: ["js"],
+  watchExtensions: ["js"],
+  spec: ["test/insured-bridge/e2e/e2e_InsuredBridge.js"],
```
mrice32, line 13: Similar to the above, instead of specifying tests individually, maybe we could have a standard way of recognizing them so we wouldn't need to update any of these configs when we add new integration tests of this type.

packages/core/test/insured-bridge/e2e/helpers/MultiProviderHelper.js
```diff
@@ -0,0 +1,67 @@
...
+
+  files.forEach((file) => {
+    if (fs.statSync(dirPath + "/" + file).isDirectory()) arrayOfFiles = getAllFiles(dirPath + "/" + file, arrayOfFiles);
+    else arrayOfFiles.push(path.join(dirPath, "/", file));
+  });
+
+  return arrayOfFiles;
+}
+
+function findArtifactFromPaths(contractName, artifactsPath) {
```
mrice32, line 26: Are we doing all of this because these strategies won't work for optimism artifacts?

https://hardhat.org/guides/compile-contracts.html#reading-artifacts

The multiple sets of artifacts breaks a lot of assumptions. How often do you think we'll need to work around this. For instance, we can't really export multiple sets of artifacts with our current infrastructure -- is exporting optimism-specific artifacts something we should worry about?

packages/core/test/insured-bridge/e2e/helpers/MultiProviderHelper.js
```diff
@@ -0,0 +1,67 @@
...
+
+  if (desiredArtifactPaths.length !== 1)
+    throw new Error(`Couldn't find desired artifact or found too many for ${contractName}`);
+  return JSON.parse(fs.readFileSync(desiredArtifactPaths[0], "utf-8"));
+}
+
+function getOptimismArtifact(contractName, ovm = false) {
+  const artifactsPath = ovm
+    ? path.resolve(__dirname, "../../../../../../node_modules/@eth-optimism/contracts/artifacts-ovm")
+    : path.resolve(__dirname, "../../../../../../node_modules/@eth-optimism/contracts/artifacts");
```
mrice32, line 38: Hardcoded module paths can be dangerous as yarn assumes that node is smart about resolving up the directory tree. Thoughts on using `require.resolve("@eth-optimism/package.json")` to ensure you're using that same resolution strategy?

packages/core/test/insured-bridge/e2e/e2e_InsuredBridge.js
```diff
@@ -0,0 +1,118 @@
...
+
+  describe("Basic bridging functionality", async () => {
+    it("Can send tokens from L1 to L2", async () => {
+      // Wallet should have to tokens on L1 from the mint action and no tokens on L2.
+      assert.equal((await L1_ERC20.balanceOf(l1Wallet.address)).toString(), "1234");
+      assert.equal((await L2_ERC20.balanceOf(l2Wallet.address)).toString(), "0");
+
+      // Allow the gateway to lock up some of our tokens.
+      await L1_ERC20.approve(L1StandardBridge.address, 1234);
+      // await tx1.wait();
```
mrice32, line 83: nit comment

mrice32 (approved):
Looks good! A few comments around the setup, but nothing blocking!

packages/common/src/HardhatConfig.js
chrismaree: that's a _great_ idea. I was trying to think of a simple way to consistently ignore files that have a particular naming style. putting *.e2e.js` makes a lot of sense! I'll update this accordingly.

packages/core/test/insured-bridge/e2e/helpers/MultiProviderHelper.js
chrismaree: correct. the whole point of this is to let the importer explicitly choose which compilation environment to import the contacts from (`artifacts` or `artifacts-ovm` depending on the file.

In general, this might become something we need to do and factor into how we export artifacts. I can imagine a world where the `contracts-node` & `contracts-frontend` methods have a second prop that lets you define the EVM compatible compilation enviroment. i.e `optimism` as a selectable parameter, which will return the artifact compiled for that environment. For now, this is unlikely to be worth it but in the future, it might be required, especially if we have a whole set of contracts deployed on these networks.

Tracking even future-forward, it is possible (and somewhat likely) that there will be other L2's that have slightly modified opcodes & compilation processes that would necessitate this kind of exporting. 

For now though, I think it's fine to isolate this behavior to testing and we can figure out a more robust solution later. Especially seeing the export techniques are still somewhat in flux (with your PR #3229 for example).

packages/core/test/insured-bridge/e2e/helpers/MultiProviderHelper.js
chrismaree: good point. I'll see what I can do to avoid this.

packages/common/src/HardhatConfig.js
```diff
@@ -29,6 +29,9 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
     settings: { optimizer: { enabled: true, runs: 200 } },
   };
 
+  // Some tests in the should not be tested using hardhat. Define all tests that end with *e2e.js to be ignored.
```
nicholaspai, line 32:
```suggestion
  // Some tests should not be tested using hardhat. Define all tests that end with *e2e.js to be ignored.
```

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -33,16 +35,11 @@ internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network
...
-
-  return [
-    ...filePaths,
-    `${ovmPath}/insured-bridge/implementation/L1_BridgePool.sol`,
-    `${ovmPath}/insured-bridge/implementation/L1_BridgeRouter.sol`,
-    `${ovmPath}/insured-bridge/implementation/L2_BridgeDepositBox.sol`,
-  ];
-
-  // return filePaths;
+  // Temp work around to exclude some directories from truffle while compiling with hardhat. Pulls in all contracts in
```
nicholaspai, line 38: This will have a merge conflict with #3256 because I put this entire section within an `else {}` statement. See [this](https://github.com/UMAprotocol/protocol/pull/3256#discussion_r678291803) comment.

packages/core/.mocharc.js
```diff
@@ -0,0 +1,24 @@
...
+// Fetch an array of all test files in this package.
+const allTestFiles = getAllFilesInPath(path.resolve(__dirname, "./test"));
+// Select only the ones that end with `e2e.js`.
+const endToEndTests = allTestFiles.filter((path) => path.endsWith(".e2e.js"));
+
+module.exports = {
+  require: ["ts-node/register/transpile-only"],
+  extension: ["js"],
+  watchExtensions: ["js"],
+  spec: endToEndTests,
```
nicholaspai, line 22: Can you add a comment somewhere that only the files set in `spec` will be run by mocha? At least, that's how I interpret this file.

packages/common/src/hardhat/tasks/compile.js
chrismaree: ye. it's ok. merge your code in first and I'll rebase and resolve the conflict tomorrow. Your PR is through CI and mine is still blocked.

packages/core/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "UMA smart contracts and unit tests",
   "dependencies": {
     "@eth-optimism/contracts": "^0.4.5",
+    "@eth-optimism/watcher": "^0.0.1-alpha.9",
```
nicholaspai, line 7: should this be a dev dep?

packages/core/test/insured-bridge/e2e/helpers/ArtifactsHelper.js
```diff
@@ -0,0 +1,47 @@
+const hre = require("hardhat");
+
+const { findArtifactFromPath } = require("@uma/common");
+
+function getProviders() {
+  return {
+    l1RpcProvider: new hre.ethers.providers.JsonRpcProvider("http://localhost:9545"),
```
nicholaspai, line 7: Is there any way to get these urls from `HardhatConfig.js`? Just thinking ahead where we possibly change these ports and don't want unintended errors here.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -0,0 +1,113 @@
...
+    );
+
+    l1Messenger = new ethers.Contract(
+      await l2AddressManager.getAddress("OVM_L1CrossDomainMessenger"),
+      getContractInterface("OVM_L1CrossDomainMessenger"),
+      l1RpcProvider
+    );
+
+    // Tool that helps watches and waits for messages to be relayed between L1 and L2.
+    watcher = new Watcher({
```
nicholaspai, line 44: Does `Watcher` only work with `ethers` contracts out of the box? Just wondering if we can use `web3` but no big deal if not.

nicholaspai (commented):
Looks awesome, just left some nits

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
chrismaree: yup. ethers only.

PR #3256, opened 2021-07-27
improve(insured-bridge): Add state variables, events, and function interfaces to BridgeRouter

nicholaspai:
**Motivation**

Follow up to #3249 with some modifications to state variable and function names. Will explain in individual comments.

Note that I only modified `L1_BridgeRouter.sol` and `L1_BridgePool.sol`

**Summary**
I also implemented the shorter admin functions.

Still need to add unit tests for implemented functions and state vars

**Testing**

- [ ]  New unit tests created
- [X]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgePool.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.7.6;
+
+/**
+ * @notice Contract deployed on L1 that ingests liquidity from passive liquidity providers and returns them claims to
+ * withdraw their funds.
+ */
+contract BridgePool {
+    event ProvidedLiquidity(address indexed token, uint256 amount, uint256 lpTokensMinted, address liquidityProvider);
```
nicholaspai, line 9: Do we envision this event occuring in this contract or in router?

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+
+import "@eth-optimism/contracts/libraries/bridge/OVM_CrossDomainEnabled.sol";
+
+import "./L2_BridgeDepositBox.sol";
+
+/**
+ * @notice Contract deployed on L1 that has an implicit reference to a DepositBox on L2 and provides methods for
+ * "Relayers" to fulfill deposit orders to that contract. The Relayers can either post capital to fulfill the deposit
+ * instantly, or request that the funds are taken out of the passive liquidity provider pool following a challenge period.
+ * @dev A "Deposit" is an order to send capital from L2 to L1, and a "Relay" is a fulfillment attempt of that order.
```
nicholaspai, line 12: Note this point carefully: I designate all orders sent from L2 as "Deposits". Any attempt to fulfill is a deposit is a "Relay".

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+        // The following params are inferred and set by the L2 deposit contract:
+        address l2Sender;
+        uint256 depositTimestamp;
+        uint256 depositNetId;
+        // Relayer will compute the realized fee considering the amount of liquidity in this contract and the pending
+        // withdrawals at the depositTimestamp.
+        uint256 realizedFee;
+        // A deposit can have both a slow and an instant relayer if a slow relay is "sped up" from slow to instant. In
+        // these cases, we want to store both addresses for separate payouts.
+        address slowRelayer;
```
nicholaspai, line 53: I renamed "initialRelayer" to "slowRelayer" because its more consistent with the `DepositState` and `DepositType` but let me know if the `initialRelayer` is NOT always the slow relayer.

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+        address instantRelayer;
+        // TODO: Not sure how this will be used or why its stored but its in the interface doc
+        bytes priceRequestAncillaryData;
+    }
+    // Associates each deposit with a unique ID.
+    mapping(uint256 => Deposit) deposits;
+
+    event AddedDepositContract(address indexed l2DepositContract, uint256 indexed netId);
```
nicholaspai, line 61: I renamed some of these events to be more consistent with the state variables

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    // Liquidity provider functions
+
+    function deposit(address l1Token, uint256 amount) public {}
+
+    function withdraw(address lpToken, uint256 amount) public {}
+
+    // Relayer functions
+
+    function relayDeposit(
```
nicholaspai, line 112: So @chrismaree IIUC, every `l2Token` will be automatically mapped to its `l1Token` via the `whitelistedTokens` mapping. Therefore, the `sender` from L2 needs to be aware of this mapping and set the `l1Token` address correctly in their original L2 deposit?

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgePool.sol
chrismaree: I think it makes sense to happen here like you've got it

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
chrismaree: great terminology 👌

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
```diff
@@ -0,0 +1,127 @@
...
+ * instantly, or request that the funds are taken out of the passive liquidity provider pool following a challenge period.
+ * @dev A "Deposit" is an order to send capital from L2 to L1, and a "Relay" is a fulfillment attempt of that order.
+ */
+contract BridgeRouter is OVM_CrossDomainEnabled {
+    // Finder used to point to latest OptimisticOracle and other DVM contracts.
+    address public finder;
+
+    // Links deposit contract addresses to L2 networks they are deployed on.
+    // TODO: Can we convert uint256 to uint128 or smaller based on some assumption about how large netIDs get?
+    mapping(uint256 => address) public netIdToDepositContract;
```
chrismaree, line 20: Are we sure we want to do multiple `netIDs` now? otherwise, this would simply be just a 1:1 mapping if we only did one network to start with with a simple `l2DepositBox` vairable

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
chrismaree: correct. My thinking is the only way to whitelist a token on L2 is by calling `whitelistToken` on this contract which will then call over the canonical bridge the `depositContract` to append to the mapping there. That way the sender does not need to know about that on L2 as it will actually be there in the contract already & the whitelisting process will be atomic over both networks.

chrismaree: notice how the function `whitelistToken` on the `L2_bridgeDepositBox` has the modifer `onlyFromCrossDomainAccount(l1Owner)` such that only this contract can whitelist tokens on L2

packages/core/contracts-ovm/insured-bridge/implementation/L1_BridgeRouter.sol
nicholaspai: Ah yes I recall us reverting back to 1 deposit contract. Reverting.

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -39,10 +39,7 @@ internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network
 
   return [
     ...filePaths,
-    `${ovmPath}/insured-bridge/implementation/L1_BridgePool.sol`,
```
nicholaspai, line 42: @chrismaree I removed this file from this array because (1) it uses solidity 0.8 in order to import the @openzeppelin/ERC20 contract and (2) we presumably won't need to compile+deploy it to the OVM. Is this the right way to do this?

chrismaree: If this file Is still within the `contracts-ovm` directory, then this is not the right way to do this. What this modification is doing is appending additional contracts to the compilation step for hardhat. what you've done by removing this line is to actually ignore this contract from compilation in totality. To remove it from the `optimism` compilation, you simply need to remove this contract name from the hardhat config. However, the _only_ file currently in the hardhat config is `L2_BridgeDepositBox.sol` so this is the only contract that will be OVM compiled.

What this code is doing is basically enabling us to add contracts that hardhat sees to both the default `contracts` directory, as well as the `contracts-ovm` directory. I will add a follow-up PR soon to remove this logic where we have to specify each import line by line and will rather just pull in _all_ files within the `contracts-ovm` directory.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_L2BridgeDepositBox.sol
```diff
@@ -3,7 +3,7 @@ pragma solidity >=0.7.6;
 
 import { OVM_CrossDomainEnabled } from "@eth-optimism/contracts/libraries/bridge/OVM_CrossDomainEnabled.sol";
```
chrismaree, line 4: seeing you renamed this file, you will need to update the hardhat config which defines which `optimism` contracts get compiled here: https://github.com/UMAprotocol/protocol/blob/aeea8ae30c6f7d5065e3857d25c75b8d69005d81/packages/common/src/HardhatConfig.js#L75

packages/core/test/insured-bridge/OVM_L1BridgeRouter.js
```diff
@@ -0,0 +1,79 @@
...
+        await didContractThrow(bridgeRouter.methods.setDepositContract(newDepositContract).send({ from: rando })),
+        "OnlyOwner modifier not enforced"
+      );
+
+      const txn = await bridgeRouter.methods.setDepositContract(newDepositContract).send({ from: owner });
+      await assertEventEmitted(txn, bridgeRouter, "SetDepositContract", (ev) => {
+        return ev.l2DepositContract === newDepositContract;
+      });
+    });
+    it("Add whitelisted Token mapping", async () => {
```
chrismaree, line 56: looks good. next will be adding in the cross-chain logic for this to call into optimism to whitelist on the other chain. can validate this with smockit :)

packages/common/src/hardhat/tasks/compile.js
chrismaree: i.e you should _not_ remove this file from here and leave this `compilation.js` task as it is to ensure this file gets compiled correctly. I've re-worked how this kind of selective compilation works in another PR https://github.com/UMAprotocol/protocol/pull/3255. once that is in this will be a lot cleaner. for now what you should do is just revert the changes on this file and compile these contracts in hardhat.

nicholaspai: Reverted

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -32,17 +32,17 @@ internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network
       }
       return false;
     });
+    return filePaths;
+  } else {
```
nicholaspai, line 36: @chrismaree I had to make this change because `BridgePool` was getting compiled for the `--network optimism` flag but it uses Solidity 0.8. I think this change is ok because any contracts that are supposed to be compiled for the OVM should be in the `compileWhitelist`. If there is no `compileWhitelist`, then compile ALL contracts plus these extra `contracts-ovm` ones.

chrismaree (approved):
looks good, thanks for implementing the requested changes.

PR #3260, opened 2021-07-29
improve(insured-bridge): L2DepositBox Implementation and tests

chrismaree:
**Motivation**

Insured bridge requires a L2 deposit box.

**Summary**

This PR adds the implementation details and unit tests to the L2 deposit box.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/common/src/hardhat/tasks/test.js
```diff
@@ -18,14 +15,12 @@ internalTask(TASK_TEST_GET_TEST_FILES, async (_, { config, network }, runSuper)
   // Build absolute path for all directories on user-specified whitelist.
   const whitelist = config.networks[network.name].testWhitelist;
   if (whitelist && Array.isArray(whitelist)) {
-    const whitelistDirs = whitelist.map((x) => path.resolve(TESTS_DIR, x));
-
     filePaths = filePaths.filter((filePath) => {
-      for (let whitelistedDir of whitelistDirs) {
-        if (!path.resolve(process.cwd(), filePath).startsWith(whitelistedDir)) return false;
+      for (let whitelistString of whitelist) {
+        if (filePath.includes(whitelistString)) return true;
```
nicholaspai, line 20: So this changes the whitelist behavior such that any file that includes the whitelisted string is included? Previously it had to start with the whitelisted string IIUC.

Is my understanding correct?

packages/core/package.json
```diff
@@ -23,7 +23,6 @@
   "devDependencies": {
     "@eth-optimism/smock": "^1.1.9",
     "@awaitjs/express": "^0.3.0",
-    "@eth-optimism/smock": "^1.1.9",
```
nicholaspai, line 26: Why remove?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity >=0.7.6;
 
+import "./OVM_Testable.sol"; //TODO: replace this with the normal UMA Testable once we can use 0.8 solidity.
+
 import { OVM_CrossDomainEnabled } from "@eth-optimism/contracts/libraries/bridge/OVM_CrossDomainEnabled.sol";
+import { Lib_PredeployAddresses } from "@eth-optimism/contracts/libraries/constants/Lib_PredeployAddresses.sol";
+
+// Define some interfaces and helper libraries. This is temporary until such that time that we can bump the solidity
```
nicholaspai, line 9:
```suggestion
// Define some interfaces and helper libraries. This is temporary until we can bump the solidity
```

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+// version in these contracts to 0.8.x and import the rest of these libs from other UMA contracts in the rpo.
```
nicholaspai, line 10:
```suggestion
// version in these contracts to 0.8.x and import the rest of these libs from other UMA contracts in the repo.
```

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+    mapping(address => WhitelistedToken) public whitelistedTokens;
+
+    // Minimum time that must elapse between bridging actions for a given token. Used to rate limit bridging back to L1.
+    uint64 public minimumBridgingDelay;
+
+    /****************************************
+     *                EVENTS                *
+     ****************************************/
+
+    event L1WithdrawContractChanged(address oldL1WithdrawContract, address newL1WithdrawContract);
```
nicholaspai, line 74: I name the corresponding event in WithdrawalBox: `SetDepositContract` so maybe more consistent to name this `SetWithdrawalContract`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+    event TokenWhitelisted(address l1Token, address l2Token);
```
nicholaspai, line 75: Similar reasoning above, maybe rename this to `WhitelistToken`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+    event DepositsEnabled(bool enabledResultantState);
+    event L2FundsDeposited(
```
nicholaspai, line 77: I think the `L2` prefix is redundant, can just rename `FundsDeposited`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
         uint256 transferId,
```
nicholaspai, line 78: For consistency think we should use `depositId` instead of `transferId`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+     * @param _l1WithdrawContract Address of the bridge withdraw contract on L1. Tokens are sent to this address at the
+     *          conclusion of fast relay. This address is the "owner" of the deposit box.
+     * @param _minimumBridgingDelay Minimum second that must elapse between L2->L1 token transfer to prevent dos.
+     * @param timerAddress Timer used to synchronize contract time in testing. Set to 0x000... in production.
+     */
+    constructor(
+        address _l1WithdrawContract,
+        uint64 _minimumBridgingDelay,
+        address timerAddress
+    ) OVM_CrossDomainEnabled(Lib_PredeployAddresses.L2_CROSS_DOMAIN_MESSENGER) OVM_Testable(timerAddress) {
```
nicholaspai, line 108: Should I be using `Lib_PredeployAddresses.L1_CROSS_DOMAIN_MESSENGER` in `WithdrawalBox`?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+     *          ADMIN FUNCTIONS           *
+     **************************************/
 
-    function whitelistToken(address l1Token, address l2Token) public onlyFromCrossDomainAccount(l1Owner) {}
+    /**
+     * @notice Changes the L1 withdraw associated with this L2 deposit box.
+     * @dev Only callable by the existing l1WithdrawContract via the optimism cross domain messenger.
+     * @param newL1WithdrawContract address of the new L1 withdrawContract.
+     */
+    function changeL1WithdrawContract(address newL1WithdrawContract)
```
nicholaspai, line 122: Think we should rename to `setWithdrawContract`

nicholaspai: Generally think we should get rid of L1/L2 reference in this contract and just let the comments explain.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+        public
+        onlyFromCrossDomainAccount(l1WithdrawContract)
+    {
+        emit L1WithdrawContractChanged(l1WithdrawContract, newL1WithdrawContract);
```
nicholaspai, line 126: nit: should emit the event after changing state

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+     * @dev Only callable by the existing l1WithdrawContract via the optimism cross domain messenger.
+     * @param l1Token Address of the canonical L1 token. This is the token users will receive on Ethereum.
+     * @param l2Token Address of the L2 token representation. This is the token users would deposit on optimism.
+     */
+    function whitelistToken(address l1Token, address l2Token) public onlyFromCrossDomainAccount(l1WithdrawContract) {
+        whitelistedTokens[l2Token] = WhitelistedToken({ l1Token: l1Token, lastBridgeTime: uint64(getCurrentTime()) });
 
-    function transferL1Owner(address _l1Owner) public onlyFromCrossDomainAccount(l1Owner) {
-        l1Owner = _l1Owner;
+        emit TokenWhitelisted(l1Token, l2Token);
```
nicholaspai, line 139: Is it worth emitting the `lastBridgeTime` in this event?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
     }
 
-    // Depositor functions
+    /**
+     * @notice L1 owner can enable/disable deposits over all whitelisted tokens.
+     * @dev Only callable by the existing l1WithdrawContract via the optimism cross domain messenger.
+     * @param _depositsEnabled bool to set if the deposit box should accept/reject deposits.
+     */
+    function setEnableDeposits(bool _depositsEnabled) public onlyFromCrossDomainAccount(l1WithdrawContract) {
```
nicholaspai, line 147: Thoughts on renaming this to `resetEnableDeposits` or `toggleEnableDeposits` since there are no practical cases where you'd set `depositsEnabled = true` if its already `true` (same with `false`). You'd only want to flip the switch.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
     function deposit(
         address recipient,
-        address l1Token,
+        address l2Token,
         uint256 amount,
         uint256 maxFee
-    ) public {}
+    ) public onlyIfDepositsEnabled() {
+        require(isTokenWhitelisted(l2Token), "deposit token not whitelisted");
+        require(maxFee <= 1e18, "max fee can not be over 1e18");
```
nicholaspai, line 172: why this setting? what if precision is 6 decimals?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+
+        TokenHelper.safeTransferFrom(l2Token, msg.sender, address(this), amount);
```
nicholaspai, line 174: Should this external call come last to match check-effects-interaction?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+    /**
+     * @notice Called by relayer (or any other EOA) to move a batch of funds from the deposit box, through the canonical
+     *      token bridge, to the L1 Withdraw box.
+     * @dev The frequency that this function can be called is rate limited by the `minimumBridgingDelay` to prevent spam
+     *      on L1 as the finalization of a L2->L1 tx is quite expensive.
+     * @param l2Token L2 token to relay over the canonical bridge.
+     * @param l1Gas Unused by optimism, but included for potential forward compatibility considerations.
+     */
+    function bridgeTokens(address l2Token, uint32 l1Gas) public {
+        uint256 bridgeDepositBoxBallance = TokenLike(l2Token).balanceOf(address(this));
```
nicholaspai, line 202: nit: `balance` not `ballance`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+        require(bridgeDepositBoxBallance > 0, "can't bridge zero tokens");
+        require(isTokenWhitelisted(l2Token), "can't bridge non-whitelisted token");
+        require(hasEnoughTimeElapsedToBridge(l2Token), "not enough time has elapsed from previous bridge");
+
+        StandardBridgeLike(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo(
+            l2Token, // _l2Token. Address of the L2 token to bridge over.
+            l1WithdrawContract, // _to. Withdraw, over the bridge, to the l1Withdraw contract.
```
nicholaspai, line 209: Do we need to check that the withdrawContract is not 0x0? Or do we do this elsewhere? Just want to make sure we're not ever able to send tokens to the 0x0 address

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,237 @@
...
+            bridgeDepositBoxBallance, // _amount. Send the full balance of the deposit box to bridge.
+            l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+            "0x" // _data. TODO: add additional info into this data prop this.
+        );
+
+        emit TokensBridged(l2Token, bridgeDepositBoxBallance, l1Gas, msg.sender);
```
nicholaspai, line 215: Don't we need to reset `lastBridgeTime` now? Also, should make external calls come last including after events

packages/core/contracts-ovm/insured-bridge/implementation/OVM_Testable.sol
```diff
@@ -0,0 +1,31 @@
+// Temp contract until we can use the canonical UMA testable once Optimism support solidity 0.8.
```
nicholaspai, line 1: No modifications right? Worth adding a comment that its copied straight with only the version changed

packages/core/test/insured-bridge/BridgeDepositBox.js
```diff
@@ -0,0 +1,297 @@
+// These tests are meant to be run within the `hardhat` network (not OVM) and use smockit to mock L2 contracts
+// such as the cross domain message bridge ect. The context of all calls, tokens and setup are the mocked L2 network.
```
nicholaspai, line 2:
```suggestion
// such as the cross domain message bridge etc. The context of all calls, tokens and setup are the mocked L2 network.
```

packages/core/test/insured-bridge/BridgeDepositBox.js
```diff
@@ -0,0 +1,297 @@
...
+const BridgeDepositBox = getContract("OVM_BridgeDepositBox");
+
+// Helper contracts
+const Token = getContract("ExpandedERC20");
+const Timer = getContract("OVM_Timer");
+
+// Contract objects
+let depositBox, l2CrossDomainMessengerMock, l1TokenAddress, l2Token, timer;
+
+// As these tests are in the context of l2, we dont have the deployed notion of an "L1 Token". The L1 token is within
```
nicholaspai, line 25: +1 comment

packages/core/test/insured-bridge/BridgeDepositBox.js
```diff
@@ -0,0 +1,297 @@
+// These tests are meant to be run within the `hardhat` network (not OVM) and use smockit to mock L2 contracts
```
nicholaspai, line 1: nit: rename this file to match contract "OVM_BridgeDepositBox.js"

nicholaspai (commented):
Amazing you wrote all this so quickly, left a bunch of nits let me know what you think!

packages/common/src/hardhat/tasks/test.js
chrismaree: correct. Are you ok with that change? it just makes defining the whitelist a bit simpler to reason about and simplified the code.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: done.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: done

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: done

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: ideally, yes. they'll be the same on all networks and it makes things a bit cleaner imo.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: sure, agreed & done.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: I put it before intentionally. else, you'd need to store the `l1WithdrawContract` to get the `oldL1WithdrawContract` address.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: is that not just the block timestamp that you can get from the event?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: this represents a fraction of `amount` of the token that will be sent as a fee. i.e If I'm willing to take a 5% fee this will be 0.05 or 0.05e18. by definition, I cant take more than 100% as a fee (or 1e18) in the context of this scalling.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: sure. done.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: hhhmmm. that is an interesting point. we could add a modifier that checks `isWithdrawContractSet` or a `require` for this. wdyt?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: great catch! will do.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_Testable.sol
chrismaree: done.

packages/core/package.json
chrismaree: it was duplicated in this package.json :)

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: done

packages/common/src/hardhat/tasks/test.js
nicholaspai: I'm good with this!

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: I think actually you should change the event signature to only emit the new withdrawal address. Then, in the constructor call an internal method `_setWithdrawContract(address newWithdrawContract)` which will emit the address set upon construction. This follows the pattern I use in `BridgeRouter`. 

On a side note, I don't actually set the Deposit contract in the constructor. WDYT about setting Deposit/Withdraw in constructor? My reasoning was depending on the order of deployments we may not know the addresses ahead of time

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: Think its better to be explicit and emit it in the event also

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: Ok SGTM

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: Ah ok this was definitely not clear. Would recommend you declare a constant ONE_HUNDRED_PCT = 1e18 at top of file and then rename `maxFee` to `maxFeePct`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: I think use a modifier. I plan to add a modifier for checking whether the depositContract is set and using that on any functions that call `sendCrossDomainMessage`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,235 @@
...
+ */
+
+contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
+    /*************************************
+     *  OVM DEPOSIT BOX DATA STRUCTURES  *
+     *************************************/
 
-contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled {
-    address public l1Owner;
+    address public l1WithdrawContract;
```
nicholaspai, line 51: nit: rename to `withdrawContract` as l1 is redundant

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,235 @@
...
+        StandardBridgeLike(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo(
+            l2Token, // _l2Token. Address of the L2 token to bridge over.
+            l1WithdrawContract, // _to. Withdraw, over the bridge, to the l1Withdraw contract.
+            bridgeDepositBoxBalance, // _amount. Send the full balance of the deposit box to bridge.
+            l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+            "0x" // _data. TODO: add additional info into this data prop this.
+        );
+
+        emit TokensBridged(l2Token, bridgeDepositBoxBalance, l1Gas, msg.sender);
+        whitelistedTokens[l2Token].lastBridgeTime = uint64(getCurrentTime());
```
nicholaspai, line 213: Should be making external contract call to `StandardBridgeLike` after all state modifications are done.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: sure, can do the event change and call the `setWithdrawContract` in the constructor to get the event.

On your side note, I do set the `l1WithdrawContract` contract in the constructor. However, we cant have _both_ the deposit and withdrawal contract set them in the constructor as you need to deploy one contract first, then the other. ideally, we deploy on L1 first, then deploy on L2 with the withdraw contract as a param in the deposit contract constructor then finally initialize the L1 contract with the L2 contract address.

chrismaree: as a side note, I also need to add a function, event and tests for setting the `minimumBridgingDelay`. I will follow the same pattern where this is called in the constructor.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: sure. done and added a test for it.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: ok I'll change it to `maxFeePct`. would really rather not define the `ONE_HUNDRED_PCT` const though; feels like that is excessive. I'll change the require message to make it more clear.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: what I did to combat this situation is add a require check in the initialization of this state variable to ensure it is set correctly so we can never get into a state where the withdrawContract is 0x0.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: SGTM

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: changed to `bridgeRouter` to match the name of the actual contract

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: good point. done.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
+ * @title OVM Bridge Deposit Box.
+ * @notice Accepts deposits on Optimism L2 to relay to Ethereum L1 as part of the UMA insured relayer system.
+ */
+
+contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
+    /*************************************
+     *  OVM DEPOSIT BOX DATA STRUCTURES  *
+     *************************************/
+
+    address public bridgeRouter;
```
nicholaspai, line 51: nit: add comment above state variable

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
-contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled {
-    address public l1Owner;
+    bool public depositsEnabled = true;
 
-    event Deposit(
-        uint256 transferId,
+    // Track the total number of deposits. Used as a unique identifier for bridged transfers.
+    uint256 public numberOfDeposits;
+
+    struct WhitelistedToken {
```
nicholaspai, line 58: nit: add comment explaining that this struct represents the mappings to a whitelisted L2 token. FYI in the BridgeRouter I name this struct `L1TokenRelationships`, if you like that name then you could name this `L2TokenRelationships` but I can see why you'd prefer `WhitelistedToken`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
+     *          conclusion of fast relay. This address is the "owner" of the deposit box.
+     * @param _minimumBridgingDelay Minimum second that must elapse between L2->L1 token transfer to prevent dos.
+     * @param timerAddress Timer used to synchronize contract time in testing. Set to 0x000... in production.
+     */
+    constructor(
+        address _bridgeRouter,
+        uint64 _minimumBridgingDelay,
+        address timerAddress
+    ) OVM_CrossDomainEnabled(Lib_PredeployAddresses.L2_CROSS_DOMAIN_MESSENGER) OVM_Testable(timerAddress) {
+        require(_bridgeRouter != address(0), "Bad bridge router address");
```
nicholaspai, line 110: Is having the following duplicate code block preferable to calling an internal helper method? Considering bytecode, gas costs, and readability.

```
require(_bridgeRouter != address(0), "Bad bridge router address");
bridgeRouter = _bridgeRouter;
```

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
+     *          ADMIN FUNCTIONS           *
+     **************************************/
 
-    function whitelistToken(address l1Token, address l2Token) public onlyFromCrossDomainAccount(l1Owner) {}
+    /**
+     * @notice Changes the L1 withdraw associated with this L2 deposit box.
+     * @dev Only callable by the existing bridgeRouter via the optimism cross domain messenger.
+     * @param _L1WithdrawContract address of the new L1 withdrawContract.
+     */
+    function setWithdrawContract(address _L1WithdrawContract) public onlyFromCrossDomainAccount(bridgeRouter) {
```
nicholaspai, line 124: nit: rename function param to `_bridgeRouter`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: You'd also want to emit an event `SetBridgeRouter`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
+    mapping(address => WhitelistedToken) public whitelistedTokens;
+
+    // Minimum time that must elapse between bridging actions for a given token. Used to rate limit bridging back to L1.
+    uint64 public minimumBridgingDelay;
+
+    /****************************************
+     *                EVENTS                *
+     ****************************************/
+
+    event SetWithdrawContract(address newL1WithdrawContract);
```
nicholaspai, line 74: nit: rename to `SetBridgeRouter` and change param to `newBridgeRouter`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
         address recipient,
-        address l1Token,
+        address l2Token,
         uint256 amount,
-        uint256 maxFee
-    ) public {}
+        uint256 maxFeePct
+    ) public onlyIfDepositsEnabled() {
+        require(isWhitelistToken(l2Token), "deposit token not whitelisted");
+        require(maxFeePct <= 1e18, "max fee can not be over 1e18");
```
nicholaspai, line 182: I know we discussed this but can you just add a small comment that "1e18 represents 100%"

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,249 @@
...
+
+        emit FundsDeposited(
+            numberOfDeposits, // the current number of deposits acts as a deposit ID (nonce).
```
nicholaspai, line 185: This comment seems unneccessary. If you really need to clarify it why not just rename `numberOfDeposits` to `depositNonce`?

nicholaspai (commented):
Almost there I think, another round of nits

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: I have a comment on the mapping. is that not sufficient? I will change to `L2TokenRelationships` for consistency.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: ye, good point. fixed.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: yup. done. did the same for minimumBridgeDelay

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: Sure

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: I prefer it as `numberOfDeposits` as it's a counter as a variable, rather than `depositNonce`.

chrismaree: it might be worthwhile to set an offset for this at the launch of the contract in the case we re-deploy this contract. however, we can always add this at a later point if/when upgrading

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: I added a better require statement and better natspec comments.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: yeah good point.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,267 @@
...
+    uint64 public minimumBridgingDelay;
+
+    /****************************************
+     *                EVENTS                *
+     ****************************************/
+
+    event SetBridgeRouter(address newBridgeRouterContract);
+    event SetMinimumBridgingDelay(uint64 newMinimumBridgingDelay);
+    event WhitelistToken(address l1Token, address l2Token, uint64 lastBridgeTime);
+    event DepositsEnabled(bool enabledResultantState);
```
nicholaspai, line 78:
```suggestion
    event DepositsEnabled(bool depositsEnabled);
```

Simpler name IMO

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -1,47 +1,267 @@
...
         address recipient,
-        address l1Token,
+        address l2Token,
         uint256 amount,
-        uint256 maxFee
-    ) public {}
+        uint256 maxFeePct
+    ) public onlyIfDepositsEnabled() {
+        require(isWhitelistToken(l2Token), "deposit token not whitelisted");
+        require(maxFeePct <= 1e18, "maxFeePct can not be over 100% (represented as 1e18)");
```
nicholaspai, line 185:
```suggestion
        require(maxFeePct <= 1e18, "maxFeePct cannot be over 100% (represented as 1e18)");
```

nicholaspai (approved):
Last nits! LGTM

PR #3263, opened 2021-07-30
feat(fpl): add generalized success token FPL with parameterized base percentage

pemulis:
**Motivation**

We would like to have a generalized `SuccessToken` financial product library where the base percentage can be set as a parameter, with the remaining percentage acting like an embedded call option.

**Summary**

Using the `SimpleSuccessToken` as a model, this PR adds a generalized `SuccessToken` financial product library and associated tests.

**Details**

* We discussed creating an even more generalized financial product library, where the "option" part of the payout could follow arbitrary logic defined in an external financial product library, which could be a call option, binary option, range token, or something else. It was determined that this was actually harder to reason about as a deployer, in addition to being more difficult to test, so we stuck with the call option logic for the option part of the payout. The ultra-generalized version of this will be brought up in a Discourse thread for further discussion with the community.
* It might make sense to delete the `SimpleSuccessToken` contract and tests from the repo as part of this pull request, since the generalized version makes it redundant. We didn't do so yet, since that contract is still used for a proof-of-concept as part of an active deal pursuit. If we don't use the `SimpleSuccessToken` for a real contract, we should delete it later.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
@@ -1,0 +1,91 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "./LongShortPairFinancialProductLibrary.sol";
+import "../../../../common/implementation/Lockable.sol";
+
+/**
+ * @title Success Token Long Short Pair Financial Product Library.
+ * @notice Adds settlement logic to create success token LSPs. A success token pays out a fixed amount of
+ * collateral as a floor, with the remaining amount functioning like an embedded option. The embedded
+ * option can use the payout logic of other financial product libraries, and resemble a covered call, a
+ * KPI option, a range bond, a capped yield dollar, or custom payout logic in a new financial library.
+ */
+contract SuccessTokenLongShortPairFinancialProductLibrary is LongShortPairFinancialProductLibrary, Lockable {
+    using FixedPoint for FixedPoint.Unsigned;
+
+    struct SuccessTokenLongShortPairParameters {
+        uint256 strikePrice;
+        uint256 basePercentage;
+    }
+
+    mapping(address => SuccessTokenLongShortPairParameters) public longShortPairParameters;
+
+    /**
+     * @notice Enables any address to set the strike price for an associated LSP.
+     * @param longShortPair address of the LSP.
+     * @param strikePrice the strike price for the covered call for the associated LSP.
+     * @param basePercentage the base percentage of collateral per pair paid out to long tokens, expressed
+     * with 1e18 decimals. E.g., a 50% base percentage should be expressed 500000000000000000, or 0.5 with
+     * 1e18 decimals. The base percentage can not be set to 0.
+     * @dev Note: a) Any address can set the initial strike price b) A strike price cannot be 0.
+     * c) A strike price can only be set once to prevent the deployer from changing the strike after the fact.
+     * d) For safety, a strike price should be set before depositing any synthetic tokens in a liquidity pool.
+     * e) longShortPair must expose an expirationTimestamp method to validate it is correctly deployed.
+     */
+    function setLongShortPairParameters(
+        address longShortPair,
+        uint256 strikePrice,
+        uint256 basePercentage
+    ) public nonReentrant() {
+        require(ExpiringContractInterface(longShortPair).expirationTimestamp() != 0, "Invalid LSP address");
+        SuccessTokenLongShortPairParameters memory params = longShortPairParameters[longShortPair];
+        require(params.strikePrice == 0 && params.basePercentage == 0, "Parameters already set");
+        require(basePercentage != 0, "Base percentage cannot be set to 0");
+
+        longShortPairParameters[longShortPair] = SuccessTokenLongShortPairParameters({
+            strikePrice: strikePrice,
+            basePercentage: basePercentage
+        });
+    }
+
+    /**
+     * @notice Returns a number between 0 and 1e18 to indicate how much collateral each long and short token are entitled
+     * to per collateralPerPair.
+     * @param expiryPrice price from the optimistic oracle for the LSP price identifier.
+     * @return expiryPercentLong to indicate how much collateral should be sent between long and short tokens.
+     */
+    function percentageLongCollateralAtExpiry(int256 expiryPrice)
+        public
+        view
+        override
+        nonReentrantView()
+        returns (uint256)
+    {
+        SuccessTokenLongShortPairParameters memory params = longShortPairParameters[msg.sender];
+        require(params.strikePrice != 0 || params.basePercentage != 0, "Params not set for calling LSP");
+
+        // If the expiry price is less than the strike price then the long options expire worthless (out of the money).
+        // In this case, return of value of the base percentage to the long tokenholders.
+        // Note we do not consider negative expiry prices in this implementation.
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+        if (positiveExpiryPrice == 0 || uint256(positiveExpiryPrice) <= contractStrikePrice) return basePercentage;
+
+        // Else, token expires to be worth basePercentage + (1 - basePercentage) * (expiryPrice - strikePrice).
+        // E.g., if SUSHI is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667%
+        // which if the collateralPerPair is 2, is worth 1.3333 $SUSHI, which is worth $40 if 1 $SUSHI is worth $30.
+        // This return value is strictly < 1. The return value tends to 1 as the expiryPrice tends to infinity.
+        return
+            (
+                FixedPoint.Unsigned(basePercentage).add(
+                    FixedPoint
+                        .Unsigned(uint256(1000000000000000000) - basePercentage)
+                        .mul(
+                        FixedPoint.Unsigned(uint256(positiveExpiryPrice)).sub(FixedPoint.Unsigned(contractStrikePrice))
+                    )
+                        .div(FixedPoint.Unsigned(uint256(positiveExpiryPrice)))
+                )
+            )
+                .rawValue;
+    }
+}
--- a/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
+++ b/packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
@@ -69,21 +69,19 @@ contract SuccessTokenLongShortPairFinancialProductLibrary is LongShortPairFinanc
         // In this case, return of value of the base percentage to the long tokenholders.
         // Note we do not consider negative expiry prices in this implementation.
         uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
-        if (positiveExpiryPrice == 0 || uint256(positiveExpiryPrice) <= contractStrikePrice) return basePercentage;
+        if (positiveExpiryPrice == 0 || positiveExpiryPrice <= contractStrikePrice) return basePercentage;
 
         // Else, token expires to be worth basePercentage + (1 - basePercentage) * (expiryPrice - strikePrice).
-        // E.g., if SUSHI is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667%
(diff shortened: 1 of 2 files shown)


packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+     */
+    function setLongShortPairParameters(
+        address longShortPair,
+        uint256 strikePrice,
+        uint256 basePercentage
+    ) public nonReentrant() {
+        require(ExpiringContractInterface(longShortPair).expirationTimestamp() != 0, "Invalid LSP address");
+        SuccessTokenLongShortPairParameters memory params = longShortPairParameters[longShortPair];
+        require(params.strikePrice == 0 && params.basePercentage == 0, "Parameters already set");
+        require(basePercentage != 0, "Base percentage cannot be set to 0");
```
nicholaspai, line 44: Should you also require `strikePrice != 0`? This way the previous `require` always reverts after the params are set. Also in your comments you say that the strike price cannot be 0

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+
+    mapping(address => SuccessTokenLongShortPairParameters) public longShortPairParameters;
+
+    /**
+     * @notice Enables any address to set the strike price for an associated LSP.
+     * @param longShortPair address of the LSP.
+     * @param strikePrice the strike price for the covered call for the associated LSP.
+     * @param basePercentage the base percentage of collateral per pair paid out to long tokens, expressed
+     * with 1e18 decimals. E.g., a 50% base percentage should be expressed 500000000000000000, or 0.5 with
+     * 1e18 decimals. The base percentage can not be set to 0.
```
nicholaspai, line 30:
```suggestion
     * 1e18 decimals. The base percentage cannot be set to 0.
```

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
nicholaspai: Basically, repeat the logic of the `require` statement on line 66

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,89 @@
...
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+        if (positiveExpiryPrice == 0 || positiveExpiryPrice <= params.strikePrice) return params.basePercentage;
+
+        // Else, token expires to be worth basePercentage + (1 - basePercentage) * (expiryPrice - strikePrice).
+        // E.g., if $TOKEN is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667%
+        // which if the collateralPerPair is 2, is worth 1.3333 $TOKEN, which is worth $40 if 1 $TOKEN is worth $30.
+        // This return value is strictly < 1. The return value tends to 1 as the expiryPrice tends to infinity.
+        return
+            (
+                FixedPoint.Unsigned(params.basePercentage).add(
```
nicholaspai, line 80: Due to rounding down/precision errors this might return a slightly smaller value than expected. Are we ok with that?

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,141 @@
...
+
+  beforeEach(async () => {
+    successTokenLSPFPL = await SuccessTokenLongShortPairFinancialProductLibrary.new().send({ from: accounts[0] });
+    lspMock = await LongShortPairMock.new(
+      "1000000", // _expirationTimestamp
+      collateralPerPair // _collateralPerPair
+    ).send({ from: accounts[0] });
+  });
+  describe("Long Short Pair Parameterization", () => {
+    it("Can set and fetch valid strikes", async () => {
```
nicholaspai, line 38: Should add asserts that you can't set `strikePrice`/`basePercentage` equal to 0. General rule of thumb is that every single require statement should be tested once in a unit test.

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,141 @@
...
+        await didContractThrow(
+          successTokenLSPFPL.methods
+            .setLongShortPairParameters(ZERO_ADDRESS, strikePrice, basePercentage1)
+            .send({ from: accounts[0] })
+        )
+      );
+    });
+  });
+  describe("Compute expiry tokens for collateral with base percentage of 0.5", () => {
+    beforeEach(async () => {
```
nicholaspai, line 72: Can you add a test that `percentageLongCollateralAtExpiry` reverts if the params are not set yet?

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,141 @@
...
+      assert(
+        await didContractThrow(
+          successTokenLSPFPL.methods
+            .setLongShortPairParameters(ZERO_ADDRESS, strikePrice, basePercentage1)
+            .send({ from: accounts[0] })
+        )
+      );
+    });
+  });
+  describe("Compute expiry tokens for collateral with base percentage of 0.5", () => {
```
nicholaspai, line 71: Can you add a test for when the expiry price = 0 since this is an edge case in the contract?

packages/core/test/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.js
```diff
@@ -0,0 +1,141 @@
...
+    });
+    it("Should never return a value greater than 1", async () => {
+      // Create a massive expiry price. 1e18*1e18. Under all conditions should return less than 1.
+      const expiryTokensForCollateral = await successTokenLSPFPL.methods
+        .percentageLongCollateralAtExpiry(toWei(toWei("1")))
+        .call({ from: lspMock.options.address });
+      assert.isTrue(toBN(expiryTokensForCollateral).lt(toBN(toWei("1"))));
+    });
+  });
+  describe("Compute expiry tokens for collateral with base percentage of 0.2", () => {
```
nicholaspai, line 106: Can you add a test for when the expiry price = 0 since this is an edge case in the contract?

nicholaspai (commented):
Thanks for this! I have one main contract comment and then requests to add unit tests to bolster code coverage.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: My gut feeling is yes, but would be interested if @smb2796 disagrees. I'll add a note about this.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,90 @@
...
+        require(params.strikePrice != 0 || params.basePercentage != 0, "Params not set for calling LSP");
+
+        // If the expiry price is less than the strike price then the long options expire worthless (out of the money).
+        // In this case, return of value of the base percentage to the long tokenholders.
+        // Note we do not consider negative expiry prices in this implementation.
+        uint256 positiveExpiryPrice = expiryPrice > 0 ? uint256(expiryPrice) : 0;
+        if (positiveExpiryPrice == 0 || positiveExpiryPrice <= params.strikePrice) return params.basePercentage;
+
+        // Else, token expires to be worth basePercentage + (1 - basePercentage) * (expiryPrice - strikePrice).
+        // E.g., if $TOKEN is $30 and strike is $20, long token is redeemable for 0.5 + 0.5*(30-20)/30 = 0.6667%
```
chrismaree, line 75: the example you have here has `basePercentage` set to 0.5 (50%). perhaps add this to the comment?

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,90 @@
...
+
+import "./LongShortPairFinancialProductLibrary.sol";
+import "../../../../common/implementation/Lockable.sol";
+
+/**
+ * @title Success Token Long Short Pair Financial Product Library.
+ * @notice Adds settlement logic to create success token LSPs. A success token pays out a fixed amount of
+ * collateral as a floor, with the remaining amount functioning like an embedded option. The embedded
+ * option can use the payout logic of other financial product libraries, and resemble a covered call, a
+ * KPI option, a range bond, a capped yield dollar, or custom payout logic in a new financial library.
```
chrismaree, line 12: prehaps add the equation that uses the `strikePrice` `basepercentage` and `expiryPrice` to show how the `percentageLongCOllateral` is calculated in the comment here?

nicholaspai (approved):
All my comments are resolved, thanks! LGTM after @chrismaree 's comments are reviewed and after @smb2796 double checks one comment. I'll tag Sean.

chrismaree (approved):
This is great! thanks for updating the comments and all the nits. Really excited to see these used in the wild!

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
```diff
@@ -0,0 +1,91 @@
...
+     */
+    function setLongShortPairParameters(
+        address longShortPair,
+        uint256 strikePrice,
+        uint256 basePercentage
+    ) public nonReentrant() {
+        require(ExpiringContractInterface(longShortPair).expirationTimestamp() != 0, "Invalid LSP address");
+        SuccessTokenLongShortPairParameters memory params = longShortPairParameters[longShortPair];
+        require(params.strikePrice == 0 && params.basePercentage == 0, "Parameters already set");
+        require(strikePrice != 0 && basePercentage != 0, "Base percentage and strike price cannot be set to 0");
```
chrismaree, line 44: One other question: this require will let you set `strikePrice` or `basePercentage` to 0 (assuming the other is not zero). is there anything bad that can happen in this case?

If `strikePrice` is zero then the SuccessToken is always in the money. if `basePercentage` is zero this acts just like an option. Neither of these seem "bad" but just wanted to see if there is anything else to think about in this parameterization context.

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
smb2796: Seems fine to me. Is there even really a solution for this? Also this would be present in the covered call library as well, correct?

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: Hmm, I'm surprised that this would allow one parameter to be set to zero, since I thought the `&&` requires both to be non-zero?

packages/core/contracts/financial-templates/common/financial-product-libraries/long-short-pair-libraries/SuccessTokenLongShortPairFinancialProductLibrary.sol
pemulis: The same issue would be there in covered call, too.

PR #3264, opened 2021-07-30
improve(insured-bridge): Implement relayDeposit in BridgePool

nicholaspai:
**Motivation**

This PR originally was only supposed to implement `relayDeposit` in the `BridgeRouter`. However, after IRL discussions with @chrismaree and @mrice32 we consider it best to re-architect the L1 contracts such that there is one administrative factory contract, `BridgePoolFactory` which can (1) send cross-chain admin transactions to the `OVM_BridgeDepositBox` and (2) deploy new `BridgePool` children which are responsible for housing passive liquidity used to relay deposits.

This PR still only implements on new function, `relayDeposit` but it also performs the above re-architecture.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
 // SPDX-License-Identifier: AGPL-3.0-only
 pragma solidity ^0.8.0;
 
-import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
```
nicholaspai, line 4: @chrismaree Best to think of this as a child contract. It will be deployed by the `BridgePoolFactory` and only stores a reference to the parent. It reads global state variables from the parent via external getter methods.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+import "../../../contracts/oracle/implementation/Constants.sol";
+import "../../../contracts/common/interfaces/AddressWhitelistInterface.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @notice Administrative contract deployed on L1 that has an implicit reference to a DepositBox. This contract is
+ * responsible for deploying new BridgePools, which houses passive liquidity and enables relaying of L2 deposits.
+ * @dev The owner of this contract can call permissioned functions on the L2 DepositBox.
+ */
+contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDomainEnabled {
```
nicholaspai, line 21: @chrismaree This contract used to be called `BridgeRouter`. Consider how the `OVM_BridgeDepositBox` will interact with this contract:
- This contract will be the cross-chain "owner" of the `OVM_BridgeDepositBox` and can send admin calls
- This contract deploys new `BridgePool` contracts.
- `function bridgeTokens(address l2Token, uint32 l1Gas)` will need to know which `BridgePool` to slow-withdraw funds to. This is the major breaking change.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: sounds good.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+    // A Deposit represents a transfer that originated on an L2 DepositBox contract and can be bridged via this contract.
+    enum DepositState { Uninitialized, PendingSlow, PendingInstant, FinalizedSlow, FinalizedInstant }
+    enum DepositType { Slow, Instant }
+
+    // @dev: There is a limit to how many params a struct can contain. Without encapsulating some of the Deposit params
+    // inside the RelayAncillaryDataContents struct, the compiler throws an error related to this issue:
+    // https://github.com/ethereum/solidity/issues/10930.
+    struct RelayAncillaryDataContents {
+        uint256 depositId;
+        // The following params are inferred by the L2 deposit:
```
chrismaree, line 40:
```suggestion
        // The following params are inferred from the L2 deposit:
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        address l2Sender;
```
chrismaree, line 41: should this not be `l2Depositor`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        address recipient;
+        uint256 depositTimestamp;
+        address l1Token;
+        uint256 amount;
+        uint256 maxFeePct;
+        uint256 proposerRewardPct;
+        // Relayer will compute the realized fee considering the amount of liquidity in this contract and the pending
```
chrismaree, line 48: I think this might change marginally based on yesterday's discussion but it's fine for now as this is fee independent.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        uint256 realizedFeePct;
+        address slowRelayer;
+    }
+    struct Deposit {
+        DepositState depositState;
+        DepositType depositType;
+        // A deposit can have both a slow and an instant relayer if a slow relay is "sped up" from slow to instant.
+        // We want to store both addresses for separate payouts.
+        address instantRelayer;
+        // @dev: See @dev note above about why some Deposit params are collapsed into `RelayAncillaryDataContents`.
```
chrismaree, line 59:
```suggestion
        // @dev: See note above about why some Deposit params are collapsed into `RelayAncillaryDataContents`.
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+
+    /**
+     * @notice Called by Relayer to execute Slow relay from L2 to L1, fulfilling a corresponding deposit order.
+     * @dev There can only be one pending Slow relay for a deposit ID.
+     * @dev Caller must have approved this contract to spend the total bond for `l1Token`.
+     * @param depositId Unique ID corresponding to deposit order that caller wants to relay.
+     * @param depositTimestamp Timestamp of Deposit emitted by L2 contract when order was initiated.
+     * @param recipient Address on this network who should receive the relayed deposit.
+     * @param l1Token Token currency to pay recipient. This contract stores a mapping of
+     * `l1Token` to the canonical token currency on the L2 network that was deposited to the Deposit contract.
```
chrismaree, line 108:
```suggestion
     * @param l1Token Token currency to pay recipient. This contract stores a mapping of `l1Token` to the 
     *     canonical token currency on the L2 network that was deposited to the Deposit contract.
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        address recipient,
+        address l2Sender,
+        address l1Token,
+        uint256 amount,
+        // TODO: Allow caller to distinguish between slow and fast fees/rewards.
+        uint256 realizedFeePct,
+        uint256 maxFeePct,
+        uint256 proposerRewardPct
+    ) public {
+        // TODO: Should this perform a check that the amount to relay minus fees and rewards is <= current balance?
```
chrismaree, line 128: prehaps. that would happen if the withdraw contract is totally out of funds

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+                maxFeePct: maxFeePct,
+                proposerRewardPct: proposerRewardPct,
+                realizedFeePct: realizedFeePct,
+                slowRelayer: msg.sender
+            });
+        bytes memory customAncillaryData = getRelayAncillaryData(newRelayData);
+
+        // Request a price for the relay identifier and propose "true" optimistically. These methods will pull the
+        // (proposer reward + proposer bond + final fee) from the caller.
+        // TODO: Figure out what to set for price request timestamp. I don't think we should simply set it to the
```
chrismaree, line 155: I think the L1 block time is fine. that's what `getCurrentTime` on L1 will return and it means that from the proposal time (this method being called) + OO liveness for the bridging action to be verified.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        );
+    }
+
+    function speedUpRelay(uint256 depositId) public {}
+
+    function finalizeRelay(uint256 depositId) public {}
+
+    function settleDisputedRelay(uint256 depositId, address slowRelayer) public {}
+
+    function getRelayAncillaryData(RelayAncillaryDataContents memory _relayData) public view returns (bytes memory) {
```
chrismaree, line 191: very nice.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,350 @@
...
+        // Pull the total bond from the caller so that the OptimisticOracle can subsequently pull it from here.
+        IERC20(l1Token).safeTransferFrom(msg.sender, address(this), totalBond);
+        IERC20(l1Token).safeApprove(address(optimisticOracle), totalBond);
+        optimisticOracle.proposePriceFor(
+            msg.sender,
+            address(this),
+            bridgePoolFactory.getIdentifier(),
+            requestTimestamp,
+            customAncillaryData,
+            1e18
```
chrismaree, line 347: anything non-zero would work as "true". would `1` be better than `1e18`?

chrismaree: I guess that `e1e18` might be easier from a voter perspective though as this is `1` encoded by the standard convention.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+        finder = _finder;
+        require(address(_getCollateralWhitelist()) != address(0), "Invalid finder");
+        _setOptimisticOracleLiveness(_optimisticOracleLiveness);
+        _setProposerBondPct(_proposerBondPct);
+        _setIdentifier(_identifier);
+    }
+
+    // Admin functions
+
+    /**
```
chrismaree, line 57: nit on all these methods: natspec

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+
+    // L2 Deposit contract that originates deposits that can be fulfilled by this contract.
+    address private depositContract;
+
+    mapping(address => L1TokenRelationships) private whitelistedTokens;
+
+    // Set upon construction and can be reset by Owner.
+    uint256 private optimisticOracleLiveness;
+    uint256 private proposerBondPct;
+    bytes32 private identifier;
```
chrismaree, line 33: why make these private and then make the accessors? why not just make them public?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+        L1TokenRelationships storage whitelistedToken = whitelistedTokens[_l1Token];
+        whitelistedToken.l2Token = _l2Token;
+        sendCrossDomainMessage(
+            depositContract,
+            _l2Gas,
+            abi.encodeWithSignature("whitelistToken(address,address)", _l1Token, whitelistedToken.l2Token)
+        );
+
+        // TODO: This contract should deploy a new BridgePool if the address is set to 0x0 at this point.
+        whitelistedToken.bridgePool = _bridgePool;
```
chrismaree, line 123: or, if you want to have this act as a two tx process, just validate the correctness of `bridgePool` address

chrismaree (commented):
Some pretty awesome stuff in this PR! I like this re-structure; better separation of concerns. I left some comments and nits. Please re-request once those are sorted and I'll take another look.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+        bytes32 _identifier
+    ) OVM_CrossDomainEnabled(_crossDomainMessenger) {
+        finder = _finder;
+        require(address(_getCollateralWhitelist()) != address(0), "Invalid finder");
+        _setOptimisticOracleLiveness(_optimisticOracleLiveness);
+        _setProposerBondPct(_proposerBondPct);
+        _setIdentifier(_identifier);
+    }
+
+    // Admin functions
```
chrismaree, line 55: nit: these are not consistent comments with your other contracts.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,189 @@
...
+    function _getCollateralWhitelist() private view returns (AddressWhitelistInterface) {
+        return
+            AddressWhitelistInterface(
+                FinderInterface(finder).getImplementationAddress(OracleInterfaces.CollateralWhitelist)
+            );
+    }
+
+    function _setIdentifier(bytes32 _identifier) private {
+        require(_getIdentifierWhitelist().isIdentifierSupported(_identifier), "Identifier not registered");
+        // TODO: Should we validate this _identifier? Perhaps check that its not 0x?
```
chrismaree, line 173: yup. should defo not be able to set it to blank

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Just a style thing. Before I make sweeping changes do you think it should be `l2Depositor`? Or even `depositor`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Yeah wasn't sure exactly what we had in mind here.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: If the pool balance < amount to relay + proposer reward, then this method will revert. Do you think we should explicitly check `balanceOf(this) < (1 + rewardPct - realizedFeePct) * amount`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Cool that's how I have it now, will remove the comment

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: > I guess that e1e18 might be easier from a voter perspective though as this is 1 encoded by the standard convention.

Yeah that's what I was thinking. I wanted to be consistent with how Admin proposals are currently voted on.

nicholaspai: @mrice32 what do you think here?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: Because the children can't access these methods if they are public, so rather than have redundant state variables and getter methods for the children to read, I chose to just use getters. Does that make sense?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: Yeah I'm not sure what we should do. On one hand I think its clean to deploy a new contract here and keep it to 1 tx, but the gas costs and code complexity are the other issue. WDYT?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: Should I remove `Admin functions`? I added natspec

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: But then again, we check that the identifier is on the Identifier whitelist. So anything that's whitelisted is presumably already sanity checked?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: I guess `depositor` is the cleanest and most descriptive. However, looking on the `DepositBox` we still call it a `sender` so perhaps leave it for now and then we can make a sweeping PR for this kind of thing in a follow-up to make sure everything is consistent over all the contracts and actors in the system.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: fine to ignore for now. this PR does not factor in this kind of fees and it will need to be a separate process anyway

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: I think leave it for now. Prehaps add what you put there as a comment. I think this warrants more thinking and testing so perhaps leave it out the PR

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
chrismaree: why can the children not just call the default public accessors associated with public state variables?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
chrismaree: to be the _most_ consistent perhaps wrap them in the `***` like we do in the other contracts and like you did in the `OVM_CrossDomainEnabled`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
chrismaree: that's a great point. if it's on the whitelist. I it's a fine identifier

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
chrismaree: I dont think it's too much of an issue, assuming you make the deployment method an internal function that gets called if one is not already deployed. feel free to leave it for a follow on PR though.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,356 @@
...
+        require(bridgePoolFactory.getFinder() != address(0), "Invalid bridge pool factory");
+    }
+
+    // Liquidity provider functions
+
+    function deposit(address l1Token, uint256 amount) public {}
+
+    function withdraw(address lpToken, uint256 amount) public {}
+
+    // Relayer functions
```
chrismaree, line 98: nit: add `***` blocks like other contracts.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -21,7 +21,7 @@ const factory__L1StandardBridge = createOptimismEthersFactory("OVM_L1StandardBri
 const factory__L2StandardBridge = createOptimismEthersFactory("OVM_L2StandardBridge", true);
```
chrismaree, line 21: thanks for updating this set of tests 👌

chrismaree (approved):
This code is great! well commented, tested and generally pretty clean. Good work on getting this out so quickly @nicholaspai .

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Implemented a check + unit test that pool balance >= amount-relay + reward

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: I ideally wanted to just have to import the interface of `BridgePoolFactory` into `BridgePool`, but I removed the public getters for now to reduce code clutter.

mrice32: @nicholaspai you can declare public getters in the interface and override it using state variables:

> Public state variables can override external functions if the parameter and return types of the function matches the getter function of the variable:
```
// SPDX-License-Identifier: GPL-3.0
pragma solidity >=0.6.0 <0.9.0;

contract A
{
    function f() external view virtual returns(uint) { return 5; }
}

contract B is A
{
    uint public override f;
}
```
Source: https://docs.soliditylang.org/en/v0.8.6/contracts.html#function-overriding

packages/core/contracts-ovm/insured-bridge/implementation/OVM_CrossDomainEnabled.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    /***************
+     * Constructor *
+     ***************/
+
+    /**
+     * @param _messenger Address of the CrossDomainMessenger on the current layer.
+     */
+    constructor(address _messenger) {
+        messenger = _messenger;
```
mrice32, line 34: Is it guaranteed that this will never change? put differently, is this a special address on the OVM or is it just a normal contract that might be upgraded to a new address later?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -0,0 +1,187 @@
...
+        address _bridgePool,
+        uint32 _l2Gas
+    ) public onlyOwner depositContractSet {
+        require(_getCollateralWhitelist().isOnWhitelist(address(_l1Token)), "Payment token not whitelisted");
+
+        L1TokenRelationships storage whitelistedToken = whitelistedTokens[_l1Token];
+        whitelistedToken.l2Token = _l2Token;
+        sendCrossDomainMessage(
+            depositContract,
+            _l2Gas,
```
mrice32, line 136: What happens if the gas is too low and the OVM transaction fails?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+        uint256 depositId,
+        uint256 depositTimestamp,
+        address recipient,
+        address l2Sender,
+        address l1Token,
+        uint256 amount,
+        // TODO: Allow caller to distinguish between slow and fast fees/rewards.
+        uint256 realizedFeePct,
+        uint256 maxFeePct,
+        uint256 proposerRewardPct
```
mrice32, line 130: Since these will never be > 1e18, could we drop them to a lower bit value to save storage costs?

2^64 goes up to a little above 1e19, so we should be able to make them all uint64. Thoughts?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+     * @param amount Deposited amount.
+     * @param realizedFeePct Computed offchain by caller, considering the amount of available liquidity for the token
+     * currency needed to pay the recipient and the count of pending withdrawals at the `depositTimestamp`. This fee
+     * will be subtracted from the `amount`. If this value is computed incorrectly, then the relay can be disputed.
+     * @param maxFeePct Maximum fee that L2 Depositor can pay. `realizedFee` <= `maxFee`.
+     * @param proposerRewardPct Reward % of deposit amount to pay relayers.
+     */
+    function relayDeposit(
+        uint256 depositId,
+        uint256 depositTimestamp,
```
mrice32, line 122: Thoughts on dropping this to uint64 or uint32 (works until 2106) since that's more than enough space for a timestamp?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+     *     canonical token currency on the L2 network that was deposited to the Deposit contract.
+     * @param amount Deposited amount.
+     * @param realizedFeePct Computed offchain by caller, considering the amount of available liquidity for the token
+     * currency needed to pay the recipient and the count of pending withdrawals at the `depositTimestamp`. This fee
+     * will be subtracted from the `amount`. If this value is computed incorrectly, then the relay can be disputed.
+     * @param maxFeePct Maximum fee that L2 Depositor can pay. `realizedFee` <= `maxFee`.
+     * @param proposerRewardPct Reward % of deposit amount to pay relayers.
+     */
+    function relayDeposit(
+        uint256 depositId,
```
mrice32, line 121: We might also be able to drop this to a lower size depending on how it's created.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
mrice32: Agree that 1e18 is the standard "true" value.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+        // Optimistic oracle will pull proposer reward from passive LP pool.
+        uint256 proposerReward =
+            FixedPoint
+                .Unsigned(proposerRewardPct)
+                .div(FixedPoint.fromUnscaledUint(1))
+                .mul(FixedPoint.Unsigned(amount))
+                .rawValue;
+
+        // Sanity check that pool balance is enough to cover relay amount + proposer reward.
+        require(IERC20(l1Token).balanceOf(address(this)) >= amount + proposerReward, "Insufficient pool balance");
```
mrice32, line 305: Does amount include the fees or exclude the fees?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+
+    // A Deposit represents a transfer that originated on an L2 DepositBox contract and can be bridged via this contract.
+    enum DepositState { Uninitialized, PendingSlow, PendingInstant, FinalizedSlow, FinalizedInstant }
+    enum DepositType { Slow, Instant }
+
+    // @dev: There is a limit to how many params a struct can contain. Without encapsulating some of the Deposit params
+    // inside the RelayAncillaryDataContents struct, the compiler throws an error related to this issue:
+    // https://github.com/ethereum/solidity/issues/10930.
+    struct RelayAncillaryDataContents {
+        uint256 depositId;
```
mrice32, line 39: As mentioned elsewhere, we could probably save some storage costs by making some of these smaller and putting the smaller ones close together to allow solidity to pack storage.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+    struct Deposit {
+        DepositState depositState;
+        DepositType depositType;
+        // A deposit can have both a slow and an instant relayer if a slow relay is "sped up" from slow to instant.
+        // We want to store both addresses for separate payouts.
+        address instantRelayer;
+        // @dev: See note above about why some Deposit params are collapsed into `RelayAncillaryDataContents`.
+        RelayAncillaryDataContents relayData;
+        // Custom ancillary data crafted from `RelayAncillaryDataContents` data.
+        bytes priceRequestAncillaryData;
```
mrice32, line 62: Do we need to store both of these?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -1,12 +1,359 @@
...
+     * @notice Returns ancillary data containing all relevant Relay data that voters can format into UTF8 and use to
+     * determine if the relay is valid.
+     * @param _relayData Contains relevant relay data.
+     * @return bytes New ancillary data that can be decoded into UTF8.
+     */
+    function getRelayAncillaryData(RelayAncillaryDataContents memory _relayData) public view returns (bytes memory) {
+        bytes memory intermediateAncillaryData = "";
+
+        // Add relay data inferred from the original deposit on L2:
+        intermediateAncillaryData = AncillaryData.appendKeyValueUint(
```
mrice32, line 206: Could we hash _most_ of this information and just match it to a hash emitted on the deposit side? The only exception might be the fee that's computed by the relayer.

Just thinking about ways to avoid storing data on mainnet.

mrice32 (approved):
This is really awesome!

The only comments/questions I have are related to gas costs, but these do not need to be solved in this PR -- I just want to get us thinking about that.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_CrossDomainEnabled.sol
nicholaspai: I think this can change after a "regenesis" event but it should mostly be hardcoded

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: Good point. I'll add a todo. we should prepare for that

nicholaspai: Currently, the user could simply call the function again.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Great idea!

nicholaspai: Changed everything to uint64

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: doesn't include fees, but fees can only decrease the amount of funds to pull from the pool so this check is conservative

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: No i don't think so.

nicholaspai: Removed

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Great point. done

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: We definitely could. @chrismaree

PR #3266, opened 2021-07-30
feat(affiliates): 1776 add v2 dev mining based on tvl

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/1776/update-the-dev-mining-script-to-use-tvl-instead-of-tvm-in-the-calculation

**Summary**

Adds tvl based rewarding as "v2" of devmining script. Keeps in place the original dev mining calculation but adds a second version based on collaterals only.  Much of the code is the same between the original version but some modifications.

Once merged the default unless specified otherwise will be to run v2. This should not be merged until after aug 2nd, since this is not needed until after.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/1776/update-the-dev-mining-script-to-use-tvl-instead-of-tvm-in-the-calculation

### DIFF
--- a/packages/affiliates/apps/DevMiningRewards.js
+++ b/packages/affiliates/apps/DevMiningRewards.js
@@ -6,18 +6,22 @@ const assert = require("assert");
 const { getAbi } = require("@uma/core");
 const { BigQuery } = require("@google-cloud/bigquery");
 const Promise = require("bluebird");
+const Web3 = require("web3");
 
 const { DevMining } = require("../libs/affiliates");
 const { Emp } = require("../libs/contracts");
 const Queries = require("../libs/bigquery");
 const Coingecko = require("../libs/coingecko");
 const SynthPrices = require("../libs/synthPrices");
-const { getWeb3 } = require("@uma/common");
 
 const { makeUnixPipe } = require("../libs/affiliates/utils");
 
-// This is the main function which configures all data sources for the calculation.
-const App = (env) => async (params) => {
+// dont know why, but the common getWeb3 is causing contract calls to fail
+function getWeb3() {
+  return new Web3(process.env.CUSTOM_NODE_URL);
+}
+
+const v1 = async (env, params, DevMining) => {
   const web3 = getWeb3();
   const { config } = params;
   assert(config, "requires config object on params");
@@ -27,7 +31,6 @@ const App = (env) => async (params) => {
   assert(endTime, "requires endTime");
   assert(totalRewards, "requires totalRewards");
 
-  const emp = Emp({ web3 });
   const client = new BigQuery();
   const queries = Queries({ client });
   const coingecko = Coingecko();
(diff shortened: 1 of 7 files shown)


packages/affiliates/apps/DevMiningRewards.js
```diff
@@ -101,6 +105,94 @@ const App = (env) => async (params) => {
   };
 };
 
+const v2 = async (env, params, DevMining) => {
```
daywiss, line 108: this is in the dev mining application. There are 2 blocks of code that configure things for v1 (tvm based dev mining) and now v2 (tvl based dev mining). This new block is almost a copy and paste of the v1 block, except it removes any mention of synthetic tokens.

packages/affiliates/libs/affiliates/devmining_v2.js
```diff
@@ -0,0 +1,250 @@
+const highland = require("highland");
```
daywiss, line 1: this is the devmining v2 business logic. Its almost completely identical to v1, but removes any calculations based on synthetic tokens, and just converts collateral to USD

packages/affiliates/libs/affiliates/devmining_v2.js
```diff
@@ -0,0 +1,250 @@
...
+
+          // if EMP was expired this block, dont include it in further calculations.
+          if (isExpired) {
+            return result;
+          }
+          const [, closestCollateralPrice] = collateralTokenPrices[empIndex].closest(timestamp);
+          const totalCollateral = Object.values(collateral).reduce((result, value) => {
+            return result.add(toBN(value));
+          }, toBN("0"));
+          const value = calculateValue(totalCollateral, closestCollateralPrice, collateralTokenDecimals[empIndex]);
```
daywiss, line 133: the bulk of the logic was just changed here, compared to v1. Calculate value is now just collateral value, and synthetics minted are ignored. Everything else for the most part is the same

packages/affiliates/libs/contracts.js
```diff
@@ -129,6 +131,22 @@ function Emp({ abi = getAbi("ExpiringMultiParty"), web3 } = {}) {
...
+// amount: amount of token in wei
+// price: price as a float
+// decimals: decimal value for the token
+// returns value in wei 18 decimals
+function calculateValue(amount, price, decimals) {
+  price = toBN(toWei(price.toString()));
+  amount = toBN(amount);
+  return fromWei(amount.mul(price), decimals);
+}
+
```
daywiss, line 149: a couple utility functions added. web3s `fromWei` is really wierd and does not work the same way as ethers, it requires you pass in a string like `wei` or `gwei` which is really useless, so i made a utility

packages/affiliates/scripts/basic-contract.js
```diff
@@ -0,0 +1,14 @@
+require("dotenv").config();
```
daywiss, line 1: can probably ignore this, or could be deleted, but could not run the app using `getWeb3`. would get `UnhandledPromiseRejectionWarning: Error: Invalid JSON RPC response: "" ` which went away once i initialized my own web3.

packages/affiliates/test/mocha/devmining-v2.js
```diff
@@ -0,0 +1,321 @@
+const { DevMining } = require("../../libs/affiliates");
```
daywiss, line 1: this copies the tests from v1, changes some values, but does a good job of testing scenarios

mrice32 (approved):
LGTM! This isn't super dry, but I'm not sure it's worth doing the work of trying to make it dry since this is not something we expect to be iterating on a bunch.

chrismaree (approved):

daywiss:
> LGTM! This isn't super dry, but I'm not sure it's worth doing the work of trying to make it dry since this is not something we expect to be iterating on a bunch.

Thanks, yea i just didnt see the value in spending the extra time to try and reuse code, though you are right theres definatley a lot of duplicate code.

PR #3267, opened 2021-07-30
fix: fix dependabot alert around googleapis version

mrice32:
**Motivation**

There's a dep warning on the repo around the googleapis dependency used in affiliates. It doesn't actually affect us because our true version specified in the yarn.lock file is higher than the required one, but we should explicitly require this in our package.json.

**Summary**

Explicitly require 39.1.0 in the package.json

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A


daywiss (approved):
seems fine

chrismaree (approved):

PR #3269, opened 2021-08-03
feat(insured-bridge): whitelist collateral cross-chain end to end tests

chrismaree:
**Motivation**

As part of the insured bridge, we require end-to-end tests. This PR adds the first UMA contract end-to-end test to show the act of whitelisting tokens over the bridge.

**Summary**

A few other changes were made. in particular, an update to how whitelisting is done was added to further simplify the process.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/common/src/HardhatConfig.js
```diff
@@ -76,7 +76,7 @@ function getHardhatConfig(configOverrides, workingDir = "./", includeTruffle = t
         gasPrice: 0,
         // We use custom logic to only compile contracts within the listed directories, as opposed to choosing which
         // ones to ignore, because there are more contracts to ignore than to include.
-        compileWhitelist: ["insured-bridge/implementation/OVM_BridgeDepositBox.sol"],
```
chrismaree, line 79: I change the whitelist again. my current thinking is we should just have to define the contract names without any paths. wdyt @nicholaspai ?

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -1,14 +1,10 @@
 const { getAllFilesInPath } = require("@uma/common");
 
-const path = require("path");
```
chrismaree, line 3: this is even simpler now. simply looks for matches on the array of whitelisted contract names.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -47,67 +60,132 @@ describe("L1 <> L2 Deposit and Withdrawal", () => {
...
+
+    await l1Erc20.deployTransaction.wait();
 
-    await L1_ERC20.addMember(1, l1Wallet.address);
-    await L1_ERC20.mint(l1Wallet.address, 1234);
+    await l1Erc20.addMember(1, l1Wallet.address);
+    await l1Erc20.mint(l1Wallet.address, 1234);
 
     // Deploy the paired ERC20 token to L2. This takes in the address of the L2 bridge and the associated L1 token.
-    L2_ERC20 = await factory__L2_ERC20
```
chrismaree, line 57: updated some of these variables to be consistent with the rest of UMA unit test naming convention.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -47,67 +60,132 @@ describe("L1 <> L2 Deposit and Withdrawal", () => {
...
+      assert.equal(receipt.logs[1].address, l1StandardBridge.address);
 
       // Validate balances on L1 have been incremented accordingly.
-      assert.equal((await L1_ERC20.balanceOf(l1Wallet.address)).toString(), "1234");
-      assert.equal((await L2_ERC20.balanceOf(l2Wallet.address)).toString(), "0");
+      assert.equal((await l1Erc20.balanceOf(l1Wallet.address)).toString(), "1234");
+      assert.equal((await l2Erc20.balanceOf(l2Wallet.address)).toString(), "0");
+    });
+  });
+  describe("Insured bridge functionality", async () => {
```
chrismaree, line 127: the tests below show a full L1->L2 transaction flow, in the process of whitelisting collateral. As we have more functionality that can be tested e2e we can update and expand these tests

packages/core/test/insured-bridge/e2e/helpers/OptimismConstants.js
```diff
@@ -1,4 +1,10 @@
 // Pre-loaded wallet from Optimism docker containers with 10k eth on both L1 and L2.
 const DEFAULT_ADMIN_KEY = "0xac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80";
 
-module.exports = { DEFAULT_ADMIN_KEY };
+// To correctly deploy on optimism, some settings need to be set.
+const OPTIMISM_GAS_OPTS = { gasLimit: 12_500_000, gasPrice: 0 };
+
+// Address of the proxied cross-domain messenger deployed in the default Optimism container. Not exported by the package.
+const PROXY__OVM_L1_CROSS_DOMAIN_MESSENGER = "0x59b670e9fa9d0a427751af201d676719a970857b";
```
chrismaree, line 8: getting this address right was a bit of a pain. it needs to be set like this to work correctly. It also means that we cant use the `Lib_PredeployAddresses` in the withdraw box.

packages/common/src/HardhatConfig.js
nicholaspai: yeah I like that more

packages/common/src/hardhat/tasks/compile.js
```diff
@@ -29,13 +25,9 @@ internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network
   // Build absolute path for all directories on user-specified whitelist.
   const whitelist = config.networks[network.name].compileWhitelist;
   if (whitelist && Array.isArray(whitelist)) {
-    const whitelistDirs = whitelist.map((x) => {
-      return path.resolve(CONTRACTS_DIR, x);
-    });
-
     filePaths = filePaths.filter((filePath) => {
-      for (let whitelistedDir of whitelistDirs) {
```
nicholaspai, line 37: OOC, are we back to square one on this file? haha

packages/core/.mocharc.js
```diff
@@ -29,6 +29,6 @@ if (!process.argv.includes("--e2e")) {
     extension: ["js"],
     watchExtensions: ["js"],
     spec: endToEndTests, // mocha will run this sent of test files defined in this array.
-    timeout: 80000,
+    timeout: 100000,
```
nicholaspai, line 32: Why this change?

chrismaree: It's not actually required really. 80k would be fine as well. I changed it while having some issues with timeout on the bridge but the diff between 80k and 100k is really important. so no real reason to have change it

packages/common/src/hardhat/tasks/compile.js
chrismaree: it's a fair amount simpler than the original I think.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeRouter.sol
```diff
@@ -157,11 +157,11 @@ contract BridgeRouter is OVM_CrossDomainEnabled {
         whitelistedToken.bridgePool = _bridgePool;
         // We want to prevent any situation where a token mapping is whitelisted on this contract but not on the
         // corresponding L2 contract.
-        require(depositContract != address(0), "Deposit contract not set");
+        require(depositContract != address(0), "Deposit contract not set"); //TODO: turn this into a modifier
```
nicholaspai, line 160: Can you remove changes to this file and just comment on the corresponding line in #3264 so as to avoid merge conflicts? Note that this contract is now named `BridgePoolFactory`

packages/core/contracts-ovm/insured-bridge/implementation/BridgeRouter.sol
```diff
@@ -157,11 +157,11 @@ contract BridgeRouter is OVM_CrossDomainEnabled {
...
         sendCrossDomainMessage(
             depositContract,
             _l2Gas,
-            abi.encodeWithSignature("whitelistToken(address,address)", _l1Token, whitelistedToken.l2Token)
+            abi.encodeWithSignature("whitelistToken(address,address)", _l1Token, _l2Token)
```
nicholaspai, line 164: Same comment as above, but why this change? Is it gas savings? Code style? I guess it doesn't matter and _l2Token is slightly nicer to read. My thinking was that reading from the new struct `whitelistedToken` was slightly more correct but I don't think it matters. It's either set correctly or not.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeRouter.sol
chrismaree: done

packages/core/contracts-ovm/insured-bridge/implementation/BridgeRouter.sol
chrismaree: they are equivalent honestly. I did it just because it's a bit cleaner and I suspect will use a small amount less gas. In any case I've reverted it for your conflict request.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -47,67 +60,131 @@ describe("L1 <> L2 Deposit and Withdrawal", () => {
...
+    beforeEach("Deploy and configure insured bridge contracts on appropriate networks", async () => {
+      let optimisticOracleLiveness = 7200;
+      let minimumBridgingDelay = 60;
+
+      // Deploy UMA helper contracts
+      l1Finder = await factory__L1_Finder.connect(l1Wallet).deploy();
+      l2Timer = await factory__L2_Timer.connect(l2Wallet).deploy(OPTIMISM_GAS_OPTS);
+      await l2Timer.deployTransaction.wait();
+
+      // TODO: when doing full integration tests we will need the OO as well. For now show that we can do basic bridging.
```
nicholaspai, line 139: can also borrow code from #3264 so better to wait on this code.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -47,67 +60,131 @@ describe("L1 <> L2 Deposit and Withdrawal", () => {
...
+
+      // Bridging infrastructure and initialization. Note we use the l1 proxy cross-domain messenger for the routers
+      // _crossDomainMessenger. This is done to mimic the production setup which routes transactions through this proxy.
+      l1BridgeRouter = await factory__L1_BridgeRouter
+        .connect(l1Wallet)
+        .deploy(l1Finder.address, PROXY__OVM_L1_CROSS_DOMAIN_MESSENGER, l1Wallet.address, optimisticOracleLiveness);
```
nicholaspai, line 145: Nice! The merge conflicts on this file will be minimal. Just changes to the contract name and the constructor signature.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
chrismaree: 100%. that'll be a separate set of e2e tests.

packages/core/test/insured-bridge/e2e/helpers/ArtifactsHelper.js
```diff
@@ -18,7 +18,7 @@ function getOptimismArtifact(contractName, ovm = false) {
 
 function getLocalArtifact(contractName, ovm = false) {
   const coreContractsPath = findPathToRootOfPackage("@uma/core");
-  const artifactsPath = ovm ? `${coreContractsPath}/artifacts-ovm` : `${coreContractsPath}/artifacts`;
+  const artifactsPath = ovm ? `${coreContractsPath}artifacts-ovm` : `${coreContractsPath}artifacts`;
```
nicholaspai, line 21: why isn't the slash needed? I would have guessed otherwise

packages/core/test/insured-bridge/e2e/helpers/OptimismConstants.js
nicholaspai: Curious where you got this address from? And do you think we need to hardcode this address in the `BridgePoolFactory` as the `_crossDomainMessenger`?

nicholaspai (commented):
AWESOME to see this is such a small PR. Left some comments, re request when you're done reviewing them all.

packages/core/.mocharc.js
nicholaspai: nit: Ok I"m not even sure what this value does maybe add a small comment.

nicholaspai (approved):

packages/core/test/insured-bridge/e2e/helpers/ArtifactsHelper.js
chrismaree: for some reason between when I set this up and now the call to `findPathToRootOfPackage` now contains a slash. The only reasoning to why this happened is that a package was updated under the hood and impacted the behavior 🤷 . Honestly not really sure how it worked before if that assumption is not true.

chrismaree: Actually, I dont think we were importing `ovm` artifacts before (only non-ovm artifacts locally...but even then the slash change does not compute

packages/core/test/insured-bridge/e2e/helpers/OptimismConstants.js
chrismaree: I got it from one of the optimism example tests. Also, if you are running the optimism docker containers and go to http://localhost:8080/addresses.json you will see this:
```
{
  "AddressManager": "0x5FbDB2315678afecb367f032d93F642f64180aa3",
  "OVM_CanonicalTransactionChain": "0xA51c1fc2f0D1a1b8494Ed1FE312d7C3a78Ed91C0",
  "OVM_ChainStorageContainer-CTC-batches": "0x0165878A594ca255338adfa4d48449f69242Eb8F",
  "OVM_ChainStorageContainer-CTC-queue": "0x2279B7A0a67DB372996a5FaB50D91eAA73d2eBe6",
  "OVM_ChainStorageContainer-SCC-batches": "0x610178dA211FEF7D417bC0e6FeD39F05609AD788",
  "OVM_ExecutionManager": "0xa85233C63b9Ee964Add6F2cffe00Fd84eb32338f",
  "OVM_FraudVerifier": "0x7a2088a1bFc9d81c55368AE168C2C02570cB814F",
  "OVM_L1CrossDomainMessenger": "0x68B1D87F95878fE05B998F19b66F4baba5De1aed",
  "OVM_L1MultiMessageRelayer": "0xa82fF9aFd8f496c3d6ac40E2a0F282E47488CFc9",
  "OVM_SafetyChecker": "0x84eA74d481Ee0A5332c457a4d796187F6Ba67fEB",
  "OVM_StateCommitmentChain": "0x9A676e781A523b5d0C0e43731313A708CB607508",
  "OVM_StateManagerFactory": "0xc5a5C42992dECbae36851359345FE25997F5C42d",
  "OVM_StateTransitionerFactory": "0xE6E340D132b5f46d1e472DebcD681B2aBc16e57E",
  "Proxy__OVM_L1CrossDomainMessenger": "0x59b670e9fA9D0A427751Af201D676719a970857b",
  "Proxy__OVM_L1StandardBridge": "0x851356ae760d987E095750cCeb3bC6014560891C",
  "OVM_BondManager": "0x959922bE3CAee4b8Cd9a407cc3ac1C251C2007B1",
  "OVM_Sequencer": "0x70997970C51812dc3A010C7d01b50e0d17dc79C8",
  "Deployer": "0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266"
}
```

in here you can find that address!

chrismaree: also we dont wnat to hardcode it in the `BridgePoolFactory` as it will be different on mainnet. On mainnet it's here: https://etherscan.io/address/0x25ace71c97B33Cc4729CF772ae268934F7ab5fA1

you can find the details of this and other contracts here: https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/deployments/README.md

PR #3275, opened 2021-08-03
feat(insured-bridge): Implement remaining crossdomain admin functions in BridgeAdmin

nicholaspai:
**Motivation**
- Implements remaining crossdomain admin functions in BridgePoolFactory
- Rename `BridgePoolFactory` to `BridgeAdmin`
- Updated tests

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -48,8 +48,8 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
      *  OVM DEPOSIT BOX DATA STRUCTURES  *
      *************************************/
 
-    // Address of the L1 bridge router that acts as the owner of this Bridge deposit box.
-    address public bridgeRouter;
+    // Address of the L1 contract that acts as the owner of this Bridge deposit box.
+    address public bridgeAdmin;
```
nicholaspai, line 52: @chrismaree I felt this name was a bit more accurate and better than `bridgePoolFactory`

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -118,44 +119,50 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
...
      */
-    function whitelistToken(address l1Token, address l2Token) public onlyFromCrossDomainAccount(bridgeRouter) {
+    function whitelistToken(
+        address l1Token,
+        address l2Token,
+        address l1BridgePool
+    ) public onlyFromCrossDomainAccount(bridgeAdmin) {
         whitelistedTokens[l2Token] = L2TokenRelationships({
             l1Token: l1Token,
+            l1BridgePool: l1BridgePool,
```
nicholaspai, line 153: whitelisted token mapping now needs to remember each pool associated with each L1 token.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -221,7 +228,7 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
 
         StandardBridgeLike(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo(
             l2Token, // _l2Token. Address of the L2 token to bridge over.
-            bridgeRouter, // _to. Withdraw, over the bridge, to the l1Withdraw contract.
+            whitelistedTokens[l2Token].l1BridgePool, // _to. Withdraw, over the bridge, to the l1 withdraw contract.
```
nicholaspai, line 231: @chrismaree does this slight change affect any e2e tests? i don't think so but want extra eyes on it

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -47,14 +55,17 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
         _;
     }
 
+    // TODO: Switch to hardcoded OVM_L1CrossDomainMessenger: https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/deployments/README.md
```
chrismaree, line 58: I _think_ we might want to leave this parameterizable (and potentially changeable) within the contract. The reason being is right now they use a proxy for this on mainnet which could potentially change. perhaps we want to move to the underlying contract (not proxy) or there is a future pattern that requires a new address. 

Also, the proxy used on mainnet is _not_ exported anywhere within any const libs or anything like that so it becomes a bit difficult to hardcode it.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -7,21 +7,25 @@ pragma solidity ^0.8.0;
...
  * @notice Administrative contract deployed on L1 that has an implicit reference to a DepositBox. This contract is
  * responsible for deploying new BridgePools, which houses passive liquidity and enables relaying of L2 deposits.
  * @dev The owner of this contract can call permissioned functions on the L2 DepositBox.
  */
-contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDomainEnabled {
+contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, Testable, OVM_CrossDomainEnabled {
     // Finder used to point to latest OptimisticOracle and other DVM contracts.
     address public override finder;
 
+    address public timer;
```
chrismaree, line 27: you dont need to store the timer address, do you? why would you need this and cant access it from the inherited `Testable`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -107,43 +118,97 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
...
     }
 
+    /**************************************************
+     *        CROSSDOMAIN ADMIN FUNCTIONS             *
+     **************************************************/
+
+    /**
+     * @notice Set this contract as the admin address in the L2 Deposit contract.
+     * @dev Only callable by the current owner.
+     * @param _l2Gas Gas limit to set for relayed message on L2
```
chrismaree, line 128: I feel that this param (used in a few other places as well) is pretty fragile. I've tested out what happens if this is too low or too high and it can fail pretty badly. If it's too low (you set it to 0 for example) then the L1 tx will go through but nothing happens on L2 (gas too low). if it's too high (larger than 12.5m, optimism gas limit) it reverts on L1 due to the cross-domain messenger.

It might be worthwhile to set a variable such as 10m that we always use for the l2Gas in all calls to prevent the caller from messing this up. wdyt?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -107,43 +118,97 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
...
-     * @notice Privileged account can associate a whitelisted token with its linked token address on L2 and its
-     * BridgePool address on this network. The linked L2 token can thereafter be deposited into the Deposit contract
-     * on L2 and relayed via the BridgePool contract.
+     * @notice Owner can pause/unpause L2 deposits for all tokens.
+     * @dev Only callable by the current owner. Will set the same setting in the L2 Deposit contract via the cross
+     * domain messenger.
+     * @param _depositsEnabled bool to set if the deposit box should accept/reject deposits.
+     * @param _l2Gas Gas limit to set for relayed message on L2
+     */
+    function setEnableDeposits(bool _depositsEnabled, uint32 _l2Gas) public onlyOwner depositContractSet {
```
chrismaree, line 161: this is a perfect reason why we can use the `toggle` method on L2 as the L1 contract (this contract) does not know if it's enabled or disabled. we _could_ store a bool in this contract that represents the enabled/disabled status of the L2 contract and only change it when this method is called and could then have that pattern? otherwise, are you ok with this setup?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -107,43 +118,97 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
...
-        address _bridgePool,
         uint32 _l2Gas
     ) public onlyOwner depositContractSet {
         require(_getCollateralWhitelist().isOnWhitelist(address(_l1Token)), "Payment token not whitelisted");
 
         L1TokenRelationships storage whitelistedToken = whitelistedTokens[_l1Token];
         whitelistedToken.l2Token = _l2Token;
+        // If token mapping isn't associated with a BridgePool, then deploy a new one. Otherwise just update the mapping.
+        if (whitelistedToken.bridgePool == address(0)) {
+            BridgePool newPool = new BridgePool(address(this), timer);
```
chrismaree, line 191: this will clearly need to update once we've got token-specific info in the `BridgePool`. Should also consider what happens if you re-whitelist an existing token. What do you do with the existing `BridgePool`? is there some process that should be followed to update it?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: cool. do you want to rename the `BridgePoolFactory` to `BridgeAdmin` as well?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: great point. this is key when actually pulling out the funds onto L1.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: there are no e2e tests for this part just yet :) as long as we've got the unit test for this updated (which you did) we good to go

chrismaree (approved):
Nice PR! good work on fleshing out this remaining functionality. I left some comments and nits but overall excellent work 🚀

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: I think `BridgePoolFactory` is more accurate since its a factory, but from the DepositBox's perspective its the admin. What do you think?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: True

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: hmmm yeah that's reasonable, we should actually ask the Optimism team. I see that [makerdao](https://github.com/makerdao/optimism-dai-bridge/blob/master/contracts/l1/L1DAITokenBridge.sol#L84) leaves this as a parameter so i'm curious what best practices are.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: hmmm yes I think that's a good call. I can make that change.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
nicholaspai: > this will clearly need to update once we've got token-specific info in the BridgePool.

what kind of token-specific info?

> Should also consider what happens if you re-whitelist an existing token. What do you do with the existing BridgePool? is there some process that should be followed to update it?

Yeah, hm this is a good point. For now should we just always redeploy a new pool whenever the mapping is changed?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
```diff
@@ -107,43 +118,97 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
...
+    /**************************************************
+     *        CROSSDOMAIN ADMIN FUNCTIONS             *
+     **************************************************/
+
+    /**
+     * @notice Set this contract as the admin address in the L2 Deposit contract.
+     * @dev Only callable by the current owner.
+     * @param _l2Gas Gas limit to set for relayed message on L2
+     */
+    function setBridgeAdmin(uint32 _l2Gas) public onlyOwner depositContractSet {
```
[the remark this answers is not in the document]
nicholaspai: yeah that's a good point, will change.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: I see you updated this. looks good.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -118,46 +118,58 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
...
+     * @notice L1 owner can enable/disable deposits for a whitelisted tokens.
+     * @dev Only callable by the existing bridgeAdmin via the optimism cross domain messenger.
+     * @param _l2Token address of L2 token to enable/disable deposits for.
      * @param _depositsEnabled bool to set if the deposit box should accept/reject deposits.
      */
-    function setEnableDeposits(bool _depositsEnabled) public onlyFromCrossDomainAccount(bridgeRouter) {
-        depositsEnabled = _depositsEnabled;
-        emit DepositsEnabled(_depositsEnabled);
+    function setEnableDeposits(address _l2Token, bool _depositsEnabled) public onlyFromCrossDomainAccount(bridgeAdmin) {
+        L2TokenRelationships storage _token = whitelistedTokens[_l2Token];
```
chrismaree, line 170: do you have to declare the `storage` object like this? could you not just do:
```
whitelistedTokens[_l2Token].depositsEnabled = _depositsEnabled;
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
```diff
@@ -107,15 +114,67 @@ contract BridgePoolFactory is BridgePoolFactoryInterface, Ownable, OVM_CrossDoma
...
-     * BridgePool address on this network. The linked L2 token can thereafter be deposited into the Deposit contract
-     * on L2 and relayed via the BridgePool contract.
+     * @notice Owner can pause/unpause L2 deposits for a tokens.
+     * @dev Only callable by the current owner. Will set the same setting in the L2 Deposit contract via the cross
+     * domain messenger.
+     * @param _l2Token address of L2 token to enable/disable deposits for.
+     * @param _depositsEnabled bool to set if the deposit box should accept/reject deposits.
+     * @param _l2Gas Gas limit to set for relayed message on L2.
+     */
+    function setEnableDeposits(
```
chrismaree, line 157: glad to see this method changed to enable/disable a particular token.

chrismaree (approved):
PR Looks great. I'm happy with the re-naming as well. `BridgeAdmin` is more descriptive. Leaving the deployment of new pools up to the admin also seems reasonable and gives us more flexibility.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
nicholaspai: Good call, done

packages/core/contracts-ovm/insured-bridge/implementation/BridgePoolFactory.sol
chrismaree: I think we can leave it as a param for now. will ask the Optimism team about the best practices.

PR #3276, opened 2021-08-03
feat(insured-bridge): Change internal relay data structures

nicholaspai:

**Motivation**

- Deposits are mapped to relay data so that duplicate relays can't exist for a single L2 deposit. This prevents edge cases where relays with different params (e.g. `proposerRewardPct`, `realizedFeePct`) can't coexist.
- Relay price request ancillary data is mapped to deposit data so that the OptimisticOracle can delete pending relays.
- Deposit data is no longer stored in the contract, but is instead hashed to save gas. Deposit data is emitted in the `DepositRelayed` event for clients.
- This PR also renames "Deposit" language to "Relay", which enforces the idea that a "Relay" is an attempt to finalize a "Deposit", and a "Deposit" is an L2 action.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -316,40 +316,58 @@ describe("BridgePool", () => {
...
+          relayData.amount,
+          relayData.maxFeePct,
+          relayData.proposerRewardPct,
+          relayData.realizedFeePct,
+          relayData.slowRelayer,
+        ]
+      );
+      const expectedRelayHash = web3.utils.soliditySha3(abiEncodedDataPacked);
+      const relayHash = await bridgePool.methods.getRelayHash(relayData).call();
+      assert.equal(
```
nicholaspai, line 355: This test is worth looking at for how clients can use data hashed inside of the relay hash

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -47,21 +47,25 @@ contract BridgePool is Testable {
...
-        RelayAncillaryDataContents relayData;
+        // Store unique hash derived from all of the data neccessary to construct this deposit.
+        bytes32 relayHash;
     }
-    // Associates each deposit with a unique ID.
-    mapping(uint64 => Deposit) public deposits;
+    // Associates each deposit with a unique ancillary data derived from its constituent data. We need to key by the
+    // ancillary data so that the OptimisticOracle can locate deposits on callbacks using only price requests' ancillary
+    // data
+    mapping(bytes => Deposit) public deposits;
```
mrice32, line 64: It's effectively the same thing, but why not just use the relayHash directly as the key (bytes32), which would mean we don't need to reconstruct the hash to do the lookup.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -136,19 +156,7 @@ contract BridgePool is Testable {
...
-            depositTimestamp: depositTimestamp,
-            l1Token: l1Token,
-            amount: amount,
-            maxFeePct: maxFeePct,
-            proposerRewardPct: proposerRewardPct,
-            realizedFeePct: realizedFeePct,
-            slowRelayer: msg.sender
-        });
-        bytes memory priceRequestAncillaryData = getRelayAncillaryData(newDeposit.relayData);
+        newDeposit.relayHash = getRelayHash(relayData);
```
mrice32, line 159: Why do we need to store this if it's passed in to every method?

Also, it might make things easier if we add this hash to the DepositRelayed event as an indexed parameter so it's easy to find and associate to the other data offchain.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -176,11 +184,63 @@ contract BridgePool is Testable {
...
+    function priceDisputed(
+        bytes32 identifier,
+        uint256 timestamp,
+        bytes memory ancillaryData,
+        uint256 refund
+    ) public {
+        Deposit storage pendingDeposit = deposits[ancillaryData];
+        // TODO: Implement logic. For now, placeholder code demonstrates how relay can be migrated from pending to
+        // disputed mapping.
+        Deposit storage disputedDeposit = disputedDeposits[ancillaryData][msg.sender];
```
mrice32, line 211: I may be missing something, but would it work just to leave this in deposits and just store the disputer in the existing struct?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -125,9 +129,25 @@ contract BridgePool is Testable {
...
+                recipient: recipient,
+                depositTimestamp: depositTimestamp,
+                l1Token: l1Token,
+                amount: amount,
+                maxFeePct: maxFeePct,
+                proposerRewardPct: proposerRewardPct,
+                realizedFeePct: realizedFeePct,
+                slowRelayer: msg.sender
+            });
+        bytes memory priceRequestAncillaryData = getRelayAncillaryData(relayData);
```
mrice32, line 147: What do you think the pros/cons are of passing in the hash in as the only param in ancillary data and asking the voters to look up the hash to determine the validity of the data?

mrice32 (commented):
Looks awesome! A few questions/comments, but this is really cool

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: because of the `OptimisticOracle.priceDisputed()` [callback](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L411). When a relay price request is disputed, we'd like for the OO to be able to callback into this method and move the pending relay to the pending dispute mapping but this callback is only aware of the price request's ancillary data, not the relay hash in this context.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Great catch, I removed this param and added the ancillary data to the DepositRelayed event

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I actually changed this logic a bit to remove all of the hashing logic. Since we're key-ing by ancillary data, we only need to pass the ancillary data around.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Hm, doesn't that make it trickier to enable another relayer to come in after a relay has been disputed? cc @chrismaree

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I think this is a reasonable feature, just adds more client complexity in exchange for gas efficiency. @chrismaree wdyt?

chrismaree: ye, I agree the hash is better. the OO callback on `priceDisputed` returns the ancillary data from the price request which you can hash to get the key for this on the dispute side. the event emitted called `DepositRelayed` contains everything to compute the hash so no loss or big complexity I don't think.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
```diff
@@ -118,6 +118,11 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, OVM_CrossDomainEnabled {
      *        CROSSDOMAIN ADMIN FUNCTIONS             *
      **************************************************/
 
+    // TODO: In following functions, we need to consider two things concerning asynchronous cross-domain messaging:
```
chrismaree, line 121: +1

For now I think both of these can be left up to the admin (they put their own gas and deal with reverting on L2). As long as users don't have to specify this ever it should be fine.

nicholaspai: There are no cross-domain functions that users can call right?

chrismaree (approved):
This is a great change. pretty simple to reconstruct the hash from event data, making the tradeoff minimal.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
chrismaree: nope. not rn

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -47,32 +46,36 @@ contract BridgePool is Testable {
...
-    // @dev The mapping is `depositId-->disputer-->Deposit`
-    mapping(uint64 => mapping(address => Deposit)) public disputedDeposits;
+    // Associates each relay with a unique ancillary data hash derived from its constituent data. We need to key by the
+    // ancillary data so that the OptimisticOracle can locate relays on callbacks using only price requests' ancillary
+    // data
+    mapping(bytes32 => Relay) public relays;
+    // If a relay is disputed, it is removed from the `relays` mapping and added to the `disputedRelays` mapping.
+    // There can only be one disputed relay per relayer for each relay ancillary data.
+    // @dev The mapping is `relayAncillaryData-->disputer-->Relay`
+    mapping(bytes32 => mapping(address => Relay)) public disputedRelays;
```
mrice32, line 65: What happens if someone instant-relays, they get wrongfully disputed, and then someone else does the same relay? How do we prevent the pool from paying out twice?

I think we may need to have some sort of persistent mapping to ensure that if the deposit details match across the same request that it automatically pays out to the same instant-relayer.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -125,70 +128,109 @@ contract BridgePool is Testable {
...
             realizedFeePct,
-            bridgeAdmin.depositContract()
+            keccak256(getRelayAncillaryData(relayData))
         );
     }
 
-    function speedUpRelay(uint256 depositId) public {}
+    function speedUpRelay(RelayDataContents memory _relayData) public {
+        // Alternatively, force client to hash off-chain and pass `bytes32 _relayAncillaryDataHash` as param.
+        Relay storage relay = relays[keccak256(getRelayAncillaryData(_relayData))];
```
mrice32, line 192: Seems a little wasteful to hash the entire ancillary data each time since it means we have to construct the ancillary data on each call (lots of string concats). Any reason we can't just hash the values as we were before and only construct the ancillary data when calling to the OO? Or are there disadvantages to doing that?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -47,32 +46,36 @@ contract BridgePool is Testable {
...
-    // If a deposit is disputed, it is removed from the `deposits` mapping and added to the `disputedDeposits` mapping.
-    // There can only be one disputed deposit per relayer for each deposit ID.
-    // @dev The mapping is `depositId-->disputer-->Deposit`
-    mapping(uint64 => mapping(address => Deposit)) public disputedDeposits;
+    // Associates each relay with a unique ancillary data hash derived from its constituent data. We need to key by the
+    // ancillary data so that the OptimisticOracle can locate relays on callbacks using only price requests' ancillary
+    // data
+    mapping(bytes32 => Relay) public relays;
+    // If a relay is disputed, it is removed from the `relays` mapping and added to the `disputedRelays` mapping.
+    // There can only be one disputed relay per relayer for each relay ancillary data.
```
mrice32, line 63: This comment is a little confusing -- here it says that the relayer is unique but the comment on the next line seems to indicate that the disputer is unique.

mrice32 (approved):
LGTM! A few minor comments, but I think this general structure of using hashes as keys is 👌.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: > What happens if someone instant-relays, they get wrongfully disputed, and then someone else does the same relay? How do we prevent the pool from paying out twice?

Really good catch. I think to mitigate that, we can set `newRelay.instantRelayer == disputedRelay.instantRelayer` if one exists. 

@chrismaree tagging for visibility

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Ah yes I actually change this mapping to be `mapping(bytes32 => Relay) disputedRelays` in a follow up PR #3281 so I'm going to ignore this comment for now.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: Here's the path matt describes:
- Relay gets instant-relayed by someone, `recipient` receives `amount`.
- Relay gets disputed, depending on result of dispute, either the `slowRelayer` or the `disputer` gets paid out
- Another `slowRelayer2` now starts another Relay. The contract somehow knows that there is a dispute pending for this relay with an instant relayer, so the Relay becomes `type=INSTANT` and when the Relay concludes, only the relayers get paid their rewards + bonds. `recipient` receives nothing.

The tricky part is `The contract somehow knows that there is a dispute pending for this relay` because there is currently no way to locate the disputed relay when another `slowRelayer2` starts a relay.

mrice32: Exactly!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
mrice32: Sounds good!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I agree we don't need to track disputes, we can just let the OptimisticOracle pay out disputes. I ultimately settled on the following structs:

```
    // A Relay represents a an attempt to finalize a cross-chain transfer that originated on an L2 DepositBox contract
    // and can be bridged via this contract.
    enum RelayState { Uninitialized, Pending, Finalized }

    // Data from L2 deposit transaction.
    struct DepositData {
        uint64 depositTimestamp;
        uint64 maxFeePct;
        uint64 depositId;
        uint256 amount;
        address l2Sender;
        address recipient;
        address l1Token;
    }

    // A Relay is linked to a L2 Deposit.
    struct RelayData {
        RelayState relayState;
        uint256 priceRequestTime;
        uint64 proposerRewardPct;
        uint64 realizedFeePct;
        address slowRelayer;
        address instantRelayer;
    }

    // Associate deposits with pending relays. When RelayState is Uninitialized, new relay attempts can be
    // made for this deposit. Also contains information neccessary to pay out relayers on successful relay.
    mapping(bytes32 => RelayData) public relays;
    // Associates ancillary data related to relay price request with the deposit hash that the relay is attempting to 
    // fulfill. We need to key by the ancillary data so that the OptimisticOracle can locate relays on callbacks using 
    // only price requests' ancillary data. The ancillary data should contain all information required by off-chain 
    // actors (validators, DVM voters, etc.) to verify that a relay is valid.
    mapping(bytes32 => bytes32) public ancillaryDataToDepositHash;
```

This stores a forward lookup of `depositHash` to `RelayData`, so duplicate relays can't be pending for the same deposit, and a reverse lookup of `ancillaryData` to `depositHash`, so that the OO can callback into this contract and delete a pending relay--thereby freeing it up for other deposits. Any instant relayers can be carried over in the `relays` struct.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -125,108 +123,154 @@ contract BridgePool is Testable {
...
+        bytes memory ancillaryData,
+        uint256 refund
+    ) public {
+        bytes32 depositHash = ancillaryDataToDepositHash[keccak256(ancillaryData)];
+        RelayData storage relay = relays[depositHash];
 
-    function settleDisputedRelay(uint256 depositId, address slowRelayer) public {}
+        // Mark pending relay as uninitialized but do not delete instant relayer information which should be copied
+        // over to next slow relay.
+        relay.relayState = RelayState.Uninitialized;
```
chrismaree, line 219: if this has been disputed then `Uninitialized` as a state seems pretty counterintuitive. would making a more explicit `Disputed` State make more sense?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: Agreed this is a good catch. The key here is that you are setting the `proposer` in the `proposePriceFor` call so arbitration to finalize the dispute can occur directly in the OO. I think the original thinking (and the reason we stored this data) was to let this arbitration occur within the BridgePool. i.e the pool will pull rewards from the OO after the dispute, call to see who was correct and then pay out accordingly. 

This is pretty redundant seeing setting the `proposer` on the OO gives us the same arbitration mechanism without needing to store any data about the dispute within this contract.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: also an event would be very useful for this

chrismaree (approved):

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I'll add en event, I was planning to leave implementation for this function for another PR.

> if this has been disputed then Uninitialized as a state seems pretty counterintuitive. would making a more explicit Disputed State make more sense?
The problem with this is that we need to then add a check to relayDeposit from `require(relays[depositHash].relayState == RelayState.Uninitialized, "Pending relay for deposit exists")` to `require(relays[depositHash].relayState == RelayState.Uninitialized || relays[depositHash].relayState == RelayState.Disputed, "Pending relay for deposit exists")`

PR #3277, opened 2021-08-04
feat: add multichain support for price feeds

mrice32:
**Motivation**

Price feeds should support querying custom nodes.

**Summary**

Add a parameter to all price feeds that allows the user to specify an env variable that the feed should use to find the node url to pass to web3.

For example:

```
config = {
  nodeUrlEnvVar: "POLYGON_NODE_URL",
  ...
};
```

This configuration would tell the feed to use the environment variable `POLYGON_NODE_URL` to get the node url that it should use. It was done this way so that users aren't forced to hardcode their node urls in the config json.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
+++ b/packages/financial-templates-lib/src/price-feed/CreatePriceFeed.js
@@ -5,6 +5,7 @@ const { getTruffleContract } = require("@uma/core");
 const { BlockFinder } = require("./utils");
 const { getPrecisionForIdentifier, PublicNetworks } = require("@uma/common");
 const { multicallAddressMap } = require("../helpers/multicall");
+const Web3 = require("web3");
 
 // Price feed interfaces (sorted alphabetically)
 const { BalancerPriceFeed } = require("./BalancerPriceFeed");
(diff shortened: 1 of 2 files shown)


chrismaree (approved):
This is a great solution to this problem. following this, a config can be made that enables you to define and combination of chain providers. for example, could easily make an expression price feed containing L1 Ethereum and polygon with min effort by just configuring the two feeds accordingly with their provider URLs.

nicholaspai:
> This is a great solution to this problem. following this, a config can be made that enables you to define and combination of chain providers. for example, could easily make an expression price feed containing L1 Ethereum and polygon with min effort by just configuring the two feeds accordingly with their provider URLs.

+1. Another separate problem is how to deal with pricefeeds that need node data from multiple providers. There isn't a use case for this currently but I imagine will come soon.

nicholaspai (approved):
Looks great! What are your thoughts on how we handle pricefeeds that need multiple providers (and can't just use the `ExpressionPriceFeed` to define two pricefeeds with different providers)?

mrice32:
@nicholaspai 
> Looks great! What are your thoughts on how we handle pricefeeds that need multiple providers (and can't just use the `ExpressionPriceFeed` to define two pricefeeds with different providers)?

I think the solution would have to be inside the price feed itself. It would need to have two (or more) optional configuration parameters that allows you to configure the node providers. Can you think of a more elegant solution?

Can you think of a use case for this?

mrice32:
Going to go ahead and merge, but @nicholaspai these are good questions!

PR #3278, opened 2021-08-04
Updates GasEthQuery to use receipt_effective_gas_price instead of raw gas price

nateawelch:
This PR updates the query used to determine average gas prices such that it accounts for changes to the gas fee market made by EIP-1559


smb2796 (approved):
Tested on my end and looks good. Thanks @nateawelch - approving

PR #3279, opened 2021-08-04
fix: make unused variables an error in the typescript linter

mrice32:
**Motivation**

The eslint considers an unused variable a warning for typescript.

**Summary**

Added a rule to make this an error rather than warning.

Note: I think this is a warning by default because typescript, itself, can check for this. However, rather than change the tsconfigs, it seems easier to add this to eslint. Long term, we may want to have a standard tsconfig that we inherit from, but that's a slightly larger project.

This also fixes a minor issue with unexported functions in common.

The two other common warnings for ts code are:
- `@typescript-eslint/no-explicit-any` -- we still use any in a lot of places, so we either need to disable it individually or just keep this rule off and just try to avoid any as much as possible.
- `@typescript-eslint/explicit-module-boundary-types` -- I'm not sure we want to add this one to our list of errors. It essentially requires an explicit return type for each exported function in a file. These explicit types get extremely cumbersome and time-consuming to write in certain places. It would be a bigger time sink to enable this.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -23,7 +23,7 @@ module.exports = {
         "plugin:@typescript-eslint/eslint-recommended",
         "plugin:@typescript-eslint/recommended",
       ],
-      rules: { "@typescript-eslint/no-var-requires": 0 },
+      rules: { "@typescript-eslint/no-var-requires": 0, "@typescript-eslint/no-unused-vars": "error" },
     },
   ],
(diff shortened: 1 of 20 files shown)


nicholaspai (approved):
Nice! LGTM

PR #3284, opened 2021-08-04
feat(pf-configs): Add default pf configs for ibBTC/BTC, BTC/ibBTC, ibBTC/USD and USD/ibBTC

Reinis-FRP:

**Motivation**

ibBTC related price identifiers defined in [UMIP-119](https://github.com/UMAprotocol/UMIPs/blob/master/UMIPs/umip-119.md) needs default price feed configuration.

**Summary**

Default price feed configuration was added for ibBTC/BTC, BTC/ibBTC, ibBTC/USD and USD/ibBTC price identifiers.

**Details**

These price feed configurations utilizes multichain support introduced in #3277.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

NA

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.js
@@ -940,6 +940,23 @@ const defaultConfigs = {
     twapLength: 7200,
     invertPrice: true,
   },
+  "ibBTC/BTC": {
+    type: "medianizer",
+    lookback: 7200,
+    twapLength: 1800,
+    invertPrice: true,
+    medianizedFeeds: [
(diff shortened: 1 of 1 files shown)


Reinis-FRP:
One issue to consider is the naming of environment variable provided by `nodeUrlEnvVar`. There is no direct way for users of this pf configuration to know which chain the `POLYGON_NODE_URL` should be pointing at. This might become more complex when `nodeUrlEnvVar` is provided through ancillary data as environment variables should be user/application specific. Maybe a better alternative would be to set ChainID in pf configuration and then agree on the naming standards for node environment variables (e.g. NODE_URL_CHAIN_137)?

mrice32 (commented):
I can't comment on the correctness of the price feed config as @smb2796 and @pemulis are more knowledgeable about what the params should be, but @Reinis-FRP agree with your suggestion and I will add a task to use chain id instead of env var name.

smb2796 (approved):
Looks good to me

PR #3285, opened 2021-08-05
feat(insured-bridge): Implement speedUpRelay and settleRelay

nicholaspai:
**Motivation**

Instant relayer needs to pay relay amount minus realized fee.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -195,18 +184,41 @@ contract BridgePool is Testable {
...
+        require(relays[depositHash].instantRelayer == address(0), "Relay has already been instant relayed");
         relay.instantRelayer = msg.sender;
-        // TODO: Pull funds from caller.
-        emit RelaySpedUp(_depositHash, msg.sender);
+
+        // Pull relay amount minus fees from caller.
+        IERC20(_depositData.l1Token).safeTransferFrom(
+            msg.sender,
+            address(this),
+            _depositData.amount - _getRealizedFeeAmount(relay.realizedFeePct, _depositData.amount)
```
chrismaree, line 205: nit: put this on a separate line for readability

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -333,6 +345,30 @@ contract BridgePool is Testable {
                 .rawValue;
     }
 
+    function _getRealizedFeeAmount(uint64 _realizedFeePct, uint256 _amount) private pure returns (uint256) {
```
chrismaree, line 348: this computation will likely chance in the next few days but it follows the interface spec for now.

chrismaree (approved):
looks good! we will need to change the fee logic a bit once we've fully converged on exactly how we want to do that but for now this looks good.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I'm trying to keep the stack small and use fewer intermediate variables. Do you think this gives gas savings either way?

PR #3288, opened 2021-08-06
feat(api): 1884 add additional global utility calls

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/1884/add-single-contract-tvl-calls-to-global-channel

**Summary**
There were several calls requested by francesco to be added to facilitate data requests when the type of contract is unknown. These are documented in the readme.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.clubhouse.io/uma-project/story/1884/add-single-contract-tvl-calls-to-global-channel

### DIFF
--- a/packages/api/src/apps/api/README.md
+++ b/packages/api/src/apps/api/README.md
@@ -75,6 +75,7 @@ Current channels are
 
 - emp - all emp calls
 - lsp - all lsp calls
+- global - custom calls that are not contract type specific
 
 ## Action Client
 
@@ -245,3 +246,51 @@ Lsp calls are on the lsp channel. These calls are the same as the EMP calls, but
 ### lsp/totalTvlHistoryBetween(start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") => Stat[]
 
 ### lsp/totalTvlSlice(start = 0, length = 1, currency: CurrencySymbol = "usd") => Stat[]
+
+## Global
+
+Global calls allow you to make calls without knowing the type of contract you are dealing with
+
+### global/listActive()
+
+List all active contracts, both LSP or EMP
+
+### global/listExpired()
+
+List all expired contracts, both LSP or EMP
+
+### global/getState(address: string)
+
+Get the state of a known contract address.
+
+### global/globalTvl(currency: CurrencySymbol = "usd")
+
+Global tvl, or the sum of tvls of all known contracts.
+
+### global/tvl(address: string, currency: CurrencySymbol = "usd")
+
+Look up tvl for any contract address.
+
+### global/globalTvlHistoryBetween(start = 0, end: number = nowS(), currency: CurrencySymbol = "usd")
+
+Global tvl history between times in seconds.
+
+### global/tvlHistoryBetween(address: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd")
+
+TVL history for a particular address.
+
+### global/globalTvm(currency: CurrencySymbol = "usd")
+
+Global tvm or the sum of all tvms of all known contracts.
+
+### global/tvm(address: string, currency: CurrencySymbol = "usd")
+
+TVM for a particular address.
+
+### global/tvmHistoryBetween(address: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd")
+
+TVM history for a specific address between timestamps in seconds.
+
+### global/tvmHistorySlice(address: string, start = 0, length = 1, currency: CurrencySymbol = "usd")
+
+TVM history slice for a specific address.
--- a/packages/api/src/services/actions/global.ts
+++ b/packages/api/src/services/actions/global.ts
@@ -40,10 +40,8 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     async getState(address: string) {
       assert(address, "requires a contract address");
       for (const action of contractActions) {
-        try {
-          return await action.getState(address);
-        } catch (err) {
-          // do nothing
+        if (await action.hasAddress(address)) {
+          return action.getState(address);
         }
       }
       throw new Error("Unable to find contract address " + address);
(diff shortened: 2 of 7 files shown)


packages/api/src/services/actions/global.ts
```diff
@@ -1,71 +1,150 @@
 import assert from "assert";
 import { Json, Actions, AppState, CurrencySymbol, AllContractStates } from "../..";
-import * as Queries from "../../libs/queries";
 import bluebird from "bluebird";
 import { BigNumber } from "ethers";
 import { nowS } from "../../libs/utils";
+import { Handlers as EmpActions } from "./emp";
+import { Handlers as LspActions } from "./lsp";
```
daywiss, line 7: im using the other action channels within the global channel so i can reuse the logic

packages/api/src/services/actions/global.ts
```diff
@@ -1,71 +1,150 @@
...
+        async (sum, actions) => {
+          return sum.add(((await actions.tvl(undefined, currency)) as string) || "0");
         },
         BigNumber.from("0")
       );
       return sum.toString();
     },
-    async tvlHistoryBetween(start = 0, end: number = nowS(), currency: CurrencySymbol = "usd") {
+    async tvl(address: string, currency: CurrencySymbol = "usd") {
+      assert(address, "requires a contract address");
```
daywiss, line 62: these all folllow a very similar pattern, it loops through the contract type actions, and assumes if no error is thrown its a sucessful response, and keeps looping until it returns or runs out of types to check and throws an error

mrice32: Using a try-catch to weed out contract types that don't have this address seems a bit odd. Action also seems to be misnomer since each of these is just a contract type.

A structure like this would be preferable in my opinion:
```js
for (const contractType of contractTypes) {
  if (contractType.contracts.has(address)) {
    contractType.contracts.get(address).tvl();
  }
}
```

Things may not be architected this way, so that's more of an example than a concrete suggestion. Just saying that a more deliberate access mechanism makes 1) code more readable and 2) makes error detection and handling much better because we're not wrapping everything in a try-catch that uses errors as a proxy for non-existence.

I think any code that does the following should be avoided if at all possible:
```js
        } catch (err) {
          // do nothing
        }
```

mrice32 (commented):
Looks good -- one comment on the general flow of these functions!

Gamaranto (approved):
LGTM. One thing to think about for the future is if is there the need to keep the `lsp` and `emp` paths as consumers will mostly use the `global` path most times

packages/api/src/services/actions/global.ts
daywiss: yea i think you are right, will update this

packages/api/src/apps/api/README.md
```diff
@@ -245,3 +246,51 @@ Lsp calls are on the lsp channel. These calls are the same as the EMP calls, but
...
+
+### global/getState(address: string)
+
+Get the state of a known contract address.
+
+### global/globalTvl(currency: CurrencySymbol = "usd")
+
+Global tvl, or the sum of tvls of all known contracts.
+
+### global/tvl(address: string, currency: CurrencySymbol = "usd")
```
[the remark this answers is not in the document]
daywiss: i think it could be, i can add it in

packages/api/src/apps/api/README.md
```diff
@@ -245,3 +246,51 @@ Lsp calls are on the lsp channel. These calls are the same as the EMP calls, but
...
+
+### global/tvlHistoryBetween(address: string, start = 0, end: number = nowS(), currency: CurrencySymbol = "usd")
+
+TVL history for a particular address.
+
+### global/globalTvm(currency: CurrencySymbol = "usd")
+
+Global tvm or the sum of all tvms of all known contracts.
+
+### global/tvm(address: string, currency: CurrencySymbol = "usd")
```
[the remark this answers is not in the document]
daywiss: tvm isnt available for every contract but it can be undefined or null if not available

packages/api/src/apps/api/README.md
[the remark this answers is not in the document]
mrice32: Either way seems fine. Just trying to think of ways to combine calls to make things simpler on the caller.

packages/api/src/apps/api/README.md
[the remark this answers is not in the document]
mrice32: Makes sense. Either way seems fine, although I'd lean towards lumping it in so the user has fewer calls to make.

packages/api/src/services/actions/global.ts
mrice32: 👌

mrice32 (approved):
LGTM!

packages/api/src/services/actions/global.ts
```diff
@@ -1,71 +1,136 @@
...
+      for (const action of contractActions) {
+        if (await action.hasAddress(address)) {
+          return action.tvl([address], currency);
+        }
+      }
+      throw new Error("Unable to find TVL for address " + address);
```
mrice32, line 66: nit: shouldn't this just be that the address wasn't found -- it's unrelated to the action type, right?

mrice32 (commented):
One minor comment.

packages/api/src/services/actions/global.ts
daywiss: Technically you are right but I like that there is some context in the error message since you could show this directly to user and they would understand that the app was trying to get tvl for an address.  If the error was just "address not found" for all the calls, its a bit more cryptic and will make it hard to immediately know the call that's failing.

daywiss: if you feel strongly on this though lets discuss, as it will cause many of the global errors to change

packages/api/src/apps/api/README.md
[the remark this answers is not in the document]
daywiss: I added both

PR #3297, opened 2021-08-10
fix(common): fix issues with gckms typescript conversion

mrice32:
**Motivation**

The gckms system has been broken since the typescript upgrade in common.

**Summary**

This PR fixes three distinct issues:
- The previous implementation imported the gckms file incorrectly, assuming it was a JSON file.
- The previous implementation did not copy the pre-existing js hidden file to dist. It assumed it was a typescript file that was compiled. However, the tsc compiler ignores files preceded by a `.` by default, so making this work as a ts file is tricky.
- The previous implementation did an incorrect type check and string conversion in order to generate the private keys.

Note: this new process _does_ require you to run the build with the gckms override file if you intend to use it. Adding the override file _after_ building will not work.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/common/package.json
+++ b/packages/common/package.json
@@ -19,7 +19,7 @@
   },
   "scripts": {
     "test": "hardhat test ./test/*",
-    "build": "rm -rf dist && tsc"
+    "build": "rm -rf dist && tsc && (cp src/gckms/.GckmsOverride.js dist/gckms/.GckmsOverride.js || echo 'No override file found, continuing')"
   },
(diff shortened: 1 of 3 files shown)


mrice32:
> great stuff. do you see any way to write tests for these kind of files in common?

Not that I can think of since it is dependent on GCP resources. A mock would likely just replicate our bad assumptions about how it works and the return types :/.

PR #3298, opened 2021-08-11
feat(insured-brige): update depositor fee information to give more control to different fee types

chrismaree:
**Motivation**

The insured bridge should enable the depositor to specify more settings about the fees they are paying. 

**Summary**

This PR modifies how fees are specified by the depositor. Before this PR, the depositor would simply define a `maxFeePct` that would encapsulate all fees that they would pay. After this PR, the depositor defines three numbers: 
1) `slowRelayFeePct` percentage of the total amount sent that the slow relayer gets. This includes the proposer bond.
2) `instantRelayFeePct` percentage of the total amount sent that the flast relayer gets.
3) `quoteDeadline` Timestamp, after which, the L2 deposit will revert. The depositor will accept L1 LP fees for any utilization amount up to this timestamp.

In addition to these fee changes, this PR also made two other modifications to the deposit box:

1) The relayer no longer specifies a `proposerRewardPct` number. This is now baked into the `slowRelayFeePct` as this reward goes to the `slowRelayer`.
2) At the conclusion of `settleRelay`, the instant relayer is refunded for the bridged amount. This was an error in the previous implementation where they did not have the correct funds sent to them at this point. As the instant relayer paid the full bridging amount at the time of calling `speedUpRelay`, they need to be re-immersed at the settlement of a relay action.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -159,36 +163,38 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
     function relayDeposit(
         uint64 depositId,
         uint64 depositTimestamp,
         address recipient,
         address l2Sender,
         uint256 amount,
-        // TODO: Allow caller to distinguish between slow and fast fees/rewards.
-        uint64 realizedFeePct,
-        uint64 maxFeePct,
-        uint64 proposerRewardPct
```
chrismaree, line 178: note: no longer define the `proposerRewardPct`. this is baked into the `slowRelayFeePct`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -485,13 +488,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         return
             keccak256(
                 abi.encode(
-                    _depositData.depositTimestamp,
```
chrismaree, line 488: re-ordered these to match the definition of the type.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -463,19 +475,10 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         return StoreInterface(FinderInterface(bridgeAdmin.finder()).getImplementationAddress(OracleInterfaces.Store));
     }
 
-    function _getProposerRewardAmount(uint64 _proposerRewardPct, uint256 _amount) private pure returns (uint256) {
+    function _getAmountFromPct(uint64 _RewardPct, uint256 _amount) private pure returns (uint256) {
```
chrismaree, line 478: made a more general method for computing a % from a number.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -46,27 +46,28 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
         uint64 depositId;
-        uint256 amount;
+        uint64 depositTimestamp;
         address l2Sender;
         address recipient;
         address l1Token;
+        uint256 amount;
+        uint64 slowRelayFeePct;
+        uint64 instantRelayFeePct;
+        uint64 quoteDeadline;
```
mrice32, line 57: Sorry, maybe I'm totally missing something, but what does this variable do? If the deposit "reverts" what does that mean and how are things rolled back on L2? If it doesn't cause things to actually revert, how does it differ from depositTimestamp?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -214,7 +219,8 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         // Sanity check that pool has enough balance to cover relay amount + proposer reward. Reward amount will be
         // paid on settlement after the OptimisticOracle price request has passed the challenge period.
         require(
-            l1Token.balanceOf(address(this)) >= amount + _getProposerRewardAmount(proposerRewardPct, amount),
+            l1Token.balanceOf(address(this)) >=
+                amount + _getAmountFromPct(slowRelayFeePct + instantRelayFeePct, amount),
```
mrice32, line 223: I thought amount already included the relay fees? How is amount computed by the relayer based on the deposit size?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
mrice32: two nits:
- casing of `_RewardPct` should be `_rewardPct` I think.
- variable name implies this is a reward, but the function is meant to be general. Maybe it should be something more generic, like `_percent` or `_pct`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -159,36 +163,38 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-        uint64 realizedFeePct,
-        uint64 maxFeePct,
-        uint64 proposerRewardPct
+        uint64 slowRelayFeePct,
+        uint64 instantRelayFeePct,
+        uint64 quoteDeadline,
+        uint64 realizedLpFeePct
     ) public {
-        require(realizedFeePct <= maxFeePct, "Invalid realized fee");
+        require(realizedLpFeePct <= 0.5e18, "Invalid realized LP fee");
```
mrice32, line 184: Should we be doing max(fee, 0.5) instead of require? Also, should we be validating that the other fees are <= 0.25 here too to ensure this method always works?

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -183,17 +185,29 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
...
+        uint64 slowRelayFeePct,
+        uint64 instantRelayFeePct,
+        uint64 quoteDeadline
     ) public onlyIfDepositsEnabled(l2Token) {
         require(isWhitelistToken(l2Token), "deposit token not whitelisted");
-        require(maxFeePct <= 1e18, "maxFeePct cannot be over 100% (represented as 1e18)");
+        require(
+            slowRelayFeePct + instantRelayFeePct <= 0.5e18,
+            "sum of slow & instant relayer fees can not exceed 50%"
+        );
```
mrice32, line 205: I thought we wanted to make sure each individually was < 0.25, so that we could call `max(fee, 0.25)` on each on L1. I'm cool either way, just thought keeping the constraints separate might make the checks a little simpler to reason about. I could see arguments either way, though.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -183,17 +185,29 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
...
+
+        // Note that the OVM's notion of `block.timestamp` is marginally different to the main ethereum L1 EVM. The OVM
+        // timestamp corresponds to the L1 timestamp of the last confirmed L1 ⇒ L2 transaction. If five minutes have
+        // passed since the last L1 ⇒ L2 transaction, the timestamp is automatically refreshed.
+        require(quoteDeadline >= getCurrentTime(), "deposit mined after deadline");
```
mrice32, line 210: Similar to above, deadline and the description is a little confusing. Is this deadline the quote timestamp? Or is it just some constraint on the quote timestamp, and the relayer uses the deposit timestamp for the quote?

mrice32 (commented):
Looks awesome!! A few minor comments, but overall this is 👌

mrice32 (approved):
Oops, meant to approve on that review!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: good points. will refactor accordingly.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: In the context of this contract, it does not do anything. It is used exclusively on the L2 contract to limit when a deposit is valid and by the relayer to define when to compute L1 pool utilization. It is included here for completeness to duplicate the exact state statement by the relayer. We could remove this as the OO disputes can fetch this from L2. wdyt?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: yup. capping the fee at 0.5 to prevent reverts sounds like a good idea. Validating the other fees here is also possible. if this was not done here anything that exceeds this should be disputed but I agree it's good idea to cap it to prevent the dispute state.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: o that is true. the original require statement was more correct. will revert that.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: yup yup. you right. I will split it into two requirements for each part.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: the deadline is a number up to which the depositor is willing to accept a quote. All we need to verify is that the current timestamp is less than or equal to the deadline such that the block inclusion time is before the deadline. The relayers will use the _block timestamp_ of the transaction to compute the LP fee for the deposit.

chrismaree: I've updated the comments to be more clear.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: regarding how it differs from `depositTimestamp`, the `depositTimestamp` will be the `block.timestamp` of that the L2 deposit is included and the `quoteDeadline` is the limit up and to which the depositor is willing to accept a quote.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: Other argument could be that under no saituations should this value ever be greater than 0.5e18 and so preventing revert might be better than a ceil.

chrismaree: After syncing on this we agreed that a `require` is sufficient.

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: After discussing this we chose to change this to `quoteTimestamp`. PR has been updated accordingly

PR #3299, opened 2021-08-11
fix(serverless-infrastructure): Add additional error modes to serverless hub

chrismaree:
**Motivation**

Production recently saw a bug in the serverless infrastructure that all UMA bots use. This PR addresses these issues identified.

**Summary**

This PR makes two changes to the serverless hub:
1) if a bot returns after execution without any `stdout`, it is considered an error.
2) if a bot returns after execution and does not contain any `started` string in the `stdout` logs, it is considered an error.

These two checks, in conjunction, should enable us to catch all failure modes that the downstream bots might present and are currently missed. empty `stdout` is an indication of some kind of silent failure. Bots missing the `started` string indicate that the bot failed before it started up fully. This could happen if the provider threw an error during boot, for example.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -132,9 +132,9 @@ hub.post("/", async (req, res) => {
     // hub is called again.
     await _saveQueriedBlockNumber(req.body.configFile, blockNumbersForChain);
 
-    // Finally, loop over all config objects in the config file and for each append a call promise to the promiseArray. Note
-    // that each promise is a race between the serverlessSpoke command and a `_rejectAfterDelay`. This places an upper
-    // bound on how long each spoke can take to respond, acting as a timeout for each spoke call.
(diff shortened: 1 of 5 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -441,6 +442,24 @@ function _processSpokeResponse(botKey, spokeResponse, validOutputs, errorOutputs
           spokeResponse.reason.response.data.execResponse),
       botIdentifier: botKey,
     };
+  } else if (spokeResponse.value && spokeResponse.value.execResponse && spokeResponse.value.execResponse.stdout == "") {
```
chrismaree, line 445: these are the additional failure mode checks we look for: `stdout` is blank or the log does not contain "started"

mrice32 (approved):
LGTM! I like how this keeps the implementation pretty much in line with what we were doing before to avoid introducing any major changes/bugs!

PR #3300, opened 2021-08-11
fix(api): 1942 add price converted gcr field to emp state

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1942/gcr-and-identifier-not-displaying-correctly

**Summary**

Does an extra calculation when identifier price is available to the gcr. So `gcr = total collateral / total tokens / identifier price`. The old gcr calculation is called `rawGcr` which is just `total collateral / total tokens`. No changes needed on frontend.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1942/gcr-and-identifier-not-displaying-correctly

### DIFF
--- a/packages/api/src/libs/queries/emp.ts
+++ b/packages/api/src/libs/queries/emp.ts
@@ -3,7 +3,7 @@ import assert from "assert";
 import * as uma from "@uma/sdk";
 import { calcGcr } from "../utils";
 import bluebird from "bluebird";
-import { BigNumber } from "ethers";
+import { BigNumber, utils } from "ethers";
 import type { AppState, CurrencySymbol, PriceSample } from "../..";
 
 const { exists } = uma.utils;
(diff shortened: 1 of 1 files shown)


packages/api/src/libs/queries/emp.ts
```diff
@@ -113,14 +113,27 @@ export default (appState: Dependencies) => {
...
+      gcrRaw = calcGcr(state).toString();
+    } catch (err) {
+      // nothing
+    }
+    // gcr now also takes into consideration the identifier price, defaults to raw if anything happens
+    let gcr = gcrRaw;
+    try {
+      gcr = utils
+        .parseUnits(gcrRaw)
+        // if theres no identifier price, just divide by 1
```
mrice32, line 128: why do we assume 1? If we don't have a price, shouldn't we just return null or undefined or something?

mrice32 (commented):
Looks good! one small comment!

packages/api/src/libs/queries/emp.ts
daywiss: its dividing out 1 eth so the result ends up being the same as the raw gcr. i figured its probably best to fall back to displaying the raw gcr rather than nothing if theres no identifier price available. what do you think?

mrice32: Hmmm, okay. We should probably align on some error handling/data availability protocol for all fields so the FE has a way to detect if the value isn't available.

Intuitively, subbing in the raw GCR seems like a decision we'd want to make on the FE instead of on the backend.

mrice32 (approved):
LGTM!

PR #3302, opened 2021-08-12
feat(insured-bridge): Exponential fee calculation

chrismaree:
**Motivation**

As part of the insured bridge design, we require a mechanism to attribute relayed fees to LPs gradually over the periods that tokens are moved from L2->L1 via the canonical bridge. This amounts to a 1-week loan over the period.

This PR adds the initial fee market to facilitate this where LP fees are gradually paid out over one week.

**Summary**

This PR adds the following methods:
- `sync` which updates the internal contracts liquid/liquid reserves based on finalized L2->L1 token transfers.
- `exchangeRateCurrent` which calculates the current redemption rate between LP tokens and L1 collateral. It factors in pending/accumulated fees.
- `getUnallocatedAccumulatedFees` computes any fees that have been accumulated from the last time the fee logic was called on the bridgePool.
- `updateFeeCounters` factors in attributed LP fees and allocates them to the appropriate internal variables.

**Details**

Please see code comments that explain how the fees are calculated.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -357,23 +371,62 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+    function exchangeRateCurrent() public returns (uint256) {
         if (totalSupply() == 0) return 1e18; //initial rate is 1 pre any mint action.
 
-        // Consider a naive rate implementation. This acts like a step function, increasing when funds hit L1 from the
-        // canonical bridge. TODO: update with a more elaborate technique that pays out gradually over the 1 week loan.
+        // First, update fee counters and local accounting of finalized transfers from L2->L1.
+        updateFeeCounters(0); // Accumulate all allocated fees from the last time this method was called.
+        sync(); // Fetch any balance changes due to token bridging finalization and factor them in.
+
+        // ExchangeRate := (liquidReserves+utilizedReserves+cumulativeLpFeesEarned-undistributedLpFees)/lpTokenSupply
```
mrice32, line 401: I'm still thinking on this implementation, but one thought that pops into my head when reading this is why do we need to track cumulative lp fees earned?

Could we not just add LP fees to the utilized reserves as they're paid out and allow the utilized reserves to go negative in the case where there are some fees yet to be paid out, but nothing in transit?

The way I'm thinking about it (and I could totally be missing something, as this is very nuanced), the cumulativeLpFeesEarned feels like it could be unnecessary complexity.

Thoughts?

mrice32 (commented):
Really cool implementation! One high-level comment.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: You cant just add them to the utilized reserves as they don't technically count as utilized reserves. if anything, they are `liquidReserves` that are in flight. I think what you propose would work numerically but it's overloading the notion of a `utalizedReserve` making it harder to reason about.

I'm going to try out what you propose to double-check it works as expected in any case.

chrismaree: I tried changing the exchange rate equation to this:
```
FixedPoint.Unsigned memory numerator =
            FixedPoint.Unsigned(liquidReserves).sub(FixedPoint.Unsigned(undistributedLpFees));
        if (utilizedReserves > 0) numerator = numerator.add(FixedPoint.Unsigned(uint256(utilizedReserves)));
        else numerator = numerator.sub(FixedPoint.Unsigned(uint256(utilizedReserves * -1)));
        return numerator.div(FixedPoint.Unsigned(totalSupply())).rawValue;
```

while updating the `updateFeeCounters` method to this:
```
function updateFeeCounters(uint256 allocatedLpFees) internal {
        // Calculate the unallocatedAccumulatedFees from the last time the contract was called.
        uint256 unallocatedAccumulatedFees = getUnallocatedAccumulatedFees();

        // Decrement the undistributedLpFees by the amount of accumulated fees.
        undistributedLpFees = undistributedLpFees > unallocatedAccumulatedFees
            ? undistributedLpFees - unallocatedAccumulatedFees
            : 0;

        lastLpFeeUpdate = getCurrentTime();

        // If this method was called from settleRelay then it includes the exact amount of fees to be attributed to the
        // Liquidity providers. add this to the associated variables.
        if (allocatedLpFees > 0) {
            undistributedLpFees += allocatedLpFees;
            utilizedReserves += int256(allocatedLpFees);
        }
    }
```

This effectively amounts to adding in the `allocatedLpFees` to the utilized reserves. After doing this, I could remove the `cumulativeLpFeesEarned` in totality.

I commented out all tests that tested `cumulativeLpFeesEarned` and `utilizedReserves` (as the former is removed after this and the latter's definition has changed somewhat) and ran all the tests. Every virtual balance accounting test works exactly as before (which is expected).

This means we can opt for this and save one local variable and make a marginal improvement in gas costs. The question need to be: is this worth the overloading and potentially higher complexity in understanding to bake this into the `utalizedReserves`? It might be but am curious to hear others thoughts.

chrismaree: I've updated the implementation to track what you recommended optimistically. let me know if this matches to your expected format.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -20,7 +20,7 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
  * @notice Contract deployed on L1 that provides methods for "Relayers" to fulfill deposit orders that originated on L2.
  * The Relayers can either post capital to fulfill the deposit instantly, or request that the funds are taken out of
  * a passive liquidity provider pool following a challenge period. Related, this contract ingests liquidity from
- * passive liquidity providers and returns them claims to withdraw their funds. Liquidity providers are incentivized
+ * passive liquidity providers and returns them claims to withdraw their funds. Liquidity providers are incentivize
```
nicholaspai, line 23:
```suggestion
 * passive liquidity providers and returns them claims to withdraw their funds. Liquidity providers are incentivized
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -357,23 +368,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+     */
+    function getUnallocatedAccumulatedFees() public view returns (uint256) {
+        // UnallocatedLpFees := min(undistributedLpFees*lpFeeRatePerSecond*timeFromLastInteraction,undistributedLpFees)
+        // The min acts to pay out all fees in the case the equation returns more than the remaining a fees.
         return
             FixedPoint
-                .fromUnscaledUint(liquidReserves)
-                .add(FixedPoint.fromUnscaledUint(utilizedReserves))
-                .div(FixedPoint.fromUnscaledUint(totalSupply()))
+                .Unsigned(undistributedLpFees)
```
nicholaspai, line 421: where does `undistributedLpFees` get updated in this contract?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -35,7 +35,19 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+    uint256 public lpFeeRatePerSecond;
+
+    // Last timestamp that LP fees were updated.
+    uint256 public lastLpFeeUpdate;
+
+    // Cumulative undistributed LP fees. As fees accumulate, they are subtracted from this number.
+    uint256 public undistributedLpFees;
+
+    // How long, in seconds, it takes for transfers to move from L2 optimism to L1 Ethereum.
+    uint256 public optimismL2toL1Time = 1 weeks;
```
nicholaspai, line 50: Should we allow this variable to be modified by an owner?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -113,12 +125,14 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
         address _l1Token,
+        uint256 _lpFeeRatePerSecond,
         address _timer
     ) Testable(_timer) ExpandedERC20("UMA Insured Bride LP Token", "UMA-LP", 18) {
         bridgeAdmin = BridgeAdminInterface(_bridgeAdmin);
-        require(bridgeAdmin.finder() != address(0), "Invalid bridge pool factory");
 
         l1Token = IERC20(_l1Token);
+        lastLpFeeUpdate = getCurrentTime();
+        lpFeeRatePerSecond = _lpFeeRatePerSecond;
```
nicholaspai, line 135: Should we constrain this %/second at all? I'm curious what's an example value that would be realistic

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -183,9 +197,9 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     ) public {
         // The realizedLPFeePct should never be greater than 0.5e18 and the slow and instant relay fees
```
nicholaspai, line 198: Why should `realizedLPFeePct < 0.5`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -357,23 +368,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         emit RelayDisputed(depositHash, keccak256(ancillaryData));
     }
 
-    /************************************
-     *           View FUNCTIONS         *
-     ************************************/
+    /**
+     * @notice Synchronize any balance changes in this contract with the utilized & liquid reserves. This would be done
+     * at the conclusion of an L2->L1 token transfer via the canonical token bridge.
```
nicholaspai, line 373: Does the L2 Bridge actually call this method? IIUC it seems that `sync` only gets called indirectly by `exchangeRateCurrent` which gets called by `add/removeLiquidity`?

If so, then I think the `exchangeRateCurrent` name is not as clear as it can be. Some possible changes:
- rename `exchangeRateCurrent` to `syncAndGetCurrentExchangeRate` to signify better that it syncs balance state
- add a `sync` modifier that calls `_sync` and add it to the `exchangeRateCurrent` method, or even the LP functions. 

I think the main confusion for me reading the code is that its not clear what triggers `sync`, but clearly `sync` needs to get called everytime the `liquidReserves` or L1 token balance for this contract changes.

chrismaree: This is not called by the L2 bridge method. There is no way to enforce that the conclusion of bridge action from L2->L1 via the canonical bridge are done from a specific contract. We need to think about the finalization of a bridging action as tokens being "dropped" onto the contract at any time. In this framing, the `sync` method can be called at any time to incorporate these balance changes.

The reason I followed this pattern (and named this method as I did) is as follows:
1) `exchangeRateCurrent` is the same name that compound uses for their cToken->underlying token exchange rate logic.
2) There might be a case where you want to `sync` and not do anything else, for example we _might_ want our bot to call `sync` after finalizing an L2->L1 transfer via the bridge.

I do see your point though that it's a bit convoluted that the _only_ time `sync` is called in this contract is via `exchangeRateCurrent`, which is called from the LP methods. The problem with adding it as a modifier is that this is actually pretty gas inefficient. For example, consider adding liquidity. you would: a) sync on the modifier on this method and then sync again on the call to the `exchangerRateCurrent`. The way it's set up now enforces that `sync` is ONLY called when it is needed.

Sync _only_ needs to be called when the L1 token balance for this contract changes, not when adding or removing `liquidReserves`.

I'm happy to rename this method though, as a middle ground. not the _biggest_ fan of `syncAndGetCurrentExchangeRate`. Any other ideas?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: I dont really think constraining this is that important. If this number is really big (like 100% per second) then all the fees are paid out instantly. This implementation limits the maximum payout to the total pending rewards so there is never a situation where we payout more than the total to be paid; it simply modifies how quickly the rewards are paid out.

We would, however, want a modifier to update this. I thought I'd leave this for later though.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: I actually realized we don't need this variable anymore! It was when lingering from a previous implementation that used it to compute the exchange rate such that all fees are paid out over the duration of the L12->L1 bidding period; now, all this information is baked into `lpFeeRatePerSecond`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: The reasoning is we constrain the slow and instant fees to each be less than 0.25e18, summing to less than 0.5e18. By definition, the `realizedLpFeePct` cant exceed 0.5e18 or the total fee percentage will exceed 1 (meaning that the total due in fees is more than the relayed amount). This require to prevent this situation.

nicholaspai: >By definition, the realizedLpFeePct cant exceed 0.5e18 or the total fee percentage will exceed 1 (meaning that the total due in fees is more than the relayed amount)

Why is this by definition? Can you explain in the comment, this part isn't that clear to me.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: >exchangeRateCurrent is the same name that compound uses for their cToken->underlying token exchange rate logic.

I think this is a good reason to ease other integrator's onboarding who are used to cTokens.

> The problem with adding it as a modifier is that this is actually pretty gas inefficient. For example, consider adding liquidity. you would: a) sync on the modifier on this method and then sync again on the call to the exchangerRateCurrent. The way it's set up now enforces that sync is ONLY called when it is needed.

Why not remove `sync` from the `exchangeRateCurrent` call, make it a modifier for the LP methods, and then also expose a public `sync` method (that calls `_sync`) for anyone to manually trigger? I'm not sure this is the best pattern IMO and I think its probably fine as is but just throwing an idea out there.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -357,23 +365,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+
+    /************************************
+     *           View FUNCTIONS         *
+     ************************************/
+
+    /**
+     * @notice Computes the current amount of unallocated fees that have accumulated from the previous time this the
+     * contract was called.
+     */
+    function getUnallocatedAccumulatedFees() public view returns (uint256) {
```
mrice32, line 413: nit: maybe we should change the function name to `getFees` or `getFeePayout` (or something simple like that)? I found the name to be a bit hard to understand.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -461,9 +506,31 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+    // Update the internal fee counter metrics by adding in any accumulated fees from the last time this logic was
+    // called. Considers any allocated fees that are assigned to the LPs at the conclusion of a bridging action.
+    function updateFeeCounters(uint256 allocatedLpFees) internal {
+        // Calculate the unallocatedAccumulatedFees from the last time the contract was called.
+        uint256 unallocatedAccumulatedFees = getUnallocatedAccumulatedFees();
+
+        // Decrement the undistributedLpFees by the amount of accumulated fees.
+        undistributedLpFees = undistributedLpFees > unallocatedAccumulatedFees
+            ? undistributedLpFees - unallocatedAccumulatedFees
+            : 0;
```
mrice32, line 521: Is the 0 case possible? Don't we take a min in `getUnallocatedAccumulatedFees`?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -461,9 +506,31 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     }
 
     /**************************************
-     *        INTERNAL FUNCTIONS          *
+     *    INTERNAL & PRIVATE FUNCTIONS    *
      **************************************/
 
+    // Update the internal fee counter metrics by adding in any accumulated fees from the last time this logic was
+    // called. Considers any allocated fees that are assigned to the LPs at the conclusion of a bridging action.
+    function updateFeeCounters(uint256 allocatedLpFees) internal {
```
mrice32, line 514: Two questions:
1. Should we make a separate method for _adding new fees_ as opposed to distributing fees? They seem like distinct actions.
2. nit: variable name seems a little confusing. Thoughts on `newFees` or something similar?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -461,9 +506,31 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+            : 0;
+
+        lastLpFeeUpdate = getCurrentTime();
+
+        // If this method was called from settleRelay then it includes the exact amount of fees to be attributed to the
+        // Liquidity providers. add this to the total undistributed LP fees and the utalized reserves. Adding it to the
+        // utalized reserves acts to book keep the fees while they are in transit.
+        if (allocatedLpFees > 0) {
+            undistributedLpFees += allocatedLpFees;
+            utilizedReserves += int256(allocatedLpFees);
```
mrice32, line 530: Note: we should be careful about casting. I think casting _makes sense_, but we need to be very very sure that it's not possible to pass in a huge number (> 2^255) here (i.e. that must _always revert_, it being disputable is not enough).

Open question whether we should always do safe casting where we revert if the uint is too large.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -357,23 +365,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-     ************************************/
+    /**
+     * @notice Synchronize any balance changes in this contract with the utilized & liquid reserves. This would be done
+     * at the conclusion of an L2->L1 token transfer via the canonical token bridge.
+     */
+    function sync() public {
+        // Check if the l1Token balance of the contract is greater than the liquidReserves. If it is then the bridging
+        // action from L2->L1 has concluded and the local accounting can be updated.
+        uint256 l1TokenBalance = l1Token.balanceOf(address(this));
+        if (l1TokenBalance > liquidReserves) {
```
mrice32, line 376: Should we do anything if `l1TokenBalance < liquidReserves`?

mrice32 (approved):
Looks good! Minor comments + questions

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: `getFees` sounds good! Have refactored

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: yup! you totally right on that. This little bit of logic was a holdout from the previous implementation I believe. I've removed this operator and all works fine in the tests & I don't think there is a situation where this becomes problematic due to the `min` taken in the previous method.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: yup. good points. I split this into two new re-named methods: 1) `updateAccumulatedLpFees()` and `allocateLpFees(uint256 allocatedLpFees)`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: agreed we need caution in this case. The only situation I can think where this would be an issue is when the `amount` transferred (and therefore the `allocatedLpFees` is absolutely massive (> 2^255). this wont really be possible given we control the token whitelisting process.

what additional checks do you think would be useful to add for additional safety?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: this should _never_ be the case. `l1TokenBalance` should always be >= the `liquidReserves`. 

Logically though if this was the case then we would need to book keep it by updating the utilized/liquid reserve variables by the shortfall.

PR #3303, opened 2021-08-13
feat(api): add lsp tvm by copying in tvl as the source

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1953/update-the-global-tvm-to-account-for-lsps

**Summary**

This replicates a lot of the infrastructure from EMP TVM to LSP TVM. The final TVM values for LSPs are just copied from its TVL, so ultimately the user just sees the LSP TVL as TVM.  The global TVM is also updated with the new global TVM values.

Global TVM value goes from 31.6 million -> 45.4 million with this new addition

**Details**

This allows the frontend to call to get an LSPs tvm per address or a group of addresses on the `lsp/tvm` action. Most likely this wont be used. The main change comes from the global tvm value which should now be slightly higher and no changes to FE necessary for this.  Another minor modification is that LSP states now have an additional `tvm` field which will be a mirror of the `tvl`. If we ever change this calculation in the future we can pass through the new tvm value with no further modifications to the api.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1953/update-the-global-tvm-to-account-for-lsps

### DIFF
--- a/packages/api/src/libs/queries/lsp.ts
+++ b/packages/api/src/libs/queries/lsp.ts
@@ -26,11 +26,18 @@ export default (appState: Dependencies) => {
     assert(appState.stats.lsp[currency], "invalid currency: " + currency);
     return appState.stats.lsp["usd"].latest.tvl.get(address);
   }
+  function getTvm(address: string, currency: CurrencySymbol = "usd") {
+    assert(appState.stats.lsp[currency], "invalid currency: " + currency);
+    return appState.stats.lsp["usd"].latest.tvm.get(address);
+  }
   async function getFullState(state: tables.lsps.Data) {
     const collateralState = state.collateralToken ? await erc20s.get(state.collateralToken).catch(() => null) : null;
     const longTokenState = state.longToken ? await erc20s.get(state.longToken).catch(() => null) : null;
     const shortTokenState = state.shortToken ? await erc20s.get(state.shortToken).catch(() => null) : null;
     const tvl = await getTvl(state.address)
+      .then((sample) => sample.value)
+      .catch(() => null);
+    const tvm = await getTvm(state.address)
       .then((sample) => sample.value)
       .catch(() => null);
     return {
@@ -45,6 +52,7 @@ export default (appState: Dependencies) => {
       shortTokenSymbol: shortTokenState?.symbol,
       collateralSymbol: collateralState?.symbol,
       tvl,
+      tvm,
       type: "lsp",
     };
   }
@@ -60,7 +68,7 @@ export default (appState: Dependencies) => {
     const tvls = await Promise.all(
       addresses.map(async (address) => {
         try {
-          const stat = await appState.stats.lsp[currency].latest.tvl.get(address);
+          const stat = await getTvl(address, currency);
           return stat.value || "0";
         } catch (err) {
           return "0";
@@ -87,15 +95,52 @@ export default (appState: Dependencies) => {
     const result = await getTotalTvlSample(currency);
     return result.value;
   }
+  // tvm calculations, these are just based on tvl
+  async function sumTvm(addresses: string[], currency: CurrencySymbol = "usd") {
+    const tvms = await Promise.all(
+      addresses.map(async (address) => {
+        try {
+          const stat = await getTvm(address, currency);
+          return stat.value || "0";
+        } catch (err) {
+          return "0";
+        }
+      })
+    );
+
+    const tvm = await tvms.reduce((sum, tvm) => {
+      return sum.add(tvm);
+    }, BigNumber.from("0"));
+
+    return tvm.toString();
+  }
+  async function totalTvm(currency: CurrencySymbol = "usd") {
+    const addresses = Array.from(appState.registeredLsps);
+    return sumTvm(addresses, currency);
+  }
+  async function getTotalTvmSample(currency: CurrencySymbol = "usd") {
+    assert(appState.stats.lsp[currency], "Invalid currency: " + currency);
+    return appState.stats.lsp[currency].latest.tvm.getGlobal();
+  }
+  async function getTotalTvm(currency: CurrencySymbol = "usd") {
+    const result = await getTotalTvmSample(currency);
+    return result.value;
+  }
   return {
     getFullState,
     getAny,
     listExpired,
     listActive,
     totalTvl,
+    totalTvm,
     sumTvl,
+    sumTvm,
     getTotalTvl,
+    getTotalTvm,
     getTotalTvlSample,
+    getTotalTvmSample,
     hasAddress,
+    getTvm,
+    getTvl,
   };
 };
--- a/packages/api/src/services/actions/lsp.ts
+++ b/packages/api/src/services/actions/lsp.ts
@@ -47,9 +47,10 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       addresses = addresses ? lodash.castArray(addresses) : [];
       return queries.sumTvl(addresses, currency);
     },
-    // lsp does not support tvm at this time, but we want a function here for consistency with emp interface
-    async tvm() {
-      return "0";
+    async tvm(addresses: string[] = [], currency: CurrencySymbol = "usd") {
+      if (addresses == null || addresses.length == 0) return queries.totalTvm(currency);
+      addresses = addresses ? lodash.castArray(addresses) : [];
+      return queries.sumTvm(addresses, currency);
     },
     async tvlHistoryBetween(
       contractAddress: string,
@@ -62,11 +63,11 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
     },
     // this is not supported
     async tvmHistoryBetween() {
-      throw new Error("LSP does not support TVM");
+      throw new Error("LSP does not support TVM History");
     },
     // this is not supported
     async tvmHistorySlice() {
-      throw new Error("LSP does not support TVM");
+      throw new Error("LSP does not support TVM History");
     },
     async tvlHistorySlice(contractAddress: string, start = 0, length = 1, currency: CurrencySymbol = "usd") {
       assert(stats.lsp[currency], "Invalid currency type: " + currency);
--- a/packages/api/src/services/stats/lsp.ts
+++ b/packages/api/src/services/stats/lsp.ts
@@ -73,6 +73,19 @@ export default (config: Config, appState: Dependencies) => {
       })
     );
   }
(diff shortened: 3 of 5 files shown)


packages/api/src/libs/queries/lsp.ts
```diff
@@ -26,13 +26,20 @@ export default (appState: Dependencies) => {
     assert(appState.stats.lsp[currency], "invalid currency: " + currency);
     return appState.stats.lsp["usd"].latest.tvl.get(address);
   }
+  function getTvm(address: string, currency: CurrencySymbol = "usd") {
```
daywiss, line 29: helper function to just get the tvm table for lsps

packages/api/src/libs/queries/lsp.ts
```diff
@@ -45,6 +52,7 @@ export default (appState: Dependencies) => {
       shortTokenSymbol: shortTokenState?.symbol,
       collateralSymbol: collateralState?.symbol,
       tvl,
+      tvm,
```
daywiss, line 55: this is the call which joins lsp state, so we add in the tvm value

packages/api/src/libs/queries/lsp.ts
```diff
@@ -60,7 +68,7 @@ export default (appState: Dependencies) => {
     const tvls = await Promise.all(
       addresses.map(async (address) => {
         try {
-          const stat = await appState.stats.lsp[currency].latest.tvl.get(address);
+          const stat = await getTvl(address, currency);
```
daywiss, line 71: small cleanup here, no logic change

packages/api/src/libs/queries/lsp.ts
```diff
@@ -87,15 +95,52 @@ export default (appState: Dependencies) => {
     const result = await getTotalTvlSample(currency);
     return result.value;
   }
+  // tvm calculations, these are just based on tvl
+  async function sumTvm(addresses: string[], currency: CurrencySymbol = "usd") {
```
daywiss, line 99: the following functions are almost identical to the previous tvl calculations. basically just helpers to sum tvm, or get various types of data

packages/api/src/services/actions/lsp.ts
```diff
@@ -47,9 +47,10 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       addresses = addresses ? lodash.castArray(addresses) : [];
       return queries.sumTvl(addresses, currency);
     },
-    // lsp does not support tvm at this time, but we want a function here for consistency with emp interface
-    async tvm() {
-      return "0";
+    async tvm(addresses: string[] = [], currency: CurrencySymbol = "usd") {
```
daywiss, line 50: this is what gets exposed in the api. it follows the pattern and logic established for emps. you can basically provide an array of addresses, or an empty array which will sum all known tvms.  This also ends up being used on the global channel in order to sum tvm across lsps and emps

packages/api/src/services/stats/lsp.ts
```diff
@@ -73,6 +73,19 @@ export default (config: Config, appState: Dependencies) => {
       })
     );
   }
+  // TVM for LSP will just be the TVL
+  async function updateTvm(address: string) {
+    const tvl = await stats[currency].latest.tvl.get(address);
+    return stats[currency].latest.tvm.upsert(address, tvl);
```
daywiss, line 79: this stats/lsp file is where the tvm values get updated and stored. This particular function just moves the tvl to tvm and nothing more. If our TVM calculation ever changes, this is the only function which would need to update how we calculate that value

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
   }
+  // tvm is just based on TVL for LSPS.
+  async function updateGlobalTvm() {
+    const value = await queries.totalTvm(currency);
+    const update = {
+      value,
+      timestamp: nowS(),
+    };
+    // normally you would upsert an lsp address where "global" is, but we are going to use a custom value to represent tvl across all addresses
+    return stats[currency].latest.tvm.upsertGlobal(update);
```
daywiss, line 119: this stores the tvm total across all lsps.

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
+  }
   async function update() {
     const addresses = Array.from(registeredLsps.values());
     await updateAllTvl(addresses).then((results) => {
       results.forEach((result) => {
-        if (result.status === "rejected") console.error("Error updating tvl: " + result.reason.message);
+        if (result.status === "rejected") console.error("Error updating TVL: " + result.reason.message);
+      });
+    });
+    await updateAllTvm(addresses).then((results) => {
```
daywiss, line 128: the update function adds this call to calculate tvm for every lsp contract

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
+      results.forEach((result) => {
+        if (result.status === "rejected") console.error("Error updating TVM: " + result.reason.message);
       });
     });
     await updateGlobalTvl().catch((err) => {
       console.error("Error updating global TVL: " + err.message);
     });
+    await updateGlobalTvm().catch((err) => {
```
daywiss, line 136: this call just sums and stores the tvm across all lsps

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
+    };
+    // normally you would upsert an lsp address where "global" is, but we are going to use a custom value to represent tvl across all addresses
+    return stats[currency].latest.tvm.upsertGlobal(update);
+  }
   async function update() {
     const addresses = Array.from(registeredLsps.values());
     await updateAllTvl(addresses).then((results) => {
       results.forEach((result) => {
-        if (result.status === "rejected") console.error("Error updating tvl: " + result.reason.message);
+        if (result.status === "rejected") console.error("Error updating TVL: " + result.reason.message);
```
mrice32, line 125: nit: should we be printing the error stack on these rather than just the message?

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
+      });
+    });
+    await updateAllTvm(addresses).then((results) => {
+      results.forEach((result) => {
+        if (result.status === "rejected") console.error("Error updating TVM: " + result.reason.message);
```
mrice32, line 130: Same comment here.

packages/api/src/services/stats/lsp.ts
```diff
@@ -95,16 +108,34 @@ export default (config: Config, appState: Dependencies) => {
...
       });
     });
     await updateGlobalTvl().catch((err) => {
       console.error("Error updating global TVL: " + err.message);
     });
+    await updateGlobalTvm().catch((err) => {
+      console.error("Error updating global TVM: " + err.message);
```
mrice32, line 137: Some comment here.

packages/api/src/libs/queries/lsp.ts
```diff
@@ -87,15 +95,52 @@ export default (appState: Dependencies) => {
...
+    const tvms = await Promise.all(
+      addresses.map(async (address) => {
+        try {
+          const stat = await getTvm(address, currency);
+          return stat.value || "0";
+        } catch (err) {
+          return "0";
+        }
```
mrice32, line 107: Similar to previous comments, why do we need to bury errors here? Is this expected in the happy path? If so, is there a different logic we could use to detect empty? If it shouldn't happen, should we be printing an error message so we can see it for debugging purposes?

packages/api/src/libs/queries/lsp.ts
```diff
@@ -26,13 +26,20 @@ export default (appState: Dependencies) => {
...
   async function getFullState(state: tables.lsps.Data) {
     const collateralState = state.collateralToken ? await erc20s.get(state.collateralToken).catch(() => null) : null;
     const longTokenState = state.longToken ? await erc20s.get(state.longToken).catch(() => null) : null;
     const shortTokenState = state.shortToken ? await erc20s.get(state.shortToken).catch(() => null) : null;
     const tvl = await getTvl(state.address)
       .then((sample) => sample.value)
       .catch(() => null);
+    const tvm = await getTvm(state.address)
+      .then((sample) => sample.value)
+      .catch(() => null);
```
mrice32, line 42: Same comment here as elsewhere on burying errors. Is this something we expect in the happy path?

mrice32 (commented):
Looks good! a few questions about error handling and messages!

packages/api/src/services/stats/lsp.ts
daywiss: it really depends. the only reason i didnt do this is because there would just be a ton of logs.  There are addresses that consistently fail when requesting data from price feeds and therefore they fail when calculating stats. Without a way to blacklist or silence certain addresses it ends up being really noisy without the stack trace.  here is a small sample, you could imagine this with stack traces for each line 
```

packages/api/src/libs/queries/lsp.ts
daywiss: This is really built for the frontend to call, IMO i dont want these calls to throw errors unless an invalid address is specified.  There are all kinds of errors on the ingestion side which can lead to no TVM/TVL for a contract, these are logged with the contract address on ingestion side.  If we passed these errors to the user on every query nothing would work. But we still have the ability to debug if the numbers seem wrong by searching through ingestion logs. See previous comment for example logs.

packages/api/src/libs/queries/lsp.ts
daywiss: Again this is mainly used to query state for the frontend. I dont want a failure in one property to prevent the rest of the state from being returned. If this is null and it shouldnt be, we can debug with logs on the ingestion side.

packages/api/src/services/stats/lsp.ts
daywiss: see comment above

packages/api/src/services/stats/lsp.ts
daywiss: see comment above

mrice32 (approved):
LGTM!

PR #3305, opened 2021-08-16
fix(GasEstimator): quick fix for EIP1559

chrismaree:
**Motivation**

Our code _is_ compatible with EIP1559 (via legacy transactions),  the etherchain API that we use primarily for gas estimation introduced a breaking change that messed up how we size gas prices. This PR is a quick fix to get transactions up and running again while we work on a more correct long term solution that uses EIP1559 tx fee model correctly.

**Summary**

The main API that we use for gas is `https://www.etherchain.org/api/gasPriceOracle`. This currently returns a data structure that looks like this:
```
{"safeLow":1,"standard":1.5,"fast":4,"fastest":6.2,"currentBaseFee":33.1,"recommendedBaseFee":67.1}
```
previously, the `fastest` was used as the current fast gas price. now, this represents the EIP1559 priority fee. To address this quickly we can swap to using the `recommendedBaseFee`. This will overprice transactions but it will ensure we are still paying the `fastest` rate for quick inclusion. 

A follow on PR should update how we send transactions to correctly include the `maxBaseFee` and `maxPrioirtyFee` in accordance with EIP1559.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/helpers/GasEstimator.js
+++ b/packages/financial-templates-lib/src/helpers/GasEstimator.js
@@ -12,12 +12,11 @@ const GAS_ESTIMATOR_MAPPING_BY_NETWORK = {
   // }
   1: {
     url: "https://www.etherchain.org/api/gasPriceOracle",
-    defaultMaxFeePerGasGwei: 50,
-    defaultMaxPriorityFeePerGasGwei: 5,
-    type: "london",
+    backupUrl: "https://api.etherscan.io/api?module=gastracker&action=gasoracle",
+    defaultFastPriceGwei: 50,
   },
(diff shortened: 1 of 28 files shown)


packages/financial-templates-lib/src/helpers/GasEstimator.js
```diff
@@ -100,7 +100,7 @@ class GasEstimator {
       //   "fastest": "39.6"
       // }
       if (json.fastest) {
-        let price = json.fastest;
+        let price = json.recommendedBaseFee;
```
chrismaree, line 103: simple 1 line change to get things back up and running.

nicholaspai (approved):

PR #3306, opened 2021-08-16
fix(GasEstimator): protocol upgrade for EIP1559

chrismaree:
**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/common/src/TransactionUtils.ts
+++ b/packages/common/src/TransactionUtils.ts
@@ -20,7 +20,7 @@ const argv = minimist(process.argv.slice(), {});
  * @notice Uses the ynatm package to retry the transaction with increasing gas price.
  * @param {*Object} web3.js object for making queries and accessing Ethereum related methods.
  * @param {*Object} transaction Transaction to call `.call()` and subsequently `.send()` on from `senderAccount`.
- * @param {*Object} transactionconfig transaction config, e.g. { gasPrice, from }, passed to web3 transaction.
+ * @param {*Object} transactionConfig config, e.g. { maxFeePerGas, maxPriorityFeePerGas, from }, passed to transaction.
  * @return Error and type of error (originating from `.call()` or `.send()`) or transaction receipt and return value.
  */
 export const runTransaction = async ({
@@ -64,7 +64,7 @@ export const runTransaction = async ({
   if (await accountHasPendingTransactions(web3, transactionConfig.from))
     transactionConfig.nonce = await getPendingTransactionCount(web3, transactionConfig.from);
   // Else, there is no pending transaction and we use the current account transaction count as the nonce.
-  // This method does not play niceley in tests. Leave the nounce null to auto fill.
+  // This method does not play nicely in tests. Leave the nonce null to auto fill.
   else if (argv.network != "test") transactionConfig.nonce = await web3.eth.getTransactionCount(transactionConfig.from);
 
   // Next, simulate transaction and also extract return value if its a state-modifying transaction. If the function is state
--- a/packages/financial-templates-lib/src/helpers/GasEstimator.js
+++ b/packages/financial-templates-lib/src/helpers/GasEstimator.js
@@ -100,7 +100,7 @@ class GasEstimator {
       //   "fastest": "39.6"
       // }
       if (json.fastest) {
-        let price = json.fastest;
+        let price = json.recommendedBaseFee;
         return price;
       } else {
         throw new Error(`Main gas station API @ ${url}: bad json response`);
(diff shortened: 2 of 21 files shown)


packages/financial-templates-lib/src/helpers/GasEstimator.js
```diff
@@ -66,21 +77,31 @@ class GasEstimator {
...
-  // Returns the current fast gas price in Wei, converted from the stored Gwei value.
+  // Returns the current fast maxFeePerGas and maxPriorityFeePerGas.
   getCurrentFastPrice() {
-    // Sometimes the multiplication by 1e9 introduces some error into the resulting number,
-    // so we'll conservatively ceil the result before returning. This output is usually passed into
-    // a web3 contract call so it MUST be an integer.
-    return Math.ceil(this.lastFastPriceGwei * 1e9);
+    // Sometimes the multiplication by 1e9 introduces some error into the resulting number, so we'll conservatively ceil
+    // the result before returning.This output is usually passed into a web3 contract call so it MUST be an integer.
+    if (this.type == "london") {
```
chrismaree, line 91: this is the core idea of this PR: return an object with a different structure depending on the network type. then, when using this variable, use the spread operator (`...`) to expand this into the transaction object's settings.

packages/financial-templates-lib/src/helpers/GasEstimator.js
```diff
@@ -66,21 +77,31 @@ class GasEstimator {
...
+      return {
+        maxFeePerGas: Math.ceil(this.latestMaxFeePerGasGwei * 1e9),
+        maxPriorityFeePerGas: Math.ceil(this.latestMaxPriorityFeePerGasGwei * 1e9),
+      };
+    } else return { gasPrice: Math.ceil(this.lastFastPriceGwei * 1e9) };
```
mrice32, line 96: Out of curiosity, is there any _harm_ in including gasPrice when you also provide the base and priority fees? What does web3 default to in that case?

not suggesting a change -- purely curious

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/helpers/GasEstimator.js
chrismaree: this is a good point. I think we can include it without any _harm_. some reasons for excluding it could be:
1) We don't have a number for this so what would we put there? null perhaps?
2) having both `maxFeePerGas` & `maxPriorityFeePerGas` and `gasPrice` makes it a bit harder to reason about what kind of transaction this is; is it an London or legacy? well now rather than looking at the keys we need to look at keys + values and see if some are null
3) I'm not 100% sure that web3 will choose the right behavior. if you give it all 3 variables, with potentially `gasPrice` set to null or `maxFeePerGas` and `maxPriorityFeePerGas` will web3 consistently know to send either legacy or London? not really sure about this one so I'd rather be explicit

packages/common/src/TransactionUtils.ts
```diff
@@ -92,6 +93,7 @@ export const runTransaction = async ({
     const minGasPrice = transactionConfig.gasPrice;
     const maxGasPrice = 2 * 3 * parseInt(minGasPrice);
 
+    // TODO: this method will need to be updated once we have updated ynatm to work with EIP1559.
```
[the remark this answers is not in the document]
chrismaree: I think this pretty much covers the required changes! I'm implementing now and will revert back if anything else needs to be changed.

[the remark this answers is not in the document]
chrismaree: Turns out we need to do a bit of typecasting to workaround the lack of London-supported transaction types inclusion in some exported types. Please see the implementation and let me know if you are ok with how it's structured right now

packages/common/src/TransactionUtils.ts
```diff
@@ -84,22 +87,44 @@ export const runTransaction = async ({
...
-      maxGasPrice,
-      gasPriceScalingFunction,
-      delay: retryDelay,
-    });
+    if (!transactionConfig.gasPrice || (!transactionConfig.maxFeePerGas && transactionConfig.maxPriorityFeePerGas))
+      throw new Error("No gas information provided");
+
+    let receipt;
+    // Sending a legacy transaction (pre-London)
+    if (transactionConfig.gasPrice) {
```
mrice32, line 100: tiny nit: why do we prioritize legacy? Totally fine with it, just curious mainly.

mrice32 (approved):
LGTM!

chrismaree:
Note that this PR is currently blocked until such time that web3.js fixes a bug they have when defaulting EIP1559/London (type-2) transactions.

pemulis:
@chrismaree is this PR still blocked?

PR #3310, opened 2021-08-17
fix(api): fix event queries for discovering contracts

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/2006/bug-emp-registry-and-lsp-creator-miss-created-contracts

**Summary**

This ensures the right block numbers are being set when calling update on services which use block events. Previously we were missing some contract creation events because some block ranges were being skipped. This just updates these calls with the correct block ranges on init and for each tick.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/2006/bug-emp-registry-and-lsp-creator-miss-created-contracts

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -140,17 +140,19 @@ export default async (env: ProcessEnv) => {
     globalStats: Services.stats.Global({ debug }, appState),
   };
 
+  const initBlock = await provider.getBlock("latest");
+
   // warm caches
-  await services.registry();
+  await services.registry(appState.lastBlockUpdate, initBlock.number);
   console.log("Got all EMP addresses");
 
-  await services.lspCreator.update();
+  await services.lspCreator.update(appState.lastBlockUpdate, initBlock.number);
   console.log("Got all LSP addresses");
 
-  await services.emps(0);
+  await services.emps(appState.lastBlockUpdate, initBlock.number);
   console.log("Updated EMP state");
 
-  await services.lsps.update(0);
+  await services.lsps.update(appState.lastBlockUpdate, initBlock.number);
   console.log("Updated LSP state");
 
   await services.erc20s.update();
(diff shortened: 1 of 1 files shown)


packages/api/src/apps/api/app.ts
```diff
@@ -140,17 +140,19 @@ export default async (env: ProcessEnv) => {
     globalStats: Services.stats.Global({ debug }, appState),
   };
 
+  const initBlock = await provider.getBlock("latest");
```
daywiss, line 143: on startup, getting the latest block number for querying events

packages/api/src/apps/api/app.ts
```diff
@@ -140,17 +140,19 @@ export default async (env: ProcessEnv) => {
...
+
   // warm caches
-  await services.registry();
+  await services.registry(appState.lastBlockUpdate, initBlock.number);
```
daywiss, line 146: these calls take in a start block and end block. just explicitly setting the range on all these calls to be sure all events are captured

packages/api/src/apps/api/app.ts
```diff
@@ -206,10 +208,10 @@ export default async (env: ProcessEnv) => {
...
       const end = profile("Updating state from block event");
       // update everyting
-      await services.registry(appState.lastBlock, blockNumber);
-      await services.lspCreator.update(appState.lastBlock, blockNumber);
-      await services.emps(appState.lastBlock, blockNumber);
-      await services.lsps.update(appState.lastBlock, blockNumber);
+      await services.registry(appState.lastBlockUpdate, blockNumber);
+      await services.lspCreator.update(appState.lastBlockUpdate, blockNumber);
+      await services.emps(appState.lastBlockUpdate, blockNumber);
+      await services.lsps.update(appState.lastBlockUpdate, blockNumber);
```
daywiss, line 214: these were using lastBlock, which updates every new block, but what we really want is the lastBlockUpdate as the start block, so we dont miss any events. lastBlockUpdate only changes when these services get updated.

nicholaspai: Ah tricky bug

nicholaspai (approved):
Nice solution LGTM

chrismaree (approved):

PR #3313, opened 2021-08-17
feat(sdk): 1882 complete google datastore store integration and tests

daywiss:

**Motivation**

https://app.clubhouse.io/uma-project/story/1882/add-google-datastore-store-in-sdk

**Summary**
This adds google data store integration based on the "store" interface.  There are some hacky things here because actually it was quite difficult to get some of the functionality cleanly out of the google datastore client.  I've commented on most of the wierdness.

**Details**

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/1882/add-google-datastore-store-in-sdk

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -14,7 +14,7 @@
     "start": "tsdx watch",
     "build": "tsdx build",
     "test": "tsdx test",
-    "test:e2e": "tsdx test --config jest-e2e-config.json",
+    "e2e": "tsdx test --config jest-e2e-config.json",
     "test:watch": "tsdx test --watch",
     "eslint": "eslint './src/**/*.ts'",
     "prettier": "prettier './**/*.md' './src/**/*.ts'",
@@ -37,13 +37,15 @@
     }
   ],
   "dependencies": {
+    "@types/highland": "^2.12.13",
     "@uma/core": "^2.7.0",
-    "ethers": "^5.4.2"
+    "ethers": "^5.4.2",
+    "highland": "^2.13.5"
   },
   "devDependencies": {
     "@ethersproject/abi": "^5.4.0",
-    "@ethersproject/contracts": "^5.4.0",
     "@ethersproject/abstract-provider": "^5.4.0",
+    "@ethersproject/contracts": "^5.4.0",
     "@ethersproject/providers": "^5.4.2",
     "@size-limit/preset-small-lib": "^4.10.2",
     "size-limit": "^4.10.2",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -14,7 +14,7 @@
     "start": "tsdx watch",
     "build": "tsdx build",
     "test": "tsdx test",
-    "e2e": "tsdx test --config jest-e2e-config.json",
+    "test:e2e": "tsdx test --config jest-e2e-config.json",
     "test:watch": "tsdx test --watch",
     "eslint": "eslint './src/**/*.ts'",
     "prettier": "prettier './**/*.md' './src/**/*.ts'",
@@ -37,12 +37,12 @@
     }
   ],
   "dependencies": {
-    "@types/highland": "^2.12.13",
     "@uma/core": "^2.7.0",
     "ethers": "^5.4.2",
     "highland": "^2.13.5"
   },
   "devDependencies": {
+    "@types/highland": "^2.12.13",
     "@ethersproject/abi": "^5.4.0",
     "@ethersproject/abstract-provider": "^5.4.0",
     "@ethersproject/contracts": "^5.4.0",
--- a/packages/sdk/src/stores/google-datastore/store.ts
+++ b/packages/sdk/src/stores/google-datastore/store.ts
@@ -43,16 +43,14 @@ export default function <D>(kind: string, store: Datastore): SortedStore<string,
   }
   // all this does is map over full entries to return the id of the entry. have not found a better way to query this.
   async function keys() {
-    const [results] = await store.createQuery(kind).run();
+    const [results] = await store.createQuery(kind).select("__key__").run();
     return results.map((result) => {
       return result[store.KEY].name;
     }) as string[];
   }
   // theres no way to really do built into the store client. Google recommends managing a size entry yourself.
-  async function size() {
+  async function size(): Promise<number> {
     throw new Error("size not supported in google store");
-    // typescript complains if theres no return value
-    return NaN;
   }
   // this actually queries all values, then batches them to delete N at a time. This is a horrible way
   // to do this, but I could not find a better way to "drop" all kinds from the table.
--- a/packages/sdk/src/stores/google-datastore/store.ts
+++ b/packages/sdk/src/stores/google-datastore/store.ts
@@ -1,15 +1,18 @@
+import { exists } from "../../utils";
 import type { SortedStore } from "..";
-import type { Datastore, Key } from "@google-cloud/datastore";
+import { Datastore, Key } from "@google-cloud/datastore";
+import highland from "highland";
 
 // TODO: this is a work in progress which will be completed during datastore integration.
 export default function <D>(kind: string, store: Datastore): SortedStore<string, D> {
   function makeKey(id: string): Key {
     return store.key([kind, id]);
   }
+  // return value or undefined if it doesnt exist
   async function get(id: string) {
     try {
-      const [first] = await store.get(makeKey(id));
-      return first;
+      const [result] = await store.get(makeKey(id));
+      return result;
     } catch (err) {
       return;
     }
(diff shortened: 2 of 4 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -249,10 +249,13 @@ hub.post("/", async (req, res) => {
       });
     }
     await delay(2); // Wait a few seconds to be sure the the winston logs are processed upstream.
-    res.status(500).send({
-      message: errorOutput instanceof Error ? "A fatal error occurred in the hub" : "Some spoke calls returned errors",
-      output: errorOutput instanceof Error ? errorOutput.message : errorOutput,
-    });
+    res
```
daywiss, line 252: i think this got linted with my project. i didnt edit this

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -60,11 +60,13 @@ spoke.post("/", async (req, res) => {
       execResponse,
     });
     await delay(2); // Wait a few seconds to be sure the the winston logs are processed upstream.
-    res.status(200).send({
-      message: "Process exited with no error",
-      childProcessIdentifier: _getChildProcessIdentifier(req),
-      execResponse,
-    });
+    res
```
daywiss, line 63: i think this got linted with my project. i didnt edit this

packages/sdk/package.json
```diff
@@ -37,13 +37,15 @@
     }
   ],
   "dependencies": {
+    "@types/highland": "^2.12.13",
```
mrice32, line 40: nit: this should be a dev dependency since the types are only needed at compile time.

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -1,15 +1,18 @@
+import { exists } from "../../utils";
 import type { SortedStore } from "..";
-import type { Datastore, Key } from "@google-cloud/datastore";
+import { Datastore, Key } from "@google-cloud/datastore";
```
mrice32, line 3: Before we get too deep into integrating Datastore (since we haven't done much work on it so far), should we use firestore rather than datastore? By my understanding, firestore is basically just the next generation of datastore and Google is trying to migrate users.

See https://cloud.google.com/datastore (they have a blurb about firestore being the next generation of google datastore).

Edit: the library you're using _may_ work with both?

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -22,37 +25,61 @@ export default function <D>(kind: string, store: Datastore): SortedStore<string,
...
+    return results.map((result) => {
+      return result[store.KEY].name;
+    }) as string[];
   }
+  // theres no way to really do built into the store client. Google recommends managing a size entry yourself.
   async function size() {
-    return 0;
+    throw new Error("size not supported in google store");
+    // typescript complains if theres no return value
+    return NaN;
```
mrice32, line 55: I think this will go away if you explicitly define the return value:
`size(): number {`

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -22,37 +25,61 @@ export default function <D>(kind: string, store: Datastore): SortedStore<string,
...
-  // TODO: implement these functions
+  // this actually queries all values, then batches them to delete N at a time. This is a horrible way
+  // to do this, but I could not find a better way to "drop" all kinds from the table.
   async function clear() {
-    throw new Error("clear is not implemented");
+    await highland(store.createQuery(kind).runStream())
+      .map((val: any) => {
+        return val[store.KEY];
+      })
+      .batch(500)
```
mrice32, line 64: Why'd you pick 500? Are we sure we need to do this? Will the library handle this sort of thing for us?

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -22,37 +25,61 @@ export default function <D>(kind: string, store: Datastore): SortedStore<string,
...
+    return results.map((result) => {
+      return [result[store.KEY].name, result];
+    }) as [string, D][];
   }
   async function values() {
-    return [];
+    const [results] = await store.createQuery(kind).run();
+    return results;
   }
+  // all this does is map over full entries to return the id of the entry. have not found a better way to query this.
```
mrice32, line 44: We might be able to use [select](https://googleapis.dev/nodejs/datastore/latest/Query.html#select) to only pull the key? But not sure if that works for keys

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -22,37 +25,61 @@ export default function <D>(kind: string, store: Datastore): SortedStore<string,
...
+        return store.delete(keys);
+      })
+      .flatMap(highland)
+      .collect()
+      .toPromise(Promise);
   }
-  async function slice() {
-    return [];
+  async function slice(id: string, length: number) {
+    const [result] = await store.createQuery(kind).filter("__key__", ">=", makeKey(id)).limit(length).run();
```
mrice32, line 73: Are we sure limit ensures that we get the _first_ `n`? As opposed to just _some_ `n` that satisfy the constraint? Unclear to me from the docs.

mrice32 (commented):
Looks good! A few comments and questions, but I think this is pretty much good to go!

packages/sdk/package.json
```diff
@@ -14,7 +14,7 @@
     "start": "tsdx watch",
     "build": "tsdx build",
     "test": "tsdx test",
-    "test:e2e": "tsdx test --config jest-e2e-config.json",
+    "e2e": "tsdx test --config jest-e2e-config.json",
```
chrismaree, line 17: we should converge on a standard for defining `e2e` tests within this repo. affiliates & core has `test-e2e`, API has `e2e`, and now the skd has `e2e`.

Following your pattern here, I'm ok with changing the others to all use `e2e` as you've defined here.

packages/sdk/src/stores/google-datastore/store.ts
chrismaree: I believe this is right. firestore is their newer noSQL data base (datastore was released in 2008 while firestore was 2017). Firestore looks to have more modern syntax, better documentation and more language support. This article is helpful in comparing the two: https://cloud.google.com/datastore/docs/firestore-or-datastore. you can actually run fire store in datastore mode

however, when I try to access firestore from our GCP account I get this:
image

indicating you need to use one _or_ the other in gcp 🤔 . this might be a per-project restriction, which is not an issue as we can make a separate project for the API (@evaldofelipe might have already set this up!)

chrismaree: @mrice32 regarding using the same library if you use firestore in "native" mode then you will need to use this sdk: https://github.com/googleapis/nodejs-firestore if you use firestore in "datastore" mode then you are correct we can keep using https://github.com/googleapis/nodejs-datastore. 

this is mentioned at the bottom of the readme for the `nodejs-firestore` package:

_Note: This Cloud Firestore Server SDK does not support Firestore databases created in Datastore mode. To access these databases, use the Datastore SDK._

daywiss: wow this is really good to know. would be happy to try firestore instead

packages/sdk/src/stores/google-datastore/store.ts
daywiss: we dont actually need this particular function in any production apps. i scoured google for a way to to clear tables or "kinds" and probably saw references to numbers in the ballpark of 500-1000 batches when deleting things. as far as i know theres no other way to do this, which seems really crazy but it was a theme when working with datastore. really simple operations i would take for granted either dont exist or are really inefficient.  

i would be happy to remove this since its not strictly needed, but its just useful for tests when clearing data.  also would be happy if theres a better way to do this

packages/sdk/src/stores/google-datastore/store.ts
daywiss: yea this might work. ill give it a try

packages/sdk/package.json
daywiss: hmm ok yea. i kept intuitively typing `yarn e2e`, which is why i changed it. we can discuss in eng meeting if anyone has any particular opinions on this. i will change it back for now

packages/sdk/src/stores/google-datastore/store.ts
evaldofelipe: Let me adjust the project automation to support that first.

packages/sdk/src/stores/google-datastore/store.ts
daywiss: its a good question, i've tried to test for this but maybe its not super precise. I basically insert a bunch of things out of order, then do a slice and check the results:
```js
    result.forEach((price: Price) => {
      assert.equal(price.identifier, "b");
      if (last == null) {
        last = price;
        return;
      }
      assert.ok(makeId(price) >= makeId(last));
    });
```

basically checking here that each result id is greater than the previous, so at the very least its ordered, though this does not check for skipped values

packages/sdk/src/stores/google-datastore/store.ts
daywiss: heres a thread that talks about it: https://groups.google.com/g/google-appengine-go/c/eOu1dDJ7MNg

packages/sdk/src/stores/google-datastore/store.ts
daywiss: ive updated the tests to check the exact values that should be returned based on the original list. things seem to check out.

packages/sdk/src/stores/google-datastore/store.ts
daywiss: im going to leave this PR as a google datastore implementation, and work on a firestore implmementation as a separate feature, most likely going to use thier updated client https://googleapis.dev/nodejs/firestore/latest/
https://app.clubhouse.io/uma-project/story/2119/add-google-firestore-store-in-sdk

packages/sdk/src/stores/google-datastore/store.ts
daywiss: updated this to select `__key__`, which is some special identifier for just keys

packages/sdk/src/stores/google-datastore/store.ts
mrice32: Either way works. It seems they recommend firestore in datastore mode for "server" applications, so we might want to stick with that and keep the current client implementation.

@daywiss however, if you find that you like the firestore sdk more, then we can swap to native mode!

packages/sdk/src/stores/google-datastore/store.ts
mrice32: Okay, then that seems fine to leave at 500, although this is something I'm surprised that the sdk/client library doesn't handle this for us.

packages/sdk/src/stores/google-datastore/store.ts
mrice32: Gotcha. I think that's a good enough test.

I'm guessing this wouldn't work if we were filtering on something _other_ than key, since this probably just returns elements in the natural ordering of the db, which is by key.

LGTM!

mrice32 (approved):
LGTM!

packages/sdk/src/stores/google-datastore/store.ts
daywiss: after our discussion offline, i think running firestore in datastore mode is right. that means this client should just work out of the box. will run some tests once we have it available in gcp

PR #3314, opened 2021-08-17
feat(fx-tunnel-relayer): Implement bot that can relay messages from Polygon ChildTunnel to Ethereum RootTunnel

nicholaspai:
**Motivation**

> Matic validators continuously monitor a contract on Ethereum chain called StateSender. Each time a registered contract on Ethreum chain calls this contract, it emits an event. Using this event Matic validators relay the data to another contract on Matic chain. This StateSync mechanism is used to send data from Ethereum to Matic.
> Matic validators also periodically submit a hash of all transactions on Matic chain to Ethereum chain. This Checkpoint can be used to verify any transaction that happened on Matic. Once a transaction is verified to have happened on Matic chain, action can be taked accordingly on Ethereum.

Cross-chain messages sent from Ethereum to Polygon are automatically relayed via the [state sync](https://docs.matic.network/docs/contribute/state-sync/state-sync) mechanism. However, going the other way around requires a third party to manually finalize a message on Ethereum that originated from Polygon. Once a Polygon message's block has been included in a Merkle root that has been checkpointed to Ethereum mainnet, it can be "finalized" via code [like this](https://docs.matic.network/docs/develop/l1-l2-communication/state-transfer/#state-transfer-from-matic-to-ethereum).

**Summary**

This bot's algorithm is very simple:
- Detect `MessageSent` events emitted by `OracleChildTunnel`. These are emitted when a price request is sent from the `OptimisticOracle` on Polygon to (eventually) the DVM on Ethereum.
- For each `MessageSent`, attempt to generate a proof that the event has been included in a checkpoint to Ethereum. If the event has not been checkpointed yet, then proof generation will fail and the bot will exit early.
- If the proof was generated successfully, then attempt to call `OracleRootTunnel.receiveMessage` on Ethereum, If this succeeds, then the price request will be sent to the DVM. If this fails, then the message has probably already been received.

**Details**

I've been testing via `ts-node ./packages/fx-tunnel-relayer/src/index.ts --network mainnet_mnemonic` but I'm not sure this is the canonical way to run these bots.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/common/src/Constants.ts
```diff
@@ -40,3 +40,8 @@ export const ZERO_ADDRESS = "0x0000000000000000000000000000000000000000";
 // Block number of first EMP created.
 // https://etherscan.io/tx/0x741ccbf0f9655b0b71e3842d788d58770bd3eb80c8f5bdf4fdec7cd74a776ea3
 export const UMA_FIRST_EMP_BLOCK = 10103723;
+
```
chrismaree, line 43: Do we want this to be in common? Will any other parts our infra need access to this? If not what do you think about placing it within the bots package?

packages/fx-tunnel-relayer/src/Relayer.ts
```diff
@@ -0,0 +1,126 @@
...
+    readonly maticPosClient: any,
+    readonly oracleChildTunnel: Contract,
+    readonly oracleRootTunnel: Contract,
+    readonly web3: Web3,
+    readonly polygonEarliestBlockToQuery: number
+  ) {}
+
+  // In order to receive a message on Ethereum from Polygon, `receiveMessage` must be called on the Root Tunnel contract
+  // with a proof derived from the Polygon transaction hash that was checkpointed to Mainnet.
+  async relayMessage(): Promise<void> {
```
chrismaree, line 19: Consider renaming this to include that it also checks for new massages. Would it make sense to split this into two methods, one that checks for messages and a second relay method?

packages/fx-tunnel-relayer/src/Relayer.ts
```diff
@@ -0,0 +1,126 @@
...
+      message: "Submitting proof",
+      proof: proof,
+      account: this.account,
+    });
+    try {
+      const { receipt } = await runTransaction({
+        web3: this.web3,
+        transaction: this.oracleRootTunnel.methods.receiveMessage(proof),
+        transactionConfig: { gasPrice: this.gasEstimator.getCurrentFastPrice(), from: this.account },
+        availableAccounts: 1,
```
chrismaree, line 105: You should not need to define this. It defaults to 1 I think

packages/fx-tunnel-relayer/src/RelayerConfig.ts
```diff
@@ -0,0 +1,25 @@
...
+  readonly errorRetries: number;
+  readonly errorRetriesTimeout: number;
+  // We reuse the infura key to connect to Polygon and Ethereum networks.
+  readonly infuraApiKey: string;
+  // Any events older than now minus this value (in seconds) will be ignored.
+  readonly lookback: number;
+
+  constructor(env: ProcessEnv) {
+    const { INFURA_API_KEY, POLLING_DELAY, ERROR_RETRIES, ERROR_RETRIES_TIMEOUT, LOOKBACK } = env;
+    assert(INFURA_API_KEY, "INFURA_API_KEY required");
```
chrismaree, line 18: This is a bit different to how we normally do this. Should we consider a more standard pattern? Perhaps we can pull the key out of the CUSTOM_NODE_URL or some other structure. Otherwise, this bot is a bit different than others requiring some duplicated information presented in the common feed. Wdyt?

packages/fx-tunnel-relayer/src/index.ts
```diff
@@ -0,0 +1,135 @@
...
+      oracleRootTunnel,
+      web3,
+      polygonEarliestBlockToQuery
+    );
+
+    for (;;) {
+      await retry(
+        async () => {
+          // Update state.
+          await Promise.all([gasEstimator.update()]);
```
chrismaree, line 82: No point in promise.all here

packages/common/src/Constants.ts
nicholaspai: Yeah its pretty niche i'll move it to the bots package.

packages/fx-tunnel-relayer/src/Relayer.ts
nicholaspai: I think if we split them into two methods they should be internal methods if anything but the "fetch" logic is super simple its just a single `getPastEvents` call. For now I"m just going to rename the method `fetchAndRelayMessages` which will call `getPastEvents` and then call the internal method `_relayMessage` for each message found.

nicholaspai: does that make sense to you?

packages/fx-tunnel-relayer/src/Relayer.ts
nicholaspai: If I remove this I get this error:

```sh
src/Relayer.ts:106:48 - error TS2345: Argument of type '{ web3: Web3; transaction: any; transactionConfig: { gasPrice: any; from: string; }; }' is not assignable to parameter of type '{ web3: Web3; transaction: ContractSendMethod; transactionConfig: AugmentedSendOptions; availableAccounts: number; }'.
  Property 'availableAccounts' is missing in type '{ web3: Web3; transaction: any; transactionConfig: { gasPrice: any; from: string; }; }' but required in type '{ web3: Web3; transaction: ContractSendMethod; transactionConfig: AugmentedSendOptions; availableAccounts: number; }'.

106       const { receipt } = await runTransaction({
                                                   ~
107         web3: this.web3,
    ~~~~~~~~~~~~~~~~~~~~~~~~
... 
109         transactionConfig: { gasPrice: this.gasEstimator.getCurrentFastPrice(), from: this.account },
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
110       });
    ~~~~~~~

  ../common/dist/TransactionUtils.d.ts:23:5
    23     availableAccounts: number;
           ~~~~~~~~~~~~~~~~~
    'availableAccounts' is declared here.


Found 1 error.

```

packages/fx-tunnel-relayer/src/RelayerConfig.ts
nicholaspai: hm we need to construct providers for two networks and you can't assume that the CUSTOM_NODE_URL is an Infura node. We have two options the way I see it:
- define a POLYGON_CUSTOM_NODE_URL variable in addition to a CUSTOM_NODE_URL
- keep INFURA_API_KEY

packages/fx-tunnel-relayer/src/Relayer.ts
chrismaree: yup. that sounds like a reasonable middle ground to me

packages/fx-tunnel-relayer/src/Relayer.ts
chrismaree: damn typescript. fine then

packages/fx-tunnel-relayer/src/RelayerConfig.ts
chrismaree: so if we did `POLYGON_CUSTOM_NODE_URL` then we assume the `CUSTOM_NODE_URL` is for Ethereum L1. that seems _reasnoble_ as the `CUSTOM_NODE_URL` is injected into the bots as part of the common config. wdyt?

chrismaree:
Regarding your comment in the PR description `I've been testing via ts-node ./packages/fx-tunnel-relayer/src/index.ts --network mainnet_mnemonic but I'm not sure this is the canonical way to run these bots.` the canonical way to run them will be to run `yarn build` in the package then run the compiled js with node (or, indirectly, use `yarn`)

packages/fx-tunnel-relayer/src/Relayer.ts
nicholaspai: Yeah, not sure there's a way to both define a default value AND make a param optional in typescript

packages/fx-tunnel-relayer/src/RelayerConfig.ts
nicholaspai: Ok i'm good with that I'll make the change.

chrismaree (approved):

PR #3315, opened 2021-08-18
feat(financial-templates-lib): convert src to typescript

mrice32:
**Motivation**

We'd like to move our core packages to typescript.

**Summary**

This moves financial-templates-lib to typescript, changing logic where required to satisfy the typescript compiler and eliminate type-related edge-cases.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -1,7 +1,62 @@
...
+import type { Abi, FinancialContractType } from "../types";
+import type Web3 from "web3";
+import type {
+  ExpiringMultiPartyWeb3,
+  ExpiringMultiPartyWeb3Events,
+  PerpetualWeb3,
+  PerpetualWeb3Events,
+} from "@uma/contracts-node";
+
+type LiquidationCreatedEvent = ExpiringMultiPartyWeb3Events.LiquidationCreated | PerpetualWeb3Events.LiquidationCreated;
```
chrismaree, line 11: sheesh. this is painful.

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -174,65 +216,63 @@ class FinancialContractEventClient {
...
-        tokensOutstanding: event.returnValues.tokensOutstanding,
-        lockedCollateral: event.returnValues.lockedCollateral,
-        liquidatedCollateral: event.returnValues.liquidatedCollateral,
+        ...event.returnValues,
       });
     }
 
     // Dispute events.
-    for (let event of disputeEventsObj) {
+    for (const event of disputeEventsObj) {
```
chrismaree, line 248: nice. changes like this, while not ts specific are great to include for general code quality.

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -241,98 +281,88 @@ class FinancialContractEventClient {
       this.newSponsorEvents.push({
         transactionHash: event.transactionHash,
         blockNumber: event.blockNumber,
-        sponsor: event.returnValues.sponsor,
+        ...event.returnValues,
```
chrismaree, line 284: this will add a few more props to this object but I dont think it matters.

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -1,7 +1,62 @@
+import type { Logger } from "winston";
+import type { Abi, FinancialContractType } from "../types";
+import type Web3 from "web3";
+import type {
+  ExpiringMultiPartyWeb3,
+  ExpiringMultiPartyWeb3Events,
+  PerpetualWeb3,
+  PerpetualWeb3Events,
+} from "@uma/contracts-node";
```
chrismaree, line 9: it's awesome to see these types imported like this!

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
```diff
@@ -1,227 +0,0 @@
-// A thick client for getting information about an OptimisticOracle. Used to get price requests and
```
chrismaree, line 1: why is this marked as deleted? it did this in a few places when you had just moved/re-named the file. I guess the diff was too large for it to see that it was from a previous file.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -0,0 +1,249 @@
...
+      return event.disputer === caller;
+    });
+  }
+
+  // Returns the last update timestamp.
+  public getLastUpdateTime(): number {
+    return this.lastUpdateTimestamp;
+  }
+
+  private _getPriceRequestKey(
```
chrismaree, line 95: this took an already messy method and made it look even scarier with the type included. not sure if there is a better way to lint this. I guess if the `OptimisticOracleWeb3Events` variable name was shorter this might be able to fit on one line?

packages/financial-templates-lib/src/helpers/GasEstimator.ts
```diff
@@ -1,9 +1,47 @@
 // This script gets the current recommended `fast` gas price from etherchain
 // to inform the Liquidator and dispute bot of a reasonable gas price to use.
 
-const fetch = require("node-fetch");
+import type { Logger } from "winston";
+import fetch from "node-fetch";
 
-const GAS_ESTIMATOR_MAPPING_BY_NETWORK = {
+interface GasEstimatorMapping {
```
chrismaree, line 7: including these types of the responses is excellent. we should try get your PR in before https://github.com/UMAprotocol/protocol/pull/3306 to avoid painful rebase.

packages/financial-templates-lib/src/helpers/allowances.ts
```diff
@@ -0,0 +1,42 @@
...
+// Sets `owner` allowance for `spender` to MAX_UINT_VAL, unless `spender` already has
+// an allowance > MAX_SAFE_ALLOWANCE. Return successful approval transaction data, or undefined
+// for skipped approvals.
+export const setAllowance = async (
+  web3: Web3,
+  gasEstimator: GasEstimator,
+  ownerAddress: string,
+  spenderAddress: string,
+  currencyAddress: string
+): Promise<
```
chrismaree, line 19: I'm really not a huge fan of how some of this stuff is linted. This looks pretty janky. Not something we need to worry about right now in this PR but I'd be curious to explore other linter settings to improve this kind of layout for ts specific stuff.

packages/financial-templates-lib/src/logger/ConsoleTransport.ts
```diff
@@ -1,9 +1,9 @@
 // This transport enables Winston logging to the console.
-const winston = require("winston");
+import winston from "winston";
 const { format } = winston;
 const { combine, timestamp, colorize, printf } = format;
 
-function createConsoleTransport() {
+export function createConsoleTransport(): winston.transports.ConsoleTransportInstance {
```
chrismaree, line 6: +1

packages/financial-templates-lib/src/logger/JsonTransport.ts
```diff
@@ -1,31 +1,26 @@
...
     handleExceptions: true,
     format: combine(
       timestamp(),
       printf((info) => {
-        let { timestamp, level, error, ...args } = info;
-        if (error) {
+        if (info.error) {
           // If there is an error then strip out all punctuation to make it easily consumable by GCP within a log json
           // object. Note this error object is assumed to be a string, converted within Logger.js that uses this transport.
-          error = error
```
chrismaree, line 17: nice

packages/financial-templates-lib/src/logger/Logger.ts
```diff
@@ -74,10 +81,8 @@ function createNewLogger(injectedTransports = [], transportsConfig = {}, botIden
       winston.format.json()
     ),
     transports: [...createTransports(transportsConfig), ...injectedTransports],
-    exitOnError: process.env.EXIT_ON_ERROR ? process.env.EXIT_ON_ERROR : false,
+    exitOnError: !!process.env.EXIT_ON_ERROR,
```
chrismaree, line 84: this is new syntax to me! pretty cool: _"Converts Object to boolean. If it was falsey (e.g. 0, null, undefined, etc.), it will be false, otherwise, true.

"_

packages/financial-templates-lib/src/logger/PagerDutyTransport.ts
```diff
@@ -3,22 +3,36 @@
...
 
-    this.pd = new pdClient(pagerDutyOptions.pdApiToken);
+    this.pd = new PagerDutyClient(pdApiToken);
 
-    this.fromEmail = pagerDutyOptions.fromEmail;
-    this.defaultServiceId = pagerDutyOptions.defaultServiceId;
-    this.customServices = pagerDutyOptions.customServices || {};
+    this.fromEmail = fromEmail;
+    this.defaultServiceId = defaultServiceId;
+    this.customServices = customServices;
```
chrismaree, line 31: will this still work fine if `customServices` is empty? in that case, we need it to default to a blank object like `{}`. I think it'll do this from your defaults in the constructor but need to double-check as this is key to this transport still working in the case the user only uses the `defaultServiceId`

packages/financial-templates-lib/src/logger/SlackTransport.ts
```diff
@@ -68,20 +91,18 @@ function slackFormatter(info) {
       }
       // If the value in the message is an object then spread each key value pair within the object.
       else if (typeof info[key] === "object" && info[key] !== null) {
-        formattedResponse.blocks.push({ type: "section", text: { type: "mrkdwn", text: ` • _${key}_:\n` } });
+        // Note: create local reference to this object, so we can modify it in the if statement.
+        const newBlock: SectionBlock = { type: "section", text: { type: "mrkdwn", text: ` • _${key}_:\n` } };
+        formattedResponse.blocks.push(newBlock);
```
chrismaree, line 96: I'm a bit confused by how this works exactly. you are creating a `newBlock` locally here, then assigning stuff to it _after _ having pushed it into the `formattedResponce`. how is this updating the f`formattedResponse`?

packages/financial-templates-lib/src/logger/Transports.ts
```diff
@@ -30,28 +44,26 @@ function createTransports(transportsConfig = {}) {
     transports.push(new LoggingWinston());
   } else if (transportsConfig.createConsoleTransport != undefined ? transportsConfig.createConsoleTransport : true) {
     // Add a console transport to log to the console.
-    transports.push(ConsoleTransport.createConsoleTransport());
```
chrismaree, line 33: this is much cleaner.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
```diff
@@ -53,17 +53,17 @@ class CoinGeckoPriceFeed extends PriceFeedInterface {
     };
   }
 
-  async update() {
-    const currentTime = this.getTime();
```
chrismaree, line 57: hhmm. was this not using `await` before when calling this?

chrismaree: looks like the case in a few places. thanks ts!

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
```diff
@@ -587,11 +627,16 @@ function isMissingField(config, requiredFields, logger) {
...
+export async function getUniswapPairDetails(
+  web3: Web3,
+  syntheticTokenAddress: string,
+  collateralCurrencyAddress: string
+): Promise<{ address?: string; pairAddress?: string; inverted?: boolean }> {
   const networkId = await web3.eth.net.getId();
 
   if (process.env.UNISWAP_ADDRESS) {
     // Used for mock uniswap pair contracts.
+    // TODO: is address the wrong name for this field?
```
chrismaree, line 639: hhmm. how should we track this todo? feels like it will just get lost in the repo once this is merged.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -148,12 +147,13 @@ class CryptoWatchPriceFeed extends PriceFeedInterface {
       } seconds), open price: ${format(after?.openPrice)}.
         To see these prices, make a GET request to:
         https://api.cryptowat.ch/markets/${this.exchange}/${this.pair}/ohlc?after=${
-        before?.openTime + this.ohlcPeriod
+        (before?.openTime || 0) + this.ohlcPeriod
```
chrismaree, line 150: I'm trying to see if this impacts anything. a bad case would be if the feed fails and then defaults to 0 as a result of the `|| 0` but this is only in the `Error` mode. other than this change I really don't think there are any issues introduced in this critical feed.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.ts
```diff
@@ -117,33 +139,32 @@ class ExpressionPriceFeed extends PriceFeedInterface {
...
     return math.bignumber(price.toString()).div(decimalsMultiplier);
   }
 
   // Takes a math.bignumber OR math.ResultSet of math.bignumbers and converts it to a fixed point number that's
   // expected outside this library. Note: if the price is a ResultSet, the last value is converted and returned.
-  _convertToFixed(price, outputDecimals) {
+  private _convertToFixed(price: math.BigNumber | ResultSet<math.BigNumber>, outputDecimals: number): BN {
     // If the price is a ResultSet with multiple entires, extract the last one and make that the price.
-    if (price.entries) {
+    if ("entries" in price) {
```
chrismaree, line 163: 🙇

chrismaree (approved):
This is the mother of all PRs. I tried to read through every single line and look for things that could introduce problems. I _think_ we are pretty much good to go throughout. I left a few questions and points through this PR.

Awesome work @mrice32

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
```diff
@@ -22,81 +87,52 @@ class FinancialContractClient {
...
-    // Store the last on-chain time the clients were updated to inform price request information.
-    this.lastUpdateTimestamp = 0;
-
-    // Helper functions from web3.
-    this.toBN = this.web3.utils.toBN;
-    this.toWei = this.web3.utils.toWei;
+    this.financialContract = (new web3.eth.Contract(
+      financialContractAbi,
+      financialContractAddress
+    ) as unknown) as FinancialContractClient["financialContract"]; // Cast to web3-specific type
```
chrismaree, line 105: why can this not just be `as  ExpiringMultiPartyWeb3 | PerpetualWeb3`? There is similar syntax where you do look up with the `["..."]`. is there a reason you do this indirectly rather than just using the type directly?

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
mrice32: For whatever reason, eslint/the typescript compiler _only_ complain about this after changing the filename to `ts` haha, which is why I changed them.

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
mrice32: Yeah, the tests actually depend on the exact members, so I'm going to update them to be more lenient as this is much more convenient to keep the types consistent.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.js
mrice32: Yeah, exactly.

packages/financial-templates-lib/src/helpers/allowances.ts
mrice32: Agreed!

packages/financial-templates-lib/src/logger/Logger.ts
mrice32: Yeah, this doesn't matter in js since things are often used as booleans implicitly, but typescript really cares about stuff like this. If you call it a bool it needs to be a bool.

packages/financial-templates-lib/src/logger/SlackTransport.ts
mrice32: I looked this up when implementing to verify I was thinking about it correctly. When you push to an array, it pushes a _reference_ to the object to the array. So if you keep the object around and keep modifying it, the object within the array will change too (since they both reference the same underlying data structure). See https://stackoverflow.com/a/8661088.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
mrice32: yep, surprised they were working!!

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
mrice32: We should just figure it out before merging. I'll use blame to find the original PR to determine if address is ever being used or if this field name was just incorrect.

packages/financial-templates-lib/src/price-feed/ExpressionPriceFeed.ts
mrice32: yeah, typescript doesn't allow you to access a property that isn't guaranteed to exist 🙃

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
mrice32: No, you're right. I was just inconsistent. I'll fix this.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.ts
```diff
@@ -0,0 +1,144 @@
...
+  constructor({
+    logger,
+    perpetualAbi,
+    web3,
+    perpetualAddress,
+    multicallAddress,
+    getTime,
+    blockFinder,
+    minTimeBetweenUpdates = 60,
+    priceFeedDecimals = 18,
```
daywiss, line 66: you arent using the auto assign trick on these vars like you have in other classes

packages/financial-templates-lib/src/price-feed/LPPriceFeed.ts
```diff
@@ -29,7 +70,7 @@ class LPPriceFeed extends PriceFeedInterface {
     blockFinder,
     minTimeBetweenUpdates = 60,
     priceFeedDecimals = 18,
-  }) {
+  }: Params) {
```
daywiss, line 73: again these arent auto assigned, though this is just a nit

packages/financial-templates-lib/src/price-feed/MedianizerPriceFeed.ts
```diff
@@ -1,30 +1,31 @@
...
-    this.priceFeeds = priceFeeds;
-    this.computeMean = computeMean;
   }
 
   // Takes the median of all of the constituent price feeds' currentPrices.
-  getCurrentPrice() {
+  public getCurrentPrice(): BN | null {
     const currentPrices = this.priceFeeds.map((priceFeed) => priceFeed.getCurrentPrice());
-    if (currentPrices.some((element) => element === undefined || element === null)) {
+    if (!currentPrices.every(isDefined)) {
```
daywiss, line 28: nice simplification

packages/financial-templates-lib/src/types.ts
```diff
@@ -0,0 +1,14 @@
...
+export type Abi = AbiItem | AbiItem[];
+
+export type BN = ReturnType<typeof toBN>;
+
+export type Awaited<T> = T extends PromiseLike<infer U> ? Awaited<U> : T;
+
+export type FinancialContractType = "ExpiringMultiParty" | "Perpetual";
+export type FinancialContractFactoryType = "PerpetualCreator" | "ExpiringMultiPartyCreator";
+
+export function isDefined<T>(val: T | undefined | null): val is T {
```
daywiss, line 12: this is great, will probably update my function in the sdk to this

packages/financial-templates-lib/src/price-feed/QuandlPriceFeed.ts
```diff
@@ -127,34 +140,34 @@ class QuandlPriceFeed extends PriceFeedInterface {
...
     // Return early if the last call was too recent.
-    if (this.lastUpdateTime !== undefined && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+    if (this.lastUpdateTime !== null && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
       this.logger.debug({
         at: "QuandlPriceFeed",
         message: "Update skipped because the last one was too recent",
         currentTime: currentTime,
         lastUpdateTimestamp: this.lastUpdateTime,
-        timeRemainingUntilUpdate: this.lastUpdateTimes + this.minTimeBetweenUpdates - currentTime,
+        timeRemainingUntilUpdate: this.lastUpdateTime + this.minTimeBetweenUpdates - currentTime,
```
daywiss, line 170: wow nice bug fix

daywiss (commented):
wow, amazing job with this, what a huge chore.  Really impressed with the quantity of little bugs that were found during conversion.  i have mostly just nits, but will leave some time for responses before approving since pr is so large.

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
```diff
@@ -107,42 +143,43 @@ class FinancialContractClient {
...
 
   // Returns the last update timestamp.
-  getLastUpdateTime() {
+  public getLastUpdateTime(): number {
     return this.lastUpdateTimestamp;
   }
 
-  getLatestCumulativeFundingRateMultiplier() {
+  getLatestCumulativeFundingRateMultiplier(): BN {
+    if (!isDefined(this.latestCumulativeFundingRateMultiplier)) throw new Error("initialSetup hasn't been called");
```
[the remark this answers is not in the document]
mrice32: The generic type can be inferred. Typescript usually does this pretty well if it's easily inferable from one of the function params.

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
```diff
@@ -315,25 +361,28 @@ class FinancialContractClient {
...
   // To accommodate these different decimal points we can normalize each term to the 10^18 basis and then apply a
   // "correction" factor due to the additional scalling from multiplying basis numbers.
-  _isUnderCollateralized(numTokens, amountCollateral, tokenRedemptionValue) {
+  private _isUnderCollateralized(numTokens: string, amountCollateral: string, tokenRedemptionValue: string) {
     // Normalize the inputs. Now all terms are 18 decimal delimited and no extra conversion is needed.
     const normalizedNumTokens = this.normalizeSyntheticDecimals(numTokens);
     const normalizedAmountCollateral = this.normalizeCollateralDecimals(amountCollateral);
     const normalizedTokenRedemptionValue = this.normalizePriceFeedDecimals(tokenRedemptionValue);
 
+    if (!isDefined(this.collateralRequirement)) throw new Error("initialSetup hasn't been called");
```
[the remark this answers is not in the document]
mrice32: Makes sense. I was mainly just copying the surrounding code's style as to not change anything. But we can build official style recommendations around this if you think it makes sense!

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -1,7 +1,62 @@
...
+type SettleExpiredEvent = ExpiringMultiPartyWeb3Events.SettleExpiredPosition;
+type SettleEmergencyShutdownEvent = PerpetualWeb3Events.SettleEmergencyShutdown;
+type FundingRateUpdatedEvent = PerpetualWeb3Events.FundingRateUpdated;
+
+interface TransactionMetadata {
+  transactionHash: string;
+  blockNumber: number;
+}
+
+type EventExport<T extends { returnValues: any }> = T["returnValues"] & TransactionMetadata;
```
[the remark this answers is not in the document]
mrice32: Sure

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
```diff
@@ -158,7 +200,7 @@ class FinancialContractEventClient {
       finalFeeEventsObj,
       liquidationWithdrawnEventsObj,
       settleExpiredPositionEventsObj,
-    ] = await Promise.all([
+    ] = (await Promise.all([
```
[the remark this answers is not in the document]
mrice32: It's a good point. Not 100% sure. One reason might be that there's a hard limit on the number of events returned by a single event query in infura. So keeping the queries smaller will keep us from hitting that limit. Other than that, your suggestion seems reasonable!

packages/financial-templates-lib/src/clients/FinancialContractFactoryClient.ts
```diff
@@ -88,21 +105,31 @@ class FinancialContractFactoryClient {
...
-          ? event.returnValues.perpetualAddress
-          : event.returnValues.expiringMultiPartyAddress;
-      this.createdContractEvents.push({
-        transactionHash: event.transactionHash,
-        blockNumber: event.blockNumber,
-        deployerAddress: event.returnValues.deployerAddress,
-        contractAddress: createdAddress,
-      });
+    if (this.contractType === "ExpiringMultiPartyCreator") {
+      for (const event of (createdContractEventsObj as unknown) as ExpiringMultiPartyCreatorWeb3Events.CreatedExpiringMultiParty[]) {
```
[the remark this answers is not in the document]
mrice32: It's required by typescript if the types have "little overlap" 🤷

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -0,0 +1,249 @@
...
+    const [requestEvents, proposalEvents, disputeEvents, currentTime] = await Promise.all([
+      (this.oracle.getPastEvents("RequestPrice", { fromBlock: earliestBlockToQuery }) as unknown) as Promise<
+        OptimisticOracleWeb3Events.RequestPrice[]
+      >,
+      (this.oracle.getPastEvents("ProposePrice", { fromBlock: earliestBlockToQuery }) as unknown) as Promise<
+        OptimisticOracleWeb3Events.ProposePrice[]
+      >,
+      (this.oracle.getPastEvents("DisputePrice", { fromBlock: earliestBlockToQuery }) as unknown) as Promise<
+        OptimisticOracleWeb3Events.DisputePrice[]
+      >,
```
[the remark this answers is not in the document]
mrice32: In one of the other files, we ran into an issue where the typescript overloads for Promise.all were limited to 10 items. This required a very specific way of casting (outside the Promise). I typically prefer casting as early as possible

packages/financial-templates-lib/src/helpers/GasEstimator.ts
```diff
@@ -45,9 +83,9 @@ class GasEstimator {
   }
 
   // Calls update unless it was recently called, as determined by this.updateThreshold.
-  async update() {
+  async update(): Promise<void> {
     const currentTime = Math.floor(Date.now() / 1000);
-    if (currentTime < this.lastUpdateTimestamp + this.updateThreshold) {
+    if (this.lastUpdateTimestamp !== undefined && currentTime < this.lastUpdateTimestamp + this.updateThreshold) {
```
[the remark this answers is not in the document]
mrice32: Because typescript can get the inference right without the type guard, so there's no reason to make it more complicated than necessary IMO.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
```diff
@@ -53,17 +53,17 @@ class CoinGeckoPriceFeed extends PriceFeedInterface {
...
+  public async update(): Promise<void> {
+    const currentTime = await this.getTime();
 
     // Return early if the last call was too recent.
-    if (this.lastUpdateTime !== undefined && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+    if (this.lastUpdateTime !== null && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
```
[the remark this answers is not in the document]
mrice32: Yeah, I totally hear you. My goal is to _at least_ make the values that are initialized internally always be null rather than undefined. The codebase has a lot of overchecking because we don't know the types of things. Part of the advantage of typescript is to eliminate this by guaranteeing that the types have a fixed domain. I wanted to add some consistency to the way we build classes so that all internally managed values are initialized to null.

As for values that are passed to functions and constructors, you're right that this is a little more tricky because we don't know if the caller is using typescript. In this case, though, we're safe since this type is fully managed from within the class and typescript will protect against us assigning it to undefined.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -1,9 +1,25 @@
...
+import { NetworkerInterface } from "./Networker";
+import { BN } from "../types";
 
 // An implementation of PriceFeedInterface that uses CryptoWatch to retrieve prices.
-class CryptoWatchPriceFeed extends PriceFeedInterface {
+export class CryptoWatchPriceFeed extends PriceFeedInterface {
+  private readonly uuid: string;
+  private readonly toBN = Web3.utils.toBN;
+  private readonly convertPriceFeedDecimals: (number: number | string | BN) => BN;
+  private currentPrice: null | BN = null;
```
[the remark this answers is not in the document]
mrice32: Yeah, I thought I set them all to null. Please let me know if you see any that I missed.

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
```diff
@@ -809,26 +854,17 @@ async function createReferencePriceFeedForFinancialContract(
...
 }
 
-function getFinancialContractIdentifierAtAddress(web3, financialContractAddress) {
+function getFinancialContractIdentifierAtAddress(web3: Web3, financialContractAddress: string) {
   try {
     const ExpiringMultiParty = getTruffleContract("ExpiringMultiParty", web3, "1.2.0");
     return new web3.eth.Contract(ExpiringMultiParty.abi, financialContractAddress);
   } catch (error) {
-    throw new Error({ message: "Something went wrong in fetching the financial contract identifier", error });
+    throw new Error(`Something went wrong in fetching the financial contract identifier ${error}`);
```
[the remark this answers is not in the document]
mrice32: Hmmm, I will print error.stack instead.

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.ts
```diff
@@ -143,28 +141,26 @@ class CoinMarketCapPriceFeed extends PriceFeedInterface {
...
     return this.priceFeedDecimals;
   }
 
-  getLookback() {
+  getLookback(): number | null {
     return this.lookback;
   }
 
-  _invertPriceSafely(priceBN) {
+  private _invertPriceSafely(priceBN: BN | undefined) {
```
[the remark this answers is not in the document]
mrice32: Agreed. Will change to null.

packages/financial-templates-lib/src/price-feed/EthVixPriceFeed.ts
```diff
@@ -122,6 +128,14 @@ class ETHVIXPriceFeed extends PriceFeedInterface {
...
-module.exports = { ETHVIXPriceFeed };
+  public getLookback(): number {
+    return this.lastUpdateTime === null || this.historicalPrices.length === 0
+      ? 0
+      : this.lastUpdateTime - moment(this.historicalPrices[this.historicalPrices.length - 1].timestamp).unix();
+  }
+
+  public getPriceFeedDecimals(): number {
+    return this.priceFeedDecimals;
+  }
```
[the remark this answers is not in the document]
mrice32: They are required by the interface that all the price feeds inherit from. This is enforced in typescript (in js it isn't).

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.ts
mrice32: It's a subtlety. Because they are passed in an object, you can't auto-assign (as far as I'm aware). If you're aware of some way to do that here, let me know!

packages/financial-templates-lib/src/price-feed/LPPriceFeed.ts
mrice32: Same as above. I don't think it's possible (or at least I don't know the syntax) when the variables are passed as a single object param.

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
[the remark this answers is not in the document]
daywiss: cool, yea i dont think we need to take any action necessarily it would be too much to refact, and this should just be a direct translation

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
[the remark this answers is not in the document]
daywiss: very cool, did not know that

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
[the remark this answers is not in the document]
daywiss: yea sounds good, that makes sense.

packages/financial-templates-lib/src/price-feed/FundingRateMultiplierPriceFeed.ts
daywiss: oh interesting ok!

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
```diff
@@ -63,6 +63,11 @@ const versionedIt = function (supportedVersions, shouldBeItOnly = false) {
   return runTestForVersion(supportedVersions, TESTED_CONTRACT_VERSIONS, iterationTestVersion) ? it : () => {};
 };
 
+const objectsInArrayInclude = (subset, superset) => {
+  assert.equal(superset.length, subset.length);
+  for (let i = 0; i < superset.length; i++) assert.deepInclude(superset[i], subset[i]);
```
daywiss, line 68: i feel like one of these should not be indexed if you are using deepInclude, probably the one with the array of possible values

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
```diff
@@ -63,6 +63,11 @@ const versionedIt = function (supportedVersions, shouldBeItOnly = false) {
   return runTestForVersion(supportedVersions, TESTED_CONTRACT_VERSIONS, iterationTestVersion) ? it : () => {};
 };
 
+const objectsInArrayInclude = (subset, superset) => {
```
daywiss, line 66: just want to make sure this is right. you have `subset, superset` here, but you always pass what looks like the superset first (array of poassible values), and the actual values second when calling this.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
```diff
@@ -113,23 +118,23 @@ contract("OptimisticOracleClient.js", function (accounts) {
...
-    assert.deepStrictEqual(result, []);
+    objectsInArrayInclude(result, []);
     result = client.getSettleableProposals(proposer);
-    assert.deepStrictEqual(result, []);
+    objectsInArrayInclude(result, []);
 
     // Should have one price request.
     result = client.getUnproposedPriceRequests();
-    assert.deepStrictEqual(result, [
+    objectsInArrayInclude(result, [
```
daywiss, line 137: it seems like the parameters are reversed from the other comment

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
```diff
@@ -16,6 +16,11 @@ const Timer = getTruffleContract("Timer", web3);
 const Store = getTruffleContract("Store", web3);
 const MockOracle = getTruffleContract("MockOracleAncillary", web3);
 
+const objectsInArrayInclude = (superset, subset) => {
```
daywiss, line 19: shoudl probably pull this out as a utility if its used a lot

packages/financial-templates-lib/test/clients/OptimisticOracleEventClient.js
```diff
@@ -16,6 +16,11 @@ const Timer = getTruffleContract("Timer", web3);
 const Store = getTruffleContract("Store", web3);
 const MockOracle = getTruffleContract("MockOracleAncillary", web3);
 
+const objectsInArrayInclude = (superset, subset) => {
```
daywiss, line 19: same comments apply here

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
```diff
@@ -128,28 +129,26 @@ class CoinGeckoPriceFeed extends PriceFeedInterface {
...
+  public getLookback(): number {
     return this.lookback;
   }
 
-  _invertPriceSafely(priceBN) {
+  private _invertPriceSafely(priceBN: BN): BN | undefined {
     if (priceBN && !priceBN.isZero()) {
       return this.convertPriceFeedDecimals("1").mul(this.convertPriceFeedDecimals("1")).div(priceBN);
     } else {
       return undefined;
```
daywiss, line 151: this returns undefined rather than null

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.ts
```diff
@@ -1,7 +1,16 @@
...
+import type Web3 from "web3";
+import { NetworkerInterface } from "./Networker";
+import type { BN } from "../types";
+
+export class CoinMarketCapPriceFeed extends PriceFeedInterface {
+  private readonly uuid: string;
+  private readonly priceHistory: { time: number; price: BN }[];
+  private readonly convertPriceFeedDecimals: (number: number | string | BN) => BN;
+  private lastUpdateTime: number | null = null;
+  private currentPrice: BN | undefined;
```
daywiss, line 13: this is undefined ratehr than null

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
mrice32: I made an aliased type. Let me know if you like this better.

packages/financial-templates-lib/src/logger/SlackTransport.ts
mrice32: Does that make sense?

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
mrice32: This is wrong. Will update the variable name.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
mrice32: I think this is implicitly what it was doing before since it was previously passing undefined or null to an arithmetic operation. I can change the logic to something else if you think there's something else we should do if this is null/undefined?

Thoughts?

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
mrice32: Fixed.

packages/financial-templates-lib/src/clients/FinancialContractEventClient.ts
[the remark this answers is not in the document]
mrice32: Done.

packages/financial-templates-lib/src/price-feed/BasketSpreadPriceFeed.ts
```diff
@@ -19,34 +29,32 @@ class BasketSpreadPriceFeed extends PriceFeedInterface {
...
-    // Helper modules.
-    this.web3 = web3;
-    this.toBN = this.web3.utils.toBN;
-    this.toWei = this.web3.utils.toWei;
-    this.logger = logger;
-
     // The precision that the user wants to return prices in must match all basket constituent price feeds and the denominator.
-    this.decimals = this.allPriceFeeds[0].getPriceFeedDecimals();
+    const decimals = this.allPriceFeeds[0].getPriceFeedDecimals();
+    if (decimals === null) throw new Error(`BasketSpreadPriceFeed -- first price feed has null decimals`);
```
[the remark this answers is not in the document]
mrice32: No. The return type of the getPriceFeedDecimals function is defined as null | number. So we should only need to check against null (assuming no price feeds are lying about the types they return).

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
```diff
@@ -481,59 +503,69 @@ async function createPriceFeed(logger, web3, networker, getTime, config) {
...
-          })
+          )
         )
-      ).filter((el) => el !== null)
+      ).filter(isDefined)
     );
 
     // Return null if any of the price feeds in the map are null (meaning there was an error).
-    if (Object.values(priceFeedMap).some((priceFeed) => priceFeed === null)) return null;
+    if (!Object.values(priceFeedMap).every(isDefined)) return null;
```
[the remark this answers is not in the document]
mrice32: There's a subtlety here. To simplify, the inner most function is returning `[string, PriceFeed | null] | null`. The first filter drops the outer null, the second check is checking for the inner null. The distinction is important because a failure to find a price feed for a symbol in the expression isn't necessarily an error (the symbol could be a built-in or resolved in some other way). If we try to generate an inner price feed and it fails and then returns null, we want to propagate that error by continuing to return null up the stack. 

In hindsight, we should probably throw instead of returning nulls, but that's the pattern we used 🤷 .

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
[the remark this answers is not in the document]
mrice32: Done.

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.ts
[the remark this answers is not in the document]
mrice32: Fixed.

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
mrice32: Hmm, what do you mean. The expectation is that you have an array of subsets and supersets and each subsequent subset should have each of its values in the corresponding superset.

I may be missing something, though.

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
mrice32: It's actually the opposite I think. The values we explicitly write in the test may be a subset of the values produced by the implementation (aka the implementation may have more values than specified in the test and that's okay). Does that make sense? Or did I misunderstand your comment.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
mrice32: Yeah, different tests had used different orderings of these values. That's why the order of the args is swapped in one file. I can go through and correct this -- I just wanted to avoid deeper modifications of these tests.

Thoughts?

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
mrice32: The issue is (as stated above) that different tests order their args differently. I can refactor this if you think it's important, was just trying to avoid unnecessary changes.

packages/financial-templates-lib/test/clients/OptimisticOracleEventClient.js
mrice32: See comments above, and let me know your thoughts.

packages/financial-templates-lib/src/price-feed/CoinGeckoPriceFeed.ts
mrice32: Fixed.

packages/financial-templates-lib/src/price-feed/CoinMarketCapPriceFeed.ts
mrice32: Fixed.

mrice32:
@chrismaree @daywiss I think I've fixed and/or responded to all comments. Please let me know if there are any additional changes you'd like me to make or if you find any of the fixes/responses unsatisfactory!

packages/financial-templates-lib/src/clients/FinancialContractClient.ts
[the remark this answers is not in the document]
chrismaree: ye, I agree we should standardize. @daywiss has called me out on this before and I think it's a great point to make as it can make debugging harder if we do things differently, over and above a desire for just general consistency.

Case and point are how we catch errors in the serverless hub. ideally, we throw the same way everywhere so we can catch the same way (in a perfect world we can catch all types of errors but that's not always the case).

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
chrismaree: looks much better now, thanks!

packages/financial-templates-lib/src/logger/PagerDutyTransport.ts
chrismaree: looking at it again, I agree. the defaults make this work fine.

packages/financial-templates-lib/src/logger/SlackTransport.ts
chrismaree: It does make sense. I hate that that's a thing haha. Really not a fan of how JS treats this differently depending on the type. It might be useful to add a comment about this pushing in a _pointer_ and that means you are modifying the original data.

mrice32: Will add a comment!

packages/financial-templates-lib/src/price-feed/CreatePriceFeed.ts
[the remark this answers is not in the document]
daywiss: yea thanks for the explanation

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
daywiss: yea that makes sense

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
daywiss: yea i misunderstood what was happening here, after you explained it, and looking at chai docs again it makes sense.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
daywiss: np just wantd to call it out incase it was not intended

daywiss (approved):
🥇

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
chrismaree: no, I think it's all good and does not introduce anything of issue. After looking at it again I think we are good to go. Was just trying to think of potential issues this could introduce.

chrismaree: no, I think it's all good and does not introduce anything of issue. After looking at it again I think we are good to go. Was just trying to think of potential issues this could introduce.

PR #3320, opened 2021-08-18
feat(ancillary): first pass at an ancillary data parser in typescript

pemulis:

**Motivation**

We need a generalized ancillary data parser in `@uma/common` so that bots and front-end interfaces can easily convert on-chain data into something useful for display or processing.

**Summary**

Converted a JavaScript parser that @Reinis-FRP wrote and has been using into a TypeScript version, with tests.

**Details**

The first application of the generalized parser will be the optimistic oracle bot, but the goal is to be able to use this pretty much everywhere.

There are currently tests for successful cases based on real ancillary data or UMIP example data, with more tests incoming for edge cases where the parser should throw an error or parse stringified data in a certain way.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+  keyValues.forEach((keyValue: any) => {
+    const [key, value] = parseKeyValue(keyValue);
+    ancillaryObject[key] = value;
+  });
+  return ancillaryObject;
+}
+
+// Escapes double quoted keys/values and values enclosed in curly/square brackets.
+function markEscapes(stringObject: any) {
+  stringObject.forEach((charObject: any, openIndex: number, stringObject: any) => {
```
mrice32, line 43: You don't _always_ have to provide types for these types of functions. Often times, typescript can infer the types without you having to explicitly specify them. Its ability to do this is not 100%, so you won't _always_ be able to rely on this.

In this case, assuming stringObject has the correct type, it should infer (and you can check this by hovering over in VSCode):
````ts
(charObject: { character: string; escape: boolean; skip: false }, openIndex: number, stringObject: { character: string; escape: boolean; skip: false }[]) => {
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+    // Skip searching in already escaped characters or closing double quotes:
+    if (charObject.escape || charObject.skip) return;
+
+    // Escape keys: opening quotes should be after comma (,) separator or start.
+    if (
+      charObject.character == '"' &&
```
mrice32, line 49: nit: prefer `===`

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+      escapeJSON(stringObject, openIndex);
+
+    // Escape JSON values: first opening square brackets should be after column (:) separator.
+    if (charObject.character == "[" && isNextChar(stringObject, openIndex, ":", false))
+      escapeJSON(stringObject, openIndex, false);
+  });
+}
+
+// Splits ancillary data object into key-value pairs.
+function splitKeyValues(stringObject: any): any {
```
mrice32, line 69: nit: similar to other comments, I think we can type this more explicitly than any.

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+  const keyValues: any = [];
+  for (let startIndex = 0; startIndex < stringObject.length; startIndex++) {
+    const charObject: any = stringObject[startIndex];
```
mrice32, line 72: nit: any

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+        endIndex == stringObject.length - 1 ||
+        isNextEnd(stringObject, endIndex) ||
+        isNextChar(stringObject, endIndex, ",")
+      ) {
+        // Copy the identified key-value pair:
+        const pairIndex = keyValues.length;
+        keyValues.push([]);
+        for (let i = startIndex; i <= endIndex; i++) {
+          keyValues[pairIndex].push(stringObject[i]);
+        }
```
mrice32, line 89:
```suggestion
        keyValues.push(stringObject.slice(startIndex, endIndex));
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+  for (let startIndex = 0; startIndex < stringObject.length; startIndex++) {
+    const charObject: any = stringObject[startIndex];
+
+    // If reached unescaped comma (,) continue with the next key-value pair:
+    if (!skipWhitespace(charObject) || (charObject.character == "," && !charObject.escape)) continue;
+
+    for (let endIndex = startIndex; endIndex < stringObject.length; endIndex++) {
+      // Search for next unescaped comma (,) delimiter or end of object:
+      if (
+        endIndex == stringObject.length - 1 ||
```
mrice32, line 80:
```suggestion
        endIndex === stringObject.length - 1 ||
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  // Remove enclosing double quotes.
+  keyValues.forEach((keyValue: any, index: number, keyValues: any) => {
+    keyValues[index] = keyValue.filter(removeDoubleQuotes);
+  });
+  return keyValues;
+}
+
+// Tries to parse key:value pair.
+function parseKeyValue(keyValue: any): any {
```
mrice32, line 106: nit: can we make these types more specific?

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+
+        // Skip start index to the end of current key-value pair:
+        startIndex = endIndex;
+        break;
+      }
+    }
+  }
+
+  // Remove enclosing double quotes.
+  keyValues.forEach((keyValue: any, index: number, keyValues: any) => {
```
mrice32, line 99: nit: more specific types

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+  return keyValues;
+}
+
+// Tries to parse key:value pair.
+function parseKeyValue(keyValue: any): any {
+  let key = "";
+  let value = "";
+
+  // Skip unescaped whitespace:
+  let index = keyValue.findIndex(skipWhitespace) == -1 ? keyValue.length : keyValue.findIndex(skipWhitespace);
```
mrice32, line 111:
```suggestion
  let index = keyValue.findIndex(skipWhitespace) === -1 ? keyValue.length : keyValue.findIndex(skipWhitespace);
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+
+  while (index < keyValue.length) {
+    const skip =
+      keyValue.slice(index).findIndex(skipWhitespace) == -1 ? 0 : keyValue.slice(index).findIndex(skipWhitespace);
```
mrice32, line 115:
```suggestion
      keyValue.slice(index).findIndex(skipWhitespace) === -1 ? 0 : keyValue.slice(index).findIndex(skipWhitespace);
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    // Reached unescaped column (:) delimiter:
+    if (keyValue[index + skip].character == ":" && !keyValue[index + skip].escape) {
```
mrice32, line 118:
```suggestion
    if (keyValue[index + skip].character === ":" && !keyValue[index + skip].escape) {
```

mrice32 (commented):
Looks great! I reviewed about half, but I realized a lot of my comments were starting to become repetitive.

The biggest issue I see is the use of `any`. We don't really get much out of typescript if we use `any`. To put it differently, if you use `any` everywhere typescript is just javascript.

There are times to use `any`, but they are only really if you have some type that _really_ could be anything and it's coming from somewhere outside your code (usually). But most types you know _something_ about. For instance, a very general object type is:
```
type SomeObject = { [key: string]: any };
```
This tells typescript that the object can be indexed by strings and that it _isn't_ null, undefined, a string, a number, etc. Interestingly, this is _a lot_ more information than `any`. This means that typescript can validate that you never set this to null or undefined and that other developers don't need to check against null or undefined (this is a big one).

From my basic understanding of this implementation, though, most of these variables have very well-defined types. So we should be able to be a lot more specific than that SomeObject type above. We should be able to avoid the use of any even within our object types, but I could be wrong.

packages/common/src/AncillaryDataParser.ts
pemulis: I'm not sure in this case, since all of the following are valid key-value pairs:

`{ key: 'value' }`
`{ key1: 'value1', key2: 'value2' }`
`{ '  key ': '  value  ' }`
`{ '"  key': 'value  "' }`
`{ key: { nestedKey: 'nestedValue' } }`
`{ key: '{nestedKey: "nestedValue",}' }`
`{ key: [ 1, 'a', null, true, {}, [] ] }`

Testing for the specific disallowed cases (key with no value, empty key, multiple column delimiters in key-value pairs, JSON object in the key, random hex) might be better than trying to strictly define what _is_ allowed through types?

pemulis: Thought a bit more, and it appears `keyValue` should be `CharObject[]` and `keyValues` should be `[CharObject[]]`. Running with that and seeing how it goes.

pemulis:
> Looks great! I reviewed about half, but I realized a lot of my comments were starting to become repetitive.
> 
> The biggest issue I see is the use of `any`. We don't really get much out of typescript if we use `any`. To put it differently, if you use `any` everywhere typescript is just javascript.
> 
> There are times to use `any`, but they are only really if you have some type that _really_ could be anything and it's coming from somewhere outside your code (usually). But most types you know _something_ about. For instance, a very general object type is:
> 
> ```
> type SomeObject = { [key: string]: any };
> ```
> 
> This tells typescript that the object can be indexed by strings and that it _isn't_ null, undefined, a string, a number, etc. Interestingly, this is _a lot_ more information than `any`. This means that typescript can validate that you never set this to null or undefined and that other developers don't need to check against null or undefined (this is a big one).
> 
> From my basic understanding of this implementation, though, most of these variables have very well-defined types. So we should be able to be a lot more specific than that SomeObject type above. We should be able to avoid the use of any even within our object types, but I could be wrong.

Made a bunch of updates! Thanks for all of the feedback.

One important note is that all of the following are valid key-value sets:

```
{ key: 'value' }
{ key1: 'value1', key2: 'value2' }
{ ' key ': ' value ' }
{ '" key': 'value "' }
{ key: { nestedKey: 'nestedValue' } }
{ key: '{nestedKey: "nestedValue",}' }
{ key: [ 1, 'a', null, true, {}, [] ] }
```

I've used `Record<string, unknown>` for the ancillary object returned by the parser, which was the recommended replacement for the `object` type, with `unknown` being a replacement for `any`.

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,256 @@
+import Web3 from "web3";
+import dotenv from "dotenv";
+dotenv.config();
```
Reinis-FRP, line 3: Sorry, seems like I left dotenv when writing this locally. Is it still needed as part of common utils?

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,256 @@
...
+}
+
+/**
+ * Finds closing quotes for keys/values and marks escaped.
+ * For values: closing quotes should be either before comma (,) or at the end.
+ * For keys: closing quotes should be before column (:).
+ */
+function escapeQuotes(stringObject: CharObject[], openIndex: number, escapeValues = true) {
+  const nextCharFn = escapeValues
+    ? function (stringObject: CharObject[], closeIndex: number) {
```
Reinis-FRP, line 208: Just learning myself on TS - why do we not set return type of boolean for this function? The same question applies to escaping keys (escapeValues = false) case below and `skipWhitespace` and `removeDoubleQuotes` that returns boolean.

packages/common/src/AncillaryDataParser.ts
pemulis: Nope, looks like we can drop it.

packages/common/src/AncillaryDataParser.ts
pemulis: Good catch! Making the return values more explicit.

chrismaree:
It looks like your PR is failing due to needing to re-base master to fix the e2e tests. I checked out your branch, updated it to master and re-pused. hopefully this sorts things out!

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,254 @@
...
+ * @notice Ancillary data parser implementation following guidelines at:
+ * https://docs.google.com/document/d/1zhKKjgY1BupBGPPrY_WOJvui0B6DMcd-xDR8-9-SPDw/edit
+ * @param {String} hex string representation of ancillaryData
+ * @return {Object} parsed ancillary data object.
+ */
+export function parseAncillaryData(ancillaryData: string): Record<string, unknown> {
+  // Some requesting contracts set the synthetic token address as ancillary data, so try to parse it first:
+  if (Web3.utils.isAddress(ancillaryData)) return { address: ancillaryData };
+  let ancillaryString;
+  try {
```
chrismaree, line 20: should this throw if it cant use `Web3.utils.hexToUtf8`? it seems like if it is an `AncillaryString` it's actually a pretty normal flow and should not throw getting here.

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,254 @@
...
+/**
+ * @title Parse ancillary data.
+ * @notice Ancillary data parser implementation following guidelines at:
+ * https://docs.google.com/document/d/1zhKKjgY1BupBGPPrY_WOJvui0B6DMcd-xDR8-9-SPDw/edit
+ * @param {String} hex string representation of ancillaryData
+ * @return {Object} parsed ancillary data object.
+ */
+export function parseAncillaryData(ancillaryData: string): Record<string, unknown> {
+  // Some requesting contracts set the synthetic token address as ancillary data, so try to parse it first:
+  if (Web3.utils.isAddress(ancillaryData)) return { address: ancillaryData };
```
chrismaree, line 18: is there any case that this will return the value when it is _not_ an address? this simply validates that: a) this is hex. b) the length is correct for an ethereum address c) the checksum on the case matches. is there a scenario where this can return for something that matches these criteria but is not an address?

packages/common/test/AncillaryDataParser.js
```diff
@@ -0,0 +1,150 @@
+// Script to test
+const AncillaryDataParser = require("../dist/AncillaryDataParser");
+
+contract("AncillaryDataParser.js", function () {
+  describe("parseAncillaryData", function () {
+    it("parses simple key-value data correctly", async function () {
+      const data = web3.utils.utf8ToHex("key:value");
+      const expectedObject = { key: "value" };
+      const parsedData = AncillaryDataParser.parseAncillaryData(data);
+      assert.equal(JSON.stringify(parsedData).toLowerCase(), JSON.stringify(expectedObject).toLowerCase());
```
chrismaree, line 10: why do you need to cast this to lowercase?

packages/common/src/AncillaryDataParser.ts
nicholaspai: > is there a scenario where this can return for something that matches these criteria but is not an address?

I think this is an assumption we can make by using `Web3.utils.isAddress`

Reinis-FRP (approved):

packages/common/src/AncillaryDataParser.ts
pemulis: Agreed with @nicholaspai that this assumption seems reasonable, and I'm not aware (per @Reinis-FRP's point) of any use case where `Web3.utils.isAddress` would not be adequate as a check or we could receive some ancillary data that passes the `isAddress` check but is not actually an address.

packages/common/test/AncillaryDataParser.js
pemulis: I don't remember why these conversions were in there originally but I removed them and everything was fine.

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,254 @@
...
+        break;
+      }
+    }
+  }
+
+  // Remove enclosing double quotes.
+  keyValues.forEach((keyValue: CharObject[], index: number) => {
+    keyValues[index] = keyValue.filter(removeDoubleQuotes);
+  });
+  return keyValues;
```
mrice32, line 101:
```suggestion
  return keyValues.map((keyValue: CharObject[]) => keyValue.filter(removeDoubleQuotes));
```

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,254 @@
...
+ * @return {Object} parsed ancillary data object.
+ */
+export function parseAncillaryData(ancillaryData: string): Record<string, unknown> {
+  // Some requesting contracts set the synthetic token address as ancillary data, so try to parse it first:
+  if (Web3.utils.isAddress(ancillaryData)) return { address: ancillaryData };
+  let ancillaryString;
+  try {
+    ancillaryString = Web3.utils.hexToUtf8(ancillaryData);
+  } catch (err) {
+    throw "Cannot parse ancillary data bytes to UTF-8!";
```
mrice32, line 23:
```suggestion
    throw new Error("Cannot parse ancillary data bytes to UTF-8!");
```

mrice32 (approved):
Looks great! Glad to see lots of good tests!!

packages/common/src/AncillaryDataParser.ts
```diff
@@ -0,0 +1,254 @@
...
+};
+
+/**
+ * @title Parse ancillary data.
+ * @notice Ancillary data parser implementation following guidelines at:
+ * https://docs.google.com/document/d/1zhKKjgY1BupBGPPrY_WOJvui0B6DMcd-xDR8-9-SPDw/edit
+ * @param {String} hex string representation of ancillaryData
+ * @return {Object} parsed ancillary data object.
+ */
+export function parseAncillaryData(ancillaryData: string): Record<string, unknown> {
```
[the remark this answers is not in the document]
pemulis: I'm also new to TypeScript, but that matches my understanding: `unknown` forces the caller to cast to a type, but doesn't specify in advance what the type will be.

PR #3323, opened 2021-08-20
feat(insured-bridge): clean comments, address some todos and general code hygiene

chrismaree:
**Motivation**

This PR cleans up some stuff in the insured bridge contracts, addresses some outstanding todos and adds some additional code hygiene.

The main(only) logic change in this PR is the introduction of a new state variable in the bridge pool called `pendingReserves`. This variable acts to track _eserves that are not yet utilized but are pre-allocated for a pending relay_. i.e if a relay action has happened, which has used funds, but it is not yet finalized, an LP should not be able to pull out their funds if it will prevent the relay action from finalizing.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
@@ -58,7 +58,7 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, OVM_CrossDomainEnabled {
      * @notice Construct the Bridge Admin
      * @param _finder DVM finder to find other UMA ecosystem contracts.
      * @param _crossDomainMessenger Optimism messenger contract used to send messages to L2.
-     * @param _optimisticOracleLiveness Timeout that all bridging actions from L2->L1 must wait for a OO response.
+     * @param _optimisticOracleLiveness Timeout that all bridging actions from L2->L1 must wait for a OptimisticOracle response.
      * @param _proposerBondPct Percentage of the bridged amount that a relayer must put up as a bond.
      * @param _identifier Identifier used when querying the OO for a cross bridge transfer action.
      */
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
@@ -54,8 +54,14 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, OVM_CrossDomainEnabled {
         _;
     }
 
-    // TODO: Consider switching to hardcoded OVM_L1CrossDomainMessenger:
-    // https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/deployments/README.md
+    /**
+     * @notice Construct the Bridge Admin
+     * @param _finder DVM finder to find other UMA ecosystem contracts.
+     * @param _crossDomainMessenger Optimism messenger contract used to send messages to L2.
+     * @param _optimisticOracleLiveness Timeout that all bridging actions from L2->L1 must wait for a OO response.
+     * @param _proposerBondPct Percentage of the bridged amount that a relayer must put up as a bond.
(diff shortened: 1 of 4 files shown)


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -238,17 +265,22 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-        // the timestamps on the L2 have an offset that are always "in the future" relative to L1 blocks, then the
-        // OptimisticOracle would always reject requests.
+        // (proposer reward + proposer bond + final fee) from the caller. We need to set a new price request timestamp
+        // instead of default setting to equal to the `depositTimestamp`, which is dependent on the L2 VM on which the
+        // DepositContract is deployed. Imagine if the timestamps on the L2 have an offset that are always "in the
+        // future" relative to L1 blocks, then the OptimisticOracle would always reject requests.
         _requestOraclePriceRelay(amount, priceRequestTime, getRelayAncillaryData(depositData, relayData));
         _proposeOraclePriceRelay(amount, priceRequestTime, getRelayAncillaryData(depositData, relayData));
 
+        pendingReserves += amount; // Book off maximum liquidity used by this relay in the pending reserves.
```
mrice32, line 282: Is there a reason we need to track this? If we add this to utilized reserves the moment the relay occurs (I assume we're doing this already), can we just drop this variable?

mrice32 (approved):
This looks great! One small question.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: the reason we need to track this separately(and cant use `utalizedReserves`) is that `utilizedReseres` is only incremented when the reserves are actually used. This happens when calling `settleRelay`. At this point, money from the pool is _used_. `pendingReserve` indicates funds from `liquidReserves` that are _pending_ to be used but are currently not used. I.e a relay action has _started_ (due to someone calling `relayDeposit`) but we are waiting for either a) OO liveness or b) someone to speed up the relay.

What this aims to do is prevent the case where a transfer is relayed, at which time the contract _thinks_ it has enough collateral to finish the relay and during the liveness period the LPs remove liquidity such that the relay cant be finished.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
```diff
@@ -54,8 +54,14 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, OVM_CrossDomainEnabled {
         _;
     }
 
-    // TODO: Consider switching to hardcoded OVM_L1CrossDomainMessenger:
-    // https://github.com/ethereum-optimism/optimism/blob/develop/packages/contracts/deployments/README.md
+    /**
+     * @notice Construct the Bridge Admin
+     * @param _finder DVM finder to find other UMA ecosystem contracts.
+     * @param _crossDomainMessenger Optimism messenger contract used to send messages to L2.
+     * @param _optimisticOracleLiveness Timeout that all bridging actions from L2->L1 must wait for a OO response.
```
nicholaspai, line 61: nit

```suggestion
     * @param _optimisticOracleLiveness Timeout that all bridging actions from L2->L1 must wait for a OptimisticOracle response.
```

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I think this is a reasonable solution. The alternative would be to move the tokens to a separate intermediary contract, which would make accounting easier, but be more expensive gas-wise.

mrice32: You're totally right. This makes sense, and I can't think of any other way to take this into account without an additional variable. Thanks for the thorough response!!

mrice32 (approved):
LGTM!

PR #3324, opened 2021-08-24
feat(insured-oracle): Added initial client implementation and tests

chrismaree:
**Motivation**

As part of building out the insured oracle, we require a series of clients to fetch information from L1 and L2 and present to the bots and price feed. This PR adds these clients & unit tests.

**Summary**

Adds clients and tests for the insured bridge.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/.mocharc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
```
chrismaree, line 1: taken from https://github.com/UMAprotocol/protocol/pull/3315 to minimize merge conflicts.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -347,7 +349,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         updateAccumulatedLpFees();
         allocateLpFees(_getAmountFromPct(relay.realizedLpFeePct, _depositData.amount));
 
-        emit SettledRelay(depositHash, keccak256(getRelayAncillaryData(_depositData, relay)), msg.sender);
```
chrismaree, line 350: re-named to be consistent with other events

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -89,21 +89,22 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     event DepositRelayed(
         uint64 depositId,
         address indexed sender,
+        address slowRelayer,
```
chrismaree, line 92: added to enable the full state re-creation from events (otherwise we would need to poll data from the client)

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -191,7 +193,7 @@ describe("BridgePool", () => {
     ).send({ from: owner });
     await bridgeAdmin.methods.setDepositContract(depositContractImpersonator).send({ from: owner });
 
-    // New BridgePool linked to BridgeFactory
```
chrismaree, line 194: lingering re-naming

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -432,11 +434,15 @@ describe("BridgePool", () => {
         return (
           ev.depositId.toString() === depositData.depositId.toString() &&
           ev.sender === depositData.l2Sender &&
+          ev.slowRelayer === relayer &&
```
chrismaree, line 437: new param in event

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -432,11 +434,15 @@ describe("BridgePool", () => {
...
           ev.depositTimestamp === depositData.depositTimestamp &&
           ev.recipient === depositData.recipient &&
           ev.l1Token === depositData.l1Token &&
           ev.amount === depositData.amount &&
-          ev.maxFeePct === depositData.maxFeePct &&
+          ev.slowRelayFeePct === depositData.slowRelayFeePct &&
```
chrismaree, line 442: previously we were testing `maxFeePct`! this is null both within the event and locally (it was a previous implementation that has been refactored away). I changed it to correctly validate this even behavour.

packages/financial-templates-lib/src/types.ts
```diff
@@ -0,0 +1,14 @@
+import type { AbiItem, toBN } from "web3-utils";
```
chrismaree, line 1: taken from https://github.com/UMAprotocol/protocol/pull/3315 to minimize merge issues.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -89,21 +89,22 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
         uint64 depositTimestamp,
         address recipient,
         address l1Token,
         uint256 amount,
         uint64 slowRelayFeePct,
         uint64 instantRelayFeePct,
-        uint64 maxLpFeePct,
+        uint64 quoteTimestamp,
```
nicholaspai, line 99: great catch, thanks

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -0,0 +1,181 @@
...
+    // Define a config to bound the queries by.
+    const blockSearchConfig = {
+      fromBlock: this.firstBlockToSearch,
+      toBlock: this.endingBlockNumber || (await this.l1Web3.eth.getBlockNumber()),
+    };
+
+    if (!this._isClientInitialized()) await this._initialSetup(blockSearchConfig.fromBlock);
+
+    // Fetch event information
+    // TODO: consider optimizing this further. Right now it will make a series of sequential BlueBird calls for each pool.
```
nicholaspai, line 104: Aren't these calls made in parallel?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -0,0 +1,181 @@
...
+          slowRelayFeePct: depositRelayedEvent.returnValues.slowRelayFeePct,
+          instantRelayFeePct: depositRelayedEvent.returnValues.instantRelayFeePct,
+          realizedLpFeePct: depositRelayedEvent.returnValues.realizedLpFeePct,
+          priceRequestAncillaryDataHash: depositRelayedEvent.returnValues.priceRequestAncillaryDataHash,
+          depositHash: depositRelayedEvent.returnValues.depositHash,
+          depositContract: depositRelayedEvent.returnValues.depositContract,
+          relayState: relayState.Pending,
+        };
+
+        // If the local data contains this deposit ID then this is a re-relay of a disputed relay. In this case, we need
```
nicholaspai, line 138: > If the local data contains this deposit ID then this is a re-relay of a disputed relay.

This assumption only holds if the `depositRelayedEvents` are sorted from oldest to newest right? Do you enforce this sorting anywhere?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -0,0 +1,181 @@
...
+        // to update the data accordingly as well as store the previous slow relayers.
+        if (this.relays[l1Token][relayData.depositHash]) {
+          // Bring the previous slow relayer from the data from the previous run and store it.
+          const previousSlowRelayer = this.relays[l1Token][relayData.depositHash].slowRelayer;
+          this.relays[l1Token][relayData.depositHash].disputedSlowRelayers.push(previousSlowRelayer);
```
nicholaspai, line 143: Do you anticipate it being useful to know the prior `disputedSlowRelayers` for a given relay?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -0,0 +1,181 @@
...
+          this.relays[l1Token][relayData.depositHash].slowRelayer = relayData.slowRelayer;
+        }
+        // Else, if this if this is the first time we see this deposit hash, then store it.
+        else this.relays[l1Token][relayData.depositHash] = relayData;
+      }
+
+      // For all RelaySpedUp, set the instant relayer. Note that if a relay action was disputed the instant
```
nicholaspai, line 150: Did you intend to finish this sentence?

nicholaspai (commented):
Great start, one high level question for me. How do you intend for the L1 and L2 clients to be constructed by an upstream pricefeed?

Would this require changing the `getWeb3` function to take in multiple networks?

cc @mrice32

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: the fetches within the `Promise.all` are parallel but there are a sequential set of parallel calls for each `bridgePool`. Put differently, if there are 5 bridge pools then there are 5 sequential sets of 4 parallel calls.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: This data comes from the `getPastEvents` call which will have the ordering in which the events were created. I dont think any issue is introduced through by this having a diffrent ordering? how do you think this would impact if the order was wrong, seeing we use the `depositHash` as the key in this object?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: it _might be_. happy to remove. if we don't need the data. I thought there could be a case where the bot needs to know this when settling disputes but I don't think it _needs_ this data. I was rather trying to build up complete information about a given request. without this you can't tell how many times a relay has been re-relayed and disputed (if that was to happen multiple times).

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: yup. will clean it up! was part of the draft still :)

chrismaree:
@nicholaspai >> https://github.com/UMAprotocol/protocol/pull/3324#pullrequestreview-737766917

The way I was thinking the price feed (and bots in general) would consume this would be as follows: we have two kinds of bots that need to use these clients: 1) the relay bot that does bridging actions from L2-> L1 and 2) the OO proposer/disputer bot that validates bot 1 and will dispute if an invalid action is done. 

For the relayer, this bot would contain two clients within it (just like the liquidator/monitor etc has clients in it). these clients would be updated within the execution loop when the bot runs. Then, this bot would feed inactive requests on L1/L2 (stuff that is not yet finalized) one a time into the price feed via a method. This feed would then make a decision on the validity of the data presented to it. This is somewhat similar to how the liquidator will make a decision based off a feed to do some action (i.e liquidate). the difference is we will now be _feeding in_ data to this feed, from the client within the bot.

For the OO bot, we wont have two initialized clients (the OO should not need to have any "warming up" for a data feed). rather, the OO should only create a new feed iff a price request comes into the OO. In this case, the OO will ask the price feed about a particular price request + timestamp + ancillary data. At this point, the feed will actually initialize _its own_ clients that will live within this feed for the duration of this price request. Using this client, the feed will figure out the correctness of the request and return to the OO.

From an implementation perspective, we can think about how the feed can "re-use" these clients. The basic framing that I have right now is that the feed will have a param that is fed in when creating the feed _if_ you have the clients already (as such in the relayer bot). If you don't have any clients, like in the OO bot, then the creator will make the clients first and pass them into the feed. This switching and construction logic will happen _outside_ of the feed, within the `createPriceFeed` utility. In doing it this way, the feeds themselves will _always_ have clients passed in (either from the relayer bot, passed via `createPriceFeed` or as "new" clients, for the OO bot, as created in `createPriceFeed`).

Does that framing make sense to you?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: agreed

nicholaspai: comment was a bit confusing.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: the `slowRelayer` of the last `event` in a set of duplicate `depositRelayedEvent` will be set as the `slowRelayer`. Normally it should be that the last `slowRelayer` is the non-disputed relayer, but this could be mixed up if the order is different?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: > without this you can't tell how many times a relay has been re-relayed and disputed (if that was to happen multiple times).

This is true, but I'm not sure its useful. We can keep for now but maybe add a TODO to remove it if not used. To call `settleRelay` you only need the `DepositData`

PR #3325, opened 2021-08-24
improve(sdk): add support for Multipart2 contract

amateima:
**Motivation**

https://app.clubhouse.io/uma-project/story/2128/add-multicall-2-support-to-multicall-sdk-library

**Summary**

- add Multipart2 contract interface & types

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/core/contracts/common/interfaces/Multicall2.sol
```diff
@@ -0,0 +1,17 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity >=0.5.0;
```
daywiss, line 2: I completely forgot about this file

packages/sdk/src/clients/multicall/client.ts
```diff
@@ -1,8 +1,8 @@
 import { EthersContracts } from "@uma/core";
 import type { SignerOrProvider } from "../..";
 
-export type Instance = EthersContracts.Multicall;
-const Factory = EthersContracts.Multicall__factory;
+export type Instance = EthersContracts.Multicall2;
+const Factory = EthersContracts.Multicall2__factory;
```
daywiss, line 5: i actually think your initial feeling here on creating a separate client is right. i think we should keep multicall 1 and 2 distinct

packages/sdk/src/clients/multicall/client.e2e.ts
```diff
@@ -5,19 +5,24 @@ import { ethers } from "ethers";
 import { emp } from "..";
 
 // multicall contract deployed to mainnet
-const address = "0xeefba1e63905ef1d7acba5a8513c70307c1ce441";
+const multicallV1Address = "0xeefba1e63905ef1d7acba5a8513c70307c1ce441";
+const multicallV2Address = "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
```
daywiss, line 9: again, i think you were right to want to create a separate client, when we talked about it i misunderstood which client you were referring to.  part of the reason we should separate it is because we need to have slightly different tests for v2, and it i think it makes more sense to keep that distinct ffrom the v1 tests.

packages/sdk/src/clients/multicall/client.e2e.ts
```diff
@@ -26,11 +31,27 @@ describe("multicall", function () {
...
+        callData: empClient.interface.encodeFunctionData(call),
+      };
+    });
+    const response = await clientV2.callStatic.tryBlockAndAggregate(false, multicalls);
+    const decoded = calls.map((call: any, i: number) => {
+      const result = response.returnData[i];
+      return empClient.interface.decodeFunctionResult(call, result[1]);
+    });
+    assert.equal(decoded.length, calls.length);
+  });
```
daywiss, line 56: It would also be good to see a test in multicall2 where one of the calls in the list actually fails while others succeed

daywiss (commented):
sorry for the misunderstanding this morning, but yea i think it makes a lot of sense to separate m1 and m2 into separate folders. In my mind i had tasked this thinking you should also update the helper multicall class in lib which is what i was thinking of, but we can do this as a separate task. 

anyways for the most part this looks good, but refactor it into separate clients

packages/sdk/src/clients/multicall/client.ts
amateima: and then `class Multicall`'s State would look like
```
interface State {
  requests: Request[];
  multicallClient: multicall.Instance | multicall2.Instance;
}
```
also `class Multicall` would have an additional method 
`public async smth_like_read_with_errors(_requests: Request[] = this.requests)` that will throw an exception if multicall client is not version 2

what do you think?

daywiss: Yea I think that we can also create a `class Multicall2` as a separate file as well which extends from `class Multicall` and just adds the new function.  I can write up a new card for this and we can also discuss more tomorrow.

packages/sdk/src/clients/multicall2/client.e2e.ts
```diff
@@ -0,0 +1,70 @@
...
+  test("multicall2 on emp with errors", async function () {
+    const calls = ["priceIdentifier", "tokenCurrency", "disputeBondPercentage"];
+    const multicalls = calls.map((call) => {
+      return {
+        target: empAddress,
+        callData: empClient.interface.encodeFunctionData(call as any),
+      };
+    });
+    const response = await client.callStatic.tryBlockAndAggregate(false, multicalls);
+    const decoded = calls.map((call: any, i: number) => {
```
daywiss, line 64: you might want to filter out errors and make sure theres at least one, rather than checking that all the calls returned, or accumulate counters for success/failure and compare against expected

daywiss (commented):
This looks great, just one comment on the new test

packages/sdk/src/clients/multicall2/client.e2e.ts
amateima: good point

packages/core/contracts/common/interfaces/Multicall2.sol
```diff
@@ -0,0 +1,24 @@
...
```
nicholaspai, line 2: Shouldn't we be using solidity 8 by default unless there's a reason not to?

packages/sdk/src/clients/multicall2/index.ts
```diff
@@ -0,0 +1 @@
+export * from "./client";
```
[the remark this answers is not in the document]
amateima: yep. forgot about it

packages/core/contracts/common/interfaces/Multicall2.sol
```diff
@@ -0,0 +1,24 @@
...
+
+contract Multicall2 {
```
nicholaspai, line 4: Is this code inspired from another external contract? If so we should note it in the comments. For example, the `Multicall.sol` contract notes that its inspired by the MakerDAO contract.

nicholaspai (commented):
Great first PR, I did one pass and mainly left comments to add more comments to the smart contracts. On second pass after you re-request my review,  I'll examine the `sdk` code more closely

packages/sdk/src/clients/multicall2/client.ts
```diff
@@ -0,0 +1,9 @@
+import { EthersContracts } from "@uma/core";
+import type { SignerOrProvider } from "../..";
+
+export type Instance = EthersContracts.Multicall2;
+const Factory = EthersContracts.Multicall2__factory;
+
+export function connect(address: string, provider: SignerOrProvider): Instance {
```
nicholaspai, line 7: @daywiss this file should have more comments right?

PR #3326, opened 2021-08-24
improve(insured-bridge): Settle resolved optimistic price requests that passed dispute period

nicholaspai:
**Motivation**

This improves slow relayer UX because they no longer need to separately call `OptimisticOracle.settle` in addition to `BridgePool.settleRelay` to recoup their price request bond.

Tests were also added in this PR to improve readability.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -259,8 +259,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     }
 
     /**
-     * @notice Instantly relay a deposit amount minus fees. Instant relayer earns a reward following the pending relay
-     * challenge period.
+     * @notice Instantly relay a deposit amount minus fees to the recipient. Instant relayer earns a reward following
+     * the pending relay challenge period.
+     * @dev We assume that the caller has performed an off-chain check that the relayed deposit data is valid that they
+     * are attempting to speed up. If the relayed deposit data is invalid, then the instant relayer has no recourse
+     * to receive their funds back after the invalid relay is disputed.
      * @dev Caller must have approved this contract to spend the deposit amount of L1 tokens to relay. There can only
      * be one instant relayer per relay attempt and disputed relays cannot be sped up.
      * @param _depositData Unique set of L2 deposit data that caller is trying to instantly relay.
@@ -295,8 +298,8 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         RelayData storage relay = relays[depositHash];
 
         require(relays[depositHash].relayState == RelayState.Pending, "State not pending");
-        // Note `hasPrice` will return false if liveness has not been passed in the optimistic oracle.
 
+        // Note `hasPrice` will return false if liveness has not been passed in the optimistic oracle.
         require(
             _getOptimisticOracle().hasPrice(
                 address(this),
@@ -307,11 +310,20 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
             "Price not yet resolved"
         );
 
+        // Optimistically settle OptimisticOracle price as a convenience for the slow relayer who will receive their
+        // dispute bond back.
+        _getOptimisticOracle().settle(
+            address(this),
+            bridgeAdmin.identifier(),
+            relay.priceRequestTime,
+            getRelayAncillaryData(_depositData, relay)
+        );
+
         // Note: Why don't we have to check the value of the price?
         // - If the OptimisticOracle has a price and the relayState is PENDING, then we can safely assume that the relay
         // was validated. This is because this contract proposes a price of 1e18, or "YES" to the identifier posing the
         // question "Is this relay valid for the associated deposit?". If the proposal is disputed, then the relayState
-        // will be reset to UNINITIALIZED. If the proposal is not disputed, and there is a price available, then the
+        // will be reset to DISPUTED. If the proposal is not disputed, and there is a price available, then the
         // proposal must have passed the dispute period, assuming the proposal passed optimistic oracle liveness.
 
         // Update the relay state to Finalized. This prevents any re-settling of a relay.
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -293,9 +293,13 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     /**
      * @notice Instantly relay a deposit amount minus fees to the recipient. Instant relayer earns a reward following
      * the pending relay challenge period.
-     * @dev We assume that the caller has performed an off-chain check that the relayed deposit data is valid that they
-     * are attempting to speed up. If the relayed deposit data is invalid, then the instant relayer has no recourse
-     * to receive their funds back after the invalid relay is disputed.
+     * @dev We assume that the caller has performed an off-chain check that the deposit data they are attempting to
+     * relay is valid. If the deposit data is invalid, then the instant relayer has no recourse
+     * to receive their funds back after the invalid deposit data is disputed. Moreover, no one will be able to
+     * resubmit a relay for the invalid deposit data because they know it will get disputed again. On the other hand,
+     * if the deposit data is valid, then even if it is falsely disputed, the instant relayer will eventually get
+     * reimbursed because someone else will be incentivized to resubmit the relay to earn slow relayer rewards. Once the
+     * valid relay is finalized, the instant relayer will be reimbursed.
      * @dev Caller must have approved this contract to spend the deposit amount of L1 tokens to relay. There can only
      * be one instant relayer per relay attempt and disputed relays cannot be sped up.
      * @param _depositData Unique set of L2 deposit data that caller is trying to instantly relay.
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -329,20 +329,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         bytes32 depositHash = _getDepositHash(_depositData);
         RelayData storage relay = relays[depositHash];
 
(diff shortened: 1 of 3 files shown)


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -291,8 +291,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     }
 
     /**
-     * @notice Instantly relay a deposit amount minus fees. Instant relayer earns a reward following the pending relay
-     * challenge period.
+     * @notice Instantly relay a deposit amount minus fees to the recipient. Instant relayer earns a reward following
+     * the pending relay challenge period.
+     * @dev We assume that the caller has performed an off-chain check that the relayed deposit data is valid that they
```
nicholaspai, line 296: @chrismaree wdyt about this way to describe the instant relayers responsibilities?

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -931,16 +931,8 @@ describe("BridgePool", () => {
       );
       assert(await didContractThrow(bridgePool.methods.settleRelay(depositData).send({ from: relayer })));
 
-      // Expire and settle proposal on the OptimisticOracle.
+      // Set time such that optimistic price request is settleable.
       await timer.methods.setCurrentTime(expectedExpirationTimestamp).send({ from: owner });
-      await optimisticOracle.methods
```
nicholaspai, line 936: Before we were manually calling `settle` to pay out the slow relayer, this PR removes that need.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -291,8 +291,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     }
 
     /**
-     * @notice Instantly relay a deposit amount minus fees. Instant relayer earns a reward following the pending relay
-     * challenge period.
+     * @notice Instantly relay a deposit amount minus fees to the recipient. Instant relayer earns a reward following
+     * the pending relay challenge period.
+     * @dev We assume that the caller has performed an off-chain check that the deposit data they are attempting to
```
nicholaspai, line 296: @chrismaree wdyt about this way to describe the instant relayers responsibilities?

chrismaree: 10/10 explanation @nicholaspai. comment looks great

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -1059,8 +1042,8 @@ describe("BridgePool", () => {
         toBN(await l1Token.methods.balanceOf(relayer).call())
           .sub(toBN(relayerBalanceBefore))
           .toString(),
-        realizedSlowRelayFeeAmount.toString(),
-        "Slow relayer should receive proposal slow relay reward"
+        toBN(totalRelayBond).add(realizedSlowRelayFeeAmount).toString(),
```
nicholaspai, line 1045: This was a false positive test that passed before because it equates `relayerBalanceBefore - relayerBalanceNow` and `realizedSlowRelayFeeAmount`, and `relayerBalanceBefore` was unintentionally being called AFTER  the `optimisticOracle.settle()` method, so the relayer had already received back their bond by that time.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -326,23 +333,24 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-                    address(this),
-                    bridgeAdmin.identifier(),
-                    relay.priceRequestTime,
-                    getRelayAncillaryData(_depositData, relay)
-                ),
-            "Relay not settleable"
+        // If relay was disputed, then parties in dispute need to go through OptimisticOracle to resolve dispute.
+        require(relays[depositHash].relayState == RelayState.Pending, "Relay not settleable");
+
+        // Attempt to settle OptimisticOracle price as a convenience for the slow relayer who will receive their
```
chrismaree, line 339: nice. this is easier to reason about now

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -326,23 +333,24 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+        // dispute bond back. This doubles as a check that the optimistic price request has resolved because this will
+        // revert if the price has not resolved yet.
+        _getOptimisticOracle().settle(
+            address(this),
+            bridgeAdmin.identifier(),
+            relay.priceRequestTime,
+            getRelayAncillaryData(_depositData, relay)
         );
 
         // Note: Why don't we have to check the value of the price?
```
chrismaree, line 349: is this comment still valid now? I guess we are not _checking_ for a price but we are _settling. We might want to re-word this to make it clear that we are not checking the outcome of the settlement because in order for this OO request to be sellable the proposal must have _not_ been disputed. If it was disputed.

nicholaspai: great call, done.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -326,23 +333,26 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-                    getRelayAncillaryData(_depositData, relay)
-                ),
-            "Relay not settleable"
+        // If relay was disputed, then parties in dispute need to go through OptimisticOracle to resolve dispute.
+        require(relays[depositHash].relayState == RelayState.Pending, "Relay not settleable");
+
+        // Attempt to settle OptimisticOracle price as a convenience for the slow relayer who will receive their
+        // dispute bond back. This doubles as a check that the optimistic price request has resolved because this will
+        // revert if the price has not resolved yet.
+        _getOptimisticOracle().settle(
```
mrice32, line 342: Sorry, this may be a dumb question, but where do we validate that the final price was 1e18 (i.e. the relay was valid)?

mrice32 (commented):
Looks great! Awesome comments. Only one minor question around the settlement logic.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -326,23 +333,26 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         bytes32 depositHash = _getDepositHash(_depositData);
         RelayData storage relay = relays[depositHash];
 
-        // Note `hasPrice` will return false if liveness has not been passed in the optimistic oracle.
-        require(
-            relays[depositHash].relayState == RelayState.Pending &&
-                _getOptimisticOracle().hasPrice(
```
chrismaree, line 332: I'm not convinced of this technique after thinking more about it. What happens if I call `settle` directly on the OO first (as soon as the request is passed liveness) before the relayer calls `settleRelay`? In this case, the price request will be settled _before_ entering this method making the subsequent call to the OO on `settle` revert. This will make this method revert.

As a result, the path forward is: a) calling `settle` if the request is settleable or using `hasPrice` b) leaving it as it was before wherein we ignore the `settle` logic thereby enforcing that it is settled outside of this contract.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: despite this making _sence_ I think it might be worthwhile to double-check the value returned, even if it is just for making the code easier to reason about without adding any extra security/functionality.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: This is a good point. We can wrap try catch around settle and then re add the hasPrice requirement?

nicholaspai: I ended up using the `settleAndGetPrice` which does not revert if the price is already settled and returns the resolved price (after optionally settling).

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I removed `settle` and replaced it with a check to `settleAndGetPrice` which will settle the price if it is not resolved, revert if the price is not settleable, and return the resolved price.

chrismaree: beautiful. looks good!

packages/core/test/insured-bridge/BridgePool.js
chrismaree: great catch. we should spend some time working on these tests and making sure they are doing what we intend. there are a number of other simplifications and syntactical changes I'm keen to make once we've got the other stuff 100% locked down.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
chrismaree: excellent. I think this is the ideal setup. Also really happy you are checking the value price to remove any complexity in _implied_ correctness.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -326,25 +333,21 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
-                    relay.priceRequestTime,
-                    getRelayAncillaryData(_depositData, relay)
-                ),
-            "Relay not settleable"
+            _getOptimisticOracle().settleAndGetPrice(
+                bridgeAdmin.identifier(),
+                relay.priceRequestTime,
+                getRelayAncillaryData(_depositData, relay)
+            ) == int256(1e18), // Canonical value representing "True"; i.e. the proposed relay is valid.
+            "Proposed relay request did not resolve to Valid"
```
chrismaree, line 348: nit: try get this to less than 32 chars. 
```suggestion
            "Relay request was not valid"
```

chrismaree (approved):
Looks good! had one final nit on a require message. Otherwise, good job on implementing, improvement in the ux is clear!

PR #3336, opened 2021-08-30
improve(sdk): add multicall2 client wrapper 

amateima:
**Motivation**
https://app.clubhouse.io/uma-project/story/2158/add-multicall-2-client-wrapper-to-sdk

**Summary**

This PR contains a wrapper implementation for the Multicall2 client which offers an easy to use interface for batching and reading multiple smart contract read calls. Besides the returned data, `readWithErrors` also returns a `success` key

**Details**

It was crucial to extend, but keep unchanged the public interface of the old Multicall wrapper. To achieve this goal, I proceed with composition instead of inheritance mechanism. I encountered some issues when I tried to call a parent class method that must return a new instance of the caller's class. It could be done with syntax like `const newObject = new (<any>this.constructor)(this)`, but composition is easier to read and understand in this case

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**


packages/sdk/src/multicall.ts
```diff
@@ -18,29 +18,29 @@ export type EncodedRequest = {
   callData: string;
 };
 
-interface State {
+export interface State {
   requests: Request[];
-  multicallClient: multicall.Instance;
+  multicallClient: multicall.Instance | multicall2.Instance;
```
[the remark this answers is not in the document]
daywiss: I think this might work, make state have a generic:
```ts
interface State <MulticallType>{
   multicallClient: MulticallType
}
```

Then in the class

```ts
class Multicall <MulticallType> implements State {
  public requests: Request[];
  public multicallClient: MulticallType;
```

Then in factory:
```ts
export default class Factory extends Multicall<multicall.Instance> {
  constructor(address: string, provider: SignerOrProvider) {
    const multicallClient = multicall.connect(address, provider);
    super({ multicallClient, requests: [] });
  }
}
```

I dont know if this actually compiles, but it might work, wdyt?

daywiss (commented):
this looks really good overall, lets just brainstorm around m1 needing to know m2

packages/sdk/src/multicall.ts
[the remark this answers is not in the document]
amateima: I agree that multicall2.Instance should be pulled out of Multicall. I'll work smth out base on your generics suggestion

packages/sdk/src/multicall.ts
```diff
@@ -18,29 +18,29 @@ export type EncodedRequest = {
...
+export interface State<MulticallType> {
   requests: Request[];
-  multicallClient: multicall.Instance;
+  multicallClient: MulticallType;
 }
 
 // Multicall class that exposes public functions to the user and recursively chains itself.  Acts immutable
 // if you store reference to the parent intsance. Children will contain mutated state.
-class Multicall implements State {
+export class Multicall<MulticallType extends multicall.Instance> implements State<multicall.Instance> {
```
mrice32, line 28: Why does this now need to be exported?

Why does it need to be a generic?

Could it just be
```ts
export class Multicall implements State<multicall.Instance> {
  public requests: Request[];
  public multicallClient: multicall.Instance;
  constructor(state: State<multicall.Instance>) {
```

Now, I _could_ see this making sense if you wanted to move most of this logic to a parent class and only implement the read call in the derived class. But that complexity probably isn't worth it.

mrice32 (commented):
Looks good! One question on some of the changes to the original multicall class.

packages/sdk/src/multicall2.ts
```diff
@@ -0,0 +1,53 @@
...
+  // reads from the contract, returns the read and error results in order that requests were queued.
+  public async readWithErrors(_requests: Request[] = this.requests) {
+    const encodedRequests = _requests.map((request) => this.encodeRequest(request));
+    const [, , returnData] = await this.multicallClient.callStatic.tryBlockAndAggregate(false, encodedRequests);
+    const zipped = zip(_requests, returnData);
+    return zipped.map(([request, response]) => {
+      if (request && response) {
+        return {
+          success: response.success,
+          result: response.success ? this.decodeResponse(request, response.returnData) : undefined,
```
daywiss, line 21: do you think we should copy the promise.allSettled return values?

```js
// [
//   {status: "fulfilled", value: 33},
//   {status: "fulfilled", value: 66},
//   {status: "fulfilled", value: 99},
//   {status: "rejected",  reason: Error: an error}
// ]
```

It doesnt change much except the actual props, just to standardize it a little.

packages/sdk/src/multicall2.ts
```diff
@@ -0,0 +1,53 @@
...
+        };
+      }
+      throw new Error("Unable to decode contract response");
+    });
+  }
+
+  // adds a new request to the queue, to be executed when read is called. Returns an instance of this class so you can chain.
+  public add(contractInstance: Contract, call: Call) {
+    const multicall = super.add(contractInstance, call);
+    return new Multicall2({
```
daywiss, line 31: does passing the multicall object in here not work (rather than restructuring it)? might be slightly cleaner if thats possible

packages/sdk/src/multicall.ts
mrice32: Okay I think this makes sense! Sorry for misunderstanding the code originally!

daywiss (commented):
Glad you could get this working, looks really good, just a couple nits

packages/sdk/src/multicall2.ts
```diff
@@ -0,0 +1,53 @@
...
+      multicallClient: multicall.multicallClient,
+      requests: multicall.requests,
+    });
```
mrice32, line 34: nit: any reason we can't just do this as a minor simplification?
```ts
return new Multicall2(super.add(contractInstance, call));
```

mrice32 (approved):
LGTM!

packages/sdk/src/multicall.ts
amateima: because not only the state depends on the multicall Instance type, but the multicallClient attribute too :(

packages/sdk/src/multicall2.ts
amateima: It was not working before adding generics. I've updated it 👍

packages/sdk/src/multicall2.ts
amateima: I've updated it. The first implementation didn't have generics and this construction didn't work

packages/sdk/src/multicall2.ts
amateima: I like more boolean flags instead of string values. And multicall result doesn't include the error in case one of the calls fails

daywiss (approved):

PR #3342, opened 2021-08-31
improve(financial-templates-lib): migrate to hardhat

mrice32:
**Motivation**

Similar to other related PRs, this PR moves financial templates lib off of any dependence on truffle.

**Summary**

Reworks tests to no longer require truffle to run and allows them to be run with mocha directly.

This also eliminates the separate truffle test job that was required.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


mrice32:
@evaldofelipe I'm not quite sure what's going wrong here. Could you take a look?

ci/tests/test-financial-templates-lib.sh
```diff
@@ -8,7 +8,7 @@ cat << EOF
         command: ganache-cli -i 1234 -l 9000000 -p 9545
     working_directory: ~/protocol
     resource_class: medium+
-    parallelism: 35
+    parallelism: 10
```
chrismaree, line 11: why change this?

packages/common/src/MultiVersionTestHelpers.ts
```diff
@@ -18,6 +18,21 @@ export const TESTED_CONTRACT_VERSIONS = [
...
 ];
 
+export const CORE_CONTRACTS_NODE_VERSION_MAPPING = {
+  "2.0.1": "0.1.0",
+};
+
+// Assumes that the contracts-node package alias is @uma/contracts-node-X-Y-Z.
+// Typical workflow is:
+// const { getAbi, getBytecode } = require(getContractsNodePackageAliasForVersion(versionString));
+export function getContractsNodePackageAliasForVerion(version: string): string {
```
chrismaree, line 28: so the `version` in this function is the version of `core` which is mapped to the `contracts-node`

packages/common/src/SolidityTestUtils.ts
```diff
@@ -158,3 +160,42 @@ export async function mineTransactionsAtTime(
     await startMining(web3);
   }
 }
+
+// Very similar to the function above, but uses a hardhat network to send hardhat node compatible RPC methods.
+export async function mineTransactionsAtTimeHardhat(
```
chrismaree, line 165: nice

chrismaree: it might be nice if there was just one method and it switched the kind of call sent depending on the network so that we don't have to have this extra function name.

chrismaree: not a big issue for now though.

packages/financial-templates-lib/test/clients/FinancialContractClient.js
```diff
@@ -71,28 +76,40 @@ const versionedIt = function (supportedVersions, shouldBeItOnly = false) {
...
+
+  let sponsor1, sponsor2;
 
   TESTED_CONTRACT_VERSIONS.forEach(function (contractVersion) {
     // Store the contractVersion.contractVersion, type and version being tested
     iterationTestVersion = contractVersion;
 
+    const { getAbi, getBytecode } = require(getContractsNodePackageAliasForVerion(contractVersion.contractVersion));
+
+    const createContract = (name) => {
```
chrismaree, line 93: would it be useful to refactor this to a common helper? it looks like you do it in two places but I know for sure we do it in other places similar to this.

You'd likely need to pass in the `getContract` method, though, which is not ideal.

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
```diff
@@ -329,22 +360,22 @@ contract("FinancialContractEventClient.js", function (accounts) {
               objectsInArrayInclude(
                 [
                   {
-                    transactionHash: newSponsorTxObj1.tx,
-                    blockNumber: newSponsorTxObj1.receipt.blockNumber,
```
chrismaree, line 333: pretty irritating that they change this syntax. with that said, the hardhat syntax is cleaner.

packages/financial-templates-lib/test/clients/OptimisticOracleEventClient.js
```diff
@@ -60,49 +59,66 @@ contract("OptimisticOracleEventClient.js", function (accounts) {
...
-    store = await Store.new({ rawValue: "0" }, { rawValue: "0" }, timer.address);
-    await finder.changeImplementationAddress(utf8ToHex(interfaceName.Store), store.address);
+    finder = await Finder.new().send({ from: accounts[0] });
+    timer = await Timer.new().send({ from: accounts[0] });
 
-    mockOracle = await MockOracle.new(finder.address, timer.address);
-    await finder.changeImplementationAddress(utf8ToHex(interfaceName.Oracle), mockOracle.address);
+    // Whitelist an initial identifier we can use to make default price requests.
+    identifierWhitelist = await IdentifierWhitelist.new().send({ from: accounts[0] });
+    await identifierWhitelist.methods.addSupportedIdentifier(identifier).send({ from: accounts[0] });
```
chrismaree, line 77: as a general comment, it's much cleaner IMHO to use the `owner` over the `account[0]`. I know you did not add this in this PR. We should consider going through and removing this syntax through the protocol tests when there is a variable declared that holds the same address.

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
```diff
@@ -331,20 +335,20 @@ contract("CreatePriceFeed.js", function (accounts) {
       withdrawalLiveness: "1000",
       collateralAddress: collateralTokenAddress,
       tokenAddress: syntheticTokenAddress,
-      finderAddress: finder.address,
+      finderAddress: finder.options.address,
       priceFeedIdentifier: padRight(utf8ToHex("ETH/BTC"), 64), // Note: an identifier which is part of the default config is required for this test.
       liquidationLiveness: "1000",
       collateralRequirement: { rawValue: toWei("1.5") },
-      disputeBondPct: { rawValue: toWei("0.1") },
```
chrismaree, line 338: why was this change required? does this imply you are now using the new version of these contracts from core/contracts-node? I might have missed that change in the `package.json`

packages/financial-templates-lib/test/price-feed/UniswapV3PriceFeed.js
```diff
@@ -1,22 +1,22 @@
...
-
-const UniswapMock = getTruffleContract("UniswapV3Mock", web3, CONTRACT_VERSION);
-const Uniswap = getTruffleContract("UniswapV3", web3, CONTRACT_VERSION);
-const Token = getTruffleContract("ExpandedERC20", web3, CONTRACT_VERSION);
+const UniswapMock = getContract("UniswapV3Mock");
+const Uniswap = getContract("UniswapV3");
+const Token = getContract("ExpandedERC20");
 
 const Convert = (decimals) => (number) => (number ? parseFixed(number.toString(), decimals).toString() : number);
+const getFutureTime = async () => Number((await web3.eth.getBlock("latest")).timestamp) + 10000;
```
chrismaree, line 16: niiice. sneaky

packages/financial-templates-lib/test/clients/FinancialContractClient.js
```diff
@@ -71,28 +76,40 @@ const versionedIt = function (supportedVersions, shouldBeItOnly = false) {
...
-contract("FinancialContractClient.js", function (accounts) {
-  const sponsor1 = accounts[0];
-  const sponsor2 = accounts[1];
+describe("FinancialContractClient.js", function () {
+  before(async function () {
+    accounts = await web3.eth.getAccounts();
+    [sponsor1, sponsor2] = accounts;
+  });
+
+  let sponsor1, sponsor2;
```
daywiss, line 85: just a nit but probably would be clearer to define this above, like you do in the next file

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
```diff
@@ -1,24 +1,34 @@
...
+const { delay } = require("../../dist/helpers/delay.js");
 
-const BalancerMock = getTruffleContract("BalancerMock", web3);
-const Balancer = getTruffleContract("Balancer", web3);
+const BalancerMock = getContract("BalancerMock");
+const Balancer = getContract("Balancer");
 
-contract("BalancerPriceFeed.js", function (accounts) {
-  const owner = accounts[0];
+const getFutureTimestamp = async () => {
```
daywiss, line 13: might want to comment on this function and why its used

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
```diff
@@ -48,28 +58,28 @@ contract("BalancerPriceFeed.js", function (accounts) {
...
     // Add an invalid price as the most recent price, which should be ignored.
-    await dexMock.setPrice(toWei("0"));
+    await dexMock.methods.setPrice(toWei("0")).send({ from: owner });
     await dexPriceFeed.update();
 
     assert.equal(dexPriceFeed.getSpotPrice().toString(), toWei("0.25"));
   });
 
   it("Selects most recent price in same block", async function () {
     // Just use current system time because the time doesn't matter.
```
daywiss, line 72: this comment should probably change

daywiss (approved):
great work as usual

ci/tests/test-financial-templates-lib.sh
mrice32: The parallelism was so high that we had _empty_ jobs that were failing because they were empty. I talked to @evaldofelipe  and he recommended that we reduce the parallelism to solve it. I also realized that each job was taking like < 2 mins (and most were like < 30s) with a pretty high variance. So I decided to reduce parallelism so ci would bunch some together without making the overall tests take longer (since startup time is the biggest cost for many of these). The results were pretty good so I kept it. We could probably reduce to even like 5 or so and we'd probably _still_ end up with the same long-pole time of like 2 mins, but 10 is probably good for now. Feel free to take a look at the job!

packages/common/src/MultiVersionTestHelpers.ts
mrice32: Exactly. It's to help bridge the gap between modules that are already using core versions.

packages/common/src/SolidityTestUtils.ts
mrice32: Is there a concrete way to tell what node you're using?

packages/financial-templates-lib/test/clients/FinancialContractEventClient.js
mrice32: Yeah, agreed. Was pretty annoying to go and fix all of these.

packages/financial-templates-lib/test/clients/OptimisticOracleEventClient.js
mrice32: Agreed. I started doing it later since I was having to define the named accounts anyway (so there was no risk of owner not existing in that file). 

We can do a big find/replace down the line, but honestly, I just want to get the tests working, and I'm not _super_ concerned as long as it's easy to understand what's going on.

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
mrice32: Yeah, we're using the new version of the EMP. I _think_ this is better because the new EMP is all we're supporting these days, so it seems better to test against it as opposed to the older, hardcoded versions that don't exist in the wild anymore.

packages/financial-templates-lib/test/price-feed/UniswapV3PriceFeed.js
mrice32: Yeah, these tests all do some very weird things. They were stuck running on oldschool truffle (no hardhat at all) because they didn't play nicely with the hardhat node. I realized that this was because 1) some of our RPC calls to move time around were not allowed/ignored by hardhat (which is why I made the change in common) and 2) we were actually mining blocks with timestamps that were before blocks that came before them. Ganache allows this, but hardhat does not. So this method ensures that each test starts pretty far past the previous one so it has space to mine pretty far before this timestamp without breaking.

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
mrice32: Great point. Will do.

packages/financial-templates-lib/test/clients/FinancialContractClient.js
mrice32: Good point. Will do.

packages/common/src/SolidityTestUtils.ts
```diff
@@ -158,3 +160,42 @@ export async function mineTransactionsAtTime(
...
+  await network.provider.send("evm_setAutomine", [false]);
+
+  try {
+    const receiptPromises = [];
+    for (const transaction of transactions) {
+      const result = transaction.send({ from: sender });
+
+      // Awaits the transactionHash, which signifies the transaction was sent, but not necessarily mined.
+      await new Promise<void>((resolve, reject) => {
+        result.on("transactionHash", function () {
```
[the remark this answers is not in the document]
mrice32: I think this is just copied from the other method. Will change it to once in both places, though, as you make a good point!

packages/common/src/SolidityTestUtils.ts
```diff
@@ -158,3 +160,42 @@ export async function mineTransactionsAtTime(
...
+
+      // result, itself, is a promise that will resolve to the receipt.
+      receiptPromises.push(result);
+    }
+
+    await network.provider.send("evm_mine", [time]);
+    const receipts = await Promise.all(receiptPromises);
+    return receipts;
+  } catch (err) {
+    throw new Error(err.message);
```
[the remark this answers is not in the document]
mrice32: This was just copied from above. I think this was done this way to throw the error but _also_ to have a finally clause where we could restart the evm miner even if an exception is thrown. Does that make sense?

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
mrice32: Agreed. will update

packages/financial-templates-lib/test/clients/FinancialContractClient.js
mrice32: Yeah, it's also tricky because the require needs to be called from this file since requires resolve packages based on the file where they're declared. So if the code were in common, it would try to find that package under uma/common. That might happen to work in our monorepo, but if there were name conflicts or the package weren't hoisted, it could cause issues (I think).

packages/common/src/MultiVersionTestHelpers.ts
```diff
@@ -18,6 +18,21 @@ export const TESTED_CONTRACT_VERSIONS = [
...
+  if (version === "latest") return "@uma/contracts-node";
+  const isKey = (input: string): input is keyof typeof CORE_CONTRACTS_NODE_VERSION_MAPPING =>
+    input in CORE_CONTRACTS_NODE_VERSION_MAPPING;
+  if (!isKey(version)) throw new Error("Unkown version!");
```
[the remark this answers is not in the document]
mrice32: Fixed!

packages/common/src/SolidityTestUtils.ts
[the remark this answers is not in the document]
mrice32: Fixed!

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
mrice32: Done!

packages/financial-templates-lib/test/price-feed/BalancerPriceFeed.js
mrice32: Fixed all occurrences!

mrice32:
@daywiss @chrismaree responded to and/or fixed all comments. PTAL and let me know if there's anything you'd like me to change, update, or further explain. Thanks for the thorough reviews!

packages/financial-templates-lib/test/clients/FinancialContractClient.js
chrismaree: ye that makes sense 100p and is a good enough reason to leave it as it is now.

packages/financial-templates-lib/test/price-feed/CreatePriceFeed.js
chrismaree: agreed! all the old ones are deprecated anyways

packages/financial-templates-lib/test/price-feed/UniswapV3PriceFeed.js
chrismaree: very nice. thanks for the context!

chrismaree:
> @daywiss @chrismaree responded to and/or fixed all comments. PTAL and let me know if there's anything you'd like me to change, update, or further explain. Thanks for the thorough reviews!

looks good to me! thanks for implementing all the requested changes!

packages/common/src/SolidityTestUtils.ts
```diff
@@ -136,10 +138,10 @@ export async function mineTransactionsAtTime(
 
       // Awaits the transactionHash, which signifies the transaction was sent, but not necessarily mined.
       await new Promise<void>((resolve, reject) => {
-        result.on("transactionHash", function () {
+        result.once("transactionHash", function () {
           resolve();
```
daywiss, line 142: just one note here, this is resolving with no hash, hopefully callers dont expect the hash value from the receipts, it will always be an empty array

packages/common/src/SolidityTestUtils.ts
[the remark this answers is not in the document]
daywiss: oh ok i meant more this: `throw err` vs `throw new Error(err.message)`. im not 100% sure, but you could probably omit the catch block entirely and just have a try/finally.  But if its just porting the code then yea probably fine as is

daywiss (approved):

packages/common/src/SolidityTestUtils.ts
mrice32: @daywiss callers _do not_ expect this. Note the await in front of the promise -- this promise's result is never being used, so we just want to wait until the txn hash has been generated, we don't actually need to use it. But good call out!

packages/common/src/SolidityTestUtils.ts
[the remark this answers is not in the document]
mrice32: You're totally right. We can do try finally and it will do what we want! And yeah, we shouldn't be dropping the trace.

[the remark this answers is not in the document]
mrice32: Fixed!

PR #3344, opened 2021-09-01
improve(serverless-orchestration): migrate to hardhat

mrice32:
**Motivation**

All packages should be fully migrated to running on hardhat without the truffle plugin.

**Summary**

These packages still require an external node to be run (either hardhat node or ganache), but they now don't require truffle.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/serverless-orchestration/test/ServerlessHub.js
+++ b/packages/serverless-orchestration/test/ServerlessHub.js
@@ -1,4 +1,8 @@
-const { toWei, utf8ToHex, padRight } = web3.utils;
+const hre = require("hardhat");
+const { getContract, web3, network } = hre;
+const { assert } = require("chai");
+const Web3 = require("web3");
+const { toWei, utf8ToHex, padRight } = Web3.utils;
 
 // Enables testing http requests to an express server.
 const request = require("supertest");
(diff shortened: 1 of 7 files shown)


packages/serverless-orchestration/test/ServerlessHub.js
```diff
@@ -69,15 +72,24 @@ contract("ServerlessHub.js", function (accounts) {
...
+    const node = ganache.server({ _chainIdRpc: chainId });
+    node.listen(port);
+    return new Web3("http://127.0.0.1:" + port);
+  };
+
   before(async function () {
+    accounts = await web3.eth.getAccounts();
+    [contractDeployer] = accounts;
     defaultChainId = await web3.eth.getChainId();
+    await runDefaultFixture(hre);
```
chrismaree, line 85: ideally, in the future, we should move away from these kinds of fixtures as they ended up slowing things down more than need be by deploying more than required a lot of the time.

chrismaree (approved):
This is great! awesome work on converting this package. I'm sure it's _much much_ faster now. I suspected there to be more pain with converting things with the network URLs but looks like hardhat is now pretty similar with the `web3.currentProvider.host to `network.config.url`.

daywiss (approved):
lgtm

packages/serverless-orchestration/test/ServerlessHub.js
mrice32: Yep! There are a bunch of simplicity-vs-speed tradeoffs here. I basically just use this when any tests (intentionally or not) depend on pre-deployed contracts. I don't include it by default, though, so most of our tests actually don't need it, which means we get most of this speedup already!

PR #3346, opened 2021-09-01
feat(trader): migrate away from truffle

mrice32:
**Motivation**

Migrate trader off of truffle.

**Summary**

Similar to other PRs, this moves off of any direct dependence on truffle. All tests can be run with mocha (which uses pulls in hardhat under the hood).

Note: there were a few minor source code changes required to get this to work.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/common/src/HardhatConfig.ts
+++ b/packages/common/src/HardhatConfig.ts
@@ -1,6 +1,8 @@
 import { HardhatConfig } from "hardhat/types";
 
 const { getNodeUrl, mnemonic } = require("./TruffleConfig");
+import { HRE } from "./hardhat/plugins/ExtendedWeb3";
+export type { HRE };
 
 export function getHardhatConfig(
   configOverrides: any,
--- a/packages/trader/src/RangeTrader.ts
+++ b/packages/trader/src/RangeTrader.ts
@@ -141,7 +141,7 @@ export class RangeTrader {
     }
 
     // Wait exactly one block to fetch events. This ensures that the events have been indexed by your node.
-    await blockUntilBlockMined(this.web3, (await this.web3.eth.getBlockNumber()) + 1);
+    await blockUntilBlockMined(this.web3, tradeExecutionTransaction.blockNumber);
 
     // Get the post trade spot price to double check deviation error.
     await this.tokenPriceFeed.update();
--- a/packages/trader/src/RangeTrader.ts
+++ b/packages/trader/src/RangeTrader.ts
@@ -141,7 +141,7 @@ export class RangeTrader {
     }
 
     // Wait exactly one block to fetch events. This ensures that the events have been indexed by your node.
-    await blockUntilBlockMined(this.web3, tradeExecutionTransaction.blockNumber);
+    await blockUntilBlockMined(this.web3, Number(tradeExecutionTransaction.blockNumber) + 1);
 
     // Get the post trade spot price to double check deviation error.
     await this.tokenPriceFeed.update();
--- a/packages/trader/src/exchange-adapters/UniswapV2Trader.ts
+++ b/packages/trader/src/exchange-adapters/UniswapV2Trader.ts
@@ -4,11 +4,13 @@ import BigNumber from "bignumber.js";
 
 import { MAX_UINT_VAL } from "@uma/common";
 import ExchangeAdapterInterface from "./ExchangeAdapterInterface";
-import { getTruffleContract } from "@uma/core";
+import { getAbi, getBytecode } from "@uma/contracts-node";
+import type { DSProxyManager } from "@uma/financial-templates-lib";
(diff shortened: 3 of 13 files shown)


packages/trader/src/RangeTrader.ts
```diff
@@ -141,7 +141,7 @@ export class RangeTrader {
     }
 
     // Wait exactly one block to fetch events. This ensures that the events have been indexed by your node.
-    await blockUntilBlockMined(this.web3, (await this.web3.eth.getBlockNumber()) + 1);
+    await blockUntilBlockMined(this.web3, tradeExecutionTransaction.blockNumber);
```
daywiss, line 144: am i missing something here, but if the `tradeExecutionTransaction` has a `blockNumber` on it, doesnt that mean its been mined and you dont need to wait?

daywiss (approved):
looks good

packages/trader/src/RangeTrader.ts
mrice32: I want @chrismaree to review this change specifically. I don't know exactly what the original intention of this was. It might've been to add a 1 block buffer after _one_ node said that it was mined because you may get load balanced to another node that may be a little behind. In that case, my solution isn't sufficient to do what @chrismaree was originally intending.

However, to answer your question @daywiss, it does mean that you've passed that block, but it doesn't mean that all infura nodes think you've passed that block. You could get load balanced to a different one. This causes issues in testing, though, because blocks are _only_ mined by hardhat when transactions are submitted (I have no idea how this ever worked in test environments).

@chrismaree may suggest that we find another workaround, which may mean that we briefly turn on scheduled mining rather than automining around these transactions in the test.

chrismaree: yup @mrice32, you are exactly right as to why this bit of code was here. What would happen in production (only some times) is the bot would execute a trade and then it would report on the trade and the logic below would fetch the state on-chain regarding the pool amounts but it would not yet have captured the trade for some or other reason. Load balancing and some kind of delay in Infura makes the most sense. the simplest solution was to just force the bot to wait ~13 seconds at which point the nodes are "up to date" and then the logging works as expected.

what you have implemented here I don't think will achieve this effect correctly and the logs will still be invalid. can you not just simply change it to be:

```js
await blockUntilBlockMined(this.web3,tradeExecutionTransaction.blockNumber + 1);

```

how it worked in tests, like you pointed out is a good question. offhand I can remember how I dealt with this but glad it's working now!

packages/trader/src/exchange-adapters/UniswapV2Trader.ts
```diff
@@ -18,13 +20,13 @@ export class UniswapV2Trader implements ExchangeAdapterInterface {
     readonly uniswapFactoryAddress: string,
     readonly tokenAAddress: string,
     readonly tokenBAddress: string,
-    readonly dsProxyManager: any
+    readonly dsProxyManager: DSProxyManager
```
chrismaree, line 23: yay for types!

chrismaree (approved):
this looks good to me! thanks for making this conversion @mrice32 ! 

Only thing is the comment thread with @daywiss might need to be changed for the logs to be produced correctly.

packages/common/src/HardhatConfig.ts
```diff
@@ -1,6 +1,8 @@
 import { HardhatConfig } from "hardhat/types";
 
 const { getNodeUrl, mnemonic } = require("./TruffleConfig");
+import { HRE } from "./hardhat/plugins/ExtendedWeb3";
```
chrismaree, line 4: could you also comment on this `HRE` vs the previously commonly used `hre`? what is the distinction?

packages/trader/src/RangeTrader.ts
daywiss: i see ok that makes sense. thanks for the context

mrice32: Thanks for the context @chrismaree. I will correct this and get your review on it next week!

packages/trader/src/exchange-adapters/UniswapV2Trader.ts
mrice32: :)

packages/common/src/HardhatConfig.ts
mrice32: HRE is a type that I created in the ExtendedWeb3 plugin to correctly reflect the components we add to the hre. So it's basically the vanilla web3 + some extra fields that we add in our setup, I will add a comment on this!

chrismaree: ok awesome, thanks!

chrismaree:
we ready to merge this in @mrice32 ?

PR #3347, opened 2021-09-01
fix(api): remove block events and update on timer

daywiss:

**Motivation**
https://app.clubhouse.io/uma-project/story/2249/post-expiry-ui-does-not-show-the-user-we-are-waiting-for-a-price-from-the-oo

**Summary**

Theres a bug where state updates stop happening on api contracts. This change removes block update events and updates on an env configurable timer instead. By default if not specified it will update every 10 minutes. 

**Details**

This also updates some logs to make it more obvious when things are happening. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.clubhouse.io/uma-project/story/2249/post-expiry-ui-does-not-show-the-user-we-are-waiting-for-a-price-from-the-oo

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -29,12 +29,12 @@ export default async (env: ProcessEnv) => {
   // we need web3 for syth price feeds
   const web3 = new Web3(env.CUSTOM_NODE_URL);
 
-  // how many blocks to skip before running updates on contract state
-  const updateBlocks = Number(env.UPDATE_BLOCKS || 1);
+  // how often to run expensive state updates, defaults to 10 minutes
+  const updateRateS = Number(env.UPDATE_RATE_S || 10 * 60);
   // default to 10 days worth of blocks
   const oldestBlock = Number(env.OLDEST_BLOCK_MS || 10 * 60 * 60 * 24 * 1000);
 
-  assert(updateBlocks > 0, "updateBlocks must be 1 or higher");
+  assert(updateRateS > 1, "UPDATE_RATE_S must be 1 or higher");
 
   // state shared between services
   const appState: AppState = {
@@ -99,7 +99,6 @@ export default async (env: ProcessEnv) => {
         },
       },
     },
-    lastBlock: 0,
     lastBlockUpdate: 0,
     registeredEmps: new Set<string>(),
     registeredLsps: new Set<string>(),
@@ -156,6 +155,9 @@ export default async (env: ProcessEnv) => {
   await services.lsps.update(appState.lastBlockUpdate, initBlock.number);
   console.log("Updated LSP state");
 
+  // we've update our state based on latest block we queried
+  appState.lastBlockUpdate = initBlock.number;
+
   await services.erc20s.update();
   console.log("Updated tokens");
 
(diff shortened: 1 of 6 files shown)


packages/api/src/apps/api/app.ts
```diff
@@ -238,6 +229,19 @@ export default async (env: ProcessEnv) => {
     await services.globalStats.update();
   }
 
+  // wait update rate before running loops, since all state was just updated on init
+  await new Promise((res) => setTimeout(res, updateRateS * 1000));
```
daywiss, line 233: previously we would do the state updates on init, then go into the loops which would immediately run the updates again. this pauses the loops until some time has passed before re-running updates. This is not part of the bugfix, but an unrelated improvement

packages/api/src/services/actions/emp.ts
```diff
@@ -32,7 +32,7 @@ export function Handlers(config: Config, appState: Dependencies): Actions {
       return Array.from(registeredEmps.values());
     },
     lastBlock() {
-      return appState.lastBlock;
+      return appState.lastBlockUpdate;
```
mrice32, line 35: What's the difference between lastBlock and the pre-existing lastBlockUpdate?

packages/api/src/apps/api/app.ts
```diff
@@ -29,12 +29,12 @@ export default async (env: ProcessEnv) => {
...
 
-  // how many blocks to skip before running updates on contract state
-  const updateBlocks = Number(env.UPDATE_BLOCKS || 1);
+  // how often to run expensive state updates, defaults to 10 minutes
+  const updateRateS = Number(env.UPDATE_RATE_S || 10 * 60);
   // default to 10 days worth of blocks
   const oldestBlock = Number(env.OLDEST_BLOCK_MS || 10 * 60 * 60 * 24 * 1000);
 
-  assert(updateBlocks > 0, "updateBlocks must be 1 or higher");
+  assert(updateRateS > 1, "UPDATE_RATE_S must be 1 or higher");
```
mrice32, line 37: This error message or condition seems to be a typo. This should either be >= 1 or it should say "must be greater than 1".

mrice32 (approved):
Looks good! Only minor comments!

amateima (approved):
Looks good to me. 
In the future, if we need to fetch data more often than `UPDATE_RATE_S`, we can emit events when the update is finished, so the next block can be processed.
Also, pushing all the state updates in a queue and processing them in order is also a solution to keep in mind.

packages/api/src/apps/api/app.ts
daywiss: good call

packages/api/src/services/actions/emp.ts
daywiss: in the previous code, we maintained a var for the last block seen based on block events (lastBlock). There was a separate counter for the block we ran a state update on, ie it would run every 100 blocks (lastBlockUpdate).  Its confusing, but now that we dont use block events, we can remove one. 

Also i mainly have this call here for debugging so i can check now how far behind our last update was through the api.

packages/api/src/apps/api/app.ts
```diff
@@ -29,12 +29,12 @@ export default async (env: ProcessEnv) => {
   // we need web3 for syth price feeds
   const web3 = new Web3(env.CUSTOM_NODE_URL);
 
-  // how many blocks to skip before running updates on contract state
-  const updateBlocks = Number(env.UPDATE_BLOCKS || 1);
+  // how often to run expensive state updates, defaults to 10 minutes
+  const updateRateS = Number(env.UPDATE_RATE_S || 10 * 60);
```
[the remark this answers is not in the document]
daywiss: I believe it was set to 100 blocks in prod, which was roughly 25 minutes. I've shortened that to 10 minutes by default because it seems like we want to get state updates in a bit quicker when things expire.  This is configurable though.

packages/api/src/services/emp-state.ts
```diff
@@ -140,6 +140,8 @@ export default (config: Config, appState: Dependencies) => {
         staticState = await readEmpStaticState(instance, address);
         // create active emp with static/dynamic state
         await emps.active.create({ ...staticState, ...dynamicState });
+      } else {
+        await emps.active.update(address, dynamicState);
```
daywiss, line 144: emps suffered from the same issue, the case where they existed and were active, they were never being re-updated with dynamic data

mrice32:
> Also, pushing all the state updates in a queue and processing them in order is also a solution to keep in mind.

Interesting idea @amateima!

packages/api/src/services/lsp-state.ts
```diff
@@ -118,20 +119,26 @@ export default (config: Config, appState: Dependencies) => {
...
-        // create active emp with static/dynamic state
+        // create active lsp with static/dynamic state
         await lsps.active.create({
           ...optionalState,
           ...staticState,
           ...dynamicState,
           totalPositionCollateral,
         });
+      } else {
+        await lsps.active.update(address, {
```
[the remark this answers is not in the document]
mrice32: Does this _only_ happen with expiry? (that seems to be the focus of this if) Or does this mean it never updates at all?

mrice32 (approved):
LGTM! Nice find!

packages/api/src/services/lsp-state.ts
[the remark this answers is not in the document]
daywiss: good question, the original bug was around expiry, and this branch does not get executed on expiry, thats handled further up in the function. Your question made me realize that this actually is a general update bug while the contract is active, it was not ever updating while it was active, only when it expired.

PR #3349, opened 2021-09-03
improve(sdk): remove @uma/core references from sdk package

mrice32:
**Motivation**

We want to remove all uses of the core artifacts directly (or through the old get* methods in core) to allow us to remove the truffle build from core.

**Summary**

This removes all direct references to core in the sdk package.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/api/src/services/emp-registry.ts
+++ b/packages/api/src/services/emp-registry.ts
@@ -9,10 +9,10 @@ interface Config extends BaseConfig {
 }
 type Dependencies = Pick<AppState, "registeredEmps" | "provider">;
 
-export default (config: Config, appState: Dependencies) => {
+export default async (config: Config, appState: Dependencies) => {
   const { network = 1 } = config;
   const { registeredEmps, provider } = appState;
-  const address = registry.getAddress(network);
+  const address = await registry.getAddress(network);
   const contract = registry.connect(address, provider);
 
   async function update(startBlock?: number | "latest", endBlock?: number) {
--- a/packages/sdk/src/clients/registry/client.ts
+++ b/packages/sdk/src/clients/registry/client.ts
@@ -1,18 +1,14 @@
-import { EthersContracts } from "@uma/core";
-import assert from "assert";
-import RegistryArtifacts from "@uma/core/build/contracts/Registry.json";
+import { RegistryEthers, RegistryEthers__factory, getAddress as getAddressForContract } from "@uma/contracts-node";
 import type { SignerOrProvider, GetEventType } from "../..";
 import { Event } from "ethers";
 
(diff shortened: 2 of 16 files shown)


nicholaspai (approved):
LGTM nice to see @uma/core continuing to be removed

packages/api/src/services/emp-registry.ts
```diff
@@ -9,10 +9,10 @@ interface Config extends BaseConfig {
 }
 type Dependencies = Pick<AppState, "registeredEmps" | "provider">;
 
-export default (config: Config, appState: Dependencies) => {
+export default async (config: Config, appState: Dependencies) => {
   const { network = 1 } = config;
   const { registeredEmps, provider } = appState;
-  const address = registry.getAddress(network);
+  const address = await registry.getAddress(network);
```
amateima, line 15: Why adding async call here? `getAddress` returns a string directly

daywiss: i think the api changed. removing core references was what this was using under the hood, and now its using `contracts-node` to provide this call, which is probably async.

packages/sdk/src/clients/registry/client.ts
```diff
@@ -1,18 +1,14 @@
...
 import type { SignerOrProvider, GetEventType } from "../..";
 import { Event } from "ethers";
 
-export type Instance = EthersContracts.Registry;
-const Factory = EthersContracts.Registry__factory;
+export type Instance = RegistryEthers;
+const Factory = RegistryEthers__factory;
 
-export type Network = keyof typeof RegistryArtifacts.networks;
+export type Network = string | number;
```
daywiss, line 8: maybe theres no way to do this now, but before, typescript would complain if you used the wrong number. this is really minor though

daywiss (approved):
i imagine we may want to change this to the frontend version of contracts at some point, but we can look into that when the time comes. thanks for doing this!

amateima (approved):

packages/api/src/services/emp-registry.ts
mrice32: Yeah, exactly.

packages/sdk/src/clients/registry/client.ts
mrice32: Oh really sorry -- I didn't mean to merge before replying to this. The package just takes a normal number rather than a const type. I _think_ the reason I didn't do this in the contracts-node package is because in tests, you can connect to a local node that might have a different network id that's not hardcoded into the package. I'm guessing the sdk always assumes that it's being run on production networks?

PR #3350, opened 2021-09-03
fix(financial-templates-lib): minor updates to mocha config

mrice32:
**Motivation**

Determine the right mocha options.

**Summary**

This removes a few elements from the mocha config:
- `require`: I think this is only needed if the tests are written in typescript. All our tests (I think) are written in js, so I don't think we need this. (Side note: the tests reference the _compiled_ js src files, not the ts files directly, so there's no need for compilation there either I don't think).
- extension: I think this is `js` by default (source: https://mochajs.org/).
- watchExtensions: I couldn't find any reference to this option in the mocha docs.

@chrismaree I wanted to get your thoughts here since you probably understand these options better than I do.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


nicholaspai (approved):
LGTM

chrismaree (approved):
LGTM.

Can you not remove this on the core `.mocharc.js` file as well?

mrice32:
> LGTM.
> 
> Can you not remove this on the core `.mocharc.js` file as well?

Probably -- didn't realize there was one! Will do this in a separate PR.

PR #3351, opened 2021-09-07
feat(insured-bridge-relayer): initial relayer logic implementation

chrismaree:
**Motivation**

As part of building out the insured bridge, we need a basic implementation of the relayer bot. This PR adds some initial logic for the relayer as well as testing stubs.

**Summary**

- Adds `Relayer.ts` and associated tests.
- Added initial branching logic for the relayer.
- Refined how ABIs are passed around.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
```diff
@@ -76,6 +76,7 @@ contract OVM_BridgeDepositBox is OVM_CrossDomainEnabled, OVM_Testable {
     event SetMinimumBridgingDelay(uint64 newMinimumBridgingDelay);
     event WhitelistToken(address l1Token, address l2Token, uint64 lastBridgeTime, address bridgePool);
     event DepositsEnabled(address l2Token, bool depositsEnabled);
+    // TODO: change the order of these to match the way they are used in the bridge pool.
     event FundsDeposited(
```
nicholaspai, line 80: is it worth putting all the uint64's together at the top of the event? We should keep them together though so that Solidity can try to optimize space.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -72,21 +74,29 @@ export class InsuredBridgeL1Client {
...
 
-  getPendingRelayedDepositsForL1Token(l1Token: string) {
+  getPendingRelayedDepositsForL1Token(l1Token: string): Relay[] {
     return this.getRelayedDepositsForL1Token(l1Token).filter((relay: Relay) => relay.relayState === relayState.Pending);
   }
 
   // TODO: we might want to add other accessors that do other forms of filtering.
 
+  calculateRealizedLpFeesPctForDeposit(/* deposit: any*/): string {
+    return this.toWei("0.05");
```
nicholaspai, line 93: good placeholder for now, this function might get chunky

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -49,6 +49,9 @@ export class InsuredBridgeL2Client {
     return this.deposits[depositId.toString()];
   }
 
+  // TODO: consider adding a method that limits how far back the deposits will be returned from. In this implementation
+  // we might hit some performance issues when returning a lot of bridging actions
```
nicholaspai, line 53: empirically I've seen issues with Infura where it tries to fetch more than X blocks into the past. We should make sure that `index` sets the `startingBlockNumber` correctly.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -73,13 +77,31 @@ export class InsuredBridgeL2Client {
         instantRelayFeePct: depositRelayedEvent.returnValues.instantRelayFeePct,
         quoteTimestamp: Number(depositRelayedEvent.returnValues.quoteTimestamp),
       };
+      depositData.depositHash = this.generateDepositHash(depositData);
+      this.deposits[depositRelayedEvent.returnValues.depositId] = depositData;
```
nicholaspai, line 81: Is it OK if a deposit data for some reason has a hash of `""` or should we explicitly error out on this case? Currently you allow it.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -1,19 +1,128 @@
...
+enum RelayerMode {
+  OnlySlow,
+  OnlySpeedUp,
+  Any,
+}
+
+enum RelayType {
+  Slow,
+  Fast,
+  No,
```
nicholaspai, line 16: Is there a better Type name than "No"?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -1,19 +1,128 @@
...
+  ) {}
+
+  // TODO: consider refactoring this further into multiple methods. One to build the object of relayableDeposits and one
+  // that processes relayableDeposits and calls respective relaying functions.
+  async checkForPendingDepositsAndRelay() {
+    this.logger.debug({ at: "Relayer", message: "Checking for pending deposits and relaying" });
+    const relayableDeposits: { [key: string]: any } = {};
+    for (const l1Token of this.whitelistedRelayL1Tokens) {
+      this.logger.debug({ at: "Relayer", message: "Checking relays for token", l1Token });
+      // TODO: consider limiting how far back we look in this call size.
```
nicholaspai, line 47: I think this limit is outside the scope of this file. It should be handled by the `startingBlockNumber` param for the L2 client, and the `index.js` file that constructs the client.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -1,19 +1,128 @@
...
+    readonly account: string,
+    readonly relayerMode: RelayerMode = RelayerMode.Any
+  ) {}
+
+  // TODO: consider refactoring this further into multiple methods. One to build the object of relayableDeposits and one
+  // that processes relayableDeposits and calls respective relaying functions.
+  async checkForPendingDepositsAndRelay() {
+    this.logger.debug({ at: "Relayer", message: "Checking for pending deposits and relaying" });
+    const relayableDeposits: { [key: string]: any } = {};
+    for (const l1Token of this.whitelistedRelayL1Tokens) {
```
nicholaspai, line 45: nit: might be worth running these promises in parallel.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -1,19 +1,128 @@
...
+
+  private shouldRelay(deposit: Deposit, botTokenBalance: string): RelayType {
+    // shouldRelay(deposit) => RelayType { No, Slow, Fast }
+    // Checks bot balance.
+    // If bot balance is > transfer size, simulate the fast relay to determine the cost.
+    // If bot balance is > bond, simulate the slow relay to determine the cost.
+    // If the max profit across the two (or all that succeeded) is negative or 0, return No.
+    // Else return the one with the most profit.
+
+    // TODO: Add logic to decide if a relay action should happen
```
nicholaspai, line 105: Do we need to do any validation of the L2 deposit here? Or is that handled by the client?

nicholaspai (commented):
Looks great, I left mostly questions. Ping me when you answer them all!

packages/core/contracts-ovm/insured-bridge/implementation/OVM_BridgeDepositBox.sol
chrismaree: ye, this is a good idea. we should leave it until later though and go through looking for these kinds of optimisations.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: ye, for sure. we still need to decide exactly how we want to structure this as well. I think it'll likely follow as closely to compounds model as posible. see https://compound.finance/documents/Compound.Whitepaper.pdf for the section on `Interest Rate Model`.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
chrismaree: 100p

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
chrismaree: ye, that's a good point and it should defo error out in that case. I can change it to error in the `this.generateDepositHash` method.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: I hope so. what do you recommend? `NotProfitable` or `DontRelay`?

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: ye, I think there are a few optimisations. I want to keep things as simple, short and clean as posible for now. we can always improve speed later on.

With that said, there are no promises in this loop. all methods calls in this particular loop are synchronous.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: I think by it including in the client it is "valid". However, we do need to build out the Insured bridge price feed that will define the "validity" of a relay action. However, this is less important in this part of the infrastructure as the relayer will just simply do nothing if a bridging action is already relayed. Only the disputer bot needs to "validate" if something is valid or not and dispute accordingly.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: yeah I think you should change that. Maybe use an `assert(hash !== "")` or something even more specific

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: Yeah maybe `DontRelay` or `IgnoreRelay` would be more explicit

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
chrismaree: cool. added `if (depositHash == "" || depositHash == null) throw new Error("Bad deposit hash");`

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: used `IgnoreRelay`

packages/financial-templates-lib/test/clients/InsuredBridgeL2Client.js
```diff
@@ -34,6 +34,24 @@ const instantRelayFeePct = toWei("0.005");
 const quoteTimestampOffset = 60; // 60 seconds into the past.
 
 describe("InsuredBridgeL2Client", () => {
+  const generateDepositHash = (depositData) => {
```
nicholaspai, line 37: Can you not re-use the L2Client's `generateDepositHash` function?

nicholaspai (approved):
Looks great, I noted one place where we can optimize

packages/financial-templates-lib/test/clients/InsuredBridgeL2Client.js
chrismaree: we could but it's a bit janky to use the thing you are testing in the test 🤔

nicholaspai: true. Can ignore this thread now.

PR #3356, opened 2021-09-07
feat(financial-templates-lib): Stub for InsuredBridgePriceFeed

nicholaspai:
**Motivation**

OptimisticOracle proposer needs to be able to determine if an insured-bridge relay request is valid. For identifiers linked to such relays, we need a pricefeed that returns TRUE/FALSE about whether a relay is valid.

**Details**

Removed any relay ancillary data params that were unused by pricefeed to validate a deposit.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -120,7 +123,14 @@ export class InsuredBridgeL1Client {
       ]);
 
       for (const depositRelayedEvent of depositRelayedEvents) {
+        // TODO: Pull this async function call into above Promise.all if possible.
+        const relayDataForDeposit = await this.bridgePools[depositRelayedEvent.returnValues.l1Token].methods
+          .relays(depositRelayedEvent.returnValues.depositHash)
+          .call();
         const relayData = {
+          bridgePoolAddress: depositRelayedEvent.address,
```
nicholaspai, line 131: These are only added for convenience if we want the InsuredBridgePriceFeed to be able to reconstruct the relay ancillary data.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  public async getHistoricalPrice(time: number | string): Promise<BN> {
+    // Grab price request for a timestamp and parse the ancillary data for the request.
+    const matchedRelays = this.relays.filter((relay: Relay) => relay.relayTimestamp === time);
+    if (matchedRelays.length > 1) throw new Error("TODO: Handle multiple relays for same price request timestamp");
+    if (matchedRelays.length === 0) throw new Error("No price request for time");
+    const relay = matchedRelays[0];
+
+    // TODO: Not 100% sure we should reconstruct the relay ancillary data using deposit/relay data from the L2/L1
```
nicholaspai, line 104: @chrismaree @mrice32 want to draw your attention here. I originally thought we needed to parse the price request's ancillary data to determine relay params, but alternatively we can just use the clients data. WDYT?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,194 @@
...
+  No,
+}
+
+interface Params {
+  logger: Logger;
+  web3: Web3;
+  bridgeAdminAddress: string;
+}
+
+interface Relay {
```
chrismaree, line 31: nit: a lot of these interfaces are exported from the clients. consider importing them from there rather than re-defining them here.

chrismaree: see here: https://github.com/UMAprotocol/protocol/blob/a350bd9d1fc9a8c58b4a57f58fee62e7cfd75141/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts#L6 as an example of this import

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,194 @@
...
+  private deposits: Deposit[] = [];
+
+  /**
+   * @notice Constructs the InsuredBridgePriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider to connect to Ethereum network.
+   * @param {String} bridgeAdminAddress Address of BridgeAdmin contract on L1.
+   * @param
+   */
+  constructor({ logger, web3, bridgeAdminAddress }: Params) {
```
chrismaree, line 86: I think it would be better to feed in the `l1Client` and `l1Client` in the constructor thereby leaving the logic for creating these clients outside of the feed, within the `CreatePriceFeed.ts` utility. In doing so, the price feed does not need to know anything about how the clients are set up and it can _assume_ that it is fed a correct price feed.

If the feed is created from the disputer bot then the create price feed will make these clients. else, if the feed is created from the relayer, then these feeds are already instantiated and can be simply loaded in.

does this make sence and do you agree?

nicholaspai: I agree

nicholaspai: Changed!

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Got it, done

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -130,13 +132,18 @@ export class InsuredBridgeL1Client {
...
+        const relayData: Relay = {
+          quoteTimestamp: Number(depositRelayedEvent.returnValues.quoteTimestamp),
+          relayTimestamp: Number(relayDataForDeposit.priceRequestTime.toString()),
           depositId: Number(depositRelayedEvent.returnValues.depositId),
           sender: depositRelayedEvent.returnValues.sender,
           slowRelayer: depositRelayedEvent.returnValues.slowRelayer,
           disputedSlowRelayers: [],
           instantRelayer: ZERO_ADDRESS,
-          depositTimestamp: depositRelayedEvent.returnValues.depositTimestamp,
+          depositTimestamp: Number(depositRelayedEvent.returnValues.depositTimestamp),
```
nicholaspai, line 146: Not sure why TS didn't catch this before but the type of `depositTimestamp` is declared `number` in the `Relay` interface.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+
+interface Params {
+  logger: Logger;
+  web3: Web3;
+  l1Client: InsuredBridgeL1Client;
+  l2Client: InsuredBridgeL2Client;
+}
+
+// Allows user to respond to a "relay" price request that was sent in response to a "deposit" on a InsuredBridge
+// deployed to an L2 network. The relay price request is submitted on L1. This pricefeed will respond True or False
```
chrismaree, line 23:
```suggestion
// deployed to an L2 network. The relay price request is submitted on L1. This pricefeed will respond "Yes" or "No"
```

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+  private readonly convertPrice: (_number: number) => BN;
+
+  /**
+   * @notice Constructs the InsuredBridgePriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider to connect to Ethereum network.
+   * @param {Object} l1Client Fetches and returns latest state of L1 pool and admin contracts.
+   * @param {Object} l2Client Fetches and returns latest state of L2 deposit contract.
+   */
+  constructor({ logger, web3, l1Client, l2Client }: Params) {
```
chrismaree, line 43: why did you make this an object? is this not somewhat non-standard with other price feeds?

chrismaree: I think it might be better to follow the pattern done in other places for this kind of constructor. it will also mean you can use the `readonly` syntax which means you wont have to define `this.x` instance of each variable.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+    super();
+
+    this.decimals = 18;
+    this.logger = logger;
+    this.web3 = web3;
+    this.l1Client = l1Client;
+    this.l2Client = l2Client;
+    this.convertPrice = (_number) => toBN(toWei(_number.toString()));
```
chrismaree, line 51: I think `convertPrice` is not a great name for this. `toBNWei` might be better. we do something like this here: https://github.com/UMAprotocol/protocol/blob/1fca52652a4995369bee2a89542f43396d502bd3/packages/trader/src/RangeTrader.ts#L6

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  // This method returns the validity of a relay price request attempt. The relay request was valid if and only if it
```
chrismaree, line 54:
```suggestion
  // This method returns the validity of a relay price request attempt. The relay request was valid if and only if it:
```

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+  // (1) corresponds with an L1 relay transaction submitted to the appropriate BridgePool contract and
+  // (2) the L1 relay corresponds with an L2 deposit transaction submitted to the appropriate DepositBox contract.
+  // If the relay request does not meet these conditions, then this method will return a price of 0, implying "No, the
+  // relay was not valid".
```
chrismaree, line 58:
```suggestion
  // relay was not valid" and the relay action will be disputed.
```

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.decimals = 18;
+    this.logger = logger;
+    this.web3 = web3;
+    this.l1Client = l1Client;
+    this.l2Client = l2Client;
+    this.convertPrice = (_number) => toBN(toWei(_number.toString()));
+  }
+
+  // This method returns the validity of a relay price request attempt. The relay request was valid if and only if it
+  // (1) corresponds with an L1 relay transaction submitted to the appropriate BridgePool contract and
```
chrismaree, line 55: There is a pretty good argument that we don't actually want to validate price requests where the `requestor` is not an official `BridgePool`. i.e if I request a price for this identifier with invalid ancillary data, rather than the bot disputing it the bot could "do nothing" and ignore it. 

This could be done by adding in a third state into the `isRelayValid` which represents an `Ignored` State (i.e something that is a price request for the given identifier but we don't want to do anything with it.

I still think the way you have it set up now is correct and logical as otherwise, it would go against the UMIP.

I Just want to start this discussion as it would not actually impact the bridge negatively if we did not dispute these requests.

I'm curious what @mrice32 thinks about this as well

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
```diff
@@ -0,0 +1,375 @@
...
+    relayAncillaryData = await bridgePool.methods.getRelayAncillaryData(depositData, relayData).call();
+    relayAncillaryDataHash = soliditySha3(relayAncillaryData);
+    return { depositHash, relayAncillaryData, relayAncillaryDataHash };
+  };
+
+  before(async function () {
+    accounts = await web3.eth.getAccounts();
+    [owner, depositor, relayer, liquidityProvider, recipient] = accounts;
+
+    finder = await Finder.new().send({ from: owner });
```
chrismaree, line 111: nothing to change right now but I think we should set aside some time to refactor out a lot of this common preamble that a lot of these bridge tests contain.

chrismaree (approved):
This is excellent work! clean code, well stubbed out for future features.

Next step should be to add the associated logic to the `CreatePriceFeed` to formulate the clients accordingly

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: I agree in general about consistency but other pricefeeds like FundingRateMultiplierPriceFeed and VaultPriceFeed also use this constructor logic and I think in general that using an object is better. WDYT?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: done

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: > I still think the way you have it set up now is correct and logical as otherwise, it would go against the UMIP.

I think this is a really key point

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
nicholaspai: yeah i copied these directly from L1Client and L2Client

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
mrice32: I think the only reason we would want to discriminate is if we were worried about someone spamming the OO with proposals, which would cost them a considerable amount of money (would be similar economics to spamming the DVM).

As discussed in 1:1 with @chrismaree, I think the only exploit scenario there would be trying to deplete bot capital so you can push a truly malicious proposal through liveness. If we assume there's enough redundancy, resiliency, and deep enough pockets in the bot network to resist an attacker, I don't think this poses too big of an issue.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
mrice32: I think either way is fine. Only thing that's a little annoying about using an object is that you end up declaring the Params type and you can't do stuff like this
```ts
constructor(
  readonly private logger: Logger,
  readonly private web3: Web3,
  ...
)
```
to automatically declare instance variables in-line so you don't need to do:
```ts
private readonly logger: Logger;
private readonly web3: Web3;

....
this.logger = logger;
this.web3 = web3;
```

So basically the other way is just a bit more concise in ts, but either way is fine, and I can still see arguments for doing it this way 🤷 .

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+  // This method returns the validity of a relay price request attempt. The relay request was valid if and only if it:
+  // (1) corresponds with an L1 relay transaction submitted to the appropriate BridgePool contract and
+  // (2) the L1 relay corresponds with an L2 deposit transaction submitted to the appropriate DepositBox contract.
+  // If the relay request does not meet these conditions, then this method will return a price of 0, implying "No, the
+  // relay was not valid" and the relay action will be disputed.
+
+  // For example, if a malicious actor were to submit a price request directly to the Optimistic
+  // Oracle using the Insured Bridge identifier, then this method would return a price of 0 since there was no
+  // associated relay for the price request.
+  public async getHistoricalPrice(time: number | string, verbose: boolean, ancillaryData: string): Promise<BN> {
```
mrice32, line 64: nit: should we move verbose to the end since it can often be ignored? This may induce a bit more churn, but might be a bit cleaner.

Other nit: isn't verbose also optional (shouldn't it be `verbose?: boolean` ?).

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+    // First get all relay events emitted at the request timestamp.
+    const matchedRelays = this.relays.filter((relay: Relay) => relay.relayTimestamp === time);
+
+    // Next, Parse ancillary data for relay request and find relay if possible with matching params.
+    const parsedAncillaryData: any = parseAncillaryData(ancillaryData);
```
mrice32, line 69: nit: I think this returns a Record<string, unknown>. I think there might be two slightly preferable options for handling this.
1. Have a type that you expect for the shape of the ancillary data and cast it to that (unknown is freely castable to anything I think).
2. (more work) have a function that validates the fields and returns a well-defined type or errors.


Thoughts?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Yeah this is future technical debt if we don't move verbose all the way to the end I'll change that. Regarding the other nit, i'll remove `verbose` from this function since its no longer neccessary to define.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+        deposit.slowRelayFeePct === relay.slowRelayFeePct &&
+        deposit.instantRelayFeePct === relay.instantRelayFeePct &&
+        deposit.quoteTimestamp === relay.quoteTimestamp
+    );
+    if (matchedDeposits.length > 1) {
+      this.logger.error({
+        at: "InsuredBridgePriceFeed",
+        message: "TODO: Handle multiple deposits associated with same relay attempt",
+      });
+      throw new Error("TODO: Handle multiple relays for same price request timestamp");
```
mrice32, line 117: Do we need to log and throw?
- Won't the throwing result in a logger.error somewhere up the stack?
- I thought throwing was better just because we can catch the errors and emit later so it doesn't interfere with other stuff later in the loop, but maybe I'm misunderstanding how our bots typically handle price feed errors.

You and @chrismaree may know better what the _right_ process is here.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+    } else if (matchedDeposits.length === 0) {
+      this.logger.debug({
+        at: "InsuredBridgePriceFeed",
+        message: "No deposit event matching relay attempt",
+      });
+      return this.toBNWei(isRelayValid.No);
+    }
+
+    // Validate relays proposed realized fee percentage.
+    const expectedRealizedFeePct = this.l1Client.calculateRealizedLpFeesPctForDeposit(/* matchedDeposits[0] */);
```
mrice32, line 127: Why is this commented out? Because this function doesn't take a parameter yet?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+    if (expectedRealizedFeePct !== relay.realizedLpFeePct) {
+      this.logger.error({
+        at: "InsuredBridgePriceFeed",
+        message: "Matched deposit with relay but realized fee % is incorrect",
+      });
+      return this.toBNWei(isRelayValid.No);
+    }
+
+    // TODO: Do we need to check other parameters like slow relayer address and deposit box contract address? These
```
mrice32, line 136: Hmm interesting point. Why did we include these in ancillary data?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+    return null;
+  }
+
+  public getLookback(): number | null {
+    // TODO.
+    return null;
+  }
+
+  public getCurrentPrice(): BN | null {
+    // TODO. This doesn't seem appropriate for this pricefeed, perhaps it should always return null.
```
mrice32, line 154: This seems like it _could_ be implemented almost the same as the historical one except we could just drop the filtering by timestamp since the request should be unique even without the relay timestamp, right?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+    return null;
+  }
+
+  public getPriceFeedDecimals(): number {
+    // TODO.
```
mrice32, line 159: This method looks to be done?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,172 @@
...
+  }
+
+  public getLastUpdateTime(): number | null {
+    // TODO.
+    return null;
+  }
+
+  public getLookback(): number | null {
+    // TODO.
+    return null;
```
mrice32, line 150: Do the clients have any notion of a lookback? It seems like that would be the only limit to the lookback on this price feed.

mrice32 (commented):
This looks awesome! Really nuanced and tricky to write this.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Yep exactly! But this is the probably interface

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: True! I'll implement. So IIUC this method will rarely if at all be used in production because the current time stamp isn't that predictable?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: yep you're right, i'll change that

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: So I thought to include the `depositBoxContract` address because we want users to know where to look on L2 for the `Deposit` event, and technically the `depositBoxAddress` mapped to the `BridgePool` can change after the `Deposit` has emitted. If the `depositBoxAddress` changes then how would users know which address to look up? I think this is probably over-engineering but curious what you think @chrismaree . One change we might consider making is adding the `depositBoxAddress` to the `Relay` struct stored in the `BridgePool` contract. This would be a straightforward change to make.

Regarding the `slowRelayer`, I think you can ignore this comment since we check the `slowRelayer` included in the `relayAncillaryData` against the `relayData.slowRelayer` fetched by the `l1Client`. I'll change that comment.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: I'm going to keep it as an object since it seems easier for clients and we should backlog a task to update all pricefeed constructors to use objects.

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
nicholaspai: I'm going to skip this in this PR for the sake of conciseness.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: I went with option 1 and casted `unknown` to a new `RelayAncillaryData` interface.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Since this is probably re-uses `getHistoricalPrice` I'll save for follow up PR, I really just want to come to consensus on the main `getHistoricalPrice` implementation

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: True. Removed comment.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Yeah this is a tricky question, but one I'll save for a follow up PR.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Actually, looks like the `depositBoxAddress` is already emitted in the `DepositRelayed` L1 event and I already check it. So this comment can be removed.

packages/core/scripts/local/getHistoricalPrice.js
```diff
@@ -11,11 +11,12 @@
  *         --network mainnet_mnemonic
  *         --identifier USDBTC
  *         --time 1601503200
+ *         --ancillaryData 0x123abc
  */
 const { fromWei } = web3.utils;
 const { createReferencePriceFeedForFinancialContract, Networker } = require("@uma/financial-templates-lib");
 const winston = require("winston");
-const argv = require("minimist")(process.argv.slice(), { string: ["identifier", "time"] });
+const argv = require("minimist")(process.argv.slice(), { string: ["identifier", "time", "ancillaryData"] });
```
nicholaspai, line 19: @smb2796 FYI this will be a new CLI flag that the user can specify

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: > If we assume there's enough redundancy, resiliency, and deep enough pockets in the bot network to resist an attacker, I don't think this poses too big of an issue.

I think this is a reasonable assumption, and one we make in using the OptimisticOracle contract in the first place. 

> There is a pretty good argument that we don't actually want to validate price requests where the requestor is not an official BridgePool. i.e if I request a price for this identifier with invalid ancillary data, rather than the bot disputing it the bot could "do nothing" and ignore it.
> This could be done by adding in a third state into the isRelayValid which represents an Ignored State (i.e something that is a price request for the given identifier but we don't want to do anything with it.

So if we were to go down this route, which type of relay requests would we ignore? 
- Those whose timestamps do not correspond to a relay event
- Those with empty or unparseable ancillary data
- Those whose ancillary data does not match any relay event

In this case, we would only return a price of 0 ("No") in the following cases:
- The matched relay event does not correspond with a deposit event
- The matched relay fee % is incorrect.

Thoughts?

packages/core/scripts/local/getHistoricalPrice.js
smb2796: Cool, thanks for the heads up. @Reinis-FRP for visibility.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
mrice32: > So IIUC this method will rarely if at all be used in production because the current time stamp isn't that predictable?

Yeah, if it's _only_ used by bots that use the historical price, then yes. But, for instance, if we wanted to validate the relay before we sent it in the relay bot, we would probably use this method.

Regardless, for this identifier, I don't think the timestamp actually matters when determining if the relay is valid or not, so this method has a reasonable value that we could return and I think it should be identical to what the historical method returns for the same ancillary data unless I'm missing something.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
chrismaree: I agree with this. We'd need to add some notion of an `Ignore` flag in the case we want the bot to do nothing 🤔

nicholaspai: As discussed offline i think we should actually return `null`, which will cause the OO proposer to ignore these requests as it does currently.

mrice32: This may be a super basic question, but what is this price feed really supposed to do/be responsible for? It seems odd that we care about what's happening on L1, since this is only trying to validate what an L1 relayer would be submitting on L1? Shouldn't it be possible to generate (and validate) that without knowing anything about what's happening on L1 (or before you even submit it to L1!)?

Some of the L1 considerations make this seem like we're mingling the responsibility of the OO or relay bot with the responsibility of a _price feed_, which is just meant to give you the price value for an identifier, timestamp, and ancillary data.

Am I missing something here?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
mrice32: SGTM!

mrice32 (approved):
LGTM! I still want to come back to this comment: https://github.com/UMAprotocol/protocol/pull/3356#discussion_r705603167. But overall, this PR looks great!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -534,6 +534,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         );
 
         // Add relay data.
+        intermediateAncillaryData = AncillaryData.appendKeyValueUint(
```
nicholaspai, line 537: By adding the request time to the ancillary data we can compare it against the time passed into `getHistoricalPrice` for further validation. This is useful for two reasons:

- By adding the timestamp to the ancillary data, any slow relayer who had their relay disputed can now submit another relay for the same deposit once the block time has changed.
- (maybe) This provides additional validation that a relay's ancillary data is valid

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  // This method returns the validity of a relay price request attempt. The relay request was valid if and only if it:
+  // The price request's ancillary data contains parameters that match with an L2 deposit event.
+  public async getHistoricalPrice(time: number | string, ancillaryData: string): Promise<BN> {
+    // Parse ancillary data for relay request and find deposit if possible with matching params.
+    const parsedAncillaryData: RelayAncillaryData = (parseAncillaryData(
+      ancillaryData
+    ) as unknown) as RelayAncillaryData;
+    const matchedDeposit = this.deposits.filter(
```
nicholaspai, line 78: Note @mrice32 and @chrismaree this method has changed. L1 relay data is no longer checked. Instead, the ancillary data is compared against L2 deposit event data. 

- If the ancillary data matches and the realized fee % is correct, then return 1.
- All other cases return 0.

This means that if someone submits a relay price request/proposal with ancillary data matching a deposit, then the OO will not dispute it, even if there no corresponding relay event. I see no risk to this.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Continuing this conversation [here](https://github.com/UMAprotocol/protocol/pull/3356/files#r705700270). Per matt's insight, I agree we don't need L1 event data. We only need to compare request ancillary data with L2 event data.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
mrice32: For 1, shouldn't this be the case regardless? The DVM/OO will consider these two distinct requests since their timestamp is different even if we don't include it in ancillary data.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
mrice32: Are there any fields we're not checking here? If so, do you think that's a signal that they can be safely removed from the ancillary data since they aren't necessary to determine whether the request was valid or not?

mrice32 (approved):
Looks great! Very simple/clean implementation

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Yeah I removed `slowrelayer`

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -484,6 +484,8 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         view
         returns (bytes memory)
     {
+        // TODO: Consider adding `chainId` and `relayNonce` to establish relay uniqueness, and removing unneccessary
```
nicholaspai, line 487: I think in a future PR we can remove a lot of these params don't establish {deposit, relay} uniqueness.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -130,13 +132,18 @@ export class InsuredBridgeL1Client {
       ]);
 
       for (const depositRelayedEvent of depositRelayedEvents) {
-        const relayData = {
+        const relayDataForDeposit = await this.bridgePools[depositRelayedEvent.returnValues.l1Token].methods
```
chrismaree, line 135: hhmmm. I think we should not be doing this. this will make this method extremely slow as we will have async calls for each and every relay. is there no way for us to pull this out of the event? perhaps we need to modify the event to include this timestamp? ideally this method should run in constant time without needing to make extra calls per relay event.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  constructor({ logger, web3, l1Client, l2Client }: Params) {
+    super();
+
+    this.decimals = 18;
+    this.logger = logger;
+    this.web3 = web3;
+    this.l1Client = l1Client;
+    this.l2Client = l2Client;
+    this.toBNWei = (_number) => toBN(toWei(_number.toString()));
```
chrismaree, line 67: nit: I dont like this method. Can you define it like [this](https://github.com/UMAprotocol/protocol/blob/8bf6271fd13f66b04ba3036fb2f9f785f076fd4f/packages/trader/src/RangeTrader.ts#L6)?

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    // Parse ancillary data for relay request and find deposit if possible with matching params.
+    const parsedAncillaryData: RelayAncillaryData = (parseAncillaryData(
+      ancillaryData
+    ) as unknown) as RelayAncillaryData;
+    const matchedDeposit = this.deposits.filter(
+      (deposit: Deposit) =>
+        deposit.depositId === parsedAncillaryData.depositId &&
+        deposit.timestamp === parsedAncillaryData.depositTimestamp &&
+        deposit.recipient.substr(2).toLowerCase() === parsedAncillaryData.recipient &&
```
chrismaree, line 83: the other option would be to cast the ancillary data part to `toChecksumAddress` and then validate the equality. I think what you have here is _ok_ but curious what you think

chrismaree (commented):
I'm pretty happy with this implementation! Please see my outstanding comment on the async call within the L1 client. that's a problem and will make the client extremely slow as the number of relays increase

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Great catch, we no longer need to modify the `relayData` to verify a relay's ancillary data! I removed all changes to this client.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: Ah yep, will remove from class.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
nicholaspai: yeah I like this its a bit more readable. will change.

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -0,0 +1,140 @@
...
+    return null;
+  }
+
+  public getLookback(): number | null {
+    // TODO. We could use the L1/L2 client's starting block number and network average block propagation time to
+    // determine this value.
+    return null;
+  }
+
+  public getCurrentPrice(): BN | null {
```
chrismaree, line 123: it could also error out. when would we be calling this on this feed? seems like something is misconfigured if that was the case

chrismaree (approved):
This is excellent. good work @nicholaspai

PR #3357, opened 2021-09-07
improve(api): add created timestamp to emps

amateima:
**Motivation**
https://app.shortcut.com/uma-project/story/2262/add-created-timestamp-to-emp-contracts-in-api

**Summary**

Now the emp-registry service of the api stores the block number for each new `NewContractRegistered` event in order to be able to compute the block timestamp that's saved later in the emps state. The sdk needed to be modified in order to introduce the new attribute: `createdTimestamp`

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/api/src/services/emp-registry.ts
```diff
@@ -22,8 +22,10 @@ export default (config: Config, appState: Dependencies) => {
       endBlock
     );
     const { contracts } = registry.getEventState(events);
-    await Promise.map(Object.keys(contracts || {}), (x) => {
-      return registeredEmps.add(x);
+    await Promise.map(Object.keys(contracts || {}), (address) => {
+      const blockNumber = (contracts as any)[address]["blockNumber"];
```
amateima, line 26: @daywiss I'm not sure how to properly assign a type to the raw event

packages/api/src/index.ts
```diff
@@ -106,6 +106,7 @@ export type AppState = {
     };
   };
   registeredEmps: Set<string>;
+  registeredEmpsMetadata: Map<string, { blockNumber: number }>;
```
amateima, line 109: @daywiss Is it ok to use the `Map` instead of the sdk's `JsMap`?

daywiss: yea its ok for now

packages/api/src/services/emp-registry.ts
daywiss: hmm yea im not sure. Maybe you could use objects.entries, which would give you the key and value, but  you might need to still cast the value.  You could also import this type from the sdk and cast to this: https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/clients/registry/client.ts#L22

packages/api/src/services/emp-state.ts
```diff
@@ -145,6 +148,21 @@ export default (config: Config, appState: Dependencies) => {
...
+      await updateCreatedTimestamp(address, emps.active);
+    }
+  }
+
+  async function updateCreatedTimestamp(address: string, table: uma.tables.emps.JsMap) {
+    if (registeredEmpsMetadata.has(address)) {
+      const emp = await table.get(address);
+      const blockMetadata = registeredEmpsMetadata.get(address);
+
+      if (!emp.createdTimestamp && blockMetadata) {
```
daywiss, line 162: i dont think you need to check for blockMetadata since you do this first `registeredEmpsMetadata.has(address)`

packages/api/src/services/emp-state.ts
```diff
@@ -145,6 +148,21 @@ export default (config: Config, appState: Dependencies) => {
...
       // add any new sponsors
       await emps.active.addSponsors(address, eventState.sponsors || []);
+
+      // set created timestamp if needed
+      await updateCreatedTimestamp(address, emps.active);
+    }
+  }
+
+  async function updateCreatedTimestamp(address: string, table: uma.tables.emps.JsMap) {
+    if (registeredEmpsMetadata.has(address)) {
```
daywiss, line 158: this is just personal preference of mine, but i like to write guards that exit early rather than nest.
for instance 
```js
if(!registeredEmpsMetadata.has(address)) return
const emp = await table.get(address);
const blockMetadata = registeredEmpsMetadata.get(address);
if (emp.createdTimestamp) return
```

This is not strictly enforced so its up to you, but i find the flatter code a bit cleaner

packages/sdk/src/tables/emps/js-map.ts
```diff
@@ -11,9 +11,14 @@ export const Table = (type = "Emp") => {
     return table.update(id, { sponsors: Array.from(set.values()) });
   }
 
+  async function setTimestamp(id: string, timestamp: number) {
```
daywiss, line 14: this is a nice touch

daywiss (commented):
this is really clean and just what i had in mind. would be nice to fix the `any` cast, maybe @mrice32 has some suggestions on it

packages/api/src/services/emp-state.ts
amateima: I know, I did it to get rid of the possible `undefined` result returned by `.get()`. `provider.getBlock` function complained about it

packages/api/src/services/emp-state.ts
amateima: I totally agree with you. Sometimes I write guards, sometimes I do it using if else statements. I'll stick with guards from now

packages/api/src/services/emp-registry.ts
mrice32: This is the ethers event, right?

Is there a central location where we could put the `GetEventType` generic so it could be used in situations like this?

mrice32 (approved):
LGTM!

packages/api/src/services/emp-registry.ts
daywiss: the type is exposed in the sdk and can be used directly, which i think would be better than fetching it from `GetEventType`. Would be something like `uma.clients.registry.NewContractRegistered`. Is this what you mean?

amateima: It looks like I was not able to get the correct type because of some Visual Studio Code issue. I went through all the source code and I knew that the type is set correctly. 
I ran `yarn build` for `sdk` and `api` again, delete Visual Studio cache and now it works.

packages/api/src/services/emp-registry.ts
```diff
@@ -22,8 +22,12 @@ export default async (config: Config, appState: Dependencies) => {
       endBlock
     );
     const { contracts } = registry.getEventState(events);
-    await Promise.map(Object.keys(contracts || {}), (x) => {
-      return registeredEmps.add(x);
+    if (!contracts) return;
```
daywiss, line 25: nice fix

PR #3358, opened 2021-09-07
feat(common): add getWeb3 by chainId

mrice32:
**Motivation**

We'd like to be able to have processes that use multiple nodes at once (beyond just price feeds).

**Summary**

This adds a function similar to getWeb3 that will allow processes to create a web3 instance connected to a node according to a provided chain id.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A


packages/common/src/ProviderUtils.ts
```diff
@@ -60,6 +65,68 @@ export function createBasicProvider(nodeRetryConfig: RetryConfig[]): RetryProvid
...
+    }
+  }
+  return "none";
+}
+
+function addMnemonicToProvider(
+  provider: AbstractProvider,
+  mnemonic: string = process.env.MNEMONIC ||
+    "candy maple cake sugar pudding cream honey rich smooth crumble sweet treat",
+  numKeys = process.env.NUM_KEYS ? parseInt(process.env.NUM_KEYS) : 2,
```
chrismaree, line 85: would it be cleaner to do this:
```suggestion
  numKeys = parseInt(process.env.NUM_KEYS || 2),
```

packages/common/src/ProviderUtils.ts
```diff
@@ -60,6 +65,68 @@ export function createBasicProvider(nodeRetryConfig: RetryConfig[]): RetryProvid
...
+      return provider;
+  }
+}
+
+/**
+ * @notice Creates a web3 instance for a particular chain.
+ * @param chainId the chain id for the network the user wants to connect to.
+ * @returns new Web3 instance.
+ */
+export function getWeb3ByChainId(chainId: number): Web3 {
```
chrismaree, line 113: nice! this is super clean. Good call on making a separate exported function. also, the notion of `nodeUrl` set to `test` is pretty good as well

chrismaree (approved):
This is excellent! nice and simple and gets the job done without too much complexity

packages/common/src/ProviderUtils.ts
```diff
@@ -8,6 +8,11 @@ import minimist from "minimist";
 import Url from "url";
 import { RetryProvider, RetryConfig } from "./RetryProvider";
 import { AbstractProvider } from "web3-core";
+import HDWalletProvider from "@truffle/hdwallet-provider";
```
nicholaspai, line 11: Does this tie us more to `truffle`? Is there a way to use hardhat?

packages/common/src/ProviderUtils.ts
mrice32: Oh actually, I don't think this compiles because parseInt doesn't take ints. Will convert to a string.

packages/common/src/ProviderUtils.ts
mrice32: Great point!

- I'm not aware of any alternatives provided by hardhat or otherwise, especially that support web3 and not just Ethers.
- This is mainly just a simple web3 provider that just happens to be part of the truffle monorepo. It shouldn't add any reliance to the core of truffle or to using truffle tools since it is immediately plugged into web3.

If you have an equivalent alternative, I'm happy to use it! I think hardhat deploy has an hd wallet config that it plugs into an internal hd wallet implementation, but I don't think that implementation is exported on its own so that you can use it outside the hardhat context (directly with web3, for instance).

nicholaspai: Yeah, while I was digging through the `hardhat` docs I couldn't find an equivalent for HDWalletProvider and I agree that it doesn't expand the truffle dependence anyways.

nicholaspai (approved):

PR #3359, opened 2021-09-10
feat(insured-bridge-relayer): Add additional relay logic, should relay and slow relay implementation

chrismaree:
**Motivation**

As part of the relayer bot, there is a number of key pieces of logic that need to be built out. This pr implements the notion of `shouldRelay` which takes in relay information and makes a decision regarding if a relay should be bridged over based on its profitability.

**Summary**
This PR also makes a number of ts and interface changes for the relayer bot.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -1,18 +1,28 @@
...
+import { Deposit } from "./InsuredBridgeL2Client";
+
 import type { BridgeAdminWeb3, BridgePoolWeb3 } from "@uma/contracts-node";
-import Web3 from "web3";
+import type { BN } from "@uma/common";
 import type { Logger } from "winston";
-import { Deposit } from "./InsuredBridgeL2Client";
 
-enum relayState {
+enum RelayState {
```
chrismaree, line 12: enums should be caps

packages/financial-templates-lib/src/proxy-transaction-handler/DSProxyManager.ts
```diff
@@ -1,12 +1,13 @@
 import assert from "assert";
 
 import { getFromBlock, runTransaction } from "@uma/common";
-import type { Logger } from "winston";
+import { TransactionReceipt } from "web3-eth";
```
chrismaree, line 4: small refactor now that the types in common contain `TransactionType`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -1,53 +1,61 @@
...
+import {
+  InsuredBridgeL1Client,
+  InsuredBridgeL2Client,
+  GasEstimator,
+  Deposit,
+  RelayAbility,
+} from "@uma/financial-templates-lib";
+import { getTokenBalance } from "./RelayerHelpers";
 
-enum RelayerMode {
```
chrismaree, line 7: I decided we dont need this for v1

packages/insured-bridge-relayer/src/RelayerHelpers.ts
```diff
@@ -0,0 +1,47 @@
+import Web3 from "web3";
```
chrismaree, line 1: I added this set of helpers to do some common things the relayer will need to do. in particular, see the `approveL1Tokens` method

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -1,17 +1,27 @@
...
   Pending,
   SpedUp,
   Disputed,
   Finalized,
 }
 
+export enum RelayAbility {
```
nicholaspai, line 19: This is kind of a weird name but I get what you're going for. You want to enumerate whether a Relay can be submitted, and what type. Why can't you just re-use the following `RelayState` types:

```
Any --> Pending || (some state before Pending like Uninitialized)
SpeedUpOnly --> Pending
None --> SpedUp || Disputed || Finalized
```

You'd only need to add an Uninitialized state

nicholaspai: Or another idea, since I also don't like the idea of mixing _contract_ state and off-chain client state, rename this enum to `ClientRelayState {Uninitialized, Pending, Finalized}` to distinguish between on-chain and off-chain state. We could moreover rename `RelayState` to `ContractRelayState`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
       this.logger.debug({ at: "Relayer", message: "Processing relayable deposits for L1Token", l1Token });
       for (const relayableDeposit of relayableDeposits[l1Token]) {
-        const botL1TokenBalance = await this.fetchBotL1TokenBalance(l1Token);
-        const relayType = this.shouldRelay(relayableDeposit, botL1TokenBalance);
-        switch (relayType) {
-          case RelayType.IgnoreRelay:
-            this.logger.info({
-              at: "Relayer",
+        const realizedLpFeePct = await this.l1Client.calculateRealizedLpFeePctForDeposit(relayableDeposit.deposit);
+        const desiredRelayType = await this.shouldRelay(
```
nicholaspai, line 68: I think it might make sense to rename this variable to `shouldRelay` and rename the `RelayType` enum to `ShouldRelay{No, Slow, Instant}`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
+    const [l1TokenBalance, proposerBondPct] = await Promise.all([
+      getTokenBalance(this.l1Client.l1Web3, deposit.l1Token, this.account),
+      this.l1Client.getProposerBondPct(),
+    ]);
+    const relayTokenRequirement = this.getRelayTokenRequirement(deposit, proposerBondPct, realizedLpFeePct);
+
+    // There are three different kinds of profits that the bot can produce:
+    let slowProfit = toBN("0");
+    let speedUpProfit = toBN("0");
+    let instantProfit = toBN("0");
```
nicholaspai, line 125: How is instant profit diff from speedup profit?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
+    // Based on the bots token balance, and the relay state we can compute the profitability of each action.
+
+    // a) Balance is large enough to do a slow relay. No relay action has happened on L1 yet for this deposit.
+    if (l1TokenBalance.gte(relayTokenRequirement.slow) && relayAbility === RelayAbility.Any)
+      slowProfit = toBN(deposit.amount).mul(toBN(deposit.slowRelayFeePct)).div(fixedPointAdjustment);
+
+    // b) Balance is large enough to instant relay. Deposit is in any state except finalized (i.e can be slow relayed
+    // and sped up or only sped up.)
+    if (l1TokenBalance.gte(relayTokenRequirement.instant) && relayAbility !== RelayAbility.None)
```
nicholaspai, line 134: This is a case where renamed enums like I mention [here](https://github.com/UMAprotocol/protocol/pull/3359/files#r707783410) would improve readability.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
-
-    // TODO: Add logic to decide if a relay action should happen
-    console.log(deposit, botTokenBalance);
-    return RelayType.Slow;
+  private async shouldRelay(deposit: Deposit, relayAbility: RelayAbility, realizedLpFeePct: BN): Promise<RelayType> {
+    const [l1TokenBalance, proposerBondPct] = await Promise.all([
+      getTokenBalance(this.l1Client.l1Web3, deposit.l1Token, this.account),
+      this.l1Client.getProposerBondPct(),
+    ]);
+    const relayTokenRequirement = this.getRelayTokenRequirement(deposit, proposerBondPct, realizedLpFeePct);
```
nicholaspai, line 120: This is a nice way to compute and fetch these relay token amount requirements.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
+    // c) Balance is large enough to slow relay and then speed up. Only considered if no L1 action has happened yet as
+    // wont be able to do an instant relay if the relay has already been slow relayed. In that case, should speedUp.
+    if (
+      l1TokenBalance.gte(relayTokenRequirement.slow.add(relayTokenRequirement.instant)) &&
+      relayAbility == RelayAbility.Any
+    )
+      instantProfit = slowProfit.add(speedUpProfit);
+
+    // Finally, decide what action to do based on the relative profits.
+    if (instantProfit.gt(speedUpProfit) && instantProfit.gt(slowProfit)) return RelayType.Instant;
```
nicholaspai, line 146: I think you should use BN.max here for readbility and reducing any edge cases.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
+        web3: this.l1Client.l1Web3,
+        transaction: (slowRelayTx as unknown) as TransactionType,
+        transactionConfig: { gasPrice: this.gasEstimator.getCurrentFastPrice().toString(), from: this.account },
+        availableAccounts: 1,
+      });
+
+      if (receipt.events)
+        this.logger.info({
+          at: "InsuredBridgeRelayer#Relayer",
+          type: "Slow Relay executed  🐌",
```
nicholaspai, line 178: +1 emoji

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -56,73 +64,165 @@ export class Relayer {
...
 
-    return await l1Token.methods.balanceOf(this.account).call();
+  private getRelayTokenRequirement(
+    deposit: Deposit,
+    proposerBondPct: BN,
+    realizedLpFeePct: BN
+  ): { slow: BN; instant: BN } {
+    // bridged amount - the LP fee, - slow relay fee, - instant relay fee
+    return {
+      slow: toBN(deposit.amount).mul(proposerBondPct).muln(2).div(fixedPointAdjustment),
```
nicholaspai, line 217: Why divide proposer bond by 2?

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -37,22 +37,16 @@ export async function run(logger: winston.Logger, web3: Web3): Promise<void> {
...
-    // Fetch the deposit contract address from the bridge admin.
-    const bridgeDepositBoxAddress = await new web3.eth.Contract(
-      getAbi("BridgeAdmin"),
-      config.bridgeAdminAddress
-    ).methods
-      .depositContract()
-      .call();
+    const l1Client = new InsuredBridgeL1Client(logger, web3, config.bridgeAdmin);
 
     // TODO: this is right now using the same web3 object. this is wrong. it should use an L2web3 object.ƒ
```
nicholaspai, line 42:
```suggestion
    // TODO: this is right now using the same web3 object. this is wrong. it should use an L2web3 object.
```

nicholaspai (approved):
Great work, seems pretty straightforward and readable for the most part but I have mostly naming suggestions

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: ye, this is something I went back and forth a few times on and is something that is not immediately clear from the review. as you pointed out, we already have `RelayState` which contains pretty much the states you discussed:
```ts
enum RelayState {
  Pending,
  SpedUp,
  Disputed,
  Finalized,
}
```

my thinking with the `RelayAbility` structure is that there is a state _before_ `RelayState.Pending` when the deposit has been sent on L2 but nothing has been brought onto L1. this is encapsulated by `RelayAbility.Any` meaning that the state is pre-relay.

With that said, `ClientRelayState {Uninitialized, Pending, Finalized}` seems pretty reasonable and fits the same framing that I require. I've refactored accordingly :)

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: the diff is that: a) speedup is considering a slow relay that has not been sped up and b) instant is a deposit that has no relay on L1 at all. this bot needs to consider the 3 cases:
a) slow relay
b) speed up a slow relay and
c) instantly relay a deposit with no actions done on it.

these three capture these profits accordingly

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: for sure. it looks better now.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: ye, that's a mistake. I've fixed it an added additional comments.

PR #3360, opened 2021-09-10
improve(api): add created timestamp for LSPs

amateima:

**Motivation**

https://app.shortcut.com/uma-project/story/2281/add-created-timestamp-to-lsp-contracts-in-sdk

**Summary**

The api now holds additional info about the created event for LSPs contract. Based on the event's block number we are now able to serve the created timestamp for this type of financial contract

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/clients/lsp-creator/client.ts
+++ b/packages/sdk/src/clients/lsp-creator/client.ts
@@ -29,7 +29,7 @@ export type CreatedLongShortPair = GetEventType<Instance, "CreatedLongShortPair"
 
 export interface EventState {
   contracts?: {
-    [lspAddress: string]: CreatedLongShortPair["args"];
+    [lspAddress: string]: CreatedLongShortPair;
   };
 }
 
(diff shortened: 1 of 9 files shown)


packages/sdk/src/clients/lsp-creator/client.ts
```diff
@@ -29,7 +29,7 @@ export type CreatedLongShortPair = GetEventType<Instance, "CreatedLongShortPair"
 
 export interface EventState {
   contracts?: {
-    [lspAddress: string]: CreatedLongShortPair["args"];
+    [lspAddress: string]: CreatedLongShortPair;
```
amateima, line 32: @daywiss I needed to change the type of this event in order to have access to the block number. From what I saw this type is not imported in other places. `yarn build` runs successfully so I think we're fine.

daywiss: yea i think its fine. originally i changed this pattern after writing the emp registry because i wanted the data to be serializable, but it doesnt really matter for now.

daywiss (approved):
👍

PR #3369, opened 2021-09-16
feat(insured-bridge): Refactoring for gas savings

nicholaspai:
**Motivation**

The motivation for this PR is gas costs. This reduces the `relayDeposit` average gas cost from 1.2mil to 1mil.

**Summary**

`BridgePool.sol`:
- Removes `priceRequestAncillaryDataHash => depositHash` mapping and instead call the OO directly in `relayDeposit` and `speedupRelay` to determine if a relay was disputed.  
- Remove requirement that you can only speedup undisputed relays
- Remove L1 token from deposit data struct since its stored as contract state variable

`Relayer.ts`:
- Validate that pending relay is valid to avoid speeding up invalid relays

`L2Client`
- Add `depositContract` to deposit struct

`L1Client`
- Add `getRelayForDeposit` helper method that `Relayer` can use to validate relays against deposits.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -299,6 +299,9 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
      * if the deposit data is valid, then even if it is falsely disputed, the instant relayer will eventually get
      * reimbursed because someone else will be incentivized to resubmit the relay to earn slow relayer rewards. Once the
      * valid relay is finalized, the instant relayer will be reimbursed.
+     * @dev We also assume that the caller has validated off-chain that the relay data that they are speeding up is
+     * valid. If the relay is disputed (or eventually gets disputed), then the caller has no recourse to recover
+     * their funds. Therefore, the caller has the same responsibility as the disputer in validating the relay data.
      * @dev Caller must have approved this contract to spend the deposit amount of L1 tokens to relay. There can only
      * be one instant relayer per relay attempt.
      * @param _depositData Unique set of L2 deposit data that caller is trying to instantly relay.
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -302,21 +302,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
      * reimbursed because someone else will be incentivized to resubmit the relay to earn slow relayer rewards. Once the
      * valid relay is finalized, the instant relayer will be reimbursed.
      * @dev Caller must have approved this contract to spend the deposit amount of L1 tokens to relay. There can only
-     * be one instant relayer per relay attempt and disputed relays cannot be sped up.
+     * be one instant relayer per relay attempt.
      * @param _depositData Unique set of L2 deposit data that caller is trying to instantly relay.
      */
     function speedUpRelay(DepositData memory _depositData) public {
         bytes32 depositHash = _getDepositHash(_depositData);
         RelayData storage relay = relays[depositHash];
-        require(
-            _getOptimisticOracle().getState(
-                address(this),
-                bridgeAdmin.identifier(),
-                relay.priceRequestTime,
-                getRelayAncillaryData(_depositData, relay)
-            ) != OptimisticOracleInterface.State.Disputed,
-            "Cannot speed up disputed relay"
-        );
         require(
             relays[depositHash].relayState == RelayState.Pending && relays[depositHash].instantRelayer == address(0),
             "Relay can not be sped up"
--- a/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
+++ b/packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
@@ -65,7 +65,6 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
         uint64 depositId;
         address l2Sender;
         address l1Recipient;
-        address l1Token; // todo: we can remove this prop. it's not needed as this contract has a unique l1 token.
         uint256 amount;
         uint64 slowRelayFeePct;
         uint64 instantRelayFeePct;
@@ -230,7 +229,6 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
                 depositId: depositId,
                 l2Sender: l2Sender,
                 l1Recipient: l1Recipient,
-                l1Token: address(l1Token),
                 amount: amount,
                 slowRelayFeePct: slowRelayFeePct,
                 instantRelayFeePct: instantRelayFeePct,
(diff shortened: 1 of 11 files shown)


packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -312,6 +308,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
     function speedUpRelay(DepositData memory _depositData) public {
         bytes32 depositHash = _getDepositHash(_depositData);
         RelayData storage relay = relays[depositHash];
+        require(
```
nicholaspai, line 311: This adds 300k gas to this function. The total gas cost to slow+instant relay is roughly the same as before.

Before:
- relayDeposit = 1.2mil, speedupRelay = 0.01mil

After:
- relayDeposit = 1mil, speedupRelay = 0.3 mil

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -312,6 +308,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
+            _getOptimisticOracle().getState(
```
mrice32, line 312: Might be a dumb question, but why do we need to check this? Don't we throw out a request if it's disputed, but _save_ the fast relay? So theoretically (probably would never happen) a fast relayed could speed up a relay that's disputed, and that fast relay would apply whenever it gets re-proposed? Basically just wondering why we need to check this if the fast relater is responsible for checking the request validity anyway.

Not saying we should change this now -- just that we might want to discuss this more.

mrice32 (approved):
Looks good!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: >  So theoretically (probably would never happen) a fast relayed could speed up a relay that's disputed, and that fast relay would apply whenever it gets re-proposed?

This is a good point. I'll think on this I think you. might be on to something,

nicholaspai: I guess you could think of this line as protecting the fast relayer in the case where the (invalid) relay has an incorrect fast relay fee % and the fast relayer would receive an incorrect amount of fees. This protection might not be necceessary.

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
```diff
@@ -246,10 +237,19 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20 {
...
         bytes32 depositHash = _getDepositHash(depositData);
+
+        // If relay exists for deposit, check if it is disputed. If its disputed, then we can relay again, otherwise
+        // the relay is pending valid and we cannot re-relay.
         require(
-            (relays[depositHash].relayState == RelayState.Uninitialized ||
-                relays[depositHash].relayState == RelayState.Disputed),
-            "Deposit already relayed"
+            relays[depositHash].relayState == RelayState.Uninitialized ||
+                _getOptimisticOracle().getState(
```
chrismaree, line 245: how would it be posible for the relay state to be not `Uninitialized` AND the OO having a disputed Price? put another way, why do you need to check the OO here?

nicholaspai: 1) call `relayDeposit` for a deposit to make the relay state `Pending`.
2) Someone disputes the relay directly against the OO
3) Relay state is still `Pending` but there is an OO dispute

Thoughts?

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: I ended up removing the check for whether a relay is disputed because I agree that the onus is on the caller to determine whether a relay should be sped up. If the relay is valid, then the `instantRelayer` address is carried over. If the relay is invalid, then the instant relayer loses their funds.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -64,22 +65,36 @@ export class Relayer {
     for (const l1Token of Object.keys(relayableDeposits)) {
       this.logger.debug({ at: "Relayer", message: "Processing relayable deposits for L1Token", l1Token });
       for (const relayableDeposit of relayableDeposits[l1Token]) {
+        // If deposit has a pending relay, then validate its relay params (i.e. any data not included in the deposit
```
nicholaspai, line 68: This check is needed since `speedupRelay` can be called now on disputed relays. If the relay is invalid (i.e. it is disputable), then this bot should ignore it and not speed it up.

I'm actually not sure how we handled Disputed relays previously, as I didn't see any tests for it. Was this an oversight @chrismaree ?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -108,6 +123,14 @@ export class Relayer {
     }
   }
 
+  private async isRelayValid(relay: Relay, deposit: Deposit): Promise<boolean> {
+    return (
+      relay.realizedLpFeePct.toString() ===
+        (await this.l1Client.calculateRealizedLpFeePctForDeposit(deposit)).toString() &&
+      relay.depositContract === deposit.depositContract
```
nicholaspai, line 130: Testing an incorrect `depositContract` is a bit tricky. I'm not sure how to do it because the L1BridgePool links the L2BridgeDepositBox address at the time of the relay. Is this either a param that we don't need to check in this method or is there a way to test it?

mrice32 (approved):
LGTM!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
nicholaspai: > The only way the OO could have invalid for this exact request is if the relay has never been relayed before, right?

Correct. But imagine that a relay is `Pending` and is then disputed directly through the OO. The `RelayState` is still `Pending`, but we should now be able to call `relayDeposit` again on it.

nicholaspai: Note that the above is possible and creates two separate price requests because the `relayId` nonce is incremented between relays.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: great addition @nicholaspai . you are 100% correct that before now the bot actually did not validate the correctness of a relay before speeding it up. the previous implementation just processed the client information and assumed correctness.

the implementation you've got here seems good however I have a few comments on the `isRelayValid` method below.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: ye, I think we need to validate other params as well. Do you see a way we could use the `InsuredBridgePriceFeed` for this? Otherwise, we need to validate some of the other params as well IMHO.

chrismaree (approved):
changes here are good simplifications! We need to continue playing the gas golf to further decrease gas costs of the bridge methods when calling the OO.

I'm keen to run our contract calls through tenderly to profile the function calls and identify further optimizations!

packages/core/contracts-ovm/insured-bridge/implementation/BridgePool.sol
mrice32: Seems reasonable to me. I also see your logic @nicholaspai.

PR #3370, opened 2021-09-16
fix(api): better state updating loops and timeouts

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/stories/space/unsaved

**Summary**
This separates asyncronous loops between checking for new contracts and updating contract state. It seems like the contract state update loop stops after some point, so this will now throw an error if it takes too long.  Some utility functions were added to support these changes.

This also adds some event driven updates as we want to get contract state quickly after discovery.  Its important to note that this change introduces process dependencies between services which use events to communicate. The emp registry and emp state services must now always be within the same process, and the lsp-creator and lsp-state services must likewise also be in the same process. This tradeoff imo is acceptable since the services are highly related and this allows more realtime state updates.
**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/stories/space/unsaved

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -1,6 +1,7 @@
 import assert from "assert";
 import { ethers } from "ethers";
 import moment from "moment";
+import Events from "events";
 
 import { tables, Coingecko, utils, Multicall } from "@uma/sdk";
 
@@ -40,6 +41,9 @@ export default async (env: ProcessEnv) => {
   assert(updateRateS >= 1, "UPDATE_RATE_S must be 1 or higher");
   assert(detectContractsUpdateRateS >= 1, "DETECT_CONTRACTS_UPDATE_RATE_S must be 1 or higher");
   assert(priceUpdateRateS >= 1, "PRICE_UPDATE_RATE_S must be 1 or higher");
+
+  // services can emit events when necessary, though for now any services that depend on events must be in same process
+  const serviceEvents = new Events();
 
   // state shared between services
   const appState: AppState = {
@@ -126,7 +130,7 @@ export default async (env: ProcessEnv) => {
     // these services can optionally be configured with a config object, but currently they are undefined or have defaults
     blocks: Services.Blocks(undefined, appState),
     emps: Services.EmpState({ debug }, appState),
-    registry: await Services.Registry({ debug }, appState),
+    registry: await Services.Registry({ debug }, appState, (data) => serviceEvents.emit("empRegistry", data)),
     collateralPrices: Services.CollateralPrices({ debug }, appState),
     syntheticPrices: Services.SyntheticPrices(
       {
@@ -141,7 +145,9 @@ export default async (env: ProcessEnv) => {
     erc20s: Services.Erc20s({ debug }, appState),
     empStats: Services.stats.Emp({ debug }, appState),
     marketPrices: Services.MarketPrices({ debug }, appState),
-    lspCreator: await Services.MultiLspCreator({ debug, addresses: lspCreatorAddresses }, appState),
+    lspCreator: await Services.MultiLspCreator({ debug, addresses: lspCreatorAddresses }, appState, (data) =>
+      serviceEvents.emit("multiLspCreator", data)
+    ),
     lsps: Services.LspState({ debug }, appState),
     lspStats: Services.stats.Lsp({ debug }, appState),
     globalStats: Services.stats.Global({ debug }, appState),
@@ -156,7 +162,7 @@ export default async (env: ProcessEnv) => {
   await services.lspCreator.update(appState.lastBlockUpdate, initBlock.number);
   console.log("Got all LSP addresses");
 
-  await services.emps(appState.lastBlockUpdate, initBlock.number);
+  await services.emps.update(appState.lastBlockUpdate, initBlock.number);
   console.log("Updated EMP state");
 
   await services.lsps.update(appState.lastBlockUpdate, initBlock.number);
@@ -221,7 +227,7 @@ export default async (env: ProcessEnv) => {
   async function updateContractState(startBlock: number, endBlock: number) {
     await services.blocks.handleNewBlock(endBlock);
     // update everyting
-    await services.emps(startBlock, endBlock);
+    await services.emps.update(startBlock, endBlock);
     await services.lsps.update(startBlock, endBlock);
     await services.erc20s.update();
     await services.blocks.cleanBlocks(oldestBlock);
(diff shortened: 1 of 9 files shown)


amateima (approved):

packages/api/src/apps/api/app.ts
```diff
@@ -120,7 +130,7 @@ export default async (env: ProcessEnv) => {
     // these services can optionally be configured with a config object, but currently they are undefined or have defaults
     blocks: Services.Blocks(undefined, appState),
     emps: Services.EmpState({ debug }, appState),
-    registry: await Services.Registry({ debug }, appState),
+    registry: await Services.Registry({ debug }, appState, (data) => serviceEvents.emit("empRegistry", data)),
```
daywiss, line 133: some context here. this pr splits up contract discovery from contract state fetching, which means a contract could be discovered at some point, but then not have state until state is fetched during its separate update loop.  This adds a pattern to services where they can emit any kind of event to trigger external processing. In this case when a new contract is discovered, it emits an event, which gets tagged and passed to a global event emitter `serviceEvents`. This is listened to further down to trigger a singular update on the new contract.

packages/api/src/apps/api/app.ts
```diff
@@ -233,17 +249,65 @@ export default async (env: ProcessEnv) => {
 
   console.log("Starting API update loops");
 
+  // listen for new lsp contracts since after we have started api, and make sure they get state updated asap
+  // These events should only be bound after startup, since initialization above takes care of updating all contracts on startup
+  // Because there is now a event driven depdency, the lsp creator and lsp state updater must be in same process
+  serviceEvents.on("multiLspCreator", (data) => {
```
daywiss, line 255: here we listen to the event emitter, and call appropriate update functions to fetch state

packages/api/src/apps/api/app.ts
```diff
@@ -233,17 +249,65 @@ export default async (env: ProcessEnv) => {
 
   console.log("Starting API update loops");
 
+  // listen for new lsp contracts since after we have started api, and make sure they get state updated asap
+  // These events should only be bound after startup, since initialization above takes care of updating all contracts on startup
+  // Because there is now a event driven depdency, the lsp creator and lsp state updater must be in same process
```
mrice32, line 254:
```suggestion
  // Because there is now a event driven dependency, the lsp creator and lsp state updater must be in same process
```

packages/api/src/apps/api/app.ts
mrice32: `multiLspCreator` seems like an odd name for an event. Shouldn't events generally be actions?

packages/api/src/apps/api/app.ts
```diff
@@ -233,17 +249,65 @@ export default async (env: ProcessEnv) => {
...
+  serviceEvents.on("multiLspCreator", (data) => {
+    console.log("LspCreator found a new contract", JSON.stringify(data));
+    services.lsps.updateLsp(data.address, data.startBlock, data.endBlock);
+  });
+
+  // listen for new emp contracts since after we start api, make sure they get state updates asap
+  serviceEvents.on("empRegistry", (data) => {
```
mrice32, line 261: Similar comment here.

packages/api/src/apps/api/app.ts
```diff
@@ -233,17 +249,65 @@ export default async (env: ProcessEnv) => {
...
+    return block.number;
+  }
+
+  const newContractBlockTick = BlockInterval(detectNewContracts, initBlock.number);
+  const updateContractStateTick = BlockInterval(updateContractState, initBlock.number);
+
+  // detect contract loop
+  utils.loop(async () => {
+    const end = profile("Detecting New Contracts");
+    // adding in a timeout rejection if the update takes too long
```
mrice32, line 277: It seems like we've had a few issues with hanging throughout the API application? Do you think there's a root cause for all of these problems? It seems a little strange that we're hanging rather than seeing some sort of error. Are we ignoring errors somewhere? Do we have promises where we only provide a `.then` (with a single arg) and no `.catch` so the promise just remains unresolved forever?

mrice32 (commented):
Looks good! One high-level comments and a few nits.

packages/api/src/apps/api/app.ts
daywiss: I'm not sure of any conventions around this, but definately could be something like "lspFound" or something like that.  I was kind of thinking that events on a single emitter could be pathed, so maybe u could listen to them by service. A future emitter you might listen to something like `on('serviceName',(eventType,data)=>{})`, to differentiate between identical event names on different services. But this is so basic right now that im happy to change it to whatever you think is best

packages/api/src/apps/api/app.ts
daywiss: The symptom im seeing and concluding things are hanging, is that i've added a specific log for when state updates run. After the servers are up for some time, i stop seeing that log, though price updates still run ( they are in a separate loop). I've been pretty careful around making sure we have visibility around errors, so i would expect to see them if there was one.  If there are any spots where we don't catch an error and one is thrown, it would cause the api to crash, or at least log that theres an uncaught promise error, but logs dont show this. 

The one place where funny business can happen though is around the websocket connection to ethers. Last PR updated the websocket options for web3, but ethers under the hood is using an infura websocket connection. If for some reason the websocket gets disconnected, this is the most likely place where we could just hang forever in a promise without throwing an error, we make tons of requests, then get disconnected before all are resolved ( and ethers swallows the error).  

I didnt want to dig into the guts of this, but there are events on the websocket to listen for disconnections. I just wanted to be sure that we are really getting blocked here first.  If thats the case it would warrant a deeper look into the websocket.

mrice32: Are you sure the websocket hangs on disconnect? Whenever we've had a websocket disconnect in the bots, we get a bunch of throws due to contract calls failing. Is there a way we can verify that the websocket is the source of the hanging as opposed to something that's more directly under our control?

packages/api/src/apps/api/app.ts
mrice32: Ahhhh, that makes sense! I think we should either do service name + event type or a global event name (some sort of action). Service name alone is confusing IMO since it doesn't imply _what_ is happening with the service.

packages/api/src/apps/api/app.ts
mrice32: Either way, this architectural change seems good to me! The hanging issue seems to keep coming up, though, so my hope is that we can dig into the source of the issue so we don't keep seeing related issues.

mrice32 (approved):
LGTM! Replied to the threads

packages/api/src/apps/api/app.ts
daywiss: just to be clear here, the type of blocking im talking about for this issue is different from the api server completely locking up (which actually hasn't happened in weeks now).  This issue im seeing is that a particular update loop just stops looping and therefore contracts are not being updated/discovered. 

> Are you sure the websocket hangs on disconnect? Whenever we've had a websocket disconnect in the bots, we get a bunch of throws due to contract calls failing.

I am totally guessing as to what might be causing this, but the possibility remains open that this is a websocket issue, because most processes have been using web3, not ethers, so its possible there is different behavior from what we have seen in the past. 

>  Is there a way we can verify that the websocket is the source of the hanging as opposed to something that's more directly under our control?

Part of this pr is kind of separating out possible problem areas of code to narrow down where the issue is. The culprit is most likely in the emp-state/lsp-state update services, and the major dependency there is ethers and infura. There is more work i can do to dig into the websocket but would like to see first if any loops time out.

packages/api/src/apps/api/app.ts
daywiss: ive added the event type, so the service can specify the type of event that happened.  Currently its just named "created" to signify that a contract was created within the service.

PR #3371, opened 2021-09-16
perf(api): perform EMPs calls in batch

amateima:

**Motivation**

https://app.shortcut.com/uma-project/story/1502/use-multicall-to-prevent-high-infura-usage-for-api-calls-to-contracts

**Summary**

This PR introduces the Multicall2 contract for getting the state of an EMP contract by wrapping the calls into batches instead of using multiple `asyncValues()` calls.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -2,7 +2,7 @@ import assert from "assert";
 import { ethers } from "ethers";
 import moment from "moment";
 
-import { tables, Coingecko, utils, Multicall } from "@uma/sdk";
+import { tables, Coingecko, utils, Multicall, Multicall2 } from "@uma/sdk";
 
 import * as Services from "../../services";
 import Express from "../../services/express-channels";
@@ -17,6 +17,7 @@ export default async (env: ProcessEnv) => {
   assert(env.EXPRESS_PORT, "requires EXPRESS_PORT");
   assert(env.zrxBaseUrl, "requires zrxBaseUrl");
   assert(env.MULTI_CALL_ADDRESS, "requires MULTI_CALL_ADDRESS");
+  assert(env.MULTI_CALL_2_ADDRESS, "requires MULTI_CALL_2_ADDRESS");
   const lspCreatorAddresses = parseEnvArray(env.lspCreatorAddresses || "");
 
   // debug flag for more verbose logs
@@ -109,6 +110,7 @@ export default async (env: ProcessEnv) => {
     longAddresses: new Set<string>(),
     shortAddresses: new Set<string>(),
     multicall: new Multicall(env.MULTI_CALL_ADDRESS, provider),
+    multicall2: new Multicall2(env.MULTI_CALL_2_ADDRESS, provider),
     lsps: {
       active: lsps.JsMap("Active LSP"),
       expired: lsps.JsMap("Expired LSP"),
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -2,7 +2,7 @@ import assert from "assert";
 import { ethers } from "ethers";
 import moment from "moment";
 
-import { tables, Coingecko, utils, Multicall, Multicall2 } from "@uma/sdk";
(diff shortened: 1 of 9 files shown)


packages/api/src/services/emp-state.ts
```diff
@@ -1,95 +1,72 @@
...
+    collateralAddresses,
+    syntheticAddresses,
+    multicall2,
+  } = appState;
   const profile = Profile(config.debug);
+  // default props we want to query on contract
+  const staticProps: [string, (x: any) => any][] = [
+    ["priceIdentifier", parseBytes],
+    ["expirationTimestamp", toNumber],
+    ["withdrawalLiveness", toNumber],
```
amateima, line 35: @daywiss The current emp state attribute is named `withdrawLiveness`, but now the name is changed to the actual name found in the contract: `withdrawalLiveness `. How should we deal with this? Rename it to the old value or update the emp state type with the new name and inform the clients of the API about this change

daywiss: hey thanks for finding this. i would say change it to align with the contract. its a minor change that does not affect anything in use (not used in umaverse)

packages/api/src/apps/api/app.ts
```diff
@@ -17,6 +17,7 @@ export default async (env: ProcessEnv) => {
   assert(env.EXPRESS_PORT, "requires EXPRESS_PORT");
   assert(env.zrxBaseUrl, "requires zrxBaseUrl");
   assert(env.MULTI_CALL_ADDRESS, "requires MULTI_CALL_ADDRESS");
+  assert(env.MULTI_CALL_2_ADDRESS, "requires MULTI_CALL_2_ADDRESS");
```
daywiss, line 20: should we just require a multicall 2 address and use it for both multicall1 and 2? Also i dont remember if multicall 2 is just a drop in replacement for 1, maybe we should just replace the `multicall` reference with the `multicall2` implementation?

packages/api/src/services/emp-state.ts
daywiss: if you want we can make a separate issue for this

daywiss (commented):
this looks great.  were you able to test this and if so did you see any improvements?

packages/api/src/libs/utils.ts
```diff
@@ -131,6 +131,28 @@ export const BatchRead = (multicall: uma.Multicall) => async (
   );
 };
 
+export const BatchReadWithErrors = (multicall2: uma.Multicall2) => async (
+  calls: [string, (x: any) => any][],
```
mrice32, line 135: Annoyingly, typescript doesn't have strong enough inference for us to make this a generic type and have typescript infer the return type for the user :/.

mrice32 (approved):
LGTM! Very nice and clean!

chrismaree:
nit: this PR is not titled in the standard format

packages/api/src/apps/api/app.ts
amateima: we can definitely do that

amateima:
> nit: this PR is not titled in the standard format

`perf` is just a broken down type of `improve`. It's specified in the Conventional Commits standard

amateima:
> this looks great. were you able to test this and if so did you see any improvements?

I ran the api locally. The EMP state for all contracts was updated 24s faster (`2m 0s` vs `1m36s`)

packages/api/src/libs/utils.ts
```diff
@@ -131,6 +131,28 @@ export const BatchRead = (multicall: uma.Multicall) => async (
...
+  // multicall batch takes array of {method} objects
+  const results = await multicall2
+    .batch(
+      contract,
+      calls.map(([method]) => ({ method }))
+    )
+    .readWithErrors();
+  // convert results of multicall, an array of responses, into a key value, keyed by contract method
+  return Object.fromEntries(
+    lodash.zip(calls, results).map(([method, result]) => {
```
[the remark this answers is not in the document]
amateima: I changed the naming 👍

daywiss (approved):

PR #3374, opened 2021-09-17
fix(cryptowatch-pricefeed): Optimistically match price period without buffer first before resorting to buffer

nicholaspai:
**Motivation**

Currently when `historicalTimestampBuffer` is set > 0, the CW pricefeed expands the valid price period through which to look for a historical price. It matches the earliest price period, which potentially returns the incorrect price if the exact historical price is available.

**Summary**

This PR changes the CW pricefeed to first match the earliest price period without accounting for the historical timestamp buffer, then if nothing is found, it uses the buffer to expand the period.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -115,7 +115,6 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
-    // console.log(this.historicalPricePeriods[this.historicalPricePeriods.length-11, this.historicalPricePeriods.length-1])
     const match = this.historicalPricePeriods.find((pricePeriod) => {
       return time <= pricePeriod.closeTime && time >= pricePeriod.openTime;
     });
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -115,59 +115,57 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return (
-        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
-        time >= pricePeriod.openTime - this.historicalTimestampBuffer
-      );
+      return time < pricePeriod.closeTime && time >= pricePeriod.openTime;
     });
 
+    // If match doesn't succeed, then we can still try to find the nearest price and use it if its within the caller's
+    // allowed margin of error.
+    let returnPrice;
     if (match === undefined) {
-      // If match doesn't succeed, then give caller details about closest price periods.
+      // Reverse sort the historical price periods from newest to oldest. Find the first time period whose open time,
+      // minus the buffer, is before the target time. Use this matched period's close time.
       const before = this.historicalPricePeriods
         .slice()
         .reverse()
         .find((pricePeriod) => time >= pricePeriod.openTime - this.historicalTimestampBuffer);
-      const after = this.historicalPricePeriods.find(
-        (pricePeriod) => time < pricePeriod.closeTime + this.historicalTimestampBuffer
-      );
 
-      const format = (value: BN | null | undefined) => {
-        const unformattedValue = value && (this.invertPrice ? this._invertPriceSafely(value) : value);
-        if (!unformattedValue) return "[no value found]";
-        return formatFixed(unformattedValue.toString(), this.priceFeedDecimals);
-      };
-
-      throw new Error(`
-        Cryptowatch price feed ${this.uuid} didn't return an ohlc for time ${time}.
-        Closest price point before is close time: ${before?.closeTime} (- buffer of ${
-        this.historicalTimestampBuffer
-      } seconds), close price: ${format(before?.closePrice)}.
-        Closest price point after is open time: ${after?.openTime} (+ buffer of ${
-        this.historicalTimestampBuffer
-      } seconds), open price: ${format(after?.openPrice)}.
-        To see these prices, make a GET request to:
-        https://api.cryptowat.ch/markets/${this.exchange}/${this.pair}/ohlc?after=${
-        (before?.openTime || 0) + this.ohlcPeriod
-      }&before=${after?.closeTime}&periods=60
-      `);
+      // If the closest price period before the desired timestamp is within the historical timestamp buffer,
+      // then use its close price.
+      if (before && before.closeTime >= time - this.historicalTimestampBuffer) {
+        returnPrice = this.invertPrice ? this._invertPriceSafely(before.closePrice) : before.closePrice;
+        if (!returnPrice) throw new Error(`${this.uuid} -- invalid price returned`);
+        if (verbose) await this._printVerbose(before.closeTime, returnPrice);
+      }
+      // If before is still not within the buffer, then return the user a detailed error message.
+      else {
+        const after = this.historicalPricePeriods.find(
+          (pricePeriod) => time < pricePeriod.closeTime + this.historicalTimestampBuffer
+        );
+        const format = (value: BN | null | undefined) => {
+          const unformattedValue = value && (this.invertPrice ? this._invertPriceSafely(value) : value);
+          if (!unformattedValue) return "[no value found]";
+          return formatFixed(unformattedValue.toString(), this.priceFeedDecimals);
+        };
+        throw new Error(`
+          Cryptowatch price feed ${this.uuid} didn't return an ohlc for time ${time}.
+          Closest price point before is close time: ${before?.closeTime} (- buffer of ${
+          this.historicalTimestampBuffer
+        } seconds), close price: ${format(before?.closePrice)}.
+            Closest price point after is open time: ${after?.openTime} (+ buffer of ${
+          this.historicalTimestampBuffer
+        } seconds), open price: ${format(after?.openPrice)}.
+            To see these prices, make a GET request to:
+            https://api.cryptowat.ch/markets/${this.exchange}/${this.pair}/ohlc?after=${
+          (before?.openTime || 0) + this.ohlcPeriod
+        }&before=${after?.closeTime}&periods=60
+        `);
+      }
+    } else {
+      returnPrice = this.invertPrice ? this._invertPriceSafely(match.openPrice) : match.openPrice;
+      if (!returnPrice) throw new Error(`${this.uuid} -- invalid price returned`);
+      if (verbose) await this._printVerbose(match.openTime, returnPrice);
     }
 
-    const returnPrice = this.invertPrice ? this._invertPriceSafely(match.openPrice) : match.openPrice;
-    if (!returnPrice) throw new Error(`${this.uuid} -- invalid price returned`);
-    if (verbose) {
-      console.group(`\n(${this.exchange}:${this.pair}) Historical OHLC @ ${match.closeTime}`);
-      console.log(`- ✅ Open Price: ${formatFixed(returnPrice.toString(), this.priceFeedDecimals)}`);
-      console.log(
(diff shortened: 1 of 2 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,17 +109,23 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
-        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
-        time >= pricePeriod.openTime - this.historicalTimestampBuffer
-      );
+    // This finds the first pricePeriod whose closeTime is after the provided time. First try to find the match without
+    // resorting to using the historical timestamp buffer, then if there is no match found try using the buffer.
+    let match = this.historicalPricePeriods.find((pricePeriod) => {
+      return time < pricePeriod.closeTime && time >= pricePeriod.openTime;
     });
+    if (match === undefined && this.historicalTimestampBuffer > 0) {
+      match = this.historicalPricePeriods.find((pricePeriod) => {
```
mrice32, line 122: Nice catch.

One oddity remaining is that if this buffer were large, let's say an hour, it would grab the fall back to the _first_ valid timestamp, meaning if the data point didn't exist it would fall back to the one one hour ago rather than the price period just before the missing one.

Does that make sense? And can you think of any easy way to fix this?

mrice32 (commented):
Nice find @nicholaspai @smb2796! One high-level comment.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
Reinis-FRP: In my specific use case fork here https://github.com/Reinis-FRP/protocol/blob/75eb066e5180409964991d02a642ecd65c5e2853/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L132 I have separately searched for nearest timestamp backwards or nearest timestamp forward. Though in my use case I afterwards use only prior close price, but it could be extended to use also forward open price if needed.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +109,63 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
     }
 
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return (
-        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
-        time >= pricePeriod.openTime - this.historicalTimestampBuffer
-      );
+      return time < pricePeriod.closeTime && time >= pricePeriod.openTime;
```
nicholaspai, line 118: First try to match the target time to the first price period who’s open time is before the target time.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +109,63 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
-      console.log(
-        '- This will return an OHLC data packet as "result", which contains in order: \n- [CloseTime, OpenPrice, HighPrice, LowPrice, ClosePrice, Volume, QuoteVolume].'
-      );
-      console.log(
-        "- We use the OpenPrice to compute the median. Note that you might need to invert the prices for certain identifiers like USDETH."
-      );
-      console.groupEnd();
+      // If the closest price period before the desired timestamp is within the historical timestamp buffer,
+      // then use its close price.
+      if (before && before.closeTime >= time - this.historicalTimestampBuffer) {
```
nicholaspai, line 134: If there is no match, then look through the price periods from newest to oldest, and return the first price period whose open time (minus the buffer) is before the target time

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: @Reinis-FRP do you think its a mistake in your fork that you first try to `match` using the buffer in [this line](https://github.com/Reinis-FRP/protocol/blob/75eb066e5180409964991d02a642ecd65c5e2853/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.js#L124) and then if that fails then you try to match `before`? It seems like no matter what if the `time` is within a `pricePeriodTimestamp - historicalTimestampBuffer` that you'll match the earliest price (but not the nearest)?

nicholaspai: Say the price periods have the following timestamps:

```
[
     {openTime: 0, closeTime: 1, price: 1}
     {openTime: 1, closeTime: 2, price: 2}
     {openTime: 2, closeTime: 3, price: null}
     {openTime: 3, closeTime: 4, price: 4}
]
```

And say we're trying to find `getHistoricalTime(time=2)`. Let `historicalTimestampBuffer=1`. In your fork it seems that you'll match the first time period where `price=1` because of this code:

```
      return (
        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
        time >= pricePeriod.openTime - this.historicalTimestampBuffer
      );
    });
```

But the correct price period to match will be 2, right?

Reinis-FRP: I did not change that part - its the same [here](https://github.com/UMAprotocol/protocol/blob/65600aec5f53ae17792b4191a8dd03a03deceba7/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts#L119). It always seemed weird to me why the first timestamp should be taken, but did not fully understand the intended use case of `historicalTimestampBuffer`. So I did not use it and instead created `maxBeforeTimestampBuffer` and if time was not within any open-close period then returned first matching price backwards if it was within `maxBeforeTimestampBuffer`.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +110,67 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
     }
 
     // historicalPricePeriods are ordered from oldest to newest.
     // This finds the first pricePeriod whose closeTime is after the provided time.
     const match = this.historicalPricePeriods.find((pricePeriod) => {
-      return (
-        time < pricePeriod.closeTime + this.historicalTimestampBuffer &&
-        time >= pricePeriod.openTime - this.historicalTimestampBuffer
-      );
+      return time <= pricePeriod.closeTime && time >= pricePeriod.openTime;
```
nicholaspai, line 119: I think this check before was expanding the range with the buffer prematurely. This will return the earliest period that matches the `time` including the buffer, but not neccessarily the latest period right before `time`.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +110,67 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
 
+    // If match doesn't succeed, then we can still try to find the nearest price and use it if its within the caller's
+    // allowed margin of error.
+    let returnPrice;
     if (match === undefined) {
-      // If match doesn't succeed, then give caller details about closest price periods.
-      const before = this.historicalPricePeriods
-        .slice()
-        .reverse()
-        .find((pricePeriod) => time >= pricePeriod.openTime - this.historicalTimestampBuffer);
```
nicholaspai, line 129: I think this check was implemented incorrectly. To find the price period before `time`, we need to expand the period's `closeTime` not its `openTime`. To find the price period after `time` we should expand the `openTime`.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +110,72 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
+      // Accounting for the buffer, find the price periods before and after the target time. These will be undefined if
+      // not found. Reverse the search for the period so that we match the latest period that encompasses the target
+      // time.
       const before = this.historicalPricePeriods
         .slice()
         .reverse()
-        .find((pricePeriod) => time >= pricePeriod.openTime - this.historicalTimestampBuffer);
+        .find(
+          (pricePeriod) =>
+            time <= pricePeriod.closeTime + this.historicalTimestampBuffer && time >= pricePeriod.openTime
```
mrice32, line 134: So this means the result time must _strictly_ be before the requested time, right?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -109,65 +110,72 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
-        `- ⚠️  If you want to manually verify the specific exchange prices, you can make a GET request to: \n- https://api.cryptowat.ch/markets/${this.exchange}/${this.pair}/ohlc?after=${match.closeTime}&before=${match.closeTime}&periods=60`
-      );
-      console.log(
-        '- This will return an OHLC data packet as "result", which contains in order: \n- [CloseTime, OpenPrice, HighPrice, LowPrice, ClosePrice, Volume, QuoteVolume].'
-      );
-      console.log(
-        "- We use the OpenPrice to compute the median. Note that you might need to invert the prices for certain identifiers like USDETH."
-      );
-      console.groupEnd();
+      // If there is a `before` period, then use its close time, otherwise use the `after` period's open time.
```
mrice32, line 157: Can you explain what changed in these logs/checks?

mrice32 (approved):
LGTM!! Thanks for doing this @nicholaspai!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: Yes exactly, if its not `before`, then it could match as the `after` time. Do you think this is a problem?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: The `if` statement logic is: 
- If there is a `before` period, then use its `close` price, otherwise if there is an `after` period, use its `open` price. If none, throw.

nicholaspai:
Requesting @chrismaree for another set of eyes since this PR affects so many production apps.

mrice32 (approved):
LGTM! Thanks for the explanation!

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
chrismaree: ye, I agree beforehand it looked wrong. I don't think it would have caused any major issues given how the previous equalities were set up but you change here is _more_ correct. this is pretty subtle!

packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js
```diff
@@ -184,16 +194,122 @@ describe("CryptoWatchPriceFeed.js", function () {
...
+          ],
+        },
+      },
+      { result: { price: 1.5 } },
+    ];
 
     await cryptoWatchPriceFeed.update();
 
-    // Before period 1 should succeed when accounting for historical timestamp buffer.
+    // An input timestamp more than the buffer length before the first period should throw an error.
```
chrismaree, line 225: thank you for adding these detailed comments. They really helped with understanding how this was wrong and in following your solution.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -231,6 +239,26 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
     this.lastUpdateTime = currentTime;
   }
 
+  private async _printVerbose(pricePeriod: PricePeriod, returnPrice: BN) {
```
chrismaree, line 242: good refactor

chrismaree (approved):
This is excellent. the bug before we extremely subtle. you did a great job at explaining the problem & your solution via comments.

PR #3376, opened 2021-09-20
improve(sdk)!: use dependency injection for using stores inside tables

amateima:
**Motivation**

https://app.shortcut.com/uma-project/story/2163/refactor-table-api-to-require-user-to-configure-and-inject-stores

**Summary**

We want the store to be manually injected into the tables, so tables won't be responsible for creating the store instance. Any type of store is expected to work fine: Google Datastore or in-memory JsMap, practically anything that implements [these](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/stores/index.ts#L6) interfaces

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/sdk/src/tables/blocks/js-map.ts
```diff
@@ -1,9 +1,10 @@
 import { Data, makeId } from "./utils";
```
daywiss, line 1: can we rename the files in each table type to be just `table.ts` from `js-map`

packages/sdk/src/tables/blocks/js-map.ts
```diff
@@ -1,9 +1,10 @@
...
-import { JsMap } from "../generic";
+import JsMap from "../base";
```
daywiss, line 2: maybe we should change this to `import Base from "../base"`

packages/sdk/src/tables/blocks/js-map.ts
```diff
@@ -1,9 +1,10 @@
...
+import { Store } from "../../stores";
```
daywiss, line 3: `import type` to be more specific

packages/sdk/src/tables/emps/table.ts
```diff
@@ -1,8 +1,10 @@
...
+import type { Store } from "../../stores";
+import { JsMap } from "../../stores";
 
-export const Table = (type = "Emp") => {
-  const table = JsMap<string, Data>(type, makeId);
+export const Table = (type = "Emp", store: Store<string, Data> = JsMap()) => {
+  const table = BaseTable<string, Data, Store<string, Data>>({ type, makeId }, store);
 
   async function addSponsors(id: string, sponsors: string[]) {
     const data = await table.get(id);
```
amateima, line 10: 2. here is the **issue**: `get()` returns `Promise<Data | undefined>`, so the type is inherited from the store object spread [here](https://github.com/UMAprotocol/protocol/pull/3376/files#diff-e3d44be706c5c9db01b1a83dffaac767e06f1e3c74737c1fc6003d1c78ffea43R31), not the `get()` function that overwrites it
And this happens in multiple cases, this is just an example.

I think I can avoid this, but I want to understand why it happens. It's related to the use of Generics I think

packages/sdk/src/tables/base/table.ts
```diff
@@ -2,7 +2,10 @@ import assert from "assert";
 import { exists } from "../../utils";
 import type { stores, MakeId, MaybeId, HasId } from "../..";
 
-export default function Table<I, D>(config: { makeId: MakeId<I, D>; type: string }, store: stores.Store<I, D>) {
+export default function Table<I, D, S extends stores.Store<I, D>>(
```
amateima, line 5: `S extends stores.Store<I, D>`
I use Generics for the store because I want the returned type to specify only the supported methods by the Store. 
`Store vs SortedStore`: `SortedStore` has additional functions: `slice()` and `between()`

amateima:
@mrice32 @daywiss I'm stuck with the issue described above. Maybe a fresh pair of eyes sees something that I'm missing here

packages/sdk/src/tables/base/table.ts
daywiss: i considered this previously, but maybe we can just have a single store interface where a non-sorted store either throws an error or returns an empty array for unsupported calls

packages/sdk/src/tables/base/table.ts
```diff
@@ -34,3 +38,5 @@ export default function Table<I, D>(config: { makeId: MakeId<I, D>; type: string
     type,
   };
 }
+
+export type Table = ReturnType<typeof Table>;
```
daywiss, line 42: i wonder if this matters:

`export type Table = ReturnType<typeof Table<I, D, S extends stores.Store<I, D>>;`

mrice32:
@amateima @daywiss I've encountered similar problems before I think. I'm a little hazy on _exactly_ why typescript inference falls back to the store, but the basic issue is (as David mentioned) combining these lines:
```ts
export type Table = ReturnType<typeof Table>;
```
with generics.

What's happening is that `typeof` drops the generic. The first thing I've tried to fix this is to do what @daywiss suggests [here](https://github.com/UMAprotocol/protocol/pull/3376#discussion_r715088847), but the issue is that typeof _cannot_ produce a generic type. So the only thing it can do when passed a generic function that is a value with a generic, is to drop all parts of that type (I'm a little hazy on this part) that it can't infer without knowing the generic args. As a side note: this is where classes have a big advantage. Typescript creates the type for you, so if they are generic, typescript generates the _generic_ type without needing to use `typeof` and dropping the generic args.

Anyway, there is a way around this. We just need to modify the type we export.

I created a new branch (since I didn't want to cause any issues on your branch @amateima). Here's the commit that fixes the problem you pointed out [here](https://github.com/UMAprotocol/protocol/pull/3376#discussion_r714829889):

https://github.com/mrice32/protocol/commit/1f5c8d70c801eea065df0a1fd94cc353dbb0e04f.

Basically, we just have to create our own generic types manually to workaround the limitations of the inference, which is annoying and verbose, but it seems to work. Not sure if this is what you had in mind for a workaround @amateima.

mrice32 (commented):
See https://github.com/UMAprotocol/protocol/pull/3376#issuecomment-926211110.

packages/api/src/apps/api/app.ts
```diff
@@ -45,6 +46,11 @@ export default async (env: ProcessEnv) => {
   // services can emit events when necessary, though for now any services that depend on events must be in same process
   const serviceEvents = new Events();
 
+  const empsActiveStore = stores.JsMap<string, tables.emps.Data>();
+  const empsExpiredStore = stores.JsMap<string, tables.emps.Data>();
+  const empsActiveTable = tables.emps.JsMap("Active Emp", empsActiveStore);
+  const empsExpiredTable = tables.emps.JsMap("Expired Emp", empsExpiredStore);
```
daywiss, line 52: can we change the table exports to not specify the store type. Might want to just consider exposing them like this

`tables.emps.Table("Expired Emp", empsExpiredStore)`  or `tables.Emps("Expired Emp", empsExpiredStore)` .

First option gives some better flexibility export things other than table, but the first one is a bit cleaner if we only expect to use table.  wdyt?

Also you may not explicitly need to define stores, maybe you did it here for testing, but since you have defaults you can probably just instance the tables, like with blocks/erc20s.

packages/api/src/apps/api/app.ts
```diff
@@ -53,8 +59,8 @@ export default async (env: ProcessEnv) => {
     zrx: new Zrx(env.zrxBaseUrl),
     blocks: tables.blocks.JsMap(),
```
daywiss, line 60: blocks and erc20 tables should have the jsmap part removed or changed

daywiss (commented):
Thanks for working on this, it ended up being quite challenging. 

If you decide to change the path of the tables in the sdk, this commit should probably include a `!` or `BREAKING CHANGE` in the pr title.

Could you also take a look at the readmes for the tables and make sure they match this new design

mrice32 (approved):
LGTM!

daywiss (approved):
looks good. id just suggested editing the github title to include the `!` or add it when you merge so the sdk version gets bumped on release

PR #3377, opened 2021-09-21
improve(disputer): remove truffle

mrice32:
**Motivation**

Remove truffle from the disputer package.

**Summary**

Removed all truffle dependencies.

This required a small change to contracts-node to get it to recognize the hardhatTestingAddresses loaded into the global object.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/common/src/hardhat/tasks/artifacts.ts
+++ b/packages/common/src/hardhat/tasks/artifacts.ts
@@ -231,9 +231,9 @@ interface HRE {
 }
 export async function getAddress(name: DeploymentName | ContractName, chainId: number): Promise<string> {
   if (typeof chainId !== "number") throw new Error("chainId must be a number");
-  const hre = (global as unknown as { hre?: HRE }).hre;
-  const hreDeployment = hre && parseInt(await hre.getChainId()) === chainId && await hre.deployments.getOrNull(name);
-  if (hreDeployment) return hreDeployment.address;
+  const { hre, hardhatTestingAddresses } = (global as unknown as { hre?: HRE; hardhatTestingAddresses?: { [name: string]: string } });
+  const hreAddress = hre && parseInt(await hre.getChainId()) === chainId && ((await hre.deployments.getOrNull(name))?.address || hardhatTestingAddresses?.[name]);
+  if (hreAddress) return hreAddress;
   if (!isDeploymentName(name)) throw new Error(\`No deployments for name: \${name}\`);
   const fn = addressFunctions[name];
   return fn(chainId);
--- a/packages/disputer/.mocharc.js
+++ b/packages/disputer/.mocharc.js
@@ -1,0 +1,3 @@
+module.exports = {
+  timeout: 100000,
+};
--- a/packages/disputer/.mocharc.js
+++ b/packages/disputer/.mocharc.js
@@ -1,3 +1,4 @@
 module.exports = {
   timeout: 100000,
+  exit: true,
 };
--- a/packages/disputer/package.json
+++ b/packages/disputer/package.json
@@ -6,6 +6,8 @@
     "@uma/common": "^2.6.0",
     "@uma/core": "^2.7.0",
     "@uma/financial-templates-lib": "^2.6.1",
+    "@uma/contracts-node": "^0.1.0",
+    "@uma/contracts-node-0-1-0": "npm:@uma/contracts-node@0.1.0",
     "async-retry": "^1.3.1",
     "dotenv": "^6.2.0"
   },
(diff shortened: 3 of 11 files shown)


packages/common/src/hardhat/tasks/artifacts.ts
```diff
@@ -231,9 +231,9 @@ interface HRE {
 }
 export async function getAddress(name: DeploymentName | ContractName, chainId: number): Promise<string> {
   if (typeof chainId !== "number") throw new Error("chainId must be a number");
-  const hre = (global as unknown as { hre?: HRE }).hre;
-  const hreDeployment = hre && parseInt(await hre.getChainId()) === chainId && await hre.deployments.getOrNull(name);
-  if (hreDeployment) return hreDeployment.address;
+  const { hre, hardhatTestingAddresses } = (global as unknown as { hre?: HRE; hardhatTestingAddresses?: { [name: string]: string } });
+  const hreAddress = hre && parseInt(await hre.getChainId()) === chainId && ((await hre.deployments.getOrNull(name))?.address || hardhatTestingAddresses?.[name]);
```
nicholaspai, line 235: Can you add a comment explaining all the different ways that `hreAddress` can be set? This line is very complex

packages/disputer/.mocharc.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  timeout: 100000,
```
nicholaspai, line 2: why?

packages/disputer/package.json
```diff
@@ -6,6 +6,8 @@
     "@uma/common": "^2.6.0",
     "@uma/core": "^2.7.0",
     "@uma/financial-templates-lib": "^2.6.1",
+    "@uma/contracts-node": "^0.1.0",
+    "@uma/contracts-node-0-1-0": "npm:@uma/contracts-node@0.1.0",
```
nicholaspai, line 10: why do we need this import?

packages/monitors/package.json
```diff
@@ -9,12 +9,12 @@
     "@uma/common": "^2.6.0",
     "@uma/core": "^2.7.0",
     "@uma/contracts-node": "^0.1.0",
+    "@uma/contracts-node-0-1-0": "npm:@uma/contracts-node@0.1.0",
```
nicholaspai, line 12: Same question as above why do we need this import

packages/disputer/.mocharc.js
chrismaree: default mocha time is too slow for some of the our tests.

packages/disputer/src/proxyTransactionWrapper.js
```diff
@@ -141,14 +144,13 @@ class ProxyTransactionWrapper {
     const callCode = this.ReserveCurrencyDisputer.bytecode;
 
     const dsProxyCallReturn = await this.dsProxyManager.callFunctionOnNewlyDeployedLibrary(callCode, callData);
-    const blockAfterDispute = await this.web3.eth.getBlockNumber();
 
     // Wait exactly one block to fetch events. This ensures that the events have been indexed by your node.
-    await blockUntilBlockMined(this.web3, blockAfterDispute + 1);
```
chrismaree, line 147: nice! this removes a call as well.

packages/disputer/src/proxyTransactionWrapper.js
```diff
@@ -141,14 +144,13 @@ class ProxyTransactionWrapper {
...
+    await blockUntilBlockMined(this.web3, dsProxyCallReturn.blockNumber + 1);
 
     const DisputeEvent = (
       await this.financialContract.getPastEvents("LiquidationDisputed", {
-        fromBlock: blockAfterDispute - 1,
+        fromBlock: dsProxyCallReturn.blockNumber,
```
chrismaree, line 153: ye, I guess this is also sufficient .

chrismaree (approved):
Looks pretty good! painful work but we almost there! once this is merged & the liquidator PR we are done 🎉

packages/disputer/.mocharc.js
mrice32: Yeah, agreed. We could probably set this to something lower than 100s, but we set it to 100s elsewhere, and it shouldn't change performance unless we have a timeout.

packages/disputer/package.json
mrice32: When we dynamically look up the abi/bytecode of specific versions, our method is to create an alias that has that version number in its package name. So 2.0.1 maps to contracts-node@0.1.0, which we alias to contracts-node-0-1-0. 

See this function for details: https://github.com/UMAprotocol/protocol/blob/06094f2b2c9defbcc332da4de64e2bf1a6ed91fa/packages/common/src/MultiVersionTestHelpers.ts#L25-L34.

PR #3378, opened 2021-09-21
feat(liquidator): remove truffle

chrismaree:
**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


nicholaspai (approved):

packages/liquidator/index.js
```diff
@@ -105,10 +111,8 @@ async function run({
       );
 
     // Setup contract instances. This uses the contract version pulled in from previous step.
-    const financialContract = new web3.eth.Contract(
-      getAbi(liquidatorConfig.contractType, liquidatorConfig.contractVersion),
-      financialContractAddress
-    );
+    const { getAbi } = require(getContractsNodePackageAliasForVerion(liquidatorConfig.contractVersion));
```
mrice32, line 114: You might want to assign this to a different name and import the normal getAbi from `@uma/contracts-node` the top. The reason I suggest this is that this causes us to unintentionally use the _versioned_ `getAbi` for all the contracts that come up later in this function. Since the old code used the _current_ abis for these other contracts, it might make sense to keep that consistent.

Does that make sense?

I did something similar in the disputer package.

packages/liquidator/package.json
```diff
@@ -31,7 +31,7 @@
   ],
   "bin": "index.js",
   "scripts": {
-    "test": "yarn mocha-test && yarn hardhat-test",
+    "test": "yarn mocha-test && mocha 'test/**/*.js'",
```
mrice32, line 34: If we're using mocha for _all_ the tests, can we combine them into a single call to mocha?

mrice32 (approved):
LGTM! Two minor comments

packages/liquidator/index.js
chrismaree: Hhhmm. did you do that in the disputer? my implementation looks identical to what you put in the disputer package. the disputer has:
```ts
// Setup contract instances.
    const { getAbi } = require(getContractsNodePackageAliasForVerion(disputerConfig.contractVersion));
    const financialContract = new web3.eth.Contract(getAbi(disputerConfig.contractType), financialContractAddress);
```

packages/liquidator/package.json
chrismaree: good point! done.

packages/liquidator/index.js
mrice32: Ha, odd, I must've not committed the change. I will change it in a follow-up :).

mrice32 (approved):
Just need to resolve this thread: https://github.com/UMAprotocol/protocol/pull/3378#discussion_r714040792.

packages/liquidator/index.js
chrismaree: yup, I agree with you to avoid any confusion or unintentionally calling it. what shall we call this? `getVersionedAbi` or what do you think?

chrismaree: after discussing this offline we will leave this as it is (for now) and then update both this package and the disputer in the same PR later on.

PR #3381, opened 2021-09-21
feat(contracts-node): export similar functions in contracts-node as contracts-frontend

mrice32:
**Motivation**

To allow backend packages (like the SDK) to gracefully switch between contracts-node and contracts-frontend depending on the env, we would like the functions in contracts-frontend to be available in contracts-node.

**Summary**

- `get[ContractName]{Address, Bytecode, Abi}` functional forms should all be available in contracts-node.
- `get[ContractName]Address` has a minor difference between the two packages -- it returns a `Promise<string>` in contracts-node to allow it to work gracefully in hardhat tests. Rather than making things more annoying on the FE packages, I figure the packages that want to incorporate both can either manually handle both cases with a wrapper or just put an await in front of the call, which will work regardless of whether a value or Promise is returned. 

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/common/src/hardhat/tasks/artifacts.ts
+++ b/packages/common/src/hardhat/tasks/artifacts.ts
@@ -246,10 +246,7 @@ export async function getAddress(name: DeploymentName | ContractName, chainId: n
 `
     );
 
-    // Write abi and bytecode for the browser file.
-    // Note: the idea behind writing the functions this way is to make them as optimized as possible for tree-shaking
-    // to remove any unused json files. In modern versions of webpack, this should allow absolutely _no_ artifact
-    // information that isn't needed to be pulled in.
+    // Write abi, bytecode, and address functions to match the contracts-frontend package api.
     artifacts.forEach(({ contractName }) => {
       fs.appendFileSync(
         out,
--- a/packages/common/src/hardhat/tasks/artifacts.ts
+++ b/packages/common/src/hardhat/tasks/artifacts.ts
@@ -203,7 +203,7 @@ export type { ${normalizeClassName(contractName)}Web3Events };\n`
     // Creates get[name]Address(chainId) using switch statements.
     // Note: don't export these functions as they are only used internally.
     for (const [name, addressesByChain] of Object.entries(addresses)) {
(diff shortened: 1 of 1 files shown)


packages/common/src/hardhat/tasks/artifacts.ts
```diff
@@ -245,6 +245,31 @@ export async function getAddress(name: DeploymentName | ContractName, chainId: n
 }
 `
     );
+
+    // Write abi and bytecode for the browser file.
```
daywiss, line 249: might want to update this comment to reference matching the frontend api

packages/common/src/hardhat/tasks/artifacts.ts
```diff
@@ -213,7 +213,7 @@ export type { ${normalizeClassName(contractName)}Web3Events };\n`
 
     // Constructs a mapping of name to address function for nodejs.
     fs.appendFileSync(out, "const addressFunctions = {\n");
-    Object.keys(addresses).forEach((name) => fs.appendFileSync(out, `  ${name}: get${name}Address,\n`));
+    Object.keys(addresses).forEach((name) => fs.appendFileSync(out, `  ${name}: get${name}StaticAddress,\n`));
```
daywiss, line 216: is this to match the fe api?

daywiss (approved):
glad to see this was a fairly simple change

packages/common/src/hardhat/tasks/artifacts.ts
mrice32: We had internal functions called the same thing that are used for something else. So I renamed those functions to make room for the new ones.

These functions differ in that they map only the static addresses, not the dynamic ones (i.e. hardhat testing addresses).

packages/common/src/hardhat/tasks/artifacts.ts
mrice32: Yep, will fix!

daywiss (approved):

PR #3384, opened 2021-09-21
feat(common): HIGH-RISK remove dependence on truffle config for getWeb3

mrice32:
**Motivation**

We'd like to drop truffle related components from the repo.

**Summary**

To drop the truffle config from common, we need all packages to no longer require it.

I wanted to separate out this sensitive change (i.e. could break bots!) to getWeb3() as a precursor to a larger refactor where all truffle-related components are dropped.

My suggested path forward is to carefully review this, commit it and verify that once deployed, all types of bots run correctly.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/serverless-orchestration/test/ServerlessSpoke.js
+++ b/packages/serverless-orchestration/test/ServerlessSpoke.js
@@ -186,26 +186,6 @@ describe("ServerlessSpoke.js", function () {
     ); // Check the process logger contained the error.
     assert.isTrue(lastSpyLogIncludes(spy, "Process exited with error")); // Check the process logger contains exit error.
   });
-  it("Serverless Spoke can correctly returns errors over http calls(invalid network identifier)", async function () {
-    // Invalid price feed config should error out before entering main while loop
-    const invalidPriceFeed = {
-      serverlessCommand: "yarn --silent monitors --network INVALID",
-      environmentVariables: {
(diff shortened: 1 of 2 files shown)


chrismaree (approved):
this seems like a pretty reasonable way to remove this dependency. I agree the safest way to merge is to roll out and watch the bots to ensure they behave as expected.

nicholaspai (approved):
Yeah this looks like a great simplification. Very powerful line of code you're removing relative to raw amount of characters haha

packages/serverless-orchestration/test/ServerlessSpoke.js
```diff
@@ -186,26 +186,6 @@ describe("ServerlessSpoke.js", function () {
     ); // Check the process logger contained the error.
     assert.isTrue(lastSpyLogIncludes(spy, "Process exited with error")); // Check the process logger contains exit error.
   });
-  it("Serverless Spoke can correctly returns errors over http calls(invalid network identifier)", async function () {
```
mrice32, line 189: As a result of this change, the getWeb3() call became more permissive. The network name always falls back to localhost. It's no longer required to be in some standard list of allowed networks (since we don't want to rely on the truffle config anymore).

PR #3385, opened 2021-09-22
feat(insured-bridge): add arbitrum contract support to deposit box inheritance path

chrismaree:
**Motivation**

The across bridge requires to work on arbitrum. This PR makes initial steps towards setting up this support.

**Summary**

This PR refactors the `OVM_BridgeDepositBox.sol` into a two-layered inheritance path system: `BridgeDepositBox`, which contains common bridge logic on all L2 chains and an `OVM_BridgeDepositBox` which inherits from `BridgeDepositBox` and adds Optimism specific logic such as access controls and token bridging.

This pattern lets us define an `AVM_BridgeDepositBox.sol` that inherits from `BridgeDepositBox` and defines the arbitrum specific access controls and bridging without needing to re-define any of the underlying bridge pool logic.

**Testing**

Check a box to describe how you tested these changes and list the steps for reviewers to test.

- [ ]  Ran end-to-end test, running the code as in production
- [ ]  New unit tests created
- [ ]  Existing tests adequate, no new tests required
- [X]  All existing tests pass
- [ ]  Untested

PR #3387, opened 2021-09-22
feat(insured-bridge): Refactor BridgeAdmin into shared logic contract across rollups

nicholaspai:
**Motivation**

Very similar to #3385 in creating shared logic for insured-bridge L1 contracts that can be re-used for different rollup VMs. 

**Summary**

This PR touches a lot of files because it removes any assumptions, or clarifies them with comments, about using Optimism vs Arbitrum vs other rollups.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/hardhat/tasks/artifacts.ts
```diff
@@ -188,7 +188,7 @@ export type { ${normalizeClassName(contractName)}Web3Events };\n`
       fs.appendFileSync(out, `  ${contractName}: "${relativePath}",\n`)
     );
     fs.appendFileSync(out, "};\n");
-    fs.appendFileSync(out, "type ContractName = keyof typeof artifactPaths;\n");
+    fs.appendFileSync(out, "export type ContractName = keyof typeof artifactPaths;\n");
```
nicholaspai, line 191: I thought I needed this type in `InsuredBridgeL1Client` but I realized that instead of making the user specify the type of `BridgeAdmin` (OVM vs. AVM), we could just use the `BridgeAdminInterface` if we made a small tweak to the interface. More on this contract change below.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
```diff
@@ -1,253 +0,0 @@
-// SPDX-License-Identifier: AGPL-3.0-only
```
nicholaspai, line 1: All shared logic in this file moved to `BridgeAdminBase`

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdminBase.sol
```diff
@@ -0,0 +1,152 @@
...
+import "../../../contracts/oracle/interfaces/IdentifierWhitelistInterface.sol";
+import "../../../contracts/oracle/interfaces/FinderInterface.sol";
+import "../../../contracts/oracle/implementation/Constants.sol";
+import "../../../contracts/common/interfaces/AddressWhitelistInterface.sol";
+import "../../../contracts/common/implementation/Lockable.sol";
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @notice Shared administrative logic among L1 BridgeAdmin contracts that have implicit references to L2 DepositBoxes.
```
nicholaspai, line 15: Only change is top-level comments about contract

packages/core/contracts-ovm/insured-bridge/implementation/ovm/OptimismBridgeAdmin.sol
```diff
@@ -0,0 +1,124 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
nicholaspai, line 1: This file only copies code from what was previously `BridgeAdmin`

packages/core/contracts-ovm/insured-bridge/interfaces/BridgeAdminInterface.sol
```diff
@@ -14,6 +24,8 @@ interface BridgeAdminInterface {
         address bridgePool;
     }
 
+    function whitelistedTokens(address) external view returns (L1TokenRelationships memory);
```
nicholaspai, line 27: Adding this function + the events above enables the off-chain clients to use the `BridgeAdminInterface` instead of having to specify `OptimismBridgeAdmin` vs `ArbitrumBridgeAdmin`.

PR #3389, opened 2021-09-23
feat(insured-bridge): Add "ownable" bridge deposit box to simplify testing

chrismaree:
**Motivation**

As part of testing, we often require a setup that is not dependent on one L1/L2 implementation. Rather than mimicking every possible combination, it would be nice to have a bridge deposit box designed for this kind of environment. Think front end testing. To this effect, this PR adds a `Ownable_BridgeDepositBox` which is meant to be deployed on the same provider as the other across contracts to aid in testing.

**Summary**

This PR also moved the `StandardBridgeLike` interface into the `OVM_BridgeDepositBox` file as this is where it is used.

**Testing**

- [ ]  New unit tests created
- [X]  Untested


nicholaspai (approved):
This makes sense, do you think we need to make a similar generic "testing" implementation for the BridgeAdmin's MessengerInterface?

PR #3390, opened 2021-09-23
feat(insured-bridge): BridgeAdmin supports multiple L2s

nicholaspai:
**Motivation**

BridgeAdmin currently hardcodes how it relays messages to L2, and that implementation assumes that the L2 is Optimism. This PR refactors the relay implementation to an external `MessengerContractInterface` that carries out the relay, thereby allowing one `BridgeAdmin` on L1 to send admin messages to any number of L2 `BridgeDepositBoxes`.

**Summary**

Added comments in off-chain client tests (e.g. `InsuredBridgeL1Client`, `InsuredBridgeL2Client`, `InsuredBridgePriceFeed`, etc.) that we use the Optimism contracts for testing purposes. These clients are not opinionated about which L2 they are connected to.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
```diff
@@ -126,87 +123,110 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, OVM_CrossDomainEnabled, L
...
-        emit WhitelistToken(_l1Token, _l2Token, _bridgePool);
+    /**************************************
+     *        VIEW FUNCTIONS              *
+     **************************************/
+    function depositContracts(uint256 _chainId) external view override returns (DepositUtilityContracts memory) {
+        return _depositContracts[_chainId];
+    }
+
+    function whitelistedTokens(address _l1Token) external view override returns (L1TokenRelationships memory) {
+        return _whitelistedTokens[_l1Token];
```
chrismaree, line 229: why not just make the underlying data type public? is there a gas tradeoff with making a function like this rather? I guess it does let you `override` the interface but having a public variable would do the same thing, no?

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
```diff
@@ -0,0 +1,19 @@
...
+
+import "../external/OVM_CrossDomainEnabled.sol";
+
+/**
+ * @notice Sends cross chain messages Optimism L2 network.
+ */
+contract OptimismMessenger is OVM_CrossDomainEnabled {
+    constructor(address _crossDomainMessenger) OVM_CrossDomainEnabled(_crossDomainMessenger) {}
+
+    function relayMessage(
```
chrismaree, line 12: nit: natspec

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrismaree, line 1: this is super clean, I really like this pattern.

I dont think that the Arbitrum one will be as clean though...

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
```diff
@@ -0,0 +1,19 @@
...
+pragma solidity ^0.8.0;
+
+import "../external/OVM_CrossDomainEnabled.sol";
+
+/**
+ * @notice Sends cross chain messages Optimism L2 network.
+ */
+contract OptimismMessenger is OVM_CrossDomainEnabled {
```
chrismaree, line 9: nit: should this not also be an instance of `MessengerInterface`

packages/core/test/insured-bridge/BridgeAdmin.js
```diff
@@ -395,8 +472,12 @@ describe("BridgeAdmin", () => {
...
-          await bridgeAdmin.methods.setEnableDeposits(l2Token, false, customGasLimit).send({ from: owner });
+          await bridgeAdmin.methods
+            .setDepositContract(chainIds.optimism, depositBoxImpersonator, optimismMessenger.options.address)
+            .send({ from: owner });
+          await bridgeAdmin.methods
+            .setEnableDeposits(chainIds.optimism, l2Token, false, customGasLimit)
+            .send({ from: owner });
           assert.equal(
             l1CrossDomainMessengerMock.smocked.sendMessage.calls[0]._gasLimit,
             customGasLimit,
```
chrismaree, line 483: do you think it would be worthwhile splitting up these tests into separate files?

it might be nice to have a `BridgeAdmin`, an `Optimism_BridgeAdmin` and `Arbitrum_BridgeAdmin`? otherwise this file will get super long.

also fine to leave this refactor for later.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
```diff
@@ -23,13 +23,16 @@ const {
   PROXY__OVM_L1_CROSS_DOMAIN_MESSENGER,
 } = require("./helpers/OptimismConstants");
 
+const chainId = "10";
```
chrismaree, line 26: did you actually run these tests out of interest?

chrismaree: Also, I cant wait until the optimism compiler is updated. We can ditch a lot of the painful docker stuff.

we'll also want to have different integration tests for each L2 to ensure sufficient coverage.

packages/financial-templates-lib/test/clients/InsuredBridgeL2Client.js
```diff
@@ -109,7 +112,7 @@ describe("InsuredBridgeL2Client", () => {
 
     let expectedDeposits = [
       {
-        chainId: 10,
+        chainId: chainId,
```
chrismaree, line 115:
```suggestion
        chainId,
```

packages/financial-templates-lib/test/clients/InsuredBridgeL2Client.js
```diff
@@ -142,7 +145,7 @@ describe("InsuredBridgeL2Client", () => {
 
     await client.update();
     expectedDeposits.push({
-      chainId: 10,
+      chainId: chainId,
```
chrismaree, line 148:
```suggestion
      chainId,
```

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
```diff
@@ -13,6 +13,10 @@ const { deployOptimismContractMock } = require("../../../core/test/insured-bridg
 const winston = require("winston");
 const { assert } = require("chai");
 
+// Note: This bot should not be opinionated about whether it is connected to Optimism or Arbitrum, so we'll default to
```
chrismaree, line 16: yup! Ideally each is tested. can add more tests later.

chrismaree (approved):
This is excellent work! thanks for re-factoring after the previous PR. I think this is a great pattern.

Feel free to merge once you've integrated nits and replied to questions.

packages/core/contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol
nicholaspai: Yeah I couldn't get this to work when using `override` on the `public` mappings. Here's the compile error:

```
TypeError: Overriding public state variable return types differ.
  --> contracts-ovm/insured-bridge/implementation/BridgeAdmin.sol:33:5:
   |
33 |     mapping(address => L1TokenRelationships) public override whitelistedTokens;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
Note: Overridden public state variable is here:
  --> contracts-ovm/insured-bridge/interfaces/BridgeAdminInterface.sol:35:5:
   |
35 |     function whitelistedTokens(address) external view returns (L1TokenRelationships memory);
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

@mrice32 I recall you might have a solution for this? I think part of the issue is that the interface expects a `struct memory` to be returned, but mappings don't return `memory` types..?

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
nicholaspai: Arbitrum one will require additional params like `gasPrice`, so I was thinking that either we create a `MessageParams` struct or just add more params to `relayMessage` that we can throwaway in the Optimism case. Luckily, the optimism params are a strict subset of the arbitrum ones...

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
nicholaspai: Great catch thanks

packages/core/test/insured-bridge/BridgeAdmin.js
nicholaspai: Yeah great point I was being lazy. I built a `MessengerMock` that allows us to split up the `BridgeAdmin` tests from the `OptimismMessenger` logic and used it across all off-chain client testing as well.

packages/core/test/insured-bridge/e2e/InsuredBridge.e2e.js
nicholaspai: No I haven't, I'm not sure how to actually run these tests

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
nicholaspai: I addressed this comment and removed all assumptions about which network we're testing

PR #3393, opened 2021-09-24
feat(sdk): add frontend build process to sdk

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/2271/prepare-sdk-for-use-on-a-fe

**Summary**

This uses rollup-replace to go through and replace instances of `@uma/contracts-node` with `@uma/contracts-frontend` when building the frontend version. There ends up being 2 versions of the sdk in `dist/web` and `dist/node`.  Importing remains the same, so this seems to still run correctly on both api and a test frontend app without any changes.

This adds 2 new build commands
- `yarn build:web` - which builds to the frontend
- `yarn build:node` - which builds to node
- `yarn build` - which has been modified to run a script `build.js` so both builds are moved to independent folders

**Details**

Shows sdk size in frontend
image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2271/prepare-sdk-for-use-on-a-fe

### DIFF
--- a/packages/sdk/build.js
+++ b/packages/sdk/build.js
@@ -1,0 +1,77 @@
+// this file is adapted from https://blog.unterholzer.dev/cross-compatible-typescript-libraries/
+// It basically moves files around so that we can have a separate web and node sdk version in different directories
+const proc = require("child_process");
+const fs = require("fs");
+const path = require("path");
+
+const moveFile = (from, to, filename) => {
+  from = from.split("/");
+  to = to.split("/");
+
+  if (filename) {
+    from.push(filename);
+    to.push(filename);
+  }
+
+  fs.renameSync(path.resolve(...from), path.resolve(...to));
+};
+
+console.log("Building web version...");
+
+// running web build
+// building automatically removes any existing 'dist' folder
+proc.execSync("yarn run build:web");
+
+// copying into temporary folder
+// so it doesn't get overwritten by next build step
+moveFile("dist", "dist-web");
+
+console.log("Building node version...");
+
+// running node build
+// building automatically removes any existing 'dist' folder
+proc.execSync("yarn run build:node");
+
+// create temporary folder 'dist-node' where we move our built node files
+fs.mkdirSync(path.resolve("dist-node"));
+
+console.log("Creating common types...");
+
+// Ok, now it's going to be weird
+// Let me explain...
+// Each build (web and node) contains its own typings
+// but this does not make sense, as both typings are exactly the same
+// that's why we are disposing one set of typings (node's typings)
+// and only use the typings generated along with the web build
+// therefore a lot of weird copying and moving is done here...sorry...
+
+// moving all node files into folder 'dist-node'
+// reason: we want to get rid of all typings
+moveFile("dist", "dist-node", "index.cjs.development.js");
+moveFile("dist", "dist-node", "index.cjs.development.js.map");
+moveFile("dist", "dist-node", "index.cjs.production.min.js");
+moveFile("dist", "dist-node", "index.cjs.production.min.js.map");
+moveFile("dist", "dist-node", "index.js");
+
+// finally we delete the folder with all node-typings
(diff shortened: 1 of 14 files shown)


packages/api/src/index.ts
```diff
@@ -30,7 +30,7 @@ export type Actions = { [key: string]: Action };
 // this represents valid currencies to check prices against on coingecko
 // see: https://www.coingecko.com/api/documentations/v3#/asset_platforms/get_asset_platforms
 export type CurrencySymbol = "usd";
-export type PriceSample = [timestamp: number, price: string];
```
mrice32, line 33: Why drop the names?

packages/sdk/tsdx.config.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = {
+  rollup(config) {
+    // removes the esm prefix -> that just makes importing the module more difficult
+    config.output.file = config.output.file.replace('.esm', '');
+
+    if (envType == 'web')
+      // unshift so that this plugin gets run first
+      config.plugins.unshift(
+        replace({ 
+          '@uma/contracts-node': "@uma/contracts-frontend",
```
mrice32, line 13: This is cool. Just to double check, is a string replace the standard way to replace imports? Also, does this replace it in the compiled output or does it modify the ts _before_ compilation? I assume the latter right?

packages/sdk/build.js
```diff
@@ -0,0 +1,86 @@
+// this file is adapted from https://blog.unterholzer.dev/cross-compatible-typescript-libraries/
```
mrice32, line 1: Out of curiousity, why is this necessary -- shouldn't rollup create two separate builds for us and we should just need to point module and main at the different ones?

packages/sdk/src/index.ts
```diff
@@ -8,11 +8,11 @@ export { default as Multicall2 } from "./multicall2";
 
 // types
 import type { TypedEventFilterEthers as TypedEventFilter, TypedEventEthers as TypedEvent } from "@uma/contracts-node";
-import { Contract } from "ethers";
-import { Result } from "@ethersproject/abi";
```
mrice32, line 12: Does removing these impact the bundle sizes?

mrice32 (approved):
Looks great! Only minor questions.

amateima (approved):

packages/api/src/index.ts
daywiss: i was getting a build error for some reason. going back to see if i still get it

packages/sdk/tsdx.config.js
daywiss: i didnt find any sources to show how to change imports based on environment during the build process. I kind of just tried this based on a slightly different tutorial for changing out code based on environment.  So no i dont think this is standard, but was not able to find a standard way to do this. 

I specifically had to push this plugin to the front of the stack so it ran before build. If you push this to the back the plugin array the build will fail!

packages/sdk/build.js
daywiss: This was copied from a tutorial. The problem is the sdk is using tsdx, which outputs the build to a single folder and will override the results of a previous build.  So this hacks around this by moving files around so previous build does not get overrwritten. There may be a better way to do this though

packages/sdk/src/index.ts
daywiss: no it doesnt, this was an artifact of me trying to remove the ethersproject dependency.  I will leave it in since i found that you can access these things from the `ethers` library itself.

packages/sdk/build.js
daywiss: oh this file also copies the types file so that both builds have the same interface

packages/api/src/index.ts
daywiss: looks like it does build with this in now

PR #3394, opened 2021-09-27
fix(api): discovery events and block interval check

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/2354/fix-lsp-api-contract-discovery

**Summary**

There are 2 minor issues here:
1. lsp app incorrectly passed events
2. block interval was opinionated on block numbers, throwing a  lot of errors during testing

Fixed the lsp event issue by passing data correctly into the events. Second issue was fixed by removing assert from utility and handling the issue by the caller. In this case we choose to just not run updates if startblock >= endblock. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2354/fix-lsp-api-contract-discovery

### DIFF
--- a/packages/api/src/apps/api/app.ts
+++ b/packages/api/src/apps/api/app.ts
@@ -221,12 +221,18 @@ export default async (env: ProcessEnv) => {
   console.log("Started Express Server, API accessible");
 
   async function detectNewContracts(startBlock: number, endBlock: number) {
+    // ignore case when startblock == endblock, this can happen when loop is run before a new block has changed
+    if (startBlock === endBlock) return;
+    assert(startBlock < endBlock, "Startblock must be lower than endBlock");
     await services.registry(startBlock, endBlock);
     await services.lspCreator.update(startBlock, endBlock);
   }
 
   // break all state updates by block events into a cleaner function
   async function updateContractState(startBlock: number, endBlock: number) {
+    // ignore case when startblock == endblock, this can happen when loop is run before a new block has changed
+    if (startBlock === endBlock) return;
+    assert(startBlock < endBlock, "Startblock must be lower than endBlock");
     await services.blocks.handleNewBlock(endBlock);
     // update everyting
     await services.emps.update(startBlock, endBlock);
--- a/packages/api/src/apps/lsp_api/app.ts
+++ b/packages/api/src/apps/lsp_api/app.ts
@@ -131,7 +131,9 @@ export default async (env: ProcessEnv) => {
     // these services can optionally be configured with a config object, but currently they are undefined or have defaults
     blocks: Services.Blocks(undefined, appState),
     emps: Services.EmpState({ debug }, appState),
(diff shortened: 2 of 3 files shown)


packages/api/src/apps/api/app.ts
```diff
@@ -221,12 +221,16 @@ export default async (env: ProcessEnv) => {
   console.log("Started Express Server, API accessible");
 
   async function detectNewContracts(startBlock: number, endBlock: number) {
+    // ignore case if somehow start block and end block are out of whack
+    if (startBlock >= endBlock) return;
```
mrice32, line 225: When does this happen? Why does this happen?

packages/api/src/apps/lsp_api/app.ts
```diff
@@ -131,7 +131,9 @@ export default async (env: ProcessEnv) => {
     // these services can optionally be configured with a config object, but currently they are undefined or have defaults
     blocks: Services.Blocks(undefined, appState),
     emps: Services.EmpState({ debug }, appState),
-    registry: await Services.Registry({ debug }, appState, (data) => serviceEvents.emit("empRegistry", data)),
+    registry: await Services.Registry({ debug }, appState, (event, data) =>
+      serviceEvents.emit("empRegistry", event, data)
```
mrice32, line 135: Are event and data different types? Surprised typescript didn't help here

mrice32 (commented):
Looks good! Two quick questions

packages/api/src/apps/api/app.ts
daywiss: `startBlock == endBlock` happens only during testing where the loop runs several time before anything triggers a new block to be mined in the test.

packages/api/src/apps/lsp_api/app.ts
daywiss: yea that is actually really wierd. typescript should have saw that, will look into it

mrice32 (approved):
Looks great! One minor suggestion: https://github.com/UMAprotocol/protocol/pull/3394#discussion_r717592559.

packages/api/src/apps/api/app.ts
daywiss: added an assert for that case and clarified comment to say this can happen if a block isnt advanced before loop runs again

amateima (approved):

packages/api/src/apps/lsp_api/app.ts
daywiss: for some reason its treating functional arguments as optional, even though they should be defined:
`(event: Events, data: EmitData) => void` is the callback signature, but i can just pass in `()=>{..etc}` and typescript does not complain

daywiss: https://github.com/Microsoft/TypeScript/wiki/FAQ#why-are-functions-with-fewer-parameters-assignable-to-functions-that-take-more-parameters

PR #3395, opened 2021-09-27
feat(insured-bridge): Make XMessenger contracts Ownable

nicholaspai:
**Motivation**

The Messenger contracts are the ones to actually relay cross-chain functions, so in order to maintain the L1BridgeAdmin --> L2DepositBox admin link we should make the XMessenger contracts owned by the BridgeAdmin so that only the BridgeAdmin can originate cross-chain admin transactions.

**Details**

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/test/insured-bridge/ArbitrumMessenger.js
```diff
@@ -81,7 +81,22 @@ describe("ArbitrumMessenger integration with BridgeAdmin", () => {
       from: owner,
     });
   });
+  describe("relayMessage only callable by owner", function () {
```
chrismaree, line 84: small nit: do you really need both the `describe` and `it` here? fine to leave it as is just it's a bit redundant seeing there is only one test in the `describe`.

packages/core/contracts-ovm/insured-bridge/implementation/OptimismMessenger.sol
```diff
@@ -23,7 +26,7 @@ contract OptimismMessenger is OVM_CrossDomainEnabled, MessengerInterface {
         uint256 gasLimit,
         uint256 gasPrice,
         bytes memory message
-    ) external override {
+    ) external override onlyOwner {
```
chrismaree, line 29: 👍

packages/core/test/insured-bridge/ArbitrumMessenger.js
nicholaspai: removed

PR #3397, opened 2021-09-27
feat(core): move all active scripts out of core and deprecate rest

mrice32:
**Motivation**

We need all active scripts to no longer rely on truffle so we can drop truffle compilation.

**Summary**

This PR does a few things:
- It moves active scripts to the scripts pacakge and makes them runnable via hardhat or node.
- It moves inactive scripts (ones we don't use any more) to a deprecated folder in core, where they will stay until we decide we are confident enough to remove them completely.
- It updates the scripts in the trader package to drop direct reliance on core and truffle.
- It moves some functionality from common to financial templates lib so common no longer has an implicit runtime dependence on truffle contracts in core.
- Related to the above, it adds some functionality to contracts-node and contracts-frontend to get a list of all contracts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Note: most of these scripts were run on a hardhat node to test the changes. A few were not due to the complexity of setup. These include emergency shutdown, the trader scripts, and the gas ETH query script.

**Issue(s)**

N/A


packages/scripts/src/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,107 @@
...
+ * @dev This service account will need GCP admin or BigQuery permissions. This guide provides further instructions: https://cloud.google.com/docs/authentication/getting-started
+ * @dev How to run: ./scripts/local/getGasEthHistoricalPrice.js --time 1612137600 --identifier GASETH-TWAP-1Mx1M
+ */
+const { BigQuery } = require("@google-cloud/bigquery");
+const highland = require("highland");
+const moment = require("moment");
+const Web3 = require("web3");
+const { fromWei } = Web3.utils;
+const argv = require("minimist")(process.argv.slice(), { string: ["identifier", "time"] });
+const { createQuery } = require("./GasEthQuery");
```
smb2796, line 17: This depends on the separate `GasEthQuery` which is no longer in the same directory.

packages/scripts/src/local/getGasEthHistoricalPrice.js
```diff
@@ -0,0 +1,107 @@
...
+
+/**
+ * @notice This is an example script demonstrating how to get a historical median gas price at a specific timestamp.
+ * It could be used for example to query a GASETH-TWAP-1Mx1M price before committing a vote for a DVM price request.
+ *
+ * @dev Prequisites: before running, you will need to set a GOOGLE_APPLICATIONS_CREDENTIALS environment variable for a service account.
+ * @dev This service account will need GCP admin or BigQuery permissions. This guide provides further instructions: https://cloud.google.com/docs/authentication/getting-started
+ * @dev How to run: ./scripts/local/getGasEthHistoricalPrice.js --time 1612137600 --identifier GASETH-TWAP-1Mx1M
+ */
+const { BigQuery } = require("@google-cloud/bigquery");
```
smb2796, line 11: Not positive, but it looks like these dependencies may be broken.

smb2796:
Couple of small comments for `getGasEthHistoricalPrice.js`, but overall looks good to me! Would like @pemulis to take a look at the OO deposit box and proposal script changes though, as my review is not as helpful on those.

packages/scripts/src/local/getGasEthHistoricalPrice.js
mrice32: Good point! Thanks for pointing this out. Now I remember that I didn't test this script because it requires special GCP permissions to run. I should have mentioned that in the description.

packages/scripts/src/local/getGasEthHistoricalPrice.js
mrice32: Should be fixed!

packages/scripts/src/local/getGasEthHistoricalPrice.js
mrice32: Fixed.

mrice32:
@smb2796 I confirmed the GasEth script works now:
```bash
./src/local/getGasEthHistoricalPrice.js 
Optional '--identifier' flag not specified, defaulting to: GASETH-TWAP-1Mx1M
Optional '--time' flag not specified, defaulting to the current Unix timestamp (in seconds): 1632846816
⏰ Fetching nearest prices for the timestamp: Tue, 28 Sep 2021 16:33:36 GMT

⚠️ Truncating price to 18 decimals
```

packages/scripts/src/local/getHistoricalPrice.js
```diff
@@ -0,0 +1,120 @@
...
+/**
+ * @notice This is an example script demonstrating how to get a historical price at a specific timestamp.
+ * It could be used for example to query a price before committing a vote for a DVM price request.
+ * We provide default configurations for querying median prices on specific markets across some exchanges and markets.
+ * This script should serve as a template for constructing other historical median queries.
+ *
+ * @notice This script will fail if the `--time` is not within a the caller's configured lookback window. The caller
+ * set this via the PRICEFEED_CONFIG={"lookback":x} environment variable.
+ * @dev How to run:
+ *     HARDHAT_NETWORK=mainnet ./src/local/getHistoricalPrice.js
```
nicholaspai, line 12: nit: don't you need to run this command with with `node x`? Or can you just run the `.js` file directly?

nicholaspai (approved):
I mainly looked at `getHistoricalGasPrice` which I know is used by the community and @smb2796

smb2796 (approved):
Thanks for addressing

packages/scripts/src/local/getHistoricalPrice.js
mrice32: I put the shebang at the top so you don't need it.

pemulis (approved):
LGTM. The deposit box tutorial updates seem reasonable and there were no changes to the admin proposal scripts besides the directory change.

PR #3398, opened 2021-09-27
improve(dependencies): drop truffle from repo and remove truffle build

mrice32:
**Motivation**

We would like to no longer depend on truffle.

**Summary**

This drops the truffle dependency from the repo, removes the truffle build, and drops anything else that was only used by truffle or required truffle.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


package.json
```diff
@@ -15,7 +15,7 @@
     "load-addresses": "yarn workspace @uma/core load-addresses",
     "create-release": "lerna version --no-git-tag-version --conventional-commits --no-push",
     "publish-release": "./scripts/run_release.sh",
-    "qbuild": "yarn workspace @uma/common build && yarn workspace @uma/core build && yarn lerna run build --stream --ignore '*/*dapp*' --ignore '@uma/common' --ignore '@uma/core'",
+    "qbuild": "yarn lerna run build --stream",
```
mrice32, line 18: We can now simplify the build command because there are no circular deps!

packages/core/index.ts
```diff
@@ -1,158 +1 @@
...
-  return truffleContract.abi;
-};
-
-const globalArtifactsExists = Boolean(((global as unknown) as { artifacts: any }).artifacts);
-const getAbi = globalArtifactsExists ? getAbiTest : getAbiProd;
-const getAddress = globalArtifactsExists ? getAddressTest : getAddressProd;
-const getTruffleContract = globalArtifactsExists ? getTruffleContractTest : getTruffleContractProd;
-
-export { getAbi, getAddress, getTruffleContract, findContractVersion };
+export { findContractVersion } from "./src/FindContractVersion";
```
mrice32, line 1: This is the only function we're still using in core. Not too concerned about relying on it. Maybe down the road, we can move it out.

packages/core/migrations/10_deploy_designated_voting_factory.js
```diff
@@ -1,11 +0,0 @@
-const DesignatedVotingFactory = artifacts.require("DesignatedVotingFactory");
```
mrice32, line 1: All migrations were removed.

packages/core/package.json
```diff
@@ -4,24 +4,17 @@
   "description": "UMA smart contracts and unit tests",
   "dependencies": {
     "@openzeppelin/contracts": "4.1.0",
-    "@truffle/contract": "^4.3.15",
     "@uma/common": "^2.7.0",
-    "@uma/core-1-1-0": "npm:@uma/core@1.1.0",
```
mrice32, line 9: No longer need older versions of core to support the index.ts file functions.

packages/core/package.json
```diff
@@ -4,24 +4,17 @@
...
     "@uniswap/v2-periphery": "1.1.0-beta.0",
     "@uniswap/v3-core": "^1.0.0-rc.2",
-    "@uniswap/v3-periphery": "^1.0.0-beta.23",
-    "arb-ts": "^0.0.42",
-    "decimal.js": "^10.2.1",
-    "ethereum-waffle": "^3.4.0"
+    "@uniswap/v3-periphery": "^1.0.0-beta.23"
   },
   "devDependencies": {
+    "arb-ts": "^0.0.42",
```
mrice32, line 15: I don't think these are needed at runtime (i.e. in contracts), so they were moved down to dev dependencies.

packages/core/package.json
```diff
@@ -70,30 +61,27 @@
...
+    "test": "yarn hardhat-test",
     "test-e2e": "mocha -test ./.mocharc.js --e2e",
-    "truffle-test": "truffle test ./truffle-test/scripts/*",
     "hardhat-test": "hardhat test --network hardhat",
     "mocha-test": "mocha ./mocha-test --recursive",
-    "load-addresses": "yarn run apply-registry && yarn hardhat load-addresses",
-    "clean": "rm -rf build && rm -rf contract-types && rm -rf artifacts && rm -rf deployments && rm -rf dist",
-    "build-contracts": "yarn truffle compile && yarn hardhat compile && yarn hardhat compile --network optimism",
+    "load-addresses": "yarn hardhat load-addresses",
+    "clean": "rm -rf contract-types && rm -rf artifacts && rm -rf deployments && rm -rf dist",
```
mrice32, line 79: Most of this is just cleaning up old truffle commands.

packages/trader/src/index.ts
```diff
@@ -4,7 +4,7 @@ import retry from "async-retry";
 import { config } from "dotenv";
 
 import { getWeb3 } from "@uma/common";
-import { getAbi, getAddress } from "@uma/core";
+import { getAbi, getAddress } from "@uma/contracts-node";
```
mrice32, line 7: This is a drop-in replacement almost everywhere.

package.json
nicholaspai: v exciting

packages/common/src/ProviderUtils.ts
```diff
@@ -92,10 +92,13 @@ function getDefaultKeyType(): typeof KEY_TYPES[number] {
   return "none";
 }
 
+export function getMnemonic(): string {
```
nicholaspai, line 95: +1 useful function

packages/api/package.json
```diff
@@ -5,6 +5,7 @@
   "author": "David Adams",
   "license": "AGPL-3.0-or-later",
   "devDependencies": {
+    "@types/cors": "^2.8.12",
```
chrismaree, line 8: why was this needed to be added here?

packages/core/README.md
```diff
@@ -4,155 +4,82 @@ This package contains contract artifacts for all UMA smart contracts.
...
 yarn add @uma/core
 ```
 
-## Importing artifacts directly
+## Deploying the contracts
 
-The easiest way to import contracts is to do so by importing the file from
-`@uma/core/build/contracts/ContractName.json`. This is especially useful in certain contexts where imports cannot be
-dynamically loaded (generally non-nodejs contexts):
+To deploy the contracts, we use the `hardhat-deploy` package. This may be failiar to users of `hardhat`. The deployment
```
nicholaspai, line 19:
```suggestion
To deploy the contracts, we use the `hardhat-deploy` package. This may be familiar to users of `hardhat`. The deployment
```

packages/core/package.json
nicholaspai: !!

packages/core/package.json
nicholaspai: Agreed

packages/common/src/SolcoverConfig.ts
```diff
@@ -1,10 +0,0 @@
-export const SolcoverConfig = {
```
chrismaree, line 1: this is not truffle specific. it actually looks like the last time our coverage ran was feb of this year: https://coveralls.io/github/UMAprotocol/protocol?branch=master

image


we should consider adding a task to get this back up and running.

packages/core/README.md
mrice32: Lol failiar

packages/common/src/TruffleConfig.ts
```diff
@@ -1,242 +0,0 @@
-/**
```
chrismaree, line 1: 👋

nicholaspai (approved):
VERY satisfying PR this is amazing. I'm so excited, LGTM

packages/core/index.ts
chrismaree: ye, could easily move it common

chrismaree (approved):
this is amazing! the culmination of many many hours of painful work all come together in this RR. really happy to see all these migrations and truffle dependent things being removed.

In other related news, I've spent a fair amount of time getting familiar with the hardhat console and I'm happy to report that for many actions it's actually much better than the truffle console!

packages/api/package.json
mrice32: Good question. The api build began failing because it couldn't resolve the types for the cors package. My (and @daywiss's) assumption is that this was previously getting this dep via some nested package that was installed under truffle, so once it was removed this began failing.

Anyway, the package uses the cors types, so it should be including them as a dep. It was a little odd to see it sporadically fail like this though.

PR #3399, opened 2021-09-28
feat(insured-bridge): War games 1

chrismaree:
**Motivation**

This PR adds required changes after the completion of War games 1. For across, done on Kovan.

**Summary**

- A few bugs and todos were implemented.
- migration scripts added.
- task for deploying across pools created.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/HardhatConfig.ts
```diff
@@ -70,13 +72,30 @@ export function getHardhatConfig(
         testBlacklist,
       },
       localhost: { url: "http://127.0.0.1:9545", timeout: 1800000, testBlacklist },
+      mainnet: { chainId: 1, url: getNodeUrl("mainnet", true), accounts: { mnemonic } },
```
chrismaree, line 75: orded by chain ID now

packages/common/src/hardhat/tasks/acrossPool.ts
```diff
@@ -0,0 +1,35 @@
+import { task, types } from "hardhat/config";
```
chrismaree, line 1: hard hat task to create new pools.

packages/common/src/hardhat/tasks/acrossPool.ts
```diff
@@ -0,0 +1,35 @@
...
+      bridgeAdmin.options.address, // _bridgeAdmin
+      l1tokenaddress, // _l1Token
+      lpfeeratepersecond, // _lpFeeRatePerSecond
+      ZERO_ADDRESS, // _timer
+    ];
+
+    const bridgePool = await deploy("BridgePool", { from: deployer, args, log: true });
+
+    console.log("Bridge pool deployed @ ", bridgePool.address);
+  });
```
chrismaree, line 35: the FPL below was renamed. it had the word "copy" in it, which was clearly wrong.

packages/core/deploy/035_deploy_across_l1_bridge_admin.js
```diff
@@ -0,0 +1,19 @@
+const func = async function (hre) {
```
chrismaree, line 1: admin migration!

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
+// This deploy script should be run on an optimism provider.
```
chrismaree, line 1: This script follows a bit of a different pattern to what we've done before: it checks the `chainID` against a set of known IDs and uses this the find the associated `BridgeAdmin` address on L1.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -150,9 +151,6 @@ export class InsuredBridgeL1Client {
     const whitelistedTokenEvents = await this.bridgeAdmin.getPastEvents("WhitelistToken", blockSearchConfig);
     for (const whitelistedTokenEvent of whitelistedTokenEvents) {
       const l1Token = whitelistedTokenEvent.returnValues.l1Token;
-      // If the data structure already contains information on this l1Token( re-whitelist of an existing token) continue.
```
chrismaree, line 153: this was actually wrong before. If there are multiple white-listed events we should take the _newest_ event.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -85,7 +86,11 @@ export class InsuredBridgeL2Client {
 
     this.firstBlockToSearch = blockSearchConfig.toBlock + 1;
 
-    this.logger.debug({ at: "InsuredBridgeL2Client", message: "Insured bridge l2 client updated" });
+    this.logger.debug({
```
chrismaree, line 89: enhance logs with `chainid`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -63,20 +63,28 @@ export class Relayer {
       return;
     }
     for (const l1Token of Object.keys(relayableDeposits)) {
-      this.logger.debug({ at: "Relayer", message: "Processing relayable deposits for L1Token", l1Token });
+      this.logger.debug({
```
chrismaree, line 66: all changes in this file are to enhance logs.

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -1,10 +1,21 @@
...
+const { isAddress, toChecksumAddress, toBN } = Web3.utils;
 
 import assert from "assert";
 
 import type { RateModel } from "@uma/financial-templates-lib";
 
+// Check each token rate model contains the expected data.
+const expectedRateModelKeys = ["UBar", "R0", "R1", "R2"];
+
+// const supported L2 Chain IDS:
```
chrismaree, line 11: limit the chain ids the user can provide to only chains we explicitly support

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -35,23 +47,26 @@ export class RelayerConfig {
...
+        `${toChecksumAddress(l1Token)} does not contain the required rate model keys ${expectedRateModelKeys}`
       );
-      this.rateModels[l1Token] = {
-        UBar: processingRateModels[l1Token].UBar,
-        R0: processingRateModels[l1Token].R0,
-        R1: processingRateModels[l1Token].R1,
-        R2: processingRateModels[l1Token].R2,
+      this.rateModels[toChecksumAddress(l1Token)] = {
+        UBar: toBN(processingRateModels[l1Token].UBar),
+        R0: toBN(processingRateModels[l1Token].R0),
```
chrismaree, line 60: this was an error before. the `rateModel` should be a bn

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -27,28 +28,44 @@ export async function run(logger: winston.Logger, web3: Web3): Promise<void> {
...
+    const [accounts] = await Promise.all([l1Web3.eth.getAccounts()]);
 
     const gasEstimator = new GasEstimator(logger);
 
     // Create L1/L2 clients to pull data to inform the relayer.
     // todo: add in start and ending block numbers (if need be).
-    const l1Client = new InsuredBridgeL1Client(logger, web3, config.bridgeAdmin, config.rateModels);
+    const l1Client = new InsuredBridgeL1Client(logger, l1Web3, config.bridgeAdmin, config.rateModels);
+
+    // TODO: Add a method to fetch all registered chainIDs from bridge admin to let the bot default to all chains when
```
chrismaree, line 42: note for now we only use the chainIds provided by the user. in the future this will auto-detect all loaded chain-ids on L1 and runn accordingly.

packages/common/src/HardhatConfig.ts
nicholaspai: Don't you need `Legacy_Testable` and `Legacy_Lockable` too?

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
...
+
+  if (chainId != "69" || chainId != "10") throw new Error("This deploy only works on optimism chains");
+
+  // Map L2 chain IDs to L1 chain IDs to find associated bridgeAdmin addresses for a given L2 chain ID.
+  const l2ChainIdToL1 = {
+    69: 42, // optimism testnet -> kovan
+    10: 1, // optimism mainnet -> mainnet
+  };
+
+  const brideAdminAddress = await getAddress("BridgeAdmin", l2ChainIdToL1[chainId]);
```
nicholaspai, line 19:
```suggestion
  const bridgeAdminAddress = await getAddress("BridgeAdmin", l2ChainIdToL1[chainId]);
```

packages/core/networks/42.json
```diff
@@ -98,5 +98,9 @@
   {
     "contractName": "RangeBondLongShortPairFinancialProductLibrary",
     "address": "0x09381C827438605B5ea3e760A27666Be3497602C"
+  },
+  {
+    "contractName": "BridgeAdmin",
```
nicholaspai, line 103: Why not add BridgePool to this file?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Now we will overwrite the whitelist events and use the newest one right?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -127,12 +135,21 @@ export class Relayer {
...
-        (await this.l1Client.calculateRealizedLpFeePctForDeposit(deposit)).toString() &&
-      relay.depositContract === deposit.depositContract
-    );
+  private async isRelayValid(relay: Relay, deposit: Deposit): Promise<{ isValid: boolean; reason: string }> {
+    const relayRealizedLpFeePct = relay.realizedLpFeePct.toString();
+    const expectedRelayRealizedLpFeePct = (await this.l1Client.calculateRealizedLpFeePctForDeposit(deposit)).toString();
+    if (relayRealizedLpFeePct !== expectedRelayRealizedLpFeePct)
+      return {
+        isValid: false,
+        reason: `relayRealizedLpFeePct: ${relayRealizedLpFeePct}!=expectedRelayRealizedLpFeePct: ${expectedRelayRealizedLpFeePct}`,
```
nicholaspai, line 144:
```suggestion
        reason: `relayRealizedLpFeePct: ${relayRealizedLpFeePct} != expectedRelayRealizedLpFeePct: ${expectedRelayRealizedLpFeePct}`,
```

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -127,12 +135,21 @@ export class Relayer {
...
+      };
+    if (relay.depositContract !== deposit.depositContract)
```
nicholaspai, line 146: nit: why not use `else if`?

nicholaspai: Also this test case seems really hard to recreate, I'm curious if you tested it?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -127,12 +135,21 @@ export class Relayer {
...
+      return {
+        isValid: false,
+        reason: `relay.depositContract: ${relay.depositContract}!=deposit.depositContract: ${deposit.depositContract}`,
```
nicholaspai, line 149:
```suggestion
        reason: `relay.depositContract: ${relay.depositContract} != deposit.depositContract: ${deposit.depositContract}`,
```

nicholaspai (commented):
Thanks for doing this, left some nits. Re request after you review!

packages/common/src/HardhatConfig.ts
chrismaree: I think that these are compiled indirectly due to the inheritance path. I aded them anyways

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: correct. during testing I had to re-whitelist (due to re-deplyoing) and it did not use the new version. after removing this code it always uses the newest one. this could could be bad if people deposit into the old one but we should ideally disable the old one and finish all bridging actions before re-whitelisting.

packages/core/networks/42.json
chrismaree: there is more than 1 `BridgePool` per network! how would we add `n` number of them to this file?

nicholaspai: I guess not since we'll never just call `getAddress("BridgePool")` since the caller will need to know which one they want!

packages/common/package.json
```diff
@@ -55,7 +55,7 @@
   "devDependencies": {
     "@eth-optimism/hardhat-ovm": "^0.2.2",
     "@nomiclabs/hardhat-ethers": "^2.0.2",
-    "@nomiclabs/hardhat-etherscan": "^2.1.4",
+    "@nomiclabs/hardhat-etherscan": "^2.1.6",
```
mrice32, line 58: Out of curiosity, what did this fix?

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
+// This deploy script should be run on an optimism provider.
+const { getAddress } = require("@uma/contracts-node");
```
mrice32, line 2: We might want to be cautious about depending on contracts-node in core since it creates a circular dep. Is there a way to use hardhat directly to grab the admin address (there might not be, and we might be able to just leave this as is and get away with it)?

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -1,10 +1,21 @@
...
+// Check each token rate model contains the expected data.
+const expectedRateModelKeys = ["UBar", "R0", "R1", "R2"];
+
+// const supported L2 Chain IDS:
+const supportedChainIds = [
+  10, // optimism mainnet
+  69, // optimism testnet
+  42161, // arbitrum mainnet
+  421611, // arbitrum testnet
+];
```
mrice32, line 17: Why hardcode this? Should this list be an optional config param instead?

packages/common/src/HardhatConfig.ts
```diff
@@ -70,13 +72,30 @@ export function getHardhatConfig(
...
-      matic: { chainId: 137, url: getNodeUrl("polygon-matic", true), accounts: { mnemonic } },
-      mainnet: { chainId: 1, url: getNodeUrl("mainnet", true), accounts: { mnemonic } },
+
+      kovan: { chainId: 42, url: getNodeUrl("kovan", true), accounts: { mnemonic } },
       optimism: {
+        ovm: true,
+        chainId: 10,
+        url: getNodeUrl("optimism", true),
+        accounts: { mnemonic },
+        gasPrice: 15000000,
```
mrice32, line 85: Out of curiosity, why do we need a hardcoded gas price?

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
...
+const { ZERO_ADDRESS } = require("@uma/common");
+const func = async function (hre) {
+  const { deployments, getNamedAccounts } = hre;
+  const { deploy } = deployments;
+
+  const { deployer } = await getNamedAccounts();
+  const chainId = await web3.eth.net.getId();
+
+  if (chainId != "69" || chainId != "10") throw new Error("This deploy only works on optimism chains");
```
mrice32, line 11: Interesting, will this break tests that run all deployments by default? Is there a way, instead to have a switch where we deploy the OVM version or the non-OVM version of the deposit contract depending on which chain rather than erroring?

mrice32 (commented):
This is awesome! Great work @chrismaree! A few minor comments

nicholaspai (approved):
LGTM after you review matt's comments

packages/common/package.json
chrismaree: in theory, this gives us optimism/arbitrum verification. in practice, this only works on their mainnet

packages/common/src/HardhatConfig.ts
chrismaree: all configs I've seen do this for optimism. the hardhat hat defaults are not representative of what the mainnet is like. This is what all the tutorial/example repos have so I just copied that.

packages/insured-bridge-relayer/src/RelayerConfig.ts
chrismaree: in the future, yes. In the future, the bot can figure out which chain IDs are whitelisted on the contract and then run there. Right now, though, to move as fast as possible, we don't do any auto-chainId detection and force the user to provide

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
chrismaree: I'm not really sure what the best pattern is for this and you do raise an excellent point.

More generally, this kind of migration does not follow the patterns our other ones do as it returns early if not running on specific chain IDs. it might be a good idea to put this in a different directory `ovm_deploy` or something to this effect that does not run as part of the normal migration and gives us the space to separate out these kinds of migrations. wdyt?

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
chrismaree: ye, this broke things. I've changed it now to ignore these migrations unless you are on those chains. Not sure what the best pattern is for this tbh.

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
+// This deploy script should be run on an optimism provider.
+const { getAddress } = require("@uma/contracts-node");
+const { ZERO_ADDRESS } = require("@uma/common");
+const func = async function (hre) {
```
mrice32, line 4: How does this get ignored for chains that don't have the right chainIds?

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
```diff
@@ -0,0 +1,30 @@
...
+
+  const { deployer } = await getNamedAccounts();
+
+  // Map L2 chain IDs to L1 chain IDs to find associated bridgeAdmin addresses for a given L2 chain ID.
+  const l2ChainIdToL1 = {
+    69: 42, // optimism testnet -> kovan
+    10: 1, // optimism mainnet -> mainnet
+  };
+
+  const bridgeAdminAddress = await getAddress("BridgeAdmin", l2ChainIdToL1[chainId]);
```
mrice32, line 18: This may be a little hacky, but maybe the simplest way to do this without having to change the way we do migrations _and_ have this contract _always_ be available is that if this deploy script is called on a chain without a pre-existing mapping, we just set it to map to _its_ own BridgeAdmin so we have the full ecosystem operating on a single chain.

```js
const bridgeAdminAddress = l2ChainIdToL1[chainId] ? await getAddress("BridgeAdmin", l2ChainIdToL1[chainId]): (await deployments.get("BridgeAdmin")).address;
```

Thoughts?

packages/common/src/HardhatConfig.ts
mrice32: Makes sense!

packages/common/src/hardhat/fixtures/default.ts
```diff
@@ -4,14 +4,32 @@ import { RegistryRolesEnum } from "../../Enums";
 import { CombinedHRE } from "../tasks/types";
 import { DeploymentsExtension } from "hardhat-deploy/types";
 
+const defaultFixtureWhitelist = [
```
mrice32, line 7: Not sure I like having deployments _not_ added to this whitelist by default. Seems like it could end up being error-prone. Maybe we could consider something less invasive as suggested below or maybe a blacklist?

chrismaree: whitelisting will include less than blacklisting so hopefully will be lighter/faster.

If we did a blacklist, how would you like to construct this? ideally, we'd want a methods that can list all Tagged scripts and then we can iterate over them and remove them. I'll look into the hardhat settings to facilitate this.

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
chrismaree: In the form it has now, there is no ignoring; if you run the deploy script it will run no matter the chainID. previously, it had return early code. now, we just work around the problem with the default fixtures implementation.

packages/core/deploy/036_deploy_across_ovm_bridge_deposit_box.js
chrismaree: sure. that sounds reasonable.

packages/common/src/hardhat/fixtures/default.ts
mrice32: If we do this https://github.com/UMAprotocol/protocol/pull/3399#discussion_r718697665, can we just _not do either_ for now?

mrice32: Personally, I like the idea of deployment making all singleton contracts available, and if you need to do something more specialized, you write a special fixture for it, rather than making the default fixture opinionated. But I may be in the minority on that opinion :). Either way you decide to go is okay with me!

mrice32 (approved):
LGTM! Just one remaining thread about the blacklist/whitelist stuff: https://github.com/UMAprotocol/protocol/pull/3399#discussion_r719713128. Either way you decide to go, LGTM!

PR #3406, opened 2021-09-28
feat(insured-bridge): gas optimizations in bridge pool

mrice32:
**Motivation**

We'd like across relays to have low gas costs.

**Summary**

This PR includes various gas optimizations. The biggest is that we compress ancillary data into a single hash, which means we have less data to convert to utf8 (which was the source of a lot of our gas costs). Relatedly, we had some repeated in-line calls to `getRelayAncillaryData` within the same method. Removing those saved a lot of gas because of how expensive those conversions were.

Gas costs for a happy-path relay cycle before this PR:
```
Relay: 904376
Speed up: 94194
Settle: 644354
Total: 1642924
```

Gas costs after this change:
```
Relay: 542766
Speed up: 74980
Settle: 312570
Total: 930316
```

This amounts to a savings of `710108` or ~43%.

Note: there were some minor changes to events and storage that may impact bots and tests.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -92,25 +92,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     event LiquidityRemoved(address indexed token, uint256 amount, uint256 lpTokensBurnt, address liquidityProvider);
     event DepositRelayed(
         uint256 indexed relayId,
-        uint8 chainId,
-        uint64 depositId,
-        address indexed l2Sender,
+        DepositData depositData,
```
nicholaspai, line 95: I'm surprised you don't need to specify this as a bytes32 type. I guess solidity automatically casts it? It might be better to explicitly use bytes32

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -257,22 +245,18 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-            RelayData({
-                relayId: numberOfRelays,
-                relayState: RelayState.Pending,
-                priceRequestTime: priceRequestTime,
-                realizedLpFeePct: realizedLpFeePct,
-                slowRelayer: msg.sender,
-                instantRelayer: relays[depositHash].instantRelayer
-            });
-        relays[depositHash] = relayData;
+        RelayData storage relayData = relays[depositHash];
```
nicholaspai, line 248: This syntax produces gas savings..?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -281,15 +265,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
+        _requestOraclePriceRelay(amount, priceRequestTime, ancillaryData);
+        _proposeOraclePriceRelay(amount, priceRequestTime, ancillaryData);
 
         pendingReserves += amount; // Book off maximum liquidity used by this relay in the pending reserves.
 
         // We use an internal method to emit this event to overcome Solidity's "stack too deep" error.
-        _emitDepositRelayedEvent(depositData, realizedLpFeePct, depositHash);
-
-        numberOfRelays += 1;
```
nicholaspai, line 292: +1

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -497,72 +480,56 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-            "instantRelayFeePct",
-            uint256(_depositData.instantRelayFeePct)
-        );
-        intermediateAncillaryData = AncillaryData.appendKeyValueUint(
-            intermediateAncillaryData,
-            "quoteTimestamp",
-            uint256(_depositData.quoteTimestamp)
-        );
+        return
+            AncillaryData.getKeyValueBytes32(
```
nicholaspai, line 484: To maintain consistency with the rest of AncillaryDataLib I think you should rename this function to "append..."

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -497,72 +480,56 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        intermediateAncillaryData = AncillaryData.appendKeyValueUint(
-            intermediateAncillaryData,
-            "realizedLpFeePct",
-            uint256(_relayData.realizedLpFeePct)
-        );
+    // Note: this method is identical to the one above, but it allows storage to be passed in, which saves some gas (3-4k)
+    // when called internally due to solidity not needing to copy the entire data structure and just lazily read data
+    // when requested.
+    function _getRelayAncillaryData(bytes32 relayHash) private view returns (bytes memory) {
+        // TODO: Consider adding BridgePool address to the ancillary data packet.
```
nicholaspai, line 508: I think this todo is not necessary and can be removed. The bridge pool address is always linked to the event that it emits

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -497,72 +480,56 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        );
-        intermediateAncillaryData = AncillaryData.appendKeyValueUint(
-            intermediateAncillaryData,
-            "quoteTimestamp",
-            uint256(_depositData.quoteTimestamp)
-        );
+        return
+            AncillaryData.getKeyValueBytes32(
+                "hash",
+                keccak256(
```
nicholaspai, line 486: Why not use internal method _getEelayHash?

nicholaspai (commented):
Woah this is an awesome start. I think this simplifies the relayer bot logic a lot too since wut only needs to match the hash of the deposit event params and the relay hash. Obviously next steps are to change the off chain client logic and tests

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Hmmm, as far as I'm aware the struct gets encoded into the Abi so clients can decide the full struct transparently. I think this would be the same cost as a bytes array, but clients would just have an easier time deciding the info in web3 or ethers.

Maybe I'm misunderstanding your point tho

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Kind of. I think it produces gas savings if you're not overwriting all fields (in this case we aren't overwriting the instant relayer. So, it doesn't need to read/write that field at all, which should save at least a few hundred gas.

When I tested it seemed to save a few k. I can get exact numbers if you'd like!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Good point. Will try making it append and see if that costs much gas.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Sounds good -- will remove!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: I changed the internal ones to take storage variables to save gas (I should go and see how much). Since these are being called from offchain, the inputs aren't storage, so those methods can't be used.

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: > Hmmm, as far as I'm aware the struct gets encoded into the Abi so clients can decide the full struct transparently. 

Ah I see what you're saying. So the entire `depositData` struct is emitted (cheaply I'm hoping). Didn't know you could do this haha

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: That makes sense no need to get exact numbers if thats painful although that'd be insightful.

packages/core/contracts/common/implementation/AncillaryData.sol
```diff
@@ -29,6 +29,18 @@ library AncillaryData {
         return s;
     }
 
+    function toUtf8Bytes(bytes32 x) internal pure returns (bytes memory) {
```
nicholaspai, line 32: So this is the same implementation as `toUtf8BytesAddress` except that `s` is of size 64? Maybe we should refactor them to both use the same internal method `_createNewBytes(uint size)` to emphasize that the logic is the same

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -257,22 +245,18 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-                relayId: numberOfRelays,
-                relayState: RelayState.Pending,
-                priceRequestTime: priceRequestTime,
-                realizedLpFeePct: realizedLpFeePct,
-                slowRelayer: msg.sender,
-                instantRelayer: relays[depositHash].instantRelayer
-            });
-        relays[depositHash] = relayData;
+        RelayData storage relayData = relays[depositHash];
+        relayData.relayId = numberOfRelays++;
```
chrismaree, line 249: this post increment is pretty sneaky! I'd say a comment is needed to make it clear what this is doing.

packages/core/contracts/common/implementation/AncillaryData.sol
chrismaree: also, some comments would be good! if you wrote this then awesome else we should reference where you got it from.

mrice32: @chrismaree agreed. I just copied from the other implementation in the file. Will add comments.

@nicholaspai it might be a bit harder to generalize because I had to change the size of the uint to cast to. Will think about this a little more tho.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: +1 good call

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: I think it saves a SLOAD (aka a read).

Before:
```
Relay: 545882
```

After:
```
Relay: 543100
```

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Making this an "append" method costs about 400 gas (200 for relay and 200 for settle). Thoughts?

packages/core/contracts/common/implementation/AncillaryData.sol
mrice32: Done.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Done!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: I added this, but we might reconsider re-removing the append if we want to make this a bit cheaper.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Yeah, it should be the same cost, since we just removed variables from the event and put them into the struct.

chrismaree:
Strange. this is now failing in some other packages that relay on the `getContractVersion` code. it might be because you modified `packages/core/contracts/common/implementation/AncillaryData.sol` which has changed the downstream bytecode of the other contracts but the tests should have accommodated this with the `findContractVersion`. however, this util _might_ be broken after the most recent hardhat/truffle migration.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -35,7 +35,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     IERC20 public override l1Token;
 
     // Track the total number of relays and uniquely identifies relays.
-    uint256 public numberOfRelays;
+    uint32 public numberOfRelays;
```
chrismaree, line 38: what? you dont think we are going to have `2^(256)-1` relays?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -75,12 +75,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
     struct RelayData {
-        uint256 relayId;
         RelayState relayState;
-        uint256 priceRequestTime;
-        uint64 realizedLpFeePct;
         address slowRelayer;
+        uint32 relayId;
+        uint64 realizedLpFeePct;
         address instantRelayer;
+        uint256 priceRequestTime;
```
chrismaree, line 83: is it not better to put all the `uintn` next to each other in the `struct`? also, put. the `address` together? I know with params this saves gas but not sure about struct values?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -91,31 +91,16 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        address indexed caller
+        bytes32 indexed ancillaryDataHash
     );
-
-    modifier onlyFromOptimisticOracle() {
-        require(msg.sender == address(_getOptimisticOracle()), "Caller must be OptimisticOracle");
-        _;
-    }
+    event RelaySpedUp(bytes32 indexed depositHash, address indexed instantRelayer);
+    event RelaySettled(bytes32 indexed depositHash, bytes32 indexed relayHash, address indexed caller);
```
chrismaree, line 103: much cleaner

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -281,15 +266,23 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         // instead of default setting to equal to the `depositTimestamp`, which is dependent on the L2 VM on which the
         // DepositContract is deployed. Imagine if the timestamps on the L2 have an offset that are always "in the
         // future" relative to L1 blocks, then the OptimisticOracle would always reject requests.
-        _requestOraclePriceRelay(amount, priceRequestTime, getRelayAncillaryData(depositData, relayData));
-        _proposeOraclePriceRelay(amount, priceRequestTime, getRelayAncillaryData(depositData, relayData));
+        bytes32 relayHash = _getRelayHash(depositData, relayData);
+        bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
+        _requestOraclePriceRelay(amount, priceRequestTime, ancillaryData);
+        _proposeOraclePriceRelay(amount, priceRequestTime, ancillaryData);
```
chrismaree, line 272: looking forward to collapsing these in a later PR

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -1,6 +1,6 @@
 const { assert } = require("chai");
 const hre = require("hardhat");
-const { web3 } = require("hardhat");
+const { web3 } = hre;
```
chrismaree, line 3: I think we do this in a few places wrong! I'll send out a quick PR to update the repo for anywhere we do this incorrectly.

chrismaree: PR: https://github.com/UMAprotocol/protocol/pull/3418

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Yeah i guess if there's no starting ancillary data to append to, then no need to append! Let's revert the append.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -391,7 +384,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         updateAccumulatedLpFees();
         allocateLpFees(_getAmountFromPct(relay.realizedLpFeePct, _depositData.amount));
 
-        emit RelaySettled(depositHash, keccak256(getRelayAncillaryData(_depositData, relay)), msg.sender);
+        emit RelaySettled(depositHash, relayHash, msg.sender);
+
+        delete relay.realizedLpFeePct;
```
nicholaspai, line 389: Why delete if you can't do anything with a `finalized` relay?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -690,28 +671,4 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
             1e18 // Canonical value representing "True"; i.e. the proposed relay is valid.
         );
     }
-
-    function _emitDepositRelayedEvent(
```
nicholaspai, line 694: +1 I only added this to beat the stack too deep compile issue, which seems to have disappeared

nicholaspai (approved):
LGTM once you resolve chris' comments and the tests pass!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Yeah, the struct helped a lot with that.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Gas + being a good ETH citizen and reducing overall state bloat :)

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Yeah, I reordered them a little bit to try and 1) reduce the amount of space this takes up 2) to reduce the number of `delete` calls I need to make in the settlement function to clear a slot. There are probably multiple orderings that would do this, but we can't put the addresses together without causing solidity to create an extra slot I don't think.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: :)

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: This saves gas...?

PR #3409, opened 2021-09-30
feat(optimistic-oracle): Introduce gas-lite version of OptimisticOracle

nicholaspai:
**Motivation**

Can be used by financial contracts (such as BridgePool) that want to make optimistic price requests secured by the DVM but are willing to make tradeoffs in order to reduce gas costs for making external calls.

**Summary**

Gas savings are summarized and based on unit tests. In parantheses are listed minimum, maximum, and average gas cost.
- `requestPrice`:
    - `OptimisticOracle`: (208504, 534950, 252616)
    - `OptimisticOracleLite`: (130000, 416000, 160000)
- `proposePrice`:
    - `OptimisticOracle`: (184820, 207461, 192174)
    - `OptimisticOracleLite`: (68000, 87000, 75000)
- `settle`:
    - `OptimisticOracle`: (126667, 262749, 179783)
    - `OptimisticOracleLite`: (63000, 163000, 117000)

New Features:
- In `OptimisticOracleLite`, `setBond`/`setCustomLiveness` logic is rolled into the `requestPrice` logic.
- Callbacks after `proposePrice`, `disputePrice` and `settle` to the `requester` address are removed.
- `refundAfterDispute` logic is removed
- Helper method `requestAndProposePriceFor` combines logic of `requestPrice` and `proposePrice` and calling it saves 30% gas compared to calling the two sequentially as two separate external calls.
- 
**Details**

The main gas savings is changing the internal mapping of `requests` from storing the entire `Request` struct to storing a hash of the `Request` struct. This means that all of the external functions require more parameters, therefore off-chain client code will be more complex.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -159,7 +159,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         IERC20 currency,
         uint256 reward
     ) external override nonReentrant() returns (uint256 totalBond) {
-        require(_getState(msg.sender, identifier, timestamp, ancillaryData) == State.Invalid, "requestPrice: Invalid");
```
nicholaspai, line 162: These are minor changes we can make to the `OptimisticOracle` that saves 10,000 gas. I can remove since its probably not worth it.

chrismaree: I'd say rather not seeing that would mean re-deploying the OO to get things up to the latest version in production.

nicholaspai: Reverted

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
```diff
@@ -0,0 +1,240 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "../interfaces/OptimisticOracleInterface.sol";
+
+/**
+ * @title Interface for the gas-cost-reduced version of the OptimisticOracle
```
chrismaree, line 8:
```suggestion
 * @title Interface for the gas-cost-reduced version of the OptimisticOracle.
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,658 @@
...
+                    _request.currency,
+                    _request.settled,
+                    _request.proposedPrice,
+                    _request.resolvedPrice,
+                    _request.expirationTime,
+                    _request.reward,
+                    _request.finalFee,
+                    _request.bond,
+                    _request.customLiveness
+                )
```
mrice32, line 502: Not really a gas optimization, but would this be simpler?
```
return keccak256(abi.encode(_request));
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,658 @@
...
+     * @param _proposedPrice price being proposed.
+     * @return totalBond the amount that's pulled from the caller's wallet as a bond. The bond will be returned to
+     * the proposer once settled if the proposal is correct.
+     */
+    function proposePriceFor(
+        address _requester,
+        bytes32 _identifier,
+        uint256 _timestamp,
+        bytes memory _ancillaryData,
+        Request memory _request,
```
mrice32, line 157: Not a gas optimization, but was curious if you think it would be simpler to bundle these params into a struct:
```
        address _requester,
        bytes32 _identifier,
        uint256 _timestamp,
        bytes memory _ancillaryData,
        Request memory _request,
```

I only say this because they're passed around a lot and emitted together in a bunch of events. Definitely not strongly opinionated though.

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
```diff
@@ -0,0 +1,240 @@
...
+    // Struct representing a price request. Note that this differs from the OptimisticOracleInterface's Request struct
+    // in that refundOnDispute is removed.
+    struct Request {
+        address proposer; // Address of the proposer.
+        address disputer; // Address of the disputer.
+        IERC20 currency; // ERC20 token used to pay rewards and fees.
+        bool settled; // True if the request is settled.
+        int256 proposedPrice; // Price that the proposer submitted.
+        int256 resolvedPrice; // Price resolved once the request is settled.
+        uint256 expirationTime; // Time at which the request auto-settles without a dispute.
```
mrice32, line 21: If we do change this to uint32, we should move it next to the boolean above, so it can be packed tightly and save some bytes.

mrice32 (approved):
This is awesome! Very well done!

mrice32:
@nicholaspai the gas numbers on this are crazy good.

I just want to present them in a different way so we can see the total savings:
Request: `252616 - 157794 = 94822`
Propose:  `192174 - 71901 = 120273`
Settle: `179783 - 117037 = 62746`

Total: `277841`!!!!

If all goes right (optimistically), that should get us to around 632k total (although I suspect we will end up with a few extra complications in the caller that will bump that a little). Wild stuff!!

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,658 @@
...
+            "Ancillary Data too long"
+        );
+        uint256 finalFee = _getStore().computeFinalFee(address(_currency)).rawValue;
+
+        // Associate new request with ID
+        Request memory request;
+        request.currency = _currency;
+        request.reward = _reward;
+        request.finalFee = finalFee;
+        request.bond = _bond != 0 ? _bond : finalFee;
```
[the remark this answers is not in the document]
nicholaspai: Do you think we could instead just use whatever bond the user provides, even if it's less than the final fee?

nicholaspai:
I just realized that there’s no trivial way to (1) determine from a requesting contract if a request was disputed (i.e. the `BridgePool` needs to know this) and (2) its difficult to settle a price request because the price request might have been disputed and the requesting contract doesn’t know who the disputer was.

I think we need to re-implement the OO dispute + settle callbacks so that requesting contract can track the state of requests and determine once they have been settled (and fetch its `resolvedPrice`)

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
nicholaspai: I changed this line to just set `request.bond = _bond` which I believe is consistent with original OO

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
nicholaspai: Would we also need to change the order of function params to have uint32's next to each other?

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,658 @@
...
+ */
+contract SkinnyOptimisticOracle is SkinnyOptimisticOracleInterface, Testable, Lockable {
+    using SafeMath for uint256;
+    using SafeERC20 for IERC20;
+    using Address for address;
+
+    event RequestPrice(
+        address indexed requester,
+        bytes32 indexed identifier,
+        uint256 timestamp,
```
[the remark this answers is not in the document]
nicholaspai: This line is giving me trouble, how would you convert `uint256` returned by `getCurrentTime` to `uint32`?

```
expirationTime: uint32(getCurrentTime()).add(
                    _request.customLiveness != 0 ? _request.customLiveness : defaultLiveness
                ), // Modified
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
nicholaspai: Yeah definitely more readable, fixed

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
mrice32: Hmmm, yeah, if we have to do a bunch of unsafe conversions, it may not be worth it.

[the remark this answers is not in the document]
mrice32: Oh wait, why the .add rather than just +?

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
```diff
@@ -0,0 +1,240 @@
...
+ * @title Interface for the gas-cost-reduced version of the OptimisticOracle.
+ * @dev Interface used by financial contracts to interact with the Oracle. Voters will use a different interface.
+ */
+abstract contract SkinnyOptimisticOracleInterface {
+    // Struct representing a price request. Note that this differs from the OptimisticOracleInterface's Request struct
+    // in that refundOnDispute is removed.
+    struct Request {
+        address proposer; // Address of the proposer.
+        address disputer; // Address of the disputer.
+        IERC20 currency; // ERC20 token used to pay rewards and fees.
```
chrismaree, line 17: is there any gas difference with storing this as an address vs an interface like this?

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
chrismaree: I think keeping them separate as params makes the interface a bit easier to work with externally personally.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+    function proposePriceFor(
+        address _requester,
+        bytes32 _identifier,
+        uint256 _timestamp,
+        bytes memory _ancillaryData,
+        Request memory _request,
+        address _proposer,
+        int256 _proposedPrice
+    ) public override nonReentrant() returns (uint256 totalBond) {
+        require(_proposer != address(0), "proposer address must be non 0");
```
chrismaree, line 212:
```suggestion
        require(_proposer != address(0), "Proposer address must be non 0");
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+        require(
+            _getState(_requester, _identifier, _timestamp, _ancillaryData, _request) ==
+                OptimisticOracleInterface.State.Requested,
+            "proposePriceFor: Requested"
```
chrismaree, line 216: this is a different `require` style to some of the other methods. should we convert on something more consistant?

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+        request.bond = _bond;
+        request.customLiveness = _customLiveness;
+        request.proposer = _proposer;
+        request.proposedPrice = _proposedPrice;
+        request.expirationTime = getCurrentTime().add(_customLiveness != 0 ? _customLiveness : defaultLiveness);
+
+        // Pull reward from requester, who is the caller.
+        if (_reward > 0) {
+            _currency.safeTransferFrom(msg.sender, address(this), _reward);
+        }
```
chrismaree, line 335:
```suggestion
        if (_reward > 0) _currency.safeTransferFrom(msg.sender, address(this), _reward);
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+        // Pull proposal bond from caller.
+        totalBond = request.bond.add(request.finalFee);
+        if (totalBond > 0) {
+            _currency.safeTransferFrom(msg.sender, address(this), totalBond);
+        }
```
chrismaree, line 340:
```suggestion
        if (totalBond > 0) _currency.safeTransferFrom(msg.sender, address(this), totalBond);
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+    }
+
+    /**
+     * @notice Disputes a price request with an active proposal on another address' behalf. Note: this address will
+     * receive any rewards that come from this dispute. However, any bonds are pulled from the caller.
+     * @param _identifier price identifier to identify the existing request.
+     * @param _timestamp timestamp to identify the existing request.
+     * @param _ancillaryData ancillary data of the price being requested.
+     * @param _request price request parameters whose hash must match the request that the caller wants to
+     * dispute.
```
chrismaree, line 360:
```suggestion
     *              dispute.
     
```

or try get it under 120 chars on one line.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+                bond: _request.bond,
+                customLiveness: _request.customLiveness
+            });
+
+        uint256 finalFee = disputedRequest.finalFee;
+        uint256 bond = disputedRequest.bond;
+        totalBond = bond.add(finalFee);
+        if (totalBond > 0) {
+            disputedRequest.currency.safeTransferFrom(msg.sender, address(this), totalBond);
+        }
```
chrismaree, line 404:
```suggestion
        if (totalBond > 0) disputedRequest.currency.safeTransferFrom(msg.sender, address(this), totalBond);
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+    }
+
+    /**
+     * @notice Disputes a price request with an active proposal where caller is the disputer.
+     * @param _requester sender of the initial price request.
+     * @param _identifier price identifier to identify the existing request.
+     * @param _timestamp timestamp to identify the existing request.
+     * @param _ancillaryData ancillary data of the price being requested.
+     * @param _request price request parameters whose hash must match the request that the caller wants to
+     * dispute.
```
chrismaree, line 443:
```suggestion
     *             dispute.
```
or try get it under 120 chars on one line.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+    }
+
+    /**
+     * @notice Attempts to settle an outstanding price request. Will revert if it isn't settleable.
+     * @param _requester sender of the initial price request.
+     * @param _identifier price identifier to identify the existing request.
+     * @param _timestamp timestamp to identify the existing request.
+     * @param _ancillaryData ancillary data of the price being requested.
+     * @param _request price request parameters whose hash must match the request that the caller wants to
+     * settle.
```
chrismaree, line 465: same nit

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+     */
+    function stampAncillaryData(bytes memory _ancillaryData, address _requester)
+        public
+        pure
+        override
+        returns (bytes memory)
+    {
+        return _stampAncillaryData(_ancillaryData, _requester);
+    }
+
```
chrismaree, line 541: do you want to add a comment block to show you are moving into private functions?
```suggestion
    /****************************************
     *    PRIVATE AND INTERNAL FUNCTIONS    *
     ****************************************/
```

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+                finalFee: _request.finalFee,
+                bond: _request.bond,
+                customLiveness: _request.customLiveness
+            });
+
+        OptimisticOracleInterface.State state =
+            _getState(_requester, _identifier, _timestamp, _ancillaryData, _request);
+        if (state == OptimisticOracleInterface.State.Expired) {
+            // In the expiry case, just pay back the proposer's bond and final fee along with the reward.
+            resolvedPrice = settledRequest.proposedPrice;
```
chrismaree, line 591: question: is there a gas difference between using `settledRequest.proposedPrice` and `request.proposedPrice`?

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+            resolvedPrice = _getOracle().getPrice(
+                _identifier,
+                _timestamp,
+                _stampAncillaryData(_ancillaryData, _requester)
+            );
+            settledRequest.resolvedPrice = resolvedPrice;
+            bool disputeSuccess = settledRequest.resolvedPrice != settledRequest.proposedPrice;
+            uint256 bond = settledRequest.bond;
+
+            // Unburned portion of the loser's bond = 1 - burned bond.
```
chrismaree, line 606: I feel that this comment does not match to the "1-" part. the bond is not "1"? I know this comment is the same in the main OO but still could be improved.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+        } else if (state == OptimisticOracleInterface.State.Resolved) {
+            // In the Resolved case, pay either the disputer or the proposer the entire payout (+ bond and reward).
+            resolvedPrice = _getOracle().getPrice(
+                _identifier,
+                _timestamp,
+                _stampAncillaryData(_ancillaryData, _requester)
+            );
+            settledRequest.resolvedPrice = resolvedPrice;
+            bool disputeSuccess = settledRequest.resolvedPrice != settledRequest.proposedPrice;
+            uint256 bond = settledRequest.bond;
```
chrismaree, line 604: nit: you can remove this variable and just reference `settledRequest.bond` on lines 607 and 614 thereby saving the declaration of a variable.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,716 @@
...
+        if (_request.settled) {
+            return OptimisticOracleInterface.State.Settled;
+        }
+
+        if (_request.disputer == address(0)) {
+            return
+                _request.expirationTime <= getCurrentTime()
+                    ? OptimisticOracleInterface.State.Expired
+                    : OptimisticOracleInterface.State.Proposed;
+        }
```
chrismaree, line 682:
```suggestion
        if (address(_request.currency) == address(0)) return OptimisticOracleInterface.State.Invalid;

        if (_request.proposer == address(0)) return OptimisticOracleInterface.State.Requested;

        if (_request.settled) return OptimisticOracleInterface.State.Settled;

        if (_request.disputer == address(0)) return _request.expirationTime <= getCurrentTime()
                    ? OptimisticOracleInterface.State.Expired
                    : OptimisticOracleInterface.State.Proposed;
```

this feels more "skinny" to me. might need to be linted to pass CI.

chrismaree (approved):
This is awesome awesome stuff! I really like this implementation. It would be useful for us to document the differences between the "normal" and "skinny" OOs and what the restrictions are between the two. this could perhaps go in the readme file for `core` or the docs site?

My review is mainly nits/styling and commenting related.

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
nicholaspai: Not sure, but i kept this to maintain as much consistency with the OO implementation as possible

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
nicholaspai: Yeah do you prefer having the function name in the error message? I see this sometimes used in the OO @mrice32

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
nicholaspai: Fixed!

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
mrice32: Tbh, it's not a huge deal either way. We just want the error messages to be as unique as possible so a caller can find the line of code where it was emitted.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
mrice32: Sorta, except that it doesn't allow a default. Basically, it can't match the original OO because there was always the option of _not_ customizing the bond. Thinking about this more, I liked the way you were originally doing it. I think a default is preferable to just allowing 0.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
mrice32: SGTM

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
mrice32: I looked into the amount of gas this would save and it's not a huge amount unless you're actually storing the struct. We can take a look at it later, but I don't think it's going to impact the bottom line too much. Feel free to ignore!

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
nicholaspai: Looks like it saves < 100 gas i'll make the change though

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
nicholaspai: I think allowing 0 is fine to be honest, I'd rather give caller more flexibility. Setting a 0 bond means that the request is more likely to get proposed an incorrect price @chrismaree wdyt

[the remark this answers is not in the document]
chrismaree: I think 0 could be problematic as it means there is no cost with bad proposals. however, how do you choose a reasonable default for all currencies, especially given different decimals for different erc20s?

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
nicholaspai: Ah this was only an issue because I made the livenss params also 32 bit. I might not do that since we perform math on them

packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
chrismaree: ok cool.

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -0,0 +1,699 @@
...
+                finalFee: _request.finalFee,
+                bond: _request.bond,
+                customLiveness: _request.customLiveness
+            });
+        _storeRequestHash(requestId, proposedRequest);
+
+        totalBond = _request.bond.add(_request.finalFee);
+        if (totalBond > 0) {
+            _request.currency.safeTransferFrom(msg.sender, address(this), totalBond);
+        }
```
chrismaree, line 241:
```suggestion
        if (totalBond > 0) _request.currency.safeTransferFrom(msg.sender, address(this), totalBond);
```

chrismaree (approved):
Thank you for implementing outstanding nits! I left one tiny other nit on formatting. Other than that this looks great!

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
nicholaspai: Yup I made all the changes!

packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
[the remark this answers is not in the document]
nicholaspai: But with the OO you can set `bond` to be 0 via `setCustomLiveness`. The only difference here is that there is not default. The old default was the final fee for the erc20 so choosing for each token is easy.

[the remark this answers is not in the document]
mrice32: @nicholaspai I'm a fan of your original implementation `request.bond = _bond != 0 ? _bond : finalFee;`. If people really want 0, they can do 1 wei (they have to pay the final fee anyway, so it won't be like the proposer is actually paying nothing).

mrice32 (approved):
LGTM! This is awesome @nicholaspai!

PR #3412, opened 2021-09-30
feat(across): Update the bridge pool to deal with the scenario where tokens are not relayed before they come through the canonical bridge

chrismaree:
**Motivation**

It is possible for a bridging action to conclude via that caoninical bridge before the relay action (over across) is finalized. This could happen due to the depositor underpricing their fees and no relayer picks it up and then they bridge it themselves after the duration or if there is no liquidity in the pool. In either case, the pool should correctly account for this in the exchange rate (which embodies the fees & pending/liquid and utilized reserves).

**Summary**

It turns out that the implementation right now actually deals with this problem perfectly, as verified in the unit tests. Therefore this PR makes no contract changes and simply updates the tests to show the contracts can deal with this case.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


nicholaspai:
I just realized that there’s no trivial way to (1) determine from a requestin contract if a request was disputed (i.e. the `BridgePool` needs to know this) and (2) its difficult to settle a price request because the price request might have been disputed and the requesting contract doesn’t know who the disputer was.

I think we need to re-implement the OO dispute + settle callbacks so that requesting contract can track the state of requests and determine once they have been settled (and fetch its `resolvedPrice`)

mrice32:
> It turns out that the implementation right now actually deals with this problem perfectly, as verified in the unit tests. Therefore this PR makes no contract changes and simply updates the tests to show the contracts can deal with this case.

Lol @chrismaree this is a great outcome. Nice work!

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -1431,4 +1429,100 @@ describe("BridgePool", () => {
...
+      // without introducing a step in the rate at any point.
+
+      // Advance time by 1 week past the end of the of the L2->L1 liveness period.
+      await advanceTime(604800);
+      assert.equal((await bridgePool.methods.exchangeRateCurrent().call()).toString(), toWei("1"));
+
+      // Now, simulate the finalization of of the bridge action by the canonical bridge by minting tokens to the pool.
+      await l1Token.methods.mint(bridgePool.options.address, toWei("100")).send({ from: owner });
+
+      // The exchange rate should not have updated.
```
mrice32, line 1453: Could you briefly explain how this is possible? Looking at the math, it was a little difficult for me to figure out.

I think it's because utilizedReserves goes _negative_, which drags down the numerator by the same amount of unexpected tokens dumped on the contract, but I want to make sure your intuition matches mine.

mrice32 (approved):
Very cool! I like that we're testing various stages and not just the one we previously discussed.

packages/core/test/insured-bridge/BridgePool.js
chrismaree: yup! that maps exactly to why I think this works. It's all about the negative `utilizedReserves` that captures this perfectly. when we discussed this IRL we neglected this fact :)

nicholaspai (approved):
Nice tests! Thanks for illustrating this case

packages/core/test/insured-bridge/BridgePool.js
mrice32: 👌

PR #3413, opened 2021-09-30
feat(across): add re entrancy guards

chrismaree:
**Motivation**

adding of liquidity should have re-entrancy guards.

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -153,7 +153,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
      * @dev Reentrancy guard not added to this function because this indirectly calls sync() which is guarded.
      * @param l1TokenAmount Number of l1Token to add as liquidity.
      */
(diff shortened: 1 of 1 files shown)


nicholaspai (approved):

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -153,7 +153,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
      * @dev Reentrancy guard not added to this function because this indirectly calls sync() which is guarded.
      * @param l1TokenAmount Number of l1Token to add as liquidity.
      */
-    function addLiquidity(uint256 l1TokenAmount) public {
+    function addLiquidity(uint256 l1TokenAmount) public nonReentrant() {
```
nicholaspai, line 156: Ah I left these off on purpose because they're already reentrancy-guarded by `exchangeRateCurrent()`

chrismaree: great point @nicholaspai ! This can be seen by the failing unit tests as well. I'm going to close off this PR.

PR #3414, opened 2021-09-30
feat(sdk): add bridge deposit box client

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/2411/add-bridge-deposit-box-client-to-sdk

**Summary**

Add a basic client and event decoder to sdk for bridge deposit box. Needed to support frontend across work.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2411/add-bridge-deposit-box-client-to-sdk

### DIFF
--- a/packages/sdk/src/clients/bridgeDepositBox/client.ts
+++ b/packages/sdk/src/clients/bridgeDepositBox/client.ts
@@ -1,0 +1,111 @@
+import { BridgeDepositBoxEthers, BridgeDepositBoxEthers__factory } from "@uma/contracts-node";
+import type { SignerOrProvider, GetEventType } from "../..";
+import { Event, BigNumber } from "ethers";
+
+export type Instance = BridgeDepositBoxEthers;
+const Factory = BridgeDepositBoxEthers__factory;
+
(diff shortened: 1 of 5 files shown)


amateima (approved):

packages/sdk/src/clients/bridgeDepositBox/client.ts
```diff
@@ -0,0 +1,114 @@
...
+            amount,
+            slowRelayFeePct,
+            instantRelayFeePct,
+            quoteTimestamp,
+          },
+        },
+      };
+    }
+    default: {
+      throw new Error("Unknown event " + event.event);
```
amateima, line 107: Is it ok to throw the error here? I know it's the client's job to handle it, but I'm wondering if we can log it silently or smth similar.

daywiss: yea good point. we can probably silently ignore errors, it would just be difficult to know if we ever miss an event. im not sure what to do here actually

daywiss: i removed the error

PR #3418, opened 2021-10-01
fix(hardhat): update HRE syntax

chrismaree:
**Motivation**

There are some cases where we should use the `const { web3 } = hre;` (see @mrice32 's PR here: https://github.com/UMAprotocol/protocol/pull/3406#discussion_r719995423 where the changes were made on another PR)

**Summary**

Update the case where we already import `hre` to more correctly import `web3`. Note that there are still many files that use the old syntax. This is the case if we don't first import `const hre = require("hardhat");`. For example, see here: https://github.com/UMAprotocol/protocol/blob/8773494d29cf0428ca6d65f0272b135ba3dafcbf/packages/financial-templates-lib/test/price-feed/CryptoWatchPriceFeed.js#L1

**Testing**

- [ ]  New unit tests created
- [X]  Untested


nicholaspai (approved):
LGTM thanks!

mrice32 (approved):

PR #3420, opened 2021-10-01
improve(sdk): refactor fee calc to use ethers and bignumber

daywiss:
**Motivation**

https://app.shortcut.com/uma-project/story/2406/refactor-across-fees-in-sdk-to-remove-web3-dependency

**Summary**

Was able to switch to using ethers native bignumber and utilities, ditching both web3 and bn.js. Was not able to remove the decimal.js dependency, since whats available in ethers (fixednumber) does not have `pow`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2406/refactor-across-fees-in-sdk-to-remove-web3-dependency

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -38,11 +38,9 @@
     "@uma/contracts-frontend": "^0.1.2",
     "@uma/contracts-node": "^0.1.2",
     "axios": "^0.21.4",
-    "bn.js": "^5.2.0",
     "decimal.js": "^10.3.1",
     "highland": "^2.13.5",
-    "lodash-es": "^4.17.21",
-    "web3": "^1.5.3"
+    "lodash-es": "^4.17.21"
   },
   "peerDependencies": {
     "ethers": "^5.4.2"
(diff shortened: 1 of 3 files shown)


packages/sdk/package.json
```diff
@@ -42,11 +42,9 @@
     "@uma/contracts-frontend": "^0.1.2",
     "@uma/contracts-node": "^0.1.2",
     "axios": "^0.21.4",
-    "bn.js": "^5.2.0",
     "decimal.js": "^10.3.1",
     "highland": "^2.13.5",
-    "lodash-es": "^4.17.21",
-    "web3": "^1.5.3"
+    "lodash-es": "^4.17.21"
```
chrismaree, line 47: Why do you need this?

chrismaree (approved):
This is excellent! I really like the refactor. We use ethers in a few places as well so removing web3 for this is a big win. Also, I've been having issues with the new version of BN in the latest web3 version. Hopefully this sorts it out with that as well.

packages/sdk/package.json
daywiss: it wasnt added, the line was just modified, but lodash is used other places in the sdk. i think in multicall libraries

PR #3425, opened 2021-10-03
fix(insured-bridge): Instant relayer should only receive refund iff they sped up valid relay

nicholaspai:
**Motivation**

If an instant relayer underpays a recipient, perhaps because the realizedLpFeePct specified by the Relay is set too high, then currently they will receive the full relay amount when the Relay eventually settles. They will net the amount that they underpaid.

We need a way to keep track of whether instant relayers paid the correct amount, and to make sure that users always receive the correct relayed amount.

**Summary**

This PR introduces a new mapping of hashed `{depositHash, relay.realizedLpFeePct}` to `instantRelayer` address. Whenever a relay is proposed for a deposit, this `instantRelayHash` is deterministically linked to it and allows any `instantRelayer` to speed up the relay. If the relay params end up getting finalized, then the instant relayer who sped up those exact relay params will get refunded. If the instant relayer sped up an incorrect relay that ended up getting disputed, then their address won't be linked to a subsequent valid relay unless they speed it up again.

In either case, the responsibility still lies on the `instantRelayer` to validate the relay off-chain before speeding it up. 

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -100,7 +100,8 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         address slowRelayer,
         address l1Token,
         uint64 realizedLpFeePct,
-        bytes32 indexed depositHash
+        bytes32 indexed depositHash,
+        bytes32 indexed relayHash
     );
     event RelaySpedUp(bytes32 indexed depositHash, address indexed instantRelayer, uint64 realizedLpFeePct);
     event RelaySettled(bytes32 indexed depositHash, bytes32 indexed relayHash);
@@ -390,7 +391,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         updateAccumulatedLpFees();
         allocateLpFees(_getAmountFromPct(relay.realizedLpFeePct, _depositData.amount));
 
-        emit RelaySettled(depositHash, relayHash, msg.sender);
+        emit RelaySettled(depositHash, relayHash);
 
         delete relay.realizedLpFeePct;
         delete relay.instantRelayer;
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -368,8 +368,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
                 _getAmountFromPct(relay.realizedLpFeePct + _depositData.slowRelayFeePct, _depositData.amount);
 
         // Refund the instant relayer iff the instant relay params match the approved relay.
-        bytes32 instantRelayHash = keccak256(abi.encode(depositHash, relay.realizedLpFeePct));
-        address instantRelayer = instantRelays[instantRelayHash];
+        address instantRelayer = instantRelays[keccak256(abi.encode(depositHash, relay.realizedLpFeePct))];
 
         l1Token.safeTransfer(
             instantRelayer != address(0) ? instantRelayer : _depositData.l1Recipient,
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -323,7 +323,6 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
 
         l1Token.safeTransferFrom(msg.sender, _depositData.l1Recipient, _depositData.amount - feesTotal);
 
-        // TODO: does this need more info?
         emit RelaySpedUp(depositHash, msg.sender, relay.realizedLpFeePct);
     }
 
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -390,6 +390,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
 
         emit RelaySettled(depositHash, relayHash, msg.sender);
 
+        delete instantRelays[keccak256(abi.encode(depositHash, relay.realizedLpFeePct))];
         delete relay.realizedLpFeePct;
         delete relay.priceRequestTime;
     }
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -80,13 +80,16 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         uint256 priceRequestTime;
         uint64 realizedLpFeePct;
         address slowRelayer;
-        address instantRelayer;
     }
 
     // Associate deposits with pending relay data. When RelayState is Uninitialized, new relay attempts can be made for
     // this deposit. Contains information necessary to pay out relayers on successful relay. Deposits get reset to the
     // "Uninitialized" state when they are disputed on the OptimisticOracle.
     mapping(bytes32 => RelayData) public relays;
+
+    // Map hash of deposit and realized-relay fee to instant relayers. This mapping is checked at settlement time
+    // to determine if there was a valid instant relayer.
+    mapping(bytes32 => address) public instantRelays;
 
     event LiquidityAdded(address indexed token, uint256 amount, uint256 lpTokensMinted, address liquidityProvider);
     event LiquidityRemoved(address indexed token, uint256 amount, uint256 lpTokensBurnt, address liquidityProvider);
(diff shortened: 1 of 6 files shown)


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -322,7 +324,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         l1Token.safeTransferFrom(msg.sender, _depositData.l1Recipient, _depositData.amount - feesTotal);
 
         // TODO: does this need more info?
```
chrismaree, line 326: you can get rid of this todo I think

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -366,8 +368,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
             _depositData.amount -
                 _getAmountFromPct(relay.realizedLpFeePct + _depositData.slowRelayFeePct, _depositData.amount);
 
+        // Refund the instant relayer iff the instant relay params match the approved relay.
+        bytes32 instantRelayHash = keccak256(abi.encode(depositHash, relay.realizedLpFeePct));
+        address instantRelayer = instantRelays[instantRelayHash];
```
chrismaree, line 373:
```suggestion
        address instantRelayer = instantRelays[keccak256(abi.encode(depositHash, relay.realizedLpFeePct))];
```

will this use less gas?

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -545,8 +563,156 @@ describe("BridgePool", () => {
...
+      // Instant relayer address should be empty for most recent relay with valid params.
+      const instantRelayHash = generateInstantRelayHash(depositHash, relayData);
+      assert.equal(await bridgePool.methods.instantRelays(instantRelayHash).call(), ZERO_ADDRESS);
+
+      // Can speed up relay with new (valid) params.
+      await l1Token.methods.mint(instantRelayer, instantRelayAmountSubFee).send({ from: owner });
+      await l1Token.methods
+        .approve(bridgePool.options.address, instantRelayAmountSubFee)
+        .send({ from: instantRelayer });
+      await bridgePool.methods.speedUpRelay(depositData).send({ from: instantRelayer });
```
chrismaree, line 675: this test case has the initial (invalid) instant relayer and subsequent (valid) instant relayer as the same EOA address. it might be a good idea to have a separate test case such that these are different EOAs. I'm pretty sure this will behave exactly the same but it won't hurt to validate it, especially given that is a likely flow in production.

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -545,8 +563,156 @@ describe("BridgePool", () => {
...
+        totalDisputeRefund.toString()
+      );
+
+      // BridgePool should have balance reduced by full amount sent to to l1Recipient.
+      assert.equal(
+        (await l1Token.methods.balanceOf(bridgePool.options.address).call()).toString(),
+        toBN(initialPoolLiquidity)
+          .sub(toBN(instantRelayAmountSubFee).add(realizedInstantRelayFeeAmount).add(realizedSlowRelayFeeAmount))
+          .toString()
+      );
```
chrismaree, line 715: this test is really good and showcases the flow well.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -216,6 +237,13 @@ export class InsuredBridgeL1Client {
     this.logger.debug({ at: "InsuredBridgeL1Client", message: "Insured bridge l1 client updated" });
   }
 
+  private _getInstantRelayHash(depositHash: string, realizedLpFeePct: string): string | null {
+    const instantRelayDataHash = this.web3.utils.soliditySha3(
```
chrismaree, line 241: do you want to add `soliditySha3` to the top set of `const { toBN } = Web3.utils;` to clean this up a little bit?

same with `eth.abi.encodeParameters`

packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
```diff
@@ -380,7 +379,7 @@ describe("InsuredBridgeL1Client", function () {
       await bridgePool.methods.settleRelay(depositData).send({ from: rando });
 
       await client.update();
-      expectedRelayedDepositInformation.relayState = 3; // settled
+      expectedRelayedDepositInformation.relayState = ClientRelayState.Finalized;
```
chrismaree, line 382: 👍

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -94,6 +99,16 @@ export class InsuredBridgeL1Client {
     return this.relays[l1Token][deposit.depositHash];
   }
 
+  hasInstantRelayer(l1Token: string, depositHash: string, realizedLpFeePct: string): boolean {
```
chrismaree, line 102: nit: why are you taking in `realizedLpFeePct`? it might be better to pass this around as a `BN`, even though you are casing it to a hash later on. Reasoning is mainly just to maintain consistency with the rest of numbers in this client, implementation and other downstream usage of `Pct` variables.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -170,7 +181,11 @@ export class Relayer {
 
     // b) Balance is large enough to instant relay. Deposit is in any state except finalized (i.e can be slow relayed
```
chrismaree, line 182:
```suggestion
    // b) Balance is large enough to instant relay and the relay does not have an instant relayer. Deposit is in any state except finalized (i.e can be slow relayed
```

chrismaree: is this always the case? what if the deposit was relayed, sped up, and disputed. can the bot not then consider this to re-speedup?

packages/insured-bridge-relayer/test/Relayer.ts
```diff
@@ -291,32 +291,38 @@ describe("Relayer.ts", function () {
...
-        await relayer.shouldRelay(deposit, clientRelayState, toBN(defaultRealizedLpFeePct)),
+        await relayer.shouldRelay(deposit, clientRelayState, toBN(defaultRealizedLpFeePct), false),
         ShouldRelay.Instant
       );
 
-      // Modifying any of the above 3 conditions should make the bot not instant relay.
+      // Modifying any of the above 4 conditions should make the bot not instant relay.
 
-      // a) ClientRelayState set to SpeedUpOnly
+      // a) There  already exists an instant relayer for this relay.
```
chrismaree, line 300:
```suggestion
      // a) There already exists an instant relayer for this relay.
```

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Barely, but its more concise.

packages/core/test/insured-bridge/BridgePool.js
nicholaspai: Yeah my test case checks that the `instantRelayer` only gets a refund for the second correct instant relay so I think this test logic won't change

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: I was matching the type of `realizedLpFeePct` in the `Relay` struct isn't that more consistent?

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: If the second relay has different params, then there won't be an instant relayer. If the relay has the same params, there will be an instant relayer. WDYT?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: I can't figure out how to get `encodeParameters`. [docs](https://web3js.readthedocs.io/en/v1.2.6/web3.html#web3-modules) do you see a way? Do we need to import a new `web3-eth` package?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
chrismaree: everything in that struct is a string to make the storage cheaper in memory.

chrismaree: otherwise, what you have will also be fine. it's not too different to how other params are setup

chrismaree (approved):
This is excellent! we should have done this kind of change in the beginning. Great work on updating everything and the associated off-chain implementations.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -387,7 +391,6 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         emit RelaySettled(depositHash, relayHash, msg.sender);
 
         delete relay.realizedLpFeePct;
-        delete relay.instantRelayer;
```
mrice32, line 390: Why can't we still delete the instant relayer (to get a refund)? Once it's settled, there should be no way to create a request with this deposit hash anyway, right?

mrice32 (approved):
LGTM!

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: true! I forgot about the other mapping

PR #3426, opened 2021-10-04
feat(insured-bridge): Add the ability to transfer bridgeAdmin in the bridgePool to enable upgradability

chrismaree:
**Motivation**

The bridge pool currently has a set owner that cant be changed. There might be cases where we wish to upgrade the bridge admin to a new address, if we change just the bridge admin contract implementation. 

**Summary**

This PR adds a small change to the bridge Pool and Bridge admin to enable the upgradability of the bridgeAdmin address

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgeAdmin.sol
+++ b/packages/core/contracts/insured-bridge/BridgeAdmin.sol
@@ -116,6 +116,19 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
         emit SetDepositContracts(_chainId, _depositContract, _messengerContract);
     }
 
+    /**
+     * @notice Enables the current owner to transfer ownership of all owned bridge pools to a new owner.
+     * @dev Only callable by the current owner.
+     * @param _bridgePools array of bridge pools to transfer ownership.
+     * @param _newAdmin new admin contract to set ownership to.
+     */
+    function transferBridgePoolAdmin(address[] memory _bridgePools, address _newAdmin) public onlyOwner nonReentrant() {
+        for (uint8 i = 0; i < _bridgePools.length; i++) {
+            BridgePoolInterface(_bridgePools[i]).changeAdmin(_newAdmin);
+        }
+        emit BridgePoolsAdminTransferred(_bridgePools, _newAdmin);
+    }
+
     /**************************************************
      *        CROSSDOMAIN ADMIN FUNCTIONS             *
      **************************************************/
--- a/packages/core/contracts/insured-bridge/BridgeAdmin.sol
+++ b/packages/core/contracts/insured-bridge/BridgeAdmin.sol
@@ -123,7 +123,7 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
      * @param _newAdmin new admin contract to set ownership to.
      */
     function transferBridgePoolAdmin(address[] memory _bridgePools, address _newAdmin) public onlyOwner nonReentrant() {
-        for (uint16 i = 0; i < _bridgePools.length; i++) {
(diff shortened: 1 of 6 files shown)


packages/core/contracts/insured-bridge/interfaces/BridgePoolInterface.sol
```diff
@@ -5,4 +5,6 @@ import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 
 interface BridgePoolInterface {
```
chrismaree, line 6: right now this interface is not extensive and only covers the two functions called in the `BridgeAdmin`. do we think we should expand this to be a full interface like we have in the bridgeAdmin? I think we should, agreed?

packages/core/contracts/insured-bridge/BridgeAdmin.sol
```diff
@@ -116,6 +116,19 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
...
     }
 
+    /**
+     * @notice Enables the current owner to transfer ownership of a set of owned bridge pools to a new owner.
+     * @dev Only callable by the current owner.
+     * @param _bridgePools array of bridge pools to transfer ownership.
+     * @param _newAdmin new admin contract to set ownership to.
+     */
+    function transferBridgePoolAdmin(address[] memory _bridgePools, address _newAdmin) public onlyOwner nonReentrant() {
+        for (uint8 i = 0; i < _bridgePools.length; i++) {
```
nicholaspai, line 126: What if we have more than 2^8 bridge pools? This seems potentially too tight, why not use uint16

packages/core/contracts/insured-bridge/interfaces/BridgePoolInterface.sol
nicholaspai: I think we should only add methods to here that we know we need to use.

packages/core/contracts/insured-bridge/BridgeAdmin.sol
chrismaree: can just call the function multiple times but I'm happy to increase it to uint16

packages/core/contracts/insured-bridge/interfaces/BridgePoolInterface.sol
chrismaree: sure. that's how it is now so that's easy.

packages/core/contracts/insured-bridge/BridgeAdmin.sol
```diff
@@ -116,6 +116,19 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
...
     }
 
+    /**
+     * @notice Enables the current owner to transfer ownership of a set of owned bridge pools to a new owner.
+     * @dev Only callable by the current owner.
+     * @param _bridgePools array of bridge pools to transfer ownership.
+     * @param _newAdmin new admin contract to set ownership to.
+     */
+    function transferBridgePoolAdmin(address[] memory _bridgePools, address _newAdmin) public onlyOwner nonReentrant() {
+        for (uint16 i = 0; i < _bridgePools.length; i++) {
```
nicholaspai, line 126: Oh I didn't see that you pass `_bridgePools` in as a function param, I think its OK to revert back to uint8. Your call, sorry

packages/core/contracts/insured-bridge/BridgeAdmin.sol
nicholaspai: Yeah my bad you can revert I missed that part

nicholaspai (approved):
LGTM I think you're ok to revert back to uint8

PR #3428, opened 2021-10-04
nit(insured-bridge): refactor test syntax to avoid runDefaultFixture

chrismaree:
**Motivation**

Some of the insured bridge tests use `runDefaultFixture` and others don't. This PR removes the need for this call to: a) enable tests to be run faster b) enable tests to easily run with mocha.

**Summary**

Small JS refactor to insured bridge tests.

**Testing**

- [ ]  New unit tests created
- [X]  Untested


nicholaspai (approved):
Thanks for this LGTM. How much speedup do we get out of curiosity?

chrismaree:
> Thanks for this LGTM. How much speedup do we get out of curiosity?

it's pretty meaningful when running locally (2 or 3x) but not too much in CI (a few seconds)

PR #3430, opened 2021-10-04
feat(insured-bridge): Integrate SkinnyOptimisticOracle with BridgePool

nicholaspai:
**Motivation**

The SkinnyOptimisticOracle introduced a gas-lite implementation of the OptimisticOracle, intended to be used with the BridgePool. This PR formally integrates the two contracts.

**Summary**

Gas costs with SkinnyOptimisticOracle:
```
BridgePool              ·  addLiquidity                 ·     126977  ·     136577  ·         131421  ·           27  ·          -  │
|  BridgePool              ·  exchangeRateCurrent          ·      54255  ·      66105  ·          58798  ·           10  ·          -  │
|  BridgePool              ·  relayDeposit                 ·     320980  ·     421491  ·         399570  ·           31  ·          -  │
|  BridgePool              ·  removeLiquidity              ·      70774  ·     102779  ·          85188  ·            5  ·          -  │
|  BridgePool              ·  settleRelay                  ·     120876  ·     159877  ·         152814  ·           19  ·          -  │
|  BridgePool              ·  speedUpRelay                 ·      77836  ·      94936  ·          91122  ·            7  ·          -  │
|  SkinnyOptimisticOracle  ·  settle                       ·      86412  ·     208429  ·         115508  ·           22  ·          -  │
```

Gas costs with OptimisticOracle:
```
 BridgePool           ·  addLiquidity                 ·     126926  ·     136526  ·         131925  ·           24  ·          -  │
························|·······························|·············|·············|·················|···············|··············
|  BridgePool           ·  exchangeRateCurrent          ·      54255  ·      66105  ·          58798  ·           10  ·          -  │
························|·······························|·············|·············|·················|···············|··············
|  BridgePool           ·  relayDeposit                 ·     518276  ·     577536  ·         549038  ·           26  ·          -  │
························|·······························|·············|·············|·················|···············|··············
|  BridgePool           ·  removeLiquidity              ·      70733  ·     102728  ·          85139  ·            5  ·          -  │
························|·······························|·············|·············|·················|···············|··············
|  BridgePool           ·  settleRelay                  ·     242602  ·     446528  ·         303222  ·           18  ·          -  │
························|·······························|·············|·············|·················|···············|··············
|  BridgePool           ·  speedUpRelay                 ·      77742  ·      94842  ·          91028  ·            7  ·          -  │
```

**Details**

This PR reduces the `relayDeposit` and `settleRelay` costs by 150,000 gas each, resulting in a total cost of 730,000 down from 910,000 to `relayDeposit`, `speedUpRelay` and `settleRelay`.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -254,6 +259,9 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         relayData.realizedLpFeePct = realizedLpFeePct;
         relayData.slowRelayer = msg.sender;
 
+        bytes memory ancillaryData = getRelayAncillaryData(depositData, relayData);
```
mrice32, line 262: By calling this, I think we're doing the hexToUtf8 call twice (once here and once in the event where we compute the relayHash).

I think this will reduce gas by ~50k:
```
bytes32 relayHash = _getRelayHash(depositData, relayData);
bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
relayRequestAncillaryData[ancillaryData] = depositHash;
...
emit DepositRelayed(
    relayId,
    depositData,
    msg.sender,
    address(l1Token),
    realizedLpFeePct,
    depositHash,
    relayHash
);
```

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -388,13 +383,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+     */
+    function priceDisputed(
+        bytes32 _identifier,
+        uint32 _timestamp,
+        bytes memory _ancillaryData,
+        SkinnyOptimisticOracleInterface.Request memory _request
+    ) external onlyFromOptimisticOracle {
+        bytes32 depositHash = relayRequestAncillaryData[_ancillaryData];
+        RelayData storage relay = relays[depositHash];
+        require(relay.relayState == RelayState.Pending);
```
mrice32, line 411: What happens if this require fails? Just the relay state won't get set?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -388,13 +383,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+     * @notice Callback for settlements, marks relay as ready for finalization if the relay was resolved as valid.
+     * @dev Reverts if relay is uninitialized or already settled.
+     * @dev _timestamp and _identifier are unused because _ancillaryData contains a relay nonce and uniquely
+     * identifies a relay request.
+     * @param _identifier price identifier for relay request.
+     * @param _timestamp timestamp for relay request.
+     * @param _ancillaryData ancillary data for relay request.
+     * @param _request settled relay request params.
+     */
+    function priceSettled(
```
mrice32, line 425: Does this mean that a separate settle call in the OO is required for a full lifecycle? What impact does that have on total gas costs?

If this is necessary, why not just do the full settlement in this callback?

Also, I thought our previous model threw out disputes and didn't consider them. It seems like they can now be used to resolve requests? Is that correct? If so, what happens if a dispute comes back in the _middle_ of a follow-up proposal?

mrice32 (commented):
Looks good! A few questions.

packages/common/src/Enums.ts
```diff
@@ -44,5 +44,6 @@ export const InsuredBridgeRelayStateEnum = {
   UNINITIALIZED: "0",
   PENDING: "1",
   DISPUTED: "2",
-  FINALIZED: "3",
+  PENDING_FINALIZATION: "3",
+  FINALIZED: "4",
```
chrismaree, line 48: this does not impact anything else, right?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -231,13 +241,8 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
         require(
             relays[depositHash].relayState == RelayState.Uninitialized ||
-                _getOptimisticOracle().getState(
-                    address(this),
-                    bridgeAdmin.identifier(),
-                    relays[depositHash].priceRequestTime,
-                    _getRelayAncillaryData(_getRelayHash(depositData, relays[depositHash]))
-                ) ==
-                OptimisticOracleInterface.State.Disputed
+                relays[depositHash].relayState == RelayState.Disputed,
```
chrismaree, line 244: can you explain why we dont need to check the disputed state? it's due to the `relayRequestAncillaryData`  I'm pretty sure but a comment would be a good idea.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -87,6 +87,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     // "Uninitialized" state when they are disputed on the OptimisticOracle.
     mapping(bytes32 => RelayData) public relays;
 
+    // Associates a relay request's ancillary data with the deposit hash that the relay request was linked with. This
+    // mapping is used by the OptimisticOracle callback functions (i.e. priceDisputed, priceSettled) to identify the
+    // relay request that was disputed or settled.
+    mapping(bytes => bytes32) public relayRequestAncillaryData;
```
chrismaree, line 93: would changing this to a `bytes32 => bytes32` save enough gas to justify it? you could store a hash of the ancillary perhaps? wdyt?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -262,15 +270,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        // Request a price for the relay identifier and propose "true" optimistically. This method will pull the
         // (proposer reward + proposer bond + final fee) from the caller. We need to set a new price request timestamp
         // instead of default setting to equal to the `depositTimestamp`, which is dependent on the L2 VM on which the
         // DepositContract is deployed. Imagine if the timestamps on the L2 have an offset that are always "in the
         // future" relative to L1 blocks, then the OptimisticOracle would always reject requests.
-        bytes32 relayHash = _getRelayHash(depositData, relayData);
-        bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
-        _requestOraclePriceRelay(amount, priceRequestTime, ancillaryData);
-        _proposeOraclePriceRelay(amount, priceRequestTime, ancillaryData);
+        _requestAndProposeOraclePriceRelay(amount, priceRequestTime, ancillaryData);
```
chrismaree, line 278: ❤️

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: it wont impact the OO as it's within a try catch that side. It would mean the relay remains in `Pending` State I guess.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -388,13 +383,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        RelayData storage relay = relays[depositHash];
+        require(relay.relayState == RelayState.Pending);
+        relay.relayState = RelayState.Disputed;
+    }
+
+    /**
+     * @notice Callback for settlements, marks relay as ready for finalization if the relay was resolved as valid.
+     * @dev Reverts if relay is uninitialized or already settled.
+     * @dev _timestamp and _identifier are unused because _ancillaryData contains a relay nonce and uniquely
+     * identifies a relay request.
```
chrismaree, line 419: nit: formatting

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -388,13 +383,60 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
         delete instantRelays[keccak256(abi.encode(depositHash, relay.realizedLpFeePct))];
         delete relay.realizedLpFeePct;
         delete relay.priceRequestTime;
     }
 
+    /**
+     * @notice Callback for disputes, marks relay as disputed.
+     * @dev Reverts if relay is not pending.
+     * @dev _timestamp and _identifier are unused because _ancillaryData contains a relay nonce and uniquely
+     * identifies a relay request.
```
chrismaree, line 397: nit: formatting

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: I guess one argument with not doing the full settlement here is that it will increase the cost of the OO settlement.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -616,63 +658,43 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
                 .rawValue;
-
-        // Pull the total bond from the caller so that the OptimisticOracle can subsequently pull it from here.
         l1Token.safeTransferFrom(msg.sender, address(this), totalBond);
         l1Token.safeApprove(address(optimisticOracle), totalBond);
-        optimisticOracle.proposePriceFor(
-            msg.sender,
-            address(this),
+
+        optimisticOracle.requestAndProposePriceFor(
```
chrismaree, line 682: this is great

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -616,63 +658,43 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+            // proposal has passed the challenge period.
+            0,
+            // Set the Optimistic oracle proposer bond for the price request.
+            _getProposerBond(amount),
+            // Set the Optimistic oracle liveness for the price request.
+            uint256(bridgeAdmin.optimisticOracleLiveness()),
+            // Caller is proposer.
+            msg.sender,
+            // Canonical value representing "True"; i.e. the proposed relay is valid.
+            1e18
```
chrismaree, line 697: is it not `int(1e18)`?

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -1037,6 +1278,19 @@ describe("BridgePool", () => {
 
       const relayerBalanceBefore = await l1Token.methods.balanceOf(relayer).call();
       const instantRelayerBalanceBefore = await l1Token.methods.balanceOf(instantRelayer).call();
+
+      // Settle OO request.
+      const proposeEvent = (await optimisticOracle.getPastEvents("ProposePrice", { fromBlock: 0 }))[0];
```
chrismaree, line 1283: for this and all following calls like this: do you need to provide the `fromBlock: 0`? does it not default to this?

chrismaree (commented):
This is awesome! the gas saving is excellent!

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Yeah I forget exactly what the situation was that prompted me to add this `require`, but the following `require(relay.relayState == RelayState.Pending || relay.relayState == RelayState.Disputed)` in `priceSettled` addresses the following situation:

- Relay request 1 is sent --> `RelayState = Pending`
- Someone disputes relay request 1 --> `RelayState = Disputed`
- Relay request 2 is sent with same params as request 1 --> `RelayState = Pending`
- Request 2 passes challenge period --> `RelayState = Pending`
- OO settles price request originating from request 2 --> `RelayState = PendingFinalization
- Someone calls `settleRelay` on BridgePool --> `RelayState = Finalized`
- Someone resolves dispute of relay request 1 

Without the above `require`, this last settlement would reset the `RelayState` to `PendingFinalization` and allow it to be settled again. It's possible because the following code in `priceSettled` would point to the same `relay` (that has already been Finalized!):

```
function priceSettled(
        bytes32 _identifier,
        uint32 _timestamp,
        bytes memory _ancillaryData,
        SkinnyOptimisticOracleInterface.Request memory _request
    ) external onlyFromOptimisticOracle {
        bytes32 depositHash = relayRequestAncillaryData[_ancillaryData];
```

nicholaspai: @chrismaree @mrice32 can you let me know if you agree with the above logic? The require statement that Matt mentioned I think can be removed, do you see a situation where removing it creates a case where a RelayState gets stuck?

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: > Also, I thought our previous model threw out disputes and didn't consider them. It seems like they can now be used to resolve requests? Is that correct? If so, what happens if a dispute comes back in the middle of a follow-up proposal?

The only way for a follow-up proposal to occur is if the original relay gets disputed, due to this line in `relayDeposit`:
```
        require(
            relays[depositHash].relayState == RelayState.Uninitialized ||
                relays[depositHash].relayState == RelayState.Disputed,
            "Pending relay exists"
        );
```

So at this point, either the original dispute can get resolved by the OO, or the follow-up relay can pass liveness and get resolved (by the OO). If the original relay and the follow-up relay have the same params, then the dispute should resolve to 1, and the relay can be settled. If they have the same params, then the payouts are correct and this results in an "early" settlement for the follow-up relay. If the original relay was incorrect, then the dispute will resolve to 0 and the follow-up relay can't be settled yet until it passes liveness.

nicholaspai: > Does this mean that a separate settle call in the OO is required for a full lifecycle? What impact does that have on total gas costs?

> If this is necessary, why not just do the full settlement in this callback?

We could do this, thoughts?

packages/common/src/Enums.ts
nicholaspai: Nope

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: So this check allows a disputed relay to be "re-relayed" I explain that above. Did you mean something else?

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Doesn't this automatically cast it? I can be more explicit I'll change it

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Yep this saves a bit of gas (<1000) but I'll take it!

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: > Does this mean that a separate settle call in the OO is required for a full lifecycle? What impact does that have on total gas costs?

> If this is necessary, why not just do the full settlement in this callback?

Actually, i'm not sure there is an easy way to do this. The OO callsback into `priceSettled` with the ancillary data. We can't really back out the `DepositData` from the `ancillaryData`, as we currently map `ancillaryData => depositHash` and getting `DepositData` from `depositHash` is non-trivial. Without `DepositData` we can't settle easily since that data stores the relay amount for example.

packages/core/test/insured-bridge/BridgePool.js
nicholaspai: Nope it doesn't default to this like it used to with truffle

chrismaree: o ok. pitty

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -308,7 +314,9 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         RelayData storage relay = relays[depositHash];
         bytes32 instantRelayHash = keccak256(abi.encode(depositHash, relay.realizedLpFeePct));
         require(
-            relays[depositHash].relayState == RelayState.Pending && instantRelays[instantRelayHash] == address(0),
+            (relays[depositHash].relayState != RelayState.Uninitialized ||
```
chrismaree, line 317: it might be good to add a comment to each of these requires explaining why they are needed, even if they are inline with each. for example:

```
                instantRelays[instantRelayHash] == address(0), // instantRelayer must be zero address (i.e no previous instant relayer)
```

or something to this effect for each require

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: no your explanation is good! I was more hoping for a comment in the code to this effect.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -334,21 +342,9 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        // dispute resolved to a price of 0), then the disputer needs to go through OptimisticOracle to settle their
-        // payout.
-        bytes32 relayHash = _getRelayHash(_depositData, relay);
-        bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
         require(
-            _getOptimisticOracle().settleAndGetPrice(bridgeAdmin.identifier(), relay.priceRequestTime, ancillaryData) ==
-                int256(1e18), // Canonical value representing "True"; i.e. the proposed relay is valid.
-            "Relay request was not valid"
+            relays[depositHash].relayState == RelayState.PendingFinalization,
+            "Can only settle relay if price is resolved True on OptimisticOracle"
```
chrismaree, line 347: nit: make this require message less than 32 chars to save gas

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: ye, I think the flow you pointed out makes sence and having the modifier on the `priceSettled` callback is sufficient for this. I agree without the modifier we'd have issues with re-setting this state to.

I cant right now think of a case where this will get stuck (I'll spend some more time looking at this particular bit of code though).

One really nice sideffect of this change is the logic is really really simple now: only if it's in `RelayState.PendingFinalization` can it be settled.

Please can you add a comment above the require explaining how this works in the re-relayed case to make it clear for us in the future & to help with audit?

mrice32:
> This PR reduces the relayDeposit and settleRelay costs by 150,000 gas each, resulting in a total cost of 610,000 down from 910,000 to relayDeposit, speedUpRelay and settleRelay.

@nicholaspai I think this gas analysis is misleading because someone has to pay for the transaction to settle the price request in a separate transaction. I think we should add this amount to the total gas cost. Can you add this in?

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: I'm going to change this logic a lot in the next PR so I'm going to skip comments for now

mrice32 (approved):
LGTM!

PR #3433, opened 2021-10-05
fix(insured-bridge): remove fixedpoint from InsuredBridge math to save gas

chrismaree:
**Motivation**

Previously, the insured-bridge used `FixedPoint` for computations. Removing this from the math makes it a bit harder to read but saves a non negligible amount of gas. 

**Summary**

Re-writes numeric operations to work without `FixedPoint`. Note we still need to include the fixed point library to extract the `finalFee` from the call to the `Store` via `_getStore().computeFinalFee(address(l1Token)).rawValue`

**Gas Difference:**
`relayDeposit` before PR:542868. after PR: 539943. saving: 2925
`speedUpRelay` before PR: 94842. after PR: 93958. saving: 884
`settleRelay` before PR: 309697. after PR: 306287. saving: 3410
total saving: 7219 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


nicholaspai (approved):
LGTM this was much needed!

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -144,8 +144,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     function addLiquidity(uint256 l1TokenAmount) public {
         l1Token.safeTransferFrom(msg.sender, address(this), l1TokenAmount);
 
-        uint256 lpTokensToMint =
-            FixedPoint.Unsigned(l1TokenAmount).div(FixedPoint.Unsigned(exchangeRateCurrent())).rawValue;
+        uint256 lpTokensToMint = (l1TokenAmount * 1e18) / exchangeRateCurrent();
```
nicholaspai, line 147: This is all possible because solidity 8 uses safe math by default right?

chrismaree: yup :)

chrismaree:
@mrice32 I'd just like to double-check you are OK with how we are doing math in this PR.

mrice32 (approved):
LGTM!

PR #3436, opened 2021-10-05
fix(insured-bridge): fix whitelist to work with multiple L2 chains

chrismaree:
**Motivation**

The current Bridge Admin has no means to support a 1:n relationship between L1 tokens and L2 tokens. For example, consider USDC on arbitrum L2 and optimism L2. This PR updates the implementation to support this kind of whitelisting behaviour.
**Summary**

In addition to solving the relationship problem within the data structure, this PR also adds a unit test to show how the `whitelistToken` method should be used when updating the `bridgePool` address within L2 `BridgeDepositBox`es.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgeAdmin.sol
+++ b/packages/core/contracts/insured-bridge/BridgeAdmin.sol
@@ -222,7 +222,7 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
         require(address(BridgePoolInterface(_bridgePool).l1Token()) == _l1Token, "Bridge pool has different L1 token");
 
         L1TokenRelationships storage l1TokenRelationships = _whitelistedTokens[_l1Token];
-        l1TokenRelationships.l2Token[_chainId] = _l2Token; // Set the L2Token at the index of the chainId.
+        l1TokenRelationships.l2Tokens[_chainId] = _l2Token; // Set the L2Token at the index of the chainId.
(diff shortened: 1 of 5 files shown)


packages/core/contracts/insured-bridge/BridgeAdmin.sol
```diff
@@ -231,14 +235,19 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
...
 
-    function whitelistedTokens(address _l1Token) external view override returns (L1TokenRelationships memory) {
-        return _whitelistedTokens[_l1Token];
+    function whitelistedTokens(address _l1Token, uint256 chainId)
+        external
+        view
+        override
+        returns (address l2Token, address bridgePool)
+    {
+        return (_whitelistedTokens[_l1Token].l2Tokens[chainId], _whitelistedTokens[_l1Token].bridgePool);
```
nicholaspai, line 250: You can't return `L1TokenRelationships`?

nicholaspai (approved):
This is awesome, does it not break the existing off-chain tests in `financial-templates-lib`?

packages/core/contracts/insured-bridge/BridgeAdmin.sol
chrismaree: nope. this does not work. you get this:

image

PR #3438, opened 2021-10-05
feat(insured-bridge): Reduce function gas costs by storing hash of Relay params instead of full struct

nicholaspai:
**Motivation**

New gas costs:
```
|  BridgePool              ·  relayDeposit                 ·     331973  ·     375793  ·         362410  ·           29  ·          -  │
|  BridgePool              ·  speedUpRelay                 ·      78853  ·      95941  ·          91493  ·            6  ·          -  │
|  BridgePool              ·  settleRelay                  ·     174741  ·     235642  ·         213350  ·           18  ·          -  │
```

Total flow cost to relay, speed up and settle is now 667,253 on average, down from 730,000 in #3430 .

**Summary**

`speedUpRelay` and `settleRelay` now require the Relay params to be passed as function input. Additionally, the `priceSettled` callback is removed since its no longer possible to locate a Relay hash using just the callback's ancillary data. Instead, `settleRelay` takes in price request params so that it can try to optimistically settle and fetch a relay request's price. This additional gas cost is offset by reduced storage.

There are some additional changes worth pointing out:
- The `Disputed` state is removed. `priceDisputed` now deletes a relay hash. `relayDeposit` now checks if the relay hash is empty.
- The `PendingFinalization` state is removed because `settleRelay` settles directly with the OO or checks if it has a price. 

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -325,27 +309,70 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+     * for this deposit.
+     * @param ancillaryData Relay price request's ancillary data. Used to check status of relay price request.
+     * @param request Relay price request struct. Used to check status of relay price request.
      */
-    function settleRelay(DepositData memory depositData) public nonReentrant() {
+    function settleRelay(
+        DepositData memory depositData,
+        RelayData memory relayData,
+        bytes memory ancillaryData,
+        SkinnyOptimisticOracleInterface.Request memory request
```
nicholaspai, line 334: Have fun calling this method

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -325,27 +315,70 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+                _getOptimisticOracle().hasPrice(
+                    address(this),
+                    bridgeAdmin.identifier(),
+                    uint32(relayData.priceRequestTime),
+                    ancillaryData,
+                    request
+                ) && request.resolvedPrice == int256(1e18),
+                "Settle relay iff price is True"
+            );
+        }
```
mrice32, line 370: Just out of curiosity, why'd you choose to do this using a try-catch?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -379,17 +413,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        allocateLpFees(_getAmountFromPct(relay.realizedLpFeePct, depositData.amount));
+        allocateLpFees(_getAmountFromPct(relayData.realizedLpFeePct, depositData.amount));
 
-        emit RelaySettled(depositHash, _getRelayHash(depositData, relay), msg.sender);
+        emit RelaySettled(depositHash, msg.sender, relayData);
 
-        delete instantRelays[keccak256(abi.encode(depositHash, relay.realizedLpFeePct))];
-        delete relay.realizedLpFeePct;
-        delete relay.priceRequestTime;
+        delete instantRelays[instantRelayHash];
```
mrice32, line 420: Can we add:
```
delete relayRequestAncillaryData[keccak256(ancillaryData)];
```

to get a refund for this storage?

mrice32 (approved):
This looks great!!!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: A lot of the information in request, we might be able to infer from relay and deposit data, so (in a follow-up) we may be able to get away with having the user only providing some of these fields.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -59,7 +59,15 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     BridgeAdminInterface public bridgeAdmin;
 
     // A Relay represents an attempt to finalize a cross-chain transfer that originated on an L2 DepositBox contract.
-    enum RelayState { Uninitialized, Pending, Disputed, PendingFinalization, Finalized }
+    // The flow chart between states is as follows:
+    // - Begin at Uninitialized.
+    // - When relayDeposit() is called, a new relay is created with state Pending and mapped to the L2 deposit hash.
+    // - If the relay is disputed, the relay gets deleted and the L2 deposit hash has no relay mapped to it anymore.
+    // - When settleRelay() is successfully called, the relay state gets set to Finalized and cannot change from there.
```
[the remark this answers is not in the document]
nicholaspai: I'm more worried about the latter. One assumption we make is that state can never be reset once it gets set to `Finalized`. I don't think the Skinny OO can cause this to happen (via a `priceDisputed` callback) but I'm not 100% sure.

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
```diff
@@ -122,7 +122,6 @@ describe("InsuredBridgePriceFeed", function () {
         toBNWei("0.1")
       ).toString(),
       slowRelayer: relayer,
-      instantRelayer: ZERO_ADDRESS,
```
nicholaspai, line 125: This is no longer stored in relay data

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -99,16 +109,14 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     event LiquidityAdded(address indexed token, uint256 amount, uint256 lpTokensMinted, address liquidityProvider);
     event LiquidityRemoved(address indexed token, uint256 amount, uint256 lpTokensBurnt, address liquidityProvider);
     event DepositRelayed(
-        uint32 indexed relayId,
+        bytes32 indexed depositHash,
         DepositData depositData,
-        address slowRelayer,
         address l1Token,
-        uint64 realizedLpFeePct,
```
chrismaree, line 106: hhmm. not having this could be an issue. it makes the off-chain bot work a bit harder. what exactly is the flow now the bot must go through te get this data to validate it in the price feed in the OO dispute bot.

chrismaree: nvm this is baked into the `RelayData`!

nicholaspai: image

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -246,29 +254,22 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        relayData.relayId = relayId;
-        relayData.relayState = RelayState.Pending;
-        relayData.priceRequestTime = priceRequestTime;
-        relayData.realizedLpFeePct = realizedLpFeePct;
-        relayData.slowRelayer = msg.sender;
+        // Save hash of new relay attempt parameters.
+        RelayData memory relayData =
+            RelayData({
+                relayState: RelayState.Pending,
+                slowRelayer: msg.sender, // Note: This increments the storage variable at the same time as setting relayId.
```
chrismaree, line 267: this comment is on the wrong line

packages/common/src/HardhatConfig.ts
```diff
@@ -76,8 +76,11 @@ export function getHardhatConfig(
       mainnet: { chainId: 1, url: getNodeUrl("mainnet", true), accounts: { mnemonic } },
       rinkeby: { chainId: 4, url: getNodeUrl("rinkeby", true), accounts: { mnemonic } },
       goerli: { chainId: 5, url: getNodeUrl("goerli", true), accounts: { mnemonic } },
-
       kovan: { chainId: 42, url: getNodeUrl("kovan", true), accounts: { mnemonic } },
+      arbitrumRinkeby: {
```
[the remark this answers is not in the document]
chrismaree: also you should use the `getNodeUrl` like the other chains.

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: ye... this is going to be painful. ideally re-constructed in the bot so no on-chain calls

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -335,28 +329,73 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
         bytes32 depositHash = _getDepositHash(depositData);
-        RelayData storage relay = relays[depositHash];
-
-        require(relays[depositHash].relayState == RelayState.PendingFinalization, "Settle iff price resolved True");
+        _validateRelayDataHash(depositHash, relayData);
+        require(relayData.relayState == RelayState.Pending, "Already settled");
+
+        // Optimistically try to settle relay. If this transaction settles the relay, then check the newly resolved
+        // price. Otherwise grab the already resolved price.
+        try
```
chrismaree, line 362: this is a really cool way of doing this.

packages/common/src/HardhatConfig.ts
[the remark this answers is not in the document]
nicholaspai: Oops sorry i'll remove

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -409,39 +450,19 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
      * @param request disputed relay request params.
      */
     function priceDisputed(
+        // Unused, but the identifier should be same as bridgeAdmin.identifier()
```
chrismaree, line 453: is there any case where this could be called in the callback on a different identifier?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -256,11 +264,16 @@ export class Relayer {
...
+          relayId: receipt.events.RelaySpedUp.returnValues.relay.relayId,
+          relayState: receipt.events.RelaySpedUp.returnValues.relay.relayState,
+          priceRequestTime: receipt.events.RelaySpedUp.returnValues.relay.priceRequestTime,
+          slowRelayer: receipt.events.RelaySpedUp.returnValues.relay.slowRelayer,
           transactionConfig,
         });
       else throw receipt;
     } catch (error) {
-      this.logger.error({ at: "InsuredBridgeRelayer#Relayer", type: "Something errored instantly relaying!", error });
+      this.logger.error({ at: "InsuredBridgeRelayer#Relayer", type: "Something errored speeding up relay!", error });
```
chrismaree, line 276: good catch

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -36,18 +30,25 @@ export interface Relay {
...
+export interface BridgePoolData {
+  contract: BridgePoolWeb3;
+  currentTime: number;
+  relayNonce: number;
+}
+
 export class InsuredBridgeL1Client {
   public readonly bridgeAdmin: BridgeAdminInterfaceWeb3;
-  public bridgePools: { [key: string]: BridgePoolWeb3 }; // L1TokenAddress=>BridgePoolClient
+  public bridgePools: { [key: string]: BridgePoolData }; // L1TokenAddress=>BridgePoolData
```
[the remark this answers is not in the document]
nicholaspai: Because we need the Relayer to know about the current contract time and relay nonce when instant relaying (slow + speedupRelay) since `speedUpRelay` takes the `RelayData` as a function param. So, instead of mapping L1 tokens to just the bridgepool contracts I thought it best to map to the contract, + those two state values mentioned. WDYT?

mrice32:
> 562,279 on average, down from 730,000 in #3430.

Can we see a before-after _just on this PR_? I think multiple PRs may be taken into account when you just compare to that other PR. The reason I think this is important is so that we 1) make sure that we're not _gaining_ gas over any previous PRs and 2) to make sure the savings match our intuition (aka there's not some weird effect that is causing them to be larger or smaller than expected that could help us).

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -580,36 +602,22 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-    // Note: this method is identical to the one above, but it allows storage to be passed in, which saves some gas
-    // (3-4k) when called internally due to solidity not needing to copy the entire data structure and just lazily read
-    //  data when requested.
+    /**************************************
+     *    INTERNAL & PRIVATE FUNCTIONS    *
+     **************************************/
+
+    // Note: this method is identical to `getRelayAncillaryData`, but it allows storage to be passed in, which saves
+    // some gas (~3-4k) when called internally due to solidity not needing to copy the entire data structure and just
+    // lazily read data when requested.
```
mrice32, line 614: (not something in your PR) this comment is out of date.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -335,28 +328,73 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+     * @param relayData Parameters of Relay that caller is attempting to settle. Must hash to the stored relay hash
+     * for this deposit.
+     * @param ancillaryData Relay price request's ancillary data. Used to check status of relay price request.
+     * @param request Relay price request struct. Used to check status of relay price request.
      */
-    function settleRelay(DepositData memory depositData) public nonReentrant() {
+    function settleRelay(
+        DepositData memory depositData,
+        RelayData memory relayData,
+        bytes memory ancillaryData,
```
mrice32, line 352: Is there anything to verify that this is the correct ancillary data for this request?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -335,28 +328,73 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
-        RelayData storage relay = relays[depositHash];
-
-        require(relays[depositHash].relayState == RelayState.PendingFinalization, "Settle iff price resolved True");
+        _validateRelayDataHash(depositHash, relayData);
+        require(relayData.relayState == RelayState.Pending, "Already settled");
+
+        // Optimistically try to settle relay. If this transaction settles the relay, then check the newly resolved
+        // price. Otherwise grab the already resolved price.
+        try
+            optimisticOracle.settle(
```
mrice32, line 362: Can we avoid the try catch by:
```
if (!request.settled) {
  (, in256 resolvedPrice) = optimisticOracle.settle(...);
  require(resolvedPrice == int256(1e18), "Settle relay iff price is True");
else {
  require(optimisticOracle.hasPrice(...) && request.resolvedPrice == int256(1e18));
}
```

Not sure if try-catch is cheaper, but I assume it isn't. Probably worth a quick test.

nicholaspai: This is about 500 gas cheaper for `settleRelay` on average. Question is, do you think checking the `request.settled` param opens the attack surface? Or are the assumptions the same about the link between `depositData`, `relayData`, `ancillaryData`, and `request`?

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Two requests with different details? One valid and one invalid?

mrice32: Like send a request that's valid. At settlement time, send a new request that's invalid (pays me $1MM), then settle the second with ancillary data from the first? I guess the timestamps might be an issue there, though. Not sure if there's an obvious exploit, but it does seem plausible to point at a different request w the same timestamp, which seems a little scary?

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: I think the assumptions are the same because in both cases we're asking the OO to verify that those request params match a valid request, right?

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: As discussed offline, this exploit is possible if you:
- send relay 1 at time 0
- dispute relay 1 at time 0
- send relay 2 at time 0
- resolve disputed relay 1 at any time
- before relay 2 expires, you can take the ancillary data from the resolved relay 1 request and call `settleRelay` with the relay 2 request.

Therefore, the contract should derive the ancillary data from the `depositData` and `relayData` rather than allowing the caller to pass it in.

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Yeah fair, done and commented. Also added a comment about pros/cons to trying to derive the `request` from just the contract state + function params

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -335,28 +328,73 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        require(relayData.relayState == RelayState.Pending, "Already settled");
+
+        // Resolve request if it is not settled or grab its resolved price if it is. If the request.settled param does
+        // not match the one stored in the relay price request on the OptimisticOracle, then this will revert. The
+        // relay price request must be linked to ancillary data that is derived from the `depositData` and `relayData`,
+        // so there is no way for the caller to modify the request params.
+        bytes memory ancillaryData = _getRelayAncillaryData(_getRelayHash(depositData, relayData));
+        if (!request.settled) {
+            (
+                ,
```
chrismaree, line 370: the linting on this is a bit nasty

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: ye, that makes sence to me. I agree with your logic. I'm still not convinced that everything works exactly as I understand it does but for now I'm happy.

PR #3441, opened 2021-10-06
fix(core): remove migrations contract

chrismaree:
**Motivation**

Archaeological artifact from the long-extinct lesser-spotted Truffle.


mrice32 (approved):
> long-extinct lesser-spotted Truffle.

PR #3442, opened 2021-10-06
feat(api): use Google Datastore as store provider for tables

amateima:
**Motivation**

https://app.shortcut.com/uma-project/story/934/integrate-with-google-datastore

**Summary**

I created a separate file for the api version that uses Google Datastore as data storage service. I kept this implementation in a separate file because most likely it will suffer modifications and optimisations in the future, so for now it would be pretty hard to have a single file for the `api` app that serves all our purposes (testing environment, in memory storage, datastore storage, etc).

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


amateima:
This is how the Google Datastore dashboard looks like 
Screenshot 2021-10-06 at 20 26 37

packages/api/src/apps/datastore_api/README.md
```diff
@@ -0,0 +1,305 @@
...
+
+- emp - all emp calls
+- lsp - all lsp calls
+- global - custom calls that are not contract type specific
+
+## Action Client
+
+See [../../examples/action-client.ts](Action Client Code) for an example of how to interface with the api from the client side.
+
+## Emp Api
```
daywiss, line 93: you can link to the api readme for these docs, similarly to what i did in the lsp readme

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -0,0 +1,327 @@
...
+          latest: {
+            tvl: [0, "0"],
+          },
+          history: {
+            tvl: empStatsHistory.Table("Tvl Global History"),
+          },
+        },
+      },
+    },
+    lastBlockUpdate: 0,
```
daywiss, line 111: we can make a separate issue for this, but the lastblockupdate should default to the latest block processed and persisted in store, or fallback to 0 if none was saved

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -0,0 +1,327 @@
...
+  console.log("Updated LSP state");
+
+  // we've update our state based on latest block we queried
+  appState.lastBlockUpdate = initBlock.number;
+
+  await services.erc20s.update();
+  console.log("Updated tokens");
+
+  // backfill price histories, disable if not specified in env
+  if (env.backfillDays) {
```
daywiss, line 180: we are going to also have to deal with backfilling based on the latest data we know about vs the time the service starts. this also can be a separate issue

packages/api/src/tables/stores-factory.ts
```diff
@@ -0,0 +1,23 @@
+import { Datastore } from "@google-cloud/datastore";
+import { tables, stores } from "@uma/sdk";
+import { empStats, empStatsHistory, lsps } from ".";
+
+const { GoogleDatastore } = stores;
+const datastoreClient = new Datastore();
+
+export const datastores = {
+  empsActiveStore: GoogleDatastore<string, tables.emps.Data>("Active Emp", datastoreClient),
```
daywiss, line 9: this file in general looks great, personally though i would not include "Stores" in the name of each propery, since you know that context from the import.

I would also suggest potentially doing this init in a function that takes in a datastoreClient parameter, so that could be configured independently by the caller

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -4,38 +4,38 @@ import { Datastore, Key } from "@google-cloud/datastore";
...
     } catch (err) {
       return;
     }
   }
-  async function set(id: string, data: D) {
+  async function set(id: I, data: D) {
     await store.save({
       key: makeKey(id),
-      method: "insert",
+      // method: "insert",
```
daywiss, line 23: u can delete this line

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -0,0 +1,327 @@
...
+  // services can emit events when necessary, though for now any services that depend on events must be in same process
+  const serviceEvents = new Events();
+  const { datastores } = storesFactory;
+  // state shared between services
+  const appState: AppState = {
+    provider,
+    web3,
+    coingecko: new Coingecko(),
+    zrx: new Zrx(env.zrxBaseUrl),
+    blocks: tables.blocks.Table("Block", datastores.blockStore),
```
daywiss, line 54: as mentioned before this can probably be removed, its not used anywhere

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -4,38 +4,38 @@ import { Datastore, Key } from "@google-cloud/datastore";
 import highland from "highland";
 
 // TODO: this is a work in progress which will be completed during datastore integration.
-export default function <D>(kind: string, store: Datastore): SortedStore<string, D> {
-  function makeKey(id: string): Key {
+export default function <I extends string | number, D>(kind: string, store: Datastore): SortedStore<I, D> {
```
daywiss, line 7: i was afraid this might have to change, all good

packages/sdk/src/stores/google-datastore/store.ts
```diff
@@ -4,38 +4,38 @@ import { Datastore, Key } from "@google-cloud/datastore";
 import highland from "highland";
 
 // TODO: this is a work in progress which will be completed during datastore integration.
```
daywiss, line 6: u can prob remove this comment

daywiss (approved):
this is awesome, it took a lot of detailed work to get this all working, but the results look really good.  theres still some tweaks to the app business logic to take advantage of this new persistence. i'd suggest doing those in a different PR unless its just minor updates, so up to you

packages/sdk/src/stores/google-datastore/store.ts
amateima: yeah, Datastore accepts string and number as key type

packages/api/src/apps/datastore_api/app.ts
amateima: I will make a separate PR for optimising the backfill and other function calls

packages/api/src/services/blocks.ts
```diff
@@ -1,29 +0,0 @@
-import { AppState } from "../types";
```
mrice32, line 1: Was this just unused?

mrice32 (approved):
This looks awesome! Very nice work

packages/api/src/services/blocks.ts
amateima: Yes

PR #3445, opened 2021-10-07
feat(LSP): Enable LSP to support early expiration

chrismaree:
**Motivation**

As part of doing more with the LPS, we require the ability to conditionally expire a contract early. This PR adds this ability to the LPS.

**Summary**

Adds a `requestEarlyExpiration` method to the LSP to enable early expiration. See code for more details.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
+++ b/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
@@ -47,7 +47,7 @@ contract LongShortPair is Testable, Lockable {
         IERC20 collateralToken; // Collateral token used to back LSP synthetics.
         LongShortPairFinancialProductLibrary financialProductLibrary; // Contract providing settlement payout logic.
         bytes customAncillaryData; // Custom ancillary data to be passed along with the price request to the OO.
-        uint256 prepaidProposerReward; // Preloaded reward to incentivize settlement price proposals.
+        uint256 proposerReward; // Optimistic oracle reward amount, pulled from the caller of the expire function.
         uint256 optimisticOracleLivenessTime; // OO liveness time for price requests.
         uint256 optimisticOracleProposerBond; // OO proposer bond for price requests.
         FinderInterface finder; // DVM finder to find other UMA ecosystem contracts.
@@ -79,7 +79,7 @@ contract LongShortPair is Testable, Lockable {
 
     // Optimistic oracle customization parameters.
     bytes public customAncillaryData;
-    uint256 public prepaidProposerReward;
+    uint256 public proposerReward;
     uint256 public optimisticOracleLivenessTime;
     uint256 public optimisticOracleProposerBond;
 
@@ -124,7 +124,7 @@ contract LongShortPair is Testable, Lockable {
      *    - `collateralToken`: Collateral token used to back LSP synthetics.
      *    - `financialProductLibrary`: Contract providing settlement payout logic.
      *    - `customAncillaryData`: Custom ancillary data to be passed along with the price request to the OO.
-     *    - `prepaidProposerReward`: Preloaded reward to incentivize settlement price proposals.
+     *    - `proposerReward`: Preloaded reward to incentivize settlement price proposals.
      *    - `optimisticOracleLivenessTime`: OO liveness time for price requests.
      *    - `optimisticOracleProposerBond`: OO proposer bond for price requests.
      *    - `finder`: DVM finder to find other UMA ecosystem contracts.
@@ -161,7 +161,7 @@ contract LongShortPair is Testable, Lockable {
         );
 
         customAncillaryData = params.customAncillaryData;
-        prepaidProposerReward = params.prepaidProposerReward;
+        proposerReward = params.proposerReward;
         optimisticOracleLivenessTime = params.optimisticOracleLivenessTime;
         optimisticOracleProposerBond = params.optimisticOracleProposerBond;
     }
@@ -365,14 +365,17 @@ contract LongShortPair is Testable, Lockable {
     function _requestOraclePrice(uint256 requestTimestamp, bytes memory requestAncillaryData) internal {
         OptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
 
-        // Use the prepaidProposerReward as the proposer reward.
-        if (prepaidProposerReward > 0) collateralToken.safeApprove(address(optimisticOracle), prepaidProposerReward);
+        // If the proposer reward was set then pull it from the caller of the function.
+        if (proposerReward > 0) {
+            collateralToken.safeTransferFrom(msg.sender, address(this), proposerReward);
+            collateralToken.safeApprove(address(optimisticOracle), proposerReward);
+        }
         optimisticOracle.requestPrice(
             priceIdentifier,
             requestTimestamp,
             requestAncillaryData,
             collateralToken,
-            prepaidProposerReward
+            proposerReward
         );
 
         // Set the Optimistic oracle liveness for the price request.
--- a/packages/core/test/financial-templates/long-short-pair/LongShortPair.js
+++ b/packages/core/test/financial-templates/long-short-pair/LongShortPair.js
@@ -1,7 +1,7 @@
 const hre = require("hardhat");
 const { web3 } = hre;
 const { getContract, assertEventEmitted } = hre;
-const { toWei, utf8ToHex, toBN, padRight } = web3.utils;
+const { toWei, fromWei, utf8ToHex, toBN, padRight } = web3.utils;
 const { assert } = require("chai");
 
 // Libraries and helpers
@@ -45,12 +45,12 @@ let store;
 
 let optimisticOracleLivenessTime = 7200;
 let optimisticOracleProposerBond = "0";
+let proposerReward = toWei("100"); // set to zero. test directly later.
 
 const startTimestamp = Math.floor(Date.now() / 1000);
 const expirationTimestamp = startTimestamp + 10000;
 const priceIdentifier = padRight(utf8ToHex("TEST_IDENTIFIER"), 64);
 const collateralPerPair = toWei("1"); // each pair of long and short tokens need 1 unit of collateral to mint.
-const prepaidProposerReward = toWei("100");
 const pairName = "Long Short Pair Test";
 
 describe("LongShortPair", function () {
@@ -136,7 +136,7 @@ describe("LongShortPair", function () {
       collateralToken: collateralToken.options.address,
       financialProductLibrary: longShortPairTestLibrary.options.address,
       customAncillaryData,
-      prepaidProposerReward,
+      proposerReward,
       optimisticOracleLivenessTime,
       optimisticOracleProposerBond,
       finder: finder.options.address,
@@ -144,13 +144,16 @@ describe("LongShortPair", function () {
     };
 
     longShortPair = await LongShortPair.new(constructorParams).send({ from: deployer });
-    await collateralToken.methods.mint(longShortPair.options.address, toWei("100")).send({ from: deployer });
 
     // Add mint and burn roles for the long and short tokens to the long short pair.
     await longToken.methods.addMember(1, longShortPair.options.address).send({ from: deployer });
     await shortToken.methods.addMember(1, longShortPair.options.address).send({ from: deployer });
     await longToken.methods.addMember(2, longShortPair.options.address).send({ from: deployer });
     await shortToken.methods.addMember(2, longShortPair.options.address).send({ from: deployer });
+
+    // Mint tokens to the deployer to pay for the proposer reward when calling expire.
+    await collateralToken.methods.mint(deployer, proposerReward).send({ from: deployer });
+    await collateralToken.methods.approve(longShortPair.options.address, proposerReward).send({ from: deployer });
   });
   describe("Basic Functionality", () => {
     it("Constructor params are set correctly", async function () {
@@ -167,7 +170,7 @@ describe("LongShortPair", function () {
         longShortPairTestLibrary.options.address
       );
       assert.equal(await longShortPair.methods.customAncillaryData().call(), customAncillaryData);
-      assert.equal(await longShortPair.methods.prepaidProposerReward().call(), prepaidProposerReward);
+      assert.equal(await longShortPair.methods.proposerReward().call(), proposerReward);
       assert.equal(await longShortPair.methods.optimisticOracleLivenessTime().call(), optimisticOracleLivenessTime);
       assert.equal(await longShortPair.methods.optimisticOracleProposerBond().call(), optimisticOracleProposerBond);
       assert.equal(await longShortPair.methods.enableEarlyExpiration().call(), false);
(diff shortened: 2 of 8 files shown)


packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
+    /**
+     * @notice Enables the LPS to request early expiration. This initiates a price request to the optimistic oracle at
+     * the provided timestamp with a modified version of the ancillary data that includes the key "earlyExpiration:1"
+     * which signals to the OO that this is an early expiration request, rather than standard settlement.
+     * @dev Will revert if: a) the contract is already early expire, b) it is after the expiration timestamp, c)
+     * early expiration is disabled for this contract, d) the proposed expiration timestamp is in the future.
+     * e) an early expiration attempt has already been made (in pending state).
+     * @param _earlyExpirationTimestamp timestamp at which the early expiration is proposed.
+     */
+    function requestEarlyExpiration(uint64 _earlyExpirationTimestamp)
```
chrismaree, line 279: hopefully this is simple enough. I think it's about as clean as we'll get it.

note that this does _not_ enable the requestor to propose a price at the same time. this was done intentionally to mimic the same flow with _all_ other settlement within uma contracts (i.e first settle is called and separately a price is proposed)

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
+
+        emit EarlyExpirationRequested(msg.sender, _earlyExpirationTimestamp);
+    }
+
+    /**
+     * @notice Expire the LSP contract. Makes a request to the optimistic oracle to inform the settlement price.
+     * @dev Will revert if: a) the contract is already early expire, b) it is before the expiration timestamp or c)
+     * an expire call has already been made.
+     */
+    function expire() public nonReentrant() notEarlyExpired() postExpiration() {
```
chrismaree, line 300: this behaves the same as before, assuming the contract was not early expired

packages/core/test/financial-templates/long-short-pair/LongShortPair.js
```diff
@@ -115,8 +143,8 @@ describe("LongShortPair", function () {
       timerAddress: timer.options.address,
     };
 
-    longShortPair = await LongShortPair.new(constructorParams).send({ from: accounts[0] });
-    await collateralToken.methods.mint(longShortPair.options.address, toWei("100")).send({ from: accounts[0] });
+    longShortPair = await LongShortPair.new(constructorParams).send({ from: deployer });
```
chrismaree, line 146: more nits

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -40,6 +41,7 @@ contract LongShortPair is Testable, Lockable {
         uint64 expirationTimestamp; // Unix timestamp of when the contract will expire.
         uint256 collateralPerPair; // How many units of collateral are required to mint one pair of synthetic tokens.
         bytes32 priceIdentifier; // Price identifier, registered in the DVM for the long short pair.
+        bool enableEarlyExpiration; // Enables the LPS contract to be settled early.
```
nicholaspai, line 44: Can you replace all "LPS" in this contract with "LSP"? Sorry to nit pick...

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -52,32 +54,28 @@ contract LongShortPair is Testable, Lockable {
         address timerAddress; // Timer used to synchronize contract time in testing. Set to 0x000... in production.
     }
 
-    enum ContractState { Open, ExpiredPriceRequested, ExpiredPriceReceived }
-    // @dev note contractState and expirationTimestamp are declared in this order so they use the same storage slot.
-    ContractState public contractState;
+    bool public receivedSettlementPrice;
 
+    // Contract parametrization.
```
nicholaspai, line 59:
```suggestion
```

I don't think this comments help

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
+     * @notice Generates a modified ancillary data that indicates the contract is being expired early.
+     */
+    function getEarlyExpirationAncillaryData() public view returns (bytes memory) {
+        return AncillaryData.appendKeyValueUint(customAncillaryData, "earlyExpiration", 1);
+    }
+
+    /**
+     * @notice Defines a special number that, if returned during an attempted early expiration, will cause the contract
+     * to do nothing and not expire. This enables the OO (and DVM voters in the case of a dispute) to choose to keep
+     * the contract running, thereby denying with the early settlement request.
```
nicholaspai, line 335:
```suggestion
     * the contract running, thereby denying the early settlement request.
```

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -223,23 +224,20 @@ contract LongShortPair is Testable, Lockable {
...
-            expiryPercentLong = Math.min(
-                financialProductLibrary.percentageLongCollateralAtExpiry(expiryPrice),
-                FixedPoint.fromUnscaledUint(1).rawValue
-            );
-            contractState = ContractState.ExpiredPriceReceived;
-        }
+        // Either time must have passed the expirationTimestamp (normal expiration) or it is before expirationTimestamp
+        // and early expiration is enabled enabled.
+        require(
+            getCurrentTime() > expirationTimestamp || (getCurrentTime() < expirationTimestamp && enableEarlyExpiration),
```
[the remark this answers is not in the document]
nicholaspai: I would have expected the RHS of the || in the `if` statement to look like: `getCurrentTime() < expirationTimestamp && getCurrentTime() >= earlyExpirationTimestamp && earlyExpirationTimstamp != 0`

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -223,23 +224,20 @@ contract LongShortPair is Testable, Lockable {
...
+            "Can not settle"
+        );
+
+        // Get the settlement price and store it. Also sets expiryPercentLong to inform settlement. Reverts if either:
+        // a) the price request has not resolved (either a normal expiration call or early expiration call) or b) If the
+        // the contract was attempted to be settled early but the price returned is the ignore oracle price.
+        if (!receivedSettlementPrice) getExpirationPrice();
```
nicholaspai, line 240: Why can't you check `expiryPrice == 0`?

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
+    /**
+     * @notice If the earlyExpirationTimestamp is != 0 then a previous early expiration OO request might still be in the
+     * pending state. Check if the OO contains the ignore early price. If it does not contain this then the contract
+     * was early expired correctly. Note that _getOraclePrice call will revert if the price request is still pending,
+     * thereby reverting all upstream calls pre-settlement of the early expiration price request.
+     */
+    function isContractEarlyExpired() public returns (bool) {
+        return (earlyExpirationTimestamp != 0 &&
+            _getOraclePrice(earlyExpirationTimestamp, getEarlyExpirationAncillaryData()) !=
+            ignoreEarlyExpirationPrice());
```
mrice32, line 350: Does this mean that if someone triggers an invalid early expiration just before real expiration, they could delay real expiration for up to 4 days? I don't really have a suggestion, just double checking my thinking.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
-            optimisticOracleProposerBond
+        optimisticOracle.setBond(priceIdentifier, requestTimestamp, requestAncillaryData, optimisticOracleProposerBond);
+    }
+
+    // Fetch the optimistic oracle expiration price. If the timestamp is before expiration tries to fetch the
+    // early expiration timestamp and ancillary data. Else, fetch the price for the provided expiration timestamp and
+    // ancillary data. If the time is before the expiration timestamp (i.e early expiration) and the price is the ignore
+    // early expiration price then revert. This acts to block settlement on bad early expiration attempts.
+    function getExpirationPrice() internal {
+        if (getCurrentTime() < expirationTimestamp) {
```
mrice32, line 394: This condition strikes me a bit odd. It doesn't seem like this perfectly sorts between early and regular expiry (unless I'm missing something).

Couldn't the current time be after the expiration timestamp _even if_ we expired it early?

mrice32 (approved):
Looks awesome!

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: yes! that's not a nit pick sir, that's a great call out!

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: yup! normal settlement is blocked until such time that the early expiration settlement has concluded. I had the unit test: `Can not call expire (normal expiration) if an early expiration is pending or correctly passed` which validated that it's blocked in the way you describe.

However, you are describing a subset of this test which is the expiration happens _just before_ warrants an additional test. I'm going to add one now to add additional colour on this to be 100% sure it behaves the way you described here.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +267,143 @@ contract LongShortPair is Testable, Lockable {
...
+        returns (uint256 longTokens, uint256 shortTokens)
+    {
         return (longToken.balanceOf(sponsor), shortToken.balanceOf(sponsor));
     }
 
+    /**
+     * @notice Generates a modified ancillary data that indicates the contract is being expired early.
+     */
+    function getEarlyExpirationAncillaryData() public view returns (bytes memory) {
+        return AncillaryData.appendKeyValueUint(customAncillaryData, "earlyExpiration", 1);
```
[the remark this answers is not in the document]
nicholaspai: OH you're right, pls ignore

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
nicholaspai: Ah

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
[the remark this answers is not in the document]
nicholaspai: > This behaviour is the same as if it is post-expiration where in expire has been called but no price has been returned from the OO. In this case, you can enter the function (pas this require) and then getExpirationPrice will revert as there is not yet a price.

This was a really helpful comparison to think about, also worth adding a comment about.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
nicholaspai: Can you add this comment above `receivedSettlementPrice`? Otherwise it seems redundant at first, until you realize 0 is a valid price.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -223,23 +223,20 @@ contract LongShortPair is Testable, Lockable {
...
-            expiryPercentLong = Math.min(
-                financialProductLibrary.percentageLongCollateralAtExpiry(expiryPrice),
-                FixedPoint.fromUnscaledUint(1).rawValue
-            );
-            contractState = ContractState.ExpiredPriceReceived;
-        }
+        // Either time must have passed the expirationTimestamp (normal expiration) or it is before expirationTimestamp
+        // and early expiration is enabled enabled.
+        require(
+            getCurrentTime() > expirationTimestamp || (getCurrentTime() < expirationTimestamp && enableEarlyExpiration),
```
nicholaspai, line 232: nit: it's more readable IMO to have `enableEarlyExpiration` appear first in the && block of the RHS of the || block. It also might exit faster on average since that bool is easier to check than the "<" comparison.

chrismaree: sure.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
```diff
@@ -269,66 +268,156 @@ contract LongShortPair is Testable, Lockable {
...
+        nonReentrant()
+        notEarlyExpired()
+        preExpiration()
+    {
+        require(enableEarlyExpiration, "Early expiration disabled");
+        require(_earlyExpirationTimestamp <= getCurrentTime(), "Only propose expire in the past");
+
+        earlyExpirationTimestamp = _earlyExpirationTimestamp;
+
+        _requestOraclePrice(earlyExpirationTimestamp, getEarlyExpirationAncillaryData());
```
mrice32, line 291: I think this breaks the reward system.

The deployer deposits a prepaid reward. That's only enough to reward a single proposer. If I run this multiple times, the contract will pay the reward multiple times, which will dip into the users' collateral.

At worst, this is an exploit where I can repeatedly request early expiry and resolve it to false (which would be correct) to steal the reward amount from the contract.

It also seems slightly odd that we don't bond the person who is requesting the early expiry. Not sure what the right solution is -- maybe it means we need to have the person propose a price as a part of this call. If we did that, we could also use the feature refundOnDispute in the OO that would refund the reward to the requester if the price request is disputed, which would ensure we would always have a reward for any follow-up requests.

Another solution is just to disable the reward on early expiries. That seems a bit inconsistent and strange, but it might be the simplest option.

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
mrice32: Can we not check the early expiration timestamp or something? Is there no state that tells us which one has been triggered so this flow is simplified? Could we take advantage of the normal expire function to wipe away any early expiry state that might be hanging around?

chrismaree: I've drastically simplified this. PTAL

packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol
chrismaree: this is a great point! I will solve this in a separate PR.

chrismaree: ended up fixing this in the same PR. now the caller of this method (and the expire method) pays the reward.

PR #3449, opened 2021-10-08
feat(across): allow relay and speed up to be called simultaneously

mrice32:
**Motivation**

We'd like relay and speed up to be able to be called in the same transaction. Using multicall is intractible because it requires the caller to predict the time.

**Summary**

This adds a method relayAndSpeedUp to allow relay and speed up to be done simultaneously. There's a fair amount of repeated code, but the advantage is that there are some gas savings that we get from not having to do repeated transfers or verifications.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -258,6 +258,13 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         bytes32 relayHash = _getRelayHash(depositData, relayData);
         bytes memory ancillaryData = _getRelayAncillaryData(relayHash);
 
+        bytes32 instantRelayHash = _getInstantRelayHash(depositHash, relayData);
+        require(
+            // Can only speed up a pending relay without an existing instant relay associated with it.
+            instantRelays[instantRelayHash] == address(0),
+            "Relay cannot be sped up"
+        );
+
         relays[depositHash] = _getRelayDataHash(relayData);
         relayRequestAncillaryData[keccak256(ancillaryData)] = depositHash;
 
@@ -295,19 +302,12 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
 
         pendingReserves += depositData.amount; // Book off maximum liquidity used by this relay in the pending reserves.
 
-        bytes32 instantRelayHash = _getInstantRelayHash(depositHash, relayData);
-        require(
-            // Can only speed up a pending relay without an existing instant relay associated with it.
-            instantRelays[instantRelayHash] == address(0),
-            "Relay cannot be sped up"
-        );
         instantRelays[instantRelayHash] = msg.sender;
 
         // If the L1 token is WETH then: a) pull WETH from instant relayer b) unwrap WETH c) send ETH to recipient.
         uint256 recipientAmount = depositData.amount - feesTotal;
         if (isWethPool) {
-            WETH9Like(address(l1Token)).withdraw(recipientAmount);
-            depositData.l1Recipient.transfer(recipientAmount);
+            _unwrapWETHTo(depositData.l1Recipient, recipientAmount);
         }
         // Else, this is a normal ERC20 token. Send to recipient.
         else l1Token.safeTransfer(depositData.l1Recipient, recipientAmount);
@@ -448,8 +448,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         uint256 recipientAmount = depositData.amount - feesTotal;
         if (isWethPool) {
             l1Token.safeTransferFrom(msg.sender, address(this), recipientAmount);
-            WETH9Like(address(l1Token)).withdraw(recipientAmount);
-            depositData.l1Recipient.transfer(recipientAmount);
+            _unwrapWETHTo(depositData.l1Recipient, recipientAmount);
         }
         // Else, this is a normal ERC20 token. Send to recipient.
         else l1Token.safeTransferFrom(msg.sender, depositData.l1Recipient, recipientAmount);
@@ -842,6 +841,11 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
             // Canonical value representing "True"; i.e. the proposed relay is valid.
             int256(1e18)
         );
+    }
+
+    function _unwrapWETHTo(address payable to, uint256 amount) internal {
+        WETH9Like(address(l1Token)).withdraw(amount);
+        to.transfer(amount);
     }
 
     // Added to enable the BridgePool to receive ETH. used when unwrapping Weth.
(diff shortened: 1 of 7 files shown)


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,104 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+     * @notice Called by Relayer to execute a slow + fast relay from L2 to L1, fulfilling a corresponding deposit order.
+     * @dev There can only be one pending relay for a deposit. This method is effectively the relayDeposit and
+     * speedUpRelay methods concatenated. This could potentially be refactored to just call each method, but there
+     * are some gas savings in combining the transfers and hash computations.
+     * @dev Caller must have approved this contract to spend the total bond + amount - fees for `l1Token`.
+     * @param depositData the deposit data struct containing all the user's deposit information.
+     * @param realizedLpFeePct LP fee calculated off-chain considering the L1 pool liquidity at deposit time, before
+     *      quoteTimestamp. The OO acts to verify the correctness of this realized fee. Can not exceed 50%.
+     */
+    function relayAndSpeedUp(DepositData memory depositData, uint64 realizedLpFeePct) public nonReentrant() {
```
nicholaspai, line 231: DO you think we could also change the `relayDeposit` inputs to take in `DepositData memory depositData` instead of the destructured params?

packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -742,6 +752,144 @@ describe("BridgePool", () => {
...
+        .send({ from: rando });
+      await didContractThrow(
+        bridgePool.methods.relayAndSpeedUp(defaultDepositData, relayAttemptData.realizedLpFeePct).send({ from: rando })
+      );
+
+      // Reset params to allow for a normal relay.
+      await l1Token.methods.transfer(instantRelayer, instantRelayAmountSubFee).send({ from: rando });
+      await l1Token.methods.approve(bridgePool.options.address, totalRelayBond).send({ from: rando });
+
+      await bridgePool.methods.relayDeposit(...generateRelayParams()).send({ from: rando });
```
nicholaspai, line 836: nit: I think its more explicit if you call `relayDeposit` here with `...generateRelayParams({}, relayAttemptData)` to make it clear that you can slow relay the original, disputed relay, but you cannot `relayAndSpeedUp` or `speedUpRelay` it.

nicholaspai (commented):
Looks great, how's the gas savings? Left some nits on readability

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: We can -- just didn't want to modify more than necessary. I did it with the new method bc of stack too deep errors.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,104 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        // (proposer reward + proposer bond + final fee) from the caller. We need to set a new price request timestamp
+        // instead of default setting to equal to the `depositTimestamp`, which is dependent on the L2 VM on which the
+        // DepositContract is deployed. Imagine if the timestamps on the L2 have an offset that are always "in the
+        // future" relative to L1 blocks, then the OptimisticOracle would always reject requests.
+        _requestAndProposeOraclePriceRelay(proposerBond, priceRequestTime, ancillaryData);
+
+        pendingReserves += depositData.amount; // Book off maximum liquidity used by this relay in the pending reserves.
+
+        bytes32 instantRelayHash = _getInstantRelayHash(depositHash, relayData);
+        require(
```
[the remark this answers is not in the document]
mrice32: We can!

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -560,9 +665,16 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     function syncWithFinderAddresses() public {
         FinderInterface finder = FinderInterface(bridgeAdmin.finder());
 
+        // Remove approval for old OO.
+        if (address(optimisticOracle) != address(0)) l1Token.safeApprove(address(optimisticOracle), 0);
```
chrismaree, line 669: nice

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,104 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        instantRelays[instantRelayHash] = msg.sender;
+
+        // If the L1 token is WETH then: a) pull WETH from instant relayer b) unwrap WETH c) send ETH to recipient.
+        uint256 recipientAmount = depositData.amount - feesTotal;
+        if (isWethPool) {
+            WETH9Like(address(l1Token)).withdraw(recipientAmount);
+            depositData.l1Recipient.transfer(recipientAmount);
+        }
+        // Else, this is a normal ERC20 token. Send to recipient.
+        else l1Token.safeTransfer(depositData.l1Recipient, recipientAmount);
```
chrismaree, line 313: this chunk of code is repeated. refactor to internal method?

packages/core/test/insured-bridge/BridgePool.js
mrice32: Sounds good!

mrice32:
> Looks great, how's the gas savings? Left some nits on readability

Biggest gas savings are on WETH since we avoid a transfer. Outside of that, they are pretty marginal. I was seeing like 1 or 2k for an arbitrary token.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Will do!

mrice32: Actually, I think there are subtle differences that make this a little bit less tractable.

chrismaree:
> > Looks great, how's the gas savings? Left some nits on readability
> 
> Biggest gas savings are on WETH since we avoid a transfer. Outside of that, they are pretty marginal. I was seeing like 1 or 2k for an arbitrary token.

honestly I'm not sure if that amount of saving is worth it. wdyt?

mrice32:
> > > Looks great, how's the gas savings? Left some nits on readability
> > 
> > 
> > Biggest gas savings are on WETH since we avoid a transfer. Outside of that, they are pretty marginal. I was seeing like 1 or 2k for an arbitrary token.
> 
> honestly I'm not sure if that amount of saving is worth it. wdyt?

Just to tie this up. @chrismaree tested this out and found the gas savings for normal ERC20s to be around ~13k. We expect this to be larger for ETH. We've decided that the repeated code is worthwhile.

PR #3450, opened 2021-10-08
feat(across): slow relayer forfeits their reward after 15 minutes of no settlement

mrice32:
**Motivation**

There should be an incentive to settle relays quickly.

**Summary**

This adds a quick check that gives the exclusive right of the relayer to finalize their relay within 15 minutes. After that, their relay fee can be "stolen" by anyone who completes the relay.

Note: the relayer _always_ gets their bond back.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
+++ b/packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
@@ -395,7 +395,7 @@ describe("InsuredBridgeL1Client", function () {
       // Settle relay.
       await bridgePool.methods
         .settleRelay(depositData, relayAttemptData, proposeEvent.returnValues.request)
-        .send({ from: rando });
+        .send({ from: relayer });
 
       await client.update();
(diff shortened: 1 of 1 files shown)


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -375,6 +375,13 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         _validateRelayDataHash(depositHash, relayData);
         require(relayData.relayState == RelayState.Pending, "Already settled");
 
+        // Note: this check is to give the relayer a small, but reasonable amount of time to complete the relay before
+        // before it can be "stolen" by someone else. This is to ensure there is an incentive to settle relays quickly.
+        require(
+            msg.sender == relayData.slowRelayer || getCurrentTime() > request.expirationTime + 15 minutes,
```
mrice32, line 381: Note: one good thing about this check (gas-wise) is that we only pay for the second part of the check if the the original relayer is not the caller (rare).

PR #3454, opened 2021-10-08
feat(insured-bridge): Move contracts-ovm files to contracts/ and bump to ^0.8.0

nicholaspai:
**Motivation**

OVM 2.0 will support contracts compiled in ^0.8.0, and AVM already supports such contracts, so there's no longer a need to segregate contracts that can only be compiled < 0.8 for the OVM.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/HardhatConfig.ts
```diff
@@ -78,7 +74,6 @@ export function getHardhatConfig(
       goerli: { chainId: 5, url: getNodeUrl("goerli", true), accounts: { mnemonic } },
       kovan: { chainId: 42, url: getNodeUrl("kovan", true), accounts: { mnemonic } },
       optimism: {
-        ovm: true,
```
nicholaspai, line 81: I don't think we'll need this option in OVM 2.0 but I'm double checking.

packages/common/src/HardhatConfig.ts
```diff
@@ -61,7 +58,6 @@ export function getHardhatConfig(
         "contracts/insured-bridge/BridgePool.sol": LARGE_CONTRACT_COMPILER_SETTINGS,
       },
     },
-    ovm: { solcVersion: "0.7.6" },
```
nicholaspai, line 64: Also guessing this might not be needed with OVM 2.0

packages/common/src/HardhatConfig.ts
```diff
@@ -41,14 +41,11 @@ export function getHardhatConfig(
   // Some tests should not be tested using hardhat. Define all tests that end with *e2e.js to be ignored.
   const testBlacklist = [".e2e.js"];
 
-  const optimismCompileWhitelist = ["OVM_BridgeDepositBox", "Legacy_Timer", "Legacy_Lockable", "Legacy_Testable"];
+  const optimismCompileWhitelist = ["OVM_BridgeDepositBox", "Testable", "Lockable"];
```
nicholaspai, line 44: Same comment as [here](https://github.com/UMAprotocol/protocol/pull/3454/files#r725259682)

packages/core/contracts/external/ovm/OVM_CrossDomainEnabled.sol
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: MIT
```
mrice32, line 1: Is this a new contract or was it moved from somewhere?

ci/truffle_workaround.sh
```diff
@@ -1,16 +0,0 @@
-#!/bin/bash
```
mrice32, line 1: Interesting -- was this just a random find? Or is it related to the rest of the PR?

(Either way, good removal!!)

packages/common/src/hardhat/tasks/compile.ts
```diff
@@ -11,22 +9,16 @@ interface ExtendedConfig extends HttpNetworkConfig {
...
 // defaultConfig = {
 //     networks: {
-//         optimism: { ..., compileWhitelist: ["ovm"] }
+//         optimism: { ..., compileWhitelist: ["contracts_to_whitelist"] }
 //     }
 // }
 internalTask(TASK_COMPILE_SOLIDITY_GET_SOURCE_PATHS, async (_, { config, network }, runSuper) => {
   let filePaths = await runSuper();
 
-  // Temp work around to exclude some directories from truffle while compiling with hardhat. Pulls in all contracts in
```
chrismaree, line 24: 👍

packages/core/contracts/insured-bridge/BridgeDepositBox.sol
```diff
@@ -31,7 +31,7 @@ interface TokenLike {
  * @notice Accepts deposits on Optimism L2 to relay to Ethereum L1 as part of the UMA insured bridge system.
  */
 
-abstract contract BridgeDepositBox is Legacy_Testable, Legacy_Lockable {
+abstract contract BridgeDepositBox is Testable, Lockable {
```
chrismaree, line 34: One problem with this change is it blocks us from testing on the current optimism test net. This should be fine for now(we can just use the code before this pr). 

I would ask, though, that we get all remaining across PRs merged in before merging this PR. In particular, @mrice32 has two important pending prs that we should merge first. This way we can check out the code before this pr and continue to test while we wait for the next version of optimism to be launched

chrismaree (approved):
This is great! Please see my comment on not merging until all other across prs are in.

packages/core/contracts/external/ovm/OVM_CrossDomainEnabled.sol
nicholaspai: Ah copied from [here](https://github.com/ethereum-optimism/optimism/blob/5fc728da7381eda6b6f858c091636c947186a3ce/packages/contracts/contracts/optimistic-ethereum/libraries/bridge/OVM_CrossDomainEnabled.sol). I added a comment.

ci/truffle_workaround.sh
nicholaspai: This was a random find where one of the Hardhat fixtures failed and made me aware of this script

PR #3455, opened 2021-10-09
Remove the OO from the relay happy path

mrice32:
**Motivation**

We'd like to reduce gas costs as much as possible.

**Summary**

To further reduce gas costs, we can move all of the OO-related operations to the dispute method so a typical relay won't have to call into the OO.

In practice, this saves a good bit of gas. The most recent estimate from #3449 (which this PR is built on top of) leaves the total cost at `~450k`. This (fairly simple) PR brings the gas costs down to `~283k`.

Note: no tests have been updated for this PR, yet. Please comment and let me know what y'all think!

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
+++ b/packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
@@ -69,6 +69,10 @@ const defaultSlowRelayFeePct = toWei("0.01");
 const defaultInstantRelayFeePct = toWei("0.01");
 const lpFeeRatePerSecond = toWei("0.0000015");
 const finalFee = toWei("1");
+const proposerBond = toBN(defaultProposerBondPct)
+  .mul(toBN(relayAmount))
+  .div(toBN(toWei("1")))
+  .toString();
 const defaultGasLimit = 1_000_000;
 const defaultGasPrice = toWei("1", "gwei");
 const minimumBridgingDelay = 60; // L2->L1 token bridging must wait at least this time.
@@ -122,6 +126,8 @@ describe("InsuredBridgePriceFeed", function () {
         toBNWei("0.1")
       ).toString(),
       slowRelayer: relayer,
+      finalFee,
+      proposerBond,
       ...relayDataOverride,
     };
 


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -50,28 +50,32 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
     // Reserves that are not yet utilized but are pre-allocated for a pending relay.
     uint256 public pendingReserves;
 
+    // If this pool contains WETH. If the withdrawn token is WETH then unwrap and send ETH  when finalizing withdrawal.
+    bool public isWethPool;
```
mrice32, line 54: Moved these variables to save a few thousand gas since they're often read together.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,119 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+
+        // Save hash of new relay attempt parameters.
+        RelayData memory relayData =
+            RelayData({
+                relayState: RelayState.Pending,
+                slowRelayer: msg.sender, // Note: Increment numberOfRelays at the same time as setting relayId to its current value.
+                relayId: numberOfRelays++,
+                realizedLpFeePct: realizedLpFeePct,
+                priceRequestTime: priceRequestTime,
+                bond: proposerBond + finalFee
```
mrice32, line 260: One minor problem is that the OO decides the final fee at request time, so if the final fee changes between proposal and dispute, this could mean that the contract loses more or less of its balance than expected. Not sure how to get around this.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,119 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+    }
+
+    function disputeRelay(DepositData memory depositData, RelayData memory relayData) public nonReentrant() {
+        // Validate the input data.
+        bytes32 depositHash = _getDepositHash(depositData);
+        _validateRelayDataHash(depositHash, relayData);
+
+        // Submit the proposal and dispute to the OO.
+        bytes32 relayHash = _getRelayHash(depositData, relayData);
+        _requestProposeDispute(
```
mrice32, line 321: Another minor problem is what happens if the collateral token stops being whitelisted or similar. Instead of just blocking relays, this would block only disputes.

One idea is to try-catch here and automatically pause settlements if a dispute fails. Then unpause once one succeeds.

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Yeah this is an issue, perhaps you could reserve some X multiple of `bond` for now, that allows the OO final fee to change by +/-X%,  and returns to the proposer the excess after sending the final fee to the OO

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: > One idea is to try-catch here and automatically pause settlements if a dispute fails. Then unpause once one succeeds.

This is useful because if we pause it, then we can re-whitelist the collateral, and then unpause it?

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Or, we could just make an assumption that LPs take on the risk that final fee increases and they collectively pay the diff

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Exactly. The only issue is that it requires someone to issue a dispute to unpause :/, which has some weird incentives now that I think about it.

mrice32:
@nicholaspai this is really helpful. I left some thoughts below.

> My thoughts are roughly:
> 
> * The code seems a lot cleaner and uses a "blackbox oracle" better than before, in only sending a price request at dispute time. For ordinary relays, its not intuitive that there should be a price request, so this implements that idea.

Agree!

> * Our "blackbox oracle", the SkinnyOptimisticOracle, is not well suited for this construction because of how OO state can change between "relay" and "dispute" time. If OO state such as the whitelist and final fees change between "relay" and "dispute", then the contract is basically unable to use the OO. It seems that this BridgePool would work better with an OptimisticOracle that could guarantee or "snapshot" state at the time of the "relay". Imagine if the BridgePool could reserve certain OO state, (i.e. "I would like `finalFee` and `whitelist` to be frozen and need a guarantee that its state will not change in the next X seconds"), then if the BridgePool comes back within X seconds and "redeems" the reservation, then it can send an OO request with the reserved `finalFee` and `whitelist` values.
> * The problem with the above solution is that it re-introduces OptimisticOracle external calls at the time of "relay" which is sort of the point of this PR.

Exactly. That is the ideal solution, but it's messy and expensive to implement.

> * Another idea is to make guarantees in the SkinnyOO about how frequently state can change. What if we knew ahead of time that state cannot change without a 2 hour notice, and that there are no pending state changes at the moment. Therefore, one could make a relay that could be disputed via the SkinnyOO provided that the period to dispute the relay is < 2 hours.

Yeah, I thought about your timelock idea a lot, but it seems like a pretty big implementation hurdle and a decent amount of complexity in the OO. It would work, but to get the guarantees we want, it would need to track multiple values at one (old + new) and basically allow you to pick between them.

> * Another idea is to change the SkinnyOO to only set certain values at dispute time like the `finalFee` and not pull that from the proposer until the `dispute` time.

The trouble with that is that the proposer can move the money out of their account (or cancel the approval) before the dispute. Also, it's a little non-kosher to use approvals to pull funds in a transaction that the user didn't initiate.


My idea (that is inspired by your comment and now in the PR) is to:
1. Store the bond and final fee separately in the request struct at proposal time.
2. Approve exactly the bond + final fee before the proposal.
3. Put a try-catch around the proposal call.
4. If it fails, refund the relayer and delete the proposal (throw it out and assume that nobody was right or wrong). Don't do any of the following steps. Note: we want to return as early as possible so the cancellation isn't super expensive to trigger.
5. If it succeeds, but pulls _less_ than the proposer initially sent, refund the proposer the remainder and remember the _new_ bond amount.
6. Pull funds from disputer (corrected bond amount if necessary).
7. Approve exactly the bond amount (corrected if necessary).
8. Send the dispute.

Note: this also handles cases of de-whitelisting, etc. Although those are a bit scarier since it could be days that bots need to dispute at no profit to protect the contract. I see this as unlikely, mitigable (by spending some ETH), and costly for the griefer, so I don't think it's a huge risk.

Thoughts?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -459,29 +562,6 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
         // Clean up state storage and receive gas refund. This also prevents `priceDisputed()` from being able to reset
         // this newly Finalized relay state.
         delete instantRelays[instantRelayHash];
-        delete relayRequestAncillaryData[keccak256(ancillaryData)];
```
chrismaree, line 462: The simplification on these methods is nice.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -723,37 +787,101 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+                // If the OO pulls less (due to a change in final fee), refund the proposer.
+                uint256 refund = totalBond - bondSpent;
+                l1Token.safeTransfer(proposer, refund);
+                l1Token.safeApprove(address(optimisticOracle), 0);
+            }
+        } catch {
+            // If there's an error in the OO, this means something has changed to make this request undisputable.
+            // To ensure the request does not go through by default, refund the prposer and return early, allowing
+            // the calling method to delete the request, but with no additional recourse by the OO.
+            l1Token.safeTransfer(proposer, totalBond);
```
chrismaree, line 828: Do we want to emit a particular error in the catch case?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -723,37 +787,101 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
     }
 
     // Added to enable the BridgePool to receive ETH. used when unwrapping Weth.
     receive() external payable {}
 }
+
+/**
+ * @notice This is the BridgePool contract that should be deployed on live networks. It is exactly the same as the
+ * regular BridgePool contract, but it overrides getCurrentTime to make the call a simply return block.timestamp with
+ * no branching or storage queries.
```
chrismaree, line 871:
```suggestion
 * no branching or storage queries. This is done to save gas. 
```

chrismaree:
> @nicholaspai this is really helpful. I left some thoughts below.
> 
> 
> 
> > My thoughts are roughly:
> 
> > 
> 
> > * The code seems a lot cleaner and uses a "blackbox oracle" better than before, in only sending a price request at dispute time. For ordinary relays, its not intuitive that there should be a price request, so this implements that idea.
> 
> 
> 
> Agree!
> 
> 
> 
> > * Our "blackbox oracle", the SkinnyOptimisticOracle, is not well suited for this construction because of how OO state can change between "relay" and "dispute" time. If OO state such as the whitelist and final fees change between "relay" and "dispute", then the contract is basically unable to use the OO. It seems that this BridgePool would work better with an OptimisticOracle that could guarantee or "snapshot" state at the time of the "relay". Imagine if the BridgePool could reserve certain OO state, (i.e. "I would like `finalFee` and `whitelist` to be frozen and need a guarantee that its state will not change in the next X seconds"), then if the BridgePool comes back within X seconds and "redeems" the reservation, then it can send an OO request with the reserved `finalFee` and `whitelist` values.
> 
> > * The problem with the above solution is that it re-introduces OptimisticOracle external calls at the time of "relay" which is sort of the point of this PR.
> 
> 
> 
> Exactly. That is the ideal solution, but it's messy and expensive to implement.
> 
> 
> 
> > * Another idea is to make guarantees in the SkinnyOO about how frequently state can change. What if we knew ahead of time that state cannot change without a 2 hour notice, and that there are no pending state changes at the moment. Therefore, one could make a relay that could be disputed via the SkinnyOO provided that the period to dispute the relay is < 2 hours.
> 
> 
> 
> Yeah, I thought about your timelock idea a lot, but it seems like a pretty big implementation hurdle and a decent amount of complexity in the OO. It would work, but to get the guarantees we want, it would need to track multiple values at one (old + new) and basically allow you to pick between them.
> 
> 
> 
> > * Another idea is to change the SkinnyOO to only set certain values at dispute time like the `finalFee` and not pull that from the proposer until the `dispute` time.
> 
> 
> 
> The trouble with that is that the proposer can move the money out of their account (or cancel the approval) before the dispute. Also, it's a little non-kosher to use approvals to pull funds in a transaction that the user didn't initiate.
> 
> 
> 
> 
> 
> My idea (that is inspired by your comment and now in the PR) is to:
> 
> 1. Store the bond and final fee separately in the request struct at proposal time.
> 
> 2. Approve exactly the bond + final fee before the proposal.
> 
> 3. Put a try-catch around the proposal call.
> 
> 4. If it fails, refund the relayer and delete the proposal (throw it out and assume that nobody was right or wrong). Don't do any of the following steps. Note: we want to return as early as possible so the cancellation isn't super expensive to trigger.
> 
> 5. If it succeeds, but pulls _less_ than the proposer initially sent, refund the proposer the remainder and remember the _new_ bond amount.
> 
> 6. Pull funds from disputer (corrected bond amount if necessary).
> 
> 7. Approve exactly the bond amount (corrected if necessary).
> 
> 8. Send the dispute.
> 
> 
> 
> Note: this also handles cases of de-whitelisting, etc. Although those are a bit scarier since it could be days that bots need to dispute at no profit to protect the contract. I see this as unlikely, mitigable (by spending some ETH), and costly for the griefer, so I don't think it's a huge risk.
> 
> 
> 
> Thoughts?

This flow seems good. I like using limited approvals to block particular actions in the case params change up stream. A downside with this is it does increase gas costs as we need to made additional approval calls than we otherwise would have. 

A softer solution to all this is _assume_ things won't change and if we need to change them (final fee, whitelist etc) we simply pause the contracts until the changes are made. This will result in the simplest possible contracts which will also be the cheapest to operate but does introduce administrative burden down the line making it less than ideal. Only reason I'm even mention this is in the whole time we've had eth whitlisted we've only change the final fee one time so it's not like it's a very common action.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +219,135 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        require(relays[depositHash] == bytes32(0), "Pending relay exists");
+
+        uint256 proposerBond = _getProposerBond(depositData.amount);
+        uint256 finalFee = store.computeFinalFee(address(l1Token)).rawValue;
+
+        // Save hash of new relay attempt parameters.
+        RelayData memory relayData =
+            RelayData({
+                relayState: RelayState.Pending,
+                slowRelayer: msg.sender, // Note: Increment numberOfRelays at the same time as setting relayId to its current value.
```
chrismaree, line 257: This comment is on the wrong line

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +219,135 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        uint256 totalBond = proposerBond + finalFee;
+        l1Token.safeTransferFrom(msg.sender, address(this), depositData.amount + totalBond);
+        bonds += totalBond;
+
+        // Request a price for the relay identifier and propose "true" optimistically. This method will pull the
+        // (proposer reward + proposer bond + final fee) from the caller. We need to set a new price request timestamp
+        // instead of default setting to equal to the `depositTimestamp`, which is dependent on the L2 VM on which the
+        // DepositContract is deployed. Imagine if the timestamps on the L2 have an offset that are always "in the
+        // future" relative to L1 blocks, then the OptimisticOracle would always reject requests.
+        // _requestAndProposeOraclePriceRelay(proposerBond, priceRequestTime, ancillaryData);
```
chrismaree, line 285: Commented out line. Not relevant in your implementation. Same true for the comment above

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +219,135 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+
+        pendingReserves += depositData.amount; // Book off maximum liquidity used by this relay in the pending reserves.
+
+        bytes32 instantRelayHash = _getInstantRelayHash(depositHash, relayData);
+        require(
```
chrismaree, line 290: Surly this require is redundant for relay and speed up? You have already validated this by the fact that there is no relay data, no? Alternatively, If you think you need this check, you should put it higher up if possible to revert sooner

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: Ah! Nice catch!

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: haha, ty

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: I think this check is necessary. You could have a previous identical relay that was falsely disputed that was already sped up.

Will move up (@nicholaspai made a similar comment in the other PR)!

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +218,119 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
      *          RELAYER FUNCTIONS         *
      **************************************/
 
+    /**
+     * @notice Called by Relayer to execute a slow + fast relay from L2 to L1, fulfilling a corresponding deposit order.
```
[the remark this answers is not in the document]
mrice32: Will do!

packages/core/contracts/insured-bridge/BridgePool.sol
chrismaree: like emit an error in the catch case to indicate that the happy path did not happen.

chrismaree: you could emit something in the non-catch case (happy path) and then emit a separate require in the event it reverts. this way we can detect if something undesirable happened. otherwise, it's somewhat hard to figure this out

chrismaree: it would be a standard event but just something named to indicate that the desired flow did not happen.

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -218,6 +219,130 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, MultiCaller
...
+        uint256 recipientAmount = depositData.amount - feesTotal;
+        if (isWethPool) {
+            WETH9Like(address(l1Token)).withdraw(recipientAmount);
+            depositData.l1Recipient.transfer(recipientAmount);
+        }
+        // Else, this is a normal ERC20 token. Send to recipient.
+        else l1Token.safeTransfer(depositData.l1Recipient, recipientAmount);
+
+        emit DepositRelayed(depositHash, depositData, address(l1Token), relayData, relayHash);
+        emit RelaySpedUp(depositHash, msg.sender, relayData);
```
chrismaree, line 308: do we want an event that indicates this is a `relayAndSpeedUp` more than a sprat `DepositRelayed` and `RelaySpedUp`? this makes it a bit hard to know that it was the `relayAndSpeedUp` function called.

Also, emitting two events costs more gas

chrismaree (approved):

PR #3457, opened 2021-10-11
perf(api): optimize api for Google datastore

amateima:
**Motivation**

https://app.shortcut.com/uma-project/story/2493/optimize-api-for-googlestore

**Summary**

The AppState for Datastore API application has a new table `appStats`, used to keep track of various variables that need to be pulled out from the in-memory state. Now, the number of the last block updated is kept in the appStats table.
More over, since the application state is not kept in the memory anymore, we don't need to initialise the state at launch.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/api/src/apps/datastore_api/app.ts
```diff
@@ -154,27 +156,6 @@ export default async (env: ProcessEnv) => {
...
-  console.log("Updated LSP state");
-
-  // we've update our state based on latest block we queried
-  appState.lastBlockUpdate = initBlock.number;
-
-  await services.erc20s.update();
-  console.log("Updated tokens");
-
   // backfill price histories, disable if not specified in env
   if (env.backfillDays) {
```
daywiss, line 160: we might need to either get rid of this, or set some kind of flag where this doesnt run if there is already data in there. or maybe we just remmeber to remove the env once it runs once

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -154,27 +156,6 @@ export default async (env: ProcessEnv) => {
     globalStats: Services.stats.Global({ debug }, appState),
   };
 
-  const initBlock = await provider.getBlock("latest");
-
-  // warm caches
-  await services.registry(appState.lastBlockUpdate, initBlock.number);
```
daywiss, line 160: this might be tricky but these state updates need to run in a certain order, and the loops introduce some race conditions.  would feel good if you could run a test against empty DB and make sure it eventually gets into a good state. we can also test this on dev and do queries against it

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -279,11 +243,14 @@ export default async (env: ProcessEnv) => {
...
 
-  const newContractBlockTick = BlockInterval(detectNewContracts, initBlock.number);
-  const updateContractStateTick = BlockInterval(updateContractState, initBlock.number);
+  const lastBlockUpdate = await appState.appStats.getLastBlockUpdate();
+  const newContractBlockTick = BlockInterval(detectNewContracts, lastBlockUpdate);
+  const updateContractStateTick = BlockInterval(updateContractState, lastBlockUpdate);
 
   // detect contract loop
   utils.loop(async () => {
+    if (appState.loopsLock) return;
```
mrice32, line 252: What does loopsLock do?

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -123,6 +124,8 @@ export default async (env: ProcessEnv) => {
       active: lsps.Table("Active LSP", datastores.lspsActive),
       expired: lsps.Table("Expired LSP", datastores.lspsExpired),
     },
+    appStats: appStats.Table("App Stats", datastores.appStats),
+    loopsLock: false,
```
mrice32, line 128: Just want to be sure -- does this imply that we _always_ use serverless and the datastore together? i.e. there's no way to run a looping api that ingests to the datastore?

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -154,57 +157,21 @@ export default async (env: ProcessEnv) => {
...
-  console.log("Updated tokens");
-
   // backfill price histories, disable if not specified in env
   if (env.backfillDays) {
     console.log(`Backfilling price history from ${env.backfillDays} days ago`);
     await services.collateralPrices.backfill(moment().subtract(env.backfillDays, "days").valueOf());
     console.log("Updated Collateral Prices Backfill");
+
+    // backfill price history only if runs for the first time
+    if (await appState.appStats.getLastBlockUpdate()) return;
```
daywiss, line 167: i dont think we can just return early here because the rest of the app needs to init. maybe we just add this check to line 161 to not enter this block if it exists

packages/api/src/apps/datastore_api/app.ts
```diff
@@ -297,10 +264,13 @@ export default async (env: ProcessEnv) => {
     )
       .catch(console.error)
       .finally(end);
+    appState.loopsLock = false;
   }, detectContractsUpdateRateS * 1000);
 
   // main update loop for all state, executes immediately and waits for updateRateS
   utils.loop(async () => {
+    if (appState.loopsLock) return;
```
daywiss, line 272: tbh im not a huge fan of this solution. it only really matters that we detect contracts before we check the state for the first time.  maybe we just initialize the second loop after the first runs or something. or maybe we run the functions in the loops in order and then init the loops, similar to how it was before

amateima:
> i dont think we can just return early here because the rest of the app needs to init. maybe we just add this check to line 161 to not enter this block if it exists
@daywiss  I replaced the return with a conditional check. Wasn't my intention to stop the execution of the program 😄 One mention tho, collateralPrices are still kept in memory so those ones need to be backfilled

packages/api/src/apps/datastore_api/app.ts
amateima: I replaced the return with a conditional check. Wasn't my intention to stop the execution of the program 😄 One mention tho, collateralPrices are still kept in memory so those ones need to be backfilled

packages/api/src/apps/datastore_api/app.ts
amateima: I switched back to the older implementation with running the functions before the loops. Just re-organised the code a little bit. Maybe you can double check the functions execution order

packages/api/src/apps/datastore_api/app.ts
amateima: The usage of serverless is not tied to storing the data in Google Datastore

PR #3460, opened 2021-10-12
feat(across): war games 2

chrismaree:
**Motivation**

Additional testing was done on kovan to ensure our latest contracts work as expected. This PR makes a few updates after these tests.

**Summary**

Updates contracts and scripts after additional kovan testing.


packages/core/contracts/insured-bridge/ovm/OVM_CrossDomainEnabled.sol
```diff
@@ -6,16 +6,16 @@
 pragma solidity ^0.8.0;
 
 /* Interface Imports */
-import "./iOptimism_CrossDomainMessenger.sol";
+import "./iOVM_CrossDomainMessenger.sol";
 
 /**
- * @title Optimism_CrossDomainEnabled
+ * @title OVM_CrossDomainEnabled
```
chrismaree, line 12: I experience the strangest bug. when these contracts were called `Optimism_x` they would generate errors when calling the canonical bridge. after renaming them `OVM_x` it worked again. I have no idea why this is the case. I tested out a file named `Chandler_x` and it also reverted in the same way...it makes no sense but it _does_ work with `OVM_x`

packages/common/src/hardhat/tasks/acrossPool.ts
```diff
@@ -8,27 +8,30 @@ task("deploy-across-pool", "Deploys an L1 across pool and whitelists it within t
...
     const { deployments, getNamedAccounts, web3 } = hre;
     const { deploy } = deployments;
     const { deployer } = await getNamedAccounts();
-    const { lptokenname, lptokensymbol, l1tokenaddress, lpfeeratepersecond } = taskArguments;
+    const { lptokenname, lptokensymbol, l1tokenaddress, lpfeeratepersecond, iswethpool } = taskArguments;
 
     const BridgeAdmin = await deployments.get("BridgeAdmin");
     const bridgeAdmin = new web3.eth.Contract(BridgeAdmin.abi, BridgeAdmin.address);
     console.log(`Loaded BridgeAdmin @ ${bridgeAdmin.options.address}`);
-    console.log(`Deploying bridgePool`);
```
nicholaspai, line 21: Why not make this script a `deploy` script?

chrismaree: need params!

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -633,9 +633,13 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, Lockable {
 
         // The liquidity utilization ratio is the ratio of utilized liquidity (pendingReserves + relayedAmount
         // +utilizedReserves) divided by the liquid reserves.
-        uint256 numerator = pendingReserves + relayedAmount;
-        if (utilizedReserves > 0) numerator += uint256(utilizedReserves);
-        else numerator -= uint256(utilizedReserves * -1);
+        int256 numerator = int256(pendingReserves + relayedAmount);
```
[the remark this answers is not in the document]
mrice32: Should we add a check here to make sure this cast is safe?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -645,7 +649,7 @@ contract BridgePool is Testable, BridgePoolInterface, ExpandedERC20, Lockable {
         if (numerator > 0 && liquidReserves == 0) return 1e18;
 
         // In all other cases, return the utilization ratio.
-        return (numerator * 1e18) / liquidReserves;
+        return (uint256(numerator) * 1e18) / liquidReserves;
```
mrice32, line 652: Same question here about the cast?

packages/core/contracts/insured-bridge/BridgePool.sol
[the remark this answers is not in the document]
chrismaree: will test this and get back to you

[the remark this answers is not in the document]
chrismaree: after playing around with this I'm pretty sure that this does not introduce any additional risks as the user is never given the ability to feed in this kind of data. all conversions happen in internal methods in a safe way.

packages/core/contracts/insured-bridge/BridgePool.sol
mrice32: That seems sufficient!

mrice32 (approved):
LGTM!

PR #3462, opened 2021-10-12
feat(sdk): add bridge pool contract client

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/2473/add-bridgepool-contract-to-sdk

**Summary**
This adds a mostly done client for the bridge pool, but unsure how to deal with decoding some events.  Will comment on them in pr.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2473/add-bridgepool-contract-to-sdk

### DIFF
--- a/packages/sdk/src/clients/bridgePool/client.ts
+++ b/packages/sdk/src/clients/bridgePool/client.ts
@@ -1,0 +1,167 @@
+import { BridgePoolEthers, BridgePoolEthers__factory } from "@uma/contracts-node";
+import type { SignerOrProvider, GetEventType } from "../..";
+import { Event, BigNumber } from "ethers";
+import { Balances } from "../../utils";
+
+export type Instance = BridgePoolEthers;
+const Factory = BridgePoolEthers__factory;
+
+export function connect(address: string, provider: SignerOrProvider): Instance {
+  return Factory.connect(address, provider);
+}
+
+export type LiquidityAdded = GetEventType<Instance, "LiquidityAdded">;
+export type LiquidityRemoved = GetEventType<Instance, "LiquidityRemoved">;
+export type DepositRelayed = GetEventType<Instance, "DepositRelayed">;
+export type RelaySpedUp = GetEventType<Instance, "RelaySpedUp">;
+export type RelayDisputed = GetEventType<Instance, "RelayDisputed">;
+export type RelayCanceled = GetEventType<Instance, "RelayCanceled">;
+export type RelaySettled = GetEventType<Instance, "RelaySettled">;
+export type BridgePoolAdminTransferred = GetEventType<Instance, "BridgePoolAdminTransferred">;
+
+export type DepositData = {
+  chainId: number;
+  depositId: BigNumber;
+  l1Recipient: string;
+  l2Sender: string;
+  amount: BigNumber;
+  slowRelayFeePct: BigNumber;
+  instantRelayFeePct: BigNumber;
+  quoteTimestamp: number;
+};
+export type RelayData = {
+  relayState: number;
+  slowRelayer: string;
+  relayId: number;
+  realizedLpFeePct: BigNumber;
+  priceRequestTime: number;
+  proposerBond: BigNumber;
+  finalFee: BigNumber;
+};
+export interface EventState {
+  oldAdmin?: string;
+  newAdmin?: string;
+  tokens: Record<string, Balances>;
+  lpTokens: Record<string, Balances>;
+  deposits: Record<string, DepositData>;
+  relays: Record<string, RelayData>;
+  instantRelays: Record<string, string>;
+}
+
+function eventStateDefaults() {
+  return {
+    tokens: {},
+    lpTokens: {},
+    deposits: {},
+    relays: {},
+    instantRelays: {},
+  };
+}
+
+export function reduceEvents(state: EventState, event: Event): EventState {
+  switch (event.event) {
+    // event LiquidityAdded(address indexed token, uint256 amount, uint256 lpTokensMinted, address liquidityProvider);
+    case "LiquidityAdded": {
+      const typedEvent = event as LiquidityAdded;
+      const { token, amount, lpTokensMinted, liquidityProvider } = typedEvent.args;
+      const tokens = Balances(state.tokens[token] || {});
+      const lpTokens = Balances(state.lpTokens[token] || {});
+      tokens.add(liquidityProvider, amount.toString());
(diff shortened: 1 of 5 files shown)


packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
...
+}
+
+export function reduceEvents(state: EventState, event: Event): EventState {
+  switch (event.event) {
+    // event LiquidityAdded(address indexed token, uint256 amount, uint256 lpTokensMinted, address liquidityProvider);
+    case "LiquidityAdded": {
+      const typedEvent = event as LiquidityAdded;
+      const { token, amount, lpTokensMinted, liquidityProvider } = typedEvent.args;
+      const tokens = Balances(state.tokens[token] || {});
+      const lpTokens = Balances(state.lpTokens[token] || {});
```
daywiss, line 68: I am assuming that token balances and lp token balances are distinct for each token address

packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
...
+        lpTokens: {
+          ...state.lpTokens,
+          [token]: lpTokens.balances,
+        },
+      };
+    }
+    // event DepositRelayed( bytes32 indexed depositHash, DepositData depositData, address l1Token, RelayData relay, bytes32 relayAncillaryDataHash);
+    case "DepositRelayed": {
+      const typedEvent = event as DepositRelayed;
+      const { depositHash, depositData, relay, relayAncillaryDataHash } = typedEvent.args;
```
daywiss, line 106: unsure how to use the `l1Token` parameter in the event

packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
...
+      // TODO: finish implementation
+      // const typedEvent = event as RelayDisputed;
+      // const { depositHash, relayHash, disputer } = typedEvent.args;
+      return state;
+    }
+    // event RelayCanceled(bytes32 indexed depositHash, bytes32 indexed relayHash, address indexed disputer);
+    case "RelayCanceled": {
+      // TODO: finish implementation
+      // const typedEvent = event as RelayCanceled;
+      // const { depositHash, relayHash, disputer } = typedEvent.args;
```
daywiss, line 142: unsure how to signify relay canceled in state

packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
...
+      return state;
+    }
+    // event RelaySettled(bytes32 indexed depositHash, address indexed caller, RelayData relay);
+    case "RelaySettled": {
+      // TODO: finish implementation
+      // const typedEvent = event as RelaySettled;
+      // const { depositHash, caller, relay } = typedEvent.args;
+      return state;
```
daywiss, line 150: unsure of how to signify relay settled

packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
+import { BridgePoolEthers, BridgePoolEthers__factory } from "@uma/contracts-node";
```
mrice32, line 1: It's not the exact same, but it seems that we're duplicating a lot of work by building bot clients in `financial-templates-lib` and then a frontend-focused client in the sdk that have a lot of repeated functionality (both primarily track state using events). See https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts.

@daywiss I know it's complicated to bridge the gap in these packages, but can you think of a _process_ s.t. we could unify these to take advantage of shared work? Maybe we could build all new clients in the sdk and import them in the bots?

Thoughts?

mrice32 (commented):
Just one high-level comment

packages/sdk/src/clients/bridgePool/client.ts
daywiss: yea this is a tough question. the financial templates clients are written with a specific purpose in mind and therefore have a lot of functionality built into them, much more than whats provided in the sdk.  One place though where there could be more alignment is around the shape of event data, ie having a common definition of the type of state that events should generate.  From there we could have a common event processor that could be shared between code bases, which is really the only value the sdk is providing at moment on contracts.

Anything more than that though, for instance if financial-templates relied on featureful clients from sdk, i would fear that it could create a bottleneck in the bot work.  If this is important enough though, i think we could put more effort into this.

mrice32: Even if one is a superset of the other or they're fairly disjoint, I think we could see real advantages from using the same underlying client and processing logic with the ability to turn on or off certain features or extend/compose the classes.

I think what we're doing now is "shipping the org chart", where the code reflects the organizational structure of those who wrote it. We should really avoid the tendency to write each application or feature completely fresh and re-use our existing work.

I think this is the same thing that has gotten us into trouble previously with price feeds where different applications support different types of price feeds because they didn't all use the same underlying implementation.

What's the urgency of this PR? Can we unify these now?

mrice32 (commented):
Just more on that thread!

packages/sdk/src/clients/bridgePool/client.ts
```diff
@@ -0,0 +1,167 @@
...
+          ...(state.instantRelays || {}),
+          [depositHash]: instantRelayer,
+        },
+      };
+    }
+    // event RelayDisputed(bytes32 indexed depositHash, bytes32 indexed relayHash, address indexed disputer);
+    case "RelayDisputed": {
+      // TODO: finish implementation
+      // const typedEvent = event as RelayDisputed;
+      // const { depositHash, relayHash, disputer } = typedEvent.args;
```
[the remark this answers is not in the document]
mrice32: Can you find and modify the relay? Or is this append-only state? If append-only, can you find and then copy the relay and append a new relay that is the same, but disputed?

packages/sdk/src/clients/bridgePool/client.ts
mrice32: Same q as above.

packages/sdk/src/clients/bridgePool/client.ts
mrice32: Same comment as above.

mrice32 (approved):
Looks good!

PR #3467, opened 2021-10-14
feat(across): enable weth transfers to smart contracts

chrismaree:
**Motivation**

If the recipient of a bridging action is a smart contract, it could revert on receiving ETH. This PR adds a check to validate if the recipient is a smart contract. If it is then WETh is sent over ETH.

**Summary**

Note that a try/catch block could be added here. However, this introduces some inconsistent behaviour. I opted to _always_ sent WETH to smart contracts over _sometimes_ sending eth, if the recipient contract can accept it.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


nicholaspai (approved):
Nice tests!

PR #3470, opened 2021-10-14
improve(insured-bridge-relayer): Refactor main loop logic in preparation for dispute functionality

nicholaspai:
…

**Motivation**

- Renamed enums
- Refactored `checkForPendingDepositsAndRelay` into `getRelayableDeposits`, `getRelayStatus`, and `sendRelayTransaction` internal methods that process relayable deposits from L2, check pending relay status for those deposits, and submit relay transactions respectively.
- Added logic to skip relayable deposits that have pending EXPIRED relays.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
+++ b/packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
@@ -301,6 +301,9 @@ describe("InsuredBridgeL1Client", function () {
     // Before the client is updated, client should error out.
     assert.throws(client.getBridgePoolsAddresses, Error);
 
+    // Before updating, optimistic oracle liveness defaults to 0.
+    assert.equal(client.optimisticOracleLiveness, 0);
+
     // After updating the client it should contain the appropriate addresses.
     await client.update();
     assert.equal(JSON.stringify(client.getBridgePoolsAddresses()), JSON.stringify([bridgePool.options.address]));
@@ -315,6 +318,9 @@ describe("InsuredBridgeL1Client", function () {
       client.getBridgePoolForDeposit(depositData).relayNonce,
       (await bridgePool.methods.numberOfRelays().call()).toString()
     );
+
+    // OptimisticOracle liveness should be reset.
+    assert.equal(client.optimisticOracleLiveness, defaultLiveness);
   });
   describe("Lifecycle tests", function () {
     it("Relayed deposits: deposit, speedup finalize lifecycle", async function () {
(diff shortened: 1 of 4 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -42,66 +47,73 @@ export class Relayer {
...
 
-  // TODO: consider refactoring this further into multiple methods. One to build the object of relayableDeposits and one
-  // that processes relayableDeposits and calls respective relaying functions.
-  async checkForPendingDepositsAndRelay() {
+  async checkForPendingDepositsAndRelay(): Promise<undefined> {
     this.logger.debug({ at: "Relayer", message: "Checking for pending deposits and relaying" });
-    const relayableDeposits: { [key: string]: [{ status: ClientRelayState; deposit: Deposit }] } = {};
-    for (const l1Token of this.whitelistedRelayL1Tokens) {
-      this.logger.debug({ at: "Relayer", message: "Checking relays for token", l1Token });
-      // TODO: consider limiting how far back we look in this call size.
```
nicholaspai, line 52: @chrismaree I think this comment can be ignored now since we pass an `l2StartBlock` into the L2 relay client. We can also consider adding an `l1StartBlock` but I haven't had issues querying L1 events.

packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
```diff
@@ -315,6 +318,9 @@ describe("InsuredBridgeL1Client", function () {
       client.getBridgePoolForDeposit(depositData).relayNonce,
       (await bridgePool.methods.numberOfRelays().call()).toString()
     );
+
+    // OptimisticOracle liveness should be reset.
+    assert.equal(client.optimisticOracleLiveness, defaultLiveness);
```
chrismaree, line 323: 👍

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -42,66 +47,73 @@ export class Relayer {
...
             this.logger.debug({
               at: "Relayer",
-              message: "Pending relay is invalid, ignoring",
+              message: "Pending relay is invalid",
               pendingRelay,
               relayableDeposit,
-              reason: isRelayValid.reason,
+              reason: relayStatus.relayDisputable.reason,
             });
+            // TODO: Optionally dispute the relay here if DISPUTE_MODE=true
```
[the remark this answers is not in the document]
nicholaspai: I think Relay and Dispute are related since we should only speed up VALID relays, and determining whether a relay is valid is the same logic for determining when to dispute.

PR #3473, opened 2021-10-15
fix(insured-bridge): remove pending reserves on dispute

mrice32:
**Motivation**

All successful disputes (those that cancel and those that trigger an OO request) should remove the pending reserves that were being set aside for the payout at settlement.

**Summary**

This adds a single line to the contract to subtract off the deposit amount in that case.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/contracts/insured-bridge/test/OVM_L1CrossDomainMessengerMock.sol
```diff
@@ -11,6 +11,6 @@ contract OVM_L1CrossDomainMessengerMock is iOVM_CrossDomainMessenger {
         address _target,
         bytes calldata _message,
         uint32 _gasLimit
-    ) external {}
+    ) external override {}
```
mrice32, line 14: Not sure why I had to add this -- was having trouble compiling locally without it.

nicholaspai (approved):
LGTM

PR #3474, opened 2021-10-15
feat(across): Add Disputer mode to bot

nicholaspai:
**Motivation**

Relayer bot now has a Disputer and Relayer mode which enable dispute and slow/speedup relay functionality respectively.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
@@ -44,12 +44,10 @@ export class InsuredBridgeL2Client {
   }
 
   getAllDeposits() {
-    this._throwIfNotInitialized();
     return Object.keys(this.deposits).map((depositId: string) => this.deposits[depositId]);
   }
 
   getDepositByID(depositId: string | number) {
-    this._throwIfNotInitialized();
     return this.deposits[depositId.toString()];
   }
 
@@ -114,9 +112,4 @@ export class InsuredBridgeL2Client {
     if (depositHash == "" || depositHash == null) throw new Error("Bad deposit hash");
     return depositHash;
   };
-
-  private _throwIfNotInitialized() {
-    if (Object.keys(this.deposits).length == 0)
-      throw new Error("InsuredBridgeClient method called before initialization! Call `update` first.");
-  }
 }


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -392,6 +417,44 @@ export class Relayer {
...
+            tx: receipt.transactionHash,
+            depositHash: receipt.events.RelayDisputed.returnValues.depositHash,
+            relayHash: receipt.events.RelayDisputed.returnValues.relayHash,
+            disputer: receipt.events.RelayDisputed.returnValues.disputer,
+            transactionConfig,
+          });
+        } else if (receipt.events.RelayCanceled) {
+          this.logger.info({
+            at: "InsuredBridgeRelayer#Relayer",
+            type: "Dispute failed to send to OO. Relay was deleted. 🧑🏻‍✈️",
```
chrismaree, line 444: diff emoji please

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -392,6 +417,44 @@ export class Relayer {
...
+            tx: receipt.transactionHash,
+            depositHash: receipt.events.RelayCanceled.returnValues.depositHash,
```
chrismaree, line 446: are there more params we can include in this?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -37,14 +37,19 @@ export class Relayer {
...
    */
   constructor(
     readonly logger: winston.Logger,
     readonly gasEstimator: GasEstimator,
     readonly l1Client: InsuredBridgeL1Client,
     readonly l2Client: InsuredBridgeL2Client,
     readonly whitelistedRelayL1Tokens: string[],
-    readonly account: string
+    readonly account: string,
+    readonly relayerEnabled: boolean,
```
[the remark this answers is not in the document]
nicholaspai: I implemented (1) PTAL

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,16 +97,32 @@ export class Relayer {
               relayableDeposit,
               reason: relayStatus.relayDisputable.reason,
             });
-            // TODO: Optionally dispute the relay here if DISPUTE_MODE=true
+            // If Disputer feature is enabled, submit a dispute on-chain. Otherwise skip to the next deposit.
+            if (this.botModes.disputerEnabled) {
```
mrice32, line 101: We're only considering _existing deposits_ in this loop, right? What happens if someone relays a completely fabricated deposit. For instance, a totally invalid chainId or a deposit Id that is way past the current deposit Id?

nicholaspai: > We're only considering existing deposits in this loop, right?

Yep IIUC. The outermost loop considers all `relayableDeposits` which are constructed from the list of `FundsDeposited` events emitted by the `BridgeDepositBox`.

> What happens if someone relays a completely fabricated deposit
If the deposit parameters were not emitted from a `BridgeDepositBox` then we won't consider them in this loop.

And, if someone relays incorrectly a _valid_ deposit, then this PR will either skip or dispute the relay.

mrice32: Isn't this a problem? If someone fabricates a non-existent deposit, it wouldn't be disputed by this bot?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -267,7 +371,7 @@ export class Relayer {
 
       if (receipt.events)
         this.logger.info({
-          at: "InsuredBridgeRelayer#Relayer",
+          at: "Relayer",
```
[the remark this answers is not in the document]
chrismaree: actually how you have it is cleaner

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -76,7 +76,27 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
           // Update state.
           await Promise.all([gasEstimator.update(), l1Client.update(), l2Client.update()]);
 
-          await relayer.checkForPendingDepositsAndRelay();
+          if (config.botModes.relayerEnabled) {
```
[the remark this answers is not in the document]
chrismaree: I think you can also clean up this syntax a lot if you use inline ifs with no `{}` and collapse the log messages to one line

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -138,12 +132,120 @@ export class Relayer {
         );
       }
     }
+
+    return;
+  }
+
+  async checkForPendingRelaysAndDispute(): Promise<undefined> {
```
mrice32, line 139:
```suggestion
  async checkForPendingRelaysAndDispute(): Promise<void> {
```

nit: prefer void for return values for functions that don't return anything.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -138,12 +132,120 @@ export class Relayer {
...
+      return;
+    }
+    this.logger.debug({
+      at: "Disputer",
+      message: `Processing ${pendingRelays.length} pending relays`,
+    });
+
+    for (const relay of pendingRelays) {
+      // Construct deposit from relay params if the deposit is undefined.
+      const deposit =
```
mrice32, line 156: This looks like it looks at _all_ relays, but then only checks against a _single_ l2 client? If the L2 client is pointed at arbitrum, but the bot sees a valid relay from optimism, will it dispute it?

Put another way, how do we deal with multiple L2s here?

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: done

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: Really good catch @mrice32 , with the latest commits do you think there are any weird edge cases now where we can misconfigure the L2/L1 clients and create false disputes unintentionally?

@chrismaree worth reviewing this thread and re reviewing

PR #3475, opened 2021-10-15
fix(insured-bridge): Handle case where OO.dispute reverts after OO.requestAndPropose succeeds

nicholaspai:
**Motivation**

This edge case is possible if the `OO` is unregistered to submit price requests to the Voting contract. Without catching the `OO.dispute` call, this situation causes pending relays to be undisputable, but still sends proposals to the OO, which is a bad situation.

This PR adds a method to the SkinnyOO that requests+proposes+disputes atomically. Previously, `requestAndProposePriceFor` and `disputePrice` had to be called sequentially, since Solidity cannot group two external calls into the same try-catch block. Adding this method `requestProposeAndDisputePriceFor` prevents the situation where `requestAndProposePriceFor` succeeds but `disputePrice` reverts, resulting in a useless pending proposal in the OO that can't be disputed. 

Combining the three OO methods also reduces gas cost for the `disputeRelay` method moderately:

```
Previous:
|  Contract                ·  Method                       ·  Min        ·  Max        ·  Avg            ·  # calls      ·  eur (avg)  │
|  BridgePool              ·  disputeRelay                 ·     196676  ·     527538  ·         450217  ·           12  ·          -  │

New:
|  BridgePool              ·  disputeRelay                 ·     207453  ·     501794  ·         427936  ·           12  ·          -  │

```

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -612,9 +691,19 @@ contract SkinnyOptimisticOracle is SkinnyOptimisticOracleInterface, Testable, Lo
         return request.bond.div(2);
     }
 
-    function _validateLiveness(uint256 liveness) private pure {
```
nicholaspai, line 615: This method is only used in constructor and we can't use it in other methods that take in `customLiveness` because we allow `customLiveness` to be equal to 0.

mrice32:
@nicholaspai I'm assuming we're not going to do this. Should we close this PR?

mrice32:
Closing because I don't think we're going to do this. @nicholaspai feel free to reopen if you disagree!

PR #3490, opened 2021-10-24
feat: decentralized proposal system

mrice32:
**Motivation**

Decentralize the proposal system.

**Summary**

This adds a new proposer contract that should allow anyone to propose governance actions using a bond. They get the bond back if the proposal is accepted.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/core/contracts/oracle/implementation/AdminIdentifierLib.sol
```diff
@@ -0,0 +1,64 @@
...
+    // The identifier is of the form "Admin n", where n is the proposal id provided.
+    function _constructIdentifier(uint256 id) internal pure returns (bytes32) {
+        bytes32 bytesId = _uintToUtf8(id);
+        return _addPrefix(bytesId, "Admin ", 6);
+    }
+
+    // This method converts the integer `v` into a base-10, UTF-8 representation stored in a `bytes32` type.
+    // If the input cannot be represented by 32 base-10 digits, it returns only the highest 32 digits.
+    // This method is based off of this code: https://ethereum.stackexchange.com/a/6613/47801.
+    function _uintToUtf8(uint256 v) internal pure returns (bytes32) {
```
nicholaspai, line 15: Would this method be helpful to add to `AncillaryDataLib`?

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+        emit BondSet(_bond);
+    }
+
+    /**
+     * @notice Propose a new set of governance transactions for vote.
+     * @dev Pulls bond from the caller.
+     * @param transactions list of transactions for the governor to execute.
+     * @return id the id of the governor proposal.
+     */
+    function propose(Governor.Transaction[] memory transactions) public returns (uint256 id) {
```
nicholaspai, line 57: Should we reentrancy-guard these public methods?

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+        bondedProposals[id] = BondedProposal({ sender: msg.sender, lockedBond: bond, time: uint64(getCurrentTime()) });
+        governor.propose(transactions);
+    }
+
+    /**
+     * @notice Resolves a proposal by checking the status of the request in the Voting contract.
+     * @dev Pulls bond from the caller.
+     * @param id proposal id.
+     */
+    function resolveProposal(uint256 id) external payable {
```
nicholaspai, line 69: Why make this `payable` if its only transferring `token.safeTransfer`?

chrismaree: governor can transfer eth.

packages/core/contracts/oracle/implementation/AdminIdentifierLib.sol
chrismaree: well that would mean that the `Governor.sol` also needs to import `AncillaryDataLib` and these particular methods are only used in the `Governor`.

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+        uint256 _bond,
+        Governor _governor,
+        Finder _finder,
+        address _timer
+    ) Testable(_timer) {
+        token = _token;
+        bond = _bond;
+        governor = _governor;
+        finder = _finder;
+        emit BondSet(_bond);
```
chrismaree, line 48: nit: use `setBond` so you dont need to emit yourself here.

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+import "./Voting.sol";
+import "./AdminIdentifierLib.sol";
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title Proposer contract that allows anyone to make governance proposals with a bond.
+ */
+contract Proposer is Ownable, Testable {
```
chrismaree, line 16: the owner of this is intended to be the DVM, yes? it's up to the deployer to transfer this ownership? would it not be better to do the transfer in the constructor?

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+        uint64 time;
+        uint256 lockedBond;
+    }
+    mapping(uint256 => BondedProposal) public bondedProposals;
+
+    event BondSet(uint256 bond);
+    event ProposalResolved(uint256 indexed id, bool success);
+
+    /**
+     * @notice Construct the Proposer contract.
```
chrismaree, line 35: natspec

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+     * @notice Propose a new set of governance transactions for vote.
+     * @dev Pulls bond from the caller.
+     * @param transactions list of transactions for the governor to execute.
+     * @return id the id of the governor proposal.
+     */
+    function propose(Governor.Transaction[] memory transactions) public returns (uint256 id) {
+        id = governor.numProposals();
+        token.safeTransferFrom(msg.sender, address(this), bond);
+        bondedProposals[id] = BondedProposal({ sender: msg.sender, lockedBond: bond, time: uint64(getCurrentTime()) });
+        governor.propose(transactions);
```
chrismaree, line 61: this is super clean

packages/core/contracts/oracle/implementation/Proposer.sol
```diff
@@ -0,0 +1,94 @@
...
+    /**
+     * @notice Resolves a proposal by checking the status of the request in the Voting contract.
+     * @dev Pulls bond from the caller.
+     * @param id proposal id.
+     */
+    function resolveProposal(uint256 id) external payable {
+        BondedProposal storage bondedProposal = bondedProposals[id];
+        Voting voting = Voting(finder.getImplementationAddress(OracleInterfaces.Oracle));
+        require(
+            voting.hasPrice(AdminIdentifierLib._constructIdentifier(id), bondedProposal.time, ""),
```
chrismaree, line 73: The only downside to this is that this contract needs to be a registered contract (can request prices). I don't think this is too big a deal.

packages/core/contracts/oracle/implementation/AdminIdentifierLib.sol
mrice32: Yeah, I was trying to keep the changes as minimal as possible. We could move them into ancillary data lib, but they are mostly unrelated to ancillary data since they are used to construct the identifier. Note: there is some functionality in the ancillary data lib that does a similar thing (more efficiently), but I wanted to be super confident that this works _exactly_ the same way the governor does.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Actually, I don't think it needs to be payable. This is a relic from the initial implementation.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Yeah :/. I couldn't find a way around that

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Probably!

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: I remember initially trying this and then deciding against it for some reason. I can't remember why now, so i'll give it another go :).

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Hmmmm, yeah, I think this makes sense. The only downside I can imagine is that I could imagine us wanting more flexibility in certain deployment scenarios. I can't think of any examples, though, so I'll go ahead and make this change.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Done.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Removed payable.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Seems to work. Fixed.

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Done. It made the tests a little trickier to implement since you have to go through the governor to test the bond changing functionality, but it seems to work fine!

packages/core/contracts/oracle/implementation/Proposer.sol
mrice32: Done.

mrice32:
@chrismaree @nicholaspai thanks for the comments. I've responded and/or fixed everything requested and added tests. PTAL.

packages/core/contracts/oracle/implementation/AdminIdentifierLib.sol
nicholaspai: Agreed with both of you, nothing to do!

packages/core/contracts/oracle/implementation/Proposer.sol
nicholaspai: Hmm I would have leaned to letting deployer transfer ownership, not doing it automatically in constructor. Up to you @mrice32

packages/core/contracts/oracle/test/MockOracleCombined.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "./MockOracleAncillary.sol";
+
+// A mock oracle used for testing. Allows both ancillary and non-ancillary methods to be called.
+contract MockOracleCombined is MockOracleAncillary {
+    constructor(address _finderAddress, address _timerAddress) MockOracleAncillary(_finderAddress, _timerAddress) {}
```
nicholaspai, line 8: +1000

nicholaspai (approved):
Awesome work and thanks for the tests, LGTM

packages/core/test/oracle/Proposer.js
```diff
@@ -0,0 +1,182 @@
...
+
+  let mockOracle;
+  let votingToken;
+  let finder;
+  let governor;
+
+  before(async function () {
+    accounts = await web3.eth.getAccounts();
+    [owner, submitter, rando] = accounts;
+    await runDefaultFixture(hre);
```
chrismaree, line 34: nit: where possible I think we should avoid these defaultFixtures and deploy all the contracts required by the tests _in_ the tests. we are trying to do this more and more with the new code.

chrismaree (approved):
This is excellent. Really excited for what this brings!

packages/core/test/oracle/Proposer.js
mrice32: Why do you think we should avoid these?
I can give two reasons why I think it's a _good_ starting point:
- Less test code.
- More accurately reflects the deployed system in production. Put differently, you have to deliberately change things where you want the test to differ from our standard production deployment.

Thoughts?

chrismaree: main downside from my experience is that it is slower to use this. This should not be the case with how hardhat snapshotting is meant to work but in my experience it is slower.

either way, I think in time we'd like to use _more_ of these kinds of fixtures. would be great to come up with a set that enable us to delete big chunks from before each blocks.

mrice32: I do try to avoid some of the costs by ensuring this only runs in the before and not the beforeEach, but I hear you. This does interact with a fair number of UMA contracts (store, governor, etc), so I think it makes sense to leave this using this same fixture, especially since the same fixture is used throughout the oracle suite.

It does illustrate that we need a more concrete plan for how to work with fixtures and test environments generally.

On a positive note: our tests are pretty quick nowadays!

PR #3516, opened 2021-10-29
improve(insured-bridge): BridgePool requires relay.quoteTime >= contract deployment time

nicholaspai:
**Motivation**

Making this contract change would allow us to remove a block of code from the `insured-bridge-relayer`'s dispute logic. Currently the disputer must check if a relay's `quoteTime < BridgePool.deploymentTime` and if so, then it should always dispute the relay because it will not be able to compute a `realizedLpFeePct`.

Instead, if we block relays where `relay.depositData.quoteTime < BridgePool.deploymentTime` then we don't have to deal with this case.

The downside is (1) marginal gas cost increase for relay methods and (2) when a `BridgePool` is upgraded, existing un-relayed deposits must be relayed on the old deprecated `BridgePool` instead of the new one.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


chrismaree (approved):
very good

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -385,6 +389,10 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
      *      quoteTimestamp. The OO acts to verify the correctness of this realized fee. Can not exceed 50%.
      */
     function relayDeposit(DepositData memory depositData, uint64 realizedLpFeePct) public nonReentrant() {
+        // If deposit quote time is before this contract's deployment, then clients will not be able to compute
+        // a realized LP fee %, so we should block such relays.
+        require(depositData.quoteTimestamp >= deploymentTimestamp, "Deposit before deployment");
```
chrismaree, line 394: we could make this a modifier? guess it's less gas to have it this way though.

nicholaspai: Yea let's leave it like this for gas reasons. There are other require statements we could make modifiers but we chose to keep it inline for this same reason

PR #3518, opened 2021-10-29
improve(insured-bridge-relayer): Prevent unexpected errors from causing bot loop to end prematurely

nicholaspai:
**Motivation**

Whenever the `Relayer.ts` attempts to submit an on-chain transaction, it correctly catches any on-chain errors and prevents the outer `index.ts` loop from failing early. However, there are many places in `Relayer.ts` where a failure can occur and we don't want these to similarly cause the outer `index.ts` loop to fail. Therefore this PR makes sure to catch all errors that can occur.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -62,7 +62,80 @@ export class Relayer {
...
+            message: "Pending relay is invalid",
+            pendingRelay,
+            relayableDeposit,
+            reason: relayDisputable.reason,
+          });
+          return;
+        }
+
+        // If we reach here, then pending relay has not expired and its valid, so there is a chance we can speed
+        // it up.
```
chrismaree, line 104: try get this onto one line?

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -90,15 +90,34 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
           // Update state.
           await Promise.all([gasEstimator.update(), l1Client.update(), l2Client.update()]);
 
-          // Start bots that are enabled.
-          if (config.botModes.relayerEnabled) await relayer.checkForPendingDepositsAndRelay();
+          // Gather promises from all bots that are enabled and Promise.allSettled() them all in parallel. This means
```
chrismaree, line 93:
```suggestion
          // Gather promises from all modules that are enabled and Promise.allSettled() them all in parallel. This means
```

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -62,6 +62,77 @@ export class Relayer {
     this.l2BlockFinder = new BlockFinder<BlockTransactionBase>(this.l2Client.l2Web3.eth.getBlock);
   }
 
+  // Evaluates given `relayableDeposit` for the `l1Token` and determines whether to submit a slow or fast relay.
+  private async _relayPendingDeposit(l1Token: string, relayableDeposit: RelayableDeposit): Promise<void> {
```
chrismaree, line 66: this is pretty clean! I really like the logic flow this enables in `checkForPendingDepositsAndRelay`.

chrismaree: one small nit: can you put this method down below `checkForPendingDepositsAndRelay`? ideally all "public" functions should come before the "private" ones.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -96,228 +167,179 @@ export class Relayer {
...
+          depositHash: relay.depositHash,
+          l1Recipient: relay.l1Recipient,
+          l2Sender: relay.l2Sender,
+          l1Token: relay.l1Token,
+          amount: relay.amount,
+          slowRelayFeePct: relay.slowRelayFeePct,
+          instantRelayFeePct: relay.instantRelayFeePct,
+          quoteTimestamp: relay.quoteTimestamp,
+          depositContract: (await this.l1Client.bridgeAdmin.methods.depositContracts(relay.chainId).call())[0],
+          // `depositContracts()` returns [depositContract, messengerContract] and we want the first arg.
```
chrismaree, line 220: can we remove this?

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: Agreed! easier to reason about and maintain to with this refactor

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: hmmm I think this comment is worth keeping

packages/insured-bridge-relayer/test/Relayer.ts
```diff
@@ -783,6 +836,68 @@ describe("Relayer.ts", function () {
       await relayer.checkForPendingRelaysAndDispute();
       assert.isTrue(lastSpyLogIncludes(spy, "Disputed pending relay"));
     });
+    it("Two pending relays with invalid relay data, one dispute succeeds, one fails", async function () {
```
chrismaree, line 839: these tests are great. as everything else passes without need to be changed it implies that we have not broken anything else.

chrismaree (approved):
This is great! good work on changing up this ordering.

Only outstanding issue is just to place the public functions at the top of the `Relayer` and leave the private methods at the bottom.

PR #3520, opened 2021-10-29
improve(common): Speed up default behavior for findBlockNumberAtTImestamp

nicholaspai:
**Motivation**

This method is the compute-time bottle neck in the `Relayer`.

**Summary**

One way to speed up this method is to increase the number of blocks that we travel per search. Currently it binary searches up and down incrementing/decrementing by only 1 block. Empirically increasing this value speeds up time significantly.

Additionally, increasing averageBlockTime seems to also speed up time significantly. I believe this is the case because making more queries in the initial loop to find a `block.timestamp < targetTimestamp` is much less costly than overshooting on the first query and needing to crawl back.

For example, using an `averageBlockTime = 14` finds a `block.timestamp < targetTimestamp` in 4 tries and completes the function in 2.5 seconds. On the other hand, using `averageBlockTime = 13` finds a `block.timestamp < targetTimestamp` in one try but then takes 100 seconds to complete the function. This example uses a `blockDelta` step of 4.

From my experimenting, I can locate a block for a timestamp 1 week ago in 2 seconds using a `blockDelta=5` and `averageBlockTime=14`. This same call in current state takes > 5 minutes.

**Details**

I think simultaneously increasing the `blockDelta` and increasing `averageBlockTime` have offsetting effects and generally reduce in pretty low error. The former increases error, but the latter decreases error. And they both speed up the function!

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/common/src/TransactionUtils.ts
+++ b/packages/common/src/TransactionUtils.ts
@@ -211,7 +211,6 @@ export async function findBlockNumberAtTimestamp(
   blockDelta = 4,
   averageBlockTime = 14
 ): Promise<{ blockNumber: number; error: number }> {
-  const start = Date.now();
   const higherLimitStamp = targetTimestamp + higherLimitMax;
   const lowerLimitStamp = targetTimestamp - lowerLimitMax;
 
@@ -262,6 +261,5 @@ export async function findBlockNumberAtTimestamp(
       }
     }
   }
-  console.log(`Elapsed time: ${Date.now() - start} ms`);
   return { blockNumber: block.number, error: Math.abs(targetTimestamp - parseInt(block.timestamp.toString())) };
 }
--- a/packages/common/src/TransactionUtils.ts
+++ b/packages/common/src/TransactionUtils.ts
@@ -200,7 +200,7 @@ export const blockUntilBlockMined = async (web3: Web3, blockerBlockNumber: numbe
  * @param blockDelta Amount of blocks to hop when binary searching for block. Increasing this increases error
  * but significantly reduces time to compute.
  * @param averageBlockTime Decreasing average block size will decrease precision and also decrease the amount of
- * requests made in order to find the closest block.
+ * requests made in order to find the closest block. Increasing this reduces time to compute but increases requests.
  * @returns {number, number} Block height and difference between block timestamp and target timestamp
  */
 export async function findBlockNumberAtTimestamp(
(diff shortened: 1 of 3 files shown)


packages/common/src/TransactionUtils.ts
```diff
@@ -188,23 +188,40 @@ export const blockUntilBlockMined = async (web3: Web3, blockerBlockNumber: numbe
...
+ * requests made in order to find the closest block. Increasing this reduces time to compute but increases requests.
+ * @returns {number, number} Block height and difference between block timestamp and target timestamp
+ */
 export async function findBlockNumberAtTimestamp(
   web3: Web3,
   targetTimestamp: number,
   higherLimitMax = 15,
-  lowerLimitMax = 15
+  lowerLimitMax = 15,
+  blockDelta = 4,
```
daywiss, line 211: Why not default this to 1 to keep previous behavior the same

packages/common/src/TransactionUtils.ts
```diff
@@ -188,23 +188,40 @@ export const blockUntilBlockMined = async (web3: Web3, blockerBlockNumber: numbe
...
+  averageBlockTime = 14
```
daywiss, line 212: Same with this 13

daywiss (approved):

daywiss:
Would be a bit of work, but i wonder if you can get an approximate block by calculting a percent, given the timestamp of first known block, the time of the last block, and the total number of blocks. This might a starting point closer to the block you are looking for.

packages/common/src/TransactionUtils.ts
nicholaspai: Good point will change this and the 13

nicholaspai:
> Would be a bit of work, but i wonder if you can get an approximate block by calculting a percent, given the timestamp of first known block, the time of the last block, and the total number of blocks. This might a starting point closer to the block you are looking for.

Can you explain this a bit more?

daywiss:
> Can you explain this a bit more?

Yea what i guess im saying, if its beneficial to have a closer block to the solution when you start your search, then maybe you can do an estimation to get it.  Take the timestamp to find / (current block timestamp - block(0).timestamp). This should give you a percentage.   then take percent * latestblock.number to get you an estimated starting block to start your search.  Theoretically it should be closer than starting your search at block 0 or latest block.

PR #3521, opened 2021-10-30
improve(common): Improve precision of block finder common util

nicholaspai:
**Motivation**

Adds more precision to #3520 to ensure that `InsuredBridgeL1Client` finds the exact block containing the `deposit.quoteTime`

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/affiliates/gas-rebate/VoterGasRebate.js
+++ b/packages/affiliates/gas-rebate/VoterGasRebate.js
@@ -634,9 +634,9 @@ async function Main(callback) {
       console.log(`- Using start date: ${moment.unix(startDate).toString()}`);
       console.log(`- Using end date: ${moment.unix(endDate).toString()}`);
 
-      endBlock = (await FindBlockAtTimestamp._findBlockNumberAtTimestamp(web3, Number(endDate), 15, 15, 1, 14))
+      endBlock = (await FindBlockAtTimestamp._findBlockNumberAtTimestamp(web3, Number(endDate), 1, 1, 1, 14))
         .blockNumber;
-      startBlock = (await FindBlockAtTimestamp._findBlockNumberAtTimestamp(web3, Number(startDate), 15, 15, 1, 14))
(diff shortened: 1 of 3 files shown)


packages/common/src/TransactionUtils.ts
```diff
@@ -242,15 +242,15 @@ export async function findBlockNumberAtTimestamp(
   if (higherLimitStamp) {
     if (block.timestamp >= higherLimitStamp) {
       while (block.timestamp >= higherLimitStamp) {
-        blockNumber -= blockDelta;
+        blockNumber -= 1;
```
nicholaspai, line 245: These last two binary searches should always use 1 as the block delta since we are now trying to pinpoint the exact block.

mrice32: I think I agree with this.

packages/affiliates/gas-rebate/VoterGasRebate.js
```diff
@@ -634,9 +634,9 @@ async function Main(callback) {
       console.log(`- Using start date: ${moment.unix(startDate).toString()}`);
       console.log(`- Using end date: ${moment.unix(endDate).toString()}`);
 
-      endBlock = (await FindBlockAtTimestamp._findBlockNumberAtTimestamp(web3, Number(endDate), 15, 15, 4, 14))
+      endBlock = (await FindBlockAtTimestamp._findBlockNumberAtTimestamp(web3, Number(endDate), 1, 1, 1, 14))
```
mrice32, line 637: Wasn't the default for the higher limit 15 before? Either way, I think this is more precise, but will probably kill our compute time. But this is good for now as it gives us the accuracy back.

mrice32 (approved):
LGTM!

packages/affiliates/gas-rebate/VoterGasRebate.js
nicholaspai: empirically the compute time is fine setting it as is!

PR #3522, opened 2021-10-31
feat: use faster block finder

mrice32:
**Motivation**

We'd like to speed up our block searching by timestamp in the relay bots.

**Summary**

This PR changes two things:
- Swaps out the finder algorithm in the relay client for one that uses binary search rather than a linear search as the current one does.
- Makes the BlockFinder a true class rather than a function that returns a function. This makes it easier to work with in typescript. There was a little bit of cleanup in the price feeds related to this change. Overall, this simplified the integration in the price feeds and the relayer bot.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -1,7 +1,7 @@
 import Web3 from "web3";
 const { toBN, soliditySha3 } = Web3.utils;
 
-import { findBlockNumberAtTimestamp } from "@uma/common";
+import { BlockFinder } from "../price-feed/utils";
 import { getAbi } from "@uma/contracts-node";
 import { Deposit } from "./InsuredBridgeL2Client";
 import { RateModel, calculateRealizedLpFeePct } from "../helpers/acrossFeesCalculator";
@@ -9,6 +9,7 @@ import { RateModel, calculateRealizedLpFeePct } from "../helpers/acrossFeesCalcu
 import type { BridgeAdminInterfaceWeb3, BridgePoolWeb3 } from "@uma/contracts-node";
 import type { Logger } from "winston";
 import type { BN } from "@uma/common";
+import type { BlockTransactionBase } from "web3-eth";
 
 export enum ClientRelayState {
   Uninitialized, // Deposit on L2, nothing yet on L1. Can be slow relayed and can be sped up to instantly relay.
@@ -64,6 +65,7 @@ export class InsuredBridgeL1Client {
 
   private firstBlockToSearch: number;
   private web3: Web3;
+  private readonly blockFinder: BlockFinder<BlockTransactionBase>;
 
   constructor(
     private readonly logger: Logger,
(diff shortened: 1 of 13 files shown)


packages/financial-templates-lib/src/price-feed/utils.ts
```diff
@@ -199,41 +199,70 @@ export const PriceHistory = <T>(
...
- */
-export const BlockFinder = <T extends { number: number; timestamp: number | string }>(
-  requestBlock: (requestedBlock: string | number) => Promise<T>,
-  blocks: T[] = []
-): {
-  getBlockForTimestamp: (timestamp: number) => Promise<T>;
-} => {
-  assert(requestBlock, "requestBlock function must be provided");
+export class BlockFinder<T extends { number: number; timestamp: number | string }> {
+  constructor(
```
mrice32, line 203: Note: there were very few real changes here. This code was just moved around and put into a class. No functional changes were made.

nicholaspai (commented):
I'm not seeing how this algorithm is binary search can you clarify?

Also, have you tried running this vs the existing algo on say time stamps 2 hours back and compared time?

mrice32:
> I'm not seeing how this algorithm is binary search can you clarify?
> 
> Also, have you tried running this vs the existing algo on say time stamps 2 hours back and compared time?

I have not. I've run it against a timestamp a week ago and saw it return in < 2 seconds. I can give individual benchmarks as well. Would 5 min, 2 hours, and 1 week work?

nicholaspai (commented):
I'm good with switching to this algo because binary search seems more scalable than linear search as long as the compute time and error is similar or better to the current algo.

Can you share some time + error results when searching for timestamps 2 hours, 1 day, and 1 week ago? I picked those times arbitrarily

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -149,8 +152,7 @@ export class InsuredBridgeL1Client {
 
     // The block number must be exactly the one containing the deposit.quoteTimestamp, so we use the lowest block delta
     // of 1. Setting averageBlockTime to 14 increases the speed at the cost of more web3 requests.
-    const quoteBlockNumber = (await findBlockNumberAtTimestamp(this.l1Web3, deposit.quoteTimestamp, 15, 15, 1, 14))
-      .blockNumber;
+    const quoteBlockNumber = (await this.blockFinder.getBlockForTimestamp(deposit.quoteTimestamp)).number;
```
chrismaree, line 155: 👌

packages/financial-templates-lib/src/price-feed/utils.ts
```diff
@@ -199,41 +199,70 @@ export const PriceHistory = <T>(
...
+    private readonly blocks: T[] = []
+  ) {
+    assert(requestBlock, "requestBlock function must be provided");
+  }
+
+  /**
+   * @notice Gets the latest block whose timestamp is <= the provided timestamp.
+   * @param {number} timestamp timestamp to search.
+   */
+  public async getBlockForTimestamp(timestamp: number | string): Promise<T> {
```
chrismaree, line 214: this is good and will work well for blocks close to the current timestamp, which is the situation we are in for the relayer.

chrismaree (approved):

nicholaspai (approved):

PR #3527, opened 2021-11-01
fix(lsp): Fix incorrect inequality for expiration timestamp in settle method

chrismaree:
**Motivation**

OZ identified the following issue:

_The settle function of the LongShortPair contract considers settlement conditions when the current time is strictly before or after the expiration timestamp. However, it incorrectly reverts when the current time matches the expiration timestamp._

This PR fixes this by making a small modification to the settle method, as recommended by OZ.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


nicholaspai (approved):
Nice fix lgtm

PR #3532, opened 2021-11-01
fix(ancillary-data-lib): Fix faulty append function

nicholaspai:
**Motivation**

The `appendKeyValueBytes32` function should combine its inputs into a formatted bytes array. However, the `currentAncillaryData` is incorrectly discarded.

Since the ancillary data influences the oracle resolution process, an incorrect value could undermine the oracle results. Fortunately, there is only one call to `appendKeyValueBytes32` in the codebase, and it uses an empty `currentAncillaryData` buffer, so the bug does not affect this case.

This PR updates the `appendKeyValueBytes32` function so that the `currentAncillaryData` is included in the returned bytes array.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


chrismaree (approved):
great catch by OZ! we should have caught this in unit tests.

PR #3534, opened 2021-11-01
fix(skinny-optimistic-oracle): Default proposer bond to final fee in requestAndProposePriceFor

nicholaspai:
…

**Motivation**

We default the proposer bond amount to the final fee in the case where the caller sets it to 0. Original discussion for this decision in the `requestPrice` method can be found [here](https://github.com/UMAprotocol/protocol/pull/3409#discussion_r719747949).

We forgot to do the same default logic in `requestAndProposePriceFor`.

From the audit:

> The `requestPrice` function of the `SkinnyOptimisticOracle` contract uses the final fee as the bond if the bond is not specified. However, the `requestAndProposePriceFor` function may use a zero bond, which contradicts its `@notice` and `@param` comments. A zero bond weakens the incentive against invalid proposal or disputes.
Fortunately, the only call to this function in the code base sets a proposer bond. Nevertheless, consider using the final fee if the bond is not specified.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
@@ -318,7 +318,7 @@ contract SkinnyOptimisticOracle is SkinnyOptimisticOracleInterface, Testable, Lo
         request.currency = currency;
         request.reward = reward;
         request.finalFee = finalFee;
-        request.bond = bond;
+        request.bond = bond != 0 ? bond : finalFee;
         request.customLiveness = customLiveness;
         request.proposer = proposer;
         request.proposedPrice = proposedPrice;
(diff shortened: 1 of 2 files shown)


packages/core/contracts/oracle/implementation/SkinnyOptimisticOracle.sol
```diff
@@ -318,7 +318,7 @@ contract SkinnyOptimisticOracle is SkinnyOptimisticOracleInterface, Testable, Lo
         request.currency = currency;
         request.reward = reward;
         request.finalFee = finalFee;
-        request.bond = bond;
+        request.bond = bond != 0 ? bond : finalFee;
```
chrismaree, line 321: 👌

chrismaree (approved):

PR #3536, opened 2021-11-01
improve(insured-bridge): Add require messages

nicholaspai:
**Motivation**

There are require statements in the `BridgePool `contract without error messages.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -269,7 +269,8 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
         require(
             depositData.slowRelayFeePct <= 0.25e18 &&
                 depositData.instantRelayFeePct <= 0.25e18 &&
-                realizedLpFeePct < 0.5e18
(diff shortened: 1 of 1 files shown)


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -269,7 +269,8 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
         require(
             depositData.slowRelayFeePct <= 0.25e18 &&
                 depositData.instantRelayFeePct <= 0.25e18 &&
-                realizedLpFeePct < 0.5e18
+                realizedLpFeePct < 0.5e18,
```
chrismaree, line 272: side comment: should `realizedLpFeePct` not be `<=`?

chrismaree (approved):

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Hmmmmm I think it should I'll change

PR #3539, opened 2021-11-01
fix(skinny-optimistic-oracle): Reentrancy guard requestAndProposePriceFor

nicholaspai:
**Motivation**

The `requestAndProposePriceFor` function of the `SkinnyOptimisticOracle` contract makes a call to an untrusted `msg.sender` but is not guarded by a `nonReentrant` modifier. While, in this instance, this does not seem to be a security concern, this can introduce unexpected behavior.

Consider adding the `nonReentrant` modifier to all functions which make calls to possibly untrusted contracts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


chrismaree (approved):

chrismaree:
This is the only function that makes external calls _and_ does not have the modifier, right?

PR #3541, opened 2021-11-01
improve(arbitrum-messenger): Emit event including return value of Arbitrum sys contract call

nicholaspai:
**Motivation**

The `relayMessage` function of the `Arbitrum_Messenger` contract does not emit a relevant event after executing a sensitive action. The `relayMessage` function calls as a subroutine `sentTxToL2NoAliasing` which itself returns the `uint256` value `seqNum`, but this return value is not logged in the `relayMessage` function.
Consider emitting events after sensitive changes take place, to facilitate tracking and notify off-chain clients following the contract's activity.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/avm/Arbitrum_Messenger.sol
+++ b/packages/core/contracts/insured-bridge/avm/Arbitrum_Messenger.sol
@@ -56,15 +56,7 @@ contract Arbitrum_Messenger is Ownable, Arbitrum_CrossDomainEnabled, MessengerIn
     ) external payable override onlyOwner {
         // Since we know the L2 target's address in advance, we don't need to alias an L1 address.
         uint256 seqNumber =
-            sendTxToL2NoAliassing(
-                target,
-                userToRefund,
-                l1CallValue,
-                maxSubmissionCost, // TODO: Determine the max submission cost. From the docs: "current base submission fee is queryable via ArbRetryableTx.getSubmissionPrice"
-                gasLimit,
-                gasPrice,
-                message
-            );
+            sendTxToL2NoAliassing(target, userToRefund, l1CallValue, maxSubmissionCost, gasLimit, gasPrice, message);
         emit RelayedMessage(
             msg.sender,
             target,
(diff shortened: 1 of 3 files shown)


packages/core/contracts/insured-bridge/avm/Arbitrum_Messenger.sol
```diff
@@ -43,13 +55,25 @@ contract Arbitrum_Messenger is Ownable, Arbitrum_CrossDomainEnabled, MessengerIn
         bytes memory message
     ) external payable override onlyOwner {
         // Since we know the L2 target's address in advance, we don't need to alias an L1 address.
-        sendTxToL2NoAliassing(
+        uint256 seqNumber =
+            sendTxToL2NoAliassing(
+                target,
+                userToRefund,
+                l1CallValue,
+                maxSubmissionCost, // TODO: Determine the max submission cost. From the docs: "current base submission fee is queryable via ArbRetryableTx.getSubmissionPrice"
```
chrismaree, line 63: shall we remove this todo?

packages/core/contracts/insured-bridge/avm/Arbitrum_Messenger.sol
```diff
@@ -13,6 +13,18 @@ import "@openzeppelin/contracts/access/Ownable.sol";
  * owner for L2 contracts that expect to receive cross domain messages.
  */
 contract Arbitrum_Messenger is Ownable, Arbitrum_CrossDomainEnabled, MessengerInterface {
+    event RelayedMessage(
```
chrismaree, line 16: this is great!

chrismaree (approved):
is there any similar info we'd like to add to the optimism messenger?

packages/core/contracts/insured-bridge/avm/Arbitrum_Messenger.sol
nicholaspai: Done

nicholaspai:
> is there any similar info we'd like to add to the optimism messenger?

Eh the Optimism sys contracts don't return any value, so we'd only be emitting the function params. I think we don't make this change until we test it more in the wild.

PR #3544, opened 2021-11-02
fix(across-bots): Address issue where relayer produces an error after every relay

chrismaree:
**Motivation**
Every time the relayer runs and sends relays it generates an error right after the fact. This is irritating. This PR fixes this behaviour.
**Summary**

The issue boiled down to double counting deposits due to incorrect filtering in the `Relayer.ts`. See code for more details on the bug and the solution.

Note that this issue was not shown in unit tests as it is an artefact of having multiple pools (more than 1 token) on one relayer.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -64,7 +64,6 @@ export class InsuredBridgeL1Client {
   private instantRelays: { [key: string]: { [key: string]: InstantRelay } } = {}; // L1TokenAddress=>{depositHash, realizedLpFeePct}=>InstantRelay.
 
   private firstBlockToSearch: number;
-  private web3: Web3;
   private readonly blockFinder: BlockFinder<BlockTransactionBase>;
 
   constructor(
@@ -83,8 +82,7 @@ export class InsuredBridgeL1Client {
     this.bridgePools = {}; // Initialize the bridgePools with no pools yet. Will be populated in the _initialSetup.
 
     this.firstBlockToSearch = startingBlockNumber;
-    this.web3 = l1Web3;
-    this.blockFinder = new BlockFinder<BlockTransactionBase>(this.web3.eth.getBlock);
+    this.blockFinder = new BlockFinder<BlockTransactionBase>(this.l1Web3.eth.getBlock);
   }
 
   // Return an array of all bridgePool addresses
@@ -188,6 +186,11 @@ export class InsuredBridgeL1Client {
 
   async getProposerBondPct(): Promise<BN> {
     return toBN(await this.bridgeAdmin.methods.proposerBondPct().call());
+  }
+
+  async getL2DepositBoxAddress(chainId: number): Promise<string> {
+    const depositContracts = (await this.bridgeAdmin.methods.depositContracts(chainId).call()) as any;
+    return depositContracts.depositContract || depositContracts[0];
   }
 
   async update(): Promise<void> {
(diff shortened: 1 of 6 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -64,7 +64,6 @@ export class InsuredBridgeL1Client {
   private instantRelays: { [key: string]: { [key: string]: InstantRelay } } = {}; // L1TokenAddress=>{depositHash, realizedLpFeePct}=>InstantRelay.
 
   private firstBlockToSearch: number;
-  private web3: Web3;
```
chrismaree, line 67: I made a small clean up here. For some reason we diverged on how we declared web3 instances. within the clients we should _not_ refer to this as `web3` but rather define the network (l1/l2) to make it very clear in the code what we are talking to/about.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -190,6 +188,11 @@ export class InsuredBridgeL1Client {
     return toBN(await this.bridgeAdmin.methods.proposerBondPct().call());
   }
 
+  async getL2DepositBoxAddress(chainId: number): Promise<string> {
```
chrismaree, line 191: this method was in the l2 client. however, it should be here in the l1 client as it is done against l1 contracts. more over, we already have the l1 web3 stance and the BridgeAdmin so no passing in required. I made associated changes to the `index` for the relayer and the `CreatePriceFeed` logic that called these methods.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -660,7 +660,7 @@ export class Relayer {
     const relayableDeposits: RelayableDeposits = {};
     for (const l1Token of this.whitelistedRelayL1Tokens) {
       this.logger.debug({ at: "Relayer", message: "Checking relays for token", l1Token });
-      const l2Deposits = this.l2Client.getAllDeposits();
+      const l2Deposits = this.l2Client.getAllDepositsForL1Token(l1Token);
```
chrismaree, line 663: this bit of code encapsulates the bug that the bot had: we are iterating over all `this.whitelistedRelayL1Tokens` and for _each_ pulling out _all_ deposits. Rather, we should only be looking at the deposits for the given `l1Token` to prevent duplication.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -47,6 +47,10 @@ export class InsuredBridgeL2Client {
     return Object.keys(this.deposits).map((depositHash: string) => this.deposits[depositHash]);
   }
 
+  getAllDepositsForL1Token(l1TokenAddress: string) {
```
chrismaree, line 50: this method enables the fix for the bug. see https://github.com/UMAprotocol/protocol/pull/3544/files#r740999431 to see how it's used.

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -65,9 +65,9 @@ export class RelayerConfig {
       RATE_MODELS,
       CHAIN_IDS,
       L2_BLOCK_LOOKBACK,
-      DISPUTER_ENABLED,
```
chrismaree, line 68: I moved this ordering around to match the order in which they are used.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -190,6 +188,12 @@ export class InsuredBridgeL1Client {
     return toBN(await this.bridgeAdmin.methods.proposerBondPct().call());
   }
 
+  // Returns the L2 Deposit box address for a given bridgeAdmin on L1.
+  async getL2DepositBoxAddress(chainId: number): Promise<string> {
+    const depositContracts = (await this.bridgeAdmin.methods.depositContracts(chainId).call()) as any;
+    return depositContracts.depositContract || depositContracts[0];
```
nicholaspai, line 194:
```suggestion
    return depositContracts.depositContract || depositContracts[0]; // When latest BridgeAdmin is redeployed, can remove the "|| depositContracts[0]".
```

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: ahhhhhhh we're so dumb! nice catch

nicholaspai: This also makes sense why duplicate txns were sending only on relays but not on settlements or disputes

nicholaspai (approved):
Great catch LGTM

PR #3550, opened 2021-11-02
improve(insured-bridge-relayer): Remove deploymentTimestamp hard-coded mapping in favor of web3.eth.getCode check

nicholaspai:
**Motivation**

@mrice32 made a great point that there's a much easier way to check if a deposit.quoteTime is before a contract's deployment time without having to store a hardcoded mapping of contract deployment timestamps. The solution is to check the result of `web3.eth.getCode(bridgePoolAddress, blockForDepositQuoteTime)` and if it returns `0x` then the contract did not exist and we can act accordingly.

This PR therefore reverts #3516 in favor of simplifying the relayer logic to use `web3.eth.getCode` instead of the hardcoded `deploymentTimestamp`.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -38,11 +38,6 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
     using FixedPoint for FixedPoint.Unsigned;
     using Address for address;
 
-    // Useful for clients to know if any quote times are for a time prior to the contract's deployment. These
-    // deposits will be impossible to compute realized LP fee % for since there will be no utilization at that quote
-    // time. Therefore, clients need to deal with this special case.
-    uint32 public deploymentTimestamp;
-
     // Token that this contract receives as LP deposits.
     IERC20 public override l1Token;
 
@@ -180,7 +175,6 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
         lastLpFeeUpdate = uint32(getCurrentTime());
         lpFeeRatePerSecond = _lpFeeRatePerSecond;
         isWethPool = _isWethPool;
-        deploymentTimestamp = lastLpFeeUpdate;
 
         syncUmaEcosystemParams(); // Fetch OptimisticOracle and Store addresses and L1Token finalFee.
         syncWithBridgeAdminParams(); // Fetch ProposerBondPct OptimisticOracleLiveness, Identifier from the BridgeAdmin.
--- a/packages/core/contracts/insured-bridge/BridgePool.sol
+++ b/packages/core/contracts/insured-bridge/BridgePool.sol
@@ -256,10 +256,6 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
      *      quoteTimestamp. The OO acts to verify the correctness of this realized fee. Cannot exceed 50%.
      */
     function relayAndSpeedUp(DepositData memory depositData, uint64 realizedLpFeePct) public nonReentrant() {
-        // If deposit quote time is before this contract's deployment, then clients will not be able to compute
-        // a realized LP fee %, so we should block such relays.
-        require(depositData.quoteTimestamp >= deploymentTimestamp, "Deposit before deployment");
-
         // If no pending relay for this deposit, then associate the caller's relay attempt with it.
         uint32 priceRequestTime = uint32(getCurrentTime());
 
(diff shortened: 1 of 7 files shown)


packages/core/test/insured-bridge/BridgePool.js
```diff
@@ -55,6 +55,7 @@ const lpFeeRatePerSecond = toWei("0.0000015");
 const defaultProposerBondPct = toWei("0.05");
 const defaultSlowRelayFeePct = toWei("0.01");
 const defaultInstantRelayFeePct = toWei("0.01");
+const defaultQuoteTimestamp = "100000"; // no validation of this happens on L1.
```
nicholaspai, line 58: Changes in this file revert #3516

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -38,11 +38,6 @@ contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
     using FixedPoint for FixedPoint.Unsigned;
     using Address for address;
 
-    // Useful for clients to know if any quote times are for a time prior to the contract's deployment. These
-    // deposits will be impossible to compute realized LP fee % for since there will be no utilization at that quote
-    // time. Therefore, clients need to deal with this special case.
-    uint32 public deploymentTimestamp;
```
nicholaspai, line 44: Changes in this file revert #3516

packages/financial-templates-lib/test/clients/InsuredBridgeL1Client.js
```diff
@@ -277,8 +277,7 @@ describe("InsuredBridgeL1Client", function () {
       amount: relayAmount,
       slowRelayFeePct: defaultSlowRelayFeePct,
       instantRelayFeePct: defaultInstantRelayFeePct,
-      // quote time must be <= deployment time of BridgePool for relayDeposit to succeed
-      quoteTimestamp: Number((await bridgePool.methods.deploymentTimestamp().call()).toString()),
+      quoteTimestamp: (await web3.eth.getBlock("latest")).timestamp, // set this to the current block timestamp.
```
nicholaspai, line 280: Changes in this file revert #3516

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -794,16 +800,15 @@ export class Relayer {
...
+    // TODO: We could improve this search by only looking +/- 10 minutes within the quote time, as this is a constraint
+    // enforced by the BridgeDepositBox. The hard part is that determining the average time per L2 block is not trivial.
     else {
-      const bridgePoolDeploymentTime = (
-        await this.l2BlockFinder.getBlockForTimestamp(this.deployTimestamps[relay.l1Token])
-      ).number;
       let blockSearchConfig = {
-        fromBlock: bridgePoolDeploymentTime,
-        toBlock: bridgePoolDeploymentTime + this.l2LookbackWindow,
+        fromBlock: 0,
```
nicholaspai, line 810: This is the one part I'm not happy about. This event querying seems expensive in the worst case. Arbitrum team did state that Infura won't have limited event block search configs soonTM so maybe this code can be deprecated shortly, but I'm not sure what the best way to optimize this search is. We could try to look +/- 10 minutes around the deposit.quoteTime but timing on L2 is a dangerous assumption to make.

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -16,22 +16,6 @@ const supportedChainIds = [
   421611, // arbitrum testnet
 ];
 
-// This struct is a hack right now but prevents an edge case bug where a deposit.quoteTime < bridgePool.deployment time.
```
nicholaspai, line 19: Changes in this file revert #3516

chrismaree (approved):

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: ye, and I feel that the call to `getCodeDeployedAtTimestamp` is also pretty slow.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -714,6 +713,11 @@ export class Relayer {
     };
   }
 
+  private async getCodeDeployedAtTimestamp(address: string, timestamp: number): Promise<string> {
```
chrismaree, line 716: I really dont like that this has to call `getBlockForTimestamp`. I feel storing a list of constants in the file is better.

Prehaps you should call this _once_ on the initialization of the pool within the L1Client?

nicholaspai: Hm yeah it's a trade off of more hard coding and required environment variables vs runtime. 

I think what I'll do is open another PR where I just make the changes to the BridgePool and then leave this open as a potential bot "optimization"

PR #3553, opened 2021-11-03
feat(insured-bridge): add back safe multicaller

chrismaree:
**Motivation**

To enable more scalable off-chain infrastructure, it would be useful to be able to send batches of transactions. We can do this easily with multicall. However, multicall runs a risk of being used along side `msg.value`. to deal with this, we simply can prevent multicall usage with any transaction with `msg.value`.

**Summary**

Adds back multicall with a block on any tx with `msg.value`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/common/implementation/MultiCaller.sol
+++ b/packages/core/contracts/common/implementation/MultiCaller.sol
@@ -1,11 +1,13 @@
 // This contract is taken from Uniswaps's multi call implementation (https://github.com/Uniswap/uniswap-v3-periphery/blob/main/contracts/base/Multicall.sol)
-// and was modified to be solidity 0.8 compatible. No other changes were made.
+// and was modified to be solidity 0.8 compatible. Additionally, the method was restricted to only work with msg.value
+// set to 0 to avoid any nasty attack vectors on function calls that use value sent with deposits.
 pragma solidity ^0.8.0;
 
 /// @title MultiCaller
 /// @notice Enables calling multiple methods in a single call to the contract
 contract MultiCaller {
     function multicall(bytes[] calldata data) external payable returns (bytes[] memory results) {
+        require(msg.value == 0, "Only multicall with 0 value");
         results = new bytes[](data.length);
         for (uint256 i = 0; i < data.length; i++) {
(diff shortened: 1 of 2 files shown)


packages/core/contracts/common/implementation/MultiCaller.sol
```diff
@@ -1,11 +1,13 @@
...
-// and was modified to be solidity 0.8 compatible. No other changes were made.
+// and was modified to be solidity 0.8 compatible. Additionally, the method was restricted to only work with msg.value
+// set to 0 to avoid any nasty attack vectors on function calls that use value sent with deposits.
 pragma solidity ^0.8.0;
 
 /// @title MultiCaller
 /// @notice Enables calling multiple methods in a single call to the contract
 contract MultiCaller {
     function multicall(bytes[] calldata data) external payable returns (bytes[] memory results) {
+        require(msg.value == 0, "Only multicall with 0 value");
```
nicholaspai, line 10: Does this break functionality of any existing contracts that inherit MutliCaller?

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -33,7 +34,7 @@ interface WETH9Like {
  * to post collateral by earning a fee per fulfilled deposit order.
  * @dev A "Deposit" is an order to send capital from L2 to L1, and a "Relay" is a fulfillment attempt of that order.
  */
-contract BridgePool is Testable, BridgePoolInterface, ERC20, Lockable {
+contract BridgePool is MultiCaller, Testable, BridgePoolInterface, ERC20, Lockable {
```
nicholaspai, line 37: +1

nicholaspai (approved):
Any gas changes?

packages/core/contracts/common/implementation/MultiCaller.sol
chrismaree: under all conditions multicall should not use any `msg.value`! this is where the dragon is!

also...we don't use this lib anywhere else in the repo :)

PR #3557, opened 2021-11-03
fix(insured-bridge-relayer): Key deployTimestamps by lowercase addresses

nicholaspai:
**Motivation**

`deposit.l1Token` has checksummed address, while `RelayerConfig.deployTimestamps` keys by lower-case.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
@@ -980,7 +980,7 @@ export const defaultConfigs: { [name: string]: { type: string; [key: string]: an
     type: "insuredbridge",
     l2BlockLookback: 99000, // Arbitrum only supports lookbacks of 100,000 blocks
     l2NetId: 42161,
-    bridgeAdminAddress: "0xde7130ceaf4345a12a2764f0dc50bc4899d7ec08",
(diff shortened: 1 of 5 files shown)


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
```diff
@@ -980,7 +980,7 @@ export const defaultConfigs: { [name: string]: { type: string; [key: string]: an
     type: "insuredbridge",
     l2BlockLookback: 99000, // Arbitrum only supports lookbacks of 100,000 blocks
     l2NetId: 42161,
-    bridgeAdminAddress: "0xde7130ceaf4345a12a2764f0dc50bc4899d7ec08",
+    bridgeAdminAddress: "0x30B44C676A05F1264d1dE9cC31dB5F2A945186b6",
```
nicholaspai, line 983: Unrelated change to this file that we should make now that we've ugpraded BridgeAdmin

chrismaree (approved):
I modified this to use `replaceAddressCase` to not place restrictions on how the user configures the bot. In all cases, internally the bot should use check sum formatted addresses.

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
chrismaree: yup, this is clutch.

PR #3558, opened 2021-11-04
feat(optimistic-oracle-proposer): Can propose, dispute, and settle SkinnyOptimisticOracle price requests

nicholaspai:
**Motivation**

User can configure the `index.js` file to work with a "SkinnyOptimisticOracle" contract. A new unit test file that replicates the testing for the normal "OptimisticOracle" is created.

**Summary**

All changes to `proposer.js` and `index.js` are backwards compatible with existing bots and they will not require any modifications to keep working with the `OptimisticOracle`.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
@@ -49,7 +49,11 @@ interface SkinnyRequest {
 }
 
 type OptimisticOracleContract = SkinnyOptimisticOracleWeb3 | OptimisticOracleWeb3;
-type OptimisticOracleType = "OptimisticOracle" | "SkinnyOptimisticOracle";
+
+export enum OptimisticOracleType {
+  OptimisticOracle,
+  SkinnyOptimisticOracle,
+}
 export class OptimisticOracleClient {
   public readonly oracle: OptimisticOracleContract;
   public readonly voting: VotingAncillaryInterfaceTestingWeb3;
@@ -74,7 +78,7 @@ export class OptimisticOracleClient {
    * @param {String} votingAddress Ethereum address of the Voting contract deployed on the current network.
    * @param {Number} lookback Any requests, proposals, or disputes that occurred prior to this timestamp will be ignored.
    * Used to limit the web3 requests made by this client.
-   * @param {OptimisticOracleType} oracleType Type of OptimisticOracle to query state for.
+   * @param {OptimisticOracleType} oracleType Index of allowed OptimisticOracle types.
    * @return None or throws an Error.
    */
   constructor(
@@ -85,7 +89,7 @@ export class OptimisticOracleClient {
     oracleAddress: string,
     votingAddress: string,
     public readonly lookback: number = 604800, // 1 Week
-    public readonly oracleType: OptimisticOracleType = "OptimisticOracle"
+    public readonly oracleType: OptimisticOracleType = OptimisticOracleType.OptimisticOracle
   ) {
     // Optimistic Oracle contract:
     this.oracle = (new web3.eth.Contract(oracleAbi, oracleAddress) as unknown) as OptimisticOracleContract;
@@ -114,7 +118,7 @@ export class OptimisticOracleClient {
   // Returns an array of expired Price Proposals that can be settled and that involved
   // the caller as the proposer
   public getSettleableProposals(caller: string): ProposePriceReturnValues[] {
-    if (this.oracleType === "SkinnyOptimisticOracle") {
+    if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
       return (this.expiredProposals as SkinnyProposePrice["returnValues"][]).filter((event) => {
         return ((event.request as unknown) as SkinnyRequest).proposer === caller;
       });
@@ -127,7 +131,7 @@ export class OptimisticOracleClient {
 
   // Returns disputes that can be settled and that involved the caller as the disputer
   public getSettleableDisputes(caller: string): DisputePriceReturnValues[] {
-    if (this.oracleType === "SkinnyOptimisticOracle") {
+    if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
       return (this.settleableDisputes as SkinnyDisputePrice["returnValues"][]).filter((event) => {
         return ((event.request as unknown) as SkinnyRequest).disputer === caller;
       });
@@ -217,7 +221,7 @@ export class OptimisticOracleClient {
 
     // Filter proposals based on their expiration timestamp:
     const isExpired = (proposal: ProposePriceReturnValues): boolean => {
-      if (this.oracleType === "SkinnyOptimisticOracle") {
+      if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
         return (
           Number(
             ((((proposal as unknown) as SkinnyProposePrice["returnValues"]).request as unknown) as SkinnyRequest)
@@ -290,7 +294,7 @@ export class OptimisticOracleClient {
     this.logger.debug({
       at: "OptimisticOracleClient",
       message: "Optimistic Oracle state updated",
-      oracleType: this.oracleType,
+      oracleType: OptimisticOracleType[this.oracleType],
       oracleAddress: this.oracle.options.address,
       lastUpdateTimestamp: this.lastUpdateTimestamp,
     });
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
@@ -12,24 +12,27 @@ import {
   SkinnyOptimisticOracleWeb3Events,
 } from "@uma/contracts-node";
 
-type RequestPrice =
-  | OptimisticOracleWeb3Events.RequestPrice["returnValues"]
-  | SkinnyOptimisticOracleWeb3Events.RequestPrice["returnValues"];
-type ProposePrice =
-  | OptimisticOracleWeb3Events.ProposePrice["returnValues"]
-  | SkinnyOptimisticOracleWeb3Events.ProposePrice["returnValues"];
-type DisputePrice =
-  | OptimisticOracleWeb3Events.DisputePrice["returnValues"]
-  | SkinnyOptimisticOracleWeb3Events.DisputePrice["returnValues"];
+type SkinnyRequestPrice = SkinnyOptimisticOracleWeb3Events.RequestPrice;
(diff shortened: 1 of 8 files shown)


packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -51,10 +84,12 @@ export class OptimisticOracleClient {
     public readonly web3: Web3,
     oracleAddress: string,
     votingAddress: string,
-    public readonly lookback = 604800 // 1 Week
+    public readonly lookback: number = 604800, // 1 Week
+    public readonly oracleType: OptimisticOracleType = "OptimisticOracle"
```
chrismaree, line 88: making this an `enum` of the two possible types is probs better.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -67,28 +102,40 @@ export class OptimisticOracleClient {
...
   }
 
   // Returns an array of expired Price Proposals that can be settled and that involved
   // the caller as the proposer
-  public getSettleableProposals(caller: string): ProposePrice[] {
-    return this.expiredProposals.filter((event) => {
-      return event.proposer === caller;
-    });
+  public getSettleableProposals(caller: string): ProposePriceReturnValues[] {
+    if (this.oracleType === "SkinnyOptimisticOracle") {
```
chrismaree, line 117: having this not be string comparison would be nice.

packages/financial-templates-lib/test/clients/OptimisticOracleClient.js
```diff
@@ -126,203 +127,418 @@ describe("OptimisticOracleClient.js", function () {
...
-    await client.update();
-    result = client.getSettleableProposals(proposer);
-    objectsInArrayInclude(result, []);
-  });
-
-  it("Basic dispute lifecycle: request, propose, dispute, resolve & settle", async function () {
-    // Initial update.
-    await client.update();
+  describe("Default OptimisticOracle contract", function () {
+    it("Basic proposal lifecycle: request, propose, expire without dispute", async function () {
```
chrismaree, line 131: was confused why the diff was so large. I see now 👀

packages/optimistic-oracle/index.js
```diff
@@ -32,10 +33,11 @@ async function run({
   commonPriceFeedConfig,
   optimisticOracleProposerConfig,
   oracleType = "Voting",
+  optimisticOracleType = "OptimisticOracle",
```
chrismaree, line 36: should probs verify the type passed in here to ensure it matches with either `OptimisticOracle` or `SkinnyOptimisticOracle`

packages/optimistic-oracle/index.js
```diff
@@ -148,6 +153,11 @@ async function Poll(callback) {
       // Type of "Oracle" set for this network's Finder, default is "Voting". Other possible types include "SinkOracle",
       //  "OracleChildTunnel", and "MockOracleAncillary"
       oracleType: process.env.ORACLE_TYPE ? process.env.ORACLE_TYPE : "Voting",
+      // Type of "OptimisticOracle" to load in client, default is "OptimisticOracle". The only other possible type is
+      // "SkinnyOptimisticOracle."
+      optimisticOracleType: process.env.OPTIMISTIC_ORACLE_TYPE
```
chrismaree, line 158: could do the validation here

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
chrismaree: where this is used is in js and a string is probs better there...I think that's ok just please add validation where it's read in within the `index` file and in this file that it is one of the two possible valid options and error if not.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -199,16 +202,29 @@ class OptimisticOracleProposer {
...
-    const proposal = this.optimisticOracleContract.methods.proposePrice(
-      priceRequest.requester,
-      this.utf8ToHex(priceRequest.identifier),
-      priceRequest.timestamp,
-      priceRequest.ancillaryData,
-      proposalPrice
-    );
+    let proposal;
+    if (this.optimisticOracleClient.oracleType === "SkinnyOptimisticOracle") {
+      proposal = this.optimisticOracleContract.methods.proposePrice(
```
chrismaree, line 207: very nice.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -220,19 +236,35 @@ class OptimisticOracleProposer {
...
+      if (this.optimisticOracleClient.oracleType === "SkinnyOptimisticOracle") {
+        logResult = {
+          tx: receipt.transactionHash,
+          requester: receipt.events.ProposePrice.returnValues.requester,
+          proposer: receipt.events.ProposePrice.returnValues.request.proposer,
+          identifier: this.hexToUtf8(receipt.events.ProposePrice.returnValues.identifier),
+          ancillaryData: receipt.events.ProposePrice.returnValues.ancillaryData || "0x",
+          timestamp: receipt.events.ProposePrice.returnValues.timestamp,
+          request: receipt.events.ProposePrice.returnValues.request,
+          proposedPrice: receipt.events.ProposePrice.returnValues.request.proposedPrice,
```
chrismaree, line 249: oof to this change. otherwise, they are identical

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
nicholaspai: How can we check type conditionally?

packages/optimistic-oracle/index.js
nicholaspai: We should do validation for the `oracleType` as well, will review

packages/optimistic-oracle/src/proposer.js
nicholaspai: yep...

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
nicholaspai: Actually, enums are easy. I changed to enums

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
nicholaspai: Made it an enum!

packages/optimistic-oracle/index.js
nicholaspai: Done by changing to an enum

packages/optimistic-oracle/index.js
nicholaspai: Actually we should make that a different PR so it doesn't break production bots

packages/optimistic-oracle/src/proposer.js
```diff
@@ -314,19 +368,34 @@ class OptimisticOracleProposer {
...
-          ancillaryData: receipt.events.DisputePrice.returnValues.ancillaryData || "0x",
-          timestamp: receipt.events.DisputePrice.returnValues.timestamp,
-          proposedPrice: receipt.events.DisputePrice.returnValues.proposedPrice,
-        };
+        let logResult;
+        if (this.optimisticOracleClient.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
+          logResult = {
+            tx: receipt.transactionHash,
+            requester: receipt.events.DisputePrice.returnValues.requester,
+            proposer: receipt.events.DisputePrice.returnValues.request.proposer,
```
mrice32, line 376: This may be more work than it's worth, but for most of these logs, we could avoid the if entirely if we did something like this:
```js
const logResult = {
  tx: receipt.transactionHash,
  ...receipt.events.DisputePrice.returnValues.request, // This may be undefined sometimes, but that's fine!
  ...receipt.events.DisputePrice.returnValues,
  // We only need to manually list off any fields that require a transformation
  ancillaryData: receipt.events.DisputePrice.returnValues.ancillaryData || "0x",
  identifier: this.hexToUtf8(receipt.events.DisputePrice.returnValues.identifier)
};
```

I think this strategy should work for most, if not all, of the logs.

I'm totally fine leaving things the way they are, but just wanted to point this out in case you felt it was cleaner.

mrice32 (approved):
LGTM! Two minor comments

packages/optimistic-oracle/index.js
```diff
@@ -32,10 +40,11 @@ async function run({
   commonPriceFeedConfig,
   optimisticOracleProposerConfig,
   oracleType = "Voting",
+  optimisticOracleType = OptimisticOracleType.OptimisticOracle,
 }) {
   try {
     const [accounts, networkId] = await Promise.all([web3.eth.getAccounts(), web3.eth.net.getId()]);
-    const optimisticOracleAddress = await getAddress("OptimisticOracle", networkId);
+    const optimisticOracleAddress = await getAddress(OptimisticOracleType[optimisticOracleType], networkId);
```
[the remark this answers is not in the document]
nicholaspai: Oh you're right I thought mistakenly it was an int enum

packages/optimistic-oracle/src/proposer.js
nicholaspai: Oh great call, this is much cleaner.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -376,20 +428,20 @@ class OptimisticOracleProposer {
 
       const logResult = {
         tx: receipt.transactionHash,
-        requester: receipt.events.Settle.returnValues.requester,
-        proposer: receipt.events.Settle.returnValues.proposer,
-        disputer: receipt.events.Settle.returnValues.disputer,
+        // This is undefined unless the oracle type is SkinnyOptimisticOracle.
```
mrice32, line 431: +1 very helpful comment

packages/optimistic-oracle/src/proposer.js
```diff
@@ -118,6 +121,7 @@ class OptimisticOracleProposer {
   async settleRequests() {
     this.logger.debug({
       at: "OptimisticOracleProposer#settleRequests",
+      oracleType: OptimisticOracleType[this.optimisticOracleClient.oracleType],
```
mrice32, line 124: Could we swap all these for just `oracleType: this.optimisticOracleClient.oracleType` here and in the rest of the logs (looks like there are about 9 spots)?

mrice32 (approved):
LGTM!

packages/optimistic-oracle/src/proposer.js
nicholaspai: Oh shoot thought I did this, will fix

PR #3559, opened 2021-11-04
feat(across-relayer): Enable batch relays via multicall and improve log production

chrismaree:
**Motivation**

This PR enables batches of transactions to be sent at once using multicall. It also improves how logs are produced, leveraging the markdown slack transport.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/common/src/FormattingUtils.ts
```diff
@@ -71,7 +71,6 @@ export const formatWithMaxDecimals = (
 };
 
 export const createFormatFunction = (
-  _web3: Web3 | undefined, // unused
```
chrismaree, line 74: this field is unused. I removed it from here an all places that use this util to simplify things a bit.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -53,6 +53,8 @@ export interface BridgePoolData {
   contract: BridgePoolWeb3;
   currentTime: number;
   relayNonce: number;
+  poolCollateralDecimals: number;
```
chrismaree, line 56: we need these to enhance logging.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -111,13 +113,16 @@ export class InsuredBridgeL1Client {
   }
 
   hasInstantRelayer(l1Token: string, depositHash: string, realizedLpFeePct: string): boolean {
+    this._throwIfNotInitialized();
+    return this.getInstantRelayer(l1Token, depositHash, realizedLpFeePct) !== undefined;
```
chrismaree, line 117: ideally re-use existing methods.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -174,9 +179,17 @@ export class InsuredBridgeL1Client {
     return ClientRelayState.Finalized;
   }
 
+  getBridgePoolCollateralInfoForDeposit(l2Deposit: Deposit): { collateralDecimals: number; collateralSymbol: string } {
```
chrismaree, line 182: returns the new data added for the decimals and symbol

packages/financial-templates-lib/src/logger/SpyTransport.ts
```diff
@@ -18,7 +18,12 @@ export class SpyTransport extends Transport {
 
   async log(info: any, callback: () => void): Promise<void> {
     // Add an `logInTest` option to help with debugging to bots in tests by printing all logs received by winston.
-    if (argv._.includes("logInTest")) console.log(JSON.stringify(info, null, 2));
+    if (argv._.includes("logInTest"))
```
chrismaree, line 21: this does not work correctly if there is an error in the log. took me a while to figure out why I was not seeing the logs.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -57,27 +58,27 @@ export class Relayer {
   ) {}
 
   async checkForPendingDepositsAndRelay(): Promise<void> {
-    this.logger.debug({ at: "InsuredBridgeRelayer#Relayer", message: "Checking for pending deposits and relaying" });
+    this.logger.debug({ at: "AcrossRelayer#Relayer", message: "Checking for pending deposits and relaying" });
```
chrismaree, line 61: I renamed these to be more "correct" in slack.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,58 +94,48 @@ export class Relayer {
           continue;
         }
         try {
-          await this._relayPendingDeposit(l1Token, relayableDeposit);
+          relayTransactions.push(await this._generateRelayTransactionForPendingDeposit(l1Token, relayableDeposit));
```
chrismaree, line 97: we simply build an array of relay transactions and then send them all at once using the `processTransactionBatch` Method.

This is the main change in this PR: previously this method would actually run the transactions. now, it simply generates the transaction objects and passes them to the `processTransactionBatch` method to deal with them.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,58 +94,48 @@ export class Relayer {
...
 
   async checkForPendingRelaysAndDispute(): Promise<void> {
-    this.logger.debug({ at: "InsuredBridgeRelayer#Disputer", message: "Checking for pending relays and disputing" });
+    this.logger.debug({ at: "AcrossRelayer#Disputer", message: "Checking for pending relays and disputing" });
 
-    // Build dictionary of pending relays keyed by l1 token and deposit hash. We assume that getPendingRelays() filters
+    // Build dictionary of pending relays keyed by l1 token and deposit hash. getPendingRelays() filters
     // out Finalized relays.
-    const pendingRelays: Relay[] = this.getPendingRelays();
+    const pendingRelays: Relay[] = this.l1Client.getPendingRelayedDeposits();
```
chrismaree, line 114: I removed this internal method. it simply called a function on the `l1Client`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -155,21 +146,24 @@ export class Relayer {
...
       if (settleableRelays.length == 0) {
-        this.logger.debug({ at: "InsuredBridgeRelayer#Finalizer", message: "No settleable relays" });
+        this.logger.debug({ at: "AcrossRelayer#Finalizer", message: "No settleable relays" });
         return;
       }
 
       for (const settleableRelay of settleableRelays) {
         try {
-          await this.settleRelay(this.l2Client.getDepositByHash(settleableRelay.depositHash), settleableRelay);
+          settleRelayTransactions.push(
```
chrismaree, line 155: as with relay, build an array then send all at once.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -230,7 +224,7 @@ export class Relayer {
     }
 
     // Fetch deposit for relay.
-    const deposit = await this.matchRelayWithDeposit(relay);
+    const deposit = await this._matchRelayWithDeposit(relay);
```
chrismaree, line 227: I did a fair bit of re-naming as well for consistency with the rest of this file.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -461,167 +460,28 @@ export class Relayer {
     return RelaySubmitType.Ignore;
   }
 
-  private async slowRelay(deposit: Deposit, realizedLpFeePct: BN) {
```
chrismaree, line 464: great thing about this PR is we can remove all this custom tx processing logic for each transaction we can send!

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -461,167 +460,28 @@ export class Relayer {
...
-        });
-      else throw receipt;
-    } catch (error) {
-      this.logger.error({ at: "InsuredBridgeRelayer#Finalizer", message: "Something errored settling relay!", error });
-    }
+  private async _generateSettleTransactionForSettleableRelay(deposit: Deposit, relay: Relay) {
+    return {
+      transaction: this._generateSettleRelayTx(deposit, relay),
+      message: "Relay settled 💸",
+      mrkdwn: this._generateMarkdownForSettle(deposit, relay),
```
chrismaree, line 467: rather, we simply define a `message` and some `mrkdwn`. This information is far more rich and detailed than before, in a more ocnsumable way than it was.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -461,167 +460,28 @@ export class Relayer {
...
+    };
   }
 
-  private async disputeRelay(deposit: Deposit, relay: Relay) {
+  private async _disputeRelay(deposit: Deposit, relay: Relay) {
```
chrismaree, line 471: note that sending disputes does not have any batching options. hopefully we are not sending batches of disputes...

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
         } else throw receipt;
       } else throw receipt;
     } catch (error) {
-      this.logger.error({ at: "InsuredBridgeRelayer#Disputer", message: "Something errored disputing relay!", error });
+      this.logger.error({ at: "AcrossRelayer#Disputer", message: "Something errored disputing relay!", error });
     }
   }
 
-  private generateSlowRelayTx(deposit: Deposit, realizedLpFeePct: BN): TransactionType {
+  private async _processTransactionBatch(
```
chrismaree, line 509: this method contains the real logic in this PR on dealing with sets of transactions. Please look at it closely!

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -727,66 +670,121 @@ export class Relayer {
...
-        break;
+        return {
+          transaction: this._generateInstantRelayTx(relayableDeposit.deposit, realizedLpFeePct),
+          message: "Relay instantly sent 🚀",
+          mrkdwn,
+        };
     }
   }
 
+  private _generateMarkdownForRelay(deposit: Deposit, realizedLpFeePct: BN) {
```
chrismaree, line 740: some markdown magic for a relay :)

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: oh boy

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,58 +94,48 @@ export class Relayer {
...
         try {
-          await this._relayPendingDeposit(l1Token, relayableDeposit);
+          relayTransactions.push(await this._generateRelayTransactionForPendingDeposit(l1Token, relayableDeposit));
         } catch (error) {
-          this.logger.error({
-            at: "InsuredBridgeRelayer#Relayer",
-            message: "Unexpected error processing deposit",
-            error,
-          });
+          console.log("error", error);
```
nicholaspai, line 99: leftover log

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
     }
   }
 
-  private generateSlowRelayTx(deposit: Deposit, realizedLpFeePct: BN): TransactionType {
+  private async _processTransactionBatch(
+    transactions: { transaction: TransactionType | any; message: string; mrkdwn: string }[]
+  ) {
+    // Remove any undefined transaction objects or objects that contain null transactions.
+    transactions = transactions.filter((transaction) => {
+      if (!transaction) return false;
```
nicholaspai, line 514: why not do `return transaction && transaction.transaction`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+      const { receipt } = await runTransaction({
+        web3: this.l1Client.l1Web3,
+        transaction,
+        transactionConfig: { ...this.gasEstimator.getCurrentFastPrice(), from: this.account },
+        availableAccounts: 1,
+      });
+      if (receipt) {
+        this.logger.info({
+          at: "AcrossRelayerTxProcessor",
+          message,
```
nicholaspai, line 576: Is the event receipt still shown?

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -80,7 +80,6 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
       config.whitelistedRelayL1Tokens,
       accounts[0],
       config.whitelistedChainIds,
-      config.deployTimestamps,
```
nicholaspai, line 83: Why remove? We still use this in Relayer

packages/insured-bridge-relayer/test/Relayer.ts
```diff
@@ -1344,4 +1357,78 @@ describe("Relayer.ts", function () {
       assert.equal(targetLog.length, 1);
     });
   });
+  describe("Mutlicall Batching", function () {
```
nicholaspai, line 1360:
```suggestion
  describe("Multicall Batching", function () {
```

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: sure.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+    transactions: { transaction: TransactionType | any; message: string; mrkdwn: string }[]
+  ) {
+    // Remove any undefined transaction objects or objects that contain null transactions.
+    transactions = transactions.filter((transaction) => {
+      if (!transaction) return false;
+      if (!transaction.transaction) return false;
+      return true;
+    });
+
+    if (transactions.length == 0) return;
```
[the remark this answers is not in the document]
chrismaree: it can get here with no transactions (in fact it will get here every time the bot runs!) as each block that _can_ send transactions simply calls this method with an empty array if there is nothing to send. 

the return early logic could be placed where this function is called by that would be more repetitive than this implementation.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: nope! It's up to the `mrkdwn` to contain enough logging information. to inform the user of what happened. in the tx.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+    }
+    if (transactions.length > 1) {
+      // The `to` field in the transaction must be the same for all transactions or the batch processing will not work.
+      // This should be a MultiCaller enabled contract.
+      const targetMultiCallerContract = new this.l1Client.l1Web3.eth.Contract(
+        getAbi("MultiCaller"),
+        transactions[0].transaction._parent._address
+      );
+
+      for (const transaction of transactions) {
```
mrice32, line 532: nit: could be a little simpler if we do `if(transactions.some(...)) throw ...;`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+        if (targetMultiCallerContract.options.address != transaction.transaction._parent._address)
```
mrice32, line 533: Just for the sake of consistency and using as few private variables as possible.
```suggestion
        if (targetMultiCallerContract.options.address != transaction.transaction._parent.options.address)
```

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
       } else throw receipt;
     } catch (error) {
-      this.logger.error({ at: "InsuredBridgeRelayer#Disputer", message: "Something errored disputing relay!", error });
+      this.logger.error({ at: "AcrossRelayer#Disputer", message: "Something errored disputing relay!", error });
     }
   }
 
-  private generateSlowRelayTx(deposit: Deposit, realizedLpFeePct: BN): TransactionType {
+  private async _processTransactionBatch(
+    transactions: { transaction: TransactionType | any; message: string; mrkdwn: string }[]
```
mrice32, line 510: nit: Why did we need to add `| any` (just curious -- I know this is sometimes necessary to get things to compile).

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+
+      // Iterate over all transactions and build up a set of multicall blocks and a block of markdown to send to slack
+      // to make the set of transactions readable.
+      const multiCallTransaction = transactions.map((transaction) => transaction.transaction.encodeABI());
+
+      let mrkdwnBlock = "*Transactions sent in batch:*\n";
+      transactions.forEach((transaction) => {
+        mrkdwnBlock += `  • ${transaction.message}:\n`;
+        mrkdwnBlock += `    - ${transaction.mrkdwn}\n`;
+      });
```
mrice32, line 545: +1 very important to get the logging right on these as they're hard to read in etherscan

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+        mrkdwnBlock
+      );
+
+      // In the event the batch transaction was unsuccessful, iterate over all transactions and send them individually.
+      if (!batchTxSuccess) {
+        for (const transaction of transactions) {
+          this.logger.info({ at: "AcrossRelayerTxProcessor", message: "Sending batched transactions individually😷" });
+          await this._sendTransaction(transaction.transaction, transaction.message, transaction.mrkdwn);
+        }
+      }
```
mrice32, line 560: If the goal is to have bad transactions not block others, should we fire these off using `Promise.allSettled` and then send out errors using `logger.error()` for any that fail? Also, doing these in parallel will probably speed the bot up a lot.

packages/insured-bridge-relayer/src/Relayer.ts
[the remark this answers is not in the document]
nicholaspai: Hmm I seem ok sgtm

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -642,11 +502,94 @@ export class Relayer {
...
+      if (receipt) {
+        this.logger.info({
+          at: "AcrossRelayerTxProcessor",
+          message,
+          mrkdwn: mrkdwn + " tx: " + createEtherscanLinkMarkdown(receipt.transactionHash),
+        });
+        return true;
+      } else throw receipt;
+    } catch (error) {
+      console.log("error", error);
```
nicholaspai, line 582: Leftover log

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,58 +94,48 @@ export class Relayer {
...
-          this.logger.error({
-            at: "InsuredBridgeRelayer#Relayer",
-            message: "Unexpected error processing deposit",
-            error,
-          });
+          console.log("error", error);
+          this.logger.error({ at: "AcrossRelayer#Relayer", message: "Unexpected error processing deposit", error });
         }
       }
+      await this._processTransactionBatch(relayTransactions as any);
```
nicholaspai, line 103: Shouldn't we try-catch this in case `_processTransactionBatch` fails, and the relayer fails, but the disputer/finalizer would still succeed?

packages/insured-bridge-relayer/test/Relayer.ts
```diff
@@ -456,6 +456,9 @@ describe("Relayer.ts", function () {
       assert.isTrue(lastSpyLogIncludes(spy, "No relayable deposits"));
     });
     it("Two deposits, one relay fails, one succeeds", async function () {
+      // Validates that one failed tx does not break the other in multi execution. Also validates unbundeling with
```
nicholaspai, line 459:
```suggestion
      // Validates that one failed tx does not break the other in multi execution. Also validates unbundling with
```

mrice32 (approved):
LGTM!

packages/insured-bridge-relayer/test/Relayer.ts
```diff
@@ -1344,4 +1357,78 @@ describe("Relayer.ts", function () {
...
+
+      // Advance time such that relay has expired and check that bot correctly identifies it as expired.
+      const expirationTime = Number(relayTime.toString()) + defaultLiveness;
+      await bridgePool.methods.setCurrentTime(expirationTime).send({ from: l1Owner });
+      await l1Client.update();
+      await relayer.checkForPendingDepositsAndRelay();
+      assert.isTrue(lastSpyLogIncludes(spy, "Pending relay has expired"));
+
+      // Finally, settle. We should be able to do both of these in one tx.
+      await relayer.checkforSettleableRelaysAndSettle();
```
nicholaspai, line 1427: Right before this line, what would happen if you sent a `deposit` that caused the Relayer to fail? Would it still be able to finalize the expired relays, despite failing to relay the deposit?

packages/insured-bridge-relayer/src/index.ts
chrismaree: was a mistake. fixed!

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: these are sent separately to avoid this exact kind of flow. Originally when designing this I wanted to batch _all_ transactions to the `processTransactionBatch` method for one particular contract. I decided agains this as you'd need the exact kind of logic you describe. the PR in it's current form will only send one kind of transaction to one destination at a time in a batch.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: reason is two fold: when we dont have transactions to append we use `null` and secondly the `._parent.*` props, used later on, are not included in the `TransactionType`.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: in future, yes! that would be great. for now though this won't work as we dont have a robust enough way to deal with nonce collisions for parallel transactions being sent. I spent a few days earlier this week working on a technique to send this kind of transaction. I build out a system were you can "fire and forget" transactions; i.e only wait until they are in the mempool and you have a tx hash and not block for block inclusion. however, I realised that how we detect pending nonces (transactions in the mempool) is not robust enough for this kind of thing. it would _sometimes_ find the pending transactions and other times would not. This is a symptom of how quickly I was sending transactions.

To really implement this kind of logic you need some object that will: a) track the starting nonce b) send transactions and increment it's local nonce c) manage nonce incrementing for the set of transactions.

I'd like to build this out as a full utility that the transaction management offers in the `common` package. 

for now, though, this will just be a bit slow :)

packages/insured-bridge-relayer/test/Relayer.ts
chrismaree: yup! those are separate actions and do not interfere with one another. batches of txs are sent for: a) relay and b) settlement separately.

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: I don't think I fully understand, since `_processTransactionBatch` is outside a try-catch, won't this failure on line 104 cause the Finalizer/Disputer to be blocked?

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: Yes great call out. Can you add a comment about this? We shouldn't be batching disputes because we don't expect to send batches of them

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: Can you extend the `TransactionType` to include those other props to avoid using `any`?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -727,66 +652,159 @@ export class Relayer {
...
+          message: "Relay instantly sent 🚀",
+          mrkdwn,
+        };
     }
   }
 
+  private _generateMarkdownForRelay(deposit: Deposit, realizedLpFeePct: BN) {
+    return (
+      "Relayed " +
+      this._generateMrkdwnDepositIdNetworkSizeFromTo(deposit) +
```
nicholaspai, line 725: As mentioned IRL I think its important to at least include the deposit and relay ID for easier tracking

nicholaspai: I suppose you could find those easily in the etherscan link of the `tx`

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -727,66 +652,159 @@ export class Relayer {
...
+    );
+  }
+
+  private _generateMrkdwnForDispute(deposit: Deposit, relay: Relay) {
+    return (
+      "Disputed " +
+      this._generateMrkdwnDepositIdNetworkSizeFromTo(deposit) +
+      "DepositHash " +
+      deposit.depositHash +
+      " relayAncillaryDataHash " +
```
nicholaspai, line 752: +1. This ancillary data can be matched against a SkinnyOO price request's ancillary data by parsing the price request's UTF8-decoded ancillary data.

packages/insured-bridge-relayer/test/Relayer.ts
nicholaspai: What if the batch of `relays` throws an error?

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: done

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: done!

packages/insured-bridge-relayer/test/Relayer.ts
chrismaree: The part of the code where the error would be thrown has a  try catch in it. it behaves the same way that the current pr would though I belive.

nicholaspai (approved):
LGTM thanks for reviewing my comments

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -93,58 +96,51 @@ export class Relayer {
...
         } catch (error) {
-          this.logger.error({
-            at: "InsuredBridgeRelayer#Relayer",
-            message: "Unexpected error processing deposit",
-            error,
-          });
+          this.logger.error({ at: "AcrossRelayer#Relayer", message: "Unexpected error processing deposit", error });
         }
       }
+      try {
```
chrismaree, line 104: @nicholaspai wrapping the `_processTransactionBatch` that could error. With this the bot should never halt to block another action.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: yup. Let me get that right.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: Deposit ID is included already. see `_generateMrkdwnDepositIdNetworkSizeFromTo`

Relay ID is not know until we get the events. I can refactor this to capture this data but would like to leave that for a subsequent PR.

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: after discussing more we decided to add additional try catch blocks.

PR #3562, opened 2021-11-04
improve(across-relayer): Handle quote times in future

nicholaspai:
**Motivation**

It is possible that a deposit can have a quote timestamp that is in the future and therefore would produce unexpected realized LP fee %'s (or cause errors) and we should handle them.

If a deposit.quoteTime is > relay.blockTime, then the pending relay should be disputed, because the realized LP fee % would have been impossible to compute at relay time.

If the deposit.quoteTime <= relay.blockTime, but deposit.quoteTime > latest web3 block time, then ignore it in both the Relayer and Disputer because we cannot compute a realized LP fee for it yet.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -249,12 +252,17 @@ export class Relayer {
...
-      if (deposit.quoteTimestamp < this.deployTimestamps[deposit.l1Token].timestamp) {
+      const latestBlockTime = Number((await this.l1Client.l1Web3.eth.getBlock("latest")).timestamp);
+      if (
+        deposit.quoteTimestamp < this.deployTimestamps[deposit.l1Token].timestamp ||
+        deposit.quoteTimestamp > latestBlockTime
+      ) {
         this.logger.debug({
           at: "Disputer",
-          message: "Deposit quote time < bridge pool deployment for L1 token, disputing",
+          message: "Deposit quote time < bridge pool deployment for L1 token or > latest block time, disputing",
```
chrismaree, line 262: This makes a lot of sense. Good catch!

chrismaree (approved):
Good catch. This kind of variability in the deposit data could be very problematic.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -81,17 +81,25 @@ export class Relayer {
...
             at: "InsuredBridgeRelayer#Relayer",
-            message: "Deposit quote time < bridge pool deployment for L1 token, skipping",
+            message: "Deposit quote time < bridge pool deployment for L1 token or > latest block time, skipping",
             deposit: relayableDeposit.deposit,
             deploymentTime: this.deployTimestamps[relayableDeposit.deposit.l1Token].timestamp,
+            latestBlockTime,
           });
           continue;
         }
+        // Now, check if deposit.quoteTimestamp > latestBlockTime
```
nicholaspai, line 98: This PR is in draft mode until I implement this new edge case handling @chrismaree @mrice32 

Also similar situation in Disputer flow

nicholaspai: Done

PR #3565, opened 2021-11-05
feat(sdk): add tx receipt decoding events to bridgepool client

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/2955/improve-event-handling-in-bridgepool-sdk-client

**Summary**

Adds another function to process events from transaction receipt to the PoolEventState class.  This also dedupes events in both event handling functions to ensure we dont get wrong balance. Order of events for these particular states is not important. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/2955/improve-event-handling-in-bridgepool-sdk-client

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -2,10 +2,11 @@ import assert from "assert";
 import { bridgePool } from "../../clients";
 import { BigNumber, Signer } from "ethers";
 import { toBNWei, fixedPointAdjustment, calcInterest, calcApy, fromWei, BigNumberish } from "../utils";
-import { BatchReadWithErrors, loop } from "../../utils";
+import { BatchReadWithErrors, loop, exists } from "../../utils";
 import Multicall2 from "../../multicall2";
 import TransactionManager from "../transactionManager";
-import { TransactionRequest, TransactionReceipt, Provider } from "@ethersproject/abstract-provider";
+import { ethers } from "ethers";
+import { TransactionRequest, TransactionReceipt, Provider, Log } from "@ethersproject/abstract-provider";
 import set from "lodash/set";
 import get from "lodash/get";
 import has from "lodash/has";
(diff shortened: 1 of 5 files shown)


packages/sdk/src/index.ts
```diff
@@ -28,6 +28,12 @@ export interface HasId<I> {
 export interface Callable {
   (...args: any[]): any;
 }
+
+export type SerializableEvent = Omit<
```
daywiss, line 32: this is needed because we cant easily make full `Event` types from `Logs` or `LogDescription`, this type is compatible with `Event`, and `Logs & LogDescription` types

amateima (approved):

packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -96,30 +97,77 @@ class PoolState {
...
-      if (a.transactionIndex < b.transactionIndex) return -1;
-      if (a.logIndex < b.logIndex) return -1;
-      return 1;
-    });
+    )
+      .filter(this.filterSeen)
+      .sort((a, b) => {
+        if (a.blockNumber < b.blockNumber) return -1;
+        if (a.transactionIndex < b.transactionIndex) return -1;
+        if (a.logIndex < b.logIndex) return -1;
```
mrice32, line 137: Does this work for primary + secondary sorting?

For instance, what happens if `a.blocknumber > b.blockNumber`, but `a.transactionIndex < b.transactionIndex`?

I think the correct implementation is something more like:
```ts
if (a.blockNumber !== b.blockNumber) return a.blockNumber - b.blockNumber;
if (a.transactionIndex !== b.transactionIndex) return a.transactionIndex - b.transactionIndex;
...
```

mrice32 (commented):
Looks great! One comment on the correctness of sorting.

packages/sdk/src/across/clients/bridgePool.ts
daywiss: ohh wow yea you are right. good catch!

PR #3585, opened 2021-11-08
feat(across-relayer): Add ability to finalize L2->L1 transfer actions

chrismaree:
**Motivation**

This PR does three things:
- Adds the ability for the relayer to send L2->L1 relay actions over the canonical Optimism/Arbitrum bridge. In this implementation, this tx type will be generated as soon as there are funds to be relayed.
- Refactors Smart contract imports to remove a temp library
- Adds a modification to contract interfaces in the `BridgePool` to include the `bridgeTokens` method to simplify implementers interface.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,268 @@
+import {
```
chrismaree, line 1: these tests are somewhat bloated. The reason is they'll need to consider the implementation of the `checkForFinalizedCanonicalRelaysAndFinalize` method, which will require all the L1 contracts deployed here.

packages/core/contracts/insured-bridge/BridgeDepositBox.sol
```diff
@@ -4,23 +4,8 @@ pragma solidity ^0.8.0;
 import "../common/implementation/Testable.sol";
 import "../common/implementation/Lockable.sol";
 
-// Define some interfaces and helper libraries. This is temporary until we can bump the solidity version in these
```
chrismaree, line 7: we should have removed this before.

packages/core/contracts/insured-bridge/BridgeDepositBox.sol
```diff
@@ -155,6 +141,8 @@ abstract contract BridgeDepositBox is Testable, Lockable {
         emit DepositsEnabled(l2Token, depositsEnabled);
     }
 
+    function bridgeTokens(address l2Token, uint32 l2Gas) public virtual;
```
chrismaree, line 144: this is the virtual interface I added. this means that implementers only need to use `BridgeDepositBox` when calling methods and can ignore chain-specific interfaces.

Note that all other important methods in this file contain the same type of interface

packages/core/contracts/insured-bridge/BridgeDepositBox.sol
```diff
@@ -204,7 +192,7 @@ abstract contract BridgeDepositBox is Testable, Lockable {
         // Else, it is a normal ERC20. In this case pull the token from the users wallet as per normal.
         // Note: this includes the case where the L2 user has WETH (already wrapped ETH) and wants to bridge them. In
         // this case the msg.value will be set to 0, indicating a "normal" ERC20 bridging action.
-        else TokenHelper.safeTransferFrom(l2Token, msg.sender, address(this), amount);
+        else IERC20(l2Token).safeTransferFrom(msg.sender, address(this), amount);
```
chrismaree, line 195: somewhat unfortunate artifact of the transfer helper removal is the need to cast these addresses to interfaces.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,83 @@
...
+import type { TransactionType } from "@uma/common";
+
+export enum RelaySubmitType {
+  Slow,
+  SpeedUp,
+  Instant,
+  Ignore,
+}
+
+export class CrossDomainFinalizer {
```
chrismaree, line 16: new class. will add comments soon.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+  constructor(
+    readonly logger: winston.Logger,
+    readonly gasEstimator: GasEstimator,
+    readonly l1Client: InsuredBridgeL1Client,
+    readonly l2Client: InsuredBridgeL2Client,
+    readonly account: string
+  ) {}
+  async checkForBridgeableL2TokensAndBridge() {
+    this.logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "Checking bridgeable L2 tokens" });
```
nicholaspai, line 23: This is cool to think of this bot as part of the `AcrossRelayer` bot but just provides another feature, which is x-domain finalizing!

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    const whitelistedTokenEvents = await this.l1Client.bridgeAdmin.getPastEvents("WhitelistToken", {
```
nicholaspai, line 25: Although its inconsistent to query events in bots in this repo, I think this is a performance optimization as we don't want to query `WhitelistToken` in the InsuredBridgeL1Client

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+      fromBlock: 0,
+      toBlock: await this.l1Client.l1Web3.eth.getBlockNumber(),
```
nicholaspai, line 27: We should ideally parameterize the `fromBlock` as `this.firstBlockToSearch` and `toBlock` as `this.endingBlockNumber` so that on each subsequent run, this bot only looks up new events.

Perhaps we should build a `BridgeAdminClient` that handles event processing....

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    });
+
+    // Extract the l2Tokens that have been whitelisted. The new Set syntax acts to remove any duplicates from the array.
+    // This would be the case if a l2 token was re - whitelisted after upgrading the bridgePool contract on L1.
```
nicholaspai, line 31: nice comment

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    const whitelistedL2Tokens = [...new Set(whitelistedTokenEvents.map((event) => event.returnValues.l2Token))];
+
+    // Check if any of the whitelisted l2Tokens are bridgeable. Do this in one parallel call. Returns an array of bool
+    // for each l2Token, describing if it can be bridged from L2->L1.
+    const canBridge = await Promise.all(
+      whitelistedL2Tokens.map((l2Token) => this.l2Client.bridgeDepositBox.methods.canBridge(l2Token).call())
```
nicholaspai, line 37: Very nicely written. +1 parallelization

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    );
+
+    // For each canBridge result, check if it is true. If so, then we can bridge the token.
+    const bridgeableL2Tokens = whitelistedL2Tokens.filter((l2Token, index) => canBridge[index]);
```
nicholaspai, line 41:
```suggestion
    const bridgeableL2Tokens = whitelistedL2Tokens.filter((, index) => canBridge[index]);
```

I think this works?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
nicholaspai, line 23: I think you should emit the L1 client's bridge admin address somewhere, perhaps in this log

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    // Finally, iterate over the bridgeable l2Tokens and bridge them.
+    if (bridgeableL2Tokens.length == 0) {
+      this.logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "No bridgeable L2 tokens" });
+      return;
+    }
+    let nonceCounter = await this.l2Client.l2Web3.eth.getTransactionCount(this.account);
```
nicholaspai, line 48: nit: Just rename to `nonce`. Nonce and counter are the same thing no?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    for (const l2Token of bridgeableL2Tokens) {
+      // Track the account nonce and manually increment on each TX. We need to do this because the L2 transactions
+      // process quicker than the infura node updates and we need to avoid the nonce collision.
+      try {
+        // Note that this tx sending method is NOT using TransactionUtils runTransaction as it is not required on L2.
+        const receipt = await this.l2Client.bridgeDepositBox.methods
+          .bridgeTokens(l2Token, "0")
```
nicholaspai, line 55: Add a comment that the second param is `l2Gas` and is unused for both Arbitrum and Optimism for now

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    const whitelistedTokenEvents = await this.l1Client.bridgeAdmin.getPastEvents("WhitelistToken", {
+      fromBlock: 0,
+      toBlock: await this.l1Client.l1Web3.eth.getBlockNumber(),
+    });
+
+    // Extract the l2Tokens that have been whitelisted. The new Set syntax acts to remove any duplicates from the array.
+    // This would be the case if a l2 token was re - whitelisted after upgrading the bridgePool contract on L1.
+    const whitelistedL2Tokens = [...new Set(whitelistedTokenEvents.map((event) => event.returnValues.l2Token))];
+
+    // Check if any of the whitelisted l2Tokens are bridgeable. Do this in one parallel call. Returns an array of bool
```
nicholaspai, line 34: Do you think we should add an `if` statement that checks if the bridge deposit box's balance is > some threshold? If its too small then perhaps we hold off bridging to save gas? I know L2 gas is negligible but still

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,91 @@
...
+    // Finally, iterate over the bridgeable l2Tokens and bridge them.
+    if (bridgeableL2Tokens.length == 0) {
+      this.logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "No bridgeable L2 tokens" });
+      return;
+    }
+    let nonceCounter = await this.l2Client.l2Web3.eth.getTransactionCount(this.account);
+    for (const l2Token of bridgeableL2Tokens) {
+      // Track the account nonce and manually increment on each TX. We need to do this because the L2 transactions
+      // process quicker than the infura node updates and we need to avoid the nonce collision.
+      try {
```
nicholaspai, line 52: nit: put the logic inside this `try{}` block into an internal method `_bridgeL2Token` to make this function more readable and easier to maintain.

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,257 @@
...
+      await l1Token.methods.mint(l1LiquidityProvider, initialPoolLiquidity).send({ from: l1Owner });
+      await l1Token.methods
+        .approve(bridgePool.options.address, initialPoolLiquidity)
+        .send({ from: l1LiquidityProvider });
+      await bridgePool.methods.addLiquidity(initialPoolLiquidity).send({ from: l1LiquidityProvider });
+
+      // Mint some tokens for the L2 depositor.
+      await l2Token.methods.mint(l2Depositor, depositAmount).send({ from: l2Owner });
+    });
+    it("Can correctly detect and cross-domain transfers", async function () {
```
nicholaspai, line 216: Can you add a set of test where there are two L2 whitelisted tokens:
- Both L2 tokens can be bridged
- Only one can be bridged because `canBridge` returns false for one of them so only one bridging action is sent
- Both L2 tokens can be bridged but one fails to send for some reason. For example, you could whitelist one L2 token that always reverts on `safeTransferFrom`? I know this is annoying but definitely increases coverage.

nicholaspai (commented):
Amazing work, Chris. I left a lot of comments that but this is going to be super useful.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: correct. it's also way less code. I could move it into a helper perhaps, like that of the index helpers.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: nope. I tried that out but ts gave an error. I think a more correct way would be to use the `_` as the variable place holder? wdyt

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: yup! I will change this to be greater than x% of the amount of liquid + utilised funds on L1. i.e the funds waiting on to be bridged on the canonical bridge are greater than 5% of the funds on L1 then, and only then, bridge

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: This is emitted in the start up log of the the bot in the here: https://github.com/UMAprotocol/protocol/blob/2fc8fe4237156e68a02bf15ff9f981eed0598137/packages/insured-bridge-relayer/src/index.ts#L31

do you want it again here?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: we _could_....is it worth it though for this one look back? also this is running on L1 and infura has no problem in calling this kind of method.

we could also store the deployment block of the bridge admin in a config file to decrease how far back this has to look?

WDYT

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: done.

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,257 @@
...
+        .send({ from: l2Depositor });
+
+      // Tokens should not yet be bridgeable as not enough time has passed.
+      assert.isFalse(await bridgeDepositBox.methods.canBridge(l2Token.options.address).call());
+
+      await crossDomainFinalizer.checkForBridgeableL2TokensAndBridge();
+      assert.isTrue(lastSpyLogIncludes(spy, "No bridgeable L2 tokens"));
+
+      // Now, advance time such that the L2->L1 token bridging action should be enabled.
+      console.log("depositTime", depositTime);
```
mrice32, line 244: extra console log

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,257 @@
...
+      await l2Timer.methods.setCurrentTime(Number(depositTime) + minimumBridgingDelay + 1).send({ from: l1Owner });
+
+      // Token should show up as bridgeable.
+      assert.isTrue(await bridgeDepositBox.methods.canBridge(l2Token.options.address).call());
+
+      // Now, when running the cross-domain finalizer, should send the L2->L1 transfer via the bridgeTokens method.
+      console.log("await", (await l2Token.methods.balanceOf(bridgeDepositBox.options.address).call()).toString());
```
mrice32, line 251: console log

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
mrice32: Unrelated to the discussion above: why do we need a toBlock at all?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,95 @@
...
+    const bridgeableL2Tokens = whitelistedL2Tokens.filter((l2Token, index) => canBridge[index]);
+
+    // Finally, iterate over the bridgeable l2Tokens and bridge them.
+    if (bridgeableL2Tokens.length == 0) {
+      this.logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "No bridgeable L2 tokens" });
+      return;
+    }
+    const nonce = await this.l2Client.l2Web3.eth.getTransactionCount(this.account);
+    for (const l2Token of bridgeableL2Tokens) {
+      // Track the account nonce and manually increment on each TX. We need to do this because the L2 transactions
```
mrice32, line 50: sorry, where are we incrementing the nonce? And I thought web3 would track the nonce internally to ensure that it incremented after each send?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,95 @@
...
+      // process quicker than the infura node updates and we need to avoid the nonce collision.
+      try {
+        this._bridgeL2Token(l2Token, nonce);
```
mrice32, line 53: This is an async function. Shouldn't we be awaiting or at least tracking the promises?

mrice32 (commented):
Looks good! A few minor comments

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: should be, my mistake.

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
chrismaree: thanks

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
chrismaree: got it

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: it should...but it does not work correctly on L2 because the transactions process so quickly 🤖

chrismaree: it is incremented in the `bridgeL2Token` method.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: @mrice32 good point! removing this and leaving it just with the `fromBLock` works just fine.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
nicholaspai: oh good point, nevermind

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,118 @@
...
+    for (const l2Token of bridgeableL2Tokens) {
+      // For each bridgeable L2Token, check the balance in the deposit box. If it is greater than
+      // crossDomainFinalizationThreshold, as a percentage, then we can bridge it.
+
+      try {
+        const { symbol, decimals, l2PoolBalance } = await this._getL2TokenInfo(l2Token);
+        const l1PoolReserves = await this._getL1PoolReserves(l2Token);
+
+        if (l2PoolBalance.gt(toBN(this.crossDomainFinalizationThreshold).mul(l1PoolReserves).div(toBN(100))))
+          nonce = await this._bridgeL2Token(l2Token, nonce, symbol, decimals);
```
nicholaspai, line 56: It's kind of weird to return the `nonce`. I assume the `nonce` is getting incremented inside `_bridgeL2Token` which is potentially dangerous since it's not made explicit. I think it might be more clear to increment `nonce++` before calling `await this._bridgeL2Token()` and then passing in the new `nonce`.

This would make this logic more functional and easier to reason about

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -85,22 +88,31 @@ export class RelayerConfig {
...
     } = env;
 
+    assert(BRIDGE_ADMIN_ADDRESS, "BRIDGE_ADMIN_ADDRESS required");
+    this.bridgeAdmin = toChecksumAddress(BRIDGE_ADMIN_ADDRESS);
+
     this.botModes = {
       relayerEnabled: RELAYER_ENABLED ? Boolean(RELAYER_ENABLED) : false,
       disputerEnabled: DISPUTER_ENABLED ? Boolean(DISPUTER_ENABLED) : false,
       finalizerEnabled: FINALIZER_ENABLED ? Boolean(FINALIZER_ENABLED) : false,
+      l2FinalizerEnabled: CROSS_DOMAIN_FINALIZER_ENABLED ? Boolean(CROSS_DOMAIN_FINALIZER_ENABLED) : false,
```
nicholaspai, line 108: See fix [here](https://github.com/UMAprotocol/protocol/pull/3593/files), this is eventually going to have to change to `CROSS_DOMAIN_FINALIZER_ENABLED === "true" ? true : false`

nicholaspai (commented):
Have one comment about making the `_bridgeL2Tokens` loop be more functional programming and easier to reason about

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: this has been improved somewhat.

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -0,0 +1,119 @@
...
+    for (const l2Token of bridgeableL2Tokens) {
+      // For each bridgeable L2Token, check the balance in the deposit box. If it is greater than
+      // crossDomainFinalizationThreshold, as a percentage, then we can bridge it.
+
+      try {
+        const { symbol, decimals, l2PoolBalance } = await this._getL2TokenInfo(l2Token);
+        const l1PoolReserves = await this._getL1PoolReserves(l2Token);
+
+        if (l2PoolBalance.gt(toBN(this.crossDomainFinalizationThreshold).mul(l1PoolReserves).div(toBN(100)))) {
+          await this._bridgeL2Token(l2Token, nonce, symbol, decimals);
```
chrismaree, line 56: post increment nonce

nicholaspai (approved):
LGTM after you finish matt's open comments

packages/insured-bridge-relayer/test/CrossDomainFinalizer.ts
chrismaree: done

PR #3586, opened 2021-11-08
feat(cross-chain-oracle): Generalized cross-chain oracle and governance contracts

nicholaspai:
**Motivation**

These contracts enable the DVM to resolve prices and execute governance transactions from parent chain (i.e. Ethereum) to any number of child chains in a hub-spoke-esque model. 

Reviewers should sanity check that this would work with Polygon, Arbitrum and Optimism as a start.

**Summary**

Follow up PR will add more unit test coverage. For now, unit tests for OracleBase are copied from `/contracts/polygon` directory.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+interface ChildMessengerInterface {
```
chrismaree, line 4: very good. I'll move this around in a follow on PR that implements this interface for the different networks.

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
```diff
@@ -0,0 +1,48 @@
...
+     * @param to Contract on child chain to send message to
+     * @param data Message to send
+     * @dev Only callable by the owner (presumably the Ethereum Governor contract).
+     */
+    function relayGovernance(
+        uint256 chainId,
+        address to,
+        bytes memory data
+    ) external nonReentrant() onlyOwner {
+        messengers[chainId].sendMessageToChild(abi.encode(to, data));
```
chrismaree, line 45: is there any point in checking that the `chainId` contains a `messenger`? I guess this will revert if that does not exist due to calling a function on an address that does not implement that interface?

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -0,0 +1,127 @@
...
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+/**
+ * @title Gatekeeper contract deployed on mainnet that validates and sends price requests from sidechain to the DVM on
+ * mainnet. This is a "gate keeper" contract because it performs the final validation for any messages originating from
+ * a child chain's oracle before submitting price requests to the DVM. This contract also can publish DVM price
+ * resolution data to OracleSpokes on any chainId via the messenger for that chainId.
+ * @dev This contract must be a registered financial contract in order to make and query DVM price requests.
+ */
+contract OracleHub is OracleBase, Ownable, Lockable {
```
chrismaree, line 20: great pattern you've got here.

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -0,0 +1,127 @@
...
+     * can call this method. If the original requester on the child chain wants to expedite the Child --> Parent
+     * message, then they can call `requestPrice` on this contract for the same unique price request.
+     * @param data ABI encoded params with which to call `_requestPrice`.
+     */
+    function processMessageFromChild(uint256 chainid, bytes memory data) public nonReentrant() onlyMessenger(chainid) {
+        (bytes32 identifier, uint256 time, bytes memory ancillaryData) = abi.decode(data, (bytes32, uint256, bytes));
+        bool newPriceRequested = _requestPrice(identifier, time, ancillaryData);
+        if (newPriceRequested) {
+            _getOracle().requestPrice(identifier, time, ancillaryData);
+        }
```
chrismaree, line 90:
```suggestion
        bool newPriceRequested = _requestPrice(identifier, time, ancillaryData);
        if (_requestPrice(identifier, time, ancillaryData))  _getOracle().requestPrice(identifier, time, ancillaryData);
```

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
nicholaspai: Yeah, and only the `owner` can set the address so we can assume the `messenger` is not a malicious contract?

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     * @notice Executes governance transaction created on Ethereum.
+     * @dev Can only called by ChildMessenger contract that wants to execute governance action on this child chain that
+     * originated from DVM voters on root chain. ChildMessenger should only receive communication from ParentMessenger
+     * on mainnet.
+     * @param data ABI encoded params to include in delegated transaction.
+     */
+    function processMessageFromParent(bytes memory data) public nonReentrant() onlyMessenger() {
```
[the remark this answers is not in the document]
nicholaspai: Ah yes, I'll add these to the interfaces

packages/core/contracts/cross-chain-oracle/OracleBase.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    struct Price {
+        RequestState state;
+        int256 price;
+    }
+
+    // This value must be <= the Voting contract's `ancillaryBytesLimit` value otherwise it is possible
+    // that a price can be requested to this contract successfully, but cannot be resolved by the DVM which refuses
+    // to accept a price request made with ancillary data length over a certain size.
+    uint256 public constant ancillaryBytesLimit = 8192;
```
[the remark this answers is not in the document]
nicholaspai: Yeah maybe we should add it to that

packages/core/contracts/cross-chain-oracle/OracleBase.sol
```diff
@@ -0,0 +1,96 @@
...
+
+    /**
+     * @notice Constructor.
+     * @param _finderAddress finder to use to get addresses of DVM contracts.
+     */
+    constructor(address _finderAddress) {
+        finder = FinderInterface(_finderAddress);
+    }
+
+    /**
```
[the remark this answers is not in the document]
nicholaspai: it's internal though

packages/core/contracts/cross-chain-oracle/OracleHub.sol
nicholaspai: Hm I think this is less readable the way you have it. It's not clear what the return value of `_requestPrice` is otherwise

packages/core/contracts/cross-chain-oracle/OracleBase.sol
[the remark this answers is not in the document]
nicholaspai: I added it to Constants do you think there's a problem with this?

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
```diff
@@ -0,0 +1,139 @@
...
+     * @notice This is called to bridge a price request to mainnet. This method will enqueue a new price request
+     * or return silently if already requested. Price requests are relayed to mainnet (the "Parent" chain) via
+     * the ChildMessenger contract.
+     * @dev Can be called only by a registered contract that is allowed to make DVM price requests. Will mark this
+     * price request as Requested, and therefore able to receive the price resolution data from mainnet.
+     * @param identifier Identifier of price request.
+     * @param time Timestamp of price request.
+     * @param ancillaryData extra data of price request.
+     */
+    function requestPrice(
```
mrice32, line 52: Do you see any danger in _not_ making these backwards compatible with the DVM methods that _don't_ have ancillary data?

Put differently, are there any contracts someone might want to deploy that this would break? EMP, Perp?

cc @chrismaree

mrice32 (commented):
Looks awesome! A few comments.

packages/core/contracts/cross-chain-oracle/OracleHub.sol
chrismaree: fair.

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
chrismaree: It's less about it being malicious, more about erroring in a readable way. but I agree with you though that it's somewhat redundant given the calling context.

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
```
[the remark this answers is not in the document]
chrismaree: I think one argument to keeping them separate is the implementation on each chain will be somewhat different for the sending and receiving sides. We could roll this into one MessengerInterface pretty easily but logical separation does help a _bit_ hear. it also makes it very clear what side of the chains you are sitting on.

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
```diff
@@ -0,0 +1,10 @@
...
+interface ChildMessengerInterface {
+    // Should send cross-chain message to Parent messenger contract or revert.
+    function sendMessageToParent(bytes memory data) external;
+
+    // Should be targeted by ParentMessenger and executed upon receiving a message from root chain.
+    function processMessageFromParent(bytes memory data) external;
```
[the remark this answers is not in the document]
chrismaree: this will happen within the implementer of this interface(I'm working on this PR separately). This will work by using the notion of a "cross domain messenger". The implementation for this will be very similar to how the OVM/AVM bridge deposit boxes only accept calls cross chain from their respective L1 owners. This is restricted using a `onlyFromCrossDomainAccount` type modifier.

See here for the cross domain enabled logic and how you can restrict who the caller is from the other chain (i.e how you limit this to only be called by the _parent_: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/external/ovm/OVM_CrossDomainEnabled.sol 

and here for how this is used within a _messenger_. This is similar to what the cross-chain adapters would look like for this set of contracts: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/insured-bridge/ovm/Optimism_Messenger.sol

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
chrismaree: I think at this point we should try move all DVM contracts to using ancillary data interfaces. ideally in the near future (next audit cycle perhaps?) we can deprecate the old interfaces in favour of the ancillary data-only ones.

However, you do raise a good point on how this _could_ break some things. The latest EMP and Perp both have ancillary data support so it won't be an issue with those. it _could_ be an issue with legacy contracts but we dont want to support those anyways.

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
```diff
@@ -0,0 +1,48 @@
...
+        messengers[chainId] = messenger;
+        emit SetParentMessenger(chainId, address(messenger));
+    }
+
+    /**
+     * @notice This should be called in order to relay a governance request to the `GovernorSpoke` contract
+     * deployed to the child chain associated with `chainId`.
+     * @param chainId network that messenger contract will communicate with
+     * @param to Contract on child chain to send message to
+     * @param data Message to send
```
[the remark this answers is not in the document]
nicholaspai: Great idea done.

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -0,0 +1,57 @@
...
+        bool success;
+        assembly {
+            let inputData := add(data, 0x20)
+            let inputDataSize := mload(data)
+            // Hardcode value to be 0 for relayed governance calls in order to avoid addressing complexity of bridging
+            // value cross-chain.
+            success := call(gas(), to, 0, inputData, inputDataSize, 0, 0)
+        }
+        return success;
+    }
```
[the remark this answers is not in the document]
nicholaspai: Sure I can change to this

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
nicholaspai: > this will happen within the implementer of this interface(I'm working on this PR separately). This will work by using the notion of a "cross domain messenger". The implementation for this will be very similar to how the OVM/AVM bridge deposit boxes only accept calls cross chain from their respective L1 owners. This is restricted using a onlyFromCrossDomainAccount type modifier.

Agreed with Chris. In the Hub/Spoke contracts its important to restrict access only to the Messenger contracts. The Messenger contracts then need to make sure they can only communicate with each other. We are assuming that the messenger contracts are trusted, and accordingly only the owner of the Hub/Spoke can set the Messenger contracts

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
nicholaspai: How would we make it backwards compatible--overload the `requestPrice` method?

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -0,0 +1,127 @@
...
+
+    /**
+     * @notice Submits a price request originating from an OracleSpoke. Request data must be sent via the
+     * Child --> Parent Messenger communication channel. Returns silently if price request is a duplicate.
+     * @dev This contract must be registered to submit price requests to the DVM. Only the ParentMessenger
+     * can call this method. If the original requester on the child chain wants to expedite the Child --> Parent
+     * message, then they can call `requestPrice` on this contract for the same unique price request.
+     * @param data ABI encoded params with which to call `_requestPrice`.
+     */
+    function processMessageFromChild(uint256 chainid, bytes memory data) public nonReentrant() onlyMessenger(chainid) {
```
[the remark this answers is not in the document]
chrismaree: thank you sir

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
nicholaspai: I'm indifferent, @chrismaree I'll leave to you as you're implementing a Messenger and have a better idea of what's cleaner

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
mrice32: So just so I understand, there will be a separate messenger for the governor and for the oracle.

For Optimism for instance, we would have an oracle child messenger, a governor child messenger, an oracle parent messenger, and an oracle child messenger?

I think I had been reading it as a single messenger per chain, but it sounds like we're doing this similar to how to polygon implementation works where each contract has exactly one contract on the other side that it can communicate with.

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
mrice32: > The latest EMP and Perp both have ancillary data support

I don't think this is true.

See EMP: https://github.com/UMAprotocol/protocol/blob/2fc8fe4237156e68a02bf15ff9f981eed0598137/packages/core/contracts/financial-templates/expiring-multiparty/PricelessPositionManager.sol#L837

See Perp: https://github.com/UMAprotocol/protocol/blob/2fc8fe4237156e68a02bf15ff9f981eed0598137/packages/core/contracts/financial-templates/perpetual-multiparty/PerpetualPositionManager.sol#L658


So the question is whether we want to support these contracts short term. We could also update them and get that into this audit, but that will cause some real test churn I suspect.

mrice32 (commented):
Just replied to a few threads.

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
nicholaspai: > So just so I understand, there will be a separate messenger for the governor and for the oracle.
I think we'll combine them into one messenger with functions specific to message the Governor/Oracle.

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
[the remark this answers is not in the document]
nicholaspai: Ok I tested this locally with a new unit test and it did not work like the existing implementation. I'm gonna revert the `to.call()` syntax until we figure it out

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
nicholaspai: Well since we're not concerned about bytecode bloat let's overload the `requestPrice` method?

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
chrismaree: The implementation I'm working towards does not have separate methods for the governor and oracle but rather stores the addresses of these contracts to verify the caller has the associated permissions

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
chrismaree: @mrice32 that's a great point and is really irritating...both of those contracts _use_ ancillary data but also use the overloaded methods.

@nicholaspai I think the simplest thing to do is to overload `requestPrice` that just defaults ancillary data to blank, the same way the `Voting.sol` contracts do this to provide backwards compatibility.

packages/core/contracts/cross-chain-oracle/OracleBase.sol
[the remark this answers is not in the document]
chrismaree: nope, I think that works great.

packages/core/contracts/cross-chain-oracle/ChildMessengerInterface.sol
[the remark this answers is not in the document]
chrismaree: it's defos cleaner to have separate implementations as the L1 and L2 sides are _marginally_ difference and so having separate interfaces makes it easier.

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -0,0 +1,127 @@
...
+     * up the price resolution process. Returns silently if price request is a duplicate.
+     * @dev The caller must pay a final fee and have approved this contract to pull final fee from it.
+     * @dev If the price request params including the ancillary data does not match exactly the price request submitted
+     * on the child chain, then the child chain's price request will not resolve. The caller is recommended to use the
+     * `stampAncillaryData` method on the OracleSpoke to reconstruct the ancillary data.
+     * @param identifier Identifier for price request.
+     * @param time time for price request.
+     * @param ancillaryData Extra data for price request.
+     */
+    function requestPrice(
```
chrismaree, line 105: 👍

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
```diff
@@ -0,0 +1,48 @@
...
+    event SetParentMessenger(uint256 indexed chainId, address indexed parentMessenger);
+
+    /**
+     * @notice Set new ParentMessenger contract for chainId.
+     * @param chainId child network that messenger contract will communicate with.
+     * @param messenger ParentMessenger contract that sends messages to ChildMessenger on network with ID `chainId`.
+     * @dev Only callable by the owner (presumably the Ethereum Governor contract).
+     */
+    function setMessenger(uint256 chainId, ParentMessengerInterface messenger) public nonReentrant() onlyOwner {
+        require(address(messenger) != address(0), "Invalid messenger contract");
```
chrismaree, line 27: there could be an argument for setting this to er. this acts somewhat like a "delete" for a particular chain. consider if we wanted to remove a chain from operation, for example.

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -0,0 +1,127 @@
...
+     * The DVM price resolution is communicated to the OracleSpoke via the Parent-->Child messenger channel.
+     * @dev This method will return silently if already called for this price request, but will still attempt to call
+     * `messenger.sendMessageToChild` again even if it is a duplicate call. Therefore the Messenger contract for this
+     * `chainId` should determine how to handle duplicate calls.
+     * @param chainId Network to resolve price for.
+     * @param identifier Identifier of price request to resolve.
+     * @param time Timestamp of price request to resolve.
+     * @param ancillaryData extra data of price request to resolve.
+     */
+    function publishPrice(
```
chrismaree, line 65: in the beginning do you think we'd like to publish a bunch of prices at once to "seed" the L2 state? probs not with any of the existing products but there could be a future situation where we'd like to do that. I think this implementation will work fine for this but it might warrant adding in `multicall` to enable this to be done more easily in a batch! we can leave this for later though, once we've thought about it a bit more.

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
```diff
@@ -0,0 +1,185 @@
...
+import "../oracle/interfaces/OracleAncillaryInterface.sol";
+import "../oracle/interfaces/OracleInterface.sol";
+import "../oracle/interfaces/RegistryInterface.sol";
+import "./OracleBase.sol";
+import "../common/implementation/AncillaryData.sol";
+import "../common/implementation/Lockable.sol";
+import "./ChildMessengerInterface.sol";
+
+/**
+ * @title This contract is primarily intended to receive messages on the child chain from a parent chain and allow
```
chrismaree, line 13: nit: this is non-standard comment. put the bulk of the verbiage within a `* @notice` block. the title should just be a `title`.

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
```diff
@@ -0,0 +1,185 @@
...
+            messenger.sendMessageToParent(abi.encode(identifier, time, _stampAncillaryData(ancillaryData, msg.sender)));
+        }
+    }
+
+    /**
+     * @notice Overloaded function to provide backwards compatibility for legacy financial contracts that do not use
+     * ancillary data.
+     */
+    function requestPrice(bytes32 identifier, uint256 time) public override nonReentrant() onlyRegisteredContract() {
+        bool newPriceRequested = _requestPrice(identifier, time, "0x0");
```
chrismaree, line 69: I think the empty ancillary data should be represented as `""` not `"0x0"`. see how this is done in `Voting.sol` here: https://github.com/UMAprotocol/protocol/blob/e41ab0d60d5598a6af71405db94f4d2e8479a004/packages/core/contracts/oracle/implementation/Voting.sol#L293

chrismaree (approved):
This is awesome! there is a fair amount of follow up stuff that still needs to be added but we can do this in subsequent PRs as we go. 

Great work Nick!

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
nicholaspai: Done I've made `OracleSpoke` extend the `OracleInterface` (in addition to `OracleAncillaryInterface`) and overloaded `requestPrice`, `hasPrice`, `getPrice`

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
nicholaspai: You can also "delete" by setting the address to its own address, so I'll leave in for now.

nicholaspai: Actually I'll just remove this line for now as it doesn't make sense. It's just as easy to have a typo in the address that you pass to `setMessenger` as it is to accidentally call with a zero address. The owner can always reset the address

packages/core/contracts/cross-chain-oracle/OracleHub.sol
nicholaspai: I don't think so at the moment. There are no existing contracts that use past prices so there is no practical need to seed the L2 state.

However I see your point about this being potentially useful in the future. I'll add in a TODO to add a multicall for this reason

PR #3587, opened 2021-11-08
feat(monitors): Add SkinnyOptimisticOracle monitor

nicholaspai:
**Motivation**

Should be backwards compatible and not affect monitor in production.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/monitors/index.js
+++ b/packages/monitors/index.js
@@ -41,6 +41,7 @@ const {
  * @param {Object} logger Module responsible for sending logs.
  * @param {String} financialContractAddress Contract address of the Financial Contract.
  * @param {String} optimisticOracleAddress Contract address of the OptimisticOracle Contract.
+ * @param {OptimisticOracleType} [optimisticOracleType] Type of "OptimisticOracle" for this network, defaults to "OptimisticOracle"
  * @param {Number} pollingDelay The amount of seconds to wait between iterations. If set to 0 then running in serverless
  *     mode which will exit after the loop.
  * @param {Number} errorRetries The number of times the execution loop will re-try before throwing if an error occurs.
@@ -59,6 +60,7 @@ async function run({
   logger,
   web3,
   financialContractAddress,
+  optimisticOracleType,
   optimisticOracleAddress,
   pollingDelay,
   errorRetries,
@@ -80,6 +82,7 @@ async function run({
       message: "Monitor started 🕵️‍♂️",
       financialContractAddress,
       optimisticOracleAddress,
+      optimisticOracleType,
       pollingDelay,
       errorRetries,
       errorRetriesTimeout,
@@ -324,10 +327,10 @@ async function run({
     if (optimisticOracleAddress) {
       const optimisticOracleContractEventClient = new OptimisticOracleEventClient(
         logger,
-        getAbi("OptimisticOracle"),
+        getAbi(optimisticOracleType),
         web3,
         optimisticOracleAddress,
-        OptimisticOracleType.OptimisticOracle,
+        optimisticOracleType,
         eventsFromBlockNumber,
         endingBlock
       );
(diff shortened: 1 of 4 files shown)


packages/monitors/index.js
```diff
@@ -324,10 +327,10 @@ async function run({
     if (optimisticOracleAddress) {
       const optimisticOracleContractEventClient = new OptimisticOracleEventClient(
         logger,
-        getAbi("OptimisticOracle"),
+        getAbi(optimisticOracleType),
```
chrismaree, line 330: would it be useful to validate that this is of an appropriate type? it would be possible to provide a type that breaks the bot here (i.e a string that is not the correct type)

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -85,21 +88,32 @@ class OptimisticOracleContractMonitor {
...
+      );
       const mrkdwn =
         createEtherscanLinkMarkdown(event.requester, this.contractProps.networkId) +
         ` requested a price at the timestamp ${event.timestamp} for the identifier: ${event.identifier}. ` +
         `The ancillary data field is ${event.ancillaryData || "0x"}. ` +
-        `Collateral currency address is ${event.currency}. Reward amount is ${this.formatDecimalString(
-          convertCollateralDecimals(event.reward)
-        )} and the final fee is ${this.formatDecimalString(convertCollateralDecimals(event.finalFee))}. ` +
+        `Collateral currency address is ${
+          this.oracleType === OptimisticOracleType.OptimisticOracle ? event.currency : event.request.currency
```
chrismaree, line 99: o boy this is such a pain in the ass. thanks for sorting out this switching logic!

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -123,20 +138,30 @@ class OptimisticOracleContractMonitor {
...
+        }. ` +
         `tx: ${createEtherscanLinkMarkdown(event.transactionHash, this.contractProps.networkId)}`;
 
       // The default log level should be reduced to "info" for funding rate identifiers:
       this.logger[
         this.logOverrides.proposedPrice || (this._isFundingRateIdentifier(event.identifier) ? "info" : "error")
       ]({
         at: "OptimisticOracleContractMonitor",
+        type: this.oracleType,
```
chrismaree, line 164: I dont think you should do it like this. rather, add the `type` into the markdown. The longer is not designed to deal with more than just a `at`, `message` `mrkdwn` and `notificationPath`, in the event the log produced contains `mrkdwn`.

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -206,6 +265,7 @@ class OptimisticOracleContractMonitor {
         this.logOverrides.settledPrice || (this._isFundingRateIdentifier(event.identifier) ? "debug" : "info")
       ]({
         at: "OptimisticOracleContractMonitor",
+        type: this.oracleType,
```
chrismaree, line 268: same comment as before. Add it into the `mrkdwn`.

chrismaree (approved):
This looks great! thanks for adding in this support. Please see my comments. once those are resolved feel free to merge.

packages/monitors/index.js
nicholaspai: We already do this by making sure that `optimisticOracleType` is of type `OptimisticOracleType`. This is in javascript but if converted to typescript it would validate for free

packages/monitors/src/OptimisticOracleContractMonitor.js
nicholaspai: lol its ugly

packages/monitors/src/OptimisticOracleContractMonitor.js
nicholaspai: ok

nicholaspai: I'm going to add the `message` key for each info or higher log, ok?

nicholaspai: for example:

```
{
        at: "OptimisticOracleContractMonitor",
        message: `${this.oracleType}: Price Settlement Alert 🏧!`,
        mrkdwn,
        notificationPath: "optimistic-oracle",
}
```

PR #3588, opened 2021-11-09
fix(L1-client): optimize how the L1 client fetches some internal data

chrismaree:
**Motivation**

The current L1Client has some inefficiencies in them. This PR address these issues.

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -43,7 +43,7 @@ export interface Relay {
   proposerBond: string;
   finalFee: string;
   settleable: SettleableRelay;
-  blockTime: number;
+  blockNumber: number;
 }
 
 export interface InstantRelay {
@@ -296,7 +296,7 @@ export class InsuredBridgeL1Client {
           proposerBond: depositRelayedEvent.returnValues.relay.proposerBond,
           finalFee: depositRelayedEvent.returnValues.relay.finalFee,
           settleable: SettleableRelay.CannotSettle,
-          blockTime: Number((await this.l1Web3.eth.getBlock(depositRelayedEvent.blockNumber)).timestamp),
+          blockNumber: depositRelayedEvent.blockNumber,
         };
         this.relays[l1Token][relayData.depositHash] = relayData;
       }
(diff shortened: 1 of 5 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -296,7 +296,7 @@ export class InsuredBridgeL1Client {
           proposerBond: depositRelayedEvent.returnValues.relay.proposerBond,
           finalFee: depositRelayedEvent.returnValues.relay.finalFee,
           settleable: SettleableRelay.CannotSettle,
-          blockTime: Number((await this.l1Web3.eth.getBlock(depositRelayedEvent.blockNumber)).timestamp),
```
chrismaree, line 299: This is the offending line that made the L1Clients slow

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
```diff
@@ -516,7 +516,7 @@ describe("InsuredBridgePriceFeed", function () {
 
       // Setting quote time after the latest block time will cause pricefeed to return 0.
       const depositTimestamp = Number(await bridgePool.methods.getCurrentTime().call());
-      const quoteTimestamp = Number((await web3.eth.getBlock("latest")).timestamp) + 60;
```
mrice32, line 519: This seems like we're making an opinionated change. Why remove the 60 second cushion?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -296,7 +296,7 @@ export class InsuredBridgeL1Client {
...
+          blockNumber: depositRelayedEvent.blockNumber,
```
mrice32, line 299: One minor note: I agree that it's probably good to avoid doing node calls for each relay event in the window. However, the slowness (I think) is because we were doing this call in series rather than doing it inside a Promise.all. For instance, we could probably speed the bot up more by putting the outer loop in a Promise.all.

mrice32 (approved):
Looks good!

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
nicholaspai: Yeah why remove?

nicholaspai (approved):
LGTM. Smart to move the web3 call as late as possible

packages/financial-templates-lib/test/price-feed/InsuredBridgePriceFeed.js
nicholaspai: This test will pass as long as the quote time is > the relay block time. So yoh can grab the block time from the relay txn and just increment it. I think it's passing now by luck but it's not clear

PR #3592, opened 2021-11-09
fix(common/core): small changes post optimsim deployment

chrismaree:
**Motivation**

There are a few small changes that came about after testing on optimism earlier this week. This PR implements these.


mrice32 (approved):
LGTM!

PR #3594, opened 2021-11-10
feat: add OptimisticRewarder

mrice32:
**Motivation**

We'd like a way to create NFTs for users that can have custom reward payout structures.

**Summary**

- The OptimisticRewarder system is centered around the OptimisticRewarder contract. It allows users to mint an NFT and be rewarded in any set of ERC20 tokens.
- The contract has all parameters set on construction and has no admin privileges.
- Tokens can be added to the contract by anyone, but they cannot be withdrawn without an NFT. If the user wants an admin that is able to withdraw funds, that can be done by issuing the admin an NFT.
- Rather than changing parameters in the contract, the contract can be migrated easily without changing the NFT using the OptimisticRewarderNoToken.
- To connect a token with data that would influence its reward payouts, it's possible to call the updateToken method with the tokenId and data, which causes the contract to emit an event that can be used off-chain. It's also possible to track rewards in _any other way_ that is verifyable offchain. Rewards do not have to be tracked this way.
- When a user requests redemption, they pass in the _total_ amount of cumulative rewards they think they are owed at that point. This allows us to not have to worry about multiple requests in flight.
- When paying out rewards, the contract only pays out any rewards in the request that exceed the previous payouts.
- Any NFT can be used with this system using the OptimisticRewarderNoToken. Even if it doesn't have the mintNextToken, the redemption system will still work as designed. The NFT will just have to have another method of minting tokens to users. For instance, a uniswap v3 LP token could reasonably be used here.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A


packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+        baseUri = _baseUri;
+    }
+
+    function depositRewards(uint256 amount, IERC20 token) public {
+        token.safeTransferFrom(msg.sender, address(this), amount);
+    }
+
+    function mint(address receiver, bytes memory data) public returns (uint256 tokenId) {
+        tokenId = nextTokenId++;
+        _safeMint(receiver, tokenId);
```
chrismaree, line 37: so no data is stored on chain? you simply emit it? that's pretty clean!

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+    ) ERC721(_name, _symbol) {
+        baseUri = _baseUri;
+    }
+
+    function depositRewards(uint256 amount, IERC20 token) public {
+        token.safeTransferFrom(msg.sender, address(this), amount);
+    }
+
+    function mint(address receiver, bytes memory data) public returns (uint256 tokenId) {
+        tokenId = nextTokenId++;
```
chrismaree, line 36: I do love me a good post increment

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+        _safeMint(receiver, tokenId);
+        emit UpdateToken(tokenId, msg.sender, data);
+    }
+
+    function updateToken(uint256 tokenId, bytes memory data) public {
```
chrismaree, line 41: 🤯

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+        // This automatically checks that redemptions[redemptionId] != 0.
+        // Check that it has not passed liveness liveness.
+        require(block.timestamp < redemptions[redemptionId]);
+
+        // TODO: Make OO request
+
+        redemptions[redemptionId] = 0;
+    }
+
+    function redeem(uint256 tokenId, RedemptionAmount[] memory amounts) public {
```
chrismaree, line 69: damn this is very very clean. very cleaver.

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+        tokenId = nextTokenId++;
+        _safeMint(receiver, tokenId);
+        emit UpdateToken(tokenId, msg.sender, data);
+    }
+
+    function updateToken(uint256 tokenId, bytes memory data) public {
+        emit UpdateToken(tokenId, msg.sender, data);
+    }
+
+    function submitRedemption(uint256 tokenId, RedemptionAmount[] memory amounts) public {
```
chrismaree, line 45: assuringly we'll want to follow a similar pattern to across where we dont actually send the request to the OO and run the escalation game within the contracts, yes?

mrice32: Exacty!

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
```diff
@@ -0,0 +1,91 @@
...
+        // TODO: Make OO request
+
+        redemptions[redemptionId] = 0;
+    }
+
+    function redeem(uint256 tokenId, RedemptionAmount[] memory amounts) public {
+        bytes32 redemptionId = _redemptionId(tokenId, amounts);
+
+        // Can only be redeemed by owner.
+        require(msg.sender == ownerOf(tokenId));
```
mrice32, line 73: @chrismaree @nicholaspai maybe we should allow the msg sender to be the owner or someone who the owner has approved? Thoughts?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
```diff
@@ -0,0 +1,87 @@
...
+     */
+    function withdraw(uint256 tokenId, uint256 amount) public nonReentrant onlyTokenOwner(tokenId) {
+        balances[tokenId] -= amount;
+        stakedToken.safeTransfer(msg.sender, amount);
+        optimisticRewarder.updateToken(tokenId, msg.data);
+    }
+
+    function _depositFor(uint256 tokenId, uint256 amount) public {
+        balances[tokenId] += amount;
+        stakedToken.safeTransferFrom(msg.sender, address(this), amount);
```
chrismaree, line 84: nit: you should do the transfer before incrementing the local variable.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /**
+     * @notice Requests a redemption for any tokenId. This can be called by anyone.
+     * @dev if called by someone who doesn't own the token, they are effectively gifting their bond to the owner.
+     * @param tokenId the tokenId the redemption is for.
+     * @param cumulativeRedemptions the cumulative token addresses and amounts that this tokenId is eligible for
+     * at the current timestap. cumulative redemptions that are too low should be considered to be valid.
+     */
+    function submitRedemption(uint256 tokenId, RedemptionAmount[] memory cumulativeRedemptions) public nonReentrant {
```
chrismaree, line 141: would it be better to call this function `requestRedemption`?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+import "../../oracle/interfaces/SkinnyOptimisticOracleInterface.sol";
+import "../../common/implementation/AncillaryData.sol";
+import "../../common/interfaces/AddressWhitelistInterface.sol";
+import "./OptimisticRewarderToken.sol";
+
+/**
+ * @notice The base rewarder contract. This manages depositing rewards and paying them out to tokenholders using values
+ * backed by the OptimisticOracle's dispute process.
+ */
+abstract contract OptimisticRewarderBase is Lockable {
```
chrismaree, line 23: should this be placed in a separate file?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarderCreator.sol
```diff
@@ -0,0 +1,95 @@
...
+     * @param _name name for the ERC721 token.
+     * @param _symbol symbol for the ERC721 token.
+     * @param _baseUri prefix to each ERC721 tokenId's name.
+     * @param _liveness liveness period between submission and verification of a reward.
+     * @param _bondToken ERC20 token that the bond is paid in.
+     * @param _bond size of the bond.
+     * @param _identifier identifier that should be passed to the optimistic oracle on dispute.
+     * @param _customAncillaryData custom ancillary data that should be sent to the optimistic oracle on dispute.
+     */
+    function createOptimisticRewarder(
```
chrismaree, line 39: very nice that we are not needing to use a struct any more for these params.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    uint256 public bond;
+    IERC20 public bondToken;
+    bytes32 public identifier;
+    bytes public customAncillaryData;
+
+    // Mapping to track redemptions.
+    mapping(bytes32 => Redemption) public redemptions;
+
+    // Mapping to track the past total cumulative redemptions for tokenIds.
+    mapping(uint256 => mapping(IERC20 => uint256)) public redeemedAmounts;
```
chrismaree, line 48: can you explain what the logic is to adding the `redeemedAmounts` to the Base contract rather than adding it to the NFT itself? you could append a prop to each NFT that tracks the cumulative redemption for that token, no?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+     */
+    constructor(
+        uint256 _liveness,
+        IERC20 _bondToken,
+        uint256 _bond,
+        bytes32 _identifier,
+        bytes memory _customAncillaryData,
+        FinderInterface _finder
+    ) {
+        require(liveness > 0, "liveness == 0");
```
chrismaree, line 83:
```suggestion
        require(liveness > 0, "liveness can not be 0");
```

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    {}
+
+    /**
+     * @notice Used to mint the next ERC721 tokenId.
+     * @param recipient the recipient of the newly minted token.
+     */
+    function mintNextToken(address recipient)
+        public
+        virtual
+        override(OptimisticRewarderBase, OptimisticRewarderToken)
```
chrismaree, line 362: nice

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    }
+}
+
+/**
+ * @notice The optimistic rewarder that does not contain the ERC721 token. It allows the user to pass in an external
+ * ERC721 token.
+ * @dev this setup allows for graceful migrations to new rewarder contracts. It also allows external ERC721 tokens,
+ * like uniswap v3 positions to be rewarded.
+ */
+contract OptimisticRewarderNoToken is OptimisticRewarderBase {
```
chrismaree, line 385: this is super cool. can open up a wide use cases for this kind of reward contract.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+     * @param _token external ERC721 token for the rewarder to base redemptions on.
+     * @param _liveness liveness period between submission and verification of a reward.
+     * @param _bondToken ERC20 token that the bond is paid in.
+     * @param _bond size of the bond.
+     * @param _identifier identifier that should be passed to the optimistic oracle on dispute.
+     * @param _customAncillaryData custom ancillary data that should be sent to the optimistic oracle on dispute.
+     * @param _finder finder to look up UMA contract addresses.
+     */
+    constructor(
+        OptimisticRewarderToken _token,
```
chrismaree, line 399: do all NFTs implement the `mintNextToken` interface? how does `mintNextToken` work if this is not implemented? on a uniswap position for example, this is not added, no?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    function mintNextToken(address recipient) public virtual returns (uint256);
+
+    /**
+     * @notice Abstract function that is called to check the owner of the token.
+     * @dev this matches the ERC721 ownerOf interface.
+     * @param tokenId the tokenId to check the owner of.
+     */
+    function ownerOf(uint256 tokenId) public view virtual returns (address);
+
+    function _redemptionId(uint256 tokenId, RedemptionAmount[] memory amounts) internal pure returns (bytes32) {
```
chrismaree, line 294: any downside to making this public? or make a nonReentrant view function that lets you call this? will just make things a bit easier to re-compute off-chain to have ea view function you can easily call for this.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+     * at the current timestap. cumulative redemptions that are too low should be considered to be valid.
+     */
+    function submitRedemption(uint256 tokenId, RedemptionAmount[] memory cumulativeRedemptions) public nonReentrant {
+        bytes32 redemptionId = _redemptionId(tokenId, cumulativeRedemptions);
+        require(redemptions[redemptionId].expiryTime == 0, "Redemption already exists");
+        require(ownerOf(tokenId) != address(0), "tokenId is invalid");
+
+        uint256 time = getCurrentTime();
+
+        uint256 finalFee = _getStore().computeFinalFee(address(bondToken)).rawValue;
```
chrismaree, line 148: would it be useful to store the final fee in this contract so we dont need to make external calls to the `store` on every `request`? this would be similar to how the BridgePool stores fee information.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+
+    struct Redemption {
+        uint256 finalFee;
+        uint256 expiryTime;
+    }
+
+    // Constants.
+    FinderInterface public finder;
+    uint256 public liveness;
+    uint256 public bond;
```
chrismaree, line 39: there seems to be no way to change this bond. is this a problem?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+     * @param cumulativeRedemptions the cumulative redemptions that were provided in the original request.
+     */
+    function dispute(uint256 tokenId, RedemptionAmount[] memory cumulativeRedemptions) public nonReentrant {
+        bytes32 redemptionId = _redemptionId(tokenId, cumulativeRedemptions);
+
+        // This automatically checks that redemptions[redemptionId] != 0.
+        // Check that it has not passed liveness liveness.
+        Redemption storage redemption = redemptions[redemptionId];
+        require(getCurrentTime() < redemption.expiryTime, "redemption expired or nonexistent");
+        if (redemption.finalFee != _getStore().computeFinalFee(address(bondToken)).rawValue) {
```
chrismaree, line 170: how would the final fee have changed?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+            SkinnyOptimisticOracleInterface(finder.getImplementationAddress(OracleInterfaces.SkinnyOptimisticOracle));
+    }
+
+    function _getIdentifierWhitelist() internal view returns (IdentifierWhitelistInterface) {
+        return IdentifierWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.IdentifierWhitelist));
+    }
+
+    function _getCollateralWhitelist() internal view returns (AddressWhitelistInterface) {
+        return AddressWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.CollateralWhitelist));
+    }
```
chrismaree, line 313: these methods introduce a fair amount of gas overhead. do we want to consider the cache and sync pattern that the across contracts use?

gas might be less of a consideration here but it will still be pretty meaningful!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+            SkinnyOptimisticOracleInterface optimisticOracle = _getOptimisticOracle();
+            uint256 totalBond = bond + redemption.finalFee;
+            bondToken.safeApprove(address(optimisticOracle), totalBond * 2);
+            bytes memory ancillaryData =
+                AncillaryData.appendKeyValueBytes32(customAncillaryData, "redemptionId", redemptionId);
+            uint32 requestTimestamp = uint32(redemption.expiryTime - liveness);
+            address proposer = ownerOf(tokenId);
+
+            try
+                optimisticOracle.requestAndProposePriceFor(
```
chrismaree, line 183: The commenting on the params within this is a bit janky. consider putting the comments on the same line as the variable.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+                    0,
+                    // Set the proposer and disputer bond for the price request.
+                    bond,
+                    // Set the liveness for the price request.
+                    liveness,
+                    proposer,
+                    // Canonical value representing "True"; i.e. the proposed redemption is valid.
+                    int256(1e18)
+                )
+            returns (uint256) {} catch {
```
chrismaree, line 198: in what case would the catch be hit? if the price request had previously been requested & proposed for for one but are there other cases? we should comment these as this is a bit complex.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+        delete redemptions[redemptionId];
+    }
+
+    /**
+     * @notice Redeem a redemption request that has passed liveness.
+     * @dev returns the bond that was paid with the initial proposal.
+     * @param tokenId the tokenId the redemption is for.
+     * @param cumulativeRedemptions the cumulative redemptions that were provided in the original request.
+     */
+    function redeem(uint256 tokenId, RedemptionAmount[] memory cumulativeRedemptions) public nonReentrant {
```
chrismaree, line 241: this is pretty awesome.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Good point. Will do

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
nicholaspai: Can use a similar `sync` method

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /**
+     * @notice Allows the caller to mint a token to the receiver and include a reward-relevant update event with it.
+     * This is intended to be used when the user first interacts with a reward-granting protocol.
+     * @dev if the user prefers to only mint a new token, they should call the mintNextToken function.
+     * @param receiver user that will receive the newly minted token.
+     * @param data arbitrary caller-generated data that will be associated with this update.
+     */
+    function mint(address receiver, bytes memory data) public nonReentrant returns (uint256 tokenId) {
```
nicholaspai, line 120: Do we need to do any sanitizing of the `data`? I'm curious what an example case of the data would be

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
```diff
@@ -0,0 +1,429 @@
...
+    bytes public customAncillaryData;
+
+    // Mapping to track redemptions.
+    mapping(bytes32 => Redemption) public redemptions;
+
+    // Mapping to track the past total cumulative redemptions for tokenIds.
+    mapping(uint256 => mapping(IERC20 => uint256)) public redeemedAmounts;
+
+    // This allows other contracts to publish reward updates.
+    event UpdateToken(uint256 indexed tokenId, address indexed caller, bytes data);
```
nicholaspai, line 51: So are Redemption attempts (i.e. Submissions) linked to `UpdateToken` events? And are disputes essentially checking whether the history of `UpdateToken` events produces the proposed `cumulativeRedemptions` amounts?

nicholaspai (commented):
Really well commented thank you! I left some nits but I'm still trying to wrap my head around possible use case in production so I'll leave comments for follow up review. Is my understanding correct here:

> So are Redemption attempts (i.e. Submissions) linked to UpdateToken events? And are disputes essentially checking whether the history of UpdateToken events produces the proposed cumulativeRedemptions amounts?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
```diff
@@ -0,0 +1,87 @@
...
+
+import "../../common/implementation/Lockable.sol";
+import "./OptimisticRewarder.sol";
+import "./OptimisticRewarderToken.sol";
+
+/**
+ * @notice An example use case of the OptimisticRewarder in use by a contract that allows users to stake an ERC20 to
+ * earn rewards.
+ */
+contract OptimisticMintr is Lockable {
```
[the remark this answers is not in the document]
mrice32: Oh, it's named this way after the heavily-used synthetix mintr contract: see https://github.com/Synthetixio/synthetix-mintr.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: So I did it this way to avoid having an owner. We can institute an owner that's assumed to be the deployer if we prefer -- it just introduces a trust assumption on the funds already in the contract. If you can change these parameters, you can effectively pull all the money out.

The way I saw this working was that these parameters would change very infrequently, but when they did, you could migrate this contract to a new contract by passing the NFT into a new OptimisticRewarderNoToken contract and beginning to pay out rewards there.

Thoughts?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: There's a subtlety here. If this interface isn't implemented, the call will fail. The effect of this is that all minting functions don't work. However, everything else (including redemptions) will continue to work.

Minting is a non-essential part of the contract, so if you pass in a token that doesn't have that interface/capability, everything should work fine using updateToken to provide any updates you need. Callers, in that setup, should just avoid using that method.

It's admittedly a little hacky, but it is nice to have the option to mint tokens, but allow for use cases where tokens are not mintable or are mintable through custom interfaces.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: If we depended on this interface or data existing on the NFT, that would mean that other NFTs that were not produced by us and don't implement our custom interfaces couldn't be used in our redemption system. That was the primary reason.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Sure, was trying to keep the changes _looking_ small. I can do that :)

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Agree. Will make this change.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
mrice32: Yeah, I was just following the Check-Effects-Interaction pattern.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
```diff
@@ -0,0 +1,87 @@
...
+     * @return tokenId the token id for the freshly minted token.
+     */
+    function depositNew(uint256 amount) public nonReentrant returns (uint256 tokenId) {
+        tokenId = optimisticRewarder.mint(msg.sender, msg.data);
+        _depositFor(tokenId, amount);
+    }
+
+    /**
+     * @notice Deposit the staked token into the contract.
+     * @param tokenId the tokenId that will own this liquidity. User must be the owner of this tokenId.
```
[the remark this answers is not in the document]
mrice32: depositNew mints you a new NFT to manage your position. for deposit, you pass in a tokenId that you already control to manage the liquidity.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Right now, I've primarily been using `msg.data` which is just the abi encoded function that calls this one.

How would you suggest sanitizing it? Like turning it to utf8 or something?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: This is where the contract is really flexible, but admittedly a little confusing. 

UpdateToken events are an _easy way_ to track interactions that affect the token's rewards. However, the identifier could be specified s.t. it relies on data that is logged completely outside of this contract.

I wanted a simple way to log _something_, but it seems overly restrictive to _only allow_ UpdateToken events to affect the redemption value.

Thoughts?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: For the same reason as the relayer, we can't do this because the OptimisticOracle uses the current final fee at time of request. Unless we do an OO request at the start of each redemption (and not just on disputes) I think we have to handle the case where the final fee changes. Do you agree with this analysis?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Makes sense. We can definitely do this.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Will do.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Sounds good. I will add comments here.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
nicholaspai: I think just name it whatever you think most accurately describes it. It feels like more of a staker tbh

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
nicholaspai: So you'd add more collateral to your position, as represented by the NFT?

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
nicholaspai: If the `data` is emitted by `UpdateTokens`, which is potentially used to determine how many rewards the NFT can be redeemed for, isn't it weird that the caller can set their own `data`? Perhaps i'm confusing its intended use case

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
nicholaspai: > I wanted a simple way to log something, but it seems overly restrictive to only allow UpdateToken events to affect the redemption value.

I agree its overly restrictive to only rely on UpdateToken events. I guess I'm not fully understanding how this would work in practice

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
mrice32: The NFT is just the representation of the account. If you already have an NFT you can pass it in. If not, you can ask the contract to create you a new account. Does that make sense? It works kinda like Uni v3 positions.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
chrismaree: @nick you right. I was confused for a hot second there. you should always leave external calls for as late as posible inside a function.

However, this does _not_ have reentrency guards on it. if the intention was for this to only be called in the depositNew and deposit functions then make this internal.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: I see the confusion. This is why we log the caller. You can imagine an identifier that says to ignore all events where the caller isn't X. The idea is that there is a lot of flexibility for the users to tell this contract to emit any sort of data from this event from any sort of caller. Then offchain, the identifier can specify which callers are valid and how to decode the data they pass in.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
chrismaree: o I see that's pretty awesome. in that case please add additional comments to this effect. this is pretty subtle!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
chrismaree: ye, that makes sence! perhaps add a comment to that effect on why it does not have a modifier.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: I'm not 100% sure if we should even have these events or just have the user to use their own contract to associate events with the tokenIds. But it seemed fairly simple to add without any real downside.

The one use case that this could enable is a privileged EOA (potentially specified in the contract's customAncillaryData) could call this method to add rewards to individual tokens without needing to use or create a contract at all.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
chrismaree: I see. well this good thing is this is 
 to happen often. the bad thing is this is expensive!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
mrice32: @chrismaree that's a great catch. This function _is_ supposed to be internal, that's a bug. Thanks for the call out!!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarderCreator.sol
mrice32: Agreed!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Will do!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Will add!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Agreed. I don't think there is much we can do about this, though. We did the exact same thing in the bridgepool, albeit I tried to make the code a little simpler here.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Note: some of these are only used in the constructor, so those don't need to be cached.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
mrice32: Changed it to OptimisticStaker.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
mrice32: Fixed!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Fixed.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: Yep!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
mrice32: Done.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    modifier onlyTokenOwner(uint256 tokenId) {
+        require(optimisticRewarder.ownerOf(tokenId) == msg.sender, "caller != token owner");
+        _;
+    }
+
+    /**
+     * @notice Deposit the staked token into the contract and mint a fresh token to manage the position.
+     * @param tokenId the tokenId that will own this liquidity. User must be the owner of this tokenId.
```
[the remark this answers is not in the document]
mrice32: Fixed in my commit.

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
mrice32: The update token or the events that occur outside the contract? I can give an example of either.

packages/core/contracts/financial-templates/rewarder/Rewarder.sol
nicholaspai: Agreed,  owner might want to delegate redemption to someone else

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticMinter.sol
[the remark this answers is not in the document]
nicholaspai: Aha that makes sense. I think this comment "The NFT is just the representation of the account. If you already have an NFT you can pass it in. If not, you can ask the contract to create you a new account." is really helpful to understand and can be placed somewhere in the contract to aid reviewers!

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
nicholaspai: This makes sense

nicholaspai (approved):
I'm going to approve so as not to block you before Monday FYI

packages/core/contracts/financial-templates/optimistic-rewarder/OptimisticRewarder.sol
hlambur: Agreed

packages/core/test/financial-templates/optimistic-rewarder/OptimisticRewarder.js
```diff
@@ -0,0 +1,509 @@
...
+          .redeem(tokenId, [{ token: redemptionToken.options.address, amount: toWei("50") }])
+          .send({ from: submitter })
+      )
+    );
+
+    assert.equal(await redemptionToken.methods.balanceOf(submitter).call(), toWei("100"));
+    assert.equal(await bondToken.methods.balanceOf(submitter).call(), totalBond);
+  });
+
+  it("Redemption checks", async function () {
```
chrismaree, line 247: nice. these helped me follow the intended usecase more closely.

packages/core/test/financial-templates/optimistic-rewarder/OptimisticRewarder.js
```diff
@@ -0,0 +1,509 @@
...
+    await collateralWhitelist.methods.removeFromWhitelist(bondToken.options.address).send({ from: owner });
+    receipt = await optimisticRewarder.methods.dispute(tokenId, redemptions).send({ from: disputer });
+    await assertEventEmitted(receipt, optimisticRewarder, "Canceled");
+
+    // Balances should match pre-request balances.
+    assert.equal(await bondToken.methods.balanceOf(submitter).call(), totalBond);
+    assert.equal(await bondToken.methods.balanceOf(disputer).call(), totalBond);
+
+    // Cleanup
+    await collateralWhitelist.methods.addToWhitelist(bondToken.options.address).send({ from: owner });
```
chrismaree, line 475: did you mean to remove this? and the other Cleanup comments on other lines?

mrice32: Nope, the cleanup comment is meant to denote that it's cleaning up the state changes that might interfere with other tests. I can add a comment to that effect or just change the wording. Thoughts?

chrismaree (approved):

PR #3595, opened 2021-11-11
feat(cross-chain-oracle): add optimism chain adapters

chrismaree:
**Motivation**

The cross-chain oracle requires adapters to work on each chain we deploy on. This PR adds the cross-chain adapter support for Optimism. Interfaces mocks and unit tests are added for both the Parent and Child sides of the adapter.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -1,20 +1,21 @@
...
+import "./interfaces/ChildMessengerConsumerInterface.sol";
 import "../common/implementation/Lockable.sol";
 
 /**
- * @notice Governor contract deployed on sidechain that receives governance actions from Ethereum.
+ * @title Cross-chain Oracle L2 Governor Spoke.
+ * @notice Governor contract deployed on L2 that receives governance actions from Ethereum.
  */
-contract GovernorSpoke is Lockable {
+contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
```
chrismaree, line 11: This PR introduces two new interfaces `ChildMessengerConsumerInterface` and `ParrentMessengerConsumerInterface`. These go in the L2 an L1 contracts, respectively, where the contracts "consume" the data from the messenger. For example, in this `GovernorSpoke` we are "consuming" data from the messenger in the `processMessageFromParent` function. This lets us `override` this function here. this is required within the `messenger` contracts to let us define the interface that is called within the consumer contract.

packages/core/contracts/cross-chain-oracle/test/GovernorHubMock.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../interfaces/ParentMessengerInterface.sol";
+
+contract GovernorHubMock {
```
nicholaspai, line 6: Why do you need this mock? It seems like you can just use the GovernorHub but with one small change to setting messenger per chain ID instead of 1 messenger for all chain IDs

nicholaspai (commented):
Why do you need the OracleHubMock, OracleSpokeMock, or GovernorHubMock? They don't seem that different from the non-mocked contracts and I think this PR would be a bit easier to read if there were fewer contracts

packages/core/contracts/cross-chain-oracle/chain-adapters/ParentMessengerBase.sol
```diff
@@ -0,0 +1,80 @@
...
+
+    address public oracleHub;
+    address public governorHub;
+
+    address public oracleSpoke;
+    address public governorSpoke;
+
+    uint32 public defaultGasLimit = 5_000_000;
+
+    modifier onlyPrivilegedCaller() {
```
nicholaspai, line 19: Should we rename this modifier to indicate that its only a Hub contract that can call?

packages/core/contracts/cross-chain-oracle/chain-adapters/ParentMessengerBase.sol
```diff
@@ -0,0 +1,80 @@
...
+        require(msg.sender == oracleHub || msg.sender == governorHub, "Only privileged caller");
+        _;
+    }
+
+    /**
+     * @notice Construct the ParentMessengerBase contract.
+     * @param _chainId The chain id of the L2 network this messenger should connect to.
+     **/
+    constructor(uint256 _chainId) {
+        chainId = _chainId;
```
nicholaspai, line 29: Why not read from block.chainId like `OracleSpoke._stampAncillaryData` does?

nicholaspai: Oh I see. I think we should rename this to `childChainId` to be more clear.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,92 @@
...
+     * @dev The caller must be the OracleSpoke on L2. No other contract is permissioned to call this function.
+     * @dev The L1 target, the parent messenger, must implement processMessageFromChild to consume the message.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override {
+        require(msg.sender == oracleSpoke, "Only callable by oracleSpoke");
+        sendCrossDomainMessage(
+            parentMessenger,
+            defaultGasLimit,
+            abi.encodeWithSignature("processMessageFromChild(bytes)", data)
```
nicholaspai, line 73: Thoughts on changing this to `abi.encode(block.chainId, abi.encodeWithSignature("processMessageFromChild(bytes)", data))`?

Then the parent can `abi.decode(data, (uint256, bytes))` and call `oracleHub.processMessageFromChild(chainId, data)`.

This way the OracleHub doesn't hardcode the child messenger's chain ID, but simply passes on whatever chain ID the child messenger detects from its environment?

Or, the ParentMessenger can `require(decodedChainId == chainId)` to enforce one way communication between ParentMessenger and a ChildMessenger for the `chainId` network.

nicholaspai (commented):
Ok left some comments on how Child/Parent detect the child chain ID

packages/core/contracts/cross-chain-oracle/test/GovernorHubMock.sol
chrismaree: I added this because the actual GovernorHub does not contain all the event data that I needed. I can however refactor the tests to work without this and delete the mock.

packages/core/contracts/cross-chain-oracle/chain-adapters/ParentMessengerBase.sol
chrismaree: sure. I changed it to `onlyHubContract`

packages/core/contracts/cross-chain-oracle/chain-adapters/ParentMessengerBase.sol
chrismaree: sure. done

packages/core/contracts/cross-chain-oracle/test/GovernorHubMock.sol
chrismaree: I ended up removing all three mocks. it made the tests a bit more complex but it uses the actual contracts that would be used in production.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
chrismaree: I think validating this is somewhat redundant as the hardcoding in `onlyFromCrossDomainAccount` enforces who the `childMessenger` is;  it is true that there could be multiple chains with the same ID on this but this is simply restricted via the `_crossDomainMessenger`, which is configured to only work between two chains (eg ethereum and optimism).

Regarding changing the data structure so the oracleHub does not need to store the Id I also think makes things more complex than they should be; it is important that you can look at the ParentMessanger and see which chain this is pointing to by reading the state variable `childChainId`.

Regarding the extra encoding, I think this makes this really hard to reason about without much gain.

Overall, I think it best to _not_ make this change for the reasons listed above ^^

packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -81,9 +84,15 @@ contract OracleHub is OracleBase, Ownable, Lockable {
      * @dev This contract must be registered to submit price requests to the DVM. Only the ParentMessenger
      * can call this method. If the original requester on the child chain wants to expedite the Child --> Parent
      * message, then they can call `requestPrice` on this contract for the same unique price request.
+     * @param ChainId ChainId of the child chain that sent the price request.
      * @param data ABI encoded params with which to call `_requestPrice`.
      */
-    function processMessageFromChild(uint256 chainid, bytes memory data) public nonReentrant() onlyMessenger(chainid) {
+    function processMessageFromChild(uint256 ChainId, bytes memory data)
```
nicholaspai, line 90:
```suggestion
    function processMessageFromChild(uint256 chainId, bytes memory data)
```

nicholaspai: And then fix in rest of function.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,92 @@
...
+    constructor(address _parentMessenger) OVM_CrossDomainEnabled(L2_CROSS_DOMAIN_MESSENGER) {
+        parentMessenger = _parentMessenger;
+    }
+
+    /**
+     * @notice Changes the stored address of the Oracle spoke, deployed on L2.
+     * @dev The caller of this function must be the parent messenger, over the canonical bridge.
+     * @param newOracleSpoke address of the new oracle spoke, deployed on L2.
+     */
+    function setOracleSpoke(address newOracleSpoke) public onlyFromCrossDomainAccount(parentMessenger) {
```
nicholaspai, line 40: I think we should add en event for each of these setter methods, to be consistent with how other contracts work. Look at `BridgeAdmin` for example which has a different event for each setter.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -0,0 +1,62 @@
...
+        ParentMessengerBase(_childChainId)
+    {}
+
+    /**
+     * @notice Changes the default gas limit that is sent along with transactions to Optimism.
+     * @dev The caller of this function must be the owner. This should be set to the DVM governor.
+     * @param newDefaultGasLimit the new L2 gas limit to be set.
+     */
+    function setDefaultGasLimit(uint32 newDefaultGasLimit) public onlyOwner {
+        defaultGasLimit = newDefaultGasLimit;
```
nicholaspai, line 31: We should emit an event here.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,92 @@
...
+        defaultGasLimit = newDefaultGasLimit;
+    }
+
+    /**
+     * @notice Sends a message to the parent messenger via the canonical message bridge.
+     * @dev The caller must be the OracleSpoke on L2. No other contract is permissioned to call this function.
+     * @dev The L1 target, the parent messenger, must implement processMessageFromChild to consume the message.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override {
```
nicholaspai, line 68: We should emit an event here since this is an important state changing function that interacts with an external contract

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,92 @@
...
+     * @param target desired recipient of `data`. Target must implement the `processMessageFromParent` function. Having
+     * this as a param enables the L1 Messenger to send messages to arbitrary addresses on the L1. This is primarily
+     * used to send messages to the OracleSpoke and GovernorSpoke on L2.
+     */
+    function processMessageFromParent(bytes memory data, address target)
+        public
+        override
+        onlyFromCrossDomainAccount(parentMessenger)
+    {
+        ChildMessengerConsumerInterface(target).processMessageFromParent(data);
```
nicholaspai, line 90: I would also want to emit an event here

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,100 @@
...
+    // Messenger contract on the other side of the L1<->L2 bridge.
+    address public parentMessenger;
+
+    // Hard coded default gas limit for L1 transactions.
+    uint32 public defaultGasLimit = 5_000_000;
+
+    // TODO: import from optimism contracts when they release their latest version.
+    address internal constant L2_CROSS_DOMAIN_MESSENGER = 0x4200000000000000000000000000000000000007;
+
+    event OracleSpokeSet(address newOracleSpoke);
```
nicholaspai, line 27: nit: you're going to hate me, but usually we've been naming events as VerbNoun. For example these events should probably be renamed:

```
SetOracleSpoke
SetParentMessenger
SetDefaultGasLimit
```

On the MessageSent and MessagReceived events I'm good with leaving those as-is because they both conveniently start with "Message"

chrismaree: ok.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -0,0 +1,67 @@
...
+import "../interfaces/ParentMessengerInterface.sol";
+import "../interfaces/ParentMessengerConsumerInterface.sol";
+import "./ParentMessengerBase.sol";
+
+/**
+ * @notice Sends cross chain messages from Ethereum L1 to Optimism L2 network.
+ * @dev This contract's is ownable and should be owned by the DVM governor.
+ */
+contract Optimism_ParentMessenger is OVM_CrossDomainEnabled, ParentMessengerInterface, ParentMessengerBase {
+    event SetdefaultGasLimit(uint32 newDefaultGasLimit);
```
nicholaspai, line 15:
```suggestion
    event SetDefaultGasLimit(uint32 newDefaultGasLimit);
```

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
nicholaspai: > but this is simply restricted via the _crossDomainMessenger, which is configured to only work between two chains (eg ethereum and optimism).

This is a good point, can ignore

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,100 @@
...
+pragma solidity ^0.8.0;
+
+// This should be replaced with a "real" import when Optimism release their new contract versions.
+import "../../external/ovm/OVM_CrossDomainEnabled.sol";
+import "../interfaces/ChildMessengerInterface.sol";
+import "../interfaces/ChildMessengerConsumerInterface.sol";
+
+/**
+ * @notice Sends cross chain messages from Optimism L2 to Ethereum L1 network.
+ * @dev This contract's is ownable via the onlyCrossDomainAccount modifier, restricting ownership to the cross-domain
```
nicholaspai, line 11:
```suggestion
 * @dev This contract is ownable via the onlyCrossDomainAccount modifier, restricting ownership to the cross-domain
```

nicholaspai (approved):
LGTM

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -0,0 +1,100 @@
...
+
+    /**
+     * @notice Process a received message from the parent messenger via the canonical message bridge.
+     * @dev The caller must be the the parent messenger, sent over the canonical message bridge.
+     * @param data data message sent from the L1 messenger. Should be an encoded function call or packed data.
+     * @param target desired recipient of `data`. Target must implement the `processMessageFromParent` function. Having
+     * this as a param enables the L1 Messenger to send messages to arbitrary addresses on the L1. This is primarily
+     * used to send messages to the OracleSpoke and GovernorSpoke on L2.
+     */
+    function processMessageFromParent(bytes memory data, address target)
```
mrice32, line 92: Similar comment here as in @nicholaspai's previous PR -- can we call this method something different than the method it calls?

I think it's confusing to have both the ChildMessengerInterface and the ChildMessengerConsumerInterface have the same method name.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -0,0 +1,67 @@
...
+     * @notice Sends a message to the child messenger via the canonical message bridge.
+     * @dev The caller must be the either the OracleHub or the GovernorHub, deployed on L2. This is to send either a
+     * price or initiate a governance action on L2.
+     * @dev The recipient of this message is the child messenger. The messenger must implement processMessageFromParent
+     * which then forwards the data to the target either the OracleSpoke or the governorSpoke depending on the caller.
+     * @param data data message sent to the L2 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToChild(bytes memory data) public override onlyHubContract() {
+        address target = msg.sender == oracleHub ? oracleSpoke : governorSpoke;
+        bytes memory dataSentToChild = abi.encodeWithSignature("processMessageFromParent(bytes,address)", data, target);
```
mrice32, line 51: As discussed offline, I think the chains of contracts that these messages have to pass through:
1. Makes them a bit more error prone as there is more configuration surface area.
2. Makes them more confusing to trace/debug.
3. Creates more opportunities for spoofing where a permissioning code error allows another contract to pretend to be one of our mainnet contracts.

Have we considered shortening the chain by putting these in the _inheritance tree_ for the hub and spoke governor/oracle contracts?

This might look something like:
```
contract OptimismGovernorSpoke is GovernorSpoke, OptimismChildMessenger {
...
}
```

Then, consuming these messages would just be calling internal methods, and we would only need to make sure that the caller on the other side was the GovernorHub.

This would introduce a problem that on the Hub side, you would need to deploy a separate Oracle and Governor hub for each chain rather than just storing a mapping to the messengers.

Not strongly opinionated as I think there isn't a perfect solution, but I just wanted to bring it up since the long chains of contracts makes me a bit nervous because of the overall surface area and configuration complexity.

Thoughts?

mrice32 (commented):
This looks great! One comment on the overall architecture

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
nicholaspai: > This would introduce a problem that on the Hub side, you would need to deploy a separate Oracle and Governor hub for each chain rather than just storing a mapping to the messengers.

Yeah this is my main push back. I think this proposed architecture works better for the Governor Hub/Spoke which doesn't have to store additional price request information, but deploying a separate OracleHub for each L2 is not a great solution. Thoughts?

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
chrismaree: yup, can do. I changed the functions that call cross-chain to `processMessageFromCrossChainChild` and `processMessageFromCrossChainParent` and left the ones that operate on the "same chain" the same.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
chrismaree: Agreed with @nicholaspai . the whole point of this design is that we have one OracleHub/GovernerHub for n number of chains. I agree that in the simple 1:1 case (only L1) the pattern here is over-engineered and complex but if we onboard 5 chains do you think having it as part of inheritance will be easier to reason about and manage? I'm not too sure.

PR #3597, opened 2021-11-11
fix(sdk): add pool utilization

amateima:
**Motivation**

https://app.shortcut.com/uma-project/story/3190/add-pool-utilization-to-the-sdk

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -83,13 +84,17 @@ class PoolState {
...
     // typechain does not have complete types for call options, so we have to cast blockTag to any
-    const exchangeRatePrevious = await this.contract.callStatic.exchangeRateCurrent({
-      blockTag: previousBlock || latestBlock - 1,
-    } as any);
+    const [exchangeRatePrevious, liquidityUtilizationCurrent] = await Promise.all([
+      this.contract.callStatic.exchangeRateCurrent({
+        blockTag: previousBlock || latestBlock - 1,
+      } as any),
+      this.contract.callStatic.liquidityUtilizationCurrent(),
+    ]);
```
amateima, line 92: My question is which clients perform these calls because they require gas to be paid

mrice32: What do you mean?

It's not read-only, but that's just a quirk of how the contract works. I think your callStatic accounts for this, right?

mrice32 (approved):
LGTM!

packages/sdk/src/across/clients/bridgePool.ts
amateima: Yep, you're right

daywiss (approved):

PR #3598, opened 2021-11-11
fix: minor fix to get web3 to respect user-provided network

mrice32:
**Motivation**

When we generate keys in getWeb3 we don't respect user-provided network names.

**Summary**

This just adds a parameter to support this.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A


nicholaspai (commented):
What's an example use case of this new feature?

mrice32:
> What's an example use case of this new feature?

@nicholaspai it isn't a new feature. We already allow users to pass in a network. If they do that in the current code, some of the functions respect that provided network, but others use `argv.network`, which is a bug.

nicholaspai (approved):
LGTM

PR #3602, opened 2021-11-12
Price feed implementation for uSPAC5 - price identifier for index of 5 most active SPACs

dmitry-vsl:
Price feed implementation for uSPAC5 - price identifier for index of 5 most active SPACs

See UMIP: https://github.com/UMAprotocol/UMIPs/pull/428

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
@@ -1012,6 +1012,12 @@ export const defaultConfigs: { [name: string]: { type: string; [key: string]: an
     type: "expression",
     expression: "1 / CVXUSD",
   },
+  uSPAC10: {
+    type: "uSPAC",
+    lookback: 7200,
+    correctionFactor: "1.0",
+    symbols: ["DWAC", "IRDM", "PRIM", "TGLS", "MP", "LCID", "GDYN", "SMPL", "ENVX", "QS"],
+  },
 };
 
 // Pull in the number of decimals for each identifier from the common getPrecisionForIdentifier. This is used within the
--- a/packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
@@ -1,0 +1,243 @@
+import assert = require("assert");
+import lodash from "lodash";
+import { FixedNumber } from "@ethersproject/bignumber";
+import { PriceFeedInterface } from "./PriceFeedInterface";
+import type { Logger } from "winston";
+import Web3 from "web3";
+import { BN } from "../types";
+import { NetworkerInterface } from "./Networker";
+
+interface RapidAPIQuoteReponse {
+  quoteResponse?: {
+    result?: { regularMarketPrice: number; symbol: string }[];
+  };
+}
+
+type RapidAPIHistoryReponse = Record<
+  string,
+  {
+    timestamp: number[];
+    close: number[];
+  }
+>;
+
+export class USPACPriceFeed extends PriceFeedInterface {
+  private readonly uuid: string;
+  private readonly toBN = Web3.utils.toBN;
+  private currentPrice: BN | null = null;
+  private lastUpdateTime: number | null = null;
+
+  /**
+   * @notice Constructs the USPACPriceFeed.
+   * @param {Object} logger Winston module used to send logs.
+   * @param {Object} web3 Provider from truffle instance to connect to Ethereum network.
+   * @param {String[]} symbols Array of symbols that constitute basket
+   * @param {String} correctionFactor factor to multiply basket price
+   * @param {String} rapidApiKey Rapid API access key.
+   * @param {String} interval granularity for price data. Example: "1m 5m 15m 1d 1wk 1mo"
+   * @param {Integer} lookback How far in the past the historical prices will be available using getHistoricalPrice.
+   * @param {Object} networker Used to send the API requests.
+   * @param {Function} getTime Returns the current time.
+   * @param {Number} priceFeedDecimals Number of priceFeedDecimals to use to convert price to wei.
+   * @param {Integer} minTimeBetweenUpdates Min number of seconds between updates. If update() is called again before
+   *      this number of seconds has passed, it will be a no-op.
+   */
+  constructor(
+    private readonly logger: Logger,
+    private readonly web3: Web3,
+    private readonly symbols: string[],
+    private readonly correctionFactor: string,
+    private readonly rapidApiKey: string,
+    private readonly interval: string = "1m",
+    private readonly lookback: number,
+    private readonly networker: NetworkerInterface,
+    private readonly getTime: () => Promise<number>,
+    private readonly priceFeedDecimals = 18,
+    private readonly minTimeBetweenUpdates = 3600 // 1 hour is enough to fit in free plan (1000 requests per day)
+  ) {
+    super();
+
+    assert(symbols != null, "symbols must be provided");
+    assert(rapidApiKey != null, "rapidApiKey must be provided");
+    assert(correctionFactor != null, "correctionFactor must be provided");
+
+    this.uuid = `USPACPriceFeed-${symbols.join(",")}`;
+  }
+
+  public getCurrentPrice(): BN | null {
+    return this.currentPrice;
+  }
+
+  public async getHistoricalPrice(time: number, ancillaryData: string, verbose = false): Promise<BN> {
+    // Construct URL.
+    // https://rapidapi.com/principalapis/api/stock-data-yahoo-finance-alternative/
+    const symbolsStr = this.symbols.join("%2C");
+    /* range is a string with format `{number_of_days}d` */
+    const range = Math.ceil(this.lookback / 3600 / 24) + "d";
+    const url = `https://stock-data-yahoo-finance-alternative.p.rapidapi.com/v8/finance/spark?symbols=${symbolsStr}&range=${range}&interval=${this.interval}`;
+    const options = {
+      method: "GET",
+      headers: {
+        "x-rapidapi-host": "stock-data-yahoo-finance-alternative.p.rapidapi.com",
+        "x-rapidapi-key": this.rapidApiKey,
+      },
+    };
+    const response = (await this.networker.getJson(url, options)) as RapidAPIHistoryReponse | null | undefined;
+
+    if (response == null) {
+      throw new Error(`🚨Invalid response from url ${url}`);
+    }
+
+    if (verbose) {
+      console.log(`Rapid API response for url ${url} is ${JSON.stringify(response, null, 2)}`);
+    }
+
+    // Validate response
+    for (const symbol of this.symbols) {
+      const stockData = response[symbol];
+      if (stockData == null) {
+        throw new Error(`🚨Could not parse price result from url ${url}: missing data for symbol ${symbol}`);
+      }
+      if (stockData.timestamp.length == 0 || stockData.timestamp.length != stockData.close.length) {
+        throw new Error(`🚨Could not parse price result from url ${url}: invalid data for symbol ${symbol}`);
+      }
+      for (const ts of stockData.timestamp) {
+        if (typeof ts != "number") {
+          throw new Error("Could not parse data: invalid timestamp type");
+        }
+      }
+      for (const p of stockData.close) {
+        if (typeof p != "number") {
+          throw new Error("Could not parse data: invalid price type");
+        }
+      }
+      for (let i = 1; i < stockData.timestamp.length; i++) {
+        if (stockData.timestamp[i] <= stockData.timestamp[i - 1]) {
+          throw new Error(`Could not parse data: timestamps for ${symbol} are not increasing monotonously`);
+        }
+      }
+    }
+
+    // Get prices
+    const prices = this.symbols.map((symbol) => {
+      const stockData = response[symbol];
+
+      // iterate over timestamps and prices backwards, until we find timestamp
+      // that is earlier then given time
+      const result = lodash
+        .zip(stockData.timestamp, stockData.close)
+        .reverse()
+        .find((item) => time >= (item[0] as number));
+      if (result == null) {
+        throw new Error(`Could not get historical price for symbol ${symbol}: timestamp ${time} past last data point`);
+      }
+      const price = result[1] as number;
+      return price;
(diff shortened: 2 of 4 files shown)


packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
```diff
@@ -984,6 +984,12 @@ export const defaultConfigs: { [name: string]: { type: string; [key: string]: an
     bridgeAdminAddress: "0x30B44C676A05F1264d1dE9cC31dB5F2A945186b6",
     rateModels: across.constants.RATE_MODELS,
   },
+  uSPAC10: {
+    type: "uSPAC",
+    lookback: 7200,
+    correctionFactor: "1.0",
```
Reinis-FRP, line 990: What is `correctionFactor` parameter? There is no reference to this in the draft UMIP

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
```diff
@@ -0,0 +1,214 @@
...
+    private readonly logger: Logger,
+    private readonly web3: Web3,
+    private readonly symbols: string[],
+    private readonly correctionFactor: string,
+    private readonly marketstackApiKey: string,
+    private readonly useHttps: boolean,
+    private readonly lookback: number,
+    private readonly networker: NetworkerInterface,
+    private readonly getTime: () => Promise<number>,
+    private readonly priceFeedDecimals = 4 /* Prices in Marketstack are 2 decimals, but give some extra precision */,
```
Reinis-FRP, line 44: `priceFeedDecimals` parameter has no relation to input precision, but how to format returned price. It was important before UMA contracts could deal with different collateral decimals, but now this should default to 18

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  private async fetchPrice(time: number, verbose: boolean): Promise<BN> {
+    // Find the closest day that completed before the beginning of the lookback window, and use
+    // it as the start date.
+    // Marketstack API does not return data for non-working days. We must give
+    // lookback that is big enough to cover at least one working day. For
+    // example, NYSE exchange can have maximum three consecutive non-working
+    // day. We make some reserve and set 10 days as minimal lookback.
+    const MINIMAL_LOOKBACK = 3600 * 24 * 10;
+    const startLookbackWindow = time - Math.max(this.lookback, MINIMAL_LOOKBACK);
```
Reinis-FRP, line 110: Normally in other price feed implementations `lookback` refers to time from when liquidable position is being evaluated to current time. In order not to cause confusion by the consuming applications it should be renamed to different parameter. As an example, [TraderMadePriceFeed.ts](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/TraderMadePriceFeed.ts) has `minuteLookback` and `hourlyLookback` parameters while [CryptoWatchPriceFeed.ts](https://github.com/UMAprotocol/protocol/blob/master/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts) has `historicalTimestampBuffer` parameter to allow using earlier timestamp if the price is not available at request timestamp.

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
dmitry-vsl: Fixed

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
dmitry-vsl: I pushed new commit that removes this code, is your comment still actual?

packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
dmitry-vsl: We have added information on `correctionFactor` to UMIP

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
```diff
@@ -0,0 +1,251 @@
...
+    // Get prices
+    const prices = [];
+    for (const symbol of this.symbols) {
+      const stockData = response[symbol];
+      /*
+        if time is greater than last timestamp (that is equal to the beginning
+        of the last market open datetime) plus 24hours then new day begin and
+        we should use current price
+      */
+      if (time >= stockData.timestamp[stockData.timestamp.length - 1] + 3600 * 24) {
```
Reinis-FRP, line 126: This logic is not specified in UMIP and would result in different price for the same request timestamp just depending on when one runs this script.

Reinis-FRP, line 126: Also, last timestamp not always corresponds to the the last market open datetime. At least when I checked it during the market hours the last timestamp was current timestamp with real-time price.

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
```diff
@@ -0,0 +1,251 @@
...
+        }
+        return this.currentPrice;
+      }
+
+      // iterate over timestamps and prices backwards, until we find timestamp
+      // that is earlier then given time
+      const result = lodash
+        .zip(stockData.timestamp, stockData.close)
+        .reverse()
+        .find((item) => time >= (item[0] as number));
```
Reinis-FRP, line 138: I understand the returned `stockData.timestamp` is market open timestamp with close price for the same day returned as `stockData.close` - so if price request is for timestamp before closing this logic would return the closing price that is in the future relative to the request timestamp. Hence you would need to convert `stockData.timestamp` to close timestamp so that it can be accurately compared with request timestamp.

Also note that converting from open to close timestamp sometimes would be challenging due to having earlier closing in some of holidays. Hence alternative would be to consider intra-day price data.

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
dmitry-vsl: Reworked to use intraday price

dmitry-vsl: Reworked to use intraday price

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
dmitry-vsl: I reworked to use intraday price. Are any of your concerns still valid?

Reinis-FRP: When looking into returned data it seems that also for intraday granularity rapidapi reports 1m period opening timestamps paired with close prices for the same 1m period. This means that requesting price at 10:00:30 AM the above logic normally would match it to 10:00:00 opening price while the corresponding close price would be at 10:01:00 that is after the request timestamp. 

There is an exception to above statement if requesting price during trading hours that is just couple of seconds behind real-time price - for example, if at 10:00:45 AM one would request price at 10:00:30 then the above algorithm would correctly resolve to the closing price for the 9:59:00-10:00:00 period since 10:00:00-10:01:00 period is not yet available.

dmitry-vsl: > This means that requesting price at 10:00:30 AM the *above logic* normally would match it to 10:00:00 *opening* price

What _above logic_ do you mean? If you mean the code above, then requesting price for 10:00:30 AM, the code above will find 10:00:00 timestamp and corresponding price, which is, as you say, is *close* price for 10:00:00 - 10:01:00 interval. So the code above will give the *closing* price for the interval, not *opening* price.

Do you think that the correct price for the middle of the interval is the *open* price for the interval, and thus the code above incorrectly returns *close* price for the interval?

dmitry-vsl: The real price for the middle of the interval is somewhere between open price and close price. So neither open price nor the close price are absolutely correct. So we have no reasons to prefer one to another.

Since API gives us only close prices, adjusting timestamps trying to get open price will be difficult. We will have to treat differently timestamps that are in the middle of trading day and timestamps that are after the end of trading day, because in the first case we have to get open price, and in the second case we still have to get the closing price for the last interval (15:59:00-16:00:00)

So I think the correct decision would be to use *closing* price for the interval, as it is currently implemented

pemulis: A different way of looking at it is that the price identifier could get the _most recent_ price, which would be an open price in the middle of the day and a closing price after the end of the trading day. I could see the argument that the user would expect the most recent price when making a price request, since the settlement price would not be known at the time of the request.

Another wrinkle is that a bot that responds to price requests automatically might not be able to know how to respond to this price request if it always uses the closing price, since the closing price might not be known yet. In other words, the price request might not be resolvable, since it would often be requesting information that is (very slightly) in the future.

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  public getPriceFeedDecimals(): number {
+    return this.priceFeedDecimals;
+  }
+
+  public async update(): Promise<void> {
+    const currentTime = await this.getTime();
+
+    // Return early if the last call was too recent.
+    if (this.lastUpdateTime !== null && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
```
pemulis, line 157: Is it correct to assume that this check will only work in a set-up that retains state between runs? For instance, there are bots that run in a serverless mode and will not know when the last update time was. Would the consequence be that those bots would exceed the limit on the free plan for the API? (@Reinis-FRP does that sound correct to you?) Any other consequences I'm not considering?

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
Reinis-FRP: > Do you think that the correct price for the middle of the interval is the _open_ price for the interval, and thus the code above incorrectly returns _close_ price for the interval?

In case the liquidation bot requests price within the same 1 minute interval as the current time (e.g. requesting price for timestamp 10:00:30 AM at 10:00:45 AM)  there would not yet be the close price available for the current 1 minute interval. Instead rapidapi would return the close price of the previous interval (ending 10:00:00 AM). This could become problematic if the dispute bot uses this price feed a bit later (e.g. requesting price for timestamp 10:00:30 AM at 10:01:30 AM) - then rapidapi would return the close price for the requested interval (ending at 10:01:00 AM) and it might fire the dispute even if both bots are run honestly.

Hence the only feasible method to have deterministic price resolution (independent on when the bot queries the API) is to use close price of the previous 1 minute interval (e.g. request price for timestamp 10:00:30 AM should be resolved with close price for the interval opening at 09:59:00 AM). In case of request being made during trading hours you should also filter out the last returned real-time price since that could be changing every time the request is made).

Also you should amend the the implementation section in the `uSPAC10` price identifier UMIP to reflect this same logic (using close price of previous 1 minute interval). Otherwise this could be rightfully disputed at the DVM escalation.

Reinis-FRP: > So I think the correct decision would be to always use _closing_ price for the interval, as it is currently implemented

Due to reasons in my previous comment it would be safer to use _close_ price for the previous interval.

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
Reinis-FRP: Yes, in order to not run out of free tier limits the bot should be run in environment capable of keeping the state (i.e. using POLLING_DELAY > 0 or customizing the bot to save lastUpdateTime to some other medium when running in serverless mode).

packages/financial-templates-lib/src/price-feed/USPACPriceFeed.ts
dmitry-vsl: > In case the liquidation bot requests price within the same 1 minute interval as the current time (e.g. requesting price for timestamp 10:00:30 AM at 10:00:45 AM) there would not yet be the close price available for the current 1 minute interval.

As far as I can see, liquidation bot never calls `getHistoricalPrice`, it only calls `getCurrentPrice` which is implemented by using different call to RapidAPI. So the liquidation bot never operates with the notion of the exact price for some specific moment in time. It corresponds to the design of EMP smart contract. The function `createLiquidation` in EMP contract does not receive timestamp as an argument, the timestamp is taken from the block time. If we take into account block mining latency, then we can conclude that entire system starting from smart contract is not designed for 1min granularity.

In case you describe, when liquidation bot asks for the price in the middle of the interval, RapidAPI `Quotes` call will return the current price, which probably will be closer to the __current__ interval close price (which will be used by disputer), than __previous__ interval close time.

So I don't see any advantage in using __previous__ interval close price. But it has some difficulties. The major one is that the interval open price does not always equal to previous interval close price.

Another difficulty is that we will have two different cases. The first case is when the timestamp is in the middle of the day, the second case is when the timestamp is after the end of the day. In the last case we still have to use closing price for the current interval (which starts at 15:59:00). We have no easy way in code to decide which case do we have. We will have to either hardcode exchange closing time or implement some heuristic, like if the last interval time was more than 1 minute ago, then the day is closed.

mrice32 (approved):
LGTM!

mrice32:
@dmitry-vsl this is ready to merge, but you'll need to fix DCO by signing off your commits. See here for instructions on how to do that: https://github.com/UMAprotocol/protocol/pull/3602/checks?check_run_id=4482823940. Once that's done, I can merge! Sorry for the delay!!

dmitry-vsl:
> @dmitry-vsl this is ready to merge, but you'll need to fix DCO by signing off your commits. See here for instructions on how to do that: https://github.com/UMAprotocol/protocol/pull/3602/checks?check_run_id=4482823940. Once that's done, I can merge! Sorry for the delay!!

Fixed

mrice32:
> > @dmitry-vsl this is ready to merge, but you'll need to fix DCO by signing off your commits. See here for instructions on how to do that: https://github.com/UMAprotocol/protocol/pull/3602/checks?check_run_id=4482823940. Once that's done, I can merge! Sorry for the delay!!
> 
> Fixed

Thank you!!

PR #3608, opened 2021-11-15
feat(across-relayer): enable batch disputing

chrismaree:
**Motivation**

The disputer should be able to send batches of disputes. This PR enables this.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -123,21 +123,30 @@ export class Relayer {
 
     // Build dictionary of pending relays keyed by l1 token and deposit hash. getPendingRelays() filters
     // out Finalized relays.
-    const pendingRelays: Relay[] = this.l1Client.getPendingRelayedDeposits();
-    if (pendingRelays.length == 0) {
+    const pendingRelays: { [key: string]: Relay[] } = this.l1Client.getPendingRelayedDepositsGroupedByL1Token();
+    if (Object.keys(pendingRelays).length == 0) {
       this.logger.debug({ at: "AcrossRelayer#Disputer", message: "No pending relays" });
       return;
     }
-    this.logger.debug({ at: "AcrossRelayer#Disputer", message: `Processing ${pendingRelays.length} pending relays` });
-
-    for (const relay of pendingRelays) {
+    this.logger.debug({
+      at: "AcrossRelayer#Disputer",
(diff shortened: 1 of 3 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -123,21 +123,30 @@ export class Relayer {
...
       return;
     }
-    this.logger.debug({ at: "AcrossRelayer#Disputer", message: `Processing ${pendingRelays.length} pending relays` });
-
-    for (const relay of pendingRelays) {
+    this.logger.debug({
+      at: "AcrossRelayer#Disputer",
+      message: `Processing ${Object.keys(pendingRelays).length} l1 token pending relays`,
+    });
+    for (const l1Token of Object.keys(pendingRelays)) {
```
chrismaree, line 135: this change (and the introduction of `getPendingRelayedDepositsGroupedByL1Token`) is done to ensure that we only send the batch disputes to one contract _at a time_. i.e consider if both the USDC and WETH pool have disputes. we cant send these in a batch as the destination contract is different between the two. The implementation here now sends them based off a grouping of `BridgePool`.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -496,23 +499,12 @@ export class Relayer {
...
-    if (receipt?.events?.RelayCanceled)
-      this.logger.error({
-        at: "AcrossRelayer#Disputer",
-        message: "Dispute failed to send to OO 👀!",
-        tx: receipt.transactionHash,
-        depositHash: receipt.events.RelayCanceled.returnValues.depositHash,
-        relayHash: receipt.events.RelayCanceled.returnValues.relayHash,
-        disputer: receipt.events.RelayCanceled.returnValues.disputer,
-      });
+  private async _generateDisputeRelayTransaction(deposit: Deposit, relay: Relay) {
```
chrismaree, line 502: this method now returns the dispute transaction, just like the similar methods in this file `_generateRelayTransaction` and `_generateSettleTransactionForSettleableRelay`

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: You should add this as a comment to the code to make sure no one accidentally introduces the possiiblity that batches can be sent to multiple contracts

nicholaspai (approved):
Looks great, LGTM

PR #3609, opened 2021-11-15
feat(transaction-manager): enable early return from runTransaction method

chrismaree:
**Motivation**

To send many transactions in a quick succession we require the ability to tell a transaction to not wait to be mined in a block and rather return as soon as we have a transaction hash. This PR implements this.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/TransactionUtils.ts
```diff
@@ -101,17 +116,32 @@ export const runTransaction = async ({
...
     // If the config contains maxPriorityFeePerGas then this is a London transaction. In this case, simply use the
     // provided config settings but double the maxFeePerGas to ensure the transaction is included, even if the base fee
     // spikes up. The difference between the realized base fee and maxFeePerGas is refunded in a London transaction.
     if (transactionConfig.maxFeePerGas && transactionConfig.maxPriorityFeePerGas) {
-      receipt = await transaction.send({
-        ...transactionConfig,
-        maxFeePerGas: parseInt(transactionConfig.maxFeePerGas.toString()) * 2,
-        type: "0x2",
-      } as SendOptions);
+      if (waitForMine)
```
chrismaree, line 125: here is the main tx logic. If we are `waitForMine` true then we run the transaction exactly as before.

Else, we use the `web3.eth.sendTransaction` function. This, combined with a promise, lets us return as soon as we have a `transactionHash`

packages/common/src/TransactionUtils.ts
```diff
@@ -19,6 +19,10 @@ export interface AugmentedSendOptions {
   maxPriorityFeePerGas?: number | string;
 }
 
+interface AugmentedWeb3 extends Web3 {
```
chrismaree, line 22: to accommodate the nonce management, we append a `nonces` object to the web3 interface. This acts as a global nonce counter.

packages/common/src/TransactionUtils.ts
```diff
@@ -76,6 +86,16 @@ export const runTransaction = async ({
   // This method does not play nicely in tests. Leave the nonce null to auto fill.
   else if (argv.network != "test") transactionConfig.nonce = await web3.eth.getTransactionCount(transactionConfig.from);
 
+  // Store the transaction nonce in the web3 object so that it can be used in the future. This enables us to fire a
+  // bunch of transactions off without needing to wait for them to be included in the mempool by manually incrementing.
+  if ((web3 as AugmentedWeb3).nonces?.[transactionConfig.from])
```
nicholaspai, line 91: The stored nonces will reset on each serverless run right? So is it possible that you'll get nonce collisions if the transactions don't mine fast enough between serverless runs?

nicholaspai (commented):
waitForMine looks great, but I'm trying to understand the nonce management logic more

packages/common/src/TransactionUtils.ts
chrismaree: that's correct

packages/common/src/TransactionUtils.ts
```diff
@@ -19,6 +19,10 @@ export interface AugmentedSendOptions {
   maxPriorityFeePerGas?: number | string;
 }
 
+interface AugmentedWeb3 extends Web3 {
+  nonces: { [key: string]: number };
```
mrice32, line 23:
```suggestion
  nonces: { [address: string]: number };
```

packages/common/src/TransactionUtils.ts
```diff
@@ -76,6 +86,16 @@ export const runTransaction = async ({
   // This method does not play nicely in tests. Leave the nonce null to auto fill.
   else if (argv.network != "test") transactionConfig.nonce = await web3.eth.getTransactionCount(transactionConfig.from);
 
+  // Store the transaction nonce in the web3 object so that it can be used in the future. This enables us to fire a
+  // bunch of transactions off without needing to wait for them to be included in the mempool by manually incrementing.
+  if ((web3 as AugmentedWeb3).nonces?.[transactionConfig.from])
+    transactionConfig.nonce = ++(web3 as AugmentedWeb3).nonces[transactionConfig.from];
```
mrice32, line 92: Instead of casting in-line, could we cast at the top of the method and just use AudmentedWeb3 everywhere in this method?

packages/common/src/TransactionUtils.ts
```diff
@@ -101,17 +121,32 @@ export const runTransaction = async ({
...
     // provided config settings but double the maxFeePerGas to ensure the transaction is included, even if the base fee
     // spikes up. The difference between the realized base fee and maxFeePerGas is refunded in a London transaction.
     if (transactionConfig.maxFeePerGas && transactionConfig.maxPriorityFeePerGas) {
-      receipt = await transaction.send({
-        ...transactionConfig,
-        maxFeePerGas: parseInt(transactionConfig.maxFeePerGas.toString()) * 2,
-        type: "0x2",
-      } as SendOptions);
+      if (waitForMine)
+        receipt = await transaction.send({
```
mrice32, line 131: I know we've already discussed this a bit, but would the following work?

```ts
        receiptPromiEvent = transaction.send({
          ...transactionConfig,
          maxFeePerGas: parseInt(transactionConfig.maxFeePerGas.toString()) * 2,
          type: "0x2",
        } as SendOptions);
        hash = await new Promise((resolve) => {
         receiptPromiEvent.on("transactionHash", (hash) => {
              resolve({ receipt: { transactionHash: hash } as any, transactionConfig, returnValue: null as any });
         });
...
return { receipt: (receipt as unknown) as PromiEvent<TransactionReceipt>, hash, returnValue, transactionConfig };

```

This approach would allow us to drop the `waitForMine` argument and would mean a little less casework, but it would mean that we _always_ return and unresolved PromiEvent for the receipt that the caller would then need to await on, which might add some messiness to the existing bots.

Just wanted to throw an alternative idea out, but I think what you have here works well and might be slightly easier on the existing callers.

mrice32 (commented):
Looks awesome!!!

As we discussed offline, we probably need to have some way for the caller to grab the original send promise in the waitForMine case.

packages/common/src/TransactionUtils.ts
chrismaree: I really like this as we can store all the promises and then at the end of the execution block use a `Promose.all` to wait for all transactions to mine. If any of them revert/error out then we can log accordingly.

packages/common/src/TransactionUtils.ts
chrismaree: yup. done ptal

packages/common/src/TransactionUtils.ts
```diff
@@ -30,26 +43,30 @@ const argv = minimist(process.argv.slice(), {});
...
   availableAccounts?: number;
-}): Promise<{
-  receipt: TransactionReceipt;
-  returnValue: CallReturnValue;
-  transactionConfig: AugmentedSendOptions;
-}> => {
+  waitForMine?: boolean;
+}): Promise<executedTransaction> => {
+  // use a cast version of web3 to enable callers to not have to define the `nonces` mapping in the web3 object.
+  const augmentedWeb3 = web3 as AugmentedWeb3;
```
chrismaree, line 66: @mrice32 lmk if this is what you meant. problem is we need the interface on this method to take in a normal web3 object which we then append the `nonces` onto.

packages/common/src/TransactionUtils.ts
chrismaree: please see the refined implementation.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -33,6 +31,8 @@ export enum RelaySubmitType {
 }
 
 export class Relayer {
+  executedTransactions: Array<executedTransaction> = []; // store all submitted transactions during execution lifecycle.
```
chrismaree, line 34: this is part of the main change in this PR: store all `executedTransaction`s in an array that we can then `await` at the end of the execution loop.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -571,33 +581,37 @@ export class Relayer {
...
           message,
-          mrkdwn: mrkdwn + " tx: " + createEtherscanLinkMarkdown(receipt.transactionHash),
+          mrkdwn: mrkdwn + " tx: " + createEtherscanLinkMarkdown(executionResult.transactionHash),
         });
-        return { txStatus: true, receipt, transactionConfig };
-      } else throw receipt;
+        // Just because the transaction was successfully included in the mem pool does not mean it will be mined without
+        // reverting. Store the transaction execution result within the executedTransactions array. This is processed
+        // at the end of the bot execution loop to ensure that all submitted transactions were successfully included.
+        this.executedTransactions.push(executionResult);
```
chrismaree, line 609: this is the only part of this file that sends transactions and therefore the only place where we need to append transaction results :)

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -113,7 +113,14 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
 
           if (config.botModes.l2FinalizerEnabled) await crossDomainFinalizer.checkForBridgeableL2TokensAndBridge();
           else logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "Cross domain finalizer disabled" });
+
+          // Each of the above code blocks could have produced transactions. If they did, their promises are stored
+          // in the executed transactions array. The method below awaits all these transactions to ensure they are
+          // correctly included in a block. if any submitted transactions contains an error then a log is produced.
+          await processTransactionPromiseBatch(relayer.getExecutedTransactions(), logger);
```
chrismaree, line 120: note here: at the end of the execution loop we process the array of transactions by awaiting each. if any failed it will produce a log accordingly.

packages/common/src/TransactionUtils.ts
mrice32: Yes, this is exactly what I meant. To keep the web3 name, you might be able to remap the name of the web3 object that the method takes in like this:
```
export const runTransaction = async ({
  web3: _web3,
  transaction,
  ...

  const web3 = web3_ as AugmentedWeb3;
```

Just to reduce confusion and keep the usage consistent. Up to you, though.

packages/common/src/TransactionUtils.ts
```diff
@@ -70,11 +88,24 @@ export const runTransaction = async ({
...
+  if (await accountHasPendingTransactions(augmentedWeb3, transactionConfig.from))
+    transactionConfig.nonce = await getPendingTransactionCount(augmentedWeb3, transactionConfig.from);
   // Else, there is no pending transaction and we use the current account transaction count as the nonce.
   // This method does not play nicely in tests. Leave the nonce null to auto fill.
-  else if (argv.network != "test") transactionConfig.nonce = await web3.eth.getTransactionCount(transactionConfig.from);
+  else if (argv.network != "test" && argv._.includes("test"))
+    transactionConfig.nonce = await augmentedWeb3.eth.getTransactionCount(transactionConfig.from);
+  // Store the transaction nonce in the augmentedWeb3 object so that it can be used in the future. This enables us to fire a
+  // bunch of transactions off without needing to wait for them to be included in the mempool by manually incrementing.
+  console.log("argv", argv);
```
mrice32, line 99: console log?

packages/common/src/TransactionUtils.ts
```diff
@@ -70,11 +88,24 @@ export const runTransaction = async ({
...
+  if (argv.network != "test" && argv._.includes("test")) {
```
mrice32, line 100: What's the problem on test networks?

packages/common/src/TransactionUtils.ts
```diff
@@ -262,3 +307,29 @@ export async function findBlockNumberAtTimestamp(
...
+  });
+  const transactionResults = await Promise.allSettled(transactions.map((tx: any) => tx.receipt));
+  const revertedTransactions = (transactionResults as any).filter((result: any) => result.status === "rejected");
+  if (revertedTransactions.length == 0)
+    logger.debug({ at: "TransactionUtils", message: "Transaction batch processed without error" });
+  else
+    logger.error({
+      at: "TransactionUtils",
+      message: "Transaction batch processed with error",
+      errors: revertedTransactions.map((transaction: any) => transaction.reason),
```
mrice32, line 333: Would it be useful to include the transaction hashes for those that failed? It seems like we would get a previous log telling us the txn was sent, so it might be nice to be able to link each error back to its txn hash. Thoughts? Would this be difficult to add?

packages/common/src/TransactionUtils.ts
```diff
@@ -19,6 +21,17 @@ export interface AugmentedSendOptions {
   maxPriorityFeePerGas?: number | string;
 }
 
+interface AugmentedWeb3 extends Web3 {
+  nonces: { [address: string]: number };
+}
+
+export interface executedTransaction {
```
mrice32, line 28: nit: types should be cap case
```suggestion
export interface ExecutedTransaction {
```

packages/common/src/TransactionUtils.ts
```diff
@@ -19,6 +21,17 @@ export interface AugmentedSendOptions {
...
+  receipt: TransactionReceipt;
```
mrice32, line 29: Why not make this an `|` type with a PromiEvent so we don't have to cast to `any`? Otherwise, callers will be expecting one type but receiving another.

packages/common/src/TransactionUtils.ts
```diff
@@ -101,17 +132,32 @@ export const runTransaction = async ({
     const maximumGasPriceMultiple = 2 * 3;
     // Pre-London transactions require `gasPrice`, London transactions require `maxFeePerGas` and `maxPriorityFeePerGas`
 
-    let receipt;
+    let receipt: any;
```
mrice32, line 135: As mentioned above, it could make sense to set this receipt type to `TransactionReceipt | PromiEvent<TransactionReceipt>`.

packages/common/src/TransactionUtils.ts
```diff
@@ -101,17 +132,32 @@ export const runTransaction = async ({
...
+      else {
+        const receiptPromiseEvent = transaction.send({
+          ...transactionConfig,
+          maxFeePerGas: parseInt(transactionConfig.maxFeePerGas.toString()) * 2,
+          type: "0x2",
+        } as SendOptions);
+        const transactionHash: string = await new Promise((resolve) => {
+          receiptPromiseEvent.on("transactionHash", (transactionHash) => resolve(transactionHash));
+        });
+        return { receipt: receiptPromiseEvent as any, transactionHash, returnValue, transactionConfig };
```
mrice32, line 159: Any reason we can't just set the receipt as we do in the if above and use the existing return statement below?

mrice32 (commented):
Looks awesome! A few questions and comments

packages/common/src/TransactionUtils.ts
chrismaree: mistake. fixed

packages/common/src/TransactionUtils.ts
chrismaree: we had this before. reason is really that sometimes we re-initalize web3 objects and sometimes this does not place well with the object storing it's nonce. without this a lot of tests fail. solution is to just let test run "normally" where they get the nonce from the provider.

packages/common/src/TransactionUtils.ts
chrismaree: this `reason` will include the transaction has. it'll look something like this:
```
{
 "blockHash": "0xd1c3f3557c5819cf59b0c8d833b03439fcf458d0c1831dce0dccd6965e32b1f2",
 "blockNumber": 13707214,
 "contractAddress": null,
 "cumulativeGasUsed": 12226892,
 "effectiveGasPrice": "0x1912a70f32",
 "from": "0x428ab2ba90eba0a4be7af34c9ac451ab061ac010",
 "gasUsed": 32907,
 "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
 "status": false,
 "to": "0x7355efc63ae731f584380a9838292c7046c1e433",
 "transactionHash": "0xf9039dcef419fc43d198b2bd6c18235f1c3b50ba0ed414a6e2b533bd9b4b8b27",
 "transactionIndex": 223,
 "type": "0x2",
 "events": {}
}
```

so we get the tx hash from this for free

packages/common/src/TransactionUtils.ts
chrismaree: I tried using this syantax but this made it harder downstream in places that consume the output of the `runTransaction` function as they need to infer that status of the transaction based on the type. Rather, I'm going to add an additional prop to this interface: one for the `TransactionReceipt` and one for the `PromiEvent<TransactionReceipt>` so that we dont need to overload this param.

chrismaree: on second thought your implementation is cleaner. the case work was pretty minimal won stream and could be fixed with a few simple casts.

mrice32: Other options:
1. [Overloading the method](https://www.typescriptlang.org/docs/handbook/2/functions.html#function-overloads). This would mean that if the compiler can deduce that the input value is true (i.e. it's hardcoded at the callsite), then it can deduce the correct type of the return value.
2. Creating two distinct methods, one that waits, one that doesn't.

packages/common/src/TransactionUtils.ts
mrice32: Any thoughts on this comment?

packages/common/src/TransactionUtils.ts
mrice32: Why can't we make this `TransactionReceipt | PromiEvent<TransactionReceipt>` since it's going to be used as the return value?

mrice32 (commented):
Looks good! Only a few remaining comments from the previous review that I want to go back to. Otherwise, I think this is g2g!

packages/common/src/TransactionUtils.ts
chrismaree: you are 100% right! fixed.

packages/common/src/TransactionUtils.ts
chrismaree: yup yup. I think the original proposal you had is the cleanest.

packages/common/src/TransactionUtils.ts
chrismaree: this is added now

packages/common/src/TransactionUtils.ts
chrismaree: main reason is we get the transaction hash in a differnt way in this code block than in the normal send logic (i.e we need to wait on the `transactionHash` promise resolution, and cant pull from the `receipt`).

I'm going to refactor this to have a `transactionhash` defined above so we can have a conistant return flow.

packages/common/src/TransactionUtils.ts
```diff
@@ -101,35 +130,52 @@ export const runTransaction = async ({
...
+      // Else, waitForMine is false and we return the transaction hash immediately as soon as it is included in the
+      // mempool. Receipt is a promise of the pending transaction that can be awaited later to ensure block inclusion.
+      else {
+        receipt = (transaction.send({
+          ...transactionConfig,
+          maxFeePerGas: parseInt(transactionConfig.maxFeePerGas.toString()) * 2,
+          type: "0x2",
+        } as SendOptions) as unknown) as PromiEvent<TransactionReceipt>;
+        transactionHash = await new Promise((resolve) => {
+          (receipt as any).on("transactionHash", (transactionHash: any) => resolve(transactionHash));
```
mrice32, line 158: Does PromiEvent not have the `on` method?

mrice32 (approved):
LGTM!

nicholaspai (approved):
LGTM

packages/common/src/TransactionUtils.ts
nicholaspai: > I'm going to refactor this to have a transactionhash defined above so we can have a conistant return flow.

agreed with this

packages/common/src/TransactionUtils.ts
chrismaree: no :(

PR #3611, opened 2021-11-16
feat(across-contracts): added WETH<->ETH wrappers for optimism

chrismaree:
**Motivation**

To get the Across contracts to work with ETH on optimism we need to add a small workaround to deal with their canonical bridge not supporting WETH bridging. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/insured-bridge/ovm/OVM_BridgeDepositBox.sol
+++ b/packages/core/contracts/insured-bridge/ovm/OVM_BridgeDepositBox.sol
@@ -133,7 +133,7 @@ contract OVM_BridgeDepositBox is BridgeDepositBox, OVM_CrossDomainEnabled {
      * @param l2Token L2 token to relay over the canonical bridge.
      * @param l1Gas Unused by optimism, but included for potential forward compatibility considerations.
      */
-    function bridgeTokens(address l2Token, uint32 l1Gas) public override nonReentrant() {
+    function bridgeTokens(address l2Token, uint32 l1Gas) public virtual override nonReentrant() {
         uint256 bridgeDepositBoxBalance = TokenLike(l2Token).balanceOf(address(this));
         require(bridgeDepositBoxBalance > 0, "can't bridge zero tokens");
         require(canBridge(l2Token), "non-whitelisted token or last bridge too recent");
--- a/packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
+++ b/packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
@@ -1,0 +1,67 @@
+pragma solidity ^0.8.0;
+
+import "./OVM_BridgeDepositBox.sol";
+
+contract OVM_OETH_BridgeDepositBox is OVM_BridgeDepositBox {
+    // Stored to work around Optimism not allowing for WETH to be bridge over the canonical bridge.
+    address public l2Weth;
+    address public l2Eth;
+    address public l1EthWrapper;
+
+    constructor(
+        address _crossDomainAdmin,
+        uint64 _minimumBridgingDelay,
+        uint256 _chainId,
+        address _l1Weth,
+        address _l2Weth,
+        address _l2Eth,
+        address _l1EthWrapper,
+        address timerAddress
+    ) OVM_BridgeDepositBox(_crossDomainAdmin, _minimumBridgingDelay, _chainId, _l1Weth, timerAddress) {
+        l2Weth = _l2Weth;
+        l2Eth = _l2Eth;
+        l1EthWrapper = _l1EthWrapper;
+    }
+
+    /**
+     * @notice Called by relayer (or any other EOA) to move a batch of funds from the deposit box, through the canonical
+     *     token bridge, to the L1 Withdraw box. Implementation is exactly the same as the standard OVM_BridgeDepositBox
+     * except constructed to work with Optimism ETH by first unwrapping WETH then bridging OETH. The target on L1 is
+     * not the bridgePool but to the l1EthWrapper that takes any ETH sent to it, wraps it and sends to the BridgePool.
+     * @dev The frequency that this function can be called is rate limited by the `minimumBridgingDelay` to prevent spam
+     *      on L1 as the finalization of a L2->L1 tx is quite expensive.
+     * @param l2Token L2 token to relay over the canonical bridge.
+     * @param l1Gas Unused by optimism, but included for potential forward compatibility considerations.
+     */
+    function bridgeTokens(address l2Token, uint32 l1Gas) public override nonReentrant() {
+        uint256 bridgeDepositBoxBalance = TokenLike(l2Token).balanceOf(address(this));
+        require(bridgeDepositBoxBalance > 0, "can't bridge zero tokens");
+        require(canBridge(l2Token), "non-whitelisted token or last bridge too recent");
+
+        whitelistedTokens[l2Token].lastBridgeTime = uint64(getCurrentTime());
+
+        address bridgePool = whitelistedTokens[l2Token].l1BridgePool;
+
+        // If the L2 token is L2WETH then, to work with the canonical optimism bridge, we first unwrap it to ETH then
+        // bridge ETH over the canonical bridge. On L1 the l1EthWrapper will re-wrap the ETH to WETH and send it to
+        // the WETH bridge pool.
+        if (l2Token == l2Weth) {
+            WETH9Like(l2Token).withdraw(bridgeDepositBoxBalance);
+            l2Token = l2Eth;
+            bridgePool = l1EthWrapper;
+        }
+        StandardBridgeLike(Lib_PredeployAddresses.L2_STANDARD_BRIDGE).withdrawTo(
+            l2Token, // _l2Token. Address of the L2 token to bridge over.
+            bridgePool, // _to. Withdraw, over the bridge, to the l1 withdraw contract.
+            bridgeDepositBoxBalance, // _amount. Send the full balance of the deposit box to bridge.
+            l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+            "" // _data. We don't need to send any data for the bridging action.
+        );
+
+        emit TokensBridged(l2Token, bridgeDepositBoxBalance, l1Gas, msg.sender);
+    }
+
+    receive() external payable {}
+
+    fallback() external payable {}
+}
--- a/packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
+++ b/packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
@@ -1,0 +1,44 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+
+/**
+ * @title Optimism Eth Wrapper
+ * @dev Any ETH sent to this contract is wrapped into WETH and sent to the set bridge pool. This enables ETH to be sent
+ * over the canonical Optimism bridge, which does not support WETH bridging.
+ */
+interface WETH9Like {
+    function deposit() external payable;
+
+    function transfer(address guy, uint256 wad) external;
+
+    function balanceOf(address guy) external view returns (uint256);
+}
+
+contract Optimism_Wrapper is Ownable {
+    WETH9Like weth;
+    address bridgePool;
+
+    constructor(WETH9Like _weth, address _bridgePool) {
+        weth = _weth;
+        bridgePool = _bridgePool;
+    }
+
+    function changeBridgePool(address newBridgePool) public onlyOwner {
+        bridgePool = newBridgePool;
+    }
+
+    receive() external payable {
+        wrapAndTransfer();
+    }
+
+    fallback() external payable {
+        wrapAndTransfer();
+    }
+
+    function wrapAndTransfer() public payable {
+        weth.deposit{ value: address(this).balance }();
+        weth.transfer(bridgePool, weth.balanceOf(address(this)));
+    }
+}
--- a/packages/core/contracts/insured-bridge/BridgeDepositBox.sol
+++ b/packages/core/contracts/insured-bridge/BridgeDepositBox.sol
@@ -13,6 +13,8 @@ interface TokenLike {
 
 interface WETH9Like {
     function deposit() external payable;
+
+    function withdraw(uint256 wad) external;
 }
 
 /**


packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,81 @@
...
+        require(canBridge(l2Token), "non-whitelisted token or last bridge too recent");
+
+        whitelistedTokens[l2Token].lastBridgeTime = uint64(getCurrentTime());
+
+        address bridgePool = whitelistedTokens[l2Token].l1BridgePool;
+
+        // If the L2 token is L2WETH then, to work with the canonical optimism bridge, we first unwrap it to ETH then
+        // bridge ETH over the canonical bridge. On L1 the l1EthWrapper will re-wrap the ETH to WETH and send it to
+        // the WETH bridge pool.
+        if (whitelistedTokens[l2Token].l1Token == l1Weth) {
```
chrismaree, line 62: this is the main logic change.

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
```diff
@@ -0,0 +1,44 @@
...
+
+    receive() external payable {
+        wrapAndTransfer();
+    }
+
+    fallback() external payable {
+        wrapAndTransfer();
+    }
+
+    function wrapAndTransfer() public payable {
```
chrismaree, line 40: this function does the magic.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,81 @@
...
+            WETH9Like(l2Token).withdraw(bridgeDepositBoxBalance);
+            l2Token = l2Eth;
```
nicholaspai, line 64: This block has I think some redundant code. For example, if we're going to assume that IF `whitelistedTokens[l2Token].l1Token == l1Weth` means that we're dealing with the WETH token, then why do we need to set `l2Token = l2Eth`? 

What if we changed the if statement to: `if (l2Token == l2Eth && whitelistedTokens[l2Token].l1Token == l1Weth)`, or alternatively just checking `if (l2Token == l2Eth)`

nicholaspai (commented):
Some initial comments, I'm working on unit tests and will push them up.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,81 @@
...
+            bridgePool, // _to. Withdraw, over the bridge, to the l1 withdraw contract.
+            bridgeDepositBoxBalance, // _amount. Send the full balance of the deposit box to bridge.
+            l1Gas, // _l1Gas. Unused, but included for potential forward compatibility considerations
+            "" // _data. We don't need to send any data for the bridging action.
+        );
+
+        emit TokensBridged(l2Token, bridgeDepositBoxBalance, l1Gas, msg.sender);
+    }
+
+    receive() external payable {}
```
nicholaspai, line 78: If we intend to treat L2 WETH as an ERC20 why provide these fallback `payable` methods?

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
nicholaspai: Or, since setting a whitelisted L2 token's `l1Token = l1weth.address` is what we're assuming officially configures an L2 token as the WETH token, then we should just remove the line `l2Token = l2Eth`. 

Phrased differently: We should already be assuming that any `whitelistedTokens[l2Token]` whose `l1Token = l1WethAddress` is the L2 WETH token.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,67 @@
+pragma solidity ^0.8.0;
+
+import "./OVM_BridgeDepositBox.sol";
+
+contract OVM_OETH_BridgeDepositBox is OVM_BridgeDepositBox {
+    // Stored to work around Optimism not allowing for WETH to be bridge over the canonical bridge.
+    address public l2Weth;
+    address public l2Eth;
+    address public l1EthWrapper;
```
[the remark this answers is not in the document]
chrismaree: We can always deprecate it later and move to a new deposit box. Bridge admin can simply shut it down and upgrade by pointing the whitelist to the new address.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
chrismaree: When unwrapping weth on L2 the weth contract transfers eth to this contract. We need this to enable that transfer. We then bridge eth over the canonical bridge and re-wrap on the other side.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
chrismaree: this is not redundant code as the l2Token is set to l2Weth, not l2Eth. we still need to enable users to send L2Weth and therefore we store the l2Token as the weth deployed on L2.

This also makes the implementation behave the same as other chains (arbitrum) wherein the l2Token(for the weth pool) is set to WETH. 

This extra logic is to get the canonical bridge to take _eth_ over. 

in your other message you said setting this:
https://github.com/UMAprotocol/protocol/pull/3611#discussion_r750524744 That wont work as we need to send _eth_ over the bridge, not weth. 

Put differently we need to do two things on the L2 side:
1) we need to accommodate the ability to have users deposit WETH or ETH on L2.
2) we need to bridge over the canonical bridge using ETH.

to accommodate this we need to store both of these addresses.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
mrice32: I'm a little confused with this block -- why do we need to _check_ if the l1Token is WETH for this contract? Shouldn't that always be the case? Or am I missing something?

mrice32 (commented):
This looks awesome and the unwrapping is super simple!

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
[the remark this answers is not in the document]
nicholaspai: Agreed

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
```diff
@@ -0,0 +1,44 @@
...
+        bridgePool = newBridgePool;
+    }
+
+    receive() external payable {
+        wrapAndTransfer();
+    }
+
+    fallback() external payable {
+        wrapAndTransfer();
+    }
```
mrice32, line 38: Do we need to define both of these? Any reason not to _only_ define the payable fallback function?

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
chrismaree: This deposit box will be able to receive any erc20, just like the current one (we only have one deposit box per chain). 

Iff it's eth, then we need to change the recipient to be the unwrapped and sent eth over the bridge

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
chrismaree: True. It'll likely work with just the one. However, if there is data from the transfer then we need the other one.

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
nicholaspai: Ahh this is really tricky. I think you should add a comment explaining the subtle difference between L2Weth and L2Eth

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
nicholaspai: Ah can you add a comment explaining why these functions are neccessary since we call `WETH.withdraw()` in `bridgeTokens`?

nicholaspai (commented):
I've added unit tests for new contracts PTAL. One contract change I added is `ChangedBridgePool` event emitted in `Optimism_Wrapper`.

My only other comment is to add comments explaining:
- The subtle dIfference between l2Eth and l2Weth in `OVM_OETH_BridgeDepositBox`
- Why `payable` methods are required in `OVM_OETH_BridgeDepositBox`
- Natspec comments in `Optimism_Wrapper`

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
chrismaree: `receive() external payable` — for empty calldata (and any value)
`fallback() external payable` — when no other function matches (not even the receive function). Optionally payable.

Having both means that if the bridge includes data (it can do this) we'll still capture the eth. This means that no matter how funds are transferred into the contract we'll wrap them and send them to the pool.

chrismaree:
@nicholaspai https://github.com/UMAprotocol/protocol/pull/3611#pullrequestreview-807701520

I've added comments like you requested :)

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
nicholaspai: So for optimism in production, this will be THE only deposit box right? It just has additional logic to handle eth?

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
nicholaspai: Can you add these as comments

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,85 @@
...
+    // l2Eth is ETH on Optimism. This acts as both an ERC20 and ETH and is @ 0xdeaddeaddeaddeaddeaddeaddeaddeaddead0000.
+    address public l2Eth;
+    address public l1EthWrapper;
+
+    /**
+     * @notice Construct the Optimism Bridge Deposit Box
+     * @param _crossDomainAdmin Address of the L1 contract that can call admin functions on this contract from L1.
+     * @param _minimumBridgingDelay Minimum second that must elapse between L2->L1 token transfer to prevent dos.
+     * @param _chainId L2 Chain identifier this deposit box is deployed on.
+     * @param _l1Weth Address of Weth on L1. Used to inform if the deposit should wrap ETH to WETH, if deposit is ETH.
```
nicholaspai, line 21:
```suggestion
     * @param _l1Weth Address of Weth on L1. Used to inform if the deposit should wrap ETH to WETH, if deposit is for a whitelisted token mapped to this L1 Weth token.
```

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,85 @@
...
+     * @param _l2Eth Address of ETH on L2. Used to inform the token that should be sent over the Optimism bridge.
```
nicholaspai, line 22:
```suggestion
     * @param _l2Eth Address of ETH on L2. If L2 Weth is deposited to this contract, then L2 ETH should be sent over the Optimism bridge.
```

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,85 @@
...
+     * @param _l1EthWrapper Address of custom ETH wrapper on L1. Any ETH sent to this contract will be wrapped to WETH
+     *     and sent to the WETH Bridge Pool.
+     * @param _l1Weth Address of Weth on L1. Used to inform if the deposit should wrap ETH to WETH, if deposit is ETH.
```
nicholaspai, line 25:
```suggestion
```

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,85 @@
...
+        uint256 bridgeDepositBoxBalance = TokenLike(l2Token).balanceOf(address(this));
+        require(bridgeDepositBoxBalance > 0, "can't bridge zero tokens");
+        require(canBridge(l2Token), "non-whitelisted token or last bridge too recent");
+
+        whitelistedTokens[l2Token].lastBridgeTime = uint64(getCurrentTime());
+
+        address bridgePool = whitelistedTokens[l2Token].l1BridgePool;
+
+        // If the l1Token mapping to the l2Token is l1Weth, then to work with the canonical optimism bridge, we first
+        //  unwrap it to ETH then bridge ETH over the canonical bridge. On L1 the l1EthWrapper will re-wrap the ETH to
```
nicholaspai, line 61:
```suggestion
        //  unwrap it to ETH then bridge the newly unwraped L2 ETH over the canonical bridge. On L1 the l1EthWrapper will re-wrap the ETH to
```

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
```diff
@@ -0,0 +1,85 @@
...
+
+        emit TokensBridged(l2Token, bridgeDepositBoxBalance, l1Gas, msg.sender);
+    }
+
+    // This contract contains the receive and fallback methods to enable the contract to receive ETH sent to it via the
+    // weth unwrapping done above. When l2ETH is unwrapped from l2WETH, the l2ETH is sent to this contract before being
+    // sent over the canonical Optimims bridge.
+    receive() external payable {}
+
+    fallback() external payable {}
```
nicholaspai, line 84: Can you add a comment explaining the intended usage of `fallback` vs `receive`?

packages/core/contracts/insured-bridge/ovm/OVM_OETH_BridgeDepositBox.sol
chrismaree: correct and correct.

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
chrismaree: done!

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -2,7 +2,7 @@
 pragma solidity ^0.8.0;
 
 // This should be replaced with a "real" import when Optimism release their new contract versions.
-import "../../external/ovm/OVM_CrossDomainEnabled.sol";
```
chrismaree, line 5: note this PR also made the following change: rather than using our own `CrossDomainEnabled` file (that was copied directly from Optimism) we now import from their package.

packages/core/contracts/insured-bridge/ovm/OVM_BridgeDepositBox.sol
```diff
@@ -2,44 +2,16 @@
...
 
 import "../BridgeDepositBox.sol";
-
-import "../../external/ovm/OVM_CrossDomainEnabled.sol";
-
-/**
- * @title Lib_PredeployAddresses
- * @dev Copied from https://github.com/ethereum-optimism/optimism/blob/b6fa9e4085da581fb47993b9c895fa20a631897d/packages/contracts/contracts/libraries/constants/Lib_PredeployAddresses.sol
- */
-library Lib_PredeployAddresses {
```
chrismaree, line 12: we no longer need to define this lib like this and can pull these constants directly from their package.

packages/core/networks/1.json
```diff
@@ -123,6 +123,14 @@
     "contractName": "Arbitrum_Messenger",
     "address": "0x0d8563d3b8b3fec542aae50a530c606dac4d1c4f"
   },
+  {
```
chrismaree, line 126: added the mainnet deployment addresses

packages/core/package.json
```diff
@@ -13,7 +13,7 @@
   },
   "devDependencies": {
     "@awaitjs/express": "^0.3.0",
-    "@eth-optimism/contracts": "^0.4.13",
+    "@eth-optimism/contracts": "^0.5.2",
```
chrismaree, line 16: this version bump enables us to pull the compatible contracts directly from their exports.

packages/core/test/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.js
```diff
@@ -32,10 +32,10 @@ describe("Optimism_ChildMessenger", function () {
     const accounts = await hre.web3.eth.getAccounts();
     [l1Owner, controlledEOA, parentMessenger, rando] = accounts;
 
-    l2CrossDomainMessengerMock = await deployContractMock("OVM_L2CrossDomainMessenger", {
```
chrismaree, line 35: and of course, all these were re-named in their official package. updating versions required us to change this.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -11,7 +11,7 @@ import "../interfaces/ChildMessengerConsumerInterface.sol";
  * @dev This contract is ownable via the onlyCrossDomainAccount modifier, restricting ownership to the cross-domain
  * parent messenger contract that lives on L1.
  */
-contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterface {
+contract Optimism_ChildMessenger is CrossDomainEnabled, ChildMessengerInterface {
```
nicholaspai, line 14: Does this mean we can remove the `external` contracts for OVM now?

nicholaspai (approved):
Awesome to see we can delete some `external/ovm` contracts.

packages/core/contracts/insured-bridge/ovm/Optimism_Wrapper.sol
mrice32: @chrismaree the docs seem to say that a payable fallback function can handle _all_ cases (incoming ETH or calldata): https://docs.soliditylang.org/en/v0.8.10/contracts.html?highlight=fallback#fallback-function. I don't think we need both. Not a huge issue, though.

mrice32 (approved):
Looks good -- only one minor comment: https://github.com/UMAprotocol/protocol/pull/3611#discussion_r752262028.

PR #3615, opened 2021-11-16
improve(insured-bridge-price-feed): Use constant RATE_MODEL from @uma/sdk package

nicholaspai:
**Motivation**

We should reduce instances where we hardcode this JSON object, and read from a single source.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -6,6 +6,7 @@
     "@ethersproject/bignumber": "^5.4.2",
     "@google-cloud/logging-winston": "^4.0.4",
     "@google-cloud/trace-agent": "^4.2.5",
+    "@uma/sdk": "^0.12.3",
     "@uma/common": "^2.13.1",
     "@uma/contracts-node": "^0.1.11",
     "@uniswap/sdk": "^2.0.5",
--- a/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
+++ b/packages/financial-templates-lib/src/price-feed/DefaultPriceFeedConfigs.ts
@@ -1,4 +1,5 @@
 import { getPrecisionForIdentifier } from "@uma/common";
+import { across } from "@uma/sdk";
 
 // Default price feed configs for currently approved identifiers.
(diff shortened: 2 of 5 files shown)


nicholaspai:
@chrismaree do you think we should also import this constants file as a fallback value to `RATE_MODEL` in the `insured-bridge-relayer/RelayerConfig.ts` file?

mrice32:
> @chrismaree do you think we should also import this constants file as a fallback value to `RATE_MODEL` in the `insured-bridge-relayer/RelayerConfig.ts` file?

@nicholaspai I think so, and I think ideally, we should usually be running the bots without overriding it. Thoughts?

mrice32 (approved):
LGTM!

nicholaspai:
> > @chrismaree do you think we should also import this constants file as a fallback value to `RATE_MODEL` in the `insured-bridge-relayer/RelayerConfig.ts` file?
> 
> @nicholaspai I think so, and I think ideally, we should usually be running the bots without overriding it. Thoughts?

I'm good with that, my only concern is that currently updating the bot via a config file means that the bot can update very quickly. For example, if a new L1 token is causing the bot problems, then it can be removed instantly. 

However, requiring a change to pass CI takes longer so we'll have slower iteration speed. Thoughts?

chrismaree:
> > > @chrismaree do you think we should also import this constants file as a fallback value to `RATE_MODEL` in the `insured-bridge-relayer/RelayerConfig.ts` file?
> > 
> > 
> > @nicholaspai I think so, and I think ideally, we should usually be running the bots without overriding it. Thoughts?
> 
> I'm good with that, my only concern is that currently updating the bot via a config file means that the bot can update very quickly. For example, if a new L1 token is causing the bot problems, then it can be removed instantly.
> 
> However, requiring a change to pass CI takes longer so we'll have slower iteration speed. Thoughts?

I do think we should pull from SDK for the bots. it should be configured such that this is the _default_ but we can always append to the config enabling us to move fast on the bots without updating the SDK. i.e best of both worlds

mrice32 (approved):
LGTM! Agree w this plan!

PR #3618, opened 2021-11-17
feat(cross-chain-oracle): OracleHub should send message to child once per price request per child chain ID

nicholaspai:
**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/cross-chain-oracle/OracleHub.sol
```diff
@@ -76,10 +80,13 @@ contract OracleHub is OracleBase, ParentMessengerConsumerInterface, Ownable, Loc
...
         _publishPrice(identifier, time, ancillaryData, price);
 
-        // TODO: Consider storing all publishPrice events for each chainId, therefore we can limit the
-        // sendMessageToChild calls to one per chainId and not allow users to spam the bridge for this chainID
-        // with calls coming from this contract.
-        messengers[chainId].sendMessageToChild(abi.encode(identifier, time, ancillaryData, price));
+        // Limit the `sendMessageToChild` calls to one per unique price request per chainId and prevent users from
+        // spamming the bridges calls coming from this messenger.
+        bytes32 priceRequestKey = _encodePriceRequest(identifier, time, ancillaryData);
+        if (!hasPublishedPriceForChain[chainId][priceRequestKey]) {
```
nicholaspai, line 86: The tradeoff to consider:
- By allowing unlimited calls to `sendMessageToChild`, we can re-send messages that don't cross the bridge correctly. For example, in the `Optimism_Messenger` if the `defaultGasLimit` is set improperly.
- By allowing unlimited calls to `sendMessageToChild`, anyone can get this method to send unlimited calls over the native bridge, potentially opening this contract up to being censored by the native bridge.

Thoughts?

chrismaree: I think an unlimited call is much much better. we wont get censored by the bridge. if we do we can just deploy new bridging contracts. I'm very confident we wont get censored. 

The ability to re-send is critical if something goes wrong in the bridging.

mrice32: @chrismaree I'm not so sure. We try to upgrade to a new messenger, but we can't upgrade the L2 contracts to point to the new L2 messenger because we can't get messages to our governor contract on L2 because our old messenger is censored. I think if our messenger contract is censored, our entire L2 system is toast.

That being said, if censorship is done by a centralized party (I assume it is), we could probably ask them to let _this one upgrade message_ through. Thoughts?

mrice32 (commented):
One comment on the unlimited resubmit thread.

packages/core/contracts/cross-chain-oracle/OracleHub.sol
chrismaree: ye...but this feels like over engineering. we dont worry about our bridge contracts getting censored and they send tones of stuff. why would censorship even happen here? we are paying gas and so would the caller of the bridge and so no problems. it's not like you can dos the bridge without being smashed with huge cost.

imho this PR solves a problem that we dont think will occur and therefore should be avoided as it addeds extra complexity in a situation that won't arise in permission less systems that contain cost of sending transactions (i.e tx fees _are_ dos protection in and of themselves)

nicholaspai: I think I lean towards @chrismaree opinion actually, I highly doubt we get censored via griefing, and if someone is attempting it we'll know and can talk with the bridge operator

PR #3619, opened 2021-11-17
fix(sdk): do not pre-populate transactions, allow overrides

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3217/trezor-trezor-one-hw-wallets-cannot-add-liquidity-on-across-pool

**Summary**

This removes  signer.populateTransaction from the transaction manager. This actually caused 2 major problems. 

1.  it assumed signers were eip-1559 compatible, causing errors when sent to metamask with hardware wallets that are not compatible.
2.  it set gas estimates that might be different from what the app was estimating during a "max eth send" event, causing the transaction to fail validation before being sent to metamask.  

Now all unset fields will be set by metamask, which has more control and information about the wallet environment.

Because we remove populateTransaction, we allow manual overrides in case you want any of these fields set for any reason. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/3217/trezor-trezor-one-hw-wallets-cannot-add-liquidity-on-across-pool

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -1,21 +1,21 @@
 import assert from "assert";
 import { bridgePool } from "../../clients";
-import { BigNumber, Signer } from "ethers";
 import { toBNWei, fixedPointAdjustment, calcPeriodicCompoundInterest, calcApr, BigNumberish } from "../utils";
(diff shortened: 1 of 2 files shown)


mrice32 (approved):
LGTM! +1

amateima (approved):

PR #3620, opened 2021-11-17
feat(cross-chain-oracle): Add Polygon adapters

nicholaspai:
**Motivation**

These adapters re use contracts from `external/polygon` that currently enable the cross chain contracts in `contracts/polygon` that only work for Polygon.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
+++ b/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
@@ -100,7 +100,6 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      */
     function processMessageFromCrossChainParent(bytes memory data, address target)
         public
-        override
         onlyFromCrossDomainAccount(parentMessenger)
         nonReentrant()
     {
--- a/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
+++ b/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
@@ -28,8 +28,8 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
     event SetOracleSpoke(address newOracleSpoke);
     event SetParentMessenger(address newParentMessenger);
     event SetDefaultGasLimit(uint32 newDefaultGasLimit);
-    event MessageSentToParent(bytes data, address indexed parentAddress, uint32 gasLimit);
-    event MessageReceivedFromParent(bytes data, address indexed parentAddress);
+    event MessageSentToParent(bytes data, address indexed targetHub, uint32 gasLimit);
+    event MessageReceivedFromParent(bytes data, address indexed targetSpoke);
 
     /**
      * @notice Construct the Optimism_ChildMessenger contract.
--- a/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
+++ b/packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
@@ -5,14 +5,15 @@ pragma solidity ^0.8.0;
 import "../../external/ovm/OVM_CrossDomainEnabled.sol";
 import "../interfaces/ChildMessengerInterface.sol";
 import "../interfaces/ChildMessengerConsumerInterface.sol";
+import "../../common/implementation/Lockable.sol";
 
 /**
  * @notice Sends cross chain messages from Optimism L2 to Ethereum L1 network.
  * @dev This contract is ownable via the onlyCrossDomainAccount modifier, restricting ownership to the cross-domain
  * parent messenger contract that lives on L1.
  */
-contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterface {
-    // The only L2 contract that can send messages over the bridge via the messenger is the oracle spoke.
+contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterface, Lockable {
+    // The only child network contract that can send messages over the bridge via the messenger is the oracle spoke.
     address public oracleSpoke;
 
     // Messenger contract on the other side of the L1<->L2 bridge.
@@ -43,7 +44,7 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      * @dev The caller of this function must be the parent messenger, over the canonical bridge.
      * @param newOracleSpoke address of the new oracle spoke, deployed on L2.
      */
-    function setOracleSpoke(address newOracleSpoke) public onlyFromCrossDomainAccount(parentMessenger) {
+    function setOracleSpoke(address newOracleSpoke) public onlyFromCrossDomainAccount(parentMessenger) nonReentrant() {
         oracleSpoke = newOracleSpoke;
         emit SetOracleSpoke(newOracleSpoke);
     }
@@ -53,7 +54,11 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      * @dev The caller of this function must be the parent messenger, over the canonical bridge.
      * @param newParentMessenger address of the new parent messenger, deployed on L1.
      */
-    function setParentMessenger(address newParentMessenger) public onlyFromCrossDomainAccount(parentMessenger) {
+    function setParentMessenger(address newParentMessenger)
+        public
+        onlyFromCrossDomainAccount(parentMessenger)
+        nonReentrant()
+    {
         parentMessenger = newParentMessenger;
         emit SetParentMessenger(newParentMessenger);
     }
@@ -63,7 +68,11 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      * @dev The caller of this function must be the parent messenger, over the canonical bridge.
      * @param newDefaultGasLimit the new L1 gas limit to be set.
      */
-    function setDefaultGasLimit(uint32 newDefaultGasLimit) public onlyFromCrossDomainAccount(parentMessenger) {
+    function setDefaultGasLimit(uint32 newDefaultGasLimit)
+        public
+        onlyFromCrossDomainAccount(parentMessenger)
+        nonReentrant()
+    {
         defaultGasLimit = newDefaultGasLimit;
         emit SetDefaultGasLimit(newDefaultGasLimit);
     }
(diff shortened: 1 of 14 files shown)


packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -53,7 +54,11 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      * @dev The caller of this function must be the parent messenger, over the canonical bridge.
      * @param newParentMessenger address of the new parent messenger, deployed on L1.
      */
-    function setParentMessenger(address newParentMessenger) public onlyFromCrossDomainAccount(parentMessenger) {
+    function setParentMessenger(address newParentMessenger)
+        public
+        onlyFromCrossDomainAccount(parentMessenger)
+        nonReentrant()
```
chrismaree, line 60: I think these `nonReentrant` modifies are redundant (and potentially dangerous) for us to user here. these methods are access controlled via the `onlyFromCrossDomainAccount` modifier so why do we need this entry guard?

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ChildMessenger.sol
```diff
@@ -0,0 +1,86 @@
...
+import "../../common/implementation/Lockable.sol";
+
+/**
+ * @notice Sends cross chain messages from Polygon to Ethereum network.
+ * @dev This contract extends the `FxBaseChildTunnel` contract and therefore is 1-to-1 mapped with the
+ * `FxBaseRootTunnel` extended by the `Polygon_ParentMessenger` contract deployed on Polygon. This mapping ensures that
+ * the internal `_processMessageFromRoot` function is only callable indirectly by the `Polygon_ParentMessenger`.
+ */
+contract Polygon_ChildMessenger is FxBaseChildTunnel, ChildMessengerInterface, Lockable {
+    // The only child network contract that can send messages over the bridge via the messenger is the oracle spoke.
```
chrismaree, line 16:
```suggestion
    // The only child network contract that can send messages over the bridge via the messenger is the Oracle Spoke.
```

chrismaree: same nit throughout

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ChildMessenger.sol
```diff
@@ -0,0 +1,86 @@
...
+    /**
+     * @notice Sends a message to the OracleSpoke via the parent messenger and the canonical message bridge.
+     * @dev The caller must be the OracleSpoke on child network. No other contract is permissioned to call this
+     * function.
+     * @dev The L1 target, the parent messenger, must implement processMessageFromChild to consume the message.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override nonReentrant() {
+        require(msg.sender == oracleSpoke, "Only callable by oracleSpoke");
+        _sendMessageToRoot(abi.encode(data, oracleHub));
```
chrismaree, line 65: hhmmm. is this correct? it might be an artefact of polygon that I'm not fully understanding. why are you not calling the call data something akin to the optimism one like this:
```
bytes memory dataSentToParent = abi.encodeWithSignature("processMessageFromCrossChainChild(bytes)", data);
```
where we encode the function that is going to be called on the `ParentMessenger`

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ChildMessenger.sol
```diff
@@ -0,0 +1,86 @@
...
+     * and Polygon. This will revert if the Root chain sender is not the `fxRootTunnel` contract.
+     * @dev This call will revert if `setFxRoot` has not been called. Fx Child should be set to Polygon_ParentMessenger.
+     * @param sender The sender of `data` from the Root chain.
+     * @param data ABI encoded params with which to call function on OracleHub or GovernorHub.
+     */
+    function _processMessageFromRoot(
+        uint256, /* stateId */
+        address sender,
+        bytes memory data
+    ) internal override validateSender(sender) nonReentrant() {
```
chrismaree, line 81: where is this `validateSender` coming from? how does that work? a comment might be needed as I cant find this in the PR anywhere

packages/core/contracts/cross-chain-oracle/chain-adapters/test/Polygon_ChildMessengerMock.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../Polygon_ChildMessenger.sol";
+
+contract Polygon_ChildMessengerMock is Polygon_ChildMessenger {
```
chrismaree, line 6: kinda lame that you need extra mocks for polygon.

packages/core/contracts/cross-chain-oracle/interfaces/ChildMessengerInterface.sol
```diff
@@ -4,8 +4,4 @@ pragma solidity ^0.8.0;
 interface ChildMessengerInterface {
     // Should send cross-chain message to Parent messenger contract or revert.
     function sendMessageToParent(bytes memory data) external;
-
-    // Should be called by parent messenger over the canonical bridge. Includes a target as the L2 contract that should
-    // have data passed to it.
-    function processMessageFromCrossChainParent(bytes memory data, address target) external;
```
chrismaree, line 10: why did you remove this?

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ParentMessenger.sol
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * @notice Process a received message from the child messenger via the canonical message bridge.
+     * @dev This internal method will be called inside `FxBaseRootTunnel.receiveMessage(bytes memory inputData)`.
+     * The `inputData` is a proof of transaction that is derived from the transaction hash of the transaction on the
+     * child chain that originated the cross-chain price request via _sendMessageToRoot.
+     * @dev This call will revert if `setFxChild` has not been called. Fx Child should be set to Polygon_ChildMessenger.
+     * @param data ABI encoded params with which to call function on OracleHub or GovernorHub.
+     */
+    function _processMessageFromChild(bytes memory data) internal override nonReentrant() {
```
chrismaree, line 57: why is this internal? how does the polygon Fxtunnel work with this to know to call this function?

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
```diff
@@ -91,10 +100,10 @@ contract Optimism_ChildMessenger is OVM_CrossDomainEnabled, ChildMessengerInterf
      */
     function processMessageFromCrossChainParent(bytes memory data, address target)
         public
-        override
```
chrismaree, line 94: why is this no longer override? do we not want this to be the standard interface for inbound messages from the parent -> child?

chrismaree: same question on the parent in the following file.

packages/core/networks/5.json
```diff
@@ -22,5 +22,21 @@
   {
     "contractName": "MockOracleAncillary",
     "address": "0x20570e9e27920ac5e2601e0bef7244deff7f0b28"
+  },
+  {
+    "contractName": "ExpandedERC20",
```
chrismaree, line 27: why do we need this?

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
nicholaspai: I don't see that they're dangerous, perhaps they're overkill, but why would reentrancy ever happen?

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ChildMessenger.sol
nicholaspai: Yeah this is polygon specific. When sending a message from Polygon to Ethereum, the Polygon contract must emit `MessageSent`. Then, the user can manually relay the message when the block containing the transaction with the `MessageSent` event is checkpointed to Mainnet.

To manually relay the message, the user must call `receiveMessage` on the `Polygon_ParentMessenger` or whatever the `FxBaseRootTunnel` that is connected to this `FxBaseChildTunnel`. `receiveMessage` will then call the internal `_processMessageFromChild` method. This is why we don't need to use the `abi.encodeWithSignature` because we know already which internal method on the `ParentMessenger` will be called.

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ChildMessenger.sol
nicholaspai: Its in `external/polygon/FxBaseChildTunnel`.  I edited the comments.

packages/core/contracts/cross-chain-oracle/chain-adapters/test/Polygon_ChildMessengerMock.sol
nicholaspai: That's only because there are internal methods to test

packages/core/contracts/cross-chain-oracle/interfaces/ChildMessengerInterface.sol
nicholaspai: Because we do not need it in `Polygon` messengers, and I see no harm in adding it to the Optimism adapters

packages/core/contracts/cross-chain-oracle/chain-adapters/Polygon_ParentMessenger.sol
nicholaspai: This is how Polygon works. I tried to explain in comments can you let me know what could be changed about this language>

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ChildMessenger.sol
nicholaspai: I don't think this is standard for all networks. Only Optimism so far, and Arbitrum will work the same. But Polygon works differently so I don't think it should go in the interface

packages/core/networks/5.json
nicholaspai: We don't I'll remove

chrismaree (approved):

PR #3621, opened 2021-11-17
fix(sdk): pass empty object instead of undefined to tx calls to fix c…

daywiss:
…all errors

**Motivation**

https://app.shortcut.com/uma-project/story/3240/hotfix-sdk-transaction-calls-fail-when-passing-undefined-overrides

**Summary**

Ethers seems to count the number of arguments to function calls when you make transactions, it fails if there is a mismatch. For some reason it considers undefined where the overrides should be as an argument and will fail. This changes it to pass an empty object by default, which for some reason ethers detects as overrides and allows the call to be made without error.

Tested this on across for add/remove liquidity calls for tokens/eth.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.shortcut.com/uma-project/story/3240/hotfix-sdk-transaction-calls-fail-when-passing-undefined-overrides

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -349,7 +349,7 @@ export class Client {
     this.transactionManagers[address] = txman;
     return txman;
   }
-  async addEthLiquidity(signer: Signer, pool: string, l1TokenAmount: BigNumberish, overrides?: Overrides) {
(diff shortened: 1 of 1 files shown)


mrice32 (approved):
LGTM!

amateima (approved):

PR #3622, opened 2021-11-17
feat(across): Add Setter for LP Fee %/second in BridgePool

nicholaspai:
**Motivation**

This rate is not easy to set and forget and should be configurable by Bridge admin.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/insured-bridge/BridgeAdmin.sol
+++ b/packages/core/contracts/insured-bridge/BridgeAdmin.sol
@@ -137,9 +137,9 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
      * @param bridgePool Bridge Pool to change LP fee rate for.
      * @param newLpFeeRate The new rate to set for the `bridgePool`.
      */
-    function changeLpFeeRatePerSecond(address bridgePool, uint64 newLpFeeRate) public onlyOwner nonReentrant() {
-        BridgePoolInterface(bridgePool).changeLpFeeRatePerSecond(newLpFeeRate);
-        emit ChangedLpFeeRate(bridgePool, newLpFeeRate);
+    function setLpFeeRatePerSecond(address bridgePool, uint64 newLpFeeRate) public onlyOwner nonReentrant() {
+        BridgePoolInterface(bridgePool).setLpFeeRatePerSecond(newLpFeeRate);
+        emit SetLpFeeRate(bridgePool, newLpFeeRate);
     }
 
     /**************************************************
--- a/packages/core/contracts/insured-bridge/BridgeAdmin.sol
+++ b/packages/core/contracts/insured-bridge/BridgeAdmin.sol
@@ -130,6 +130,11 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
         emit BridgePoolsAdminTransferred(bridgePools, newAdmin);
     }
 
+    function changeLpFeeRatePerSecond(address bridgePool, uint64 newLpFeeRate) public onlyOwner nonReentrant() {
+        BridgePoolInterface(bridgePool).changeLpFeeRatePerSecond(newLpFeeRate);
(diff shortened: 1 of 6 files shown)


packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -664,6 +667,17 @@ contract BridgePool is MultiCaller, Testable, BridgePoolInterface, ERC20, Lockab
         emit BridgePoolAdminTransferred(msg.sender, _newAdmin);
     }
 
+    /**
+     * @notice Enable the bridge admin to change the decay rate at which LP shares accumulate fees. The higher this
+     * value, the faster LP shares realize pending fees.
+     * @param _newLpFeeRatePerSecond The new rate to set.
+     */
+    function changeLpFeeRatePerSecond(uint64 _newLpFeeRatePerSecond) public override nonReentrant() {
+        require(msg.sender == address(bridgeAdmin));
```
chrismaree, line 676: nit: should we make this a modifier?
nit: this should have a require message?
nit: should the natspec say only the admin can call this?

nicholaspai: > nit: should we make this a modifier?

yeah, slightly improves readability let's do that.

> nit: this should have a require message?

It definitely should have a require message, will add.

> nit: should the natspec say only the admin can call this?

Will add

chrismaree (approved):
LGTM!

packages/core/contracts/insured-bridge/BridgePool.sol
```diff
@@ -656,14 +664,25 @@ contract BridgePool is MultiCaller, Testable, BridgePoolInterface, ERC20, Lockab
...
 
+    /**
+     * @notice Enable the bridge admin to change the decay rate at which LP shares accumulate fees. The higher this
+     * value, the faster LP shares realize pending fees.
+     * @dev Caller must be BridgeAdmin contract.
+     * @param _newLpFeeRatePerSecond The new rate to set.
+     */
+    function changeLpFeeRatePerSecond(uint64 _newLpFeeRatePerSecond) public override onlyBridgeAdmin() nonReentrant() {
+        lpFeeRatePerSecond = _newLpFeeRatePerSecond;
+        emit LpFeeRateChanged(lpFeeRatePerSecond);
```
mrice32, line 683: Do you think it'd be useful to include the old LP fee rate in this event?

mrice32 (approved):
LGTM!

packages/core/contracts/insured-bridge/BridgePool.sol
nicholaspai: Not really because all historical LP fee rates are captured by this event, since its also emitted in the constructor.

nicholaspai: Do you?

mrice32: I think it's fine the way it is. When digging through events in the past, I remember wanting this info (for instance, in the store), but I think this is fine, as you can always find the event before if you need to.

mrice32 (approved):
LGTM!

packages/core/contracts/insured-bridge/BridgeAdmin.sol
```diff
@@ -130,6 +130,11 @@ contract BridgeAdmin is BridgeAdminInterface, Ownable, Lockable {
         emit BridgePoolsAdminTransferred(bridgePools, newAdmin);
     }
 
+    function changeLpFeeRatePerSecond(address bridgePool, uint64 newLpFeeRate) public onlyOwner nonReentrant() {
```
chrismaree, line 133: nit: NatSpec

nicholaspai: Changed

PR #3626, opened 2021-11-18
improve(cross-chain-oracle): Allow publishPrice to be called for any DVM resolved price

nicholaspai:
**Motivation**

`OracleBase.publishPrice` can currently only be called after a price is requested, which usually comes via a cross-chain price request. This is unnecessarily restrictive as the `OracleSpokes` should be able to make all prices available that have been resolved on the DVM, even if they weren't originally requested via the `OracleSpoke` on the child network.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/OracleBase.sol
+++ b/packages/core/contracts/cross-chain-oracle/OracleBase.sol
@@ -42,6 +42,8 @@ abstract contract OracleBase {
     /**
      * @notice Enqueues a request (if a request isn't already present) for the given (identifier, time,
      * ancillary data) combination. Will only emit an event if the request has never been requested.
+     * @return True if price request is new, false otherwise. This is useful for caller to keep track of
+     * duplicate price requests.
      */
     function _requestPrice(
         bytes32 identifier,
@@ -61,7 +63,7 @@ abstract contract OracleBase {
     }
 
     /**
-     * @notice Publishes price for a requested query. Will only emit an event if the request has never been resolved.
+     * @notice Publishes price for a requested query.
      */
     function _publishPrice(
         bytes32 identifier,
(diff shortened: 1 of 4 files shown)


packages/core/contracts/cross-chain-oracle/OracleBase.sol
```diff
@@ -71,11 +73,9 @@ abstract contract OracleBase {
...
         bytes32 priceRequestId = _encodePriceRequest(identifier, time, ancillaryData);
         Price storage lookup = prices[priceRequestId];
-        if (lookup.state == RequestState.Requested) {
-            lookup.price = price;
-            lookup.state = RequestState.Resolved;
-            emit PushedPrice(identifier, time, ancillaryData, lookup.price, priceRequestId);
-        }
+        lookup.price = price;
+        lookup.state = RequestState.Resolved;
+        emit PushedPrice(identifier, time, ancillaryData, lookup.price, priceRequestId);
```
chrismaree, line 78: I think there needs to be a bit more logic here:
1) if the `OracleBase` already has a price it should not change the stored price. 
2) we should only emit an event on the first time the function is called.

you can simply add the following to the top of this function:
```
if (lookup.state == RequestState.Resolved) return;
```

this way you exit early, don't override the price and don't double emit

chrismaree (approved):
This looks great to me! I left one comment about early exit to ensure we: a) don't override the previous price and b) only emit once.

packages/core/contracts/cross-chain-oracle/OracleBase.sol
mrice32: Agree with this comment.

mrice32 (approved):
+1 on this PR since this allows us to resolve prices faster than the L2 -> L1 bridges might allow as long as we're willing to lose the final fee.

LGTM once @chrismaree's comment is resolved.

PR #3627, opened 2021-11-18
improve(cross-chain-oracle): Change OracleSpoke stamped ancillary data contents

nicholaspai:
**Motivation**

Currently, the `OracleSpoke` stamps the child chain's network ID and the caller address (i.e. `msg.sender`) in the ancillary data for all methods `requestPrice`, `hasPrice`, and `getPrice`. This creates a problem where `hasPrice` will never stamp the correct ancillary data for any caller that is NOT the original requester. This behavior is unintended and means that the `OracleSpoke` does not behave like the `Voting` contract for the child chain.

Therefore, the first motivation of this PR is to ensure that `hasPrice` and `getPrice` can be called successfully by any registered contract. This is achieved by removing the requester address from the stamped ancillary data.

The second motivation is to change the contents of the stamped ancillary data. The entire point of stamping ancillary data is to make it easy for human DVM voters on Ethereum mainnet to be able to deterministically track a price request's path and see all of the contracts that it passed through and who originated it. Therefore, its important that the ancillary data contains (1) the child chain ID, otherwise its not obvious which child `OracleSpoke` originated the price request. Secondly, its important to also include the `OracleSpoke`'s address, because this contract can be upgraded over time for each child chain.

This should be all of the information needed to track a price request back to a particular `OracleSpoke`.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
```diff
@@ -163,29 +163,23 @@ contract OracleSpoke is
...
-    function _stampAncillaryData(bytes memory ancillaryData, address requester) internal view returns (bytes memory) {
-        // This contract should stamp its requester's address and network in the
-        // ancillary data so voters can conveniently track the requests path to the DVM.
+    function _stampAncillaryData(bytes memory ancillaryData) internal view returns (bytes memory) {
+        // This contract should stamp the child network's ID and its address so that voters on the parent network can
+        // deterministically track unique price requests back to this contract.
         return
             AncillaryData.appendKeyValueUint(
-                AncillaryData.appendKeyValueAddress(ancillaryData, "childRequester", requester),
+                AncillaryData.appendKeyValueAddress(ancillaryData, "oracleSpoke", address(this)),
```
chrismaree, line 182: excellent! I really like the idea of including this. very useful in the case we want to upgrade like you mentioned.

There is one downside to this which we should consider: if we _were_ to upgrade, you would not be able to bridge prices that were destined for the previous `OracleSpoke` over the bridge to "seed" the new spoke. infact you would be _forced_ to start with a blank slate with no pricing info and all previous prices are effectively lost.

if we exclude this `childRequester` logic, then we can enable arbitrary archeological prices (requested for and consumed by previous `OracleSpokes`) to be re-used.

another solution to this, other than removing this from the ancillary data, would be to enable chained lookbacks in the same way the Voting.sol contract enables previous Voting instances to be queried to enable data preservation in upgrades.

this, however, might not be a problem at the end of the day as most applications use price once then not again (i.e the expiration of an LSP is once off)

wdyt?

mrice32: Would this be fixed by us either 1) rebridging (if that's possible, depending on the outcome of the other PR) or 2) upgrading the hub oracle at the same time, which would allow any necessary price requests to be rebridged due to _both sides_ being a blank slate.

Thoughts?

mrice32 (approved):
LGTM!

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
nicholaspai: > There is one downside to this which we should consider: if we were to upgrade, you would not be able to bridge prices that were destined for the previous OracleSpoke over the bridge to "seed" the new spoke. infact you would be forced to start with a blank slate with no pricing info and all previous prices are effectively lost.

Great point @chrismaree 

I think this could be fixed by rebridging but that's suboptimal as we'd like to not have extra transactions. What about just removing the `oracleSpoke` address? Does just the `childChainId` suffice for ancillary price request information?

chrismaree: but we could not just re-bridge in the current implementation, no? the `oracleSpoke` address will now differ and so we'll not be able match the ancillary data.

if you remove the `oracleSpoke` then re-bridging becomes posible as the `childChainId` matches. I think this is probs the best solution. wdyt? any downsides with removing this?

nicholaspai: I agree and I don't think there are any downsides, let's remove `oracleSpoke` so that upgrades are a bit easier.

mrice32 (approved):
LGTM!

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
mrice32: Agree that adding this to the ancillary data isn't strictly necessary and removing it is worthwhile if it gets around all these complications.

PR #3628, opened 2021-11-19
feat(cross-chain-oracle): Add Arbitrum adapters

nicholaspai:
**Motivation**

Child and Parent messengers are implemented to bridge governance and oracle messages from Arbitrum to Ethereum and back.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
```diff
@@ -0,0 +1,153 @@
...
+        uint32 gasLimit,
+        uint256 gasPrice,
+        uint256 maxSubmissionCost,
+        address indexed childAddress,
+        uint256 sequenceNumber
+    );
+    event MessageReceivedFromChild(bytes data, address indexed childAddress, address indexed targetHub);
+
+    // TODO: Try to read these from Arbitrum system contracts.
+    uint32 public defaultGasLimit = 5_000_000;
```
nicholaspai, line 36: We should look into how we can dynamically read this but i wanted to test the contracts for now on test net

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -47,7 +47,7 @@ contract Optimism_ParentMessenger is OVM_CrossDomainEnabled, ParentMessengerInte
     function setChildOracleSpoke(address newOracleSpoke) public onlyOwner {
         bytes memory dataSentToChild = abi.encodeWithSignature("setOracleSpoke(address)", newOracleSpoke);
         sendCrossDomainMessage(childMessenger, defaultGasLimit, dataSentToChild);
-        emit MessageSentToChild(dataSentToChild, childMessenger, defaultGasLimit);
+        emit MessageSentToChild(dataSentToChild, childMessenger, defaultGasLimit, childMessenger);
```
nicholaspai, line 50: These events are breaking CI

packages/core/networks/4.json
```diff
@@ -9,7 +9,7 @@
   },
   {
     "contractName": "IdentifierWhitelist",
-    "address": "0xCF744280D6ee7ca04e6D4167546a8463Ad899105"
+    "address": "0x7242892cc4653f1214cf700ddcc50ad47ebff564"
```
nicholaspai, line 12: THis is the ID whitelist hardcoded in the `MockOracleAncillary` that I use in tests, so I switched Finder back to it.

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
```diff
@@ -49,9 +52,11 @@ contract GovernorHub is Ownable, Lockable {
         uint256 chainId,
         address to,
         bytes memory dataFromGovernor
-    ) external nonReentrant() onlyOwner {
+    ) external payable nonReentrant() onlyOwner {
         bytes memory dataSentToChild = abi.encode(to, dataFromGovernor);
-        messengers[chainId].sendMessageToChild(dataSentToChild);
+
+        // Pass all msg.value to Messenger
+        messengers[chainId].sendMessageToChild{ value: msg.value }(dataSentToChild);
```
chrismaree, line 59: sheesh. ok this is painful. there is no compatibility issues with the `Governor.sol` sending transactions that contain a `value`, right? it can support this in its current form?

chrismaree: Indeed the `Governor.sol`'s method includes associated `value` field to enable this: https://github.com/UMAprotocol/protocol/blob/a4e395f4a714f54f042078baac3993d339641ba6/packages/core/contracts/oracle/implementation/Governor.sol#L190

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ChildMessenger.sol
```diff
@@ -0,0 +1,87 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "../interfaces/ChildMessengerInterface.sol";
+import "../interfaces/ChildMessengerConsumerInterface.sol";
+import "../../common/implementation/Lockable.sol";
+import "../../external/avm/AVM_CrossDomainEnabled.sol";
+
+/**
+ * @notice Sends cross chain messages from Arbitrum L2 to Ethereum L1 network.
```
chrismaree, line 10:
```suggestion
 * @notice Sends and receives cross chain messages between Arbitrum L2 and Ethereum L1 network.
```

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
chrismaree: is there a solution where rather than the `value` coming from the caller it is some value that is pre-loaded into the `GovernorHub` to pay for gas on L2? this way, we dont need to make this function payable and can rather let us drop some funds onto the bridge pool to pay for requests out of the contracts balance?

chrismaree: the big advantage of doing it this way is that the `Governor` now no longer needs to pay for these actions. Moreover, the `Governor` does not need to consider the difference between the Arbitrum and other chain calls as the payment for the cross-chain actions is embedded within _this_ contract rather than the calling contract.

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ChildMessenger.sol
```diff
@@ -0,0 +1,87 @@
...
+    /**
+     * @notice Sends a message to the parent messenger via the canonical message bridge.
+     * @dev The caller must be the OracleSpoke on L2. No other contract is permissioned to call this function.
+     * @dev The L1 target, the parent messenger, must implement processMessageFromChild to consume the message.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override nonReentrant() {
+        require(msg.sender == oracleSpoke, "Only callable by oracleSpoke");
+        bytes memory dataSentToParent = abi.encodeWithSignature("processMessageFromCrossChainChild(bytes)", data);
+        uint256 id = sendCrossDomainMessage(msg.sender, parentMessenger, dataSentToParent);
```
chrismaree, line 67: so when sending from L2->L1 we dont need to send any eth to pay gas? this is only needed when sending from L1->L2?

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
```diff
@@ -0,0 +1,153 @@
...
+    uint256 public defaultMaxSubmissionCost = 1e18 / 10; // 0.1e18
```
chrismaree, line 37:
```suggestion
    uint256 public defaultMaxSubmissionCost = 1e17; // 0.1e18
```

chrismaree: no need to do the devision. you can combine decimals with the `e` syntax

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * @notice Changes the address of the oracle spoke on L2 via the child messenger.
+     * @dev The caller of this function must be the owner. This should be set to the DVM governor.
+     * @param newOracleSpoke the new oracle spoke address set on L2.
+     */
+    function setChildOracleSpoke(address newOracleSpoke) public payable onlyOwner {
+        bytes memory dataSentToChild = abi.encodeWithSignature("setOracleSpoke(address)", newOracleSpoke);
+        _sendMessageToChild(dataSentToChild, childMessenger);
```
chrismaree, line 87: nit: event

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
```diff
@@ -0,0 +1,153 @@
...
+    }
+
+    /**
+     * @notice Changes the address of the parent messenger on L2 via the child messenger.
+     * @dev The caller of this function must be the owner. This should be set to the DVM governor.
+     * @param newParentMessenger the new parent messenger contract to be set on L2.
+     */
+    function setChildParentMessenger(address newParentMessenger) public payable onlyOwner {
+        bytes memory dataSentToChild = abi.encodeWithSignature("setParentMessenger(address)", newParentMessenger);
+        _sendMessageToChild(dataSentToChild, childMessenger);
```
chrismaree, line 97: nit: event

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
chrismaree: nvm you did it in the internal function.

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
```diff
@@ -0,0 +1,153 @@
...
+    {
+        ParentMessengerConsumerInterface(oracleHub).processMessageFromChild(childChainId, data);
+        emit MessageReceivedFromChild(data, childMessenger, oracleHub);
+    }
+
+    function _sendMessageToChild(bytes memory data, address target) internal {
+        uint256 seqNumber =
+            sendTxToL2NoAliassing(
+                childMessenger,
+                owner(), // This is the address that will send ETH refunds for any failed messages.
```
chrismaree, line 135: owner() will be the owner of the `OracleBase`, right? so this, as it's the `ParentMessenger`, will be the `Governor` address, but on L2? this might be suboptimal as we won't be able to re-claim these funds, no?

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
chrismaree: ye, I fixed the same thing on one of my PRs. had the same issue

packages/core/contracts/external/avm/interfaces/ArbSys.sol
```diff
@@ -0,0 +1,79 @@
...
+
+pragma solidity ^0.8.0;
+
+/**
+ * @notice Precompiled contract that exists in every Arbitrum chain at address(100),
+ * 0x0000000000000000000000000000000000000064. Exposes a variety of system-level functionality. ArbSys provides
+ * systems functionality useful to some Arbitrum contracts. Any contract running on an Arbitrum Chain can call the
+ * chain's ArbSys.
+ */
+interface ArbSys {
```
chrismaree, line 12: is there no way to import these directly from the arbitrum package and therefore delete the `avm` directory in full? this is how we do things in the optimism contracts and it makes things clean af.

chrismaree: I'm looking into if these are exported from the `arb-bridge-eth` package that arbitrum publishes. will report back shortly.

chrismaree: damn. this package is private: https://github.com/OffchainLabs/arbitrum/blob/4b0f8d34b73036226d8a7455b5252d96e13bd7e4/packages/arb-bridge-eth/package.json?_pjax=%23js-repo-pjax-container%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%2C%20%5Bdata-pjax-container%5D#L3

packages/core/contracts/cross-chain-oracle/GovernorHub.sol
nicholaspai: Yeah good call out... I'm going to change the `Arbitrum_ParentMessenger` so that it can receive ETH and use its own balance to send to the inbox

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ChildMessenger.sol
nicholaspai: Nope, the User just needs to manually finalize the transaction L1 (and pay some gas) after the challenge period passes: https://developer.offchainlabs.com/docs/l1_l2_messages#l2-to-l1-messages-lifecycle

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
nicholaspai: aha

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ParentMessenger.sol
nicholaspai: I changed the ParentMessenger so that the owner can set a refund address

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -103,4 +103,11 @@ contract Optimism_ParentMessenger is CrossDomainEnabled, ParentMessengerInterfac
         ParentMessengerConsumerInterface(oracleHub).processMessageFromChild(childChainId, data);
         emit MessageReceivedFromChild(data, childMessenger, oracleHub);
     }
+
+    /**
+     * @notice sendMessageToChild never sends any msg.value, so this should always return 0.
+     */
+    function getL1CallValue() external pure override returns (uint256) {
```
chrismaree, line 110: can you add this function to the `ParentMessengerBase` such that it defaults to 0 in all cases except when overrides, such as in the `Arbitrum_ParentMessenger.sol`

chrismaree: In fact, can we not move a bunch of the `setChild*` functions into the `Base`?

packages/core/contracts/cross-chain-oracle/chain-adapters/Arbitrum_ChildMessenger.sol
chrismaree: ok great, yup. same as Optimism.

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
```diff
@@ -58,7 +58,7 @@ contract Optimism_ParentMessenger is CrossDomainEnabled, ParentMessengerInterfac
     function setChildParentMessenger(address newParentMessenger) public onlyOwner {
         bytes memory dataSentToChild = abi.encodeWithSignature("setParentMessenger(address)", newParentMessenger);
         sendCrossDomainMessage(childMessenger, defaultGasLimit, dataSentToChild);
-        emit MessageSentToChild(dataSentToChild, address(0), defaultGasLimit, childMessenger);
+        emit MessageSentToChild(dataSentToChild, childMessenger, defaultGasLimit, childMessenger);
```
chrismaree, line 61: I intentionally set this to the `address(0)` btw. reason is this prop is the `targetSpoke`. in this case, there is no `targetSpoke` as this is changing the `childMessenger`. does it not make more sence to be blank (i.e 0x0) to indicate there is no _target spoke_.

It might help to rename the 4th param in `MessageSentToChild` from `childAddress` to `childMessengerAddress`.

nicholaspai: hmm but isn't this more of a `targetContract`, where the final execution happens? Perhaps thats meaningless

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
nicholaspai: We can definitely move `getL1CallValue` but the other ones we cannot because the internal `sendMessage` implementations are different

packages/core/contracts/cross-chain-oracle/chain-adapters/Optimism_ParentMessenger.sol
nicholaspai: I'll revert and rename to `childMessenger`

chrismaree (approved):
This is awesome! Thanks for adding in the abstract class and the associated mock for testing.

PR #3631, opened 2021-11-19
feat: add Admin_ChildMessenger

mrice32:
**Motivation**

We would like to support EVM compatible chains that we haven't built specialized bridge integrations for.

**Summary**

This contract matches the interface of other ChildMessengers, but instead of bridging, it just emits logs on incoming messages and allows an admin to send messages out.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
+++ b/packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
@@ -1,0 +1,60 @@
+// SPDX-License-Identifier: AGPL-3.0-only
+pragma solidity ^0.8.0;
+
+import "@openzeppelin/contracts/access/Ownable.sol";
+import "../interfaces/ChildMessengerInterface.sol";
+import "../interfaces/ChildMessengerConsumerInterface.sol";
+import "../../common/implementation/Lockable.sol";
+
+/**
+ * @notice A version of the child messenger that allows an admin to relay messages on its behalf.
+ * @dev No parent messenger is needed for this case, as the admin could be trusted to manually send DVM requests on
+ * mainnet. This is intended to be used as a "beta" deployment compatible with any EVM-compatible chains before
+ * impleenting a full bridge adapter. Put simply, it is meant as a stop-gap.
+ */
+contract Admin_ChildMessenger is Ownable, Lockable, ChildMessengerInterface {
+    // The only child network contract that can send messages over the bridge via the messenger is the oracle spoke.
+    address public oracleSpoke;
+
+    event SetOracleSpoke(address newOracleSpoke);
+    event MessageSentToParent(bytes data);
+    event MessageReceivedFromParent(bytes data, address indexed targetSpoke, address indexed caller);
+
+    /**
+     * @notice Changes the stored address of the Oracle spoke, deployed on L2.
+     * @dev The caller of this function must be the admin.
+     * @param newOracleSpoke address of the new oracle spoke, deployed on L2.
+     */
+    function setOracleSpoke(address newOracleSpoke) public onlyOwner nonReentrant() {
+        oracleSpoke = newOracleSpoke;
+        emit SetOracleSpoke(newOracleSpoke);
+    }
+
+    /**
+     * @notice Logs a message to be manually relayed to L1.
+     * @dev The caller must be the OracleSpoke on L2. No other contract is permissioned to call this function.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override nonReentrant() {
+        require(msg.sender == oracleSpoke, "Only callable by oracleSpoke");
+
+        // The data formmat is kept the same to match the format in other ChildMessenger classes.
+        bytes memory dataSentToParent = abi.encodeWithSignature("processMessageFromCrossChainChild(bytes)", data);
+
+        // Note: only emit an event. These messages will be manually relayed.
+        emit MessageSentToParent(dataSentToParent);
+    }
+
+    /**
+     * @notice Process a received message from the admin.
+     * @dev The caller must be the the admin.
+     * @param data data message sent from the admin. Should be an encoded function call or packed data.
+     * @param target desired recipient of `data`. Target must implement the `processMessageFromParent` function. Having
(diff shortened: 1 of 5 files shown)


packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * @notice Logs a message to be manually relayed to L1.
+     * @dev The caller must be the OracleSpoke on L2. No other contract is permissioned to call this function.
+     * @param data data message sent to the L1 messenger. Should be an encoded function call or packed data.
+     */
+    function sendMessageToParent(bytes memory data) public override nonReentrant() {
+        require(msg.sender == oracleSpoke, "Only callable by oracleSpoke");
+
+        // The data formmat is kept the same to match the format in other ChildMessenger classes.
+        bytes memory dataSentToParent = abi.encodeWithSignature("processMessageFromCrossChainChild(bytes)", data);
```
chrismaree, line 42: this part is a bit redundant as the use of this contract is to decode whatever was sent from the spoke and then manually relay information to L1. I guess it is good to keep the data's encoding the exact same as the other spokes with the encoded function call on `processMessageFromCrossChainChild`.

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
```diff
@@ -0,0 +1,60 @@
...
+
+        // Note: only emit an event. These messages will be manually relayed.
+        emit MessageSentToParent(dataSentToParent);
```
chrismaree, line 45: seeing the `oracleSpoke ` can change with admin calls, you should include the `oracleSpoke` address in this request.

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    /**
+     * @notice Process a received message from the admin.
+     * @dev The caller must be the the admin.
+     * @param data data message sent from the admin. Should be an encoded function call or packed data.
+     * @param target desired recipient of `data`. Target must implement the `processMessageFromParent` function. Having
+     * this as a param enables the Admin to send messages to arbitrary addresses on the L2. This is primarily
+     * used to send messages to the OracleSpoke and GovernorSpoke on L2.
```
chrismaree, line 54: this comment should change a bit to match the implementation that this is called from the EOA/multisig that mimics a cross-chain function call, without going through a canonical bridge.

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
mrice32: Added it!

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
mrice32: I agree this is weird. I decided to change it to just log out the data. Thoughts?

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
mrice32: Fixed. PTAL

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
chrismaree: ye, I think that's fine. the only real complexity comes in when we are publishing prices. we need to encode things correctly but that's honestly not too difficult given the simplicity of the `processMessageFromParent` on the `OracleSpoke` and `OracleHub.

chrismaree (approved):
This is going to be super useful! very simple but enables a lot of extra functionality

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
```diff
@@ -0,0 +1,57 @@
...
+ * @dev No parent messenger is needed for this case, as the admin could be trusted to manually send DVM requests on
+ * mainnet. This is intended to be used as a "beta" deployment compatible with any EVM-compatible chains before
+ * impleenting a full bridge adapter. Put simply, it is meant as a stop-gap.
+ */
+contract Admin_ChildMessenger is Ownable, Lockable, ChildMessengerInterface {
+    // The only child network contract that can send messages over the bridge via the messenger is the oracle spoke.
+    address public oracleSpoke;
+
+    event SetOracleSpoke(address newOracleSpoke);
+    event MessageSentToParent(bytes data, address indexed oracleSpoke);
```
nicholaspai, line 20: Should we emit `oracleSpoke` in the other `ChildMessenger` contracts events? Currently we do not but I'm curious if you think its useful and you could add in this PR

nicholaspai (commented):
One nit question!

packages/core/contracts/cross-chain-oracle/chain-adapters/Admin_ChildMessenger.sol
mrice32: The only one where it's currently missing is `MessageReceivedFromParent`, right?

I would say no for that one only because it seems somewhat irrelevant. If the target is the oracle spoke, that address will be in the log, but if you're talking to the governors spoke (or any other contract), why do you care about the oracle spoke?

Note: not strongly opinionated, just seems a little irrelevant for that particular call.

Thoughts?

nicholaspai: No I'm talking about the `MessageSentToParent` events in teh `Optimstic/Polygon/Arbitrum_ChildMessenger`

mrice32: Added.

nicholaspai (approved):

PR #3635, opened 2021-11-22
feat(logging): add error logging on disputes and slow relay. Decode OO ancillary data

chrismaree:
**Motivation**

This PR does two things:
1) Generates `error` level log on disputes and slow relays.
2) fixes the OO bot to decode encoded ancillary data.

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -190,7 +190,7 @@ export class Relayer {
       }
 
       try {
-        await this._processTransactionBatch(settleRelayTransactions);
+        await this._processTransactionBatch(settleRelayTransactions as any);
       } catch (error) {
         this.logger.error({ at: "AcrossRelayer#Finalizer", message: "Unexpected error processing relay batch", error });
       }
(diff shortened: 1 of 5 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -559,7 +567,8 @@ export class Relayer {
   private async _sendTransaction(
     transaction: TransactionType,
     message: string,
-    mrkdwn: string
+    mrkdwn: string,
+    level = "info"
```
chrismaree, line 571: level defaults to `info` (what it was before). else, you can define a `level` that will be set on message emmited. we simply append "error" onto the `level` on a transaction.

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -285,6 +290,15 @@ class OptimisticOracleContractMonitor {
   _isFundingRateIdentifier(identifier) {
     return identifier.toLowerCase().endsWith("_fr");
   }
+
+  _formatAncillaryData(ancillaryData) {
```
chrismaree, line 294: simply decode ancillary and append.

nicholaspai: +1

nicholaspai (approved):
lgtm

PR #3637, opened 2021-11-23
fix(across-bots): address issues found when launching on optimism regarding multiple chainIds

chrismaree:
**Motivation**

After launching on Optimism we realized a few small issues due to multiple chain IDs. This PR addresses them.

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -159,11 +159,12 @@ export class Relayer {
 
       // Either this bot is the slow relayer for this relay OR the relay is past 15 mins and is settleable by anyone.
       const settleableRelays = this.l1Client
-        .getSettleableRelayedDepositsForL1Token(l1Token, this.l2Client.chainId)
+        .getSettleableRelayedDepositsForL1Token(l1Token)
         .filter(
           (relay) =>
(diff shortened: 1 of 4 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -159,7 +159,7 @@ export class Relayer {
 
       // Either this bot is the slow relayer for this relay OR the relay is past 15 mins and is settleable by anyone.
       const settleableRelays = this.l1Client
-        .getSettleableRelayedDepositsForL1Token(l1Token)
+        .getSettleableRelayedDepositsForL1Token(l1Token, this.l2Client.chainId)
```
nicholaspai, line 162: Why not filter on `chainId` in this file instead of updating the `InsuredBridgeL1Client. getSettleableRelayedDepositsForL1Token` function? I think its a bit "safer" since we're not updating the upstream code

nicholaspai (approved):
One nit otherwise lgtm

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: sure. please take a look.

nicholaspai (approved):

PR #3638, opened 2021-11-23
improve(across-relayer): Should prune whitelisted L1 tokens using L2 events

nicholaspai:
**Motivation**

Currently the user must specify the `RATE_MODELS` for any L2 network that hasn't whitelisted all of the tokens in the default `RATE_MODELS` dictionary in `@uma/sdk`. We should not ask the user to specify `RATE_MODELS` and they should instead be able to rely on the default in the code. This PR prunes the whitelist derived from `RATE_MODELS` for a specific L2 by removing any L1 tokens not on that chain's specific deposit box.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
@@ -27,7 +27,6 @@ export class InsuredBridgeL2Client {
   private whitelistedTokens: { [key: string]: string } = {}; // L1Token=>L2Token
 
   private firstBlockToSearch: number;
-  private lastBlockToSearch: number | null;
 
   constructor(
     private readonly logger: Logger,
@@ -43,7 +42,6 @@ export class InsuredBridgeL2Client {
     ) as unknown) as BridgeDepositBoxWeb3;
 
     this.firstBlockToSearch = startingBlockNumber;
-    this.lastBlockToSearch = endingBlockNumber;
   }
 
   getAllDeposits() {
@@ -69,7 +67,7 @@ export class InsuredBridgeL2Client {
     // Define a config to bound the queries by.
     const blockSearchConfig = {
       fromBlock: this.firstBlockToSearch,
-      toBlock: this.lastBlockToSearch || (await this.l2Web3.eth.getBlockNumber()),
+      toBlock: this.endingBlockNumber || (await this.l2Web3.eth.getBlockNumber()),
     };
     // TODO: update this state retrieval to include looking for L2 liquidity in the deposit box that can be sent over
     // the bridge. This should consider the minimumBridgingDelay and the lastBridgeTime for a respective L2Token.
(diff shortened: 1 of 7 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -62,14 +69,23 @@ export class InsuredBridgeL2Client {
     // Define a config to bound the queries by.
     const blockSearchConfig = {
       fromBlock: this.firstBlockToSearch,
-      toBlock: this.endingBlockNumber || (await this.l2Web3.eth.getBlockNumber()),
```
chrismaree, line 65: why did you need to make this change? did accessing the `endingBlockNumber` not work? also I don't see why you need to re-declare the `lastBlockToSearch`

packages/insured-bridge-relayer/src/RelayerHelpers.ts
```diff
@@ -35,6 +35,27 @@ export async function approveL1Tokens(
...
+export async function pruneWhitelistedL1Tokens(
+  logger: winston.Logger,
+  l2Client: InsuredBridgeL2Client,
+  whitelistedRelayL1Tokens: string[]
+): Promise<string[]> {
+  await l2Client.update();
+  const filteredWhitelistedRelayL1Tokens = whitelistedRelayL1Tokens.filter((l1TokenAddress: string) => {
+    return l2Client.isWhitelistedToken(l1TokenAddress);
+  });
+  logger.info({
```
chrismaree, line 51: I dont think this should be an `info` log. this means it'll log to slack on every run

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -52,14 +52,25 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
       config.activatedChainIds[0],
-      l2StartBlock
+      l2StartBlock,
+      l2EndBlock
+    );
+
+    // Update the L2 client and filter out tokens that are not whitelisted on the L2 from the whitelisted
+    // L1 relay list.
+    await l2Client.update();
+    const filteredWhitelistedRelayL1Tokens = await pruneWhitelistedL1Tokens(
```
chrismaree, line 70: if you use a smaller variable name this could fit on one line. how about `filteredL1Whitelist`

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: So i discovered this bug when I passed an `endingBlockNumber` into this function. You'll notice that after each loop, the `fromBlock` gets set to the `toBlock + 1`, but the `toBlock` does not get reset, so after the first loop, the `toBlock` remains at the same value while `fromBlock` is now `> toBlock`.

nicholaspai: So you'll notice that after each loop, `fromBlock` gets set to `toBlock + 1`, but `toBlock` is not reset. So this causes the next loop to have `fromBlock > toBlock`. Shouldn't we be incrementing `toBlock` as well? Or perhaps we should rename the constructor variable `endingBlockNumber` to `blockSearchWindowLength`, wdyt?

packages/insured-bridge-relayer/src/RelayerHelpers.ts
nicholaspai: fixed

packages/insured-bridge-relayer/src/index.ts
nicholaspai: done

chrismaree (approved):

PR #3639, opened 2021-11-23
fix(across-relayer): Key BridgePool.l2Tokens by chain ID

nicholaspai:
**Motivation**
Currently we assume there is one L2 token per BridgePool but there can be many depending on how many L2 tokens there are.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -52,7 +52,7 @@ export interface InstantRelay {
 
 export interface BridgePoolData {
   contract: BridgePoolWeb3;
-  l2Token: { [key: string]: string }; // chainID=>L1TokenAddress
+  l2Token: { [chainId: string]: string }; // chainID=>L2TokenAddress
   currentTime: number;
   relayNonce: number;
   poolCollateralDecimals: number;
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -160,7 +160,8 @@ export class InsuredBridgeL1Client {
   getSettleableRelayedDepositsForL1Token(l1Token: string): Relay[] {
     return this.getRelayedDepositsForL1Token(l1Token).filter(
       (relay: Relay) =>
-        relay.relayState === ClientRelayState.Pending && relay.settleable != SettleableRelay.CannotSettle
+        relay.relayState === ClientRelayState.Pending && // The relay is in the Pending state.
+        relay.settleable != SettleableRelay.CannotSettle // The relay is not set to CannotSettle.
     );
   }
 
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -215,7 +215,7 @@ export class InsuredBridgeL1Client {
 
   getBridgePoolForL2Token(l2Token: string, chainId: string): BridgePoolData {
     const bridgePoolData = Object.values(this.bridgePools).find((bridgePool) => {
-      return bridgePool.l2Token[chainId] === l2Token;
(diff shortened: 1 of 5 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -52,7 +52,7 @@ export interface InstantRelay {
 
 export interface BridgePoolData {
   contract: BridgePoolWeb3;
-  l2Token: string;
+  l2Token: { [key: string]: string }; // chainID=>L1TokenAddress
```
chrismaree, line 55:
```suggestion
  l2Token: { [key: string]: string }; // chainID=>L2TokenAddress
```

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -112,7 +112,7 @@ export class CrossDomainFinalizer {
 
   // Fetch L1 pool reserves for a given l2Token.
   private async _getL1PoolReserves(l2Token: string): Promise<BN> {
-    const bridgePool = this.l1Client.getBridgePoolForL2Token(l2Token).contract;
+    const bridgePool = this.l1Client.getBridgePoolForL2Token(l2Token, this.l2Client.chainId.toString()).contract;
```
chrismaree, line 115: nice. I'm pretty sure this is the only place we use this method.

chrismaree (approved):
good fix.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -52,7 +52,7 @@ export interface InstantRelay {
 
 export interface BridgePoolData {
   contract: BridgePoolWeb3;
-  l2Token: string;
+  l2Token: { [key: string]: string }; // chainID=>L2TokenAddress
```
mrice32, line 55:
```suggestion
  l2Token: { [chainId: string]: string }; // chainID=>L2TokenAddress
```

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -213,9 +213,11 @@ export class InsuredBridgeL1Client {
     return this.bridgePools[l1Token];
   }
 
-  getBridgePoolForL2Token(l2Token: string): BridgePoolData {
-    const bridgePoolData = Object.values(this.bridgePools).filter((bridgePool) => bridgePool.l2Token == l2Token)[0];
-    if (!bridgePoolData) throw new Error(`No bridge pool initialized for ${l2Token}`);
+  getBridgePoolForL2Token(l2Token: string, chainId: string): BridgePoolData {
+    const bridgePoolData = Object.values(this.bridgePools).filter(
```
mrice32, line 217: nit: find will only return one value, making this a little simpler

mrice32 (approved):
Looks good!

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: +1 fixed

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -213,9 +213,11 @@ export class InsuredBridgeL1Client {
     return this.bridgePools[l1Token];
   }
 
-  getBridgePoolForL2Token(l2Token: string): BridgePoolData {
-    const bridgePoolData = Object.values(this.bridgePools).filter((bridgePool) => bridgePool.l2Token == l2Token)[0];
-    if (!bridgePoolData) throw new Error(`No bridge pool initialized for ${l2Token}`);
+  getBridgePoolForL2Token(l2Token: string, chainId: string): BridgePoolData {
+    const bridgePoolData = Object.values(this.bridgePools).find((bridgePool) => {
```
mrice32, line 217: I think you might need to remove the braces around the inner function or add a "return". Not 100% sure, but I think the current syntax will cause the inner function to always return undefined.

PR #3640, opened 2021-11-23
fix(across-relayer): Skipped relay should add `undefined` to disputeRelay array, not `null`

nicholaspai:
**Motivation**

For the disputer, when a pending relay is skipped because its `chainID !== L2Client's chain ID` but the `chainID` is whitelisted, the bot should add an `undefined` entry to the relaysToDispute array, but incorrectly adds `null`.

I discovered this bug because the `Optimism` bot was timing out unexpectedly when there were ONLY pending relays for `Arbitrum` deposits, and therefore an array of `null` pending relays were being passed to `_generateRelayTransactionForPendingDeposit` which only filters out `undefined` relays. However, this bug mysteriously 
disappeared when an `undefined` relay was added to the array passed to `_generateRelayTransactionForPendingDeposit`, for example when the relay is for the `Optimism` chain but its already sped up!

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -130,7 +130,11 @@ export class Relayer {
     }
     this.logger.debug({
       at: "AcrossRelayer#Disputer",
-      message: `Processing ${Object.keys(pendingRelays).length} l1 token pending relays`,
+      message: `Processing pending relays for ${Object.keys(pendingRelays).length} l1 tokens`,
+      // Log # of relays for each L1 token:
+      pendingRelayCounts: Object.keys(pendingRelays).map((l1Token) => ({
+        [l1Token]: pendingRelays[l1Token].length,
+      })),
     });
     for (const l1Token of Object.keys(pendingRelays)) {
       const disputeTransactions = []; // Array of dispute transactions to send.
@@ -249,7 +253,7 @@ export class Relayer {
         l2ClientChainId: this.l2Client.chainId,
         relay,
       });
-      return null;
+      return;
     }
 
     // Fetch deposit for relay.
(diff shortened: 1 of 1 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -130,7 +130,11 @@ export class Relayer {
     }
     this.logger.debug({
       at: "AcrossRelayer#Disputer",
-      message: `Processing ${Object.keys(pendingRelays).length} l1 token pending relays`,
+      message: `Processing pending relays for ${Object.keys(pendingRelays).length} l1 tokens`,
```
nicholaspai, line 133: Unrelated but this log wasn't accurate.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -249,7 +253,7 @@ export class Relayer {
         l2ClientChainId: this.l2Client.chainId,
         relay,
       });
-      return null;
```
nicholaspai, line 252: This is the bug fix.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -513,7 +517,7 @@ export class Relayer {
     transactions: { transaction: TransactionType | any; message: string; mrkdwn: string; level: string }[]
   ) {
     // Remove any undefined transaction objects or objects that contain null transactions.
-    transactions = transactions.filter((transaction) => transaction && transaction.transaction);
+    transactions = transactions.filter((transaction) => transaction && transaction !== null && transaction.transaction);
```
nicholaspai, line 520: This is an alternative bug fix, but we might as well add it as well.

chrismaree: Good idea. Should watch out for this confusion in the future. Perhaps having a specific explicit null type.

chrismaree (approved):

PR #3644, opened 2021-11-24
fix(across-relayer): Key L1 client dictionary by checksummed addresses

nicholaspai:
**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -250,7 +253,7 @@ export class InsuredBridgeL1Client {
     // while the bot is running.
     const whitelistedTokenEvents = await this.bridgeAdmin.getPastEvents("WhitelistToken", blockSearchConfig);
     for (const whitelistedTokenEvent of whitelistedTokenEvents) {
-      const l1Token = whitelistedTokenEvent.returnValues.l1Token;
+      const l1Token = this.l1Web3.utils.toChecksumAddress(whitelistedTokenEvent.returnValues.l1Token);
```
chrismaree, line 256: nit: please add `utils.toChecksumAddress` to `const { toBN, soliditySha3 } = Web3.utils;` at the top.

chrismaree (approved):
please see nit. once refactored can merge.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -215,7 +215,10 @@ export class InsuredBridgeL1Client {
 
   getBridgePoolForL2Token(l2Token: string, chainId: string): BridgePoolData {
     const bridgePoolData = Object.values(this.bridgePools).find((bridgePool) => {
-      return bridgePool.l2Token[chainId].toLowerCase() === l2Token.toLowerCase();
```
mrice32, line 218: Didn't the error say that one of these was _undefined_? Do we know why that happened?

nicholaspai: Ah I figured it out. The issue is that there are two L2 tokens for one bridgepool, one for arbitrum and one for optimism. This errors out when searching through the bridge pools via `.find()` because calling `bridgePool.l2Token[chainId]` is undefined for the some `bridgePools` such as `BADGER` which doesn't have an entry for the network with chain ID 10. So this PR will fix it because it won't be calling a function on undefined anymore.

mrice32: What happens if you call toChecksumAddress on undefined?

nicholaspai: returns ''

chrismaree: Ye In all cases, whenever dealing with addresses, we must use this syntax. Lower case is not tie way. This is a code smell. We are good about this in the other bots and should try stay consistent

chrismaree (approved):
Good stuff. I like the additional logs as well.

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
mrice32: Cool cool, sounds good. As long as it doesn't error, SGTM

PR #3645, opened 2021-11-24
improve(across-relayer): Remove redundant L1 event query

nicholaspai:
**Motivation**

CrossDomainFinalizer queries for `WhitelistTokenEvents` but L1 client does the same thing in its `update()` method. We should cache all such events in the L1 client and read them from the CrossDomainFinalizer.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


chrismaree (approved):
great change. previously we were doing less stuff with the whitelist so made sence to have the call in the cross-domain finaliser. now that we do this and the filtering in from the rate model having this in the L1Client seems better.

mrice32 (approved):
LGTM!

PR #3646, opened 2021-11-24
feat(sdk): ability to send funds from Ethereum to Boba

amateima:
**Motivation**

https://app.shortcut.com/uma-project/story/3301/ability-to-send-funds-from-ethereum-to-boba-in-sdk

This is almost the same as L1 -> Optimism [PR](https://github.com/UMAprotocol/protocol/pull/3623). Boba and Optimism looks are very similar, they even share the same npm libraries for contract utilities.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


mrice32:
@amateima 

> Boba and Optimism looks are very similar, they even share the same npm libraries for contract utilities.

Are there any differences or reasons why we can't use the exact same code for both?

packages/sdk/src/across/clients/bobaBridge.ts
```diff
@@ -0,0 +1,101 @@
...
+  async depositERC20(l1Signer: Signer, l1Erc20Address: string, l2Erc20Address: string, amount: BigNumber) {
+    const chainId = await l1Signer.getChainId();
+    const addressManagerAddress = l1Contracts.ADDRESS_MANAGER_ADDRESS[chainId];
+    assert(typeof addressManagerAddress === "string", "Chain not supported");
+
+    const addressManager = BobaAddressManagerEthers__factory.connect(addressManagerAddress, l1Signer);
+    const l1StandardBridgeAddress = await addressManager.getAddress("Proxy__OVM_L1StandardBridge");
+    const l1StandardBridge = OptimismL1StandardBridgeEthers__factory.connect(l1StandardBridgeAddress, l1Signer);
+    const l1_ERC20 = ERC20Ethers__factory.connect(l1Erc20Address, l1Signer);
+    return l1StandardBridge.depositERC20(
```
mrice32, line 42: Do approvals need to happen before this? Should we offer a utility so users can 1) check if they've already approved the bridge to pull their tokens and 2) do the approval?

packages/sdk/src/across/clients/bobaBridge.ts
```diff
@@ -0,0 +1,101 @@
...
+export const l1Contracts: { ADDRESS_MANAGER_ADDRESS: { [chainId: number]: string } } = {
+  ADDRESS_MANAGER_ADDRESS: {
+    // mainnet
+    1: "0x8376ac6C3f73a25Dd994E0b0669ca7ee0C02F089",
+    // rinkeby
+    4: "0x93A96D6A5beb1F661cf052722A1424CDDA3e9418",
+  },
+};
+
+export class BobaBridgeClient {
```
mrice32, line 21: If this is very similar to Optimism, do you see an advantage of having them share code? Or do you think it's safer to keep them distinct in case they diverge in the future?

mrice32 (commented):

daywiss (approved):
looks good to me, but i think we will also need a way to get projected fees of transfer. we can do this as a new pr though

daywiss:
oh also i agree with matt, if possible, probably good to use the same code on optimism and boba

amateima:
There are a few slightly differences
- on Optimism, the contract's addresses are hardcoded, Boba uses a smart contract to store the addresses
- Boba examples use the encoded current date as tx data

I can extract the common logic in a base class, but in the end we'll still have 2 different implementations (classes). What do you think?

amateima:
> looks good to me, but i think we will also need a way to get projected fees of transfer. we can do this as a new pr though

are you referring to L1 transaction fees?

packages/sdk/src/across/clients/bobaBridge.ts
amateima: Approvals are required, yes, but this logic already exists in the FE app for other contracts. I think that exporting the Bridge addresses is enough

daywiss:
> I can extract the common logic in a base class, but in the end we'll still have 2 different implementations (classes). What do you think? There's also the possibility to have a single class OptimisticStandardBridge, common methods for depositing (depositEth, depositErc20), but provide a factory method to generate the arguments for each call depending on the desired L2 (bridges addresses and tx data)

If they arent identical i would say keep things how they are then and have separate implementations. we can refactor later if its a big deal

daywiss:
> are you referring to L1 transaction fees?

Well i dont exactly know whats involved, but i assume theres some fees to do initiate the transfer which may be above just the gas costs. Essentially we will need to fill this part of the form out: 
image

amateima:
> Well i dont exactly know whats involved, but i assume theres some fees to do initiate the transfer which may be above just the gas costs. Essentially we will need to fill this part of the form out: image

I made a bunch of deposits and I didn't see any fee other than the L1 tx fee

daywiss:
> I made a bunch of deposits and I didn't see any fee other than the L1 tx fee

Ok good to know

packages/sdk/src/across/clients/bobaBridge.ts
amateima: I added methods for approve and allowance just in case

PR #3653, opened 2021-11-29
improve(across): Improve L2 client provider redundancy

nicholaspai:
**Motivation**
L2 web3 providers should be expected to return bad data occasionally, this PR enhances redundancy by allowing user to specify a set of fallback l2 web3 providers. If the fallback set is specified, then all L2 events must be found in ALL fallback l2 web3 providers otherwise an error is thrown. This error is thrown instead of a false dispute being sent.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/ProviderUtils.ts
```diff
@@ -151,6 +151,25 @@ export function getWeb3ByChainId(chainId: number): Web3 {
...
+  const retryConfigJson = process.env[`RETRY_CONFIG_${chainId}`];
+  if (!retryConfigJson) return [];
+  const retryConfig: RetryConfig[] = JSON.parse(retryConfigJson);
+
+  const retryWeb3s: Web3[] = [];
+  retryConfig.forEach((config) => {
+    retryWeb3s.push(new Web3(config.url));
+  });
+
+  return retryWeb3s;
```
mrice32, line 170: I think this can be simplified to remove the special cases.

```suggestion
  const retryConfigJson = process.env[`RETRY_CONFIG_${chainId}`] || "[]";
  const retryConfig: RetryConfig[] = JSON.parse(retryConfigJson);
  return retryConfig.map((config) => new Web3(config.url));
```

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -121,6 +120,45 @@ export class InsuredBridgeL2Client {
     });
   }
 
+  async getEventsForBlockSearch(blockSearchConfig: { fromBlock: number; toBlock: number }): Promise<EventData[][]> {
```
mrice32, line 123: Do you think we could make a utility method that has the following interface? (Note: some of the types might be wrong, just wanted to jot an example down)

```ts
async getEventsForMutlipleProviders(web3s: Web3[], contract: Contract, eventName: string, eventParameters: Parameters<Contract["getPastEvents"]>[1]): Promise<EventData[][]> {
```

Or something like that? That way, we can reuse this for any event query?

I think this would allow us to separate out the event-specific logic from the retry/fallback logic.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -121,6 +120,45 @@ export class InsuredBridgeL2Client {
...
+        getAbi("BridgeDepositBox"),
+        this.bridgeDepositAddress
+      ) as unknown) as BridgeDepositBoxWeb3;
+      const [_fundsDepositedEvents, _whitelistedTokenEvents] = await Promise.all([
+        _bridgeDepositBox.getPastEvents("FundsDeposited", blockSearchConfig),
+        _bridgeDepositBox.getPastEvents("WhitelistToken", blockSearchConfig),
+      ]);
+      _fundsDepositedEvents.forEach((event) => {
+        if (!fundsDepositedTransactionHashes.includes(event.transactionHash)) {
+          throw new Error(
```
mrice32, line 145: Are we sure we want to _throw_? Assuming most of the event data is valid and agreed upon, would this cause us to block the relayer/disputer from running even on non-contested items?

mrice32 (commented):
Looks good -- only major point is around generalizing the fallback process into a utility that can be reused.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Hm what should we do if not throw? Should we return the union of all events found?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Yep done

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +127,85 @@ export class InsuredBridgeL2Client {
...
+   * Fetches specified event data for all input web3 providers. Assumes that the first fetched EventData array contains
+   * the "control" events. Compares the "control" events against each of the other EventData arrays. Returns false
+   * if any of the "control" events are NOT found exactly in ALL of the other EventData arrays.
+   * @param web3s Web3 providers to check for target event.
+   * @param contract Contract to query target event on.
+   * @param eventName Name of target event
+   * @param blockSearchConfig Target event search config
+   * @returns Object containing success of event query, missing event if not found in all providers, and all event data
+   */
+  async getEventsForMultipleProviders(
```
nicholaspai, line 168: @mrice32 here's the module we can move into some external utilities file. I haven't moved it yet bc i'm not sure where to place it. Thoughts?

mrice32: @daywiss do you have thoughts? I could see us putting this in the SDK, common, or financial-templates-lib/src/utils.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +127,85 @@ export class InsuredBridgeL2Client {
...
+    web3s: Web3[],
+    contract: Contract,
+    eventName: string,
+    blockSearchConfig: BlockSearchConfig
+  ): Promise<{ success: boolean; missingEvent: string | null; events: EventData[][] }> {
+    const getEventsPromises = [contract.getPastEvents(eventName, blockSearchConfig)];
```
mrice32, line 174: Why not _just_ query on the provided web3 objects? Seems simpler for the user to pass in all the web3s they want rather than having a special one that's included in the contract instance. Thoughts?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +127,85 @@ export class InsuredBridgeL2Client {
...
+
+    // For each fallback web3 provider, check that the specified events are also found by those providers,
+    // otherwise throw an error. This allows the caller to have additional confidence about the accuracy of fetched
+    // contract state.
+    for (let i = 0; i < web3s.length; i++) {
```
mrice32, line 179: nit: would it be simpler just do do `await Promise.all(web3.map(...` here instead of building up an array of event promises?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +127,85 @@ export class InsuredBridgeL2Client {
...
+
+    // events[0] contains the events returned by the main web3 provider at index 0. We'll compare those events
+    // against those returned by the fallback providers in events[1,...n].
+    for (let i = 1; i < events.length; i++) {
+      const fallbackEvents = events[i];
+      fallbackEvents.forEach((event) => {
+        if (!controlEvents.includes(event.transactionHash)) {
+          return {
+            success: false,
+            missingEvent: event.transactionHash,
```
mrice32, line 195: Should we wait to return so we can return a list of _all_ missing events?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +127,85 @@ export class InsuredBridgeL2Client {
...
+    const events = await Promise.all(getEventsPromises);
+
+    const controlEvents = events[0].map((event) => event.transactionHash);
+
+    // events[0] contains the events returned by the main web3 provider at index 0. We'll compare those events
+    // against those returned by the fallback providers in events[1,...n].
+    for (let i = 1; i < events.length; i++) {
+      const fallbackEvents = events[i];
+      fallbackEvents.forEach((event) => {
+        if (!controlEvents.includes(event.transactionHash)) {
```
mrice32, line 192: Would this break if there were two matching events in the same transaction, but one was dropped?

mrice32 (commented):
Looks great!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: yeah true, what do you think is the minum set of values we need to lookup?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Yeah that's good feedback, will change

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: yup good shout

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: yup good call

nicholaspai: I return all missing events now, TBD what we'll do with them but ptal at the interface

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
mrice32: I'm not sure. I think there are a few ways to approach this:
- We could use transactionIndex and logIndex. That would mean that we're ensuring the _positioning_ of each log has a match. That should give us a lot of confidence.
- In addition to the above, if we want more confidence, we might want to somehow encode the data in the returnValues of the event. This would ensure that it's not just the same event name in the same spot in the block, but that it also has the exact same arguments.

Thoughts?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +128,88 @@ export class InsuredBridgeL2Client {
...
+    // Compare unique event keys from all providers with the provider at index 0, and keep track of mismatching events.
+    const missingEvents: EventData[] = [];
+    for (let i = 1; i < events.length; i++) {
+      const _events = events[i];
+      _events.forEach((event) => {
+        if (!uniqueEventKeys.includes(event.transactionHash)) {
+          missingEvents.push(event);
+        }
+      });
+    }
```
mrice32, line 205: A few points on this loop:
- It looks like this is specifically looking for events that providers 2, 3, 4... have that are not included in provider 1. Do we care about the case where 1 has an event, but providers 2, 3, 4... _don't_? The former is the problem we've seen before, but do we want to cover the other form of "disagreement"?
- The way the loop is designed, if providers 2, 3, 4 all have the same event that 1 is missing, we would have that event show up 3+ times in our missingEvents array, right? Wouldn't we only want that to show up once since it's the same event?

mrice32 (commented):
Looks awesome!! One comment (I also responded to one of the existing threads).

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Do you think it's reasonable to just stringify the entire Event object? Would this vary between providers?

nicholaspai: Empirically this does match events across Alchemy and Infura, thoughts?

```ts
    const getUniqueEventKey = (event: EventData): string => {
      return JSON.stringify({
        transactionHash: event.transactionHash,
        transactionIndex: event.transactionIndex,
        logIndex: event.logIndex,
        returnValues: event.returnValues,
        address: event.address
      })
    }
```

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: I've updated this core logic now. This helper method now returns two sets of EventData arrays:
- `missingEvents`: any events found in provider 1 and NOT found in ALL of providers 2 through n
- `matchedEvents`: any events found in provider 1 and also found in ALL of providers 2 through n

There should be no duplicate events anymore, PTAL!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -22,6 +22,11 @@ export interface Deposit {
   depositContract: string;
 }
 
+type BlockSearchConfig = {
+  fromBlock: number;
```
mrice32, line 26: Should we allow filter params so the caller can filter events?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +128,113 @@ export class InsuredBridgeL2Client {
...
+      })
+    );
+
+    // Associate each event with a key uniquely identifying the event. We'll use this key in the next step to determine
+    // which events were returned by all providers.
+    type UniqueEventData = { [uniqueEventKey: string]: EventData };
+    const uniqueEventsForProvider: UniqueEventData[] = [];
+    // [index of web3Provider => {eventKey => event}]
+
+    const _getUniqueEventKey = (event: EventData): string => {
```
mrice32, line 201: Awesome!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +128,113 @@ export class InsuredBridgeL2Client {
...
+      eventDataForProvider.forEach((event) => {
+        const uniqueEventKey = _getUniqueEventKey(event);
+        uniqueEventsForProvider[i][uniqueEventKey] = event;
+      });
+    });
+
+    // Store only the events returned by ALL providers.
+    const eventKeysReturnedByAllProviders: string[] = [];
+    const missingEventKeys: string[] = [];
+    Object.keys(uniqueEventsForProvider[0]).forEach((eventKeysForFirstProvider: string) => {
```
mrice32, line 221: What happens if an event is missing in the 0th provider?

If you agree that this case isn't covered, I might have a slight tweak that would make it work. You could add a union map that you build in the first loop. If you encounter any events that aren't in that map, you add it in the first loop.

Then in this loop, you can loop over all the elements in that union map so you know that you're covering everything.

mrice32 (commented):
Looks great!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Sure done

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Ah yes I agree with this optimization. It adds an additional struct but no more loops

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +130,119 @@ export class InsuredBridgeL2Client {
...
+          break;
+        }
+      }
+      if (eventFoundInAllProviders) eventKeysReturnedByAllProviders.push(eventKey);
+      else missingEventKeys.push(eventKey);
+    });
+
+    return {
+      missingEvents: missingEventKeys.map((eventKey) => uniqueEventsForProvider[0][eventKey]),
+      events: eventKeysReturnedByAllProviders.map((eventKey) => uniqueEventsForProvider[0][eventKey]),
```
mrice32, line 243:
```suggestion
      missingEvents: missingEventKeys.map((eventKey) => uniqueEvents[eventKey]),
      events: eventKeysReturnedByAllProviders.map((eventKey) => uniqueEvents[eventKey]),
```

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +130,119 @@ export class InsuredBridgeL2Client {
...
+        uniqueEvents[uniqueEventKey] = event;
+      });
+    });
+
+    // Store only the events returned by ALL providers.
+    const eventKeysReturnedByAllProviders: string[] = [];
+    const missingEventKeys: string[] = [];
+    Object.keys(uniqueEvents).forEach((eventKey: string) => {
+      let eventFoundInAllProviders = true;
+      for (let providerIndex = 1; providerIndex < uniqueEventsForProvider.length; providerIndex++) {
```
mrice32, line 231:
```suggestion
      for (let providerIndex = 0; providerIndex < uniqueEventsForProvider.length; providerIndex++) {
```

mrice32 (commented):
LGTM once these quick suggestions/fixes are implemented

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +130,119 @@ export class InsuredBridgeL2Client {
...
+      });
+    };
+    allProviderEvents.forEach((eventDataForProvider, i) => {
+      uniqueEventsForProvider[i] = {};
+      eventDataForProvider.forEach((event) => {
+        const uniqueEventKey = _getUniqueEventKey(event);
+        // Add event for this provider.
+        uniqueEventsForProvider[i][uniqueEventKey] = event;
+        // Add event to union map if we haven't seen it before.
+        uniqueEvents[uniqueEventKey] = event;
```
daywiss, line 222: i think this would be useful to also store a map of  counts, and increment it every time you run into an event with that key.  This way u can just iterate over the counts in the following loop and it might simplify some of the logic there

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,14 +130,119 @@ export class InsuredBridgeL2Client {
...
+          eventFoundInAllProviders = false;
+          break;
+        }
+      }
+      if (eventFoundInAllProviders) eventKeysReturnedByAllProviders.push(eventKey);
+      else missingEventKeys.push(eventKey);
+    });
+
+    return {
+      missingEvents: missingEventKeys.map((eventKey) => uniqueEventsForProvider[0][eventKey]),
```
daywiss, line 242: Why would you care about this?

packages/common/src/ProviderUtils.ts
```diff
@@ -151,6 +151,20 @@ export function getWeb3ByChainId(chainId: number): Web3 {
   return new Web3(keyedProvider);
 }
 
+/**
+ * @notice Creates array of web3 instances for a particular chain.
+ * @dev Providers to use are described in RETRY_CONFIG_{chainId} dictionary under the "url" key.
+ * @param chainId the chain id for the network the user wants to connect to.
+ * @returns array of new readonly Web3 instances.
```
chrismaree, line 158: this is fresh.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: I'll leave it in this package for now, but probably move to a new file within `src/helpers` instead of the L2Client. And if we want to use it elsewhere in the future we can move it

chrismaree: ye I I dont think this should go in the SDK. perhaps `common` in a new util called `EventUtils` or something?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: wow great catch

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: It could be used in an error log to help debug which events are missing in some providers and not in others?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: But you'd need to remember which provider index saw that event?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: hmmm I like that better actually

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
chrismaree: that looks great! makes sence to throw error & log.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
daywiss: i didnt see where that mattered, it looks like u just care if the event exists and if its found in all providers. you already have the map of uniqueEvents to do a lookup without having to know which provider it was from.  i might be missing something though

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
daywiss: i see now, you dont include them in the event list. i missed that the first time

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: Ah I see what you're saying, this is a really clever suggestion! PTAL at [this implementation](https://github.com/UMAprotocol/protocol/pull/3653/commits/f6e5c58c8e8d6681e9a1320332385624186006e6)

packages/common/src/ContractUtils.ts
```diff
@@ -79,3 +80,79 @@ export const replaceLibraryBindingReferenceInArtitifact = <T>(artifact: T, libra
...
+      else uniqueEvents[uniqueEventKey].count++;
+    });
+  });
+
+  // Store only the events returned by ALL providers.
+  const eventKeysReturnedByAllProviders: string[] = [];
+  const missingEventKeys: string[] = [];
+  Object.keys(uniqueEvents).forEach((eventKey: string) => {
+    if (uniqueEvents[eventKey].count === web3s.length) eventKeysReturnedByAllProviders.push(eventKey);
+    else missingEventKeys.push(eventKey);
```
daywiss, line 151: Yea looks great, glad that worked

packages/common/src/ContractUtils.ts
```diff
@@ -79,3 +80,79 @@ export const replaceLibraryBindingReferenceInArtitifact = <T>(artifact: T, libra
...
+      returnValues: event.returnValues,
+      address: event.address,
+    });
+  };
+  allProviderEvents.forEach((eventDataForProvider, i) => {
+    uniqueEventsForProvider[i] = {};
+    eventDataForProvider.forEach((event) => {
+      const uniqueEventKey = _getUniqueEventKey(event);
+      // Add event for this provider.
+      uniqueEventsForProvider[i][uniqueEventKey] = event;
```
daywiss, line 139: Is this still needed?

daywiss (approved):

packages/common/src/ContractUtils.ts
nicholaspai: Nope! Great catch. This function is so much cleaner now

packages/common/src/ContractUtils.ts
```diff
@@ -79,3 +80,71 @@ export const replaceLibraryBindingReferenceInArtitifact = <T>(artifact: T, libra
...
+ * Fetches specified contract event data for all input web3 providers. Returns false if any of the events found with
+ * one provider are NOT matched exactly in ALL of the other providers' event arrays.
+ * @param web3s Web3 providers to check for target event.
+ * @param contractAbi Contract ABI to query target event on.
+ * @param contractAddress Contract address to query target event on.
+ * @param eventName Name of target event.
+ * @param eventSearchOptions Target event search options. See here for more details: https://web3js.readthedocs.io/en/v1.5.2/web3-eth-contract.html#getpastevents
+ * @returns Object containing success of event query, missing events not found in all providers, and all event data
+ */
+export async function getEventsForMultipleProviders(
```
chrismaree, line 101: this is great here. thanks for putting it in a common file.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,12 +125,60 @@ export class InsuredBridgeL2Client {
     });
   }
 
-  async getFundsDepositedEvents(blockSearchConfig: { fromBlock: number; toBlock: number }): Promise<EventData[]> {
-    return await this.bridgeDepositBox.getPastEvents("FundsDeposited", blockSearchConfig);
+  async getFundsDepositedEvents(eventSearchOptions: EventSearchOptions): Promise<EventData[]> {
+    const eventsData = await getEventsForMultipleProviders(
+      [this.l2Web3].concat(this.fallbackL2Web3s),
+      this.bridgeDepositBox.options.jsonInterface,
```
chrismaree, line 131: do you think it's better to call this using `getAbi("BridgeDepositBox")` or the way you have it now?

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -53,13 +53,21 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
     const l2Web3 = getWeb3ByChainId(config.activatedChainIds[0]);
     const latestL2BlockNumber = await l2Web3.eth.getBlockNumber();
     const l2StartBlock = Math.max(0, latestL2BlockNumber - config.l2BlockLookback);
-
+    const fallbackL2Web3s = getRetryWeb3sByChainId(config.activatedChainIds[0]);
+    // Note: This will not construct duplicate Web3 objects for URL's in the Retry config that are the same as the
+    // one used to construct l2Web3.
+    logger.debug({
+      at: "AcrossRelayer#index",
+      message: `Constructed ${fallbackL2Web3s.length} fallback L2 web3 providers`,
```
chrismaree, line 61: I wish you could extract the `rpc` url from a provider. would make the logs here and in the client so much nicer.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -122,12 +125,60 @@ export class InsuredBridgeL2Client {
...
+      this.bridgeDepositAddress,
+      "WhitelistToken",
+      eventSearchOptions
+    );
+    if (eventsData.missingEvents.length > 0) {
+      const errorMessage = `L2 RPC endpoints disagree about L2 contract events, please manually investigate.`;
+      const error = new Error(errorMessage);
+      this.logger.error({
+        at: "InsuredBridgeL2Client",
+        message:
```
chrismaree, line 169: this is a long message! it might be good to split this up. perhaps make the `message` key short. something like: `L2 RPC endpoint state disagreement! 🤺` then put a different key for the "solution" or "moreDetails". the reason I say you should do it like this is two fold:
1) you need an emoji on all slack logs ;)
2) the "message" field is sent in the top line of the slack message. the should be short and direct. This is analogous to the `Relay instantly sent` for the relay. the "details" should follow below the message

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: i think using `getAbi` is more consistent thanks for highlighing

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: ugh you're totally right i forgot about emojis

PR #3654, opened 2021-11-29
refactor(across-relayer): Remove event querying from relayer code

nicholaspai:
**Motivation**

Relayer should not query events, we can easily move event-querying-logic into L2 client if we do this refactor and add an additional L2 client external function.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
@@ -4,6 +4,7 @@
 import { getAbi } from "@uma/contracts-node";
 import type { BridgeDepositBoxWeb3 } from "@uma/contracts-node";
 import Web3 from "web3";
+import { EventData } from "web3-eth-contract";
 const { toChecksumAddress } = Web3.utils;
 import type { Logger } from "winston";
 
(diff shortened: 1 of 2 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -121,6 +119,13 @@ export class InsuredBridgeL2Client {
     });
   }
 
+  async getEventsForBlockSearch(blockSearchConfig: { fromBlock: number; toBlock: number }): Promise<EventData[][]> {
+    return await Promise.all([
+      this.bridgeDepositBox.getPastEvents("FundsDeposited", blockSearchConfig),
+      this.bridgeDepositBox.getPastEvents("WhitelistToken", blockSearchConfig),
+    ]);
+  }
```
mrice32, line 127: Two nits:
1. Should the return values be named? Without looking at the implementation, it's hard to tell what return[0] and return[1] are.
2. getEventsForBlockSearch seems like an odd method for a general-purpose client. This combined call makes sense for our relayer but may be overkill for other uses. Thoughts on 1) separating these queries into two methods and 2) naming them more specifically so it's clear what the method will be returning?

mrice32 (commented):
Only one comment!

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: On point 2 do you mean making one GET method for each event name?

mrice32: Yeah. I know that's kinda annoying since it's more verbose.

We could just do 1 and have the interface more clearly communicate what is being returned.

My main push is that the user should be able to look at the interface (`
  async getEventsForBlockSearch(blockSearchConfig: { fromBlock: number; toBlock: number }): Promise<EventData[][]>`) and know what the method is going to give them and how to use it -- any way we can achieve that SGTM.

Thoughts?

mrice32 (commented):
https://github.com/UMAprotocol/protocol/pull/3654/files#r758713949

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
nicholaspai: No I think you're right about having one method for each event name, I've implemented the change.

mrice32 (approved):
LGTM!

PR #3655, opened 2021-11-29
feat(sdk): add projected apr calculation

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3275/implement-projected-apy-calculation-in-sdk

**Summary**

Exposes a function already in the sdk for calcultaing instant apr fees. Uses this and rate model, utilization and adds it to pool data.

**Details**

image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/3275/implement-projected-apy-calculation-in-sdk

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -1,6 +1,6 @@
 import assert from "assert";
 import { bridgePool } from "../../clients";
-import { toBNWei, fixedPointAdjustment, calcPeriodicCompoundInterest, calcApr, BigNumberish } from "../utils";
+import { toBNWei, fixedPointAdjustment, calcPeriodicCompoundInterest, calcApr, BigNumberish, fromWei } from "../utils";
 import { BatchReadWithErrors, loop, exists } from "../../utils";
 import Multicall2 from "../../multicall2";
 import TransactionManager from "../transactionManager";
@@ -11,6 +11,8 @@ import { Provider, Block } from "@ethersproject/providers";
 import set from "lodash/set";
 import get from "lodash/get";
 import has from "lodash/has";
+import { calculateInstantaneousRate } from "../feeCalculator";
+import { getRateModel } from "../constants";
 
 export const SECONDS_PER_YEAR = 31557600; // based on 365.25 days per year
 export const DEFAULT_BLOCK_DELTA = 10; // look exchange rate up based on 10 block difference by default
@@ -41,6 +43,7 @@ export type Pool = {
   blocksElapsed: number;
   secondsElapsed: number;
   liquidityUtilizationCurrent: string;
+  projectedApr: string;
 };
 export type User = {
   address: string;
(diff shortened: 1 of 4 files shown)


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -253,6 +256,11 @@ function joinPoolState(
     SECONDS_PER_YEAR
   );
   const estimatedApr = calcApr(exchangeRatePrevious, exchangeRateCurrent, secondsElapsed, SECONDS_PER_YEAR);
+  let projectedApr = "N/A";
+  const rateModel = getRateModel(poolState.l1Token);
+  if (rateModel) {
+    projectedApr = fromWei(calculateInstantaneousRate(rateModel, poolState.liquidityUtilizationCurrent));
```
mrice32, line 262: Do we need to multiply by the utilization?

mrice32 (commented):
Looks good!

packages/sdk/src/across/clients/bridgePool.ts
daywiss: yea, good call

daywiss: do we also have to mutliply the current apr/apy value by utilization?

daywiss: this makes projected apr much lower fyi 
image

mrice32: @daywiss 

> this makes projected apr much lower fyi

I think that's expected. The pool utilization for ETH is 20% right now, which is pretty low, meaning we'd be charging a low fee and it's being charged on a very small amount of the capital in the pool (since very little is loaned out).

We should probably chat with other folks to ensure that we want to show this.

> do we also have to mutliply the current apr/apy value by utilization?

I don't think so. For the current, we're measuring the contract's fee appreciation divided across all shares, so this is already handled by the contract's internal math.

mrice32 (approved):
LGTM!

packages/sdk/src/across/constants.ts
```diff
@@ -34,3 +48,12 @@ export const RATE_MODELS = {
     R2: "2000000000000000000",
   },
 };
+
+export function getRateModel(address: string): RateModelBN | undefined {
```
chrismaree, line 52: do we have to do this? it feels like this is somewhat unnecessary and adding the extra type is a bit confusing.

packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -253,6 +256,15 @@ function joinPoolState(
     SECONDS_PER_YEAR
   );
   const estimatedApr = calcApr(exchangeRatePrevious, exchangeRateCurrent, secondsElapsed, SECONDS_PER_YEAR);
+  let projectedApr = "N/A";
+  const rateModel = getRateModel(poolState.l1Token);
+  if (rateModel) {
+    projectedApr = fromWei(
```
chrismaree, line 262: this looks great.

chrismaree (approved):
this looks good! just seems a bit unnecessary to have a whole extra type and method for this.

packages/sdk/src/across/constants.ts
daywiss: yea true, its just that the function required big nums and i didnt want the constants to be Bignums. i can think of a better way to do this

daywiss:
I adjusted the fee calculator functions to accept rate models which are not purely bignumber, this makes it more synergistic with the way its defined in the constants file.

PR #3656, opened 2021-12-02
feat(across-bots): add profitability module to only relay profitable relays

chrismaree:
**Motivation**

The relayer should only do relays that are profitable. this PR adds a new module that calculates relay profitability based on token price.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/helpers/GasEstimator.ts
```diff
@@ -143,12 +152,32 @@ export class GasEstimator {
...
+    // Fetch the gasPrice info from the external provider and fetch the base fee from the web3 instance. Note that the
+    // gas price provider on mainnet (etherchain) does provide a base fee but in practice it seems to lag behind the
+    // most recent blocks somewhat. Using the web3 provider is more direct. Also Note that if this is not on L1 then the
+    // web3.eth.getGasPrice call will return the current gas price, not base fee. finally, note that if web3 was not
+    // defined then the base fee will be 0.
+    const [gasInfo, baseFee] = await Promise.all([
+      this._getPrice(this.networkId),
+      this.web3 ? this.web3.eth.getGasPrice() : null, // If a web3 instance was provided, fetch the base fee.
+    ]);
+    this.latestBaseFee = Number(baseFee);
```
nicholaspai, line 175: What if `baseFee == null`? Does this set `this.latestBaseFee = 0` and is that a problem?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,227 @@
...
+    const ethRevenue = this.getEthRevenue(tokenEthPrice, slowRevenue, speedUpRevenue, instantRevenue);
+
+    // Calculate the relay profitability as the difference between revenue and cost.
+    const ethProfitability = this.getProfit(ethSubmissionCost, ethRevenue);
+
+    // Finally, determine the relay submit type based on the profitability.
+    let relaySubmitType: RelaySubmitType;
+    if (
+      ethProfitability.instantEthProfit.gt(toBN(0)) &&
+      ethProfitability.instantEthProfit.gt(ethProfitability.speedUpEthProfit) &&
```
nicholaspai, line 123: nit: should these be `gte` or `gt` you think? I don't really know myself, but currently it seems like we prefer slow relays all things being equal, but shouldn't we prefer instant relays? I guess users would prefer our bots to prefer instant relays, while relayers would rather be slow?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,227 @@
...
+      case TokenType.ERC20:
+        return {
+          slowEThCost: gasPrice.mul(toBN(across.constants.SLOW_ERC_GAS)).mul(discount).div(fixedPoint),
+          speedUpEthCost: gasPrice.mul(toBN(across.constants.SPEED_UP_ERC_GAS)).mul(discount).div(fixedPoint),
+          instantEthCost: gasPrice.mul(toBN(across.constants.FAST_ERC_GAS)).mul(discount).div(fixedPoint),
+        };
+
+      case TokenType.UMA:
+        return {
+          slowEThCost: gasPrice.mul(toBN(across.constants.SLOW_UMA_GAS)).mul(discount).div(fixedPoint),
```
nicholaspai, line 177: Does UMA transfer really have that much of a different gas cost?

nicholaspai: Also, I've talked to @daywiss about this in a previous PR but we shouldn't have to hardcode these gas costs, can't we use some sort of `contract.methods.estimateGas` to get rid of the `enum TokenType`?

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -105,6 +109,7 @@ export class Relayer {
         try {
           relayTransactions.push(await this._generateRelayTransactionForPendingDeposit(l1Token, relayableDeposit));
         } catch (error) {
+          console.log("error", error);
```
nicholaspai, line 112: leftover log

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,227 @@
...
+      cumulativeGasPrice: cumulativeGasPrice.toString(),
+      relayerDiscount: fromWei(this.relayerDiscount),
+      slowRevenue: slowRevenue.toString(),
+      speedUpRevenue: speedUpRevenue.toString(),
+      instantRevenue: instantRevenue.toString(),
+    });
+
+    // Calculate the transaction submission costs depending on the gas price and the token type being relayed. This
+    // considers the relayer discount, applied to all relay actions equally.
+    const ethSubmissionCost = this.getRelayEthSubmissionCost(cumulativeGasPrice, tokenType);
```
nicholaspai, line 111: I like splitting up the profit decision into revenue and cost computations

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -585,6 +593,7 @@ export class Relayer {
   }> {
     try {
       await this.gasEstimator.update();
+      console.log("config", { ...this.gasEstimator.getCurrentFastPrice(), from: this.account });
```
nicholaspai, line 596: leftover log

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -700,9 +704,10 @@ export class Relayer {
     hasInstantRelayer: boolean
   ) {
     const mrkdwn = this._generateMarkdownForRelay(relayableDeposit.deposit, realizedLpFeePct);
+    console.log("shouldRelay", shouldRelay);
```
nicholaspai, line 707: left over log

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -700,9 +704,10 @@ export class Relayer {
...
     switch (shouldRelay) {
       case RelaySubmitType.Ignore:
-        this.logger.debug({
+        this.logger.warn({
```
nicholaspai, line 710: +1

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -75,11 +75,20 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
+      await l1Web3.eth.getChainId(),
+      config.relayerDiscount
+    );
+
     const relayer = new Relayer(
       logger,
       gasEstimator,
       l1Client,
       l2Client,
+      profitabilityCalculator,
```
nicholaspai, line 91: Doesn't `profitabilityCalculator.update()` need to be added to the first `await Promise.all([])` in the main infinite loop?

packages/sdk/src/across/constants.ts
```diff
@@ -8,6 +8,10 @@ export const FAST_ETH_GAS = 273519;
 export const FAST_ERC_GAS = 281242;
 export const FAST_UMA_GAS = 305572;
 
+export const SPEED_UP_ETH_GAS = 195288;
```
nicholaspai, line 11: I really hate hardcoding these, especially when its so easy to estimate gas via contract.methods.estimateGas

nicholaspai (commented):
The `ProfitabilityCalculator` module is super clean. I've left some nits, please re-request when you address them all!

packages/insured-bridge-relayer/src/index.ts
chrismaree: yup! good catch

packages/sdk/src/across/constants.ts
chrismaree: ye, it's not ideal. the main argument for doing it like this is to make the bots compute gas the same way the front end does.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: I think `gte` just on the instant makes sense as you are right we should always instant relay if posible.

packages/financial-templates-lib/src/helpers/GasEstimator.ts
chrismaree: yup. it would be null. in that case base fee would indeed be 0. I can change the `getCurrentBaseFee` and `getExpectedCumulativeGasPrice` to return an error if it's 0, wdyt?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,227 @@
...
+    let relaySubmitType: RelaySubmitType;
+    if (
+      ethProfitability.instantEthProfit.gt(toBN(0)) &&
+      ethProfitability.instantEthProfit.gte(ethProfitability.speedUpEthProfit) &&
+      ethProfitability.instantEthProfit.gte(ethProfitability.slowEthProfit)
+    )
+      relaySubmitType = RelaySubmitType.Instant;
+    else if (
+      ethProfitability.speedUpEthProfit.gt(toBN(0)) &&
+      ethProfitability.speedUpEthProfit.gt(ethProfitability.slowEthProfit)
```
nicholaspai, line 129: Should this be a `gte` too?

nicholaspai: Also, how can there be a positive `speedUpEthProfit` AND a positive `slowEthProfit`? I.e., a speed up relay should only be possible if a slow relay is pending, and therefore a slow relay should be impossible?

packages/sdk/src/across/constants.ts
nicholaspai: Yeah agreed lets leave for now

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
nicholaspai: agreed

packages/financial-templates-lib/src/helpers/GasEstimator.ts
nicholaspai: Is it ok to return 0 as a base fee? Otherwise we should throw an error yes

nicholaspai:
I've resolved all comments that you addressed!

packages/sdk/src/across/feeCalculator.ts
```diff
@@ -11,6 +17,13 @@ export interface RateModel {
   R2: BigNumberish; // R_0+R_1+R_2 is the interest rate charged at 100% utilization
 }
 
+export interface Web3RateModel {
+  UBar: Web3BN; // denote the utilization kink along the rate model where the slope of the interest rate model changes.
+  R0: Web3BN; // is the interest rate charged at 0 utilization
+  R1: Web3BN; // R_0+R_1 is the interest rate charged at UBar
+  R2: Web3BN; // R_0+R_1+R_2 is the interest rate charged at 100% utilization
+}
```
daywiss, line 25: it might clean things up to treat rate models as strings in wei rather than big numbers, that way you can import them from the constants file and pass them into the calculator as is. I adjusted the calculator to use BigNumberish which allows for strings, numbers or bignumbers.  Most of your definitions of the rate models use `toBNWei` which you could probably just change to `toWei` and it should work unless you are doing math on this outside of the sdk

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,227 @@
...
+    speedUpEthCost: BN;
+    instantEthCost: BN;
+  } {
+    const discount = fixedPoint.sub(this.relayerDiscount);
+    switch (tokenType) {
+      case TokenType.WETH:
+        return {
+          slowEThCost: gasPrice.mul(toBN(across.constants.SLOW_ETH_GAS)).mul(discount).div(fixedPoint),
+          speedUpEthCost: gasPrice.mul(toBN(across.constants.SPEED_UP_ETH_GAS)).mul(discount).div(fixedPoint),
+          instantEthCost: gasPrice.mul(toBN(across.constants.FAST_ETH_GAS)).mul(discount).div(fixedPoint),
```
daywiss, line 165: just FYI these constants might not be correct, francesco found out that the constants need to be subtracted from each other to get the correct cost, which i didnt realize when i set these constants. See here, hes doing math on the constants: https://github.com/across-protocol/across-frontend/blob/master/src/utils/bridge.ts#L70

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
daywiss: i think its good to have some default values, especially for frontend where its better to have the option to get a syncronous value, but if its easy to do i think it would be great to additionally add a way to get this cost dynamically

packages/sdk/src/across/feeCalculator.ts
chrismaree: ye, that's a great point. dealing with two different types of big number (web3 & ethers/BigNumber.js) is a _huge_ pain. I'll move to using strings ideally

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: ye so the subtraction thing that Francesco did we did together IRL. that was to find the cost of the speedup, over and above the slow relay cost. he then later adds these together to get the overall "fast" cost here: https://github.com/across-protocol/across-frontend/blob/fbf8ea1cc21a6c93b8bb9305e760560c371d8810/src/utils/bridge.ts#L101

I dont think this is _strictly_ correct. Ideally, we do no addition in the front end and quote the user (and compute the slow/instant rewards) using the full amount on `FAST_x_GAS` and do no addition. @mrice32 computed these constants by looking at the _cumlative_ gas used over a full relay, including the `relayAndSpeedUp` call and finally the `settleRelay` call later on.

chrismaree: This does bring up another point that @mrice32 and myself were discussing earlier on this week. right now, the front end considers the gas price using a diffrent priority fee to what the bots will use. this means that in general the front end will be _under_ quoting the cost of relays, given the bots always use a fast priority fee. we should consider changing this in the front end to more accurately represent what the bots will be paying.

daywiss: got it good to know

packages/sdk/src/across/feeCalculator.ts
chrismaree: please take a look at how I do it now. I think it's much cleaner now

mrice32 (commented):
Really awesome work! Flushing some comments, but haven't gotten through everything yet.

packages/financial-templates-lib/src/helpers/GasEstimator.ts
chrismaree: well no. if it is not london then `web3.eth. getGasPrice` returns the most recent gas price.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: ye, it should be gte. it's really pretty immaterial here though as these are wei numbers.

regarding how there could be a positive speedUpEthProfit AND a positive slowEthProfit, I agree that logically that should not happen due to how the revenues are calculated.

would you agree that this whole block can be reduced to this:

```ts
if (
      ethProfitability.instantEthProfit.gt(toBN(0)) &&
      ethProfitability.instantEthProfit.gte(ethProfitability.speedUpEthProfit) &&
      ethProfitability.instantEthProfit.gte(ethProfitability.slowEthProfit)
    )
      relaySubmitType = RelaySubmitType.Instant;
    else if (ethProfitability.speedUpEthProfit.gt(toBN(0))) relaySubmitType = RelaySubmitType.SpeedUp;
    else if (ethProfitability.slowEthProfit.gt(toBN(0))) relaySubmitType = RelaySubmitType.Slow;
    else relaySubmitType = RelaySubmitType.Ignore;
```

packages/financial-templates-lib/src/helpers/GasEstimator.ts
```diff
@@ -143,12 +152,32 @@ export class GasEstimator {
...
   async _update() {
-    const latestGasInfo = await this._getPrice(this.networkId);
+    // Fetch the gasPrice info from the external provider and fetch the base fee from the web3 instance. Note that the
+    // gas price provider on mainnet (etherchain) does provide a base fee but in practice it seems to lag behind the
+    // most recent blocks somewhat. Using the web3 provider is more direct. Also Note that if this is not on L1 then the
+    // web3.eth.getGasPrice call will return the current gas price, not base fee. finally, note that if web3 was not
+    // defined then the base fee will be 0.
+    const [gasInfo, baseFee] = await Promise.all([
+      this._getPrice(this.networkId),
+      this.web3 ? this.web3.eth.getGasPrice() : null, // If a web3 instance was provided, fetch the base fee.
```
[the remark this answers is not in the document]
chrismaree: from experimentation using this method (I created a simple script that logged it out) over 100 blocks it looks almost exactly the same as what etherscan was reporting the base fee to be. (granted this _almost_ the same could have been the _total gas price_, like you point out, which included the priority fee)

packages/common/src/ObjectUtils.ts
```diff
@@ -44,4 +44,18 @@ export const createObjectFromDefaultProps = (overrideProps: ObjectType, defaultP
...
-module.exports = { createObjectFromDefaultProps };
+/**
+ * @notice Given an`object`, and a `mapFn`, returns a new Object having applied the mapping function to each element in
+ * the `object`. Analogous to Array.prototype.map(), but for Objects.
+ * @param {object} the Object to map over.
+ * @param {mapFn} the mapping function to apply to each element in the `object`.
+ * @returns {object} new Object with the same keys as the `object`, but each value is the result of applying the mapFn.
+ */
+export const objectMap = (object: ObjectType, mapFn: any): ObjectType => {
+  return Object.keys(object).reduce(function (result: any, key) {
```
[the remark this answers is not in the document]
chrismaree: yup! that worked just fine.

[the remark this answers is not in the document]
chrismaree: prehaps I'm missing how the rest of these params should be used though. where do I use `T` now?

image

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: also note that the `shouldRelay` method would have set the `speedUpRevenue` to 0 in the event that the relay already had been slow relayed.

chrismaree: in fact after thinking about this more I think we should only instant relay iff the profit is greater than (not greater than or equal) to the slow eth profit. Reason is if these are equal then the depositor has set the instant relay fee to 0 and in this case the relayer should do a slow relay.

I also modified the `shouldRelay` method. it now has:

`instantRevenue = hasInstantRelayer ? toBN(0) : slowRevenue.add(speedUpRevenue);` such that we only generate an instant revenue if the relay has not been instant relayed.

packages/financial-templates-lib/src/helpers/GasEstimator.ts
mrice32: Ahhh, that makes sense. My bad. 

What do you think about having an `if (this.type == NetworkType.London)` at the top of this method and making different calls and setting different variables depending on the type. It's a little confusing (IMO) to overload a variable called latestBaseFee to store the average gas price for non-london chains.

Then we could maybe (?) just have an interface that has methods for:
1. An estimate of the gas price that you will actually pay.
2. The max gas price you will pay.
3. The suggested transaction params that you should pass to web3.

These could all be computed differently depending on the type, but regardless of london or not, they should be computable, right? Note: this could be already what you're doing and I could just be misunderstanding. Thoughts?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
nicholaspai: Agreed

packages/insured-bridge-relayer/src/index.ts
nicholaspai: I still don't see this in `update` have you pushed up all changes?

nicholaspai (commented):
Have you pushed up all local commits? https://github.com/UMAprotocol/protocol/pull/3656/files/b663ffaa4ae8517324a16e89a4bbb9bf9f085041#r763330602

packages/insured-bridge-relayer/src/index.ts
chrismaree: it's there sir. see line 118.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+    const ethRevenue = this.getEthRevenue(tokenEthPrice, slowRevenue, speedUpRevenue, instantRevenue);
+
+    // Calculate the relay profitability as the difference between revenue and cost.
+    const ethProfitability = this.getProfit(ethSubmissionCost, ethRevenue);
+
+    // Finally, determine the relay submit type based on the profitability.
+    let relaySubmitType: RelaySubmitType;
+    if (
+      ethProfitability.instantEthProfit.gt(toBN(0)) &&
+      ethProfitability.instantEthProfit.gt(ethProfitability.slowEthProfit)
```
nicholaspai, line 137: Did we discuss using `gte` here?

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -122,6 +124,9 @@ export class RelayerConfig {
 
     assert(this.crossDomainFinalizationThreshold < 100, "CROSS_DOMAIN_FINALIZATION_THRESHOLD must be < 100");
 
+    this.relayerDiscount = RELAYER_DISCOUNT ? Number(RELAYER_DISCOUNT) : 0;
+    assert(this.relayerDiscount >= 0 && this.relayerDiscount <= 100, "RELAYER_DISCOUNT must be between 0 and 100");
```
nicholaspai, line 128: Can you merge master in and replace this assert with a throw?

nicholaspai (commented):
Left another round of smaller comments

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: please see here: https://github.com/UMAprotocol/protocol/pull/3656#discussion_r762763723

packages/common/src/ObjectUtils.ts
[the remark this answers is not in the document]
mrice32: Gonna try to pull this down and test locally.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+   * @param {Number} l1ChainId L1 chain id that the relayer is running on.
+   * @param {Number} relayerDiscount Relayer discount applied to relays to offset gas cost.
+   */
+  constructor(
+    readonly logger: winston.Logger,
+    readonly l1Tokens: string[],
+    readonly l1ChainId: number,
+    readonly relayerDiscountNumber: number = 0
+  ) {
+    this.relayerDiscount = toBNWei(relayerDiscountNumber).div(toBN("100"));
```
mrice32, line 41: Will this work? I thought BNs couldn't represent decimals?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+    if (Object.keys(this.l1TokenInfo).length == 0) {
+      const [umaAddress, wethAddress] = await Promise.all([
+        await getAddress("VotingToken", this.l1ChainId),
+        await getAddress("WETH9", this.l1ChainId),
+      ]);
+      for (const l1Token of this.l1Tokens) {
+        this.l1TokenInfo[l1Token] = { tokenType: TokenType.ERC20, tokenEthPrice: toBNWei("1") };
+        if (l1Token == umaAddress) this.l1TokenInfo[l1Token].tokenType = TokenType.UMA;
+        else if (l1Token == wethAddress) this.l1TokenInfo[l1Token].tokenType = TokenType.WETH;
+        else this.l1TokenInfo[l1Token].tokenType = TokenType.ERC20;
```
mrice32, line 61: Is this else needed? Looks like it already defaults to the ERC20 type.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+
+    // Fetch the prices of each token, denominated in ETH. If coingecko does not have the price or is down then this
+    // call will fail and the tokenPrices array will have this error.
+    const tokenPrices = await Promise.allSettled(
+      this.l1Tokens.map((l1Token) => this.coingecko.getCurrentPriceByContract(l1Token, "eth"))
```
mrice32, line 68: 1. Should we retry?
2. I think coingecko rate limits requests. Does this CG client make sure that we don't send requests too fast? Or do we need to manage that outside of the CG client?

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+  ): {
+    slowEThCost: BN;
+    speedUpEthCost: BN;
+    instantEthCost: BN;
+  } {
+    const discount = fixedPoint.sub(this.relayerDiscount);
+    switch (tokenType) {
+      case TokenType.WETH:
+        return {
+          slowEThCost: gasPrice.mul(toBN(across.constants.SLOW_ETH_GAS)).mul(discount).div(fixedPoint),
```
mrice32, line 172: Do you think this would be easier to read if we put all of these constants into a map, like:
```ts
const costConstants = {
  [TokenType.WETH]: {
    slowEthGas: across.constants.SLOW_ETH_GAS
    ...
```

Then down here, you could use your object mapper or some other mechanism to write the math once?

Thoughts? This just seems a little repetitive.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,236 @@
...
+    this.logger.debug({
+      at: "ProfitabilityCalculator",
+      message: "Updated prices",
+      tokenInfo: objectMap(this.l1TokenInfo, (value: { tokenType: TokenType; tokenEthPrice: BN }) => {
+        return { tokenType: TokenType[value.tokenType], tokenEthPrice: fromWei(value.tokenEthPrice) };
+      }),
+    });
+  }
+
+  getRelaySubmitTypeBasedOnProfitability(
```
mrice32, line 99: Is this the method that users call?

I could totally be wrong here, but if I were a caller, I would prefer for this class to _just_ tell me how profitable each action is and then I could decide which one to do.

For instance, what happens if the relay has already been slow relayed and _all_ I can do is speed it up. If this module told me to send an instant relay, that doesn't tell me anything about the profitability of a speed up, right? Seems like the caller might have other constraints to consider along with profitability that this module can't take into account.

Thoughts?

mrice32 (commented):
Looks awesome!

packages/financial-templates-lib/src/helpers/GasEstimator.ts
chrismaree: ok so I've got for a minimal implementation that achieves what you laid out without needing to make big changes to the GasEstimator and while considering the recommendations you had in https://github.com/UMAprotocol/protocol/pull/3656#discussion_r761994730

now, on `update`, we fetch the most "latest" block, if there is a web3 and this is london. Then, if london, we extract the `latestBaseFee` and store it.

When calling `getExpectedCumulativeGasPrice` it returns:
1) `latestBaseFee + latestMaxPriorityFeePerGasGwei` if london and
2) `lastFastPriceGwei` if we are on any non-london chain.

`lastFastPriceGwei` seems to be the most accurate version of the expected amount actually paid if it's not london as this is the prevailing gas price the bot will pay. no additional case work is needed.

Please let me know if you agree with this! I really want to keep this as simple as possible and I feel this archives this.

packages/financial-templates-lib/src/helpers/GasEstimator.ts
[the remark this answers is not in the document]
chrismaree: after thinking more about this, we now use the `baseFeePerGas` on the most recent block, like you proscribed.

packages/insured-bridge-relayer/src/RelayerConfig.ts
chrismaree: done.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: this is not storying a decimal. it is storing a Wei'd BN. `relayerDiscountNumber` is a number between 0 and 100. this is numerically taking relayerDiscountNumber*1e18/100. so if `relayerDiscountNumber` was 50 (for 50%) this would generate 50*1e18/100= 5E17 to represent 50%. 

you are right though that `relayerDiscountNumber` could be a fraction would cause issues on this. I have updated it to round down to the nearest number to avoid this.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: true! done

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: it is possible CG rate limits this. the client does not do any rate-limiting. I think it's fine for now (from my testing we can easily do 5 tokens at once). if we hit limits here we can slow down this call with a bluebird promise or add in retrys.

I added a comment to this effect.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: ye, it is a tradeoff on where that decision-making happens. this chunk of logic at the end here ( from line 134 to line 142) could easily sit in the `Relater.ts` code block (in fact that part of code was originally sitting in the relayer and I moved it here!). 

my thinking in this was that all decision-making is informed exclusively by profit, in all cases. in your example where the relay has already been slow relayed and all that can be done is sped up then the inputs to this method would have `speedUpRevenue` and `instantRevenue` set to zero so the only profitable action that would be returned would be `RelaySubmitType.SpeedUp` (note there a number of unit tests that show this particular case (as well as others) in the `ProfitabiliyCalculator.ts` tests).

At the end of the day, it's just a question of where the logic sits. the assumption is the caller has already done a fair amount of logic on the revenues for each action according to what _is_ possible and that influences the profit computation.

You do have a good argument on returning profit from this module then doing the casework within the client. I'm happy to move it there if you feel strongly about this.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: sure. I did this + a bit more of a refactor to make it even fewer lines.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
nicholaspai: Ok I agree with this

nicholaspai (approved):
All my comments are resolved so LGTM after you respond to David and Matt's comments

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
mrice32: I see. That makes sense. A few quick points:
- Seems a little convoluted to have the relayer signal impossible cases by sending 0. But it does make sense.
- If the relayer never sees profitability, this limits our flexibility to do things with that information, like log the details in an error (since we would probably want to understand that we chose not to do the relay and by what margin was that decision made).

The latter bullet is the one I care about the most, since if we really are seeing an influx of underpriced txns, we'll need to have well-tailored errors and logs to tell us what's happening and allow us to potentially override the bot's decisions. Do you think that's possible in the current setup? If so, I'm not strongly opinionated about changing this structure.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,226 @@
...
+  getRelayEthSubmissionCost(
+    gasPrice: BN,
+    tokenType: TokenType
+  ): {
+    slowEThCost: BN;
+    speedUpEthCost: BN;
+    instantEthCost: BN;
+  } {
+    const cost = across.constants;
+    const costConstants = {
```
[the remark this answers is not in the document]
chrismaree: sure and done.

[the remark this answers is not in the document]
chrismaree: I did not make it static as it meant I had to refer to the variable as `ProfitabilityCalculator.costConstants` which made the code cover way more lines than it should for this small thing.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -0,0 +1,226 @@
...
+      speedUpRevenue: speedUpRevenue.toString(),
+      instantRevenue: instantRevenue.toString(),
+    });
+
+    // Calculate the transaction submission costs depending on the gas price and the token type being relayed. This
+    // considers the relayer discount, applied to all relay actions equally.
+    const ethSubmissionCost = this.getRelayEthSubmissionCost(cumulativeGasPrice, tokenType);
+
+    // Calculate the expected revenue, in ETH, based on the amount of tokens being offered in fees.
+    const ethRevenue = this.getEthRevenue(tokenEthPrice, slowRevenue, speedUpRevenue, instantRevenue);
```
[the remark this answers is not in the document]
chrismaree: sure. I added a block to check if the token price is zero then it should error. the one case where this should not happen is if the relayerDiscount is at 100% (i.e we don't care at all about the profitability). in this case, we should still be able to continue with the relay.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: So regarding logs the way it's structured right now I actually thing provides far, far more detailed logs (this is part of the reason why I chose this pattern). There are two main logs that are produced during the profitability evaluation:
a) `Considering relay profitability` that is produced at the beginning and  b) `Considered relay profitability` which is produced at the end. BOth of these logs contain _all_ state data the `ProfitabilityCalculator` has on this particular token including. The first log contains:
```ts
	  l1Token,
      tokenType: TokenType[tokenType],
      tokenEthPrice: fromWei(tokenEthPrice),
      cumulativeGasPrice: cumulativeGasPrice.toString(),
      relayerDiscount: fromWei(this.relayerDiscount),
      slowRevenue: slowRevenue.toString(),
      speedUpRevenue: speedUpRevenue.toString(),
      instantRevenue: instantRevenue.toString(),
```

and the second log contains:
```ts
      l1Token,
      tokenType: TokenType[tokenType],
      tokenEthPrice: fromWei(tokenEthPrice),
      cumulativeGasPrice: cumulativeGasPrice.toString(),
      relayerDiscount: fromWei(this.relayerDiscount),
      ethSubmissionCost: objectMap(ethSubmissionCost, (value: BN) => fromWei(value)),
      ethRevenue: objectMap(ethRevenue, (value: BN) => fromWei(value)),
      ethProfitability: objectMap(ethProfitability, (value: BN) => fromWei(value)),
      relaySubmitType: RelaySubmitType[relaySubmitType],
```

This is far more info than what would be possible if this decision making happened within the relayer as the relayer would not have information about: 1) tokenEthPrice, 2) relayerDiscount, 3) cumulativeGasPrice or 4) ethSubmissionCost. 

Put differently, by keeping this logic here _all_ information is processed in the same place enabling for more detailed logs to help with debugging.

packages/common/src/ObjectUtils.ts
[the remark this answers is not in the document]
chrismaree: ok I was able to get close but not the whole way to your explicit definition. It now looks like this:
```
export const objectMap = <T extends ObjectType, U>(object: T, mapFn: (T: any) => U): { [key: string]: U } => {
  return Object.keys(object).reduce(function (result: { [key: string]: U }, key) {
    result[key] = mapFn(object[key]);
    return result;
  }, {});
};
```

I could not get the last part (`T:any` is what I replaced) to correctly map to those arguments. This is fine for now methinks. we can improve it later.

packages/common/src/ObjectUtils.ts
```diff
@@ -44,4 +44,18 @@ export const createObjectFromDefaultProps = (overrideProps: ObjectType, defaultP
...
+ * @param {object} the Object to map over.
+ * @param {mapFn} the mapping function to apply to each element in the `object`.
+ * @returns {object} new Object with the same keys as the `object`, but each value is the result of applying the mapFn.
+ */
+export const objectMap = <T extends ObjectType, U>(object: T, mapFn: (T: any) => U): { [key: string]: U } => {
+  return Object.keys(object).reduce(function (result: { [key: string]: U }, key) {
+    result[key] = mapFn(object[key]);
+    return result;
+  }, {});
+};
```
mrice32, line 59:
```suggestion
export const objectMap = <T extends ObjectType, U>(object: T, mapFn: (arg: T[string]) => U): { [key: string]: U } => {
  return Object.keys(object).reduce(function (result: { [key: string]: U }, key) {
    result[key] = mapFn(object[key]);
    return result;
  }, {});
};
```

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
mrice32: But these are all debug logs, right? We won't be able to relay this info to slack unless we have _context_ about whether the decision is notable enough that it needs to be escalated.

That context is what I think may be missing in the profitability calculator.

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
mrice32: Ah I missed the Wei part of the BN conversion.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -106,6 +110,7 @@ export class Relayer {
           relayTransactions.push(await this._generateRelayTransactionForPendingDeposit(l1Token, relayableDeposit));
         } catch (error) {
           this.logger.error({ at: "AcrossRelayer#Relayer", message: "Unexpected error processing deposit", error });
+          console.log("ERROR", error);
```
mrice32, line 113: nit console.log

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
chrismaree: ye, you right. these logs are solely sent as info in the profitability calculator. this means that once the call has left the profitability calculator this context is lost and all we have in the `Relayer` was the _outcome_ and no info on why that was made (that was sent as an info log in the Profitability calculator). I see your point about wanting that context in the Relayer, in particular, to inform the log sent when a relay is deemed unprofitable such as `Not relaying potentially unprofitable deposit, or insufficient balance 😖`. it might be useful to have the additional profitability logic in that log showing the costs.

however, to get that to work cleanly I think we'll need to refactor some of the flow in the `Relayer.ts` file to more consistently pass this information around.

I propose for now we leave it as this PR has it and make an overall refactor down the line to better pass around this data. this PR is already almost 800 lines and adding in better data preservation for this component will be pretty considerable (likely over 200 lines). is that ok for you?

packages/insured-bridge-relayer/src/Relayer.ts
chrismaree: ye, good catch. this was an artifact of my painful debugging with some flaky unit tests! have removed.

packages/common/src/ObjectUtils.ts
```diff
@@ -44,4 +44,18 @@ export const createObjectFromDefaultProps = (overrideProps: ObjectType, defaultP
...
 
-module.exports = { createObjectFromDefaultProps };
+/**
+ * @notice Given an`object`, and a `mapFn`, returns a new Object having applied the mapping function to each element in
+ * the `object`. Analogous to Array.prototype.map(), but for Objects.
+ * @param {object} the Object to map over.
+ * @param {mapFn} the mapping function to apply to each element in the `object`.
+ * @returns {object} new Object with the same keys as the `object`, but each value is the result of applying the mapFn.
+ */
+export const objectMap = <T, U>(object: T, mapFn: (arg: T[keyof T]) => U): { [Property in keyof T]: U } => {
```
mrice32, line 54: @chrismaree took a little extra work, but this should be working now. I have trouble intuiting these generic functions without compiling them myself. Hope you don't mind, I just pushed up the change to fix it.

mrice32: The result of this function working is typescript being able to correctly infer return types as shown here:

Screen Shot 2021-12-09 at 2 49 53 PM

Screen Shot 2021-12-09 at 2 50 00 PM


Note: it's a little confusing bc the input and prop have the same name, but the second image is the property that you're setting in the log object. This isn't too impactful here, but it is more meaningful if you were to take these output values and use them for something that required the compiler to know that they're strings.

chrismaree: thanks @mrice32 ! I see what you did there, pretty clever using `keyof`!

daywiss (approved):
im just approving based on the small part i commented on

packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
mrice32: Seems reasonable -- I don't think it's a huge deal for now as the relayer can at least tell us when it ignores something if it wants to. It just won't be able to give us the numbers via slack.

mrice32 (approved):

PR #3657, opened 2021-12-02
feat(across): add sync caller helper

chrismaree:
**Motivation**

Small helper contract to make calling sync methods on a set of bridge pools easier. When one pool needs to be
 * sync it is likely that all ecosystem pools also need to be sync. This contract lets the caller do this in a batch.


packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
```diff
@@ -0,0 +1,32 @@
+pragma solidity ^0.8.0;
+
+interface BridgePoolLike {
+    function syncUmaEcosystemParams() external;
+
+    function syncWithBridgeAdminParams() external;
+}
+
+/**
+ * @notice Small helper contract to make calling sync methods on a set of bridge pools easier. When one pool needs to be
```
nicholaspai, line 10:
```suggestion
 * @notice Small helper contract to facilitate calling sync methods on a set of bridge pools. When one pool needs to be
```

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
```diff
@@ -0,0 +1,32 @@
...
+ * sync it is likely that all ecosystem pools also need to be synced. This contract lets the caller do this in a batch.
```
nicholaspai, line 11:
```suggestion
 * synced it is likely that all ecosystem pools also need to be synced. This contract lets the caller do this in a batch.
```

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
```diff
@@ -0,0 +1,32 @@
...
+ */
+contract BridgePoolSynchronizer {
+    function syncUmaEcosystemParams(BridgePoolLike[] memory bridgePools) public {
+        for (uint256 i = 0; i < bridgePools.length; i++) {
+            bridgePools[i].syncUmaEcosystemParams();
```
nicholaspai, line 16: nitty nit: what if some bridge pools we want to skip for some reason and get some gas-savings? Is that possible?

nicholaspai (approved):
This seems potentially very useful. Small nit

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
```diff
@@ -0,0 +1,32 @@
...
+    function syncUmaEcosystemParams() external;
+
+    function syncWithBridgeAdminParams() external;
+}
+
+/**
+ * @notice Small helper contract to make calling sync methods on a set of bridge pools easier. When one pool needs to be
+ * sync it is likely that all ecosystem pools also need to be synced. This contract lets the caller do this in a batch.
+ */
+contract BridgePoolSynchronizer {
```
mrice32, line 13: Seems like overkill to deploy a full contract just to send transactions like this. Why not use Multicall2 since we already have a pretty deep integration with that in our bots and in the SDK?

chrismaree: indeed; multicall is the right way to do this or using DSProxy. in any case though you still need to "implementation" that you will be calling. I'll add a script to this PR that lets you do this.

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
chrismaree: just dont put them in the array of `bridgePools` that you feed in

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
chrismaree: in fact using multicall means we dont actually need the solidity file. this is simply a script now :)

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
nicholaspai: true!

packages/core/contracts/insured-bridge/non-bridge-helpers/BridgePoolSynchronizer.sol
mrice32: 💯 Yeah, just didn't want to deploy a single-use contract!

mrice32 (approved):
LGTM! This is great!

PR #3659, opened 2021-12-02
improve(cryptowatch-pricefeed): Use twapLength from parsed ancillary data

nicholaspai:
**Motivation**

The CryptoWatchPriceFeed should be able to read `twapLength` as a parameter specified in ancillary data if available.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -91,11 +91,11 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     let twapLength = this.twapLength;
     // Ancillary data might contain parameters that affect how we compute the historical price, such as the twapLength.
-    try {
+    if (ancillaryData) {
       const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
-      twapLength = Number(parsedAncillaryData.twapLength);
-    } catch (err) {
-      // Could not parse ancillary data into utf8 dictionary, use default parameters.
+      if (parsedAncillaryData.twapLength) {
+        twapLength = Number(parsedAncillaryData.twapLength);
+      }
     }
 
     // Return early if computing a TWAP.
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -1,5 +1,5 @@
 import { PriceFeedInterface } from "./PriceFeedInterface";
-import { parseFixed, formatFixed } from "@uma/common";
+import { parseFixed, formatFixed, parseAncillaryData } from "@uma/common";
 import { computeTWAP } from "./utils";
 import type { Logger } from "winston";
 import Web3 from "web3";
@@ -11,6 +11,11 @@ interface PricePeriod {
   closeTime: number;
   openPrice: BN;
   closePrice: BN;
+}
+
+// Options for retrieving most prices such as TWAP length
+export interface SimplePriceAncillaryData {
+  twapLength: string;
 }
 // An implementation of PriceFeedInterface that uses CryptoWatch to retrieve prices.
 export class CryptoWatchPriceFeed extends PriceFeedInterface {
@@ -84,9 +89,14 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
       throw new Error(`${this.uuid}: undefined lastUpdateTime`);
     }
 
+    // Ancillary data might contain parameters that affect how we compute the historical price, such as the twapLength.
+    const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
+    const ancillaryDataTwapLength = Number(parsedAncillaryData.twapLength);
+
     // Return early if computing a TWAP.
-    if (this.twapLength) {
-      const twapPrice = this._computeTwap(time, this.historicalPricePeriods);
+    const twapLength = ancillaryDataTwapLength ? ancillaryDataTwapLength : this.twapLength;
+    if (twapLength) {
+      const twapPrice = this._computeTwap(time, this.historicalPricePeriods, twapLength);
       if (!twapPrice) {
         throw new Error(`${this.uuid}: historical TWAP computation failed due to no data in the TWAP range`);
       }
@@ -234,7 +244,7 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     const newHistoricalPricePeriods = await this._getOhlcPricePeriods(earliestHistoricalTimestamp, currentTime);
     const newPrice = this.twapLength
-      ? this._computeTwap(currentTime, newHistoricalPricePeriods)
+      ? this._computeTwap(currentTime, newHistoricalPricePeriods, this.twapLength)
       : await this._getImmediatePrice();
 
     // 5. Store results.
@@ -344,7 +354,8 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
   private _computeTwap(
     endTime: number,
-    ohlcs: { openTime: number; closeTime: number; openPrice: BN; closePrice: BN }[]
+    ohlcs: { openTime: number; closeTime: number; openPrice: BN; closePrice: BN }[],
+    twapLength: number
   ): BN | null {
     // Combine open and close to get more data fidelity at the edges of the range.
     const priceTimes = ohlcs
@@ -355,7 +366,7 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
         ] as [number, BN][];
       })
       .flat();
-    const startTime = endTime - this.twapLength;
+    const startTime = endTime - twapLength;
     const twapPrice = computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
 
     return this.invertPrice ? this._invertPriceSafely(twapPrice) : twapPrice;
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -89,12 +89,16 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
       throw new Error(`${this.uuid}: undefined lastUpdateTime`);
     }
 
+    let twapLength = this.twapLength;
     // Ancillary data might contain parameters that affect how we compute the historical price, such as the twapLength.
-    const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
-    const ancillaryDataTwapLength = Number(parsedAncillaryData.twapLength);
+    try {
+      const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
+      twapLength = Number(parsedAncillaryData.twapLength);
+    } catch (err) {
+      // Could not parse ancillary data into utf8 dictionary, use default parameters.
+    }
 
     // Return early if computing a TWAP.
-    const twapLength = ancillaryDataTwapLength ? ancillaryDataTwapLength : this.twapLength;
     if (twapLength) {
       const twapPrice = this._computeTwap(time, this.historicalPricePeriods, twapLength);
       if (!twapPrice) {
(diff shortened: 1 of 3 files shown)


Reinis-FRP:
`this.historicalPricePeriods` is saved in `update()` function where `earliestHistoricalTimestamp` is derived from `this.twapLength` coming from stored configuration. So if price request has longer TWAP in ancillary data than stored in PF configuration then there would not be full data available to compute the TWAP for the requested period.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -84,9 +89,18 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
     }
 
+    let twapLength = this.twapLength;
+    // Ancillary data might contain parameters that affect how we compute the historical price, such as the twapLength.
+    try {
+      const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
+      twapLength = Number(parsedAncillaryData.twapLength);
+    } catch (err) {
+      // Could not parse ancillary data into utf8 dictionary, use default parameters.
+    }
```
mrice32, line 99: Are we sure we want to silently use default params if the ancillary data can't be parsed? Maybe we should enumerate the cases:
- Ancillary data is empty. Seems fine to use default params in this case. Not sure if parseAncillaryData errors in this case, but if not, this would end up looking identical to the case below.
- Ancillary data doesn't contain the twapLength field.  Seems fine to use default params in this case.
- Ancillary data can't be parsed to UTF8. Seems like this should throw, right?
- Ancillary data has a value for twapLength that can't be parsed into a number. Seems like this should also throw, right?

Thoughts?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -234,7 +248,7 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
     const newHistoricalPricePeriods = await this._getOhlcPricePeriods(earliestHistoricalTimestamp, currentTime);
     const newPrice = this.twapLength
-      ? this._computeTwap(currentTime, newHistoricalPricePeriods)
+      ? this._computeTwap(currentTime, newHistoricalPricePeriods, this.twapLength)
```
mrice32, line 251: Do we not consider ancillary data in the update function? Is that a problem? Shouldn't the current price and the historical price be computed the same way?

Or is the assumption that there are no contracts that use current price that also have ancillary data? i.e. all ancillary data requests go through the OO.

mrice32 (commented):
Looks great! A few comments

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -84,9 +89,18 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
       throw new Error(`${this.uuid}: undefined lastUpdateTime`);
     }
 
+    let twapLength = this.twapLength;
+    // Ancillary data might contain parameters that affect how we compute the historical price, such as the twapLength.
+    try {
+      const parsedAncillaryData = (parseAncillaryData(ancillaryData) as unknown) as SimplePriceAncillaryData;
```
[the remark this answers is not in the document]
nicholaspai: It returns `Record<unknown, string>`?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: > Or is the assumption that there are no contracts that use current price that also have ancillary data? i.e. all ancillary data requests go through the OO.

This is the assumption. Also, currently the `update()` function is not designed to accept any ancillary data. How would that work you think? I do think we should refactor `getCurrentPrice` to take in `ancillaryData` soon

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: I agree with all these cases, and its pretty easy to implement actually: [commit](https://github.com/UMAprotocol/protocol/pull/3659/commits/7e257ee0f110ede8b1cd99b45150032d6c956a42)

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: This issue raised by @Reinis-FRP is related: https://github.com/UMAprotocol/protocol/pull/3659#issuecomment-984988812

How do we connect any `twapLength` specified in `ancillaryData` to the `update` method in the price feed?

nicholaspai: We could add an optional input param to `update` for `ancillaryData` so that anyone can "update" it for a specific `ancillaryData`. Is that reasonable..?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -70,23 +76,31 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
   }
 
   public getCurrentPrice(): null | BN {
-    if (!this.twapLength && this.invertPrice) {
```
nicholaspai, line 73: Reading `this.twapLength` is only neccessary in this method because we compute a TWAP in `_inversePriceSafely`, which is non-intuitive. I've changed this logic so we no longer need to read this here. This also is a problem to this PR because `update` now reads `twapLength` potentially from `ancillaryData`, but does not set the class's state variable `this.twapLength`.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -205,16 +219,21 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
...
+  public async update(ancillaryData?: string): Promise<void> {
     const currentTime = await this.getTime();
 
-    // Return early if the last call was too recent.
-    if (this.lastUpdateTime !== null && this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime) {
+    // Return early if the last call for this ancillary data was too recent.
+    if (
+      this.lastUpdateTime !== null &&
+      this.lastUpdateTime + this.minTimeBetweenUpdates > currentTime &&
+      ancillaryData === this.lastAncillaryData
```
nicholaspai, line 229: Now that `ancillaryData` can be passed into `update`, we should only exit prematurely if the last update was for a recent timestamp AND the ancillary data is the same as the last one queried

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -355,10 +386,10 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
         ] as [number, BN][];
       })
       .flat();
-    const startTime = endTime - this.twapLength;
+    const startTime = endTime - twapLength;
     const twapPrice = computeTWAP(priceTimes, startTime, endTime, this.web3.utils.toBN("0"));
 
-    return this.invertPrice ? this._invertPriceSafely(twapPrice) : twapPrice;
```
nicholaspai, line 361: There was really no advantage to inverting price in this method, Instead we should do it at the `get*Price` method level

chrismaree: this seems risky...where are you now doing the inversion?

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
mrice32: What about providing an extra arg to getPrice()?

mrice32: The reason I say this is that it seems a little difficult to reason about if you pass in the ancillary data into update: whenever you call getPrice, you need to know what ancillary data was passed in the last time someone called update. Depending on where we call update, the ancillary data could have been left blank or could have been different. This means that you _always_ just need to call update() and immediately follow with getPrice().

This isn't a strong opinion, but I think a more natural way to set values like this is either:
1. In the constructor and have it be immutable.
2. As an argument to getPrice, similar to getHistoricalPrice.

I'm guessing it's difficult to remove it from update() because there are async calls in there that rely on it. Maybe the right solution is to leave this as you have it for now, but add a task to drop update entirely and make getPrice async?

mrice32 (commented):
https://github.com/UMAprotocol/protocol/pull/3659#discussion_r763057613

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: > whenever you call getPrice, you need to know what ancillary data was passed in the last time someone called update. Depending on where we call update, the ancillary data could have been left blank or could have been different. This means that you always just need to call update() and immediately follow with getPrice().

This is an assumption we make for how `getCurrentPrice` works though, so I think this is very reasonable. `getCurrentPrice` will always return an unexpected price if the caller does not know when the last time `update` was called. 

However, I could see us adding `ancillaryData` to the constructor params. 

> the right solution is to leave this as you have it for now, but add a task to drop update entirely and make getPrice async for all price feeds

This 100% is true but we should save for future big PR

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
nicholaspai: There are currently three calls to `_computeTwap`:
- Line 102 in `getHistoricalPrice` which returns the TWAP'd historical price. This now inverts the return value of `_computeTwap`
- Line 266 in `update` which sets the price returned by `getCurrentPrice`. This value gets inverted in `getCurrentPrice` now, whereas previously it was being inverted in `_computeTwap`

I think this change makes much more sense and doesn't have a hidden inversion in `_computeTwap`

mrice32 (approved):

packages/financial-templates-lib/src/price-feed/PriceFeedInterface.ts
```diff
@@ -3,12 +3,15 @@ import { BN } from "../types";
 // Price feed interface -- all price feed implementations should override all functions (except for _abstractFunctionCalled).
 export abstract class PriceFeedInterface {
   // Updates the internal state of the price feed. Should pull in any async data so the get*Price methods can be called.
+  // Will use the optional ancillary data parameter to customize what kind of data get*Price returns.
   // Note: derived classes *must* override this method.
-  public abstract update(): Promise<void>;
+  // Note: Eventually `update` will be removed in favor of folding its logic into `getCurrentPrice`.
+  public abstract update(ancillaryData?: string): Promise<void>;
```
chrismaree, line 9: not ideal to have to include this but an ok workaround for now

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
chrismaree: that's true! invert where it's used rather than hiding it here.

chrismaree: I see that the inversion now happens on line 103 right after you get the price 👌

chrismaree (approved):
nice work.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
chrismaree: ye, I agree this is not ideal and a but clunky in some cases. it's very much strapped onto our existing implementation. 

I agree with both of you what an ideal future version of this would look like, but it should totally be ignored for the time being given how much work those changes would require.

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -344,7 +374,8 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
 
   private _computeTwap(
     endTime: number,
-    ohlcs: { openTime: number; closeTime: number; openPrice: BN; closePrice: BN }[]
+    ohlcs: { openTime: number; closeTime: number; openPrice: BN; closePrice: BN }[],
+    twapLength: number
```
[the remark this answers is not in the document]
nicholaspai: Yeah I actually think thats a good design. Each price feed should handle whatever keys in the ancillary data it wants to

PR #3662, opened 2021-12-06
feat(across-bots): add ability to finalize L2->L1 transfers from arbitrum

chrismaree:
**Motivation**

As part of integrating with multiple chains, we require the ability fo finalize configured L2->L1 transfers sent over the canonical bridge. This PR adds arbitrum L2->L1 finalization to the `CrossDomainFinalizer`.
**Summary**

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/common/src/TransactionUtils.ts
```diff
@@ -318,6 +318,7 @@ export async function findBlockNumberAtTimestamp(
  * catch reverts that could not be seen at submission time.
  */
 export async function processTransactionPromiseBatch(transactions: Array<ExecutedTransaction>, logger: winston.Logger) {
+  if (transactions.length == 0) return;
```
chrismaree, line 321: we produced logs in all cases this was called, even if there is no transactions to waint on.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -23,11 +23,17 @@ export interface Deposit {
...
+  bridgedTokensTxHash: string;
+  numberOfTokens: string;
+}
+
 export class InsuredBridgeL2Client {
   public bridgeDepositBox: BridgeDepositBoxWeb3;
 
   private deposits: { [key: string]: Deposit } = {}; // DepositHash=>Deposit
   private whitelistedTokens: { [key: string]: string } = {}; // L1Token=>L2Token
+  private tokensBridgedTransactions: { [key: string]: TokensBridged[] } = {}; // L2Token=>BridgeTransactionHash[]
```
chrismaree, line 36: store additional data to find L2->L1 transfers.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -64,6 +71,35 @@ export class InsuredBridgeL2Client {
     return this.deposits[depositHash];
   }
 
+  getTokensBridgedTransactionsForL2Token(l2TokenAddress: string) {
+    if (!this.tokensBridgedTransactions[l2TokenAddress]) return [];
```
chrismaree, line 75: accessor methods. some of these are a bit convoluted due to the data structure. I played with a good number of these and this implementation worked out to be the simplest version, mainly due to the data we need to store and access without adding additional async calls.

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -124,48 +167,22 @@ export class InsuredBridgeL2Client {
...
     const eventsData = await getEventsForMultipleProviders(
       [this.l2Web3].concat(this.fallbackL2Web3s),
       getAbi("BridgeDepositBox"),
       this.bridgeDepositAddress,
-      "FundsDeposited",
+      eventName,
       eventSearchOptions
     );
     if (eventsData.missingEvents.length > 0) {
-      const errorMessage = `L2 RPC endpoints disagree about L2 contract events, please manually investigate. L2 rpcs are described in NODE_URL and RETRY_CONFIG environment variables.`;
```
chrismaree, line 136: I also refactored how we make L2 calls to have a general `getBridgeDepositBoxEvents` that uses @nicholaspai 's `getEventsForMultipleProviders`. previously this was repeated for each internal call. now that there are 3 of them we should have a common function for doing this where you pass in the `eventName`

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -3,31 +3,37 @@ const { toBN } = Web3.utils;
...
 
-    // Fetch all WhitelistToken events on L1 and filter out the ones that are not on the associated L2Client
-    // chainId.
-    const whitelistedTokenMappings = this.l1Client.getWhitelistedTokensForChainId(this.l2Client.chainId.toString());
-
-    // Extract the l2Tokens that have been whitelisted. `whitelistedTokenMappings` is a key-value dictionary of L1 to
-    // L2 token addresses so we'll extract just the values.
-    const whitelistedL2Tokens = Object.values(whitelistedTokenMappings);
+    // Fetch all whitelisted tokens on the particular l2 chainId.
+    const whitelistedL2Tokens = this.l1Client.getWhitelistedL2TokensForChainId(this.l2Client.chainId.toString());
```
chrismaree, line 36: this was simplified somewhat due to the new method I added to the client.

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -406,7 +406,7 @@ export class Relayer {
     if (hasInstantRelayer && relayableDeposit.status == ClientRelayState.Pending) {
       this.logger.debug({
         at: "AcrossRelayer#Relayer",
-        message: "Relay pending and already sped up 😖",
```
chrismaree, line 409: this is a debug and should ot have an emoji

packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -873,7 +873,10 @@ export class Relayer {
 
       // Look up all blocks from contract deployment time to latest to ensure that a deposit, if it exists, is found.
       while (deposit === undefined) {
-        const fundsDepositedEvents = await this.l2Client.getFundsDepositedEvents(l2BlockSearchConfig);
```
chrismaree, line 876: this call uses the new generalized L2client method.

Note that we should consider removing this event query block from this `Relayer.ts` file.

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -88,8 +90,9 @@ export class RelayerConfig {
     this.botModes = {
       relayerEnabled: RELAYER_ENABLED === "true" ? true : false,
       disputerEnabled: DISPUTER_ENABLED === "true" ? true : false,
-      finalizerEnabled: FINALIZER_ENABLED === "true" ? true : false,
```
chrismaree, line 91: NOTE BREAKING CHANGE!

I renamed these variables and envs to be more descriptive, given we now have a new module.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/ArbitrumBridgeAdapter.ts
```diff
@@ -0,0 +1,121 @@
+import winston from "winston";
```
chrismaree, line 1: this class has the new logic for arbitrum bridging.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/BridgeAdapterInterface.ts
```diff
@@ -0,0 +1,8 @@
+import type { TransactionType } from "@uma/common";
```
chrismaree, line 1: common interface we can extend to in the future for other chains.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -52,8 +53,6 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
 
     // Construct a web3 instance running on L2.
     const l2Web3 = getWeb3ByChainId(config.activatedChainIds[0]);
-    const latestL2BlockNumber = await l2Web3.eth.getBlockNumber();
```
chrismaree, line 55: NOTE THIS CHANGE! @nicholaspai I made this as we need the ability to see `TokensBridged` events all the way back, past the `l2BlockLookback`. This means that this implementation _only_ works on Alchemy and NOT infura as we need to look back past what that provider can support. to address this we should update the L2client to facilitate pagination to work around this restriction.

let me know what you think!

packages/common/src/TransactionUtils.ts
nicholaspai: Is it overkill to log that there are 0 transactions before returning?

packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts
```diff
@@ -38,7 +44,8 @@ export class InsuredBridgeL2Client {
     readonly chainId: number = 0,
     readonly startingBlockNumber: number = 0,
     readonly endingBlockNumber: number | null = null,
-    readonly fallbackL2Web3s: Web3[] = []
+    readonly fallbackL2Web3s: Web3[] = [],
+    readonly l2BlockLookback: number = 0
```
nicholaspai, line 48: Why can't you use `startingBlockNumber` and or `endingBlockNumber` in place of `l2BlockLookback`? Arent' these redundant?

packages/insured-bridge-relayer/src/index.ts
nicholaspai: Hm I don't like that this change would force us to no longer work with Infura. Why don't you separate event querying between two flows:
- the usual getPastEvents call that goes from `{fromBlock: l2StartBlock, toBlock: latest}` for events like `FundsDeposited` 
- Add a second event querying path that paginates from `{fromBlock: _fromBlock, toBlock: _toBlock}`, and sits inside a `for` loop that keeps decrementing `_fromBlock` until `_fromBlock <= 0` and sets `_toBlock = _fromBlock`.

I think pagination is the way to go instead of cuttting off compatibility with Infura

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -84,8 +90,64 @@ export class CrossDomainFinalizer {
...
+
+  async checkForConfirmedL2ToL1RelaysAndFinalize() {
+    // Fetch all whitelisted L2 tokens.
+    const whitelistedL2Tokens = this.l1Client.getWhitelistedL2TokensForChainId(this.l2Client.chainId.toString());
+
+    // For each whitelisted L2 token, fetch all associated "TokensBridge" transaction hashes on L2. This will return an
+    // array of arrays, with each L2 token's transaction hashes being the nested array to each L2Token.
+    const l2TokensBridgedTransactions = whitelistedL2Tokens
+      .map((l2Token) => this.l2Client.getTokensBridgedTransactionsForL2Token(l2Token))
+      .flat(2) // Flatten the array to get a 1D array of all TokenBridged transaction hashes.
```
mrice32, line 102: I think you can leave this arg out and let this default to 1:
```
> [[1,2], [3,4]].flat()
[ 1, 2, 3, 4 ]
```

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -112,6 +174,49 @@ export class CrossDomainFinalizer {
...
+
+      await this.gasEstimator.update();
+      const executionResult = await runTransaction({
+        web3: this.l1Client.l1Web3,
+        transaction: finalizationTransaction,
+        transactionConfig: { ...this.gasEstimator.getCurrentFastPrice(), from: this.account },
+        availableAccounts: 1,
+        waitForMine: false,
+      });
+      if (executionResult.receipt) {
```
mrice32, line 193: When would we _not_ have a receipt? Doesn't the function above always return some non-null/undefined value as a receipt?

packages/insured-bridge-relayer/src/canonical-bridge-adapters/ArbitrumBridgeAdapter.ts
```diff
@@ -0,0 +1,121 @@
...
+    this.bridge = await Bridge.init(l1EthersVoidSigner, l2EthersVoidSigner);
+
+    this.logger.debug({ at: "ArbitrumBridgeAdapter", message: "Initialized Arbitrum Bridge Adapter" });
+  }
+  async constructCrossDomainFinalizationTransaction(
+    l2TransactionHash: string
+  ): Promise<{ l2TransactionHash: string; finalizationTransaction: TransactionType | null }> {
+    this._throwIfNotInitialized();
+
+    if (this.bridge) {
```
mrice32, line 34: Is this just a typescript thing? I assume we know this will always be non-undefined? If so, maybe we should make this an assert rather than an if? The only reason I suggest it is that the way it reads, you need to consider the case where this condition is false (which I think should never happen).

packages/insured-bridge-relayer/src/canonical-bridge-adapters/ArbitrumBridgeAdapter.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  async initialize() {
+    const l1EthersProvider = new providers.Web3Provider(this.l1Web3.currentProvider as any);
+    const l1EthersVoidSigner = new VoidSigner(ZERO_ADDRESS, l1EthersProvider);
+    const l2EthersProvider = new providers.Web3Provider(this.l2Web3.currentProvider as any);
+    const l2EthersVoidSigner = new VoidSigner(ZERO_ADDRESS, l2EthersProvider);
+
+    this.bridge = await Bridge.init(l1EthersVoidSigner, l2EthersVoidSigner);
+
+    this.logger.debug({ at: "ArbitrumBridgeAdapter", message: "Initialized Arbitrum Bridge Adapter" });
```
mrice32, line 27: Note: if these methods + _throwIfNotInitialized() checks are annoying, one thing you could do is use async factory functions to create classes with async logic.

packages/insured-bridge-relayer/src/index.ts
mrice32: Given that it needs a pretty different lookback and it's doing a different query, do you think it makes sense to do this logic somewhere outside of this client (maybe a separate client or maybe in the finalizer logic itself)?

Agree that pagination (and adding a utility for it!) seems like a good idea here. As long as we know how many blocks we're trying to go back, we could also do this pagination in parallel (with some limit to the number of in-flight requests using bluebird, maybe?) so we don't have to worry about this bot taking a long time to run due to the serialized queries.

Going back to 0 might not be ideal, though, especially if we're doing pagination.

mrice32 (commented):
Looks good!

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: that's true! removed

packages/insured-bridge-relayer/src/canonical-bridge-adapters/ArbitrumBridgeAdapter.ts
chrismaree: ye, ts was not happy seeing it can be null. I changed this to throw if null, which should never be the case, assuming you've initiated the adapter.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/ArbitrumBridgeAdapter.ts
chrismaree: good call out. I ended up just moving the check into the `constructCrossDomainFinalizationTransaction`  method seeing this the only place we use this.

packages/insured-bridge-relayer/src/index.ts
chrismaree: @nicholaspai agreed that pagination is the solution. the way this PR is structured right now, however, makes it not work with infura as we currently don't have the pagination logic. I also agree having two flows for different kinds of events seems useful. ideally though I'd like to create a `paginationEventLookup` or something equivalent that will abstract away this from the user so it does not have to sit in the client.

@mrice32 putting the query somewhere else does make sense! this is how we had structured things originally (the finalizer itself fetched the whitelist event data). we ended up rolling this into the client when we needed access to the same data in another location. I agree though in this context we wont really need the data anywhere else and pulling it in for all bots (even those that don't run the finalizer) is a bit of a waist on resources. Regarding how far we go back we don't need to go all the way to block 0, we can simply go back to `bridgeDepositBoxDeployData` from the config file https://github.com/UMAprotocol/protocol/blob/88bd694ad916cb3ee5e97102e27d073b1b9fdf67/packages/insured-bridge-relayer/src/RelayerConfig.ts?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L32

chrismaree: After looking more at this I think the best solution is to go with what @mrice32 described by pulling the event query into the `CrossDomainFinalizer`. this has a number of advantages, least of which we don't need to fetch any of this data unless you want to run this particular bot. this makes "normal" bot operation less taxing on nodes and user resources. 

while considering this, I also realized that we capute a fair amount of unnecessary data in the `L2Client`. in particular, we fetch `WhitelistToken` and store it within the client. this is only used with the method `isWhitelistedToken` (see [here](https://github.com/UMAprotocol/protocol/blob/05930397cd67e018bc09d6c5d8169514b365a4ee/packages/financial-templates-lib/src/clients/InsuredBridgeL2Client.ts?_pjax=%23js-repo-pjax-container%3Afirst-of-type%2C%20div%5Bitemtype%3D%22http%3A%2F%2Fschema.org%2FSoftwareSourceCode%22%5D%20main%3Afirst-of-type%2C%20%5Bdata-pjax-container%5D%3Afirst-of-type#L59)) but we never use this data anywhere in the bots. Moreover, the current implementation of this method will _not_ contain all expected tokens as the lookback, as injected into this client currently in master, prevents the L2 client from looking back much past ~100k events!

I've refactored this to do the following:
1) I removed the `isWhitelistedToken` and associated event retrieval and data from the L2Client, thereby making it lighter and easier to work with.
2) I refactored the fetching logic in the L2 client, previously called `getFundsDepositedEvents` to `getBridgeDepositBoxEvents` to make things more modular and clean.
3) I pulled the logic for fetching `TokensBridged` events _directly_ into the `CrossDomainFinalizer`. while this is a bit convoluted in that we fetch this data in a location that is _not_ a "client", it does make the bots lighter (only fetch the data if you run this module) and solves where the pagination logic should be.

chrismaree: I worked on creating a basic pagination block searcher. Logic looked basically like this(rough example from testing):
```ts
const currentBlock = await web3.eth.getBlockNumber();
  const firstBlock = 2811998;
  const maxBlockChunk = 99900;

  const paginationChunks = Math.ceil((currentBlock - firstBlock) / maxBlockChunk);

  let blockSearchIndecides = [];
  for (let i = 0; i < paginationChunks; i++) {
    blockSearchIndecides[i] = {
      fromBlock: firstBlock + maxBlockChunk * i + (i == 0 ? 0 : 1),
      toBlock: firstBlock + maxBlockChunk * (i + 1),
    };
    if (blockSearchIndecides[i].toBlock > currentBlock)
      blockSearchIndecides[i].toBlock = currentBlock;
  }

  const tokensBridgedEvents = await Promise.all(
    blockSearchIndecides.map((searchSettings) =>
      depositBox.getPastEvents('TokensBridged', searchSettings)
    )
  );
```

Issue is that this does not play nicely with our RPC endpoints! when run against infura this blows up the node and you get back an invalid JSON RPC response. I added `BluebirdPromise.all` to try and set maximum `concurrency` settings but I can only run maximum 1 concurrency at the same time on infura or this errors out :/ From when the contracts were deployed, for a chunk size of 99900 we need to make 10 sequential queries to build up the full event set. This takes almost 40 seconds.


for now, I think the solution should be to not do any pagination and simply run this particular module on `alchemy` only, thereby avoiding this problem altogether. Infura in the future will support longer look-backs on Arbitrum and therefore I don't see any particular point in fighting this restriction any further seeing pagination does not work.

packages/financial-templates-lib/package.json
```diff
@@ -4,8 +4,8 @@
   "description": "Arbitrage automation and libraries for UMA financial templates",
   "dependencies": {
     "@ethersproject/bignumber": "^5.4.2",
-    "@google-cloud/logging-winston": "^4.0.4",
```
chrismaree, line 7: this is unrelated to the main change of this PR. I realized this fixes the issue we've been having for a while in production where this log is generated every time the hub or spoke runs:

image

packages/insured-bridge-relayer/src/canonical-bridge-adapters/BridgeAdapterInterface.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export default interface ExchangeAdapterInterface {
+  initialize(): void;
+  constructCrossDomainFinalizationTransaction(
+    transactionHash: string
+  ): Promise<{ l2TransactionHash: string; finalizationTransaction: TransactionType | null }>;
```
mrice32, line 7: why `| null`?

mrice32 (approved):
LGTM! This is really great work!

packages/insured-bridge-relayer/src/canonical-bridge-adapters/BridgeAdapterInterface.ts
chrismaree: if the relay cant be sent (it's not confirmed yet) it is null! this is the key mechanism on how filtering works :)

PR #3664, opened 2021-12-06
improve(across-relayer): Add log to block search in unhappy dispute case

nicholaspai:
**Motivation**

We should add log to this loop in case bot gets stuck here. This loop is highly likely to break in the future if the block history is large and RPC API endpoints are returning slowly (or we get rate-limited).

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -875,11 +875,10 @@ export class Relayer {
       while (deposit === undefined) {
         this.logger.debug({
           at: "AcrossRelayer#Disputer",
-          message: "Searching through all L2 block history for event",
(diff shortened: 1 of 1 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -891,14 +899,14 @@ export class Relayer {
           };
           _deposit.depositHash = this.l2Client.generateDepositHash(_deposit);
           if (_deposit.depositHash === relay.depositHash) {
+            deposit = _deposit;
```
nicholaspai, line 902: minor bug fix I found: `deposit` will always be `undefined` if we don't place this log after resetting `deposit = _deposit`

mrice32 (approved):
LGTM!

PR #3665, opened 2021-12-06
feat(sdk): add timestamp search for lp fee pct calculation

daywiss:
**Motivation**
https://app.shortcut.com/uma-project/story/3307/add-support-for-l2s-that-are-more-than-10-mintues-behind-in-sdk

**Summary**

Wraps the lpfee calculation to add an optional timestamp with the query. The timestamp does a block lookup using BlockFinder code in financial-templates-lib.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.shortcut.com/uma-project/story/3307/add-support-for-l2s-that-are-more-than-10-mintues-behind-in-sdk

### DIFF
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -39,7 +39,6 @@
     "@types/lodash-es": "^4.17.5",
     "@uma/contracts-frontend": "^0.1.13",
     "@uma/contracts-node": "^0.1.13",
-    "@uma/common": "^2.14.0",
     "axios": "^0.21.4",
     "bn.js": "^4.11.9",
     "decimal.js": "^10.3.1",
@@ -48,6 +47,8 @@
     "web3": "^1.6.0"
   },
   "peerDependencies": {
+    "@eth-optimism/contracts": "^0.5.5",
+    "@eth-optimism/core-utils": "^0.7.3",
     "ethers": "^5.4.2"
   },
   "devDependencies": {
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -39,7 +39,7 @@
     "@types/lodash-es": "^4.17.5",
     "@uma/contracts-frontend": "^0.1.13",
     "@uma/contracts-node": "^0.1.13",
-    "@uma/financial-templates-lib": "^2.19.0",
+    "@uma/common": "^2.14.0",
     "axios": "^0.21.4",
     "bn.js": "^4.11.9",
     "decimal.js": "^10.3.1",
--- a/packages/sdk/package.json
+++ b/packages/sdk/package.json
@@ -39,6 +39,7 @@
     "@types/lodash-es": "^4.17.5",
     "@uma/contracts-frontend": "^0.1.13",
     "@uma/contracts-node": "^0.1.13",
+    "@uma/financial-templates-lib": "^2.19.0",
     "axios": "^0.21.4",
     "bn.js": "^4.11.9",
     "decimal.js": "^10.3.1",
--- a/packages/sdk/src/across/README.md
+++ b/packages/sdk/src/across/README.md
@@ -142,7 +142,7 @@ const tokenAddress = // token address on L1 to transfer from l2 to l1
 const bridgePoolAddress = // bridge pool address on L1 with the liquidity pool
 const amount = // amount in wei for user to send across
 const timestamp = // timestamp in seconds of latest block on L2 chain
-const const percent = await calculator.getLpFeePct(tokenAddress, bridgePoolAddress, amount, timestamp)
+const percent = await calculator.getLpFeePct(tokenAddress, bridgePoolAddress, amount, timestamp)
 
 
 ```
--- a/packages/sdk/src/across/README.md
+++ b/packages/sdk/src/across/README.md
@@ -126,6 +126,27 @@ const result = utils.calculateGasFees(gas, gasPrice)
 const userDisplay = fromWei(result)
 ```
 
+### LP Fee Calculator
+
+Get lp fee calculations by timestamp.
+
+```ts
+import * as uma from "@uma/sdk"
+
+const LpFeeCalculator = uma.across.LpFeeCalculator
+
+// pass in L1 read only provider. You should only have a single instance of the calculator.
+const calculator = new LpFeeCalculator(provider)
+
+const tokenAddress = // token address on L1 to transfer from l2 to l1
+const bridgePoolAddress = // bridge pool address on L1 with the liquidity pool
+const amount = // amount in wei for user to send across
+const timestamp = // timestamp in seconds of latest block on L2 chain
+const const percent = await calculator.getLpFeePct(tokenAddress, bridgePoolAddress, amount, timestamp)
+
+
+```
+
 ### Contract Clients
 
 - [BridgePool Client](./clients/README.md)
--- a/packages/sdk/src/across/lpFeeCalculator.ts
+++ b/packages/sdk/src/across/lpFeeCalculator.ts
@@ -5,7 +5,8 @@ import { bridgePool } from "../clients";
 import { RATE_MODELS } from "./constants";
 import { BigNumberish } from "./utils";
 import { calculateRealizedLpFeePct } from "./feeCalculator";
-import { exists, BlockFinder } from "../utils";
+import { exists } from "../utils";
+import BlockFinder from "../blockFinder";
 
 export default class LpFeeCalculator {
   private blockFinder: BlockFinder<Block>;
--- a/packages/sdk/src/across/lpFeeCalculator.ts
+++ b/packages/sdk/src/across/lpFeeCalculator.ts
@@ -1,0 +1,38 @@
+import assert from "assert";
+import { Provider, Block } from "@ethersproject/providers";
+import { ethers, BigNumber } from "ethers";
+import { bridgePool } from "../clients";
+import { RATE_MODELS } from "./constants";
+import { BigNumberish } from "./utils";
+import { calculateRealizedLpFeePct } from "./feeCalculator";
+import { exists, BlockFinder } from "../utils";
+
+export default class LpFeeCalculator {
+  private blockFinder: BlockFinder<Block>;
+  constructor(private provider: Provider) {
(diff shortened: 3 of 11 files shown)


packages/sdk/src/utils.ts
```diff
@@ -4,6 +4,9 @@ import type Multicall2 from "./multicall2";
 import type { Contract } from "ethers";
 import zip from "lodash/zip";
 
+// re-use this class, its compatible with FE and BE
+export { BlockFinder } from "@uma/financial-templates-lib/dist/price-feed/utils";
```
mrice32, line 8: How does this work? Will the frontend be able to access this package if it isn't a runtime dependency of the sdk?

Also, accessing files within packages like this isn't portable (I don't think). I've seen problems when trying to do this in a frontend because the filesystem doesn't exist like it does in node.

packages/sdk/src/across/README.md
```diff
@@ -126,6 +126,27 @@ const result = utils.calculateGasFees(gas, gasPrice)
...
+const LpFeeCalculator = uma.across.LpFeeCalculator
+
+// pass in L1 read only provider. You should only have a single instance of the calculator.
+const calculator = new LpFeeCalculator(provider)
+
+const tokenAddress = // token address on L1 to transfer from l2 to l1
+const bridgePoolAddress = // bridge pool address on L1 with the liquidity pool
+const amount = // amount in wei for user to send across
+const timestamp = // timestamp in seconds of latest block on L2 chain
+const const percent = await calculator.getLpFeePct(tokenAddress, bridgePoolAddress, amount, timestamp)
```
nicholaspai, line 145:
```suggestion
const percent = await calculator.getLpFeePct(tokenAddress, bridgePoolAddress, amount, timestamp)
```

packages/sdk/src/across/lpFeeCalculator.ts
```diff
@@ -0,0 +1,38 @@
...
+import { calculateRealizedLpFeePct } from "./feeCalculator";
+import { exists, BlockFinder } from "../utils";
+
+export default class LpFeeCalculator {
+  private blockFinder: BlockFinder<Block>;
+  constructor(private provider: Provider) {
+    this.blockFinder = new BlockFinder<Block>(provider.getBlock.bind(provider));
+  }
+  async getLpFeePct(tokenAddress: string, bridgePoolAddress: string, amount: BigNumberish, timestamp?: number) {
+    const rateModel = RATE_MODELS[ethers.utils.getAddress(tokenAddress)];
```
nicholaspai, line 16: FYI we'll soon want to read rate models from an async method `getRateModel(string tokenAddress)` which will query a `RateModelStore` contract for a specific token address

packages/sdk/src/across/lpFeeCalculator.ts
```diff
@@ -0,0 +1,38 @@
...
+    assert(rateModel, "No rate model for token: " + tokenAddress);
+
+    amount = BigNumber.from(amount);
+    assert(amount.gt(0), "Amount must be greater than 0");
+
+    const { blockFinder, provider } = this;
+
+    const bridgePoolInstance = bridgePool.connect(bridgePoolAddress, provider);
+
+    const targetBlock = exists(timestamp)
```
nicholaspai, line 26: random aside: I really like that you have an `exists` exported method in utils, we should use this everywhere including in the bots!

nicholaspai: Actually, I notice that the return type of `exists` is `value`, shouldn't it be `boolean`?

nicholaspai (commented):
Left some nits, really easy code to read

packages/sdk/src/utils.ts
daywiss: im going to try it now before i merge this, will let you know. I think it might work though because the sdk gets built indepdendently before importing into FE.

packages/sdk/src/across/lpFeeCalculator.ts
daywiss: This was actually copied from some of matts code, which he has a similar function defined somewhere. If you are referring to this part ` value is T`, this is meant to let typescript know that the value is of type T, if the function returns true ( or something like that).  It does actually return a boolean, which looks really weird now thaat you bring it up

packages/sdk/src/utils.ts
daywiss: yea it doesnt work :( i will have to copy it into the sdk i think

packages/sdk/src/across/lpFeeCalculator.ts
mrice32: @nicholaspai `exists` is a generic typescript method, so the `value is T` is a typescripty way of telling the compiler that the input (value) is of type T (which is the input type without the possibility of null or undefined).

It can be used as a javascript shorthand for "is the thing I passed in not undefined/null", but the reason we wrote it this way was to help typescript infer types. Here's a dumb example that illustrates what this allows us to do:
```ts
const a: number | undefined | null = someFun();
if (exists(a)) {
  // This compiles, because typescript knows that a cannot be undefined or null in this branch.
  const b: number = a;
} else {
  // This compiles because we know that a _must be_ undefined or null in this branch.
  let c: undefined | null = a;
}
```

It feels a little stupid/obvious that `a !== null && a !== undefined` means that a is a number, but typescript often fails to infer this unless you define these special functions to help it differentiate the cases.

packages/sdk/package.json
```diff
@@ -39,6 +39,7 @@
     "@types/lodash-es": "^4.17.5",
     "@uma/contracts-frontend": "^0.1.13",
     "@uma/contracts-node": "^0.1.13",
+    "@uma/financial-templates-lib": "^2.19.0",
```
mrice32, line 42: is this still needed?

packages/sdk/src/blockFinder.ts
```diff
@@ -0,0 +1,103 @@
+import assert from "assert";
```
mrice32, line 1: If you're going to copy this, why not have financial-templates-lib pull it from here? Or have the users of that method just pull it from here? Would prefer not to have multiple copies of the same code if possible.

packages/sdk/src/utils.ts
```diff
@@ -93,3 +95,33 @@ export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contra
     })
   );
 };
+
+/**
+ * @notice Return average block-time for a period.
+ */
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export async function averageBlockTimeSeconds(lookbackSeconds?: number, networkId?: number): Promise<number> {
```
mrice32, line 103: is this copied from common? Can we just import it? Isn't common FE compatible and already used in the SDK?

packages/sdk/src/blockFinder.ts
daywiss: ok so youa re saying update financial templates to pull from sdk?  i can do that

packages/sdk/src/utils.ts
daywiss: its copied from common. i can try, i dont think its used in sdk currently but i could be wrong

packages/sdk/package.json
daywiss: nope

packages/sdk/src/utils.ts
daywiss: i updated this to import the function from common, seems to work

packages/sdk/src/blockFinder.ts
daywiss: i updated ftl to import blockfinder from sdk

packages/sdk/src/utils.ts
daywiss: i had to copy and paste the functions again due to build failing in across frontend when common was a dep

yarn.lock
```diff
@@ -3962,9 +3962,9 @@
   integrity sha512-eMk0b9ReBbV23xXU693TAIrLyeO5iTgBZGSJfpqriG8UkYvr/hC9u9pyMlAakDNHWmbhMZCDs6KQO0jzKD8OTw==
 
 "@napi-rs/triples@^1.0.3":
-  version "1.0.3"
-  resolved "https://registry.yarnpkg.com/@napi-rs/triples/-/triples-1.0.3.tgz#76d6d0c3f4d16013c61e45dfca5ff1e6c31ae53c"
-  integrity sha512-jDJTpta+P4p1NZTFVLHJ/TLFVYVcOqv6l8xwOeBKNPMgY/zDYH/YH7SJbvrr/h1RcS9GzbPcLKGzpuK9cV56UA==
+  version "1.1.0"
```
daywiss, line 3965: idk what happened here

mrice32 (approved):
LGTM!

packages/sdk/src/index.ts
```diff
@@ -6,15 +6,23 @@ export * as across from "./across";
...
 export { default as Multicall } from "./multicall";
 export { default as Multicall2 } from "./multicall2";
+export { default as BlockFinder } from "./blockFinder";
 
 // types
 import type { TypedEventFilterEthers as TypedEventFilter, TypedEventEthers as TypedEvent } from "@uma/contracts-node";
-import { Contract, ethers, Signer, providers, Event } from "ethers";
+import { Contract, ethers, providers, Event } from "ethers";
 import { Provider } from "@ethersproject/providers";
+import { Signer } from "@ethersproject/abstract-signer";
```
amateima, line 15: did this helped to get rid of the build error? I had to do the same thing with the `Provider` type. `providers.Provider` conflicted with the Provider used by the contracts & factories generated by the typechain

amateima (approved):
👍

packages/sdk/src/index.ts
daywiss: yea... i think the issue is that `@ethersproject` packages seem to be different versions, even if u have the same `ethers` defined at times.  I think the real fix was this https://github.com/across-protocol/across-frontend/pull/74. But this is a lot of guesswork from me, and  i literally kept trying tihngs till it worked. we will see what CI says

PR #3668, opened 2021-12-07
improve(cross-chain-oracle): Stamp ancillary data in hasPrice and getPrice

nicholaspai:
**Motivation**

I believe we should stamp ancillary data in these public methods. I originally removed this stamping logic in [this PR](https://github.com/UMAprotocol/protocol/pull/3627) because the previous `stampAncillaryData` implementation stamped the `msg.sender` address, which means that only the original requester could call `has/getPrice`, but now that we removed it we can add stamping for the caller’s convenience. 

This improves the caller's experience as they only need to pass the original ancillary data that was passed into `requestPrice`.

This PR is possible because `stampAncillaryData` now only stamps the `block.chainId` which is not affected by who the `msg.sender` is.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
+++ b/packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
@@ -104,7 +104,7 @@ contract OracleSpoke is
         uint256 time,
         bytes memory ancillaryData
     ) public view override nonReentrantView() onlyRegisteredContract() returns (bool) {
-        bytes32 priceRequestId = _encodePriceRequest(identifier, time, ancillaryData);
+        bytes32 priceRequestId = _encodePriceRequest(identifier, time, _stampAncillaryData(ancillaryData));
         return prices[priceRequestId].state == RequestState.Resolved;
     }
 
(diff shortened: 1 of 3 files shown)


chrismaree (approved):
This is interesting. it definitely makes interacting with the spoke directly easier. importantly, the stamping only includes the `chainId` so this is forward compatible with upgrades to the spoke, keeping previous requests still functional.

also, this change now makes the has/get price methods consistent with the inputs you would put into `requestPrice`. previously you would put in one ancillary data, which would then be stamped and would need separate ancillary data when fetching data from the contracts. this would have made implementation integrating with this contract more difficult as they would need to pre-stamp their ancillary data when calling  `hasPrice` and `getPrice`

nicholaspai:
> This is interesting. it definitely makes interacting with the spoke directly easier. importantly, the stamping only includes the `chainId` so this is forward compatible with upgrades to the spoke, keeping previous requests still functional.
> 
> also, this change now makes the has/get price methods consistent with the inputs you would put into `requestPrice`. previously you would put in one ancillary data, which would then be stamped and would need separate ancillary data when fetching data from the contracts. this would have made implementation integrating with this contract more difficult as they would need to pre-stamp their ancillary data when calling `hasPrice` and `getPrice`

Totally agreed. We originally removed the stamping in the previous PR because the stamping logic added `msg.sender`, but then we also removed the `msg.sender` from the stamping logic. I rate it an oversight that we forgot to re-add stamping in `has/getPrice`

PR #3673, opened 2021-12-09
feat(across-relayer): Fetch rate model from contract instead of SDK constant

nicholaspai:
**Motivation**

Bot runner no longer needs to pass in `RATE_MODELS` environment config, or have the bot by default read this constant from the `sdk` package. Instead, the bot will attempt to query the rate model dictionary from a `RateModelStore` contract.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -120,6 +128,70 @@ export class InsuredBridgeL1Client {
...
 
+  getRateModelForBlockNumber(l1Token: string, blockNumber: number | undefined = undefined): RateModel {
+    this._throwIfNotInitialized();
+    const l1TokenNormalized = toChecksumAddress(l1Token);
+
+    if (!this.rateModelsForToken[l1TokenNormalized])
+      throw new Error(`No updated rate model events for L1 token: ${l1TokenNormalized}`);
+
+    let rateModelToReturn: any;
+    // Rate model events are inserted into the array from oldest at index 0 to newest at index length-1, so we'll
```
nicholaspai, line 139: I wasn't sure whether this logic of parsing the rate model and checking that its in the expected shape should go in this function or the `update` function

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -115,30 +112,6 @@ export class RelayerConfig {
 
     this.l1DeployData = replaceAddressCase(L1_DEPLOY_DATA ? JSON.parse(L1_DEPLOY_DATA) : bridgePoolDeployData);
     this.l2DeployData = L2_DEPLOY_DATA ? JSON.parse(L2_DEPLOY_DATA) : bridgeDepositBoxDeployData;
-    const processingRateModels = RATE_MODELS ? JSON.parse(RATE_MODELS) : across.constants.RATE_MODELS;
```
nicholaspai, line 118: This parsing logic for the rate model dictionary now appears in the L1 client

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -172,12 +231,11 @@ export class InsuredBridgeL1Client {
   }
 
   async calculateRealizedLpFeePctForDeposit(deposit: Deposit): Promise<BN> {
-    if (this.rateModels === undefined || this.rateModels[deposit.l1Token] === undefined)
```
nicholaspai, line 175: An error will now be thrown in `getRateModelForBlockNumber` if a rate model isn't found for the token

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -120,6 +129,68 @@ export class InsuredBridgeL1Client {
     return this.whitelistedTokens[chainId];
   }
 
+  getRateModelForBlockNumber(l1Token: string, blockNumber: number | undefined = undefined): RateModel {
```
chrismaree, line 132: I dont think this is how we want to implement this. IMHO we should add the functionality to pulling the rate model from the smart contracts (pretty much what you've got here) _into_ the SDK then use the SDK function in the bots. this way, the front end and anything else that also uses the SDK will use the same logic for pulling the rate model. if we do what you've got here we'll end up having duplicated code. 

(note that the implementation you've got here I think will do the trick, it's just a question of _where_ it's located).

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -120,6 +129,68 @@ export class InsuredBridgeL1Client {
...
+        // this L1 token didn't exist in the RateModel at the block height and we shouldn't include it in the returned
+        // array.
+        if (
+          !blockNumber ||
+          this.updatedRateModelEventsForToken[l1TokenNormalized].find((event) => event.blockNumber <= blockNumber)
+        )
+          return toChecksumAddress(l1Token);
+        else return null;
+      })
+      .filter((x) => x !== null) as string[];
```
mrice32, line 191: nit: you should be able to use [this function](https://github.com/UMAprotocol/protocol/blob/3955d80038df1c54663a59b44d6e23bd09c7dcdc/packages/financial-templates-lib/src/types.ts#L12-L14) to force typescript to understand that the output will be a string.

Example of how this is used:
```ts
// can be imported from ../types
function isDefined<T>(val: T | undefined | null): val is T {
  return val !== undefined && val !== null;
}
....

.filter(isDefined);
```

If this all works correctly, no cast should be necessary. See https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates for more details on this concept.

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -299,6 +369,45 @@ export class InsuredBridgeL1Client {
...
+      // Before adding the new rate model to the array, first validate its format:
+      let rateModelFromEvent: any;
+      try {
+        rateModelFromEvent = JSON.parse(updatedRateModelEvent.returnValues.rateModel);
+      } catch (err) {
+        // Found an UpdatedRateModel event where the rate model string cannot be parsed into JSON, ignore it.
+        continue;
+      }
+
+      if (!expectedRateModelKeys.every((item) => Object.prototype.hasOwnProperty.call(rateModelFromEvent, item))) {
```
mrice32, line 391: A few questions:
1. Should an improperly formatted rate model throw an error? What happens if there's a rate model that matches our expectations at block 1, it's updated to one that has "incorrect" formatting at block 2, and then we're trying to find a rate model at block 3? Do we just drop the update and return the stale one? That seems undesirable.
2. nit: this code seems a bit convoluted. Could we just use `in`? ex: `item => item in rateModelFromEvent`?
3. Do we care if there are extra properties in the rateModelFromEvent? i.e. should we check the reverse, that each property in rateModelFromEvent is in the `expectedRateModelKeys` array?

mrice32 (commented):
Looks good -- a few comments!

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Will move this into the SDK after I finish up the implementation itself after you and matt's input

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: 1. Hm this is a great point, and one that can be resolved by lazily parsing stringified rate models. I've changed this to store the stringified rate model events in `update`, and then only attempt to lazily parse it in `getRateModelForBlockNumber` when it needs to be used. At that point, a failure to parse will throw an error. This way if an incorrect rate model is inserted in block 2, the rate model at block 3 will be this incorrect rate model and parsing it will throw an error.
2. Good point done.
3. I believe this check already allows for extra keys. It takes every key in `expectedRateModelKeys` and searches for it in `rateModelFromEvent`. Do you understand otherwise?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Oh very nice

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Done!

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,120 @@
...
+
+    if (!rateModel)
+      throw new Error(`No updated rate model events before block #${blockNumber} for L1 token: ${l1TokenNormalized}`);
+    return parseAndReturnRateModelFromString(rateModel?.rateModel);
+  }
+};
+
+/**
+ * @notice Return all L1 tokens that had a rate model associated with it at the block number.
+ * @param blockNumber Returns l1 tokens that were mapped to a rate model at this block height. If undefined,
```
daywiss, line 100: this is missing docs for rateModelEvents param.

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,120 @@
+import { expectedRateModelKeys, RateModel } from "./constants";
+import { exists } from "../utils";
+import type { RateModelStoreWeb3 } from "@uma/contracts-node";
+import Web3 from "web3";
+const { toChecksumAddress } = Web3.utils;
```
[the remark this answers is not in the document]
nicholaspai: Ah does this convert the address TI checksum?

packages/sdk/src/across/rateModel.ts
nicholaspai: Will add docs to all functions!

packages/sdk/src/across/rateModel.ts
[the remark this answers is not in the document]
daywiss: yep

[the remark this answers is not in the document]
nicholaspai: done

packages/sdk/src/across/lpFeeCalculator.ts
```diff
@@ -14,6 +14,7 @@ export default class LpFeeCalculator {
     this.blockFinder = new BlockFinder<Block>(provider.getBlock.bind(provider));
   }
   async getLpFeePct(tokenAddress: string, bridgePoolAddress: string, amount: BigNumberish, timestamp?: number) {
+    // TODO: Use `./rateModel` methods here instead of hardcoded constant.
```
nicholaspai, line 17: @daywiss what do you think is the best way to integrate `./rateModel.ts` into this method? One current issue right now is that `getAllRateModelEvents` expects a `Web3` contract object for `RateModelStore` to be passed in, while this method and the SDK in general expects an `ethers` contract. The `InsuredBridgeL1Client` and all code in `financial-templates-lib` expects to use `web3` providers, while the SDK uses `ethers` providers.

@mrice32 cc

nicholaspai: One really annoying implementation would be to implement two methods:
`getAllRateModelEventsWeb3` and `getAllRateModelEventsEthers`, or overload the function params and check for whether the provider/contract is web3/ethers. Thoughts?

packages/sdk/src/across/lpFeeCalculator.ts
```diff
@@ -14,6 +14,11 @@ export default class LpFeeCalculator {
     this.blockFinder = new BlockFinder<Block>(provider.getBlock.bind(provider));
   }
   async getLpFeePct(tokenAddress: string, bridgePoolAddress: string, amount: BigNumberish, timestamp?: number) {
+    // TODO: Accept another param, rateModelStoreAddress, construct a RateModelStore contract instance, query all
```
nicholaspai, line 17: see here @daywiss

packages/sdk/src/across/lpFeeCalculator.ts
nicholaspai: I ended up moving `getAllRateModelEvents` to the client, leaving the web3 event querying and reformatting up to the client before calling into the SDK to get the latest rate model for a block height.

nicholaspai: I also added an internal `createRateModelDictionary` helper method that standardizes how the rate model dictionary is constructed. This obfuscates the dictionary itself from the client of the `sdk`, leaving the only responsibility for the caller to slightly parse raw contract events into an array of `{blockNumber, rateModelString, tokenAddress}` objects.

packages/insured-bridge-relayer/src/RelayerHelpers.ts
```diff
@@ -35,17 +35,20 @@ export async function approveL1Tokens(
...
   logger: winston.Logger,
   l1Client: InsuredBridgeL1Client,
-  l2Client: InsuredBridgeL2Client,
-  whitelistedRelayL1Tokens: string[]
+  l2Client: InsuredBridgeL2Client
 ): Promise<string[]> {
   await Promise.all([l2Client.update(), l1Client.update()]);
+
+  // Fetch list of potential whitelisted L1 tokens from keys in the RateModelStore.
+  const whitelistedRelayL1Tokens = l1Client.getL1TokensFromRateModel();
```
mrice32, line 50: Do we care (or is there anything we should do) if the rate model doesn't include a token that's whitelisted by the bridgeAdmin?

Do you think this is something we should error or warn on?

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,143 @@
...
+
+export const parseAndReturnRateModelFromString = (rateModelString: string): RateModel => {
+  const rateModelFromEvent = JSON.parse(rateModelString);
+
+  // Rate model must contain all keys in `expectedRateModelKeys`, and extra keys are OK.
+  for (const key in expectedRateModelKeys) {
+    if (!(key in Object.keys(rateModelFromEvent))) {
+      throw new Error(
+        `Rate model does not contain all expected keys. Expected keys: [${expectedRateModelKeys}], actual keys: [${Object.keys(
+          rateModelFromEvent
```
mrice32, line 32: I asked about this in another comment, but I think I explained it badly, so I'll give an example. Do you think we should throw if the on-chain rate model has an extra key `R3` that the bot's rate model computation doesn't need or use?

Right now, it think we just check that it has the keys we use, but we don't care if there are extras, but wouldn't an extra key indicate that our rate model computation is out of date or incompatible?

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,143 @@
...
+ * @param rateModelString Stringified rate model to parse.
+ * @returns Rate model object. Must conform to `expectedRateModelKeys` format.
+ */
+
+export const parseAndReturnRateModelFromString = (rateModelString: string): RateModel => {
+  const rateModelFromEvent = JSON.parse(rateModelString);
+
+  // Rate model must contain all keys in `expectedRateModelKeys`, and extra keys are OK.
+  for (const key in expectedRateModelKeys) {
+    if (!(key in Object.keys(rateModelFromEvent))) {
```
mrice32, line 29: nit: I don't think this does what you might expect. `in` checks that the `key` is a _key_ for the other object. In this case that means, it's checking that `key` is 1, 2, 3, or 4. The `in` operator is a bit confusing because of how it deals with arrays.

I think to make this do what we want, we'd want to change this line to
```ts
if (!(key in rateModelFromEvent)) {
```

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,143 @@
...
+ * Return the rate model for L1 token set at the block height.
+ * @param rateModelEvents Reformatted unordered UpdatedRateModel events.
+ * @param l1Token L1 token address to get rate model for.
+ * @param blockNumber Block height to get rate model for.
+ * @returns Rate model object.
+ */
+export const getRateModelForBlockNumber = (
+  rateModelEvents: RateModelEvent[],
+  l1Token: string,
+  blockNumber: number | undefined = undefined
```
mrice32, line 84:
```suggestion
  blockNumber?: number
```

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -307,6 +325,12 @@ export class InsuredBridgeL1Client {
       this.instantRelays[l1Token] = {};
     }
 
+    // Fetch and store all rate model updated events, which will be used to fetch the rate model for a specific deposit
+    // quote timestamp.
+    this.updatedRateModelEventsForToken = this.updatedRateModelEventsForToken.concat(
+      await this._getAllRateModelEvents(blockSearchConfig)
+    );
```
mrice32, line 332: Do we need to sort these by block number, transaction index, and log index? See this code block where we do this elsewhere: https://github.com/UMAprotocol/protocol/blob/011905dcda88889884ec6f377d8b5cea8af20c7e/packages/financial-templates-lib/src/price-feed/UniswapPriceFeed.ts#L231-L241.

I have generally seen them to at least be sorted by block number, but I haven't seen any documentation _guarantee_ that behavior.

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,143 @@
...
+        `Block number #${blockNumber} is before first UpdatedRateModel event block ${firstEventBlockNumber}`
+      );
+    }
+
+    // We're looking for the latest rate model update that occurred at or before the block number.
+    // Rate model events are inserted into the array from oldest at index 0 to newest at index length-1, so we'll
+    // reverse the array so it goes from newest at index 0 to oldest at index length-1, and then find the first event
+    // who's block number is less than or equal to the target block number.
+    const rateModel = rateModelDictionary[l1TokenNormalized]
+      .slice() // reverse() modifies memory in place so create a copy first.
```
mrice32, line 108: +1

mrice32 (commented):
Looks great! Mostly minor comments on minor details and edge cases. All-in-all looks really great!

packages/sdk/src/across/lpFeeCalculator.ts
mrice32: I think this is a good solution here.

However, long-term, we probably need a conversion function that will take a web3 provider and create an ethers provider from it. And a similar function in the opposite direction. The bots will never be able to comfortably share functionality with the FE through the SDK if we can't have some way of sharing code that interacts with the blockchain.

packages/sdk/src/across/lpFeeCalculator.ts
daywiss: i dont understand why we need to query all events, wasnt there a function call to get the latest rate model?

packages/insured-bridge-relayer/src/RelayerHelpers.ts
nicholaspai: Hm that means we need to iterate across ALL whitelisted L1 tokens on the `bridgeAdmin` which is kind of hard to do unless we maintain a list of whitelisted tokens right? I suppose we do store all whitelisted L1 tokens in the `bridgePoolDeployData` struct in `RelayerConfig.ts`

packages/sdk/src/across/rateModel.ts
nicholaspai: > Do you think we should throw if the on-chain rate model has an extra key R3 that the bot's rate model computation doesn't need or use?

No i don't think an extra key matters, as long as the rate model can be computed using the keys it does have? I don't think this makes the rate model incompatible or out of date, this is not obvious to me

packages/sdk/src/across/rateModel.ts
nicholaspai: I don't think this works either assuming my tests are correct so I've just replaced this with the much easier to reason about `!Object.keys(rateModelFromEvent).includes(key)`

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: Yeah good point. I'm going to perform this sorting in `sdk/rateModel/createRateModelEventDictionary` but there isn't any need to sort them since the SDK function will sort them

packages/sdk/src/across/rateModel.ts
mrice32: Weird. I was able to verify this behavior in the console:
```
> const obj = { a: 1, b: 2 }
undefined
> "a" in obj
true
> "a" in Object.keys(obj)
false
```

mrice32: Either way, `includes` is totally fine to!

packages/sdk/src/across/rateModel.ts
mrice32: Let's say we want to add an R3 to make this piecewise rate curve have an additional changepoint (2 instead of 1).

I'm assuming that means we would need to change the names of the _existing_ keys too to make sure that the old clients error out instead of performing the wrong computation right?

Or do you think we should include a version number key?

Basically, I think we should have _some_ mechanism to ensure that updates to rate models are recognized in old clients so they can throw errors instead of doing the wrong computation. How do you think we should do that?

packages/insured-bridge-relayer/src/RelayerHelpers.ts
mrice32: Isn't there an event we can parse in the BridgeAdmin to grab all of the whitelisted tokens?

Not saying we _have_ to do this, just pointing out that we're using the rate model as the source of truth, but there's a similarly valid source of truth in the BridgeAdmin whitelist. Not sure if we should try to detect when those are different from one another. I could see it being fine for there to be a rate model for a token that isn't whitelisted, but a whitelisted token with no rate model seems odd at best and a security issue at worst.

Thoughts?

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,158 @@
...
+ * @param blockNumber Block height to get rate model for.
+ * @returns Rate model object.
+ */
+export const getRateModelForBlockNumber = (
+  rateModelEvents: RateModelEvent[],
+  l1Token: string,
+  blockNumber?: number
+): RateModel => {
+  const l1TokenNormalized = ethers.utils.getAddress(l1Token);
+  const rateModelDictionary = createRateModelEventDictionary(rateModelEvents);
```
mrice32, line 102: Theoretically, how many times could this rate model dictionary be created per run? Does it get rebuilt every time we need to grab a rate model for a relay? If there were 100 relays within the last 2 hours, would we create this 100 times?

Thoughts on making this a persistent class where the rate model events are passed on construction?

mrice32 (commented):
Looks good! Replied to a few threads and left one additional comment

packages/sdk/src/across/lpFeeCalculator.ts
nicholaspai: > wasnt there a function call to get the latest rate model?
The `getRateModelForBlockNumber` expects an array of `RateModelEvents` to be passed to it to find the correct one for the block number. The reason why events aren't queried in that function is because the FE and the bots would query events differently using `ethers` and `web3` contract instances respectively. Does that make sense?

packages/insured-bridge-relayer/src/RelayerHelpers.ts
nicholaspai: Ah you're right we totally can use `l1Client.getWhitelistedTokensForChainId` instead of calling `bridgeAdmin.methods.whitelistToken()` directly. I'll fix this

packages/sdk/src/across/rateModel.ts
nicholaspai: Ah i see your point. I'll just add the reverse check as well to guarantee all keys are exactly included and no extra ones

packages/sdk/src/across/rateModel.ts
nicholaspai: Yeah it does get rebuilt every time a new rate model is constructed, I can change to a persistant class with storage

nicholaspai: > Does it get rebuilt every time we need to grab a rate model for a relay?

Yes, good point.

I've added a persistent `RateModelDictionary` class in this PR, I think this means that 1 dictionary would be created per serverless run. [PR](https://github.com/UMAprotocol/protocol/pull/3673/commits/fa6270e377eb61e28c61407ae31ed03c6a2647b3)

packages/sdk/src/across/constants.ts
```diff
@@ -37,6 +39,7 @@ export const ADDRESSES = {
   USDC: "0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48",
 };
 
+// TODO: Deprecate this hardcoded mapping and read from RateModelStore contract via ./rateModel.ts helper methods.
```
mrice32, line 42: So this means that this PR will not change the behavior, yet, right?

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,174 @@
...
+// Events should be reformatted into this shape to be used as input into methods in this file.
+export type RateModelEvent = {
+  blockNumber: number;
+  transactionIndex: number;
+  logIndex: number;
+  rateModel: string;
+  l1Token: string;
+};
+
+export class RateModelDictionary {
```
mrice32, line 20: Should we add tests for this class?

mrice32 (approved):
This is awesome! +1

packages/sdk/src/across/constants.ts
nicholaspai: Nope

packages/sdk/src/across/rateModel.ts
nicholaspai: [Added tests](https://github.com/UMAprotocol/protocol/pull/3673/commits/bae72c18a055027d7f048c848e43d65167794df1)

packages/sdk/src/across/constants.ts
nicholaspai: Oh wait, yes, this PR will change the bot's behavior but not the front-end until `lpFeeCalculator.getLpFeePct` implementation is changed, see my updated comment

mrice32 (approved):
LGTM!

daywiss (approved):
lgtm, nice work

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -307,6 +331,15 @@ export class InsuredBridgeL1Client {
       this.instantRelays[l1Token] = {};
     }
 
+    // Fetch and store all rate model updated events, which will be used to fetch the rate model for a specific deposit
+    // quote timestamp.
+    // Note: there is no need to sort these events chronologically, as they will be sorted upon calling
+    // `rateModel.updateRateModelEventDictionary`
```
chrismaree, line 337: is this comment useful? are events not by default sorted chronologically?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -307,6 +331,15 @@ export class InsuredBridgeL1Client {
...
+    this.updatedRateModelEventsForToken = this.updatedRateModelEventsForToken.concat(
+      await this._getAllRateModelEvents(blockSearchConfig)
```
chrismaree, line 339: can this async call be bundled with some of the others? perhaps placed in the promise.all block below or joined with the call to fetch the OO liveness?

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -89,6 +85,9 @@ export class RelayerConfig {
     if (!BRIDGE_ADMIN_ADDRESS) throw new Error("BRIDGE_ADMIN_ADDRESS required");
     this.bridgeAdmin = toChecksumAddress(BRIDGE_ADMIN_ADDRESS);
 
+    if (!RATE_MODEL_ADDRESS) throw new Error("RATE_MODEL_ADDRESS required");
```
chrismaree, line 88: why? can we not get it from the 1.json file using getAddress?

I guess though that this is consistent with the `BRIDGE_ADMIN_ADDRESS` env which requires that this is specified. 

perhaps a future PR can remove this requirement and we place logic in the `index` to fetch the address from the known addresses if this env is not provided.

packages/insured-bridge-relayer/src/RelayerHelpers.ts
```diff
@@ -35,34 +35,42 @@ export async function approveL1Tokens(
...
-  const whitelistedTokenMappings = await Promise.all(
-    whitelistedRelayL1Tokens.map((tokenAddress) =>
-      l1Client.bridgeAdmin.methods.whitelistedTokens(tokenAddress, l2Client.chainId.toString()).call()
-    )
-  );
-  const prunedWhitelist = whitelistedRelayL1Tokens.filter(
-    (_tokenAddress, i) => whitelistedTokenMappings[i].l2Token !== ZERO_ADDRESS
+
+  // Fetch list of potential whitelisted L1 tokens from keys in the RateModelStore.
+  const whitelistedRelayL1TokensInRateModel = l1Client.getL1TokensFromRateModel();
```
chrismaree, line 50: so this implies that the l1client must be updated before this method is called. not really an issue, just a comment.

packages/insured-bridge-relayer/src/RelayerHelpers.ts
```diff
@@ -35,34 +35,42 @@ export async function approveL1Tokens(
...
+// to the list of all possible tokens provided in the on-chain RateModels dictionary. However, not all L1 tokens in this
 // dictionary will be whitelisted for all L2 deposit boxes.
 export async function pruneWhitelistedL1Tokens(
   logger: winston.Logger,
   l1Client: InsuredBridgeL1Client,
-  l2Client: InsuredBridgeL2Client,
-  whitelistedRelayL1Tokens: string[]
+  l2Client: InsuredBridgeL2Client
 ): Promise<string[]> {
   await Promise.all([l2Client.update(), l1Client.update()]);
```
chrismaree, line 47: unrelated to this PR but it's not super ideal that we do the first client updates here :/

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -35,7 +35,7 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
 
     // The logger is having issues with logging nested BNs. Remove this for now. Is indirectly fixed in PR https://github.com/UMAprotocol/protocol/pull/3656
     /* eslint-disable @typescript-eslint/no-unused-vars */
-    const { rateModels, ...logableConfig } = config;
+    const { ...logableConfig } = config;
```
chrismaree, line 38: seeing you've removed this you can ditch this thing in totality. you can remove the comment above, and change the `log` below to include `config`. the whole thing is now `logable`.

packages/sdk/src/across/constants.ts
```diff
@@ -16,6 +16,8 @@ export const SPEED_UP_UMA_GAS = 227341;
 // Bots incur lower than expected costs due to batching mulitple transactions, this roughly estimates the savings
 export const DEFAULT_GAS_DISCOUNT = 25;
 
+export const expectedRateModelKeys = ["UBar", "R0", "R1", "R2"];
```
chrismaree, line 19: 👍

packages/sdk/src/across/rateModel.ts
```diff
@@ -0,0 +1,174 @@
...
+        );
+      }
+
+      // We're looking for the latest rate model update that occurred at or before the block number.
+      // Rate model events are inserted into the array from oldest at index 0 to newest at index length-1, so we'll
+      // reverse the array so it goes from newest at index 0 to oldest at index length-1, and then find the first event
+      // who's block number is less than or equal to the target block number.
+      const rateModel = this.rateModelDictionary[l1TokenNormalized]
+        .slice() // reverse() modifies memory in place so create a copy first.
+        .reverse()
```
chrismaree, line 63:
```suggestion
        .slice()
        .reverse() // reverse() modifies memory in place so create a copy first.
```

chrismaree (approved):
this is excellent! 

one question I have is this the integration steps for the front end. this, in its current form, wont work directly with the front end I think. shall we wait until this PR is in until we update the front end?

packages/insured-bridge-relayer/src/RelayerHelpers.ts
nicholaspai: should we update the clients before calling `pruneWhitelistedL1Tokens` and removing it from this method?

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: done

packages/insured-bridge-relayer/src/RelayerHelpers.ts
nicholaspai: Thoughts? https://github.com/UMAprotocol/protocol/pull/3673/commits/b2225df1436ad895ddaf7525bb0067f5b8461141

chrismaree: ye that makes sence! thanks for making the change.

PR #3678, opened 2021-12-10
fix(cross-chain-oracle): [L03] fixed wrong commenting

chrismaree:
OZ identified the following issues:
_Here are some misleading comments we identified during our review:
ChildMessengerConsumerInterface.sol:
Line 5 says "parent messenger" instead of "child messenger"
GovernorSpoke.sol:
Lines 49-51 links to a Gnosis file even though the comment says the snippet was copied
from Governor.sol. Additionally, the snippet is not identical to the one in Governor.sol_

This PR addresses these misleading comments.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
+++ b/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
@@ -45,9 +45,7 @@ contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
 
     // Note: this snippet of code is copied from Governor.sol.
     function _executeCall(address to, bytes memory data) private returns (bool) {
-        // Note: this snippet of code is copied from Governor.sol.
-        // solhint-disable-next-line max-line-length
-        // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
+        // Note: this snippet of code is copied from Governor.sol and modified to not include any "value" field.
         // solhint-disable-next-line no-inline-assembly
 
         bool success;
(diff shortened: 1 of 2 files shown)


packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -45,9 +45,7 @@ contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
 
     // Note: this snippet of code is copied from Governor.sol.
     function _executeCall(address to, bytes memory data) private returns (bool) {
-        // Note: this snippet of code is copied from Governor.sol.
-        // solhint-disable-next-line max-line-length
-        // https://github.com/gnosis/safe-contracts/blob/59cfdaebcd8b87a0a32f87b50fead092c10d3a05/contracts/base/Executor.sol#L23-L31
+        // Note: this snippet of code is copied from Governor.sol and modified to not include any "value" field.
```
chrismaree, line 48: note @nicholaspai & @mrice32 that the assembly we have in this file does not have a `value` field, like it has in the `Governor.sol`. this means that this `_executeCall` method cant include any eth transfers on the L2. Should we consider changing this?

nicholaspai: How would we ensure that `value` could be used properly? Would we have to seed the `GovernorSpoke` with ETH?

nicholaspai: This is how the Polygon governor tunnel currently works BTW

chrismaree: indeed. we'd need to send it the ETH to use it. it basically means that having ETH in the contract is possible. there is not really a reason to have this in our current implementation but equally no reason really not to support it.

mrice32: In the interest of keeping things consistent and simple, I'd argue that we should leave this out and not worry about it until we need a feature like this.

mrice32 (approved):

PR #3680, opened 2021-12-10
fix(cross-chain-oracle): [N03] address incorrect interface

chrismaree:
**Motivation**

OZ identified the following issue: _The GovernorSpoke contract incorrectly uses the ChildMessengerConsumerInterface type to describe its messenger variable. Consider using the ChildMessengerInterface instead._

This PR fixes this by using the correct interface.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
+++ b/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
@@ -2,6 +2,7 @@
 pragma solidity ^0.8.0;
 
 import "./interfaces/ChildMessengerConsumerInterface.sol";
+import "./interfaces/ChildMessengerInterface.sol";
 import "../common/implementation/Lockable.sol";
 
 /**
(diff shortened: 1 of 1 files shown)


packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -10,12 +11,12 @@ import "../common/implementation/Lockable.sol";
...
 contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
     // Messenger contract that receives messages from root chain.
-    ChildMessengerConsumerInterface public messenger;
+    ChildMessengerInterface public messenger;
 
     event ExecutedGovernanceTransaction(address indexed to, bytes data);
     event SetChildMessenger(address indexed childMessenger);
 
-    constructor(ChildMessengerConsumerInterface _messengerAddress) {
+    constructor(ChildMessengerInterface _messengerAddress) {
```
nicholaspai, line 19: Nit: We don't even call any functions on this `messenger`, should we just store it as an `address`?

nicholaspai (approved):

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
chrismaree: i personally like storing things as their interface as it makes it more explicit.

PR #3685, opened 2021-12-10
fix(slack-logger/relay-logs): small logger fixes

chrismaree:
**Motivation**

This PR makes two logs changes:
1) slack logs sent by the relayer should not wrap to a extra line. to solve this I removed the `pct` from some logs, which was redundant, the make the logs take up a little bit less space
2) there was a subtle bug in the slack markdown creator that removed the last char from the markdown text. see example below:

this slack message:
image
shows the address as 0xb1A...A68aD but the address ends in a 4, having lost the last param. hovering shows this loss:
image

this was also fixed.


nicholaspai (approved):
LGTM

PR #3686, opened 2021-12-11
fix: harden logger waiting in serverless hub

mrice32:
**Motivation**

Hub errors have been consistently failing to bubble up to the winston transports.

**Summary**

This modifies how the logger is created in the hub (one per request), so that we can use the `waitForLogger` function.

It also makes the waiting period customizable via an environment variable `WAIT_FOR_LOGGER_DELAY`, which will allow users to increase the waiting period on the fly without rebuilding.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
This is a great idea! what you've effectively done is make each invocation of the hub have it's own logger. this should help with a few issues of the silent logger failing, in particular when a hub is re-used between runs.

PR #3688, opened 2021-12-11
feat(cross-chain-oracle): Enable GovernorHub to change child messenger on child network

nicholaspai:
**Motivation**

This PR enables the owner of `GovernorHub` to call `relayGovernance` with a new child messenger address, which will change the `messenger` address stored in the `GovernorSpoke` and `OracleSpoke` on the child chain. This allows the ChildMessenger to be upgraded without having to update the spoke contracts.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -39,8 +45,19 @@ contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
         // TODO: Consider calling this via <address>.call(): https://docs.soliditylang.org/en/v0.8.10/units-and-global-variables.html?highlight=low%20level%20call#members-of-address-types
         // to avoid inline assembly.
 
-        require(_executeCall(to, inputData), "execute call failed");
-        emit ExecutedGovernanceTransaction(to, inputData);
+        // There is a special case if `to` is this contract. If this contract is the target, then we assume that the
+        // cross-chain caller is attempting to change the child messenger.
+        if (to == address(this)) {
```
chrismaree, line 50: hhmm. ideally, we'd like a separate method that does this messenger upgrade. problem is though that our inbound calls from L2->L1 always go via the `processMessagerFromParent` call. 

I don't think we want to overload this function like this. in my mind this is a _bit_ of a code smell. it might be better to add a function on this contract called `setChildMessenger`, like you did on the `OracleSpoke`. Then, set the access control on that function to be called only by `this` contract. We'd then craft a normal `Governance` tx, sent over the bridge like any other one and it calls _back_ into the `GovernorSpoke` by setting the `to` to be the address of this contract. Setting the `messenger` would be done by calling the `OracleSpoke`'s `SetMessengerSpoke` via a normal governance call (using the function you added already).

what this means is we don't overload the function like this, the `GovernorSpoke` does not need to have an extra event, extra branching logic or need to store the `Finder`. We _do_ need to send two governance transactions at once (one for each contract's spoke setting) but I think that's fine as it makes this part of the implementation was easier. 

wdyt?

chrismaree: I've updated this function to use this implementation.

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -37,15 +48,27 @@ contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
      */
     function processMessageFromParent(bytes memory data) public override nonReentrant() onlyMessenger() {
         (address to, bytes memory inputData) = abi.decode(data, (address, bytes));
-        // TODO: Consider calling this via <address>.call(): https://docs.soliditylang.org/en/v0.8.10/units-and-global-variables.html?highlight=low%20level%20call#members-of-address-types
-        // to avoid inline assembly.
 
         require(_executeCall(to, inputData), "execute call failed");
-        emit ExecutedGovernanceTransaction(to, inputData);
+        emit ExecutedGovernanceTransaction(to, data);
```
nicholaspai, line 53: Since "to" is a part of "data" I think you should either emit "to" and "inputData" (the data you call "to" with) or just data. I prefer leaving this the way it is

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -37,15 +48,27 @@ contract GovernorSpoke is Lockable, ChildMessengerConsumerInterface {
...
+    }
+
+    /**
+     * @notice Changes the address of the child messenger contract used by both this contract and the OracleSpoke.
+     * @dev Can only be called by this contract. i.e Executed by the GovernorHub sending a message, via the messenger
+     * to the processMessageFromParent function which intern calls back into this contract.
+     * @param newChildMessenger Address of the child messenger contract to set for this contract and OracleSpoke.
+     */
+    function setChildMessenger(address newChildMessenger) public onlyThisContract {
```
nicholaspai, line 62: Nit: should this be reentrant guarded?

nicholaspai (commented):
Looks great left some nits

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
chrismaree: agreed! reverted

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
chrismaree: the method reverts if this has a reentrant guard on it :)

packages/core/test/cross-chain-oracle/GovernorSpoke.js
```diff
@@ -1,52 +1,81 @@
...
 const AddressWhitelist = getContract("AddressWhitelist");
+const Finder = getContract("Finder");
+const OracleSpoke = getContract("OracleSpoke");
 
 describe("GovernorSpoke.js", async () => {
   let accounts;
   let owner;
   let messenger;
 
-  let governor;
```
chrismaree, line 14: small nit to make tests easier to read.

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
```diff
@@ -22,29 +23,25 @@ import "./interfaces/ChildMessengerConsumerInterface.sol";
...
         _;
     }
-
-    modifier onlyMessenger() {
-        require(msg.sender == address(messenger), "Caller must be messenger");
+    modifier onlyGovernorSpoke() {
+        require(
+            msg.sender == finder.getImplementationAddress(OracleInterfaces.GovernorSpoke),
+            "Caller must be governor spoke"
+        );
```
mrice32, line 44: I may be totally missing something, but where is this used?

mrice32 (approved):
LGTM!

packages/core/contracts/cross-chain-oracle/OracleSpoke.sol
chrismaree: it's not! good catch. removed.

PR #3690, opened 2021-12-13
fix: [C01] fixes issue where disputes are impossible in the OptimisticRewarder

mrice32:
**Motivation**

```
When disputing a reward request, the OptimisticRewardBase contract first triggers a proposal on the
SkinnyOptimisticOracle and then disputes that proposal. However, the proposal sets the expiration
time as an offset from the current (dispute) time, while the dispute specifies the expiration as an offset
from the time of the original reward request. In most cases, this discrepancy will prevent the oracle
from identifying the proposal to be disputed, which means valid disputes will not be processed and
invalid reward requests will be accepted.

Consider updating the dispute invocation to correctly specify the proposal to be disputed.
```

**Summary**

This changes the timestamp passed into the Request object provided to the dispute function to match what the SkinnyOptimisticOracle sets.

**Testing**

- [ ]  New unit tests created
- [x]  All existing tests pass (updated existing tests to cover this case by modifying the dispute timestamps)
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
this catch is excellent. indeed a critical severity issue! good fix

PR #3693, opened 2021-12-13
fix(sdk): fix pool withdraw validation

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/3476/fix-bug-with-sdk-validatewithdraw-in-across-client

Someone who tried to withdraw a large lp position ran into an error and could not withdraw.  see [slack](https://risklabs.slack.com/archives/C02D94M08F7/p1638392549018600)

**Summary**

Removes the line that checks against `pool.liquidReserves`

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**
https://app.shortcut.com/uma-project/story/3476/fix-bug-with-sdk-validatewithdraw-in-across-client

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -297,10 +297,6 @@ export class ReadPoolClient {
 export function validateWithdraw(pool: Pool, user: User, lpTokenAmount: BigNumberish) {
   const l1TokensToReturn = BigNumber.from(lpTokenAmount).mul(pool.exchangeRateCurrent).div(fixedPointAdjustment);
   assert(BigNumber.from(l1TokensToReturn).gt("0"), "Must withdraw amount greater than 0");
-  assert(
-    BigNumber.from(pool.liquidReserves).gte(l1TokensToReturn.add(pool.pendingReserves)),
-    "Utilization too high to remove that amount, try lowering withdraw amount"
-  );
   assert(BigNumber.from(lpTokenAmount).lte(user.lpTokens), "You cannot withdraw more than you have");
   return { lpTokenAmount, l1TokensToReturn: l1TokensToReturn.toString() };
 }
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -85,22 +85,22 @@ class PoolState {
   public async read(latestBlock: number, previousBlock?: number) {
     if (this.l1Token === undefined) this.l1Token = await this.contract.l1Token();
     // typechain does not have complete types for call options, so we have to cast blockTag to any
-    const [exchangeRatePrevious, liquidityUtilizationCurrent] = await Promise.all([
-      this.contract.callStatic.exchangeRateCurrent({
-        blockTag: previousBlock || latestBlock - 1,
-      } as any),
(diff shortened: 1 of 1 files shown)


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -297,10 +297,6 @@ export class ReadPoolClient {
 export function validateWithdraw(pool: Pool, user: User, lpTokenAmount: BigNumberish) {
   const l1TokensToReturn = BigNumber.from(lpTokenAmount).mul(pool.exchangeRateCurrent).div(fixedPointAdjustment);
   assert(BigNumber.from(l1TokensToReturn).gt("0"), "Must withdraw amount greater than 0");
-  assert(
```
chrismaree, line 300: so there is a situation where this does not error and a user cant withdraw, no? if you want to use the contracts to validate this could you not try `.call`ing `removeLiquidity` with the `from` set to the users wallet and see if it reverts?

chrismaree (commented):
This will work fine but there is a situation where this will create transactions that revert I think (if the utilisation is legitimately too high).

packages/sdk/src/across/clients/bridgePool.ts
daywiss: it would be great if i could do that check without calling the contract, is there a way given the known vars?

mrice32: _Technically_ you could probably do this. But operationally, this would be very difficult. The contract does some math based on its balance and the current liquid reserves, bond, and utilized reserves to figure out what the new utilized reserves will be.

Generally, the easiest, most accurate, and most future-proof way to do this is to run a .call() (I know this is inconvenient for the FE since it's async and depends on what the user types in).

mrice32 (approved):

packages/sdk/src/across/clients/bridgePool.ts
daywiss: ok interesting. so i could change this validation to just do all the validation in contract in that case, since im sure the checks in the code here will be redundant

daywiss: I tested out using the contract to validate, but what i dont like is that i dont get back good error messages. I can try to clean up the error when it throws, but i would basically get this `ERC20: burn amount exceeds balance` when withdrawing more eth than deposited. Im not sure how to test for withdrawing more liquidity than what is available. For now i will add improving this function to the backlog and keep this client side implementation for now.

daywiss: https://app.shortcut.com/uma-project/story/3637/improve-pool-withdraw-validation

mrice32: SGTM!

mrice32 (approved):

daywiss:
@mrice32 pointed out that if we call `sync` on the contract before getting pool properties (in a multicall) then we will have accurate values to do client side validation.  The update here was to move `exchangeRateCurrent` to the first value to be queried in the multicall since it calls `sync` under the hood, which theoretically should give the best values for teh rest of the properties we validate pool lp withdraws with.

mrice32 (approved):
LGTM!

PR #3694, opened 2021-12-13
fix: [M01] fixes incorrect event parameters

mrice32:
**Motivation**

```
The OptimisticRewarderBase contract defines a Requested event which is emitted from the
requestRedemption function when a redemption is requested. This event is defined to emit the expiry
time of the redemption as its last parameter. However, when the event is emitted, its last parameter is
incorrectly set to the current time.
Similarly the Redeemed event reads the expiry time after the record is deleted, so it will be incorrectly
set to zero.

Given that this event can be used to trigger off-chain computations, consider updating the emitted
value appropriately.
```

**Summary**

Corrected as suggested in the Motivation.

Added tests for these events.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
🥇

PR #3699, opened 2021-12-14
fix(serverless): improve how hub and spoke logs interact

chrismaree:
**Motivation**

There are a number of cases where logs are not displayed correctly between the hub and spoke. This PR makes a few small changes on the hub/spoke implementation to try and make logs more predictable and consistent.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -50,7 +50,12 @@ const defaultHubConfig = {
   rejectSpokeDelay: 120, // 2 min.
 };
 
+const waitForLoggerDelay = process.env.WAIT_FOR_LOGGER_DELAY || 5;
+
 hub.post("/", async (req, res) => {
+  // Use a custom logger if provided. Otherwise, initialize a local logger.
+  // Note: no reason to put this into the try-catch since a logger is required to throw the error.
+  const logger = customLogger || createNewLogger();
```
chrismaree, line 58: this is taken from https://github.com/UMAprotocol/protocol/pull/3686. will rebase once that PR is merged in

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -235,24 +241,30 @@ hub.post("/", async (req, res) => {
           try {
             return {
               spokeName: spokeName,
-              errorReported: errorOutput.errorOutputs[spokeName].execResponse
-                ? errorOutput.errorOutputs[spokeName].execResponse.stderr
-                : errorOutput.errorOutputs[spokeName],
+              errorReported:
```
chrismaree, line 244: this is the one main change made here to try and more consistently produce error logs. in different testing situations the previous block would return blank.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -373,7 +385,7 @@ async function _getLastQueriedBlockNumber(configIdentifier, chainId) {
         const [dataField] = await datastore.get(key);
         return dataField[chainId];
       } else if (hubConfig.saveQueriedBlock == "localStorage") {
-        return process.env[`lastQueriedBlockNumber-${chainId}-${configIdentifier}`];
+        return process.env[`lastQueriedBlockNumber-${chainId}-${configIdentifier}`] | 0;
```
chrismaree, line 385: when running locally can have an issue when there is no config stored. should return 0 in this case.

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -446,7 +458,7 @@ function _processSpokeResponse(botKey, spokeResponse, validOutputs, errorOutputs
     errorOutputs[botKey] = {
       status: spokeResponse.status,
       execResponse: spokeResponse.value && spokeResponse.value.execResponse,
-      reason: "empty stdout",
+      message: "empty stdout",
```
chrismaree, line 458: these errors were lost as they should be described as a `message`, not a `reason`.

note that the previous implementation would have worked fine as well due to this: https://github.com/UMAprotocol/protocol/pull/3699/files#r768479146 change but `message` is more accurate.

packages/financial-templates-lib/src/logger/SlackTransport.ts
```diff
@@ -49,18 +48,6 @@ interface SlackFormatterResponse {
 }
 // Note: info is any because it comes directly from winston.
 function slackFormatter(info: any): SlackFormatterResponse {
-  // Try and fetch injected web3 which we can use to customize the transaction receipt hyperlink:
```
chrismaree, line 52: this code block is problematic. every time we send a log we are now fetching the chainId to inform logs. Originally, this was ok as we were only running bots on one chain wherein the logger could figure out what chain the tx/address was produced on and perform etherscan link formation. in the current bot context though this makes less sense as we have bots that run over multiple chains.

more over, this code block generates an error if the serverless hub is started without the `--network` flag. every execution of the hub would produce an error, as follows:

image


removing this code has the following consequence: we no longer auto-format addresses/transactions in slack messages. IMHO this is fine as all messages that we send nowadays (relays, predominantly) are formatted from the sender using markdown, which bypasses this logic anyways.

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -15,11 +15,14 @@ const spoke = express();
...
-let logger;
+const { delay, createNewLogger, waitForLogger } = require("@uma/financial-templates-lib");
+
+let customLogger;
+
+const waitForLoggerDelay = process.env.WAIT_FOR_LOGGER_DELAY || 5;
 
-const waitForLogsS = 5;
 spoke.post("/", async (req, res) => {
+  const logger = customLogger || createNewLogger();
```
chrismaree, line 25: I implemented the same logger patten seen in the hub in the spoke. this is not strictly needed here but it was done for two reasons:
1) consistency with the hub
2) in the context you run exactly one spoke (i.e you are running this setup on your local machine) having separate loggers makes things more predictable, as is the case in the hub.

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -105,18 +120,19 @@ function _execShellCommand(cmd, inputEnv, strategyRunnerSpoke = false) {
 // Format stdout outputs. Turns all logs generated while running the script into an array of Json objects.
 function _stripExecStdout(output, strategyRunnerSpoke = false) {
```
chrismaree, line 121: note how this function was changed: there is no mapping going on here. `logsArray` is the full log object generated within the spoke, returned as part of the `stdout`. depending on the log produced we can sepect what parts to emit above.

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -54,17 +57,28 @@ spoke.post("/", async (req, res) => {
     if (execResponse.error) {
       throw execResponse;
     }
+    // Log the full execResponse in this log. This enables you to retrieve the full output of the child process.
```
chrismaree, line 60: this is functionality different: the `Process exited with no error` log now contains the _complete_ `execResponse` log. this contains the _full_, non-redacted spoke output. this should make finding logs and their outputs easier within gcp.

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -54,17 +57,28 @@ spoke.post("/", async (req, res) => {
...
-    await delay(waitForLogsS); // Wait a few seconds to be sure the the winston logs are processed upstream.
+    await waitForLogger(logger);
+    await delay(waitForLoggerDelay); // Wait a few seconds to be sure the the winston logs are processed upstream.
+
+    // Send back a redacted log to the hub, if possible. This is done to decrease the amount of logs that are sent to
+    // the hub. extract only the `message` filed from the logs. This is the main log headline without any details.
     res.status(200).send({
       message: "Process exited with no error",
       childProcessIdentifier: _getChildProcessIdentifier(req),
-      execResponse,
```
chrismaree, line 67: logs returned to the hub are now mapped to contain only the `message`. if there is no `message (or there is not an array of logs) then we return the full execResonce's stdout. this would be the case if the spoke is running a bash file or something equivalent.

packages/financial-templates-lib/src/logger/SlackTransport.ts
```diff
@@ -123,20 +102,11 @@ function slackFormatter(info: any): SlackFormatterResponse {
         }
         // Else, if the input is not an object then print the values as key value pairs. First check for addresses or txs
       } else if (info[key]) {
-        const lastBlock = formattedResponse.blocks[formattedResponse.blocks.length - 1] as SectionBlock;
```
mrice32, line 126: nice simplification

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -258,10 +258,13 @@ hub.post("/", async (req, res) => {
...
     await waitForLogger(logger);
     await delay(waitForLoggerDelay); // Wait a few seconds to be sure the the winston logs are processed upstream.
-    res.status(500).send({
-      message: errorOutput instanceof Error ? "A fatal error occurred in the hub" : "Some spoke calls returned errors",
-      output: errorOutput instanceof Error ? errorOutput.message : errorOutput,
-    });
+    res
+      .status(500)
+      .send({
+        message:
```
mrice32, line 264: Was this just a formatting change? I can't spot the difference :)

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -105,18 +124,19 @@ function _execShellCommand(cmd, inputEnv, strategyRunnerSpoke = false) {
 // Format stdout outputs. Turns all logs generated while running the script into an array of Json objects.
 function _stripExecStdout(output, strategyRunnerSpoke = false) {
   if (!output) return output;
-  // Parse the outputs into a json object to get an array of logs. It is possible that the output is not in a parable form
-  // if the spoke was running a process that did not correctly generate a winston log. In this case simply return the stripped output.
+  // Parse the outputs into a json object to get an array of logs. It is possible that the output is not in a parable
```
mrice32, line 127:
```suggestion
  // Parse the outputs into a json object to get an array of logs. It is possible that the output is not in a parsable
```

packages/serverless-orchestration/src/ServerlessSpoke.js
```diff
@@ -105,18 +124,19 @@ function _execShellCommand(cmd, inputEnv, strategyRunnerSpoke = false) {
...
+  // form if the spoke was running a process that did not correctly generate a winston log. In this case simply return
+  // the stripped output. Note that we use an array to preserve the log ordering.
+  const strippedOutput = _regexStrip(output).replace(/\r?\n|\r/g, ","); // Remove escaped new line chars. Replace with comma between each log output.
```
mrice32, line 130: Not sure if you noticed or if this was intentional, but the stripped output in this line is different than the one originally in the catch -- the catch replaced with a space, but the try replaced with a `,`, so this is _technically_ a functionality change. Probably doesn't matter as I can't understand the original intention behind the difference

mrice32 (approved):
LGTM!

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: ye, and this made the linter fail. I've fixed it and am re-running through CI now.

PR #3709, opened 2021-12-16
fix: [across] add optimism L2->L1 finalizer

chrismaree:
**Motivation**

Across requires the ability to finalise L2->L1 transfers over the canonical optimism bridge. This PR adds the ability to do this for optimism, following a similar pattern to that done in the arbitrum finaliser.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.l2EthersProvider = new providers.Web3Provider(this.l2Web3.currentProvider as any);
+
+    // Fetch contract data.
+    this.l1StateCommitmentChainAddress = getDeployedContractArtifact("StateCommitmentChain", "mainnet").address;
+    this.l2CrossDomainMessengerAddress = predeploys.L2CrossDomainMessenger;
+    this.l1CrossDomainMessenger = new this.l1Web3.eth.Contract(
+      getContractDefinition("IL1CrossDomainMessenger").abi,
+      getDeployedContractArtifact("Proxy__OVM_L1CrossDomainMessenger", "mainnet").address
+    );
+  }
```
mrice32, line 36: Why do we need this method? Why can't we just put all of this in the constructor?

Or, if we just want to maintain the interface, you could just make this method empty.

It seems better to do as much as possible in the constructor since then you don't have to handle all the weird edge cases where stuff isn't defined.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+import BridgeAdapterInterface from "./BridgeAdapterInterface";
+
+import Web3 from "web3";
+import type { TransactionType } from "@uma/common";
+
+export class OptimismBridgeAdapter implements BridgeAdapterInterface {
+  private l1EthersProvider: any;
+  private l2EthersProvider: any;
+  private l1StateCommitmentChainAddress: any;
+  private l2CrossDomainMessengerAddress: any;
```
mrice32, line 19: string?

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+  private l1CrossDomainMessenger: any;
```
mrice32, line 20: Can you pull a well-defined type for this contract from `@uma/contracts-node`?

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getDeployedContractArtifact } from "@eth-optimism/contracts/dist/contract-deployed-artifacts";
+
+import BridgeAdapterInterface from "./BridgeAdapterInterface";
+
+import Web3 from "web3";
+import type { TransactionType } from "@uma/common";
+
+export class OptimismBridgeAdapter implements BridgeAdapterInterface {
+  private l1EthersProvider: any;
+  private l2EthersProvider: any;
```
mrice32, line 17: Don't we have well-defined types for Ethers providers that we could use here?

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+    this.l2CrossDomainMessengerAddress = predeploys.L2CrossDomainMessenger;
+    this.l1CrossDomainMessenger = new this.l1Web3.eth.Contract(
+      getContractDefinition("IL1CrossDomainMessenger").abi,
+      getDeployedContractArtifact("Proxy__OVM_L1CrossDomainMessenger", "mainnet").address
+    );
+  }
+
+  async constructCrossDomainFinalizationTransaction(
+    l2TransactionHash: string
+  ): Promise<{ l2TransactionHash: string; finalizationTransaction: TransactionType | null }> {
```
mrice32, line 40: When would the finalization transaction be null?

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
```diff
@@ -0,0 +1,90 @@
...
+import { getMessagesAndProofsForL2Transaction } from "@eth-optimism/message-relayer";
+import { predeploys, getContractDefinition } from "@eth-optimism/contracts";
+import { getDeployedContractArtifact } from "@eth-optimism/contracts/dist/contract-deployed-artifacts";
+
+import BridgeAdapterInterface from "./BridgeAdapterInterface";
+
+import Web3 from "web3";
+import type { TransactionType } from "@uma/common";
+
+export class OptimismBridgeAdapter implements BridgeAdapterInterface {
```
mrice32, line 15: Just OOC, do we typically test these chain-specific classes? I figure we don't just because we're directly using their library, but wanted to ask just in case.

mrice32 (commented):
Looks good -- a few comments!

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
chrismaree: ye, not tested :) worst that can happen though is it errors or the tx reverts so low risk with that. testing is not worth the pain, given the setup complexities for each chain.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
chrismaree: ye, sorry. I was lazy with this. have fixed

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
chrismaree: we need this method to make the interface conform with the `ArbitrumBridgeAdapter`, which needs this `initialize` method. they both need to conform to the `ExchangeAdapterInterface`.

you right though having it in the constructor is much cleaner. have moved it and set this to just `return`.

packages/insured-bridge-relayer/src/canonical-bridge-adapters/OptimismBridgeAdapter.ts
chrismaree: the tx is null when the L2->L1 transaction is not confirmed. it is not confirmed when not enough time has passed (less than 7 days) or has already been spent.

mrice32 (approved):
LGTM!

PR #3711, opened 2021-12-16
fix: [bots] remove wait for logger

chrismaree:
**Motivation**

The `waitForLogger` utility is extremely inconsistant in production and has proven to be dangerous to use. If this utility is called before all GCP logs have fully flushed out of the stream the process will hang until it times out, producing no logs. This makes it very difficult to figure out what's going wrong with the bots and to debug.

Unfortunately, there is no easy solution to this. Ideally, we want a utility that can halt code execution until all streams have been fully flushed (i.e all upstream messages sent). This is not posible with Winston. Note that their documentation claims this is posible but in practice it just does not work. See [this](https://github.com/winstonjs/winston/issues/1250) PR for the best solution I've found (which inspired the original implementation of `waitForLogger`). Note that the code here does not work any more with GCP's `@google-cloud/logging-winston` package.

There are a number of people complaining about the same issue with `@google-cloud/logging-winston`. see issue [502](https://github.com/googleapis/nodejs-logging-winston/issues/502 ) and issue [598](https://github.com/googleapis/nodejs-logging-winston/issues/598) and there is no clear solution for now.

**Details**

Removes all usage of `waitForLogger` instances and replaces with a `delay`.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/bot-strategy-runner/src/index.ts
+++ b/packages/bot-strategy-runner/src/index.ts
@@ -104,7 +104,7 @@ async function runStrategies(strategyRunnerConfig: strategyRunnerConfig) {
           message: "End of execution loop - terminating process",
         });
 
-      await delay(5);
+      await delay(5); // Set a delay to let the transports flush fully.
       break;
     }
     if (strategyRunnerConfig.emitRunnerLogs)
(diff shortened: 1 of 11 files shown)


mrice32 (approved):
Thanks for debugging this!

PR #3713, opened 2021-12-16
feat(sdk): add detailed gas deposit fees for across

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3644/improve-gas-fee-calculator-in-sdk-to-add-frontend-functionality

**Summary**

This adds a wrapper function to the previous bare bones fee estimator. This roughly matches the output in frontend specific function, but will require some changes for integration. This also adds a "discount" parameter which is optional, to reduce gas costs by the discount rate. 

current frontend definition:
image

sdk definition 
image

**Details**
output of this function with the same params as frontend is returning `0.053160704612800294` in eth transfer fees vs across showing `0.0506`, so its in the same ballpark                                                                                      

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

https://app.shortcut.com/uma-project/story/3644/improve-gas-fee-calculator-in-sdk-to-add-frontend-functionality

### DIFF
--- a/packages/sdk/src/across/README.md
+++ b/packages/sdk/src/across/README.md
@@ -70,7 +70,7 @@ Quickest way to get up and running.
 - Mainnet Ethers provider
 - Address of erc20 token on mainnet, if transfering a token, or `ethers.constants.AddressZero` for ETH.
 
-#### Example
+#### Basic Fees Example
 
 Calculate fees for calling deposit for initiating a relay.
 
@@ -81,7 +81,13 @@ const { gasFeeCalculator, constants, utils } = uma.across
 const totalRelayed = utils.toWei(10)
 const provider = ethers.providers.getDefaultProvider()
 const umaAddress = constants.ADDRESSES.UMA
-const { slowPct, instantPct } = await gasFeecalculator.getDepositFees(provider, totalRelayed, umaAddress)
+const optionalDiscountPercent = 25 // we may decide to discount gas fees by this percent, optionally omit this param
+const { slowPct, instantPct } = await gasFeecalculator.getDepositFees(
+  provider,
+  totalRelayed,
+  umaAddress,
+  optionalDiscountPercent
+)
 
 // example call to deposit, uses the slow/instant percentages from getDepositFees call
 // const tx = await depositBox.deposit(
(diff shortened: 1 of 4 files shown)


amateima (approved):

packages/sdk/src/across/constants.ts
```diff
@@ -13,6 +13,9 @@ export const SPEED_UP_ETH_GAS = 195288;
 export const SPEED_UP_ERC_GAS = 203011;
 export const SPEED_UP_UMA_GAS = 227341;
 
+// Bots offer a discount on gas costs when batching mulitple transactions, this roughly estimates the savings
```
mrice32, line 16:
```suggestion
// Bots incur lower than expected costs due to batching, this roughly estimates the savings
```

packages/sdk/src/across/README.md
```diff
@@ -94,6 +94,28 @@ const { slowPct, instantPct } = await gasFeecalculator.getDepositFees(provider,
...
+This uses the same underlying logic as the basic example, but returns omre information useful for display purposes.
+
+```ts
+import * as uma from "@uma/sdk"
+const { gasFeeCalculator, constants, utils } = uma.across
+
+const totalRelayed = utils.toWei(10)
+const provider = ethers.providers.getDefaultProvider()
+const umaAddress = constants.ADDRESSES.UMA
+const optionalFeeLimitPercent = 25 // this checks if fees are too high as a percentage of total amount to relay, omit to disable check
```
mrice32, line 108: Do you think we should be checking for the hard limits imposed by the contract (in addition to this)? i.e. a slow fee > 25% or fast fee > 25%?

mrice32 (commented):
Looks good! Two minor comments

packages/sdk/src/across/README.md
daywiss: theres is a value `isAmountTooLow` which tells you if you are below the minimum to send. but the actual check here does the logic differently from the contract, which kind of irks me. This function adds fast+slow to be < feeLimit. Where as contract checks fast < feeLimit && slow < feeLimit. Perhaps we can re-evaluate this in a separate pr.

PR #3718, opened 2021-12-20
fix(bots): correct optimism toBlock issue

chrismaree:
**Motivation**

If the cross-domain finaliser looks forward to the current block the OptimismBridgeAdapter will error out as L2 transactions have not yet been included in the L1 StateCommitmentChain. This makes the bot throw errors every time a L2->L1 transfer is initiated until the `StateCommitmentChain` has been updated.

The solution is to add an offset on how new the blocks should be that we look at, thereby avoiding the above situation.

**Summary**

Fixes an issue with fresh blocks in the Optimism finaliser.

### DIFF
--- a/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
+++ b/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
@@ -6,6 +6,7 @@ const argv = minimist(process.argv.slice(), {});
 
 import winston from "winston";
 
+import { across } from "@uma/sdk";
 import {
   createEtherscanLinkMarkdown,
   createFormatFunction,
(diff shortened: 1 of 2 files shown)


packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -230,9 +233,15 @@ export class CrossDomainFinalizer {
     // Note that the query below only works on particular RPC endpoints. Infura, for example, is limited to a 100k look
     // back. This means that to use this module you need to use an endpoint that supports longer lookbacks, such as
     // alchemy, which supports arbitrary long loobacks. In future, Infura will support arbitrary long lookbacks.
+    // Note2: the blockOffset is added to limit how far forward we look. We want to ideally ignore the newest blocks
+    // as these can causes errors in the Bridge Adapters if the L2 transactions are not yet checkpointed on the
+    // canonical L1 StateCommitmentChain. 5000 blocks is ~ 3 hours at current Optimism rate. This is only applied if we
+    // are not running in test mode.
+    const blockOffset =
```
nicholaspai, line 240: I think it'd be better to set this offset via an environment variable. The default value could be 5000 blocks if the bot is being run in prod or 0 in test

nicholaspai (commented):
I think this can be more generalized!

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
nicholaspai: Or we could put the constant value in the ask so the front end would be able to use it also

nicholaspai (approved):

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
mrice32: Not sure why this would ever be needed in the FE, but I think it does make sense to put in the SDK!

Agree on the env variable, but going to go ahead and merge -- we can add that later as necessary.

mrice32 (approved):

PR #3720, opened 2021-12-21
feat: add functionality for relayer to run on multiple chainIds

mrice32:
**Motivation**

We would like to run the relayer for multiple chains in the same process to avoid nonce collisions and allow better batching.

**Summary**

Much of the code was already ready for this change. It only required a top-level Promise.allSettled to loop over all the chainIds we wanted to support.

Note: this should be backwards compatible with all existing configuration. It simply removes a restriction.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/insured-bridge-relayer/src/index.ts
+++ b/packages/insured-bridge-relayer/src/index.ts
@@ -58,10 +58,10 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
     const relayers = await Promise.all(
       config.activatedChainIds.map(async (chainId: number) => {
         // Construct a web3 instance running on L2.
-        const l2Web3 = getWeb3ByChainId(config.activatedChainIds[0]);
+        const l2Web3 = getWeb3ByChainId(chainId);
         const latestL2BlockNumber = await l2Web3.eth.getBlockNumber();
         const l2StartBlock = Math.max(0, latestL2BlockNumber - config.l2BlockLookback);
-        const fallbackL2Web3s = getRetryWeb3sByChainId(config.activatedChainIds[0]);
+        const fallbackL2Web3s = getRetryWeb3sByChainId(chainId);
         // Note: This will not construct duplicate Web3 objects for URL's in the Retry config that are the same as the
         // one used to construct l2Web3.
         logger.debug({
@@ -71,8 +71,8 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
         const l2Client = new InsuredBridgeL2Client(
           logger,
           l2Web3,
-          await l1Client.getL2DepositBoxAddress(config.activatedChainIds[0]),
-          config.activatedChainIds[0],
+          await l1Client.getL2DepositBoxAddress(chainId),
+          chainId,
           l2StartBlock,
           null,
           fallbackL2Web3s
@@ -108,7 +108,7 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
           config.l2BlockLookback
         );
 
-        const canonicalBridgeAdapter = createBridgeAdapter(logger, l1Web3, l2Web3, config.activatedChainIds[0]);
+        const canonicalBridgeAdapter = createBridgeAdapter(logger, l1Web3, l2Web3, chainId);
         if (config.botModes.l1FinalizerEnabled) await canonicalBridgeAdapter.initialize();
 
         const crossDomainFinalizer = new CrossDomainFinalizer(
--- a/packages/insured-bridge-relayer/src/index.ts
+++ b/packages/insured-bridge-relayer/src/index.ts
@@ -23,7 +23,7 @@ import { RelayerConfig } from "./RelayerConfig";
 config();
 
 function isErrorOutput<T>(input: PromiseSettledResult<T>): input is PromiseRejectedResult {
-  return "reason" in input;
+  return input.status === "rejected";
 }
 
 export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
--- a/packages/insured-bridge-relayer/src/index.ts
+++ b/packages/insured-bridge-relayer/src/index.ts
@@ -2,7 +2,7 @@ import winston from "winston";
 import Web3 from "web3";
 import retry from "async-retry";
 import { config } from "dotenv";
-import assert from "assert";
+import lodash from "lodash";
 
 import { getWeb3, getWeb3ByChainId, processTransactionPromiseBatch, getRetryWeb3sByChainId } from "@uma/common";
 
@@ -21,6 +21,10 @@ import { CrossDomainFinalizer } from "./CrossDomainFinalizer";
 import { createBridgeAdapter } from "./canonical-bridge-adapters/CreateBridgeAdapter";
 import { RelayerConfig } from "./RelayerConfig";
 config();
+
+function isErrorOutput<T>(input: PromiseSettledResult<T>): input is PromiseRejectedResult {
+  return "reason" in input;
+}
 
 export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
   try {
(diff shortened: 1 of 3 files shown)


packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-      fallbackL2Web3s
-    );
-
-    // Update the L2 client and filter out tokens that are not whitelisted on the L2 from the whitelisted L1 relay list.
-    const filteredL1Whitelist = await pruneWhitelistedL1Tokens(
-      logger,
-      l1Client,
-      l2Client,
-      config.whitelistedRelayL1Tokens
+    const relayers = await Promise.all(
```
mrice32, line 58: This looks like a lot of changes -- it really just moves all of this logic into a `Promise.all(...map(` block. For vast majority of the logic, it's just an indent change.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
+          config.crossDomainFinalizationThreshold
+        );
+
+        return {
+          relayer,
+          crossDomainFinalizer,
+          l2Client,
+          profitabilityCalculator,
+          filteredL1Whitelist,
+        };
```
mrice32, line 131: This is what's inside the array that the promise.all returns. It will be used in the Promise.allSettled loop below.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-      l1Client,
-      l2Client,
-      canonicalBridgeAdapter,
-      accounts[0],
-      config.l2DeployData,
-      config.crossDomainFinalizationThreshold
-    );
-
+    const combinedFilteredL1Whitelist = lodash.uniq(relayers.map((relayer) => relayer.filteredL1Whitelist).flat());
+    await approveL1Tokens(logger, l1Web3, gasEstimator, accounts[0], config.bridgeAdmin, combinedFilteredL1Whitelist);
```
mrice32, line 138: This just combines the filtered whitelist from all of the L2s in the loop above. It flattens and uniqs the elements before passing into our pre-existing approval function.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-          else logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "L2->L1 finalizer disabled" });
-
-          // Each of the above code blocks could have produced transactions. If they did, their promises are stored
-          // in the executed transactions array. The method below awaits all these transactions to ensure they are
-          // correctly included in a block. if any submitted transactions contains an error then a log is produced.
-          await processTransactionPromiseBatch(
-            [...relayer.getExecutedTransactions(), ...crossDomainFinalizer.getExecutedTransactions()],
-            logger
+      const outputs = await Promise.allSettled(
+        relayers.map(async ({ relayer, crossDomainFinalizer, l2Client, profitabilityCalculator }) => {
```
mrice32, line 141: Similar to the above, this just moves all of the logic in here into a Promise.allSettled loop over the return values from above. The array just stores all the L2-specific objects that we'll need for this loop. Otherwise, all the logic is exactly the same as before.

Note: allSettled is used to ensure that one failing L2 doesn't stop the others from getting through their runs. Only once all runs have been completed do we throw an error.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-              at: "AcrossRelayer#index",
-              message: "An error was thrown in the execution loop - retrying",
-              error: typeof error === "string" ? new Error(error) : error,
-            });
-          },
-        }
+        })
       );
+
+      if (outputs.some(isErrorOutput))
```
mrice32, line 199: Takes the allSettled output and converts it into an error if necessary.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-    // Update the L2 client and filter out tokens that are not whitelisted on the L2 from the whitelisted L1 relay list.
-    const filteredL1Whitelist = await pruneWhitelistedL1Tokens(
-      logger,
-      l1Client,
-      l2Client,
-      config.whitelistedRelayL1Tokens
+    const relayers = await Promise.all(
+      config.activatedChainIds.map(async (chainId: number) => {
+        // Construct a web3 instance running on L2.
+        const l2Web3 = getWeb3ByChainId(config.activatedChainIds[0]);
```
nicholaspai, line 61: wait why do you use `config.activatedChainIds[0]` here shouldn't it just be `chainId`?

mrice32: Yes! Nice catch!

mrice32: This needs to be replaced below too.

mrice32: Probably why the tests are failing :)

mrice32: Fixed.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -22,6 +22,10 @@ import { createBridgeAdapter } from "./canonical-bridge-adapters/CreateBridgeAda
 import { RelayerConfig } from "./RelayerConfig";
 config();
 
+function isErrorOutput<T>(input: PromiseSettledResult<T>): input is PromiseRejectedResult {
+  return "reason" in input;
+}
```
daywiss, line 27: this is really interesting

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -51,129 +55,155 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
-    );
-
-    // Update the L2 client and filter out tokens that are not whitelisted on the L2 from the whitelisted L1 relay list.
-    const filteredL1Whitelist = await pruneWhitelistedL1Tokens(
-      logger,
-      l1Client,
-      l2Client,
-      config.whitelistedRelayL1Tokens
+    const relayers = await Promise.all(
+      config.activatedChainIds.map(async (chainId: number) => {
```
daywiss, line 59: also just want to mention, if a single chain throws an error in here, it will prevent the rest from running, it seems like this is low risk for errors

daywiss (approved):
looks good, have a high level comment. i can see 2 approaches to multiplexing this code, the changes you made here is one example.  the other would treat the entire code as an class which you could loop through all chains at the very top.  This would roughly be somehting like this:
```js
  // in file multi-chain-relayer
  import relayer from 'insured-bridge-relayer'
  
  async run(logger,l1Web3, configs){ // configs for each chain in array
     const results = await Promise.allSettled(configs.map(chainid=>  relayer.run(logger,l1Web3, config) ) 
  }
```

Something roughly like that. The only change to insured bridge relayer is that config can be optionally passed in. This has a couple benefits:
1. reduces risk to working code in relayer
2. maintains the original shape of the config file, reduces complexity
3. you can still run the relayer singly
4. you can have simpler error handling in the multichain relayer without needing special logic in the bridge relayer

I dont know the feasibilty of this approach, but just wanted to mention it

packages/insured-bridge-relayer/src/index.ts
nicholaspai: ha nice

packages/insured-bridge-relayer/src/index.ts
nicholaspai: Yeah we should add this to some general library and use it elsewehere

packages/insured-bridge-relayer/src/index.ts
nicholaspai: yeah ideally we'd perform this set up for each chain, figure out which chains succeeded in setting up, and run the main loop on them. But that code might be too convoluted?

nicholaspai (commented):
I think David's idea is really interesting and would definitely make reviewing a lot easier with a smaller diff. But I don't know if the class approach has too much overhead otherwise

mrice32:
> looks good, have a high level comment. i can see 2 approaches to multiplexing this code, the changes you made here is one example. the other would treat the entire code as an class which you could loop through all chains at the very top. This would roughly be somehting like this:
> 
> ```js
>   // in file multi-chain-relayer
>   import relayer from 'insured-bridge-relayer'
>   
>   async run(logger,l1Web3, configs){ // configs for each chain in array
>      const results = await Promise.allSettled(configs.map(chainid=>  relayer.run(logger,l1Web3, config) ) 
>   }
> ```
> 
> Something roughly like that. The only change to insured bridge relayer is that config can be optionally passed in. This has a couple benefits:
> 
> 1. reduces risk to working code in relayer
> 2. maintains the original shape of the config file, reduces complexity
> 3. you can still run the relayer singly
> 4. you can have simpler error handling in the multichain relayer without needing special logic in the bridge relayer
> 
> I dont know the feasibilty of this approach, but just wanted to mention it

@daywiss this is an interesting approach. The two issues I see with it are:
1. Part of the reason we wanted to run them in the same process with the same web3 is that we can do internal nonce management inside the web3 instance. However, for this to work, they must share a web3 instance. Otherwise, this works almost identically to just running them in completely separate jobs.
2. Another reason we wanted to make this change was so that relays coming from different chains could be batched in the same multicall txns to the bridge. See [this draft](https://github.com/UMAprotocol/protocol/pull/3723) for how that's intended to work (these changes are related, but to keep things simple they are implemented in separate PRs). Similarly to 1, for this batching to work, state needs to be shared between the different relayer runs and then a `send` step must happen afterward.

We could probably initialize the web3 _outside_ of the main relayer setup function. We could also initialize the batcher outside of the main relayer function, pass it in, and then run the send step after. However, this breaks a lot of the current structure. It muddles where we initialize objects and run business logic. It also would mean looping bots never reach the point where we fire off txns.

Basically, my feeling is that we wanted these to run in the same process because we _want_ them to share state. If they share no state, it doesn't offer many (or any?) advantages over just running them in distinct jobs.

Thoughts?

daywiss:
> Another reason we wanted to make this change was so that relays coming from different chains could be batched in the same multicall txns to the bridge. See this draft for how that's intended to work (these changes are related, but to keep things simple they are implemented in separate PRs). Similarly to 1, for this batching to work, state needs to be shared between the different relayer runs and then a send step must happen afterward.

This is a really good reason to not use the approach i outlined, there are optimizations to consider here. There probably ways to get it to work, but it would introduce changes to the business logic which kind of ruins the point of the change.

packages/insured-bridge-relayer/src/index.ts
mrice32: Yeah, this is difficult. As you implied @nicholaspai, I can't really think of a clean way to avoid this. And I do agree this is somewhat low risk for errors.

Thoughts on just leaving this as is?

packages/insured-bridge-relayer/src/index.ts
mrice32: Any suggestions as to where this should go?

daywiss (approved):

packages/insured-bridge-relayer/src/index.ts
daywiss: if we find we are needing this logic in more places id throw it in the sdk/utility file

packages/insured-bridge-relayer/src/index.ts
daywiss: the only scenario i worry about is if a particular chains provider is down, but im not 100% certain this loop would fail, and probaly would fail during the queries in the next loop where you use allSettled.   I'm ok with keeping it as is and taking a wait and see approach,

PR #3723, opened 2021-12-22
feat: add transaction bundler to relayer for better batching

mrice32:
**Motivation**

We'd like the relayer bot to do as much batching as possible.

**Summary**

Rather than doing the batching within the transaction-producing code, this adds a separate bundler that transactions are added to over the course of the bot run. Then, the bundler sorts the transactions by target and fires off batches that are as large as possible.

Note: this upgrades the typescript version to get access to the `Awaited` generic type. This caused some issues with compilation elsewhere in the repo, which were fixed in this PR.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/insured-bridge-relayer/src/MulticallBundler.ts
+++ b/packages/insured-bridge-relayer/src/MulticallBundler.ts
@@ -1,0 +1,179 @@
+import winston from "winston";
+import Web3 from "web3";
+
+import { runTransaction, createEtherscanLinkMarkdown } from "@uma/common";
+import { MultiCallerWeb3 } from "@uma/contracts-node";
+import { GasEstimator } from "@uma/financial-templates-lib";
+
+import type { TransactionType, ExecutedTransaction } from "@uma/common";
+import type { Contract } from "web3-eth-contract";
+import type { TransactionReceipt } from "web3-core";
+
+import lodash from "lodash";
+
+interface BaseCall<T extends TransactionType> {
+  transaction: T;
+  message?: string;
+  mrkdwn?: string;
+  level?: string;
+}
+
+type Call = BaseCall<TransactionType>;
+type ExtendedTransasction = TransactionType & { _parent: Contract };
+type ExtendedCall = BaseCall<ExtendedTransasction>;
+
+function isErrorSettlement<T>(input: PromiseSettledResult<T>): input is PromiseRejectedResult {
+  return input.status === "rejected";
+}
+
+async function allSettledOrError<T>(promises: T[], errorPrefix: string): Promise<Awaited<T>[]> {
+  // Allow all transactions to finish before propagating any errors.
+  const results = await Promise.allSettled(promises);
+  const errors = results.filter(isErrorSettlement).map((errorResult) => errorResult.reason);
+
+  // Throw if there are any errors.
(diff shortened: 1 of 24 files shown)


ru0xjr669 (approved):

packages/insured-bridge-relayer/src/MulticallBundler.ts
```diff
@@ -0,0 +1,179 @@
...
+  message?: string;
+  mrkdwn?: string;
+  level?: string;
+}
+
+type Call = BaseCall<TransactionType>;
+type ExtendedTransasction = TransactionType & { _parent: Contract };
+type ExtendedCall = BaseCall<ExtendedTransasction>;
+
+function isErrorSettlement<T>(input: PromiseSettledResult<T>): input is PromiseRejectedResult {
```
nicholaspai, line 25: This function is also used in `index.ts`, maybe export it there and import here?

nicholaspai (approved):
Is this just a refactor of the existing bundling logic into a separate module, or is the logic for bundling different? If the latter, which situations are bundled differently now?

mrice32:
> Is this just a refactor of the existing bundling logic into a separate module, or is the logic for bundling different? If the latter, which situations are bundled differently now?

@nicholaspai

Right now, the bundling happens per relayer object and per transaction type.

For example, let's say we're running the Optimism and Arbitrum bots in the same process and these four transactions were sent in the same run:
- WETH Relay from Optimism
- WETH Settlement for a relay sent from Optimism
- WETH Relay from Arbitrum
- WETH Settlement for relay sent from Arbitrum

Before this change, this would send 4 separate transactions. After this change, they would be bundled together.

packages/insured-bridge-relayer/src/MulticallBundler.ts
mrice32: I can make a helper file and import it.

mrice32: Done.

packages/insured-bridge-relayer/src/index.ts
```diff
@@ -170,14 +171,16 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
               if (config.botModes.l2FinalizerEnabled) await crossDomainFinalizer.checkForBridgeableL2TokensAndBridge();
               else logger.debug({ at: "AcrossRelayer#CrossDomainFinalizer", message: "L2->L1 finalizer disabled" });
 
+              // The multicall bundler likely accrued transactions over the course of the run.
+              // This call fires off those transactions, but does not wait on them to be mined.
+              // Note: we wait until this point to actually send off the transactions to
+              await multicallBundler.send();
```
nicholaspai, line 177: Doesn't this batch transactions per relayer network? So All arbitrum relays + settles are batched, and on the next run, all optimism ...

nicholaspai (approved):
Just one comment in response to your general comment

packages/insured-bridge-relayer/src/index.ts
mrice32: You're totally right. I merged this incorrectly with the PR that added multichain support.

mrice32: Fixed.

PR #3727, opened 2021-12-24
refactor(cross-chain-oracle): Add HasFinder base contract

nicholaspai:
**Motivation**

I discovered while testing out contracts on live test networks that the `GovernorSpoke` does not make its `finder` address public. This is because the `GovernorSpoke` inherits the `SpokeBase` who's `finder` global variable is private. 

I found this a puzzling design choice so I flipped it to `public`, only to run into compile errors. This is because `OracleBase` also has a `finder` global variable that is `public`, and the `OracleSpoke` inherits both `OracleBase` and `SpokeBase`.

The very simple but naive fix was to add a `finder` public variable to the one contract that _only_ inherits `SpokeBase` but not `OracleBase`: the `GovernorSpoke` contract.

However, it feels weird (cc @mrice32 ) to keep having to declare `finder` in contracts that inherit one another. Instead, we should add a `HasFinder` contract that any contract can inherit and share the same `finder` variable.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
+++ b/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
@@ -10,12 +10,17 @@ import "./SpokeBase.sol";
  * @notice Governor contract deployed on L2 that receives governance actions from Ethereum.
  */
 contract GovernorSpoke is Lockable, SpokeBase, ChildMessengerConsumerInterface {
+    // Added as a convenient view method to determine this contract's set Finder address. The finder address stored
+    // in the SpokeBase is private for reasons specific to that contract, see its comments.
+    address public finder;
     struct Call {
         address to;
         bytes data;
     }
 
-    constructor(address _finderAddress) SpokeBase(_finderAddress) {}
+    constructor(address _finderAddress) SpokeBase(_finderAddress) {
+        finder = _finderAddress;
+    }
 
     event ExecutedGovernanceTransaction(address indexed to, bytes data);
 
--- a/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
+++ b/packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
@@ -23,12 +23,12 @@ contract GovernorSpoke is Lockable, SpokeBase, ChildMessengerConsumerInterface {
      * @notice Executes governance transaction created on Ethereum.
      * @dev Can only be called by ChildMessenger contract that wants to execute governance action on this child chain
      * that originated from DVM voters on root chain. ChildMessenger should only receive communication from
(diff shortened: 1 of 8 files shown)


packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
```diff
@@ -10,12 +10,17 @@ import "./SpokeBase.sol";
  * @notice Governor contract deployed on L2 that receives governance actions from Ethereum.
  */
 contract GovernorSpoke is Lockable, SpokeBase, ChildMessengerConsumerInterface {
+    // Added as a convenient view method to determine this contract's set Finder address. The finder address stored
+    // in the SpokeBase is private for reasons specific to that contract, see its comments.
+    address public finder;
```
mrice32, line 15: The pattern of redeclaring the finder at every inheritance level seems like we're doing something wrong?

My (potentially misguided) fear is that we have multiple sources of truth in the contract as to what the _real_ finder value is. This means we might _think_ the true value is what's returned by the public function, but there is another finder being used at a different inheritance level that could be a different address. I don't think the current contract will allow them to be set differently (especially since they can't be reset), but it would only take a small typo in a constructor for one to be set to `0x0` accidentally.

Should we, instead, have a universal base class to hold and expose the finder?

I tested the following code and it seemed to work:
```
pragma solidity ^0.8.9;

// SPDX-License-Identifier: UNLICENSED

contract HasFinder {

    address public finder;

    constructor(address _finder) {
        finder = _finder;
    }
}

abstract contract OracleBase is HasFinder {
    function doThingWithFinder() public returns (bool, bytes memory) {
        return finder.call("");
    } 
}

abstract contract SpokeBase is HasFinder {
    function doOtherThingWithFinder() public returns (bool, bytes memory) {
        return finder.call("");

    }
}

contract GovernorSpoke is SpokeBase {
    constructor(address _finder) HasFinder(_finder) {}
}

contract OracleSpoke is OracleBase, SpokeBase {
    constructor(address _finder) HasFinder(_finder) {}
}
```

This would likely be a bit more churn, since a new base class for SpokeBase and OracleBase would be added and their constructors would be removed, but it would eliminate all repeated finder declarations.

Thoughts?

mrice32 (commented):
One high-level comment

packages/core/contracts/cross-chain-oracle/GovernorSpoke.sol
nicholaspai: I like this! I'll add

nicholaspai: The churn looks to be only on the contract side, no tests which would be easy

mrice32 (approved):
LGTM!

PR #3728, opened 2021-12-27
feat(sdk): basic erc20 and erc20 multicall client

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3779/add-basic-erc20-service-for-oo-in-sdk

**Summary**

This basic client will be used during the approval flow in the full oracle client eventually. Factory creates a client with or without multicall as multicall is not available on all chains.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -93,7 +93,13 @@ class PoolState {
       address: this.address,
       l1Token: this.l1Token,
       exchangeRatePrevious,
-      ...(await this.batchRead([
+      ...(await this.batchRead<{
+        exchangeRateCurrent: BigNumber;
+        liquidityUtilizationCurrent: BigNumber;
+        liquidReserves: BigNumber;
+        pendingReserves: BigNumber;
+        utilizedReserves: BigNumber;
+      }>([
         // its important exchangeRateCurrent is called first, as it calls _sync under the hood which updates the contract
         // and gives more accurate values for the following properties.
         ["exchangeRateCurrent"],
--- a/packages/sdk/src/oracle/services/erc20.ts
+++ b/packages/sdk/src/oracle/services/erc20.ts
@@ -1,0 +1,43 @@
+import { erc20 } from "../../clients";
+import Multicall2 from "../../multicall2";
+import { BatchReadWithErrors, BatchReadWithErrorsType, Calls } from "../../utils";
+import { BigNumber, Provider } from "../types/ethers";
+
+const batchProps: Calls = [["symbol"], ["name"], ["decimals"], ["totalSupply"]];
+type Props = {
+  symbol: string;
+  name: string;
+  decimals: number;
+  totalSupply: BigNumber;
+};
+export class Erc20 {
+  public contract: erc20.Instance;
+  constructor(protected provider: Provider, protected address: string) {
+    this.contract = erc20.connect(address, provider);
+  }
+  async getProps(): Promise<Props> {
+    const { contract } = this;
+    return {
+      symbol: await contract.callStatic.symbol(),
+      name: await contract.callStatic.name(),
+      decimals: await contract.callStatic.decimals(),
+      totalSupply: await contract.callStatic.totalSupply(),
+    };
+  }
+}
+export class Erc20Multicall extends Erc20 {
+  private multicall: Multicall2;
+  private batchRead: BatchReadWithErrorsType;
+  constructor(provider: Provider, address: string, private multicallAddress: string) {
+    super(provider, address);
+    this.multicall = new Multicall2(multicallAddress, this.provider);
+    this.batchRead = BatchReadWithErrors(this.multicall)(this.contract);
+  }
+  async getProps(): Promise<Props> {
+    return this.batchRead<Props>(batchProps);
+  }
+}
+export function factory(provider: Provider, address: string, multicallAddress?: string) {
+  if (!multicallAddress) return new Erc20(provider, address);
+  return new Erc20Multicall(provider, address, multicallAddress);
+}
--- a/packages/sdk/src/utils.ts
+++ b/packages/sdk/src/utils.ts
@@ -72,9 +72,12 @@ export async function loop(fn: (...args: any[]) => any, delay: number, ...args: 
   } while (true);
 }
 
-type Call = [string, ...any[]];
-type Calls = Call[];
-export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async (calls: Calls) => {
+export type Call = [string, ...BigNumberish[]];
+export type Calls = Call[];
+export type BatchReadWithErrorsType = ReturnType<ReturnType<typeof BatchReadWithErrors>>;
+export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async <R>(
+  calls: Calls
+): Promise<R> => {
   // multicall batch takes array of {method} objects
   const results = await multicall2
     .batch(
--- a/packages/sdk/src/index.ts
+++ b/packages/sdk/src/index.ts
@@ -3,6 +3,7 @@ export * as stores from "./stores";
 export * as tables from "./tables";
 export * as utils from "./utils";
 export * as across from "./across";
+export * as oracle from "./oracle";
 export { default as Coingecko } from "./coingecko";
 export { default as Multicall } from "./multicall";
 export { default as Multicall2 } from "./multicall2";
--- a/packages/sdk/src/oracle/index.ts
+++ b/packages/sdk/src/oracle/index.ts
@@ -1,0 +1,1 @@
+export * as services from "./services";
--- a/packages/sdk/src/oracle/services/index.ts
+++ b/packages/sdk/src/oracle/services/index.ts
@@ -1,0 +1,1 @@
+export * as erc20 from "./erc20";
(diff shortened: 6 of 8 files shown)


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -93,7 +93,13 @@ class PoolState {
...
       l1Token: this.l1Token,
       exchangeRatePrevious,
-      ...(await this.batchRead([
+      ...(await this.batchRead<{
+        exchangeRateCurrent: BigNumber;
+        liquidityUtilizationCurrent: BigNumber;
+        liquidReserves: BigNumber;
+        pendingReserves: BigNumber;
+        utilizedReserves: BigNumber;
+      }>([
```
daywiss, line 102: The batch read function was made generic so we can properly type the results. needed to update the call in the bridge pool

packages/sdk/src/oracle/services/erc20.ts
```diff
@@ -0,0 +1,43 @@
...
+  name: string;
+  decimals: number;
+  totalSupply: BigNumber;
+};
+export class Erc20 {
+  public contract: erc20.Instance;
+  constructor(protected provider: Provider, protected address: string) {
+    this.contract = erc20.connect(address, provider);
+  }
+  async getProps(): Promise<Props> {
```
daywiss, line 18: spent several hours trying to make this call dynamic, ie provide the props you want rather than hard coding it. could not get typescript to pass.

packages/sdk/src/oracle/services/erc20.ts
```diff
@@ -0,0 +1,43 @@
...
+    const { contract } = this;
+    return {
+      symbol: await contract.callStatic.symbol(),
+      name: await contract.callStatic.name(),
+      decimals: await contract.callStatic.decimals(),
+      totalSupply: await contract.callStatic.totalSupply(),
```
daywiss, line 24: this would have been better as promise.all, but could not get typescript to pass. its slightly more complicated because of the override in the multicall class. so ended up with this solution

packages/sdk/src/utils.ts
```diff
@@ -72,9 +72,12 @@ export async function loop(fn: (...args: any[]) => any, delay: number, ...args:
...
 
-type Call = [string, ...any[]];
-type Calls = Call[];
-export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async (calls: Calls) => {
+export type Call = [string, ...BigNumberish[]];
+export type Calls = Call[];
+export type BatchReadWithErrorsType = ReturnType<ReturnType<typeof BatchReadWithErrors>>;
+export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async <R>(
+  calls: Calls
+): Promise<R> => {
```
daywiss, line 80: improved the batch call with generic

packages/sdk/src/oracle/services/erc20.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const batchProps: Calls = [["symbol"], ["name"], ["decimals"], ["totalSupply"]];
+type Props = {
+  symbol: string;
+  name: string;
+  decimals: number;
+  totalSupply: BigNumber;
+};
+export class Erc20 {
+  public contract: erc20.Instance;
```
daywiss, line 14: could not find a way to extend from this, due to the factory being used to instantiate in typechain

packages/sdk/src/oracle/services/erc20.ts
mrice32: What do you mean "could not get typescript to pass"? Any reason not to multicall these?

packages/sdk/src/utils.ts
```diff
@@ -72,9 +72,12 @@ export async function loop(fn: (...args: any[]) => any, delay: number, ...args:
   } while (true);
 }
 
-type Call = [string, ...any[]];
-type Calls = Call[];
-export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async (calls: Calls) => {
+export type Call = [string, ...BigNumberish[]];
+export type Calls = Call[];
+export type BatchReadWithErrorsType = ReturnType<ReturnType<typeof BatchReadWithErrors>>;
+export const BatchReadWithErrors = (multicall2: Multicall2) => (contract: Contract) => async <R>(
```
mrice32, line 78: Minor note: is there a strong reason to write this function this way? Not saying we should change it since it is pre-existing, but I think the call syntax `BatchReadWithErrors(this.multicall)(this.contract)(calls)` is not very idiomatic for js/ts and can be confusing at the callsite.

packages/sdk/src/oracle/services/erc20.ts
mrice32: Doesn't seem like a huge deal, but it is odd that it was difficult considering it should be nearly identical to batchRead, right?

packages/sdk/src/oracle/services/erc20.ts
mrice32: Can you explain further?

packages/sdk/src/oracle/services/erc20.ts
```diff
@@ -0,0 +1,43 @@
...
+import { BigNumber, Provider } from "../types/ethers";
+
+const batchProps: Calls = [["symbol"], ["name"], ["decimals"], ["totalSupply"]];
+type Props = {
+  symbol: string;
+  name: string;
+  decimals: number;
+  totalSupply: BigNumber;
+};
+export class Erc20 {
```
mrice32, line 13: Why did you create two classes Erc20 and Erc20 multicall. Why not just make one class that includes all the functions? If for modularity reasons, I'm having trouble seeing the separation between the two? They both offer a client to the ERC20 token, one just seems to add a little dynamism, right?

mrice32 (approved):
Looks great! Only minor comments and questions

packages/sdk/src/oracle/services/erc20.ts
daywiss: Ideally i twould be nice to be able to extend the contract class and just add functions to it, but the implementation is required to do this and only the factory is exposed afaik.  Given that, its not clear how you would instantiate "this" with a factory, rather than a constructor.

packages/sdk/src/oracle/services/erc20.ts
daywiss: I would have to go back and recreate the conditions to know the error, but typescript would keep complaining when i tried to convert an array of values back into an object of a certain type.

packages/sdk/src/oracle/services/erc20.ts
daywiss: Part of the prolbem is that you are using user supplied prop arguments, which i couldnt figure out how to convert back into a known type to typescripts satisfaction. I dont remember the exact errors though

packages/sdk/src/utils.ts
daywiss: The only reason really is so you can build multiple batch readers from a single provider and multicall instance. In a more complex scenario, you would instantiate a single level with your multicall contract at the top of your app, and pass that down to specific components which could pass the contract they want to multicell. From there the specific component could use that to make batch calls on that contract in multiple places.

This pattern is specific to the type of client i plan to build, but most likely you would just want to use the multicall class directly.

packages/sdk/src/oracle/services/erc20.ts
daywiss: the multicall erc20 extends from the plain erc20, overriding the getProps function. Given that the app is multichain and we dont know if a chain has multicall deployed, i figured we would want a non multicall option. The mutlicall option simply needs to override calls on the base class to optimize.

packages/sdk/src/oracle/services/erc20.ts
mrice32: Interesting. I totally agree that this sort of thing is difficult to make work in ts, but I'm surprised it didn't just follow the same pattern as batchRead which takes user-defined args and produces typed outputs. I could be misunderstanding batchRead though.

packages/sdk/src/oracle/services/erc20.ts
mrice32: Oh I see. I think it's actually kind of nice to keep the client and contract objects separate. But your explanation totally makes sense.

packages/sdk/src/oracle/services/erc20.ts
mrice32: Oh interesting. Do we assume multicall exists everywhere in other parts of the SDK?

I just thought we assumed this existed everywhere. Seems reasonable to leave a non-multicall option if that's what we do elsewhere.

mrice32 (approved):

packages/sdk/src/oracle/services/erc20.ts
daywiss: i didnt try to add generics and require the user to type cast the call, like in batch read. not sure why i didnt think of it

packages/sdk/src/oracle/services/erc20.ts
daywiss: We learned from the API that when deploying to polygon, we couldn't assume multicall existed. Our solution in that case was to deploy our own multicall contract. So if i know now that the app will run on multi chain, i dont make that assumption anymore.  The only other place multicall is used is the across pool client, which its assumed the liquidity pools are on mainnet.

packages/sdk/src/oracle/services/erc20.ts
daywiss: also i dont know why typescript does not complain about the batch read function. maybe because we arent calling functions  dynamically on the multicall contract, where as in this call we would be using these user passed strings as ways to key into the contract methods dynamically

daywiss: i spent some more time on this, probably not going to keep trying. but i eventually get to the point where typescript needs me to validate the arguments to the call, which is maybe possible but i dont know

image
image

PR #3729, opened 2021-12-27
feat(sdk): add basic optimistic oracle client for getting requests

daywiss:
**Motivation**
https://app.shortcut.com/uma-project/story/3779/add-basic-erc20-service-for-oo-in-sdk

**Summary**

Very basic OO client with functionality for skinny oo ui

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/constants.ts
+++ b/packages/sdk/src/oracle/constants.ts
@@ -1,7 +1,5 @@
 import { ethers } from "ethers";
 
-export const MULTICALL2_ADDRESS = "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
-
 export enum ChainId {
   MAINNET = 1,
   RINKEBY = 4,
@@ -17,7 +15,7 @@ export const AddressZero = ethers.constants.AddressZero;
 
 export const Addresses = {
   [ChainId.MAINNET]: {
-    multicall2: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
+    multicall2: "0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696",
     optimisticOracle: "0xC43767F4592DF265B4a9F1a398B97fF24F38C6A6",
     erc20: {
       ETH: AddressZero,
--- a/packages/sdk/src/oracle/constants.ts
+++ b/packages/sdk/src/oracle/constants.ts
@@ -1,0 +1,29 @@
+import { ethers } from "ethers";
+
+export const MULTICALL2_ADDRESS = "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
+
+export enum ChainId {
+  MAINNET = 1,
+  RINKEBY = 4,
+  KOVAN = 42,
+  OPTIMISM = 10,
+  KOVAN_OPTIMISM = 69,
+  ARBITRUM = 42161,
+  ARBITRUM_RINKEBY = 421611,
+  BOBA = 288,
+}
+
+export const AddressZero = ethers.constants.AddressZero;
+
+export const Addresses = {
+  [ChainId.MAINNET]: {
+    multicall2: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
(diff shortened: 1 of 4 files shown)


packages/sdk/src/oracle/services/optimisticOracle.ts
```diff
@@ -0,0 +1,22 @@
...
+  public contract: OptimisticOracleEthers;
+  constructor(protected provider: Provider, protected address: string) {
+    this.contract = OptimisticOracleEthers__factory.connect(address, provider);
+  }
+  async getRequest(requester:string,identifier:string, timestamp:BigNumberish, ancillaryData:string): Promise<Request>{
+    return this.contract.callStatic.getRequest(requester, identifier, timestamp, ancillaryData)
+  }
+  async getProps():Promise<Props>{
+    return {
+      liveness: await this.contract.defaultLiveness()
```
mrice32, line 19: Do you think the name "liveness" is a bit misleading since some requests may specify a different liveness? Or do you think this self-explanatory since the request object should have a liveness value as well.

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/services/optimisticOracle.ts
daywiss: yea you are right, i should change this to be more literal, "defaultLiveness"

packages/sdk/src/oracle/services/optimisticOracle.ts
```diff
@@ -0,0 +1,22 @@
+import { OptimisticOracleEthers, OptimisticOracleEthers__factory } from "@uma/contracts-node";
+import { BigNumberish, BigNumber, Provider } from "../types/ethers";
+
+type Props = {
+  liveness: BigNumber;
+}
+type Request = ReturnType<OptimisticOracleEthers['getRequest']>
+
+export class OptimisticOracle {
+  public contract: OptimisticOracleEthers;
```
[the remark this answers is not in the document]
daywiss: interesting, yea can do that

packages/sdk/src/oracle/services/optimisticOracle.ts
daywiss: im not 100% sure, but it seems if a request does not override the liveness, liveness wont available on the request object which makes this call necessary.  is that the case?

packages/sdk/src/oracle/constants.ts
```diff
@@ -0,0 +1,29 @@
+import { ethers } from "ethers";
+
+export const MULTICALL2_ADDRESS = "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
+
+export enum ChainId {
+  MAINNET = 1,
```
nicholaspai, line 6: Can you grab these from `@uma/common/PublicNetworks` even though they are in reverse (map chain ID to network name)?

daywiss: yea i can take a look at that. this was actually copied from the frontend constants in the across app.

daywiss: this looks really useful, but cant be imported due to incompatibility with common and the FE.  Going to keep this like this for now because i think this is going to evolve more as this gets more built out. We should though talk about adding constants like this to the SDK in general.

packages/sdk/src/oracle/constants.ts
```diff
@@ -0,0 +1,29 @@
+import { ethers } from "ethers";
+
+export const MULTICALL2_ADDRESS = "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
```
amateima, line 3: I imagine in the future this will need to depend on the chain id (when we'll deploy OO on other chains)

amateima (approved):
👍

packages/sdk/src/oracle/constants.ts
daywiss: good call, yea i can move that now

PR #3731, opened 2021-12-28
feat: add manual relay script

mrice32:
**Motivation**

We would like to have a script to manually relay a depost.

**Summary**

Adds a script to manually relay a deposit.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/scripts/src/ManualRelay.js
+++ b/packages/scripts/src/ManualRelay.js
@@ -1,0 +1,76 @@
+#!/usr/bin/env node
+
+// This script allows you to manually relay a particular across deposit.
+// Example:
+// ./src/DecodeTransactionData.js --chainId 10 --depositId 100
+// Optional env overrides:
+//   BRIDGE_ADMIN_ADDRESS
+//   BRIDGE_DEPOSIT_ADDRESS
+//   L2_END_BLOCK_NUMBER
+
+const argv = require("minimist")(process.argv.slice(), { string: ["depositId", "chainId"] });
+
+const { getWeb3, getWeb3ByChainId } = require("@uma/common");
+const { InsuredBridgeL1Client, InsuredBridgeL2Client, Logger } = require("@uma/financial-templates-lib");
+const sdk = require("@uma/sdk");
+
+const depositBoxMapping = {
+  10: "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96",
+  42161: "0xD8c6dD978a3768F7DDfE3A9aAD2c3Fd75Fa9B6Fd",
+  288: "0xCD43CEa89DF8fE39031C03c24BC24480e942470B",
+};
+
+async function main() {
+  if (!argv.chainId) {
+    throw new Error("You must provide the --chainId argument");
+  } else if (!argv.depositId) {
+    throw new Error("--depositId argument must be provided");
+  }
+
+  const l1Web3 = getWeb3();
+  const l2Web3 = getWeb3ByChainId(Number(argv.chainId));
+  const [account] = l1Web3.eth.getAccounts();
+  const latestL2BlockNumber = process.env.L2_END_BLOCK_NUMBER | (await l2Web3.eth.getBlockNumber());
+
+  const l1Client = new InsuredBridgeL1Client(
+    Logger,
+    l1Web3,
+    process.env.BRIDGE_ADMIN_ADDRESS | "0x30B44C676A05F1264d1dE9cC31dB5F2A945186b6",
+    sdk.across.constants.RATE_MODELS
+  );
+  await l1Client.update();
+
+  const l2Client = new InsuredBridgeL2Client(
+    Logger,
+    l2Web3,
+    process.env.BRIDGE_DEPOSIT_ADDRESS || depositBoxMapping[Number(argv.chainId)],
+    Number(argv.chainId),
+    latestL2BlockNumber - 9900,
+    latestL2BlockNumber
+  );
+  await l2Client.update();
+
+  const deposit = l2Client.getAllDeposits().find((deposit) => deposit.depositId === Number(argv.depositId));
+  if (!deposit) throw new Error(`No deposit found for ${argv.depositId}`);
+
+  const txn = l1Client.bridgePools[deposit.l1Token].contract.methods.relayAndSpeedUp(
+    deposit,
+    await l1Client.calculateRealizedLpFeePctForDeposit(deposit)
+  );
+
+  // Verify the txn goes through.
+  await txn.call({ from: account });
+
+  // Send to network.
+  await txn.send({ from: account });
+}
+
+main().then(
+  () => {
+    process.exit(0);
+  },
+  (error) => {
+    console.error(error);
+    process.exit(1);
+  }
+);
--- a/packages/scripts/src/ManualRelay.js
+++ b/packages/scripts/src/ManualRelay.js
@@ -19,12 +19,7 @@ const argv = require("minimist")(process.argv.slice(), { string: ["depositId", "
 const { getWeb3ByChainId } = require("@uma/common");
 const { InsuredBridgeL1Client, InsuredBridgeL2Client, Logger } = require("@uma/financial-templates-lib");
 const sdk = require("@uma/sdk");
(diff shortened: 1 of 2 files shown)


packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,76 @@
+#!/usr/bin/env node
+
+// This script allows you to manually relay a particular across deposit.
+// Example:
+// ./src/DecodeTransactionData.js --chainId 10 --depositId 100
```
smb2796, line 5:
```suggestion
// ./src/ManualRelay.js --chainId 10 --depositId 100
```

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env node
+
+// This script allows you to manually relay a particular across deposit.
+// Example:
```
nicholaspai, line 4: So this manually relays a deposit originating from chain ID 10 on chain ID 1? Maybe its worth a quick note explaining why you specify _1 and _10 NODE_URL here

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,82 @@
...
+//   --network mainnet_gckms --keys your_gckms_key_name
+
+const argv = require("minimist")(process.argv.slice(), { string: ["depositId", "chainId"] });
+
+const { getWeb3ByChainId } = require("@uma/common");
+const { InsuredBridgeL1Client, InsuredBridgeL2Client, Logger } = require("@uma/financial-templates-lib");
+const sdk = require("@uma/sdk");
+
+const depositBoxMapping = {
+  10: "0x3baD7AD0728f9917d1Bf08af5782dCbD516cDd96",
```
nicholaspai, line 24: nit: why not read these from `getAddress`? You'd still need to use a constant string to fetch the contract like `getAddress("OVM_BridgeDepositBox")` but at least you don't need to repeat hardcode the address

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const l1Web3 = getWeb3ByChainId(1);
+  const l2Web3 = getWeb3ByChainId(Number(argv.chainId));
+  const [account] = await l1Web3.eth.getAccounts();
+  const latestL2BlockNumber = process.env.L2_END_BLOCK_NUMBER || (await l2Web3.eth.getBlockNumber());
+
+  const l1Client = new InsuredBridgeL1Client(
+    Logger,
+    l1Web3,
+    process.env.BRIDGE_ADMIN_ADDRESS || "0x30B44C676A05F1264d1dE9cC31dB5F2A945186b6",
```
nicholaspai, line 44: nit: read default bridge admin address from `getAddress`

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,82 @@
...
+    Number(argv.chainId),
+    latestL2BlockNumber - 9900,
+    latestL2BlockNumber
+  );
+  await l2Client.update();
+
+  const deposit = l2Client.getAllDeposits().find((deposit) => deposit.depositId === Number(argv.depositId));
+  if (!deposit) throw new Error(`No deposit found for ${argv.depositId}`);
+
+  const txn = l1Client.bridgePools[deposit.l1Token].contract.methods.relayAndSpeedUp(
```
nicholaspai, line 62: Should we add some logs or warning or checking about whether caller has enough funds to post a proposal bond and relay?

nicholaspai (commented):
Great structure overall, left some nits!

packages/scripts/src/ManualRelay.js
mrice32: Won't it just error? Is there a reason it's preferable to manually check for that ahead of time as opposed to just throwing the error?

packages/scripts/src/ManualRelay.js
mrice32: Oh interesting. For some reason, I thought the across addresses weren't in the main addresses files.

packages/scripts/src/ManualRelay.js
mrice32: Sure

packages/scripts/src/ManualRelay.js
mrice32: Fixed.

packages/scripts/src/ManualRelay.js
mrice32: Fixed.

packages/scripts/src/ManualRelay.js
mrice32: Done

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,85 @@
...
+
+const argv = require("minimist")(process.argv.slice(), { string: ["depositId", "chainId"] });
+
+const { getWeb3ByChainId } = require("@uma/common");
+const { InsuredBridgeL1Client, InsuredBridgeL2Client, Logger } = require("@uma/financial-templates-lib");
+const sdk = require("@uma/sdk");
+const { getAddress } = require("@uma/contracts-node");
+
+async function main() {
+  if (!argv.chainId) {
```
chrismaree, line 27: worth while validating that this `chainId` is part of the known list of: 10, 42161 or 288?

packages/scripts/src/ManualRelay.js
chrismaree: I agree it'll throw an error but perhaps add some more logs before and after the `.call` and `.send` blocks below so the runner of the script knows where it is in the execution.

packages/scripts/src/ManualRelay.js
```diff
@@ -0,0 +1,85 @@
...
+  const txn = l1Client.bridgePools[deposit.l1Token].contract.methods.relayAndSpeedUp(
+    deposit,
+    await l1Client.calculateRealizedLpFeePctForDeposit(deposit)
+  );
+
+  // Verify the txn goes through.
+  await txn.call({ from: account });
+
+  // Send to network.
+  await txn.send({ from: account });
```
chrismaree, line 74: this does not have any custom gas logic. I think this script should use our gas estimator or we'll likely have issues with underpriced transactions (hopefully web3 gracefully handles this now post 1559 but I'm not convinced it does).

mrice32: I actually used this to manually relay, and it did work gracefully, but happy to add custom gas logic.

packages/scripts/src/ManualRelay.js
mrice32: Any reason we don't want to allow people to add addresses via env and use this on a testnet deployment?

packages/scripts/src/ManualRelay.js
mrice32: Can do!

packages/scripts/src/ManualRelay.js
chrismaree: that's fair. main issue I had was you could provide an invalid chainId then it would fail in a confusing way but I guess leaving it less opinionated to open up test net usage is better.

packages/scripts/src/ManualRelay.js
chrismaree: ok cool. if it worked for you then dont worry about it :)

mrice32: Sounds good!

packages/scripts/src/ManualRelay.js
mrice32: I added a generic error message to the end of the script suggesting that the user double check their chain id.

mrice32: I added a specific error if they don't provide a custom deposit box address _and_ their chainId isn't listed in the object.

mrice32:
@chrismaree @nicholaspai any additional comments on this?

chrismaree (approved):
lgtm! thanks for adding in all the comments!

PR #3734, opened 2021-12-30
fix: fix profitability module

mrice32:
**Motivation**

There is a bug in the profitability module that affects tokens whose decimals are != 18.

**Summary**

The profitability module now uses the token's decimals to convert between the amount of the token and its corresponding value in ETH.

Note: this doesn't add new tests. Rather, it modifies existing tests to take decimals into account.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


nicholaspai (approved):
Nice fix, I really liked that we didn't have to pass `web3` into this module but i don't see a way to read `decimals` from the token otherwise, unless we do it in the L1 client

PR #3735, opened 2021-12-30
feat(core): Add OO deployment instructions

nicholaspai:
**Motivation**

I used existing hardhat `tasks` to deploy the `OptimisticOracle` to arbitrum but I formally wrote up instructions for future users in this PR.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/common/src/Constants.ts
```diff
@@ -18,6 +18,15 @@ export const interfaceName = {
   OracleHub: "OracleHub",
 };
 
+// Updated addresses can be found here: https://github.com/makerdao/multicall
+export const MULTICALL2_DEPLOYMENTS: { [chainId: number]: string } = {
+  1: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
+  42: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
+  4: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
+  5: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
+  3: "0x5ba1e12693dc8f9c48aad8770482f4739beed696",
```
chrismaree, line 27: few questions: 1) these are all the same. why do we need to store them? 2) should this not be checksum case?

packages/core/contracts/cross-chain-oracle/README.md
```diff
@@ -65,4 +70,25 @@ yarn hardhat setup-l1-arbitrum-cross-chain --network mainnet
 yarn hardhat setup-l2-arbitrum-cross-chain --network arbitrum
 ```
 
-7. At this point, the minimum set up is complete and further integration tests can be executed. Additionally, an OptimisticOracle can now be deployed to Arbitrum that can communicate back to mainnet via the newly deployed contracts.
+7. At this point, the cross chain contract suite setup is complete. We will now deploy the `OptimisticOracle` and required contracts to the L2.
```
chrismaree, line 73: this is fresh

packages/common/src/Constants.ts
nicholaspai: but what if the chain doesn't have a deployment for this contract? Should we instead just store a list of networks that do have this contract, and then hardcode the address?

mrice32: Do we not store this in `networks/` because they aren't _our_ system contracts?

If so, should we make this a standard place to store hardcoded external contract addresses by chain id?

contractName -> chainId -> address or chainId -> contractName -> address, maybe?

packages/common/src/hardhat/tasks/identifiers.ts
```diff
@@ -27,6 +28,47 @@ function isString(input: string | null): input is string {
...
+    console.log(`Using multicall contract @ ${multicallAddress} to reduce web3 requests`);
+    const Multicall2 = hre.getContract("Multicall2");
+    const multicall = new web3.eth.Contract(Multicall2.abi, multicallAddress);
+    const calls = [];
+    for (const id of identifiersToCheck) {
+      calls.push({
+        target: whitelistToCheck.options.address,
+        callData: whitelistToCheck.methods.isIdentifierSupported(id).encodeABI(),
+      });
+    }
```
mrice32, line 53: nit:

```ts
const calls = indentifiersToCheck.map((id) => ({
        target: whitelistToCheck.options.address,
        callData: whitelistToCheck.methods.isIdentifierSupported(id).encodeABI(),
}));

File: packages/common/src/hardhat/tasks/identifiers.ts [hunk 4]
```diff
@@ -27,6 +28,47 @@ function isString(input: string | null): input is string {
...
+    const result = await multicall.methods.aggregate(calls).call();
+    for (const _result of result.returnData) {
+      // Multicall contract returns results as bytes, so we need to check return value against bytes representation of
+      // Boolean value.
+      whitelistedIdentifiers.push(_result === "0x0000000000000000000000000000000000000000000000000000000000000001");
+    }
```
mrice32, line 59:
```suggestion
    return result.returnData.map((_result) => {
      // Multicall contract returns results as bytes, so we need to check return value against bytes representation of
      // Boolean value.
      return _result === "0x0000000000000000000000000000000000000000000000000000000000000001";
    });
```

packages/common/src/hardhat/tasks/identifiers.ts
```diff
@@ -27,6 +28,47 @@ function isString(input: string | null): input is string {
...
+  } else {
+    console.log(
+      `No multicall contract found for network ${networkId}, submitting ${identifiersToCheck.length} web3 requests, sit tight`
+    );
+    for (const id of identifiersToCheck) {
```
mrice32, line 64: Are these intentionally done in series to avoid overloading the node? I only ask because it could be faster if we used an `await Promise.all(identifierToCheck.map(`.

packages/core/networks/42161.json
```diff
@@ -22,5 +22,17 @@
   {
     "contractName": "OracleSpoke",
     "address": "0x196767568F848fF7607145A7C1f3CDE44bb4Ce98"
+  },
+  {
+    "contractName": "IdentifierWhitelist",
+    "address": "0x2914BE6CD56276d208299D2062848b059FC10588"
+  },
+  {
+    "contractName": "AddressWhitelist",
```
mrice32, line 31: I'm assuming we don't intend to have any automation for migrating collateral addresses and adding a final fee?

Do we have a script that could take the addresses and final fees as arguments and add them? Or do you think that's overkill? I only ask because we previously had an issue where we forgot to add initial final fees on polygon, causing free DVM requests to be possible.

mrice32 (commented):
Looks great! Just a few minor nits and comments

packages/common/src/hardhat/tasks/identifiers.ts
nicholaspai: Ah yes we should do that!

packages/common/src/Constants.ts
nicholaspai: Hm maybe we should add them to `networks` and then we can just call `getAddress()` and if it errors then we can catch the error and log that the multicall contract isn't set for this network.

packages/core/networks/42161.json
nicholaspai: > Do we have a script that could take the addresses and final fees as arguments and add them?

We should add this but I'll save it for the next PR. Would be nice to be able to pipe in some list of addresses to whitelist all at once.

mrice32: Sounds good!

packages/common/src/Constants.ts
nicholaspai: Hmm actually we shouldn't really use `getAddress` within the `common` package because then we'll have a circular dependency between `contracts-node <> common`. I'll go back to hardcoding the multicall address but i'll define it like so:

```
// Updated multicall deployments can be found here: https://github.com/makerdao/multicall
const NETWORKS_WITH_MULTICALL_DEPLOYMENTS = [1, 3, 4, 5, 42];
export const getMulticallAddress = (netId: number): string => {
  if (!NETWORKS_WITH_MULTICALL_DEPLOYMENTS.includes(netId)) throw new Error("No Multicall2 deployment for netId")
  // All Multicall2 addresses are the same.
  return "0x5ba1e12693dc8f9c48aad8770482f4739beed696";
}
```

mrice32: Seems reasonable!

mrice32: Good point on the circular dep.

mrice32 (approved):
LGTM!

PR #3738, opened 2022-01-03
feat(sdk): add oo global store and writing

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/3809/add-ability-to-write-to-global-state

**Summary**

This adds a global store to the client with a typesafe writing interface. This is similar to what redux or some other global store would do, but only uses `immer` to generate the changes needed to the state object. Changes can be hooked in with a callback to update your application. 

**Details**
Because of the dependency on immer, the actual interface to write is a bit weird, it requires making all changes in a callback that finalizes at the end of the callback. This wont be exposed to end user ultimately and will be wrapped by higher level functions to compose multiple writes internally.

` store.write(writer=>writer.user.address(newAddress))`

**Testing**

- [x]  New unit tests created
- [ ]  Untested


packages/sdk/src/oracle/services/tests/erc20.e2e.ts
```diff
@@ -3,11 +3,11 @@ import assert from "assert";
 import { ethers } from "ethers";
 import { Provider } from "@ethersproject/providers";
 import { factory, Erc20 } from "../erc20";
-import { MULTICALL2_ADDRESS } from "../../constants";
 
 dotenv.config();
 
 const wethAddress = "0x7355Efc63Ae731f584380a9838292c7046c1e433";
+const multicall2Address = process.env.multicall2Address;
```
mrice32, line 10: If you're already assuming the mainnet WETH address, is there any harm in assuming the mainnet multicall address?

packages/sdk/src/oracle/store/index.ts
```diff
@@ -0,0 +1,18 @@
+import type { State } from "../types/state";
+import Write from "./write";
+import produce from "immer";
+
+export type WriteCallback = (write: Write) => void;
+export type Store = {
+  write: (cb: WriteCallback) => void;
+};
+
+export function factory(emit: (state: State, prev: State) => void = () => ({}), state: State = {}): Store {
```
mrice32, line 10:
```suggestion
export function factory(emit: (state: State, prev: State) => void = () => undefined, state: State = {}): Store {
```

packages/sdk/src/oracle/store/index.ts
```diff
@@ -0,0 +1,18 @@
...
+  return {
+    write(cb: WriteCallback) {
+      const prevState = state;
+      state = produce(state, (draft: State) => cb(new Write(draft)));
+      emit(state, prevState);
+    },
```
mrice32, line 16: Could you add some comments around what this is doing? These steps seem a little dense.

packages/sdk/src/oracle/store/index.ts
mrice32, line 10: Could you add some natspec comments explaining how this is intended to be used? emit and state are both pretty general variable names, so it's unclear what they're meant to be set to.

packages/sdk/src/oracle/store/index.ts
```diff
@@ -0,0 +1,18 @@
+import type { State } from "../types/state";
```
mrice32, line 1: I'm a little confused about what in this file is specific to the oracle? This feels like a general module that could be generic and used for _any_ store, no? Is there a reason it isn't/can't be generalized?

mrice32 (commented):
Looks great! A few comments

amateima (approved):
Overall looks very good. I don't have other comments besides Matt's mentions.

packages/sdk/src/oracle/services/tests/erc20.e2e.ts
daywiss: yea can do that

packages/sdk/src/oracle/store/index.ts
daywiss: i took a stab at generalizing this. the problem really is the "Writer" class isnt general and is structured specifically to the state defined in types/state, which describes the shape of the data for the application.  I've made the interface for writing expose this Writer class to the caller which wraps the state to give you a safer way to do mutations.  So thats where the sticking point is for generalizing it currently.

packages/sdk/src/oracle/store/index.ts
daywiss: yes, ive refactored this into a class, and added comments

packages/sdk/src/oracle/store/index.ts
daywiss: added some comments around what this does

packages/sdk/src/oracle/store/index.ts
daywiss: i was able to break out the generic store as a separate client, and wrap this with an app specific store

packages/sdk/src/oracle/store/write.ts
```diff
@@ -0,0 +1,107 @@
...
+  }
+}
+export class Erc20 {
+  constructor(private state: Partial<state.Erc20>) {}
+  props(data: state.Erc20["props"]): void {
+    this.state.props = data;
+  }
+  balance(account: string, amount: ethers.BigNumber): void {
+    if (!this.state.balances) this.state.balances = {};
+    new Balances(this.state.balances).set(account, amount);
```
mrice32, line 42: nit: what's the point of this ephemeral balances class that's only used to set a value in a mapping? Why not store or return the balances class?

packages/sdk/src/oracle/store/write.ts
```diff
@@ -0,0 +1,107 @@
...
+  constructor(private state: Partial<state.Chain>) {}
+  erc20s(address: string): Erc20 {
+    if (!this.state?.erc20s) this.state.erc20s = {};
+    if (!this.state.erc20s?.[address]) this.state.erc20s[address] = {};
+    return new Erc20(this.state.erc20s[address]);
+  }
+  optimisticOracle(): OptimisticOracle {
+    if (!this.state?.optimisticOracle) this.state.optimisticOracle = {};
+    return new OptimisticOracle(this.state.optimisticOracle);
+  }
```
mrice32, line 79: There's a convention in these classes to use the name of the relevant state as the getter/setter name without "get" or "set". This is fine, but in some classes, these are getters and in others they are setters. Do you think this might be confusing for a user? What happens if you want a getter and a setter for the same value in the same class?

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/store/store.ts
```diff
@@ -0,0 +1,35 @@
...
+   *
+   * @param {WriteCallback} cb
+   */
+  write(cb: WriteCallback<S>) {
+    const prevState = this.state;
+
+    // immer's produce method, takes an object, and passes a draft of that object to the callback. Any changes to the draft
+    // will be returned as the new state at the end of the callback. Changes to this object are optimized to be used in state
+    // for React or other front-end state managers.
+    this.state = produce(this.state, cb);
```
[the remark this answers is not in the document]
daywiss: this `produce` function is actually doing a lot under the hood. Essentially the benefit is that it creates a new object from this.state, which is passed into cb. You can modify this new object in your callback, do whatever you want with it, and `immer's` produce function will return the new state with the path to all nested state changes as new objects. Its probably best described in their docs: https://immerjs.github.io/immer/.

The reason why we want this is 2 reasons:
1. updating nested state in an immutable way is really annoying, especially when nesting is really deep in our case
```
// example from immer docs
nextState[1] = {
    // replace element 1...
    ...nextState[1], // with a shallow clone of element 1
    done: true // ...combined with the desired update
}
```
2. In order for react to detect changes, its not good enough that you change just the leaves of the state tree, you need to show that all parents to that leaf also changed, which does not happen when you modify a nested object directly.
3. immer also has the ability to generate json diffs which are like events for applying changes or undoing changes through this produce function. Its really cool and may be useful later.

packages/sdk/src/oracle/store/write.ts
daywiss: The balance class is used for setting both balances and allowances. its also reused for allowances.

this balance class may be extended eventually to not just set an amount, but add or remove. In that case we would probably change this to return the balances class, so you could have like `balance.add(address,amount)`, `balance.sub(address,amount)`. Right now theres just a single call to `set` so i figured it could save lines by just doing the set in the function without having to return the class

packages/sdk/src/oracle/store/write.ts
daywiss: the point of this class is that these are all setters, so using the term set to prefix calls would be redundant.  The times where functions return a class, are only returning classes that are other setters. You dont have to know you are getting anything when interacting with the writer.

For example, even though this chain of calls technically 'gets' other classes, the user doesnt need to know that and its more like drilling down into a nested state to finally set something.

`store.write((write) => write.chains(1).erc20s("test").allowance("user1", "oracle", BigNumber.from(1)))` 

if we could do nest into data with `?.` syntax for setting, we could probably forgo these classes

packages/sdk/src/oracle/store/store.ts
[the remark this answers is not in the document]
mrice32: Makes sense. Great point about the `this.state` ref needing to change for react to "notice".

packages/sdk/src/oracle/store/write.ts
mrice32: This makes more sense when you describe it that way. This isn't meant to be a place where you can access state at all.

mrice32 (approved):

PR #3739, opened 2022-01-03
feat(sdk): add basic state reader to oo client

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/3810/add-ability-to-read-from-global-state

**Summary**
This adds a basic way to simplify calls to read global state by creating a type safe class with some common calls. Calls are automatically scoped based on the users current input, ie which request they are currently looking at. This means though that reads for some calls wont work unless the input is set ahead of time.  This is mainly meant as an example of how to add calls as needed.

**Testing**

- [x]  New unit tests created
- [ ]  Untested


mrice32 (approved):
LGTM!

PR #3742, opened 2022-01-04
improve(optimistic-oracle-event-client): Support paginated block search config for Arbitrum Infura

nicholaspai:
**Motivation**

Event querying on Arbitrum cannot look at more than 10,000 blocks at a time. This PR defines a new environment variable `process.env.MAX_BLOCKS_PER_EVENT_SEARCH` that can be used to control the amount of blocks to search in a single event web3 query

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/ContractUtils.ts
+++ b/packages/common/src/ContractUtils.ts
@@ -4,7 +4,7 @@ import type truffleContract_ from "@truffle/contract";
 import type { BN } from "./types";
 import Web3 from "web3";
 import type { provider as Provider } from "web3-core";
-import { EventData } from "web3-eth-contract";
+import { EventData, Contract } from "web3-eth-contract";
 
 // Truffle library types aren't specified correctly. Cast and modify to correct for this.
 export interface TruffleInstance {
@@ -88,26 +88,25 @@ export type EventSearchOptions = {
   // filters all events where "myNumber" is 12 or 13.
 };
 
+export type Web3Contract = Contract;
+
 /**
- * Return all events between block range. Will paginate the event search using the `pageSize` if specified. Takes as
- * input a list of async callbacks that are used to make event requests. This means that different event types and
- * and contracts can be queried in this function.
- * @param contractEventQueryCallbacks Array of functions that take in `EventSearchOptions` and return promises that will
- * resolve to arrays of EventData.
- * Example of one such callback: (blockSearchConfig) => return this.oracle.getPastEvents("RequestPrice", blockSearchConfig)
+ * Return all events between block range. Will paginate the event search using the `pageSize` if specified.
+ * @param contract Contract to query.
+ * @param eventName Event to query.
  * @param earliestBlockToQuery First block to query.
  * @param latestBlockToQuery Latest block to query.
  * @param pageSize Number of blocks to search for in each query. Determines how many web3 requests are sent to fetch
  * data for all blocks between `earliestBlockToQuery` and `latestBlockToQuery`.
- * @return array of event data. Array size is equal to size of `contractEventQueryCallbacks`. Also returns number of
- * web3 requests sent.
+ * @return array of event data.
  */
 export async function getEventsWithPaginatedBlockSearch(
-  contractEventQueryCallbacks: ((blockSearchConfig: EventSearchOptions) => Promise<EventData[]>)[],
+  contract: Web3Contract,
+  eventName: string,
   earliestBlockToQuery: number,
   latestBlockToQuery: number,
   pageSize: number | null = null
-): Promise<{ eventData: EventData[][]; web3RequestCount: number }> {
+): Promise<{ eventData: EventData[]; web3RequestCount: number }> {
   const blockSearchConfig = {
     fromBlock: earliestBlockToQuery,
     toBlock: latestBlockToQuery,
@@ -118,9 +117,7 @@ export async function getEventsWithPaginatedBlockSearch(
   // Construct promise array of event searches to send in parallel
   const promisesToSend = [];
   while (blockSearchConfig.fromBlock <= latestBlockToQuery) {
-    for (let i = 0; i < contractEventQueryCallbacks.length; i++) {
-      promisesToSend.push(contractEventQueryCallbacks[i](blockSearchConfig));
-    }
+    promisesToSend.push(contract.getPastEvents(eventName, blockSearchConfig));
 
     // Increment block search config. If `pageSize` is undefined, there is no need to set the `toBlock` since we're
     // already increasing the `fromBlock` such that the `while` loop will exit on the next iteration.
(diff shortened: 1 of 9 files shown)


packages/common/src/TimeUtils.ts
```diff
@@ -7,7 +7,7 @@ dotenv.config();
  * @notice Return average block-time for a period.
  */
 // eslint-disable-next-line @typescript-eslint/no-unused-vars
-export async function averageBlockTimeSeconds(lookbackSeconds?: number, networkId?: number): Promise<number> {
+export async function averageBlockTimeSeconds(networkId?: number): Promise<number> {
```
mrice32, line 10: Good call. While we're at it, should we change this variable name to chainId rather than network id?

mrice32 (approved):
Looks good! If we end up using this in more than these two places we may want to consider making this some sort of function or something so we don't have to repeat the logic. Since generalizing it seems like it would be messy, I think just repeating the logic in two places is better when it's only done in two places.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -154,32 +156,56 @@ export class OptimisticOracleClient {
 
   public async update(): Promise<void> {
     // Determine earliest block to query events for based on lookback window:
+    const netId = await this.web3.eth.getChainId();
```
chrismaree, line 159: might it be better to send this in on construction?

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -154,32 +156,56 @@ export class OptimisticOracleClient {
...
+    let disputeEvents: (DisputePrice | SkinnyDisputePrice)[] = [];
+    let settleEvents: (Settle | SkinnySettle)[] = [];
+
+    while (blockSearchConfig.fromBlock <= currentBlock.number) {
+      this.logger.debug({
+        at: "OptimisticOracleClient",
+        message: "Fetching events with block search config",
+        blockSearchConfig,
+      });
+      const eventSearchResults = await Promise.all([
```
chrismaree, line 185: this is not parallel, no? would it be better to gather up all the promises and then fire them at once? the counter-argument to doing this would be that this will place a lot more load onto the PRC provider.

chrismaree (approved):
this is pretty awesome! I do think infura will solve this problem soon but having this for now is a good stop gap solution.

packages/common/src/TimeUtils.ts
nicholaspai: Sure but is that more accurate..? I've always been confused about chain ID vs network ID

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
nicholaspai: I don't love the idea of adding another constructor variable for something we'll only use once. I feel like its cleanest to infer `chainId` from the `web` object, even if that means an extra call. What we could do is store this `chainId` and only set it once

nicholaspai: Thoughts on this commit addressing your comment? https://github.com/UMAprotocol/protocol/pull/3742/commits/245bc900b8d7aeeb2c85c9e7013de602370b2f10

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
nicholaspai: Oh good call I'll parallelize it and test in prod

nicholaspai:
> Looks good! If we end up using this in more than these two places we may want to consider making this some sort of function or something so we don't have to repeat the logic. Since generalizing it seems like it would be messy, I think just repeating the logic in two places is better when it's only done in two places.

I generalized it into a common util, PTAL!

packages/common/src/TimeUtils.ts
mrice32: They're almost never different. As far as I understand, chainId protects against replay attacks, whereas network id is how nodes find each other. For applications that operate _outside_ of the consensus layer, I think chain id is the preferred identifier.

packages/common/src/ContractUtils.ts
```diff
@@ -88,6 +88,64 @@ export type EventSearchOptions = {
...
+ * Example of one such callback: (blockSearchConfig) => return this.oracle.getPastEvents("RequestPrice", blockSearchConfig)
+ * @param earliestBlockToQuery First block to query.
+ * @param latestBlockToQuery Latest block to query.
+ * @param pageSize Number of blocks to search for in each query. Determines how many web3 requests are sent to fetch
+ * data for all blocks between `earliestBlockToQuery` and `latestBlockToQuery`.
+ * @return array of event data. Array size is equal to size of `contractEventQueryCallbacks`. Also returns number of
+ * web3 requests sent.
+ */
+export async function getEventsWithPaginatedBlockSearch(
+  contractEventQueryCallbacks: ((blockSearchConfig: EventSearchOptions) => Promise<EventData[]>)[],
```
mrice32, line 106: Just OOC, why did you decide to pass in an array of events to trigger rather than having the caller just call this function `n` times within a `Promise.all`?

mrice32 (approved):
Looks good!

packages/common/src/ContractUtils.ts
nicholaspai: Oh that's a good point I hadn't thought of that, I'll refactor as this is more sustainable

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
chrismaree: your changes are better, thanks!

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
chrismaree: excellent, thanks!

packages/common/src/ContractUtils.ts
```diff
@@ -88,6 +88,56 @@ export type EventSearchOptions = {
...
+ * Return all events between block range. Will paginate the event search using the `pageSize` if specified.
+ * @param contract Contract to query.
+ * @param eventName Event to query.
+ * @param earliestBlockToQuery First block to query.
+ * @param latestBlockToQuery Latest block to query.
+ * @param pageSize Number of blocks to search for in each query. Determines how many web3 requests are sent to fetch
+ * data for all blocks between `earliestBlockToQuery` and `latestBlockToQuery`.
+ * @return array of event data.
+ */
+export async function getEventsWithPaginatedBlockSearch(
```
chrismaree, line 103: <3 thanks for moving this into a common util. this is awesome

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -154,32 +157,57 @@ export class OptimisticOracleClient {
...
-      >,
-      (this.oracle.getPastEvents("DisputePrice", { fromBlock: earliestBlockToQuery }) as unknown) as Promise<
-        DisputePrice[] | SkinnyDisputePrice[]
-      >,
-      (this.oracle.getPastEvents("Settle", { fromBlock: earliestBlockToQuery }) as unknown) as Promise<
-        Settle[] | SkinnySettle[]
-      >,
-      this.oracle.methods.getCurrentTime().call(),
+    const eventResults = await Promise.all([
+      getEventsWithPaginatedBlockSearch(
```
chrismaree, line 169: super clean with this util. good work. 

one thing I'm not a huge fan of is the type casting you have to do in all these place with the `(this.oracle as unknown) as Web3Contract`. that's kinda painful. is there a way we dont have to do this? `this.oracle` is 
`export type OptimisticOracleContract = SkinnyOptimisticOracleWeb3 | OptimisticOracleWeb3`. is this not of type `Web3Contract`?

nicholaspai: Nah its not quite a type of Web3Contract

PR #3745, opened 2022-01-04
feat: make hub respect bot retry configs

mrice32:
**Motivation**

The hub doesn't have any capacity to retry its requests. This causes it to sometimes be flaky.

**Summary**

This allows the hub to pull in the spoke's retry config to allow it to construct a retry provider.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -33,6 +33,7 @@ const { Storage } = require("@google-cloud/storage"); // Used to get global conf
 const storage = new Storage();
 const { Datastore } = require("@google-cloud/datastore"); // Used to read/write the last block number the monitor used.
 const datastore = new Datastore();
+const { createBasicProvider } = require("@uma/common");
 
 // Web3 instance to get current block numbers of polling loops.
 const Web3 = require("web3");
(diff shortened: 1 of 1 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -397,15 +398,23 @@ async function _getLastQueriedBlockNumber(configIdentifier, chainId, logger) {
   );
 }
 
+function _getWeb3AndUrlForBot(botConfig) {
+  const retryConfig = botConfig?.environmentVariables?.NODE_RETRY_CONFIG;
+  if (retryConfig) {
+    return [new Web3(createBasicProvider(retryConfig)), retryConfig[0].url];
```
chrismaree, line 404: hmm. you are taking the 0th element from this array. does this not mean that you are only creating the web3 instance from the first provider and therefore are not creating a re-try provider, using all of the provided node URLs?

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -397,15 +398,23 @@ async function _getLastQueriedBlockNumber(configIdentifier, chainId, logger) {
...
-  const web3 = new Web3(overrideNodeUrl || customNodeUrl);
+async function _getLatestBlockNumber(web3) {
   return await web3.eth.getBlockNumber();
 }
 
-async function _getChainId(overrideNodeUrl) {
-  const web3 = new Web3(overrideNodeUrl || customNodeUrl);
+async function _getChainId(web3) {
   return await web3.eth.getChainId();
 }
```
chrismaree, line 419: this is better

packages/serverless-orchestration/src/ServerlessHub.js
mrice32: No, previously we were using the url as the key to cache information. To keep that structure, I just use the first url as the "key". The web3 being used to make the calls is the first element of the return.

Does that make sense?

mrice32: (Happy to go with a different function or key structure if you have a different preference!!)

chrismaree: O I see, ye I missed that. thanks for explaining, makes sence!

chrismaree (approved):
lgtm! this should hopefully remove some issues we've seen with flaky timeouts in the hub.

PR #3746, opened 2022-01-05
feat(sdk): expose basic oo client interface

daywiss:

**Motivation**

https://app.shortcut.com/uma-project/story/3813/wire-up-and-publicly-expose-oo-client-functions

**Summary**

Exposes basic client to frontend, allowing user to set the current request as well as the currently logged in user.  Next PR should add state updates based on collateral balances and allowances.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/README.md
+++ b/packages/sdk/src/oracle/README.md
@@ -1,0 +1,77 @@
+# Oracle Client
+
+Interact with UMA's Optimistic Oracle using an eventful client designed for a frontend dapp with global state management.
+
+# Quick Start
+
+Currently theres a single factory which creates the client and allows you to subscribe to state changes.
+
+```js
+import {oracle} from '@uma/sdk'
+
+// create your configuration object with known addresses
+type Config = oracle.types.state.Config
+
+// creating a config for mainnet usage add any additional chains you want
+export const config = {
+  chains: {
+    1:{
+      chainId: 1,
+      multicall2Address: process.env.multicall2Address,
+      optimisticOracleAddress: process.env.optimisticOracleAddress,
+      providerUrl: process.env.CUSTOM_NODE_URL,
+    },
+  },
+};
+
+let state:oracle.types.state.State = {}
+function changeHandler(nextState:oracle.types.state.State, prevState:oracle.types.state.State){
+ // dispatch to state manager. Recommended to have its own slice of state.
+ state = nextState
+}
+
+const client = Factory(config, changeHandler);
+// client contains the following classes and calls:
+//{
+//  store: { read, write, get} // get and set all parts of state
+//  update: {// all update calls} // update parts of state on chain
+//  setUser: // set the currently logged in user
+//  setActiveRequest: // set the request page the user is on
+//}
+
+const requester = "0xb8b3583f143b3a4c2aa052828d8809b0818a16e9",
+const identifier = "0x554D415553440000000000000000000000000000000000000000000000000000"
+const timestamp = 1638453600
+const ancillaryData = "0x747761704C656E6774683A33363030",
+
+const account = "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D";
+const chainId = ChainId.MAINNET;
+
+// currently the order of calls is important
+client.setActiveRequest(requester,identifier,timestamp,ancillaryData)
+client.setUser(account,chainId)
+
+
+// update parts of state from on chain. It requires that data is first set. It will throw errors if required data is unavailable.
+// inspect client.ts for all update calls.
+// updates state of current request on chain
+await client.update.request()
+// updates default liveness from oracle
+await client.update.oracle()
+// update everything, only run if both user and request has been set
+await client.update.all()
+
+// global state will be updated with the needed information
+// you can also manually fetch from store using array paths, but this isnt recommended.
+// recommended way
+let userAddress = state.user?.address
+// you can also get it from the store
+userAddress = client.store.read().userAddress()
+
+// fetched from contract
+let fullRequest = state.chains?.[chainId]?.optimisticOracle?.requests?.[requestId]
+// or get from store
+fullRequest = client.store.read().request()
+
+// inspect types/state.ts for shape of state data.
+```
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -1,0 +1,95 @@
+import Store, { Emit } from "./store";
+import type { state } from "./types";
+import { Inputs } from "./types/state";
+import { Signer } from "./types/ethers";
+
+export class Client {
+  public readonly update: Update;
+  constructor(public readonly store: Store) {
+    this.update = new Update(store);
+  }
+  setUser(address: string, chainId: number, signer: Signer): void {
+    this.store.write((write) => write.user().set({ address, chainId, signer }));
+  }
+  setActiveRequest(params: Inputs["request"]): void {
+    const { requester, identifier, timestamp, ancillaryData, chainId } = params;
+    this.store.write((write) => write.inputs().request(requester, identifier, timestamp, ancillaryData, chainId));
+  }
+}
+
+export class Update {
+  private read: Store["read"];
+  private write: Store["write"];
+  constructor(private store: Store) {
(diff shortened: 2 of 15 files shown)


packages/sdk/src/oracle/client.ts
```diff
@@ -0,0 +1,32 @@
+import Store, { Emit } from "./store";
+import type { state } from "./types";
+import { Inputs } from "./types/state";
+import { Signer } from "./types/ethers";
+
+export class Client {
```
daywiss, line 6: this is the client exposed to the frontend, currently only 2 functions, one to set your active wallet, the other to set the active request, ie which oracle request we are on

packages/sdk/src/oracle/client.ts
```diff
@@ -0,0 +1,91 @@
...
+  setUser(address: string, chainId: number, signer: Signer): void {
+    this.store.write((write) => write.user().set({ address, chainId, signer }));
+  }
+  setActiveRequest(params: Inputs["request"]): void {
+    const { requester, identifier, timestamp, ancillaryData, chainId } = params;
+    this.store.write((write) => write.inputs().request(requester, identifier, timestamp, ancillaryData, chainId));
+  }
+}
+
+export class Update {
```
daywiss, line 20: update functions are attached to the client, this allows fetching from chain and updating store

amateima (approved):

packages/sdk/src/oracle/README.md
```diff
@@ -0,0 +1,77 @@
...
+const requester = "0xb8b3583f143b3a4c2aa052828d8809b0818a16e9",
+const identifier = "0x554D415553440000000000000000000000000000000000000000000000000000"
+const timestamp = 1638453600
+const ancillaryData = "0x747761704C656E6774683A33363030",
+
+const account = "0x9A8f92a830A5cB89a3816e3D267CB7791c16b04D";
+const chainId = ChainId.MAINNET;
+
+// currently the order of calls is important
+client.setActiveRequest(requester,identifier,timestamp,ancillaryData)
```
mrice32, line 51: Interesting -- why are we having the client have a stateful "active request" OOC? I'm trying to think about this purely from a user's perspective (without considering what's happening under the hood).

Two alternatives that I could see being a little simpler to understand/use:
1. Having an async method on the client to get the raw request state `client.getRequest(...)`, since that's all you can really do with a particular request. This would drop the statefulness in the client, which seems like it might be a bit annoying to deal with if you reuse the same client across a dapp.
2. Similar to 1, but maybe a little closer to what's already here: if you want request to be a stateful object, you could have a `client.getRequest(...)` function that returns a request object that you can then use to interact with the request in a more dynamic way. Something like:
```ts
const request = client.getRequest(requester,identifier,timestamp,ancillaryData);

// Not exactly sure how the user should trigger updates, as that's the part that's very application/system specific.
await request.update();

// Some way to directly get the request state would be ideal.
const state = request.read().state();

await request.write().dispute(...);

...
```

mrice32 (commented):
Looks awesome! One high-level comment on the UX of this client.

daywiss:
> Interesting -- why are we having the client have a stateful "active request" OOC? I'm trying to think about this purely from a user's perspective (without considering what's happening under the hood).
> 
> Two alternatives that I could see being a little simpler to understand/use:
> 
> Having an async method on the client to get the raw request state client.getRequest(...), since that's all you can really do with a particular request. This would drop the statefulness in the client, which seems like it might be a bit annoying to deal with if you reuse the same client across a dapp.
> Similar to 1, but maybe a little closer to what's already here: if you want request to be a stateful object, you could have a client.getRequest(...) function that returns a request object that you can then use to interact with the request in a more dynamic way. Something like:

1. Im not sure i totally understand this suggestion

2. This is a really interesting UX @mrice32 (if i understand right) that i could consider, i dont really have a good reason why not to do it that way because it does present some nice properties from an implementation perspective, you can then instance as many "request" pages as you want, you are no longer really restricted to single view per client.  As far as the implementation goes, it would have to think about it some, but it might be pretty easy to do. One slightly annoying thing about this is that you cant create a single global object to interact with on your app, each time you go to a new page you have to instance a `request client` which would then need to be passed around to components, vs a single global client which can be imported as a singleton from any page and you call a function to update it. 

As far as why i did it this way, i was more building to the design, based on the elements on the request page. I saw the users input as  its own kind of stateful object, set by what the user was looking at currently, so it didnt occur to me to allow instances of this state ( since the user would only ever be at a single page and have a single wallet per window). I do find the current implementation a bit awkward though because we have to listen to changes on this object as the user navigates to other pages or signs in as other users. 

anyways not sure if that covers your questions, would be happy to discuss this more

mrice32:
> > Interesting -- why are we having the client have a stateful "active request" OOC? I'm trying to think about this purely from a user's perspective (without considering what's happening under the hood).
> > Two alternatives that I could see being a little simpler to understand/use:
> > Having an async method on the client to get the raw request state client.getRequest(...), since that's all you can really do with a particular request. This would drop the statefulness in the client, which seems like it might be a bit annoying to deal with if you reuse the same client across a dapp.
> > Similar to 1, but maybe a little closer to what's already here: if you want request to be a stateful object, you could have a client.getRequest(...) function that returns a request object that you can then use to interact with the request in a more dynamic way. Something like:
> 
> 1. Im not sure i totally understand this suggestion

This is just 2, but instead of returning a request _object_ you just return the raw request data. Not sure how well this integrates with the existing read/write architecture.

> 2. This is a really interesting UX @mrice32 (if i understand right) that i could consider, i dont really have a good reason why not to do it that way because it does present some nice properties from an implementation perspective, you can then instance as many "request" pages as you want, you are no longer really restricted to single view per client.  As far as the implementation goes, it would have to think about it some, but it might be pretty easy to do. One slightly annoying thing about this is that you cant create a single global object to interact with on your app, each time you go to a new page you have to instance a `request client` which would then need to be passed around to components, vs a single global client which can be imported as a singleton from any page and you call a function to update it.

Why can't you create a single global object? The client is that global object, right? I might be misunderstanding what you're getting at, though. As a side note: if you wanted to optimize this to ensure you don't end up creating multiple instances of the same client, you could have the client object keep an internal cache of all the request objects it's created. This has some obvious downsides, though, namely that if there's any _other_ state in those request objects, like a signer, you could potentially have different parts of the app "disagree" over what state they want in the object.

> 
> As far as why i did it this way, i was more building to the design, based on the elements on the request page. I saw the users input as its own kind of stateful object, set by what the user was looking at currently, so it didnt occur to me to allow instances of this state ( since the user would only ever be at a single page and have a single wallet per window). I do find the current implementation a bit awkward though because we have to listen to changes on this object as the user navigates to other pages or signs in as other users.
> 
> anyways not sure if that covers your questions, would be happy to discuss this more

Yep, I think that covers it. Are you in favor of making this change?

mrice32 (commented):
https://github.com/UMAprotocol/protocol/pull/3746#issuecomment-1006937203

daywiss:
> Yep, I think that covers it. Are you in favor of making this change?

I need to make sure i understand what the intended interface should be, maybe we can talk more offline.  Id also need to talk to jason on what type of interface he would prefer once i get a clearer picture.

based on what i think i understand and thinking about it some more, there are some hurdles which gives me hesitation

1. the signer can change at any time, so you still need some way to dynamically update the client with the current user, leading to the same kind of problems managing request as dynamic internal state, so moving it out doesnt really save much technically
2. the app doesnt really care about operating on a request itself, the request just sets the parameters, so creating a 'request client' is a bit too specific, its more like we want a "page" client that does things like finds the collateral needed, looks up approvals and balances, figures out whats blocking this users proposal submission, sees what transactions are currently in flight for the user. This may just be simply an issue of naming, but the client as it is now is supposed to encapsulate all this functionality
3. having to instantiate this new client on component load is technically a worse ux than having the client instantiated once on JS load (outside of react). This is because the request props for the page are parsed in react router and are only available inside a component.  This means it would require a useeffect hook a  useState hook to store the 'request client' in the component and also passing this around to other components through props or context, which is really annoying. Compare this to instantiating the client once outside of react, setting the request and using 'import' to access it from any component.

mrice32 (approved):
LGTM!

PR #3748, opened 2022-01-06
feat(across-bots): modify profitability model to only log once on unprofitable relay

chrismaree:
**Motivation**

The profitability module in its current form will send a log on each unprofitable relay. This makes this very noisy and hard to use. This PR modify this to only send one `warn` level log, with subsequent logs set to `debug`.  

**Summary**

Implementation saves a file to disk when running locally or saves to GCP data store when running in production by setting the `LOG_SILENCER_MODE` environment variable.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


packages/insured-bridge-relayer/src/LogHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+import path from "path";
+import { Datastore } from "@google-cloud/datastore";
+const datastore = new Datastore();
+
+enum LogHelperMode {
+  Gcp,
+  Local,
+}
+
+const logHelperMode = process.env.LOG_SILENCER_MODE == "gcp" ? LogHelperMode.Gcp : LogHelperMode.Local;
```
mrice32, line 11:
```suggestion
const logHelperMode = process.env.LOG_SILENCER_MODE === "gcp" ? LogHelperMode.Gcp : LogHelperMode.Local;
```

packages/insured-bridge-relayer/src/LogHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+
+const localSaveFilePath = `${path.resolve(__dirname)}/relayer-unprofitable-logs`;
```
mrice32, line 13: What's the point of `path.resolve` here? Isn't __dirname already an absolute path?

packages/insured-bridge-relayer/src/LogHelper.ts
```diff
@@ -0,0 +1,44 @@
+import fs from "fs";
+import path from "path";
+import { Datastore } from "@google-cloud/datastore";
+const datastore = new Datastore();
```
mrice32, line 4: Do we not have to specify the project when creating this? How does it know which project's datastore to connect to?

packages/insured-bridge-relayer/src/LogHelper.ts
```diff
@@ -0,0 +1,44 @@
...
+      const logFile = fs.readFileSync(localSaveFilePath, "utf8");
+      const pastLogs = JSON.parse("[" + logFile.replace(/\r?\n|\r/g, ",").substring(0, logFile.length - 1) + "]");
+      return pastLogs.includes(depositHash);
+    } catch (error) {
+      return false;
+    }
+  } else if (logHelperMode == LogHelperMode.Gcp) {
+    const key = datastore.key(["RelayerUnprofitableLogs", depositHash]);
+    const [dataField] = await datastore.get(key);
+    return dataField != undefined;
```
mrice32, line 42: nit: prefer `!==` since `!=` can be deceptive. If you're intending to check for null _or_ undefined, something like `!!dataField` might be better?

mrice32 (approved):
LGTM! Just a few minor nits

packages/insured-bridge-relayer/src/LogHelper.ts
chrismaree: it's based off the provided GCP service account. as this will be running in prod it will access the main GCP service account. the database name is defined in the `datastore.key` call later on, with the DB called `RelayerUnprofitableLogs`

PR #3749, opened 2022-01-06
feat(across-bots): Improve log quality in serverless

chrismaree:
**Motivation**

Serverless logs are somewhat hard to read now that we've got one unified relayer over all networks for across. See the sample logs below that show an example of these hard to read logs:

image

**Summary**

This PR aims to fix this by doing two things:
1) Serverless redacted logs now include the `at` filed, as well as the previous `message` field.
2) the across bots were modified such that the `relayer.ts` now includes the chain name the relayer is connected to in the `at` field to make these logs more readable. 

Note that this wont impact almost all slack logs as slack logs are not produced from this part of the code. For example instant, speed up and slow relay logs are produced in the `MulticallBundler. See screenshot below:
image

An additional thing this PR adds is the `configName` to the serverless hub execution. Using this, you can enrich GCP logs to with a flag to show you what config produced the execution output. this looks like this:
image

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/insured-bridge-relayer/src/Relayer.ts
+++ b/packages/insured-bridge-relayer/src/Relayer.ts
@@ -18,7 +18,7 @@ import {
 import { getTokenBalance } from "./RelayerHelpers";
 import { ProfitabilityCalculator } from "./ProfitabilityCalculator";
 
-import type { BN, TransactionType, ExecutedTransaction } from "@uma/common";
+import type { BN, TransactionType } from "@uma/common";
 import { MulticallBundler } from "./MulticallBundler";
 
 // Stores state of Relay (i.e. Pending, Uninitialized, Finalized) and linked L2 deposit parameters.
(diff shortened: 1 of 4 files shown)


packages/insured-bridge-relayer/src/Relayer.ts
```diff
@@ -66,16 +65,18 @@ export class Relayer {
...
     readonly l2LookbackWindow: number,
     readonly multicallBundler: MulticallBundler
-  ) {}
+  ) {
+    this.chainName = PublicNetworks[this.l2Client.chainId]?.name;
+  }
 
   async checkForPendingDepositsAndRelay(): Promise<void> {
-    this.logger.debug({ at: "AcrossRelayer#Relayer", message: "Checking for pending deposits and relaying" });
+    this.logger.debug({ at: `Relayer#${this.chainName}`, message: "Checking for pending deposits and relaying" });
```
mrice32, line 73: I think this is 🆗 , but as a minor note, it seems _slightly_ inconsistent with the fact that our relayer logs normally identify deposits by _chain id_ rather than chain name. I'm assuming you wanted to do the name here because it's the title of the log?

mrice32 (approved):
LGTM!

packages/insured-bridge-relayer/src/Relayer.ts
nicholaspai: I think chain ID is equally useful and in the sake of consistency we can use it over name

nicholaspai (approved):

PR #3751, opened 2022-01-06
feat(sdk): add preview flags for proposing for oo

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3781/add-approval-checking-logic

**Summary**

Adds a way to flag important errors the user has when trying to propose. This takes the shape as an array of booleans, which are keyed by an enum type which is self explanatory.  This way you can check on the frontend something like 

`if(flags[ProposalFlags.InvalidContractState]) ...`

**Testing**

- [x]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/utils.ts
+++ b/packages/sdk/src/oracle/utils.ts
@@ -1,0 +1,39 @@
+import { State } from "./types/state";
+import { Read } from "./store";
+
+export enum RequestState {
+  Invalid = 0, // Never requested.
+  Requested, // Requested, no other actions taken.
+  Proposed, // Proposed, but not expired or disputed yet.
+  Expired, // Proposed, not disputed, past liveness.
+  Disputed, // Disputed, but no DVM price returned yet.
+  Resolved, // Disputed and DVM price is available.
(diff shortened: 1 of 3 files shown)


packages/sdk/src/oracle/utils.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+export enum ProposalFlags {
+  WrongChain = 0,
+  InvalidContractState,
+  InsufficientBalance,
+  InsufficientApproval,
+  TransactionInProgress,
+}
+export function previewProposal(state: State) {
```
mrice32, line 21: Should we also simulate the txn and add an error to signal that the txn failed (for an unknown reason)?

mrice32 (approved):
LGTM!

amateima (approved):

packages/sdk/src/oracle/utils.ts
daywiss: this is really just for showing changes in the proposal form, so a failing transaction will probably be shown elsewhere. i may generalize this in a later pr, but not exactly sure yet how i want to deal with transactions.

PR #3752, opened 2022-01-06
fix(api): allow both http and ws providers

amateima:
**Motivation**

Quick PR for allowing the serverless api to use both http and ws url providers for ethers

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/api/src/libs/utils.ts
+++ b/packages/api/src/libs/utils.ts
@@ -1,7 +1,7 @@
 import lodash from "lodash";
 import { Obj } from "../types";
 import * as uma from "@uma/sdk";
-import { utils, BigNumber, Contract } from "ethers";
+import { utils, BigNumber, Contract, ethers } from "ethers";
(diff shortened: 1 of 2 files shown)


packages/api/src/libs/utils.ts
```diff
@@ -205,6 +205,13 @@ export function getWeb3(url: string, options: Obj = {}) {
   throw new Error("Web3 provider URLs not supported");
 }
 
+export function getEthers(url: string) {
+  if (url.startsWith("ws")) return new ethers.providers.WebSocketProvider(url);
+  if (url.startsWith("http")) return new ethers.providers.JsonRpcProvider(url);
```
daywiss, line 210: Might also want to check https

amateima: It's covered by `startsWith("http")` I think, right?

PR #3753, opened 2022-01-07
feat(scripts): add script to seed new address whitelist

chrismaree:
**Motivation**

When deploying to new chains it is painful to setup the new address whitelist. this PR helps this situation by providing a script to seed new address whitelists on new target chains.

The fetches the existing whitelist addresses on L1 (filtering out removed tokens) and then finds the associated L2 token addresses. for now the script works on arbitrum, optimism and boba.
for optimism and boba the scrip pulls addresses from the L2StandardBridge , looking for DepositFinalized events for the associated L1 token. On optimism and Arbitrum I also found these token lists ([arbitrum](https://bridge.arbitrum.io/token-list-42161.json) and [optimism](https://static.optimism.io/optimism.tokenlist.json)) from the [uniswap source code](https://github.com/Uniswap/interface/blob/main/src/constants/lists.ts) that includes some other tokens. using a combination of contract events and inclusion in these lists I think I’ve hit about 95% of all the tokens that are: a) in our whitelist and b) on the associated l2. you can see the outputs below:

Arbitrum:
image

Optimism:
image

Boba:
image

**Testing**

- [ ]  New unit tests created
- [X]  Untested


nicholaspai (commented):
High level comment: do you think this script is better placed in `scripts` or in the `common/hardhat/tasks` folder? The latter would be consistent with the `migrate-identifiers` script that I've used to do a similar logic in copying identifiers from one contract to another

packages/scripts/src/multi-chain/BatchCollateralWhitelist.js
```diff
@@ -0,0 +1,75 @@
...
+const argv = require("minimist")(process.argv.slice(), { number: ["l1ChainId", "l2ChainId"] });
+
+async function run() {
+  const { l1ChainId, l2ChainId } = argv;
+
+  const l1Web3 = getWeb3ByChainId(l1ChainId);
+  const l2Web3 = getWeb3ByChainId(l2ChainId);
+  const l1Account = (await l1Web3.eth.getAccounts())[0];
+
+  console.log(`Running OVM Batch Collateral whitelister from ${l1ChainId}->${l2ChainId} with`);
```
nicholaspai, line 22: I wouldn't name this just OVM batch whitelister, since it should work on other L2s right?

packages/scripts/src/multi-chain/utils.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  if (l2ChainId == 42161) {
+    return await _findL2TokenFromTokenList(l2ChainId, l1TokenAddress);
+  }
+}
+
+async function _findL2TokenFromTokenList(l2ChainId, l1TokenAddress) {
+  if (l2ChainId == 10) {
+    const response = await fetch("https://static.optimism.io/optimism.tokenlist.json");
```
nicholaspai, line 62: if l2ChainId !== 10 or 42161 where will it error?

packages/scripts/src/multi-chain/BatchCollateralWhitelist.js
chrismaree: ye, that's a mistake. when I started it was OVM spesific but I was able to generalize it.

packages/scripts/src/multi-chain/utils.js
chrismaree: if the chain id is neither of these then it returns a blank array from the token lists. for example boba's list is built up exclusively from the `_findL2TokenForOvmChain` method, not using any lists, as their is no token list like this for boba.

chrismaree:
> High level comment: do you think this script is better placed in `scripts` or in the `common/hardhat/tasks` folder? The latter would be consistent with the `migrate-identifiers` script that I've used to do a similar logic in copying identifiers from one contract to another

ye, I was not exactly sure where to put it but you are right that making this a hardhat task is better! I've refactored this to be a hard hat task, in ts, like the other ones and updated the documentation in `cross-chain-oracle` to reflect how to use this task.

packages/common/src/hardhat/tasks/collateralWhitelist.ts
```diff
@@ -0,0 +1,238 @@
...
+    );
+    assert(
+      process.env[`NODE_URL_${taskArguments.l2chainid}`],
+      `Must set NODE_URL_${taskArguments.l2chainid} in the environment`
+    );
+
+    const l1ChainId = Number(taskArguments.l1chainid);
+    const l2ChainId = Number(taskArguments.l2chainid);
+
+    const l1Web3 = new Web3(process.env[`NODE_URL_${l1ChainId}`] as string);
```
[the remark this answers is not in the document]
nicholaspai: Since you're checking NODE_URL_CHAIN_ID you're already set up to use that well

packages/common/src/hardhat/tasks/collateralWhitelist.ts
```diff
@@ -0,0 +1,238 @@
...
+      .filter((tokenList: any) => tokenList.l2TokenAddress !== ZERO_ADDRESS);
+
+    console.log(`Found the following ${combineSet.length} L1->L2 mapping and the associated final fees`);
+    console.table(combineSet);
+
+    console.log("Removing any tokens that are already on the L2 whitelist...");
+    const l2TokenWhitelistContract = await deployments.get("AddressWhitelist");
+    const l2TokenWhitelist = new l2Web3.eth.Contract(l2TokenWhitelistContract.abi, l2TokenWhitelistContract.address);
+    const currentBlock = await l2Web3.eth.getBlockNumber();
+    const blockLookBack = 99999; // We need to use paginated query on L2 as some L2s limit how far you can look back
```
nicholaspai, line 64: I like to use 99900 to be safer

packages/common/src/hardhat/tasks/collateralWhitelist.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    let nonce = await l2Web3.eth.getTransactionCount(deployer);
+    for (let index = 0; index < filteredCombinedSet.length; index++) {
+      console.log(
+        `Whitelisting ${filteredCombinedSet[index].symbol} at ${filteredCombinedSet[index].l1TokenAddress} with fee ${filteredCombinedSet[index].finalFee}...`
+      );
+      const whitelistTx = await l2TokenWhitelist.methods
+        .addToWhitelist(filteredCombinedSet[index].l2TokenAddress)
+        .send({ from: deployer, nonce });
+      nonce++;
```
nicholaspai, line 110: Why are you setting nonce? Can't you just use `.send({ from: })`?

nicholaspai (commented):
Thanks for making this a task!

packages/common/src/hardhat/tasks/collateralWhitelist.ts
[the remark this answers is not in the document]
chrismaree: I originally did do this but I had issues with the web3 instance being returned. For some reason, the web3 instance from the `getWeb3(l2ChainId)` could not accept sending of transactions. I would get the following error: `The method eth_sendTransaction does not exist/is not available`. I've seen this before; it happens when you try send a raw transaction. for some or other reason the web3 instances instantiated like this was sending txs like this. 

The solution to this was to use `web3` from the `hre` and create a new web3 instance for l1web3. I could use `getWeb3(l1ChainId)` but it's functionally equivalent to what I've got here without needed an extra import. The way the code is here is more similar to the other hardhat tasks as well.

packages/common/src/hardhat/tasks/collateralWhitelist.ts
chrismaree: for example I will often get this:
image

nicholaspai (approved):

mrice32 (commented):
Looks good! Just one comment on the env/getWeb3 usage.

PR #3758, opened 2022-01-10
feat(cross-chain-oracle): Add verification script to cross chain oracle setup

nicholaspai:
**Motivation**

There are a lot of moving parts and contract state to set before one can confidently say that an OptimisticOracleon an L2 is set up. The goal of this script is for users to have a source of truth in determining whether the setup is complete.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/hardhat/tasks/xchainSetup.ts
+++ b/packages/common/src/hardhat/tasks/xchainSetup.ts
@@ -1,10 +1,32 @@
 import { Deployment } from "hardhat-deploy/types";
+import { interfaceName } from "../../Constants";
 import { task } from "hardhat/config";
 import { Contract } from "web3-eth-contract";
 import { CombinedHRE } from "./types";
 import Web3 from "web3";
 const { utf8ToHex, toBN } = Web3.utils;
 const assert = require("assert");
+
+// Note: Make sure that the following chain names are all lower case.
+const L2_CHAIN_NAMES: { [key: string]: number } = {
+  arbitrum: 42161,
+  optimism: 10,
+  boba: 288,
+  polygon: 137,
+};
+const getChainNameForId = (chainId: number): string => {
+  const name = Object.keys(L2_CHAIN_NAMES).find((key) => L2_CHAIN_NAMES[key.toLowerCase()] === chainId);
+  if (name === undefined) throw new Error("Cannot find chain name with ID");
+  function capitalizeFirstLetter(_string: string): string {
+    return _string.charAt(0).toUpperCase() + _string.slice(1);
+  }
+  return capitalizeFirstLetter(name);
+};
+const getChainIdForName = (chainName: string): number => {
+  const id = L2_CHAIN_NAMES[chainName.toLowerCase()];
+  if (id === undefined) throw new Error("Cannot find chain ID for name");
+  return id;
+};
 
 async function setupHub(hub: Contract, deployer: string, parentMessenger: string, childChainId: number) {
   const [owner, existingParentMessenger] = await Promise.all([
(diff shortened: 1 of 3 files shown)


nicholaspai:
@chrismaree @mrice32 @Reinis-FRP Can you guys PTAL at the setup steps and let me know if i've missed anything?

packages/common/src/hardhat/tasks/xchainSetup.ts
```diff
@@ -1,11 +1,33 @@
...
 import Web3 from "web3";
 const { utf8ToHex, toBN } = Web3.utils;
 const assert = require("assert");
 
+// Note: Make sure that the following chain names are all lower case.
+const L2_CHAIN_NAMES: { [key: string]: number } = {
+  arbitrum: 42161,
+  optimism: 10,
+  boba: 288,
+  polygon: 137,
```
mrice32, line 15: nit: why do we include polygon here if it isn't currently supported by these scripts?

The double check in the verification script (one that includes polygon and one that doesn't) is a little confusing. Should we not just drop polygon (for now) and use this list as the universal source of truth?

mrice32 (approved):
Looks great! One minor comment

packages/common/src/hardhat/tasks/xchainSetup.ts
```diff
@@ -253,21 +275,239 @@ task("setup-l2-xchain", "Configures L2 cross chain smart contracts").setAction(a
...
+    console.groupEnd();
+
+    console.group(`${l2ChainName}_ChildMessenger`);
+    const childMessengerContract = new l2Web3.eth.Contract(childMessenger.abi, childMessenger.address);
+    const [childMessengerOracleSpoke, childMessengerParentMessenger] = await Promise.all([
+      childMessengerContract.methods.oracleSpoke().call(),
+      childMessengerContract.methods.parentMessenger().call(),
+    ]);
+    console.log(`- Set oracleSpoke: ${childMessengerOracleSpoke === oracleSpoke.address ? "✅" : "❌"}`);
+    console.log(`- Set parentMessenger: ${childMessengerParentMessenger === parentMessenger.address ? "✅" : "❌"}`);
```
mrice32, line 458: This name is super confusing, but I really don't have anything better in mind 😓

mrice32:
> @chrismaree @mrice32 @Reinis-FRP Can you guys PTAL at the setup steps and let me know if i've missed anything?

@nicholaspai I couldn't think of any fields/contracts missed, but there's a lot, so I definitely can't say for sure :)

packages/common/src/hardhat/tasks/xchainSetup.ts
chrismaree: as unrelated nit I think this should be placed in the constants file somewhere. we have this mapping in a few places but having a canonical `chainName->chainId` mapping would be good.

packages/common/src/hardhat/tasks/xchainSetup.ts
```diff
@@ -1,11 +1,33 @@
...
 const assert = require("assert");
 
+// Note: Make sure that the following chain names are all lower case.
+const L2_CHAIN_NAMES: { [key: string]: number } = {
+  arbitrum: 42161,
+  optimism: 10,
+  boba: 288,
+  polygon: 137,
+};
+const getChainNameForId = (chainId: number): string => {
```
chrismaree, line 17: you need to merge master. this script changed somewhat with the OP deployment. this function will be useful in re-factoring the `setupOvmBasedL1Chain` function that is currently in master. it does the exact same thing with `capitalizeFirstLetter`

packages/common/src/hardhat/tasks/xchainSetup.ts
```diff
@@ -253,21 +275,239 @@ task("setup-l2-xchain", "Configures L2 cross chain smart contracts").setAction(a
...
+      l2IdentifierWhitelist,
+      l2AddressWhitelist,
+      l2OptimisticOracle,
+      l2Finder,
+    ] = await Promise.all([
+      deployments.get("GovernorHub"),
+      deployments.get("OracleHub"),
+      deployments.get("Governor"),
+      deployments.get(`${l2ChainName}_ParentMessenger`),
+      companionNetworks[l2ChainName.toLowerCase()].deployments.get(`${l2ChainName}_ChildMessenger`),
```
chrismaree, line 325: might it be useful to define a `companionNetworkGet = companionNetworks[l2ChainName.toLowerCase()].deployments.get` to make this a bit cleaner?

chrismaree (approved):
this is pretty awesome! I really like how it makes a checklist of all things you need to have set. I'm pretty sure you hit all the details that need to be validated.

packages/common/src/hardhat/tasks/xchainSetup.ts
nicholaspai: - removed polygon and the duplicate assert that the chain name is in the L2_PUBLIC_KEYS array
- I ended up using the `PublicNetworks` file already in `common` instead of redefining this chainName --> chain ID mapping

packages/common/src/hardhat/tasks/xchainSetup.ts
nicholaspai: yeah we use this exact name in the contracts too so at least its consistent

packages/common/src/hardhat/tasks/xchainSetup.ts
nicholaspai: Nice done

packages/common/src/hardhat/tasks/xchainSetup.ts
nicholaspai: will do

PR #3763, opened 2022-01-12
 (OO-monitor): Add UTF-8 fallback decoding to OO bot

chrismaree:
**Motivation**

Sometimes the OO monitor will fail to decode data. in this cae we need to still see the ancillary data within the price request. this PR adds a fallback to use utf-8 decoding in this case so the logs are still usable.

sample log that could not be decoded:
image

Using `Web3.utils.hexToUtf8(ancillaryData)` this should be able to be decoded.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/monitors/src/OptimisticOracleContractMonitor.js
+++ b/packages/monitors/src/OptimisticOracleContractMonitor.js
@@ -299,7 +299,7 @@ class OptimisticOracleContractMonitor {
       try {
         // If that fails, try to return the ancillary data UTF-8 decoded.
         return "Ancillary could not be parsed. UTF-8 decoded data: " + this.web3.utils.hexToUtf8(ancillaryData);
-      } catch (error) {
+      } catch (_) {
         return `Could not parse ancillary data nor UTF-8 decode: ${ancillaryData || "0x"}`;
       }
     }
--- a/packages/monitors/src/OptimisticOracleContractMonitor.js
+++ b/packages/monitors/src/OptimisticOracleContractMonitor.js
@@ -300,7 +300,7 @@ class OptimisticOracleContractMonitor {
         // If that fails, try to return the ancillary data UTF-8 decoded.
         return "Ancillary could not be parsed. UTF-8 decoded data: " + this.web3.utils.hexToUtf8(ancillaryData);
       } catch (_) {
-        return `Could not parse ancillary data nor UTF-8 decode: ${ancillaryData || "0x"}`;
+        return "Could not parse ancillary data nor UTF-8 decode: " + ancillaryData || "0x";
       }
     }
   }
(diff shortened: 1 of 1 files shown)


packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -296,7 +296,12 @@ class OptimisticOracleContractMonitor {
       // Return the decoded ancillary data as a string. The `replace` syntax removes any escaped quotes from the string.
       return "Ancillary data: " + JSON.stringify(parseAncillaryData(ancillaryData)).replace(/"/g, "");
     } catch (_) {
-      return `Could not decode ancillary data: ${ancillaryData || "0x"}`;
+      try {
+        // If that fails, try to return the ancillary data UTF-8 decoded.
+        return "Ancillary could not be parsed. UTF-8 decoded data: " + this.web3.utils.hexToUtf8(ancillaryData);
```
smb2796, line 301: Thoughts on logging orig `ancillaryData` bytes as well as the decoded data? Would be long, but also mean that we wouldn't need to reference etherscan/other source to find this for proposals etc

chrismaree: why? are you not going to do the same decoding? if this decoding fails then it logs the original data. is that not sufficient?

smb2796: I'm saying it's useful to log the original and decoded data at all times. Say someone wanted to use this log to manually propose or dispute a price, then it'd be useful to have original data on hand in the log.

nicholaspai: i think as long as you provide the txn hash you can grab the ancillary data easily from etherscan

nicholaspai: i'm indifferent. These alerts are infrequent enough that im not concerned about log message. So if @smb2796 does use this ancillary data then we should include

mrice32: Agreed! Raw ancillary data would be nice!

mrice32 (approved):
LGTM once @smb2796's comment is resolved!

packages/monitors/src/OptimisticOracleContractMonitor.js
chrismaree: ye I dont agree. I think that that mades these logs really hard to see, read and makes the slack channel hard to use. If you want the raw data go get it from etherscan. These messages end up being so large and unwieldily that the channel is just destroyed by massive messages

image


For almost everyone except perhaps one person it is better to let the bot decode it and if you want the raw data get it yourself.

do you agree?

chrismaree: Also consider that slack has a max message length less than the max ancillary data! look at the screenshot above. in that case the message had to be redacted because ancillary data is so massive....

smb2796: Ok nbd. Fine without

PR #3764, opened 2022-01-12
feat(sdk): add generic state machine runner

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/3974/add-statemachine-runner-to-oo-client-for-handling-all-change-requests

**Summary**

This is a generic state machine implementation which will be used for triggering and tracking transactions, as well as fetching state on chain in reaction to user requests.  It has some features useful to tracking transactions, such as an interval value to prevent over querying, automatic error managament, and an optional memory space used for accumulating data across state transitions. 

**Details**
A statemachine is really useful for asyncronous tasks, like waiting for a pending transaction, or fetching several pieces of state conditionally or in series which may cause branching cases.  We can construct a simple transaction manager as an example:

1. Validate parameters to send transaction, send transaction, store hash in memory -> transition to 2, or end with error
2. check for transaction receipt with memory.hash , if found, transition to done, else stay in 2 and check on interval.

If we needed to execute other transactions after this one we could transition to another state, rather than ending. 

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/sdk/src/oracle/services/statemachine.ts
+++ b/packages/sdk/src/oracle/services/statemachine.ts
@@ -39,7 +39,7 @@ export class StateMachine<P = undefined, M extends Memory = undefined> {
     try {
       assert(this.handlers[context.state], "No handler for state: " + context.state);
       const handler = this.handlers[context.state];
-      // handlers take in params and return a new state or nothin
+      // handlers take in params and return a new state or nothing
       const state = await handler(context.params, context.memory);
 
       // update context state
--- a/packages/sdk/src/oracle/services/statemachine.ts
+++ b/packages/sdk/src/oracle/services/statemachine.ts
@@ -1,0 +1,108 @@
+import assert from "assert";
+import type { Context, Handlers, Memory, Emit, ContextProps } from "../types/statemachine";
+import uid from "lodash/uniqueId";
+
+export { Context, Handlers, Emit };
+
+/**
+ * StateMachine. This class will be used to handle all change requests by the user, including setting state which
+ * may require triggering fetching data from chain, or initiating transactions that require tracking.
+ *
+ * This class is meant to step through states of a Context object. This object can have variable
+ * parameters and memory. There are several concepts to understand before using this class:
+ *
+ * 1. Handlers - This is the state machine state handlers, it is an objected keyed by each state, with a function handler.
+ * 2. Params - These are common parameters set by the caller, passed into each handler function.
+ * 3. Memory - This is a writing space within the state machine handlers that accumulates through states.
+ * 4. Context - This is the sum of all data needed to transition through the state machine states. It includes params, memory and metadata about the runtime.
+ * 5. Context.done - This is a special property on context.done which represents that the state machine is done transitioning this context.
+ * 6. Context.state = "done" - This is a reserved state on context.state, if set to "done" its the same thing as done = true.
+ * 7. Context.state = "error" - This is a reserved state on context.state, if set to "error" it means the context is done, but also there is an context.error object.
+ * 8. Interval - This is a property on the context which specifies the maximum rate in MS this context will transition, based on the current timestamp.
+ */
+export class StateMachine<P = undefined, M extends Memory = undefined> {
+  private pending: Record<string, Context<P, M>> = {};
+  constructor(public type: string, private handlers: Handlers<P, M>, private emit: Emit<P, M>) {}
+  private remove(id: string): void {
+    delete this.pending[id];
+  }
+  private shouldStep(context: Context<P, M>, now: number): boolean {
+    if (context.updated === undefined) return true;
+    if (!context?.interval) return true;
+    return now - context.updated >= context.interval;
+  }
+  public hasPending(): boolean {
+    return Object.values(this.pending).length > 0;
+  }
+  private async step(context: Context<P, M>, now: number): Promise<Context<P, M>> {
+    assert(!context.done, "Context has ended");
+    try {
+      assert(this.handlers[context.state], "No handler for state: " + context.state);
+      const handler = this.handlers[context.state];
+      // handlers take in params and return a new state or nothin
+      const state = await handler(context.params, context.memory);
+
+      // update context state
+      if (state) context.state = state;
+      // check for done
+      if (state === "done") context.done = true;
+    } catch (err) {
+      context.error = err as Error;
+      context.state = "error";
+      context.done = true;
+    }
+    context.updated = now;
+    return context;
+  }
+  /**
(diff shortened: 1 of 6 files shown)


packages/sdk/src/oracle/services/statemachine.ts
```diff
@@ -0,0 +1,108 @@
...
+ * parameters and memory. There are several concepts to understand before using this class:
+ *
+ * 1. Handlers - This is the state machine state handlers, it is an objected keyed by each state, with a function handler.
+ * 2. Params - These are common parameters set by the caller, passed into each handler function.
+ * 3. Memory - This is a writing space within the state machine handlers that accumulates through states.
+ * 4. Context - This is the sum of all data needed to transition through the state machine states. It includes params, memory and metadata about the runtime.
+ * 5. Context.done - This is a special property on context.done which represents that the state machine is done transitioning this context.
+ * 6. Context.state = "done" - This is a reserved state on context.state, if set to "done" its the same thing as done = true.
+ * 7. Context.state = "error" - This is a reserved state on context.state, if set to "error" it means the context is done, but also there is an context.error object.
+ * 8. Interval - This is a property on the context which specifies the maximum rate in MS this context will transition, based on the current timestamp.
```
amateima, line 21: I would add a description for the emit param as well

packages/sdk/src/oracle/services/statemachine.ts
```diff
@@ -0,0 +1,108 @@
...
+ * StateMachine. This class will be used to handle all change requests by the user, including setting state which
+ * may require triggering fetching data from chain, or initiating transactions that require tracking.
+ *
+ * This class is meant to step through states of a Context object. This object can have variable
+ * parameters and memory. There are several concepts to understand before using this class:
+ *
+ * 1. Handlers - This is the state machine state handlers, it is an objected keyed by each state, with a function handler.
+ * 2. Params - These are common parameters set by the caller, passed into each handler function.
+ * 3. Memory - This is a writing space within the state machine handlers that accumulates through states.
+ * 4. Context - This is the sum of all data needed to transition through the state machine states. It includes params, memory and metadata about the runtime.
```
amateima, line 17: I try to build a mental model of this implementation. Is the Context smth comparable to the [Unit of Work Pattern](https://martinfowler.com/eaaCatalog/unitOfWork.html)? I see it as a db transaction

amateima (approved):

packages/sdk/src/oracle/services/statemachine.ts
daywiss: thats really cool, thanks for linking to that, it is really similar. the difference from what i can see is that unit of work drafts changes to your state for execution, and then commits them, where as the design i have in mind is each state machine is just making changes to state in real time.  These could also be called "[commands](https://martinfowler.com/bliki/CommandQuerySeparation.html)" as the broader framework follows CQRS principles.

The idea is that all requests for change that originate on the user side are commands that execute within the statemachine. This allows us an environment where when executing the commands they can operate in a space with consistent state and safely issue side effects.  This implies though that all commands must execute serially to make guarantees for state consistency.

daywiss: this is also similar to "jobs" https://cloud.google.com/kubernetes-engine/docs/how-to/jobs

packages/sdk/src/oracle/services/statemachine.ts
```diff
@@ -0,0 +1,108 @@
...
+  }
+  public hasPending(): boolean {
+    return Object.values(this.pending).length > 0;
+  }
+  private async step(context: Context<P, M>, now: number): Promise<Context<P, M>> {
+    assert(!context.done, "Context has ended");
+    try {
+      assert(this.handlers[context.state], "No handler for state: " + context.state);
+      const handler = this.handlers[context.state];
+      // handlers take in params and return a new state or nothin
```
mrice32, line 42:
```suggestion
      // handlers take in params and return a new state or nothing
```

mrice32 (approved):

packages/sdk/src/oracle/services/statemachine.ts
```diff
@@ -0,0 +1,108 @@
...
+ * 3. Memory - This is a writing space within the state machine handlers that accumulates through states.
+ * 4. Context - This is the sum of all data needed to transition through the state machine states. It includes params, memory and metadata about the runtime.
+ * 5. Context.done - This is a special property on context.done which represents that the state machine is done transitioning this context.
+ * 6. Context.state = "done" - This is a reserved state on context.state, if set to "done" its the same thing as done = true.
+ * 7. Context.state = "error" - This is a reserved state on context.state, if set to "error" it means the context is done, but also there is an context.error object.
+ * 8. Interval - This is a property on the context which specifies the maximum rate in MS this context will transition, based on the current timestamp.
+ */
+export class StateMachine<P = undefined, M extends Memory = undefined> {
+  private pending: Record<string, Context<P, M>> = {};
+  constructor(public type: string, private handlers: Handlers<P, M>, private emit: Emit<P, M>) {}
```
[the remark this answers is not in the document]
daywiss: 1. if i understand the question right, the benefit of exogenous state here is that its pure data which is useful to the user to be able to inspect, persist or even resume if needed. An example would be tracking a transaction through confirmations. You do want to know when the transaction is mined, but you may also want to know how many confirmations have happened in real time, which is not something you see waiting for a single promise to resolve. theres many ways to skin the cat here, but i thought the best way to do this was separate the stateful context (data layer) from the execution environment.

2. input is not only time, but it is difficult to see without examples.  Input also happens on the creation of a stateful context, this input is set by the user in the `create` function and `params` argument. The only reason why theres a time variable is so that you have some way to delay the machine from spamming an api like "getTransactionHash" as fast as it can waiting for something to happen. Again though its assumed once created (with user input) that the state machine has full control over writing to the object, through the "handlers" passed in the constructor.  Handlers are responsible for interacting with the world ( reading and writing externally) and then advancing state.

I have some simple examples coming in next pr which might make this make some more sense. but yea lmk if my responses are not satisfactory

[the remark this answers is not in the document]
mrice32: I don't want to get too much into the details on this because it can get really abstract and semantic, but my main confusion came from naming it a state machine, since the state machine is not contained within this class (it is not the only thing modifying the state and the state lives outside of it!).

The primary advantage of a traditional state machine is the separation of input/actions, internal state, and outputs with the main goal being that one can have strong guarantees about the path of the state. I don't think that's the goal here, though.

mrice32 (approved):
LGTM!

PR #3766, opened 2022-01-13
feat(deployments): Add xDai deploys

chrismaree:
**Motivation**

Adds xDai deployments.

### DIFF
--- a/packages/common/src/HardhatConfig.ts
+++ b/packages/common/src/HardhatConfig.ts
@@ -108,17 +108,6 @@ export function getHardhatConfig(
       rinkeby: { chainId: 4, url: getNodeUrl("rinkeby", true, 4), accounts: { mnemonic } },
       goerli: { chainId: 5, url: getNodeUrl("goerli", true, 5), accounts: { mnemonic } },
       kovan: { chainId: 42, url: getNodeUrl("kovan", true, 42), accounts: { mnemonic } },
-      arbitrum: {
-        chainId: 42161,
-        url: getNodeUrl("arbitrum", true, 42161),
-        accounts: { mnemonic },
-        companionNetworks: { mainnet: "mainnet" },
-      },
-      "arbitrum-rinkeby": {
-        chainId: 421611,
-        url: getNodeUrl("arbitrum-rinkeby", true, 421611),
-        accounts: { mnemonic },
-      },
       optimism: {
         chainId: 10,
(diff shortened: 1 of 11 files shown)


packages/common/src/HardhatConfig.ts
```diff
@@ -108,32 +108,20 @@ export function getHardhatConfig(
...
-        accounts: { mnemonic: "test test test test test test test test test test test junk" },
-        // This sets the gas price to 0 for all transactions on L2. We do this because account balances are not yet
-        // automatically initiated with an ETH balance.
-        gasPrice: 0,
-        testWhitelist: ["oracle/Finder"],
-        testBlacklist,
+      xdai: {
+        chainId: 100,
+        url: getNodeUrl("xdai", true, 100),
+        gas: 500000,
```
nicholaspai, line 121: Is it important that we keep these gas and gasPrice defaults?

nicholaspai: You'll also want to add "xdai" as a companion network to mainnet

chrismaree: ye these gas values are needed to actually send transactions on xdai. without them I was unable to send any transactions. I've added the companion network on main net.

packages/core/contracts/cross-chain-oracle/README.md
```diff
@@ -129,3 +131,65 @@ To finalize **Arbitrum** transactions see the docs [here](https://github.com/Off
...
+
+```sh
+yarn hardhat deploy --network [xdai] --tags OptimisticOracle,IdentifierWhitelist,AddressWhitelist,Store
+```
+
+8. Verify contracts:
+
+```sh
+# xDai (etherscan-verify does not work on xdai so we use hardhat verify)
+yarn hardhat --network xdai verify <EACH-DEPLOYED ADDRESS IN STEP 4> <ASSOCIATED CONSTRUCTOR PARAMS FROM 4>
```
mrice32, line 179: Does `yarn hardhat sourcify --network xdai` work? This is suggested here: https://docs.blockscout.com/for-users/smart-contract-interaction/verifying-a-smart-contract/contracts-verification-via-sourcify.

I only ask because, like `etherscan-verify`, sourcify is much less painful than the plain verify call.

mrice32 (approved):
LGTM!

PR #3769, opened 2022-01-16
fix(across-relayer): Fix various bugs in relayer and manual relay script

nicholaspai:
**Motivation**

Manual relay script wasn't using latest InsuredBridgeL1Client interface.

Unrelated, but `pruneWhitelistedRelayTokens` also can be a synchronous function.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/scripts/src/ManualRelay.js
+++ b/packages/scripts/src/ManualRelay.js
@@ -9,6 +9,7 @@
 // Optional env overrides:
 //   BRIDGE_ADMIN_ADDRESS
 //   BRIDGE_DEPOSIT_ADDRESS
+//   RATE_MODEL_ADDRESS
 //   L2_END_BLOCK_NUMBER
 //
 // Note: if you want to run with GCKMS keys, the command will look like this:
@@ -20,7 +21,6 @@ const argv = require("minimist")(process.argv.slice(), { string: ["depositId", "
 
 const { getWeb3ByChainId } = require("@uma/common");
 const { InsuredBridgeL1Client, InsuredBridgeL2Client, Logger } = require("@uma/financial-templates-lib");
-const sdk = require("@uma/sdk");
 const { getAddress } = require("@uma/contracts-node");
 
 async function main() {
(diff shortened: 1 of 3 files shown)


mrice32 (approved):
Nice catch!

packages/scripts/src/ManualRelay.js
```diff
@@ -59,7 +59,7 @@ async function main() {
     l2Web3,
     process.env.BRIDGE_DEPOSIT_ADDRESS || defaultDepositBoxMapping[Number(argv.chainId)],
     Number(argv.chainId),
-    latestL2BlockNumber - 9900,
+    latestL2BlockNumber - 99900,
```
mrice32, line 62: Is there somewhere where we should put this magic number so we don't have to remember it? Or is there at least a piece of documentation we can reference to note where it comes from?

packages/scripts/src/ManualRelay.js
```diff
@@ -50,7 +50,7 @@ async function main() {
     Logger,
     l1Web3,
     process.env.BRIDGE_ADMIN_ADDRESS || (await getAddress("BridgeAdmin", 1)),
-    sdk.across.constants.RATE_MODELS
```
mrice32, line 53: Is there a reason these constants are still there? Keeping them around seems like an easy way to create bad relays, especially since they're out of date now.

packages/scripts/src/ManualRelay.js
nicholaspai: Yeah i'll add a comment linking to the inspiration code

packages/scripts/src/ManualRelay.js
nicholaspai: @daywiss will remove this when the FE client no longer uses this. Can it be removed now?

mrice32 (approved):
LGTM!

packages/scripts/src/ManualRelay.js
daywiss: safe to remove, no longer in use on FE

daywiss (approved):

mrice32:
@nicholaspai I went ahead and merged this as I need to use it :)

PR #3770, opened 2022-01-17
improve(across-relayer): Remove hardcoded deployment data maps

nicholaspai:
**Motivation**

Its easy to forget to update this map of L1 tokens to bridge pool deployment times, and moreover its trivial to approximate these deployment times using on-chain data.

**Summary**

This adds no additional web3 requests to each bot run, it just stores extra data from the existing web3 requests.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -52,11 +52,16 @@ export interface InstantRelay {
 
 export interface BridgePoolData {
   contract: BridgePoolWeb3;
+  earliestValidDepositQuoteTime: number;
   l2Token: { [chainId: string]: string }; // chainID=>L2TokenAddress
   currentTime: number;
   relayNonce: number;
   poolCollateralDecimals: number;
   poolCollateralSymbol: string;
+}
+
+export interface BridgePoolDeploymentData {
+  [key: string]: { timestamp: number };
 }
 
 export class InsuredBridgeL1Client {
@@ -129,6 +134,14 @@ export class InsuredBridgeL1Client {
     return this.relays[l1Token][deposit.depositHash];
   }
 
+  getBridgePoolDeployData(): BridgePoolDeploymentData {
+    this._throwIfNotInitialized();
+    const deployTimestamps: BridgePoolDeploymentData = {};
+    Object.keys(this.bridgePools).map((l1Token: string) => {
+      deployTimestamps[l1Token] = { timestamp: this.bridgePools[l1Token].earliestValidDepositQuoteTime };
+    });
+    return deployTimestamps;
+  }
   getWhitelistedTokensForChainId(chainId: string): { [l1TokenAddress: string]: string } {
     this._throwIfNotInitialized();
     return this.whitelistedTokens[chainId];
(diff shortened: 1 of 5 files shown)


mrice32 (approved):
LGTM!

nicholaspai:
> LGTM!

I just realized that my current implementation is going to make one extra web3 request per WhitelistToken event, which should slow down the bot a bit since the WhitelistToken event loop is synchronous. Instead, I think we should lookup timestamps for each event in parallel upfront. This shouldn't be an issue since there won't be that many WhitelistToken events in production

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -297,7 +310,16 @@ export class InsuredBridgeL1Client {
...
     const whitelistedTokenEvents = await this.bridgeAdmin.getPastEvents("WhitelistToken", blockSearchConfig);
-    for (const whitelistedTokenEvent of whitelistedTokenEvents) {
+
+    // Lookup timestamp for each event block number in parallel instead of once per loop. We are ok fetching these
+    // upfront because there shouldn't be that many WhitelistToken events in production, so the max number of web3
+    // requests to send is anticipated to be low to lookup block numbers.
+    const getWhitelistedTokenEventBlockPromises = whitelistedTokenEvents.map((e) =>
+      this.l1Web3.eth.getBlock(e.blockNumber)
+    );
+    const whitelistedTokenBlocks = await Promise.all(getWhitelistedTokenEventBlockPromises);
```
chrismaree, line 320: put the map call within the await Promise.all?

```ts
      );
      const whitelistedTokenBlocks = await Promise.all(whitelistedTokenEvents.map((e) =>
        this.l1Web3.eth.getBlock(e.blockNumber));
```

chrismaree (approved):
this is a good cleanup. main downside is it adds at least n per whitelisted token per chain on boot. we are on 3 chains right with 5 tokens. this means 15 extra async calls per boot. I think this is fine but its a bit of a pain as we scale.

nicholaspai:
> this is a good cleanup. main downside is it adds at least n per whitelisted token per chain on boot. we are on 3 chains right with 5 tokens. this means 15 extra async calls per boot. I think this is fine but its a bit of a pain as we scale.

Yeah so we're trading off extra web3 calls here with extra dev work of updating a hardcoded mapping. I'm actually not sure which is preferable. WDYT @mrice32 @chrismaree

chrismaree:
> > this is a good cleanup. main downside is it adds at least n per whitelisted token per chain on boot. we are on 3 chains right with 5 tokens. this means 15 extra async calls per boot. I think this is fine but its a bit of a pain as we scale.
> 
> Yeah so we're trading off extra web3 calls here with extra dev work of updating a hardcoded mapping. I'm actually not sure which is preferable. WDYT @mrice32 @chrismaree

ultimately the trade off makes the bots a bit slower which is the main downside. but the way you've structured this means this is not too impactful. importantly to note is that the way you have it now also makes it easier for 3rd party bot operators to run a relay.

IMHO this is good and we should add it in as it makes the maintenance basically nothing until v2 launches. I think we ideally want to do no more bot work until v2!

this PR is good to merge from my side.

mrice32 (approved):
LGTM!

mrice32:
> > > this is a good cleanup. main downside is it adds at least n per whitelisted token per chain on boot. we are on 3 chains right with 5 tokens. this means 15 extra async calls per boot. I think this is fine but its a bit of a pain as we scale.
> > 
> > 
> > Yeah so we're trading off extra web3 calls here with extra dev work of updating a hardcoded mapping. I'm actually not sure which is preferable. WDYT @mrice32 @chrismaree
> 
> ultimately the trade off makes the bots a bit slower which is the main downside. but the way you've structured this means this is not too impactful. importantly to note is that the way you have it now also makes it easier for 3rd party bot operators to run a relay.
> 
> IMHO this is good and we should add it in as it makes the maintenance basically nothing until v2 launches. I think we ideally want to do no more bot work until v2!
> 
> this PR is good to merge from my side.

Totally agree with this tradeoff. Web3 calls are cheap relative to our time and "thinking overhead".

PR #3774, opened 2022-01-18
feat: add proposer upgrade verification script

mrice32:
**Motivation**

We want to verify that the decentralized proposal upgrade works as expected.

**Summary**

This script verifies that:
- The governance system will be functional after the upgrade.
- The governance system can revert the upgrade afterwards if any problems are discovered.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/scripts/src/admin-proposals/simulateVote.js
+++ b/packages/scripts/src/admin-proposals/simulateVote.js
@@ -33,7 +33,7 @@ const argv = require("minimist")(process.argv.slice(), {
   default: { skipExecute: false },
 });
 
-async function run() {
+async function simulateVote() {
   // Set up provider so that we can sign from special wallets. This script is designed to only run against local mainnet
   // forks.
   const { netId, web3 } = await _setupWeb3();
@@ -235,7 +235,7 @@ async function run() {
 
 function main() {
   const startTime = Date.now();
-  run()
+  simulateVote()
     .catch((err) => {
       console.error(err);
     })
@@ -249,4 +249,4 @@ if (require.main === module) {
   main();
 }
 
-module.exports = { run };
+module.exports = { simulateVote };
--- a/packages/scripts/src/admin-proposals/simulateVote.js
+++ b/packages/scripts/src/admin-proposals/simulateVote.js
@@ -223,11 +223,14 @@ async function run() {
           continue;
         }
       }
+      console.groupEnd();
     }
+    console.groupEnd();
     console.groupEnd();
   }
 
   console.log("\n😇 Success!");
+  console.groupEnd();
 }
 
 function main() {
@@ -241,4 +244,9 @@ function main() {
       console.log(`Done in ${(timeElapsed / 1000).toFixed(2)}s`);
     });
 }
-main();
+
+if (require.main === module) {
+  main();
+}
+
+module.exports = { run };
--- a/packages/scripts/src/upgrade-tests/Proposer.js
+++ b/packages/scripts/src/upgrade-tests/Proposer.js
@@ -1,0 +1,137 @@
+#!/usr/bin/env node
+
+// Run:
+//    - Run a fork in a separate terminal:
+//      HARDHAT_CHAIN_ID=1 yarn hardhat node --fork <fork url here> --no-deploy --port 9545
+//    - Run the script (assuming you're running from the root of the scripts package):
+//      CUSTOM_NODE_ULR=http://localhost:9545 HARDHAT_NETWORK=mainnet ./src/upgrade-tests/Proposer.js
+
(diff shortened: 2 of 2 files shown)


nicholaspai (approved):

packages/scripts/src/admin-proposals/simulateVote.js
```diff
@@ -241,4 +244,9 @@ function main() {
       console.log(`Done in ${(timeElapsed / 1000).toFixed(2)}s`);
     });
 }
-main();
+
+if (require.main === module) {
+  main();
+}
+
+module.exports = { run };
```
chrismaree, line 252: very nice to re-use this.

chrismaree: it might be cleaner to export it as `simulateVote` not `run` that way you dont need to do `{ run: simulateVote }` when importing and could just do `{ simulateVote }`

packages/scripts/src/upgrade-tests/Proposer.js
```diff
@@ -0,0 +1,137 @@
...
+  );
+  assert(
+    (await governor.methods.getMember(1).call()) === governorProposer,
+    "❌ Proposer not reset. Revert proposal wasn't successful."
+  );
+
+  console.log("✅ Proposal to revert upgrade to decentralized proposal system succeded.");
+}
+
+async function impersonateAccount(account) {
```
chrismaree, line 121: this is a useful method! might be nice to have it set as a common method but it is a bit confusing though as this only works in the hre context. would be nice to append it somehow to the hre object so it's always nicely useable...not something to worry about now

packages/scripts/src/upgrade-tests/Proposer.js
```diff
@@ -0,0 +1,137 @@
...
+  await governor.methods.resetMember(1, proposer.options.address).send({ from: governorOwner });
+  await governor.methods.resetMember(0, governor.options.address).send({ from: governorOwner });
+
+  // Send 10k tokens to the default account so it can make two proposals.
+  await votingToken.methods.transfer(defaultAccount, web3.utils.toWei("10000")).send({ from: largeHolder });
+
+  // Approve the proposal contract to take 10k tokens from the default wallet.
+  await votingToken.methods.approve(proposer.options.address, web3.utils.toWei("10000")).send({ from: defaultAccount });
+
+  // Create arbitrary transaction to test that the proposal system works for vanilla requests.
```
chrismaree, line 82: this is good but I feel we should also have a more UMA specific test case. please add a proposal to add an identifier to the whitelist or approve a collateral or something else as well.

packages/scripts/src/upgrade-tests/Proposer.js
```diff
@@ -0,0 +1,137 @@
...
+    (await usdc.methods.allowance(governor.options.address, defaultAccount).call()).toString() === "100",
+    "❌ Initial test transaction didn't go through!"
+  );
+
+  console.log("✅ Test proposal went through using new proposal system.");
+
+  // Make a proposal to reset the governor permissions to how they started.
+  const resetPermissionsProposal = [
+    { to: governor.options.address, data: governor.methods.resetMember(1, governorProposer).encodeABI(), value: 0 },
+    { to: governor.options.address, data: governor.methods.resetMember(0, governorOwner).encodeABI(), value: 0 },
```
chrismaree, line 103: is the ordering here important? do you have to reset 1 before 0?

mrice32: Yes. If you swapped out 0 first, you would no longer be able to swap out 1.

packages/scripts/src/admin-proposals/simulateVote.js
mrice32: Totally fair. Not sure why I kept the internal name 🤦

packages/scripts/src/upgrade-tests/Proposer.js
mrice32: Yeah, I agree. We could probably write a hardhat plugin to tack this onto the hre object.

packages/scripts/src/upgrade-tests/Proposer.js
mrice32: Will add.

mrice32: Done.

packages/scripts/src/admin-proposals/simulateVote.js
mrice32: Done.

PR #3775, opened 2022-01-18
feat: remove across contracts

mrice32:
**Motivation**

We'd like to drop all across contracts from the UMA repo.

**Summary**

There were a few build changes that needed to happen to make this fully work. The largest of which is partially disabling a generic type in the SDK because it no longer resolves correctly when typechain types are pulled from multiple sources.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/sdk/src/index.ts
```diff
@@ -46,8 +46,11 @@ export type SerializableEvent = Omit<
...
 // this convoluted type is meant to cast events to the types you need based on the contract and event name
 // example: type NewContractRegistered = GetEventType<Registry,"NewContractRegistered">;
+// TODO: the any below is a hacky solution because some typechain types fail to resolve due to
+// incompatible TypedEventFilter and TypedEvent types. This will be fixed by upgrading typechain
+// to a version where Ethers events are exported as first class types.
 export type GetEventType<ContractType extends Contract, EventName extends string> = ReturnType<
   ContractType["filters"][EventName] extends Callable ? ContractType["filters"][EventName] : never
 > extends TypedEventFilter<infer T, infer S>
-  ? TypedEvent<T & S extends Result ? T & S : never>
+  ? TypedEvent<T & S extends Result ? T & S : any>
```
daywiss, line 55: i don tthink this change is that big of a deal, would be nice to totally get rid of this with new typechain

daywiss (approved):

.circleci/config.yml
```diff
@@ -24,6 +24,7 @@ jobs:
     docker:
       - image: circleci/node:lts
     working_directory: ~/protocol
+    resource_class: large
```
mrice32, line 27: The build was running out of memory repeatedly, so I increased the docker image size.

chrismaree (approved):

PR #3778, opened 2022-01-19
feat(sdk): move set user and set request to statemachine

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/4001/move-setuser-and-setcurrentrequest-as-state-machines-in-oo-client

**Summary**

This moves `setUser` and `setActiveRequest` to the state machine framework for controlling mutations to state. This is a proof of concept for adding transactions next.  Code has been refactored to support this, with some minor modifications to statemachine and the global state object. Flags are now automatically updated in global state.  There were major difficulties working within the constraints of typescript when trying to simplify adding new types of contexts, and had to default to a less generic implementation.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/sdk/src/oracle/client.ts
```diff
@@ -1,87 +1,33 @@
...
+  private intervalStarted = false;
+  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {}
+  setUser(address: string, chainId: number, signer: Signer): string {
+    return this.sm.types.setUser.create({ address, chainId, signer });
+  }
+  setActiveRequest(params: Inputs["request"]): string {
+    return this.sm.types.setActiveRequest.create(params);
+  }
+  // runs statemachine step loop pretty fast by default.
+  startInterval(delayMs = 10) {
```
daywiss, line 20: this ensures the statemachine queue is constantly being checked. The speed doesnt really matter that much, but we do want something pretty reactive to input.  in a perfect world this would run from events rather than a timing loop.

packages/sdk/src/oracle/client.ts
```diff
@@ -1,87 +1,33 @@
...
-  setActiveRequest(params: Inputs["request"]): void {
-    const { requester, identifier, timestamp, ancillaryData, chainId } = params;
-    this.store.write((write) => write.inputs().request(requester, identifier, timestamp, ancillaryData, chainId));
-  }
-  previewProposal() {
-    return utils.previewProposal(this.store.get());
-  }
-}
-
-export class Update {
```
daywiss, line 24: this was moved to its own file

packages/sdk/src/oracle/services/index.ts
```diff
@@ -1,3 +1,3 @@
 export * as erc20 from "./erc20";
 export * as optimisticOracle from "./optimisticOracle";
-export * as statemachine from "./statemachine";
+export * as statemachines from "./statemachines";
```
daywiss, line 3: this was just renamed because there are multiple state machine contexts in this folder. having trouble coming up with good naming conventions

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,106 @@
+import { Context, Memory, ContextType } from "../../types/statemachine";
```
daywiss, line 1: this file was moved into its own folder. there are slight modifications to this class to allow multiple types of contexts in the same queue, rather than being instanced previously to a single type. This is so we can have a single queue of contexts regardless of type.

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,106 @@
...
+ * 6. Context.state = "done" - This is a reserved state on context.state, if set to "done" its the same thing as done = true.
+ * 7. Context.state = "error" - This is a reserved state on context.state, if set to "error" it means the context is done, but also there is an context.error object.
+ * 8. Interval - This is a property on the context which specifies the maximum rate in MS this context will transition, based on the current timestamp.
+ */
+export class StateMachine {
+  private pending: Context<unknown, unknown & Memory>[] = [];
+  // this needs to be updated when adding new context type
+  public types: {
+    [ContextType.setUser]: ContextManager<setUser.Params, setUser.Memory>;
+    [ContextType.setActiveRequest]: ContextManager<setActiveRequest.Params, setActiveRequest.Memory>;
```
daywiss, line 30: adding new types is a really tedious process, theres no way to build up this object at run time because you dont have access to typescript types at that time. otherwise we could just add new context types to a list and dyanmically build the structure ( like you would in js). Now we have to manually add it here and a handful of other places in what i consider infrastructure.

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,106 @@
...
+        ContextType.setUser,
+        setUser.Handlers(store),
+        setUser.initMemory,
+        this.handleCreate
+      ),
+      [ContextType.setActiveRequest]: new ContextManager<setActiveRequest.Params, setActiveRequest.Memory>(
+        ContextType.setActiveRequest,
+        setActiveRequest.Handlers(store),
+        setActiveRequest.initMemory,
+        this.handleCreate
```
daywiss, line 45: manually building up these context manager objects because we cant do it dynamically. These require types and object exported from each context type

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,106 @@
...
+   * tick. Process all pending contexts, move them to done if needed
+   *
+   * @param {} now - Specify the timestamp this tick is running on.
+   * @returns {Promise<boolean>} - Returns if there are still any pending contexts to run.
+   */
+  tick = async (now = Date.now()): Promise<boolean> => {
+    const context = this.pop();
+    if (StateMachine.shouldStep(context, now)) {
+      let next;
+      switch (context.type) {
```
daywiss, line 82: we now switch on context.type to know how to cast parameters and memory of each context type before it goes into the handler

packages/sdk/src/oracle/services/statemachines/utils.ts
```diff
@@ -0,0 +1,82 @@
+import assert from "assert";
+import uid from "lodash/uniqueId";
+
+import { Handlers, Memory, Context, Step as StepType, ContextProps, ContextType } from "../../types/statemachine";
+
+export const Step = <P = undefined, M extends Memory = undefined>(handlers: Handlers<P, M>): StepType<P, M> => async (
```
daywiss, line 6: type generic step function which will move a specific context type forward one tick

packages/sdk/src/oracle/services/statemachines/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+};
+
+export function shouldStep(context: Context<unknown, unknown & Memory> | undefined, now: number): context is Context {
+  if (!context) return false;
+  if (context.updated === undefined) return true;
+  if (!context?.interval) return true;
+  return now - context.updated >= context.interval;
+}
+
+export function create<P, M extends Memory>(
```
daywiss, line 45: generic create function to be scoped to a speicific context type.

packages/sdk/src/oracle/services/statemachines/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    type,
+    // override any part of the context, not recommended for most cases
+    ...override,
+    params,
+    memory,
+  };
+  return context;
+}
+
+export class ContextManager<P, M extends Memory> {
```
daywiss, line 67: this is complicated, but this is just a way to pull out parts of the state machine that could be generic, so we can have stronger type guarantees where we only are handling a single context type.

packages/sdk/src/oracle/services/update.ts
```diff
@@ -0,0 +1,64 @@
+import Store from "../store";
```
daywiss, line 1: this file has not changed

packages/sdk/src/oracle/utils.ts
```diff
@@ -1,39 +1,91 @@
-import { State } from "./types/state";
+import { State, RequestState, Flag } from "./types/state";
```
daywiss, line 1: these utilities have been changed to be a key:value type of flag system. Flags are now strings to make it more intuitive when logging.

packages/sdk/src/oracle/services/tests/statemachine.test.ts
```diff
@@ -1,131 +0,0 @@
-import assert from "assert";
```
daywiss, line 1: the statemachine refactors means it can no longer be unit tested unfortunately. there are new e2e tests in the client that test the statemachine implicitly as well as the individual contexts.

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+import { Handlers as GenericHandlers } from "../../types/statemachine";
+import { getFlags } from "../../utils";
+
+// required exports for state machine
+export type Params = Inputs["request"];
+export type Memory = { error?: Error };
+export function initMemory(): Memory {
+  return {};
+}
+export function Handlers(store: Store): GenericHandlers<Params, Memory> {
```
daywiss, line 13: this is the statemachine handler for the setActiveRequest call. This does not need any state transitions, so technically this is more of a proof of concept for more complex state machines. These files require certain exports to be able to be imported into the state machine: `Params`, `Memory`, `Handlers` and `initMemory`. Tried to make this as easy given the contstraints of typescript. All future context handlers should fallow this pattern.

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+  const update = new Update(store);
+  return {
+    async start(params: Params, memory: Memory) {
```
daywiss, line 16: by default all contexts start in the "start" state. so this is where you would put the initial logic

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
```diff
@@ -0,0 +1,38 @@
...
+        await update.userCollateralBalance();
+        await update.oracleAllowance();
+      } catch (err) {
+        // its ok to ignore these errors
+        memory.error = (err as unknown) as Error;
+      }
+
+      // set state flags after everything updates
+      store.write((write, state) => write.flags(getFlags(state)));
+      return "done";
```
daywiss, line 35: signal that this context is done transitioning

packages/sdk/src/oracle/services/statemachines/setUser.ts
```diff
@@ -0,0 +1,39 @@
+import { Update } from "../update";
+import Store from "../../store";
+import { Signer } from "../../types/ethers";
+import { Handlers as GenericHandlers } from "../../types/statemachine";
+import { getFlags } from "../../utils";
+
+// require exports for a new context handler
+export type Params = {
```
daywiss, line 8: this is basically identical to the setActiveRequest handler, but wiht differnet params and slightly different handler logic

mrice32 (approved):
Looks good! I think this is good to merge.

My passing comment is that it seems like the ratio of abstraction code to application code is pretty high. I think this either means that 1) there is a lot of application code planned to built on top of this or 2) the abstractions and the language don't match up very well. As I mentioned in the comments, it seems we're _avoiding_ polymorphism in a way that might be making implementation harder than it needs to be.

All that being said, I don't fully understand the vision, so I just wanted to mention these as notes to consider rather than any sort of requests for change.

LGTM!

packages/sdk/src/oracle/services/statemachines/statemachine.ts
daywiss: hmm yea i am open to any ideas here. with the current design theres a single queue of generic "contexts".  In order to create contexts and push onto queue we need to know the type to properly check parameters, and when popping off the queue and stepping through them we need to know the type.

daywiss:
> it seems we're avoiding polymorphism in a way that might be making implementation harder than it needs to be.

Thats really interesting, i wasnt explicitly trying to avoid polymorphism, its not something that i use enough to see how it would help this implementation

packages/sdk/src/oracle/client.ts
daywiss: it would just be a more complicated for some gain in performance. The contexts have this "sleep" feature which means even if you had an event driven system you still need a continuous loop to know when to "wake" the contexts to re-run, so probably not even worth persuing at this point

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
daywiss: hmm currently it is just a string, the "done" string is reserved for ending the state transitions. Otherwise you are allowed to return any string that matches the function names on the handler. This is the only string that is reserved so currently no enum.

packages/sdk/src/oracle/services/statemachines/setUser.ts
```diff
@@ -0,0 +1,38 @@
+import { Update } from "../update";
```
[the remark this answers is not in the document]
daywiss: pretty much every filename in sdk is lowercase, did not know about this

packages/sdk/src/oracle/services/statemachines/setUser.ts
```diff
@@ -0,0 +1,38 @@
...
+  signer: Signer;
+  address: string;
+};
+export type Memory = { error?: Error };
+
+export function initMemory(): Memory {
+  return {};
+}
+
+export function Handlers(store: Store): GenericHandlers<Params, Memory> {
```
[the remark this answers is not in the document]
daywiss: yea i tried to make this a class at one point but ran into a typescript problem. Typescript didnt like that i wanted to index class methods with a string, but when i return an object it seems to be ok.

packages/sdk/src/oracle/services/statemachines/setUser.ts
```diff
@@ -0,0 +1,38 @@
...
+  const update = new Update(store);
+  return {
+    async start(params: Params, memory: Memory) {
```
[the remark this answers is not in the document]
daywiss: yes ever handler needs a start, thats supposed to be enforced with the GenericHandlers as it has a `start` key as well as a `Record<string,Handler>`

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,101 @@
...
+        setActiveRequest.Handlers(store),
+        setActiveRequest.initMemory,
+        this.handleCreate
+      ),
+    };
+  }
+  private saveContext(context: Context<unknown, unknown & Memory>) {
+    this.store.write((w) => w.command(context));
+  }
+  private handleCreate = (context: Context<unknown, unknown & Memory>) => {
```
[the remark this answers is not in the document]
daywiss: yea you could be right, id maybe need to see an example of what to do though. its mainly this statemachine class and the contextmanager class which are kind of being composed to handle different parts of the logic where some is unknown and some is known

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,101 @@
...
+   * @returns {Promise<boolean>} - Returns if there are still any pending contexts to run.
+   */
+  tick = async (now = Date.now()): Promise<boolean> => {
+    const context = this.pop();
+    if (shouldStep(context, now)) {
+      let next;
+      switch (context.type) {
+        // need to update this function with the new context processor
+        case ContextType.setUser: {
+          next = await this.types[context.type].step(
```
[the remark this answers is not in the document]
daywiss: i think i kind of see what you are getting at. i can try out some things and see if it works

packages/sdk/src/oracle/services/statemachines/utils.ts
```diff
@@ -0,0 +1,82 @@
...
+    const handler = handlers[copy.state];
+    // have to copy this out because memory becomes read only for some reason?
+    const memory = Object.assign({}, copy.memory);
+    // handlers take in params and return a new state or nothin
+    const state = await handler(copy.params, memory);
+    copy.memory = memory;
+    // update context state
+    if (state) copy.state = state;
+    // check for done
+    if (state === "done") copy.done = true;
```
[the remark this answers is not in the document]
daywiss: yea i can add that

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -0,0 +1,101 @@
+import { Context, Memory, ContextType } from "../../types/statemachine";
+import type Store from "../../store";
+import { ContextManager, shouldStep } from "./utils";
+
+// context types, import new ones here
+import * as setUser from "./setUser";
```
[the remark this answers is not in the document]
daywiss: interesting, good to know. in this case though we want all the exports on these files as they are supposed to export the same interface so to speak.  although you cant create an interface that exports types

daywiss:
@mrice32 i spent some more time trying to turn handlers into classes but could not get it working. this might be something maybe we can pair on in the future, but for the sake of time im going to merge this so we can move forward

PR #3784, opened 2022-01-20
feat(sdk): add commands to for approve/propose/dispute

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/4148/enable-propose-dispute-approve-transactions-in-sdk-client

**Summary**

This adds dispute, propose, approve transaction types, as well as a clearUser for logging off.  Some minor refactors to state machine logic:

1.  support tagging a user address to a context for filtering later
2. allowing "sleep" of a context to slow down iterations within the context handler
3. step will now continue looping a context until either it sleeps or is done. there is an infininte loop check to break after X iterations with no state change

This has been integrated with the oo dapp for testing transactions

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/sdk/src/oracle/client.ts
```diff
@@ -1,23 +1,102 @@
...
+    const inputRequest = this.store.read().inputRequest();
+    const user = this.store.read().user();
+    assert(user.provider, "requires user provider");
+    assert(user.address, "requires user address");
+    assert(inputRequest.chainId, "requires active request chainId");
+    return this.sm.types.switchOrAddChain.create(
+      { chainId: inputRequest.chainId, provider: user.provider },
+      user.address
+    );
+  }
```
daywiss, line 97: all these functions create state machine calls. see the handler definitions in the files for business logic

packages/sdk/src/oracle/services/erc20.ts
```diff
@@ -10,6 +10,10 @@ export class Erc20 {
   constructor(protected provider: Provider, public readonly address: string) {
     this.contract = erc20.connect(address, provider);
   }
+  async approve(signer: Signer, spender: string, amount: BigNumberish): Promise<TransactionResponse> {
```
daywiss, line 13: adds the approve call to the erc20 class

packages/sdk/src/oracle/services/statemachines/approve.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+      // wait x seconds before running this state again
+      return context.sleep(10000);
+    },
+    async update(params: Params) {
+      const { chainId, currency, spender, account } = params;
+      await update.balance(chainId, currency, account);
+      await update.allowance(chainId, currency, account, spender);
+      return "done";
+    },
```
daywiss, line 54: approval business logic, basically sends approve tx, checks for transaction receipt in a loop, then refreshes user balance/allowance

packages/sdk/src/oracle/services/statemachines/approve.ts
```diff
@@ -0,0 +1,56 @@
...
+    async confirm(params: Params, memory: Memory, context: ContextClient) {
+      const { chainId, confirmations } = params;
+      const { hash } = memory;
+      assert(hash, "requires hash");
+      if (await update.isConfirmed(chainId, hash, confirmations)) {
+        return "update";
+      }
+      // wait x seconds before running this state again
+      return context.sleep(10000);
```
daywiss, line 47: we can sleep this state and it will resume in 10 seconds, making a basic polling function

packages/sdk/src/oracle/services/statemachines/disputePrice.ts
```diff
@@ -0,0 +1,46 @@
...
+        return "update";
+      }
+      // wait x seconds before running this state again
+      return context.sleep(10000);
+    },
+    async update(params: Params) {
+      const { chainId, currency, account } = params;
+      await update.balance(chainId, currency, account);
+      await update.request(params);
+      return "done";
```
daywiss, line 43: this follows same pattern as approve, submit tx, wait for receipt, then update relevant state from on chain

packages/sdk/src/oracle/services/statemachines/statemachine.ts
```diff
@@ -38,19 +48,49 @@ export class StateMachine {
...
+        ContextType.proposePrice,
+        proposePrice.Handlers(store),
+        proposePrice.initMemory,
+        this.handleCreate
+      ),
+      [ContextType.switchOrAddChain]: new ContextManager<switchOrAddChain.Params, switchOrAddChain.Memory>(
+        ContextType.switchOrAddChain,
+        switchOrAddChain.Handlers(store),
+        switchOrAddChain.initMemory,
+        this.handleCreate
```
daywiss, line 85: this is all just boilerplate for adding each handler

packages/sdk/src/oracle/services/statemachines/switchOrAddChain.ts
```diff
@@ -0,0 +1,46 @@
...
+      }
+    },
+    async addAndSwitch(params: Params) {
+      const { chainId, provider } = params;
+      const metadata = store.read().chainMetadata(chainId);
+      await provider.send("wallet_addEthereumChain", [metadata]);
+      await provider.send("wallet_switchEthereumChain", [
+        {
+          chainId: ethers.utils.hexValue(chainId),
+        },
```
daywiss, line 42: logic for switching chain or adding it if not avialable. uses the "metadata" which is an optional parameter in the config for each chain

packages/sdk/src/oracle/utils.ts
```diff
@@ -82,10 +83,32 @@ export function getFlags(state: State) {
...
-  flags[Flag.DisputeInProgress] = false;
-  flags[Flag.ApprovalInProgress] = false;
+  try {
+    flags[Flag.ProposalInProgress] = false;
+    flags[Flag.DisputeInProgress] = false;
+    flags[Flag.ApprovalInProgress] = false;
+    flags[Flag.ChainChangeInProgress] = false;
+
+    // get all active commands
+    const commands = read.filterCommands({ done: false, user: read.userAddress() });
```
daywiss, line 93: queries alll commands (contexts) for a user, and checks the type to see if any are in progress. sets flags accordingly

packages/sdk/src/oracle/services/statemachines/utils.ts
```diff
@@ -11,12 +19,15 @@ export const Step = <P = undefined, M extends Memory = undefined>(handlers: Hand
...
+
   try {
+    // remove sleep, assume it should be reset if needed
+    if (copy.sleep) copy.sleep = undefined;
     const handler = handlers[copy.state];
     // have to copy this out because memory becomes read only for some reason?
     const memory = Object.assign({}, copy.memory);
     // handlers take in params and return a new state or nothin
-    const state = await handler(copy.params, memory);
+    const state = await handler(copy.params, memory, new ContextClient(copy));
```
daywiss, line 30: adding this contextclient just enables updates to context from within handler. in this case it can set "sleep" for next tick, putting this at the back of the statemachine queue

packages/sdk/src/oracle/types/state.ts
```diff
@@ -17,11 +17,24 @@ export type Services = {
   chains?: Record<number, Partial<ChainServices>>;
 };
 
+export type ChainMetadata = {
```
daywiss, line 20: information required when submitting new chain data for user to add in metmask

packages/sdk/src/oracle/services/statemachines/approve.ts
```diff
@@ -0,0 +1,56 @@
...
+  return {
+    async start(params: Params, memory: Memory) {
+      const { chainId, currency, spender, amount, account, signer } = params;
+      assert(chainId === (await signer.getChainId()), "Signer on wrong chainid");
+      assert(account === (await signer.getAddress()), "Signer on wrong account");
+
+      // create service if it does not exist
+      store.write((w) => w.services(chainId).erc20s(currency));
+      const erc20 = store.read().tokenService(chainId, currency);
+      const tx = await erc20.approve(signer, spender, amount);
```
daywiss, line 35: theres a slight wrinkle here in that waiting on the users interaction here blocks the statemachine queue. This is neither good nor bad, but it implies that we cannot change state ( ie issue new transactions or change user) until the user closes or accepts this popup in metamask. I suggest keeping this behavior unless we see this is detrimental to the application. We could easily introduce a timeout here in case the user forgets to proceed.

packages/sdk/src/oracle/services/statemachines/approve.ts
mrice32: 10 seconds seems like a long time to wait for the UI to notice that your sent transaction is mined. Is there any way to make this configurable?

packages/sdk/src/oracle/services/statemachines/disputePrice.ts
```diff
@@ -0,0 +1,46 @@
...
+    async start(params: Params, memory: Memory) {
+      const { requester, identifier, timestamp, ancillaryData, chainId, signer } = params;
+      assert(chainId === (await signer.getChainId()), "Signer on wrong chainid");
+
+      const oracle = store.read().oracleService(chainId);
+      const tx = await oracle.disputePrice(signer, requester, identifier, timestamp, ancillaryData);
+      memory.hash = tx.hash;
+      return "confirm";
+    },
+    async confirm(params: Params, memory: Memory, context: ContextClient) {
```
mrice32, line 29: This is another spot where inheritance might be able to improve things. This confirm function looks identical to the other confirm function. In a class-based handler world, this could be implemented in a base class so it wouldn't need to be copy-pasted.

In fact, this file seems to be copy-pasted outside of the disputePrice call above, so I think much of this code could be eliminated in an inheritance structure.

packages/sdk/src/oracle/services/statemachines/proposePrice.ts
```diff
@@ -0,0 +1,51 @@
...
+  return {
+    async start(params: Params, memory: Memory) {
+      const { requester, identifier, timestamp, ancillaryData, chainId, signer, proposedPrice } = params;
+      assert(chainId === (await signer.getChainId()), "Signer on wrong chainid");
+      const oracle = store.read().oracleService(chainId);
+      const tx = await oracle.proposePrice(signer, requester, identifier, timestamp, ancillaryData, proposedPrice);
+      memory.hash = tx.hash;
+      return "confirm";
+    },
+    async confirm(params: Params, memory: Memory, context: ContextClient) {
```
mrice32, line 34: Same comment here.

packages/sdk/src/oracle/services/statemachines/switchOrAddChain.ts
```diff
@@ -0,0 +1,52 @@
...
+        if (err.code === -32603 || err.code === 4902) {
+          return "addAndSwitch";
+        }
+        throw err;
+      }
+    },
+    async addAndSwitch(params: Params) {
+      const { chainId, provider } = params;
+      const metadata = store.read().chainMetadata(chainId);
+      await provider.send("wallet_addEthereumChain", [
```
mrice32, line 38: What happens if the user rejects this add call? Does the next call fail gracefully? Should we detect that? Or does the error just show up in the console, so it's nothing to worry about?

packages/sdk/src/oracle/services/transaction.ts
```diff
@@ -0,0 +1,17 @@
+import { Provider, TransactionReceipt } from "../types/ethers";
+
+export class Transaction {
+  constructor(private provider: Provider) {}
```
mrice32, line 4: nit: class name seems a bit odd since it implies that this class is an abstraction for a transaction. If it were an abstraction for a transaction, I would expect it to take in the transaction hash in the constructor rather than being used for _many_ transactions.

My minor suggestion is to do one of the following:
1. Change the name.
2. Have the hash passed in in the constructor.
3. Turn this into two utility functions that take in a provider _and_ hash in each method.

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/services/statemachines/approve.ts
daywiss: currently metamask will force you to resolve the transaction request before you can change user, so it doesnt matter too much. there are some cases like "disconnect" which you could do while waiting for a tx, in this case the disconnect wont happen until you acknowledge a tx request in MM.  I think its best to show the user an error if this happens saying why they cant disconnect, ie they must accept or cancel the pending transaction.

packages/sdk/src/oracle/README.md
```diff
@@ -12,14 +12,25 @@ import {oracle} from '@uma/sdk'
...
 
-// creating a config for mainnet usage add any additional chains you want
+// creating a config for mainnet usage add any additional chains you want following the same format
 export const config = {
   chains: {
     1:{
       chainId: 1,
-      multicall2Address: process.env.multicall2Address,
-      optimisticOracleAddress: process.env.optimisticOracleAddress,
+      multicall2Address: "0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696",
```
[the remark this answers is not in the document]
daywiss: if its not syncronous it might be wierd, since theres no way to async instantiate any of this from the frontend, it currently relies on this behavior. if it is syncronous then yea we could provide some defaults here!

packages/sdk/src/oracle/README.md
```diff
@@ -12,14 +12,25 @@ import {oracle} from '@uma/sdk'
...
+        chainId: 1,
+        rpcUrls: [],
+        nativeCurrency: {
+          name: "Ether",
+          symbol: "ETH",
+          decimals: 18,
+        },
+        chainName: "Ethereum",
+        blockExplorerUrls: "https://etherscan.io"
+      }
```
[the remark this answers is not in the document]
daywiss: this was mainly tacked on as the exact data needed for requesting a user add a new provider through metamask. I agree theres a bunch of redundant fields, it might make sense to make this the default object instead.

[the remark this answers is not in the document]
daywiss: this data structure is based on this https://eips.ethereum.org/EIPS/eip-3085 so its not strictly a frontend thing

packages/sdk/src/oracle/README.md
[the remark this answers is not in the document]
daywiss: i added this change in, unfortunately it wasnt very easy. because node and frontend have different interfaces to get address (one async, one not) it forces it to either work on one or the other.  by defaulting to only using contracts-frontent, the library fails to import in node, and using contracts-node, it forces the default functions to be async.  I ended up copying just the functions i needed into the utility file to get this working on both FE and in e2e tests.

packages/sdk/src/oracle/README.md
```diff
@@ -12,14 +12,25 @@ import {oracle} from '@uma/sdk'
...
-// creating a config for mainnet usage add any additional chains you want
+// creating a config for mainnet usage add any additional chains you want following the same format
 export const config = {
   chains: {
     1:{
       chainId: 1,
-      multicall2Address: process.env.multicall2Address,
-      optimisticOracleAddress: process.env.optimisticOracleAddress,
+      multicall2Address: "0x5BA1e12693Dc8F9c48aAD8770482f4739bEeD696",
+      optimisticOracleAddress: "0xC43767F4592DF265B4a9F1a398B97fF24F38C6A6",
```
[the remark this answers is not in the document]
daywiss: yep added this in, good suggestion

packages/sdk/src/oracle/services/statemachines/disputePrice.ts
daywiss: good call, yea there is a lot of code dupe here. when i try to turn these into classes i end up getting stuck with typescript because of the way i index into class methods with a string in the statemachine. It requires more extensive changes to get that working which i am not sure how to do at the moment.  

In more sophisticated infrastructure, the way i've solved this code dupe problem in the past, rather than using inheritance, we can compose state machines, so that you have a transaction waiter, which can be called and resolved by other state machines.  This feature is really useful, so then you can treate these state machines almost as functions themselves, but the code to do this is challenging.

packages/sdk/src/oracle/services/statemachines/switchOrAddChain.ts
daywiss: The nice thing about this statemachine architecture is that you can throw errors safely in these handlers, The higher level call will catch and attach it to the context.error property, which you can then inspect to see on the command what went wrong. This gives control to the front end dev to show appropriate errors to user when needed, or ignore completely.

heres a real example of the data in state that makes it really easy to debug what happened for each users action:
image
The commands are numbeered in order and labeled by type, opening up the last approve request, we can see the user denied signing it.

daywiss:
> 10 seconds seems like a long time to wait for the UI to notice that your sent transaction is mined. Is there any way to make this configurable?

I made this configurable per chain using `checkTxIntervalSec` , so each chain can have its own speed, it will default to 5 seconds if not supplied!

packages/sdk/src/oracle/services/transaction.ts
daywiss: yea good call, it doesnt really do much other than provide some utility

daywiss: including hash and provider in same function call will require some refactoring id rather not do, but i can move this to a utility where it might make more sense, but it will have to keep the provider=>hash=>result style interface for now

PR #3790, opened 2022-01-25
feat(sdk): add typechain oo event decoder

daywiss:

**Motivation**

Add event decoding to OO events so we can get a list of requests.

**Summary**

Uses the sdk client template in src/clients to add event decoding, not to be confused with the oracle client, which is specifically a frontend client. Eventually this will be imported there to add request listing.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/sdk/src/clients/optimisticOracle/README.md
+++ b/packages/sdk/src/clients/optimisticOracle/README.md
@@ -14,12 +14,12 @@ import uma from '@uma/sdk'
 const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL)
 
 // get the contract instance
-const contractAddress:string = // assume you have an lsp address you want to connect to
+const contractAddress:string = // assume you have an optimistic oracle address you want to connect to
 const client:uma.clients.optimisticOracle.Instance = uma.clients.optimisticOracle.connect(contractAddress,provider)
 // gets all events using etheres query filter api
 const events = await client.queryFilter({})
 
-// returns EventState, defined in the lsp client
+// returns EventState, defined in the optimistic oracle client
 const state:uma.clients.optimisticOracle.EventState = uma.clients.optimisticOracle.getEventState(events)
 // see all requests given even details
 console.log(state.requests)
--- a/packages/sdk/src/clients/optimisticOracle/README.md
+++ b/packages/sdk/src/clients/optimisticOracle/README.md
@@ -1,0 +1,27 @@
+# UMA SDK Optimistic Oracle Client
+
+This is a basic wraper around the typechain Optimistic Oracle contract which adds some event decoding.
+
+## Usage
+
+Connect to a typechain contract instance and recreate state from events.
+
+```js
+import {ethers} from 'ethers'
+import uma from '@uma/sdk'
+
+// assume you have a url injected from env
+const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL)
+
+// get the contract instance
+const contractAddress:string = // assume you have an lsp address you want to connect to
+const client:uma.clients.optimisticOracle.Instance = uma.clients.optimisticOracle.connect(contractAddress,provider)
+// gets all events using etheres query filter api
+const events = await client.queryFilter({})
+
+// returns EventState, defined in the lsp client
+const state:uma.clients.optimisticOracle.EventState = uma.clients.optimisticOracle.getEventState(events)
+// see all requests given even details
+console.log(state.requests)
+
+```
--- a/packages/sdk/src/clients/optimisticOracle/client.e2e.ts
+++ b/packages/sdk/src/clients/optimisticOracle/client.e2e.ts
@@ -1,0 +1,20 @@
+require("dotenv").config();
+import assert from "assert";
+import * as Client from "./client";
+import { ethers } from "ethers";
+
+const address = "0xc43767f4592df265b4a9f1a398b97ff24f38c6a6";
+// these require integration testing, skip for ci
+describe("OptimisticOracle", function () {
+  let client: Client.Instance;
+  test("inits", function () {
+    const provider = ethers.providers.getDefaultProvider(process.env.CUSTOM_NODE_URL);
+    client = Client.connect(address, provider);
+    assert.ok(client);
+  });
+  test("getEventState", async function () {
+    const events = await client.queryFilter({});
+    const state = await Client.getEventState(events);
+    assert.ok(Object.values(state.requests || {}).length > 0);
+  });
+});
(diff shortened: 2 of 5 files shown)


packages/sdk/src/clients/optimisticOracle/README.md
```diff
@@ -0,0 +1,27 @@
...
+const provider = new ethers.providers.WebSocketProvider(env.CUSTOM_NODE_URL)
+
+// get the contract instance
+const contractAddress:string = // assume you have an lsp address you want to connect to
+const client:uma.clients.optimisticOracle.Instance = uma.clients.optimisticOracle.connect(contractAddress,provider)
+// gets all events using etheres query filter api
+const events = await client.queryFilter({})
+
+// returns EventState, defined in the lsp client
+const state:uma.clients.optimisticOracle.EventState = uma.clients.optimisticOracle.getEventState(events)
```
mrice32, line 23: Why not just have the user call your function to query and parse the events? Is it better to have them handle the raw events themselves?

packages/sdk/src/clients/optimisticOracle/client.e2e.ts
```diff
@@ -0,0 +1,20 @@
+require("dotenv").config();
```
mrice32, line 1: Just out of curiosity, what's the rationale for unit tests vs end-to-end tests that aren't run in ci? Is there a process for determining what code needs which in the sdk?

mrice32 (approved):
LGTM!

I worry a bit about the scalability of writing these light wrappers. Is there something we can do to avoid having to do this for every contract we interact with? Or is there a way to generalize event processing, so we don't have to transform the events?

I could see it taking less effort to just write code generators based on contract ABIs, kind of like what's done in contracts-node and contracts-frontend or typechain.

packages/sdk/src/clients/optimisticOracle/client.ts
```diff
@@ -0,0 +1,138 @@
...
+  payout: string;
+  state: RequestState;
+};
+
+export interface EventState {
+  requests?: Record<string, Request>;
+}
+
+export function requestId(request: Omit<RequestKey, "timestamp"> & { timestamp: BigNumberish }) {
+  return [request.requester, request.identifier, request.timestamp.toString(), request.ancillaryData].join("!");
```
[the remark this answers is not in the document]
daywiss: yea, theres not a strong reason to use any particular one. This is just a habit from creating sortable compound keys, `!` is the first somewhat uncommon visible character in the ascii table, `~` is also a good one, as its the last printable character.  These have some subtle properties when sorting all these keys, which i cant remember.  If was really important though, we would probably want timestamp first and to normalize the string lengths, but i dont think it matters.

packages/sdk/src/clients/optimisticOracle/README.md
daywiss: i think because i didnt want to assume the source of the events, and i wanted the ability to test without having a live provider. though it is really hard to serialize these events to run later.

packages/sdk/src/clients/optimisticOracle/client.e2e.ts
daywiss: well the rationale is that we dont have hardhat setup in the sdk, so its just easier to run the tests against a real data source.  ideally id like to turn these all into unit tests, and its in the backlog for now.

daywiss:
> I worry a bit about the scalability of writing these light wrappers. Is there something we can do to avoid having to do this for every contract we interact with? Or is there a way to generalize event processing, so we don't have to transform the events?
> 
> I could see it taking less effort to just write code generators based on contract ABIs, kind of like what's done in contracts-node and contracts-frontend or typechain.

I think generalizing this would require cooperation from the contract team to come up with a standard eventing pattern.  it would be nice but i dont know if its really necessary.  I dont see a way around writing bespoke event decoders now though, since the events can be arbitrary and the reduced shape of the data is arbitrary.  not a big deal imo

packages/sdk/src/clients/optimisticOracle/client.ts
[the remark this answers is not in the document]
mrice32: Do we sort these keys? Not opinionated, it was more of a curiosity than anything!

mrice32 (approved):

mrice32:
> > I worry a bit about the scalability of writing these light wrappers. Is there something we can do to avoid having to do this for every contract we interact with? Or is there a way to generalize event processing, so we don't have to transform the events?
> > I could see it taking less effort to just write code generators based on contract ABIs, kind of like what's done in contracts-node and contracts-frontend or typechain.
> 
> I think generalizing this would require cooperation from the contract team to come up with a standard eventing pattern. it would be nice but i dont know if its really necessary. I dont see a way around writing bespoke event decoders now though, since the events can be arbitrary and the reduced shape of the data is arbitrary. not a big deal imo

Yeah, I hear you. I think this makes sense. I'm not sure how we would build a standard for events, and I don't think we can really have a standard way of turning events into state. There are some parts that are generalizable, like transforming type X to type Y. Make sense.

packages/sdk/src/clients/optimisticOracle/client.ts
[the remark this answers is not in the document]
daywiss: Nope currently not being sorted, though i might reconsider that once we add these all to a table

packages/sdk/src/clients/optimisticOracle/README.md
```diff
@@ -0,0 +1,27 @@
+# UMA SDK Optimistic Oracle Client
```
chrismaree, line 1: thanks for adding the docs!

chrismaree: one question: we have all these really useful README files through the repo (mainly in the `sdk` package). I feel that people don't really see them as they are kind of hidden away, nested within a bunch of directories. do you have any ideas on a better way to export these? should these SDK docs be within https://docs.umaproject.org/ to show people how to use the SDK?

daywiss: i've tried to create a links from the readme in the root of the repo to each readmes in each major folder. in general though i dont think the sdk is quite ready for external use, so im not too concerned about making the docs visible.  i think right now the parts of the sdk which are most useful externally is the across client, which is up on the doc site, and potentially the optimistic oracle client, since both of these have an application.  idk though just my rough thoughts on this

daywiss:
> Yeah, I hear you. I think this makes sense. I'm not sure how we would build a standard for events, and I don't think we can really have a standard way of turning events into state. There are some parts that are generalizable, like transforming type X to type Y. Make sense.

I do have some ideas on how to generalize events for smart contracts, but since i dont have any experience with it im not sure if it would work well.  i think we could create a light wrapper around hash maps which create an interface like set/get and during a set, emits an event in the form Set(id,data), and delete Delete(id).  depending on the arch of the contract, you could also add a type to the event, to indicate what table the event is being emitted from. something like that maybe

PR #3791, opened 2022-01-25
fix(pf): use close price for matching cryptowatch close timestamps

Reinis-FRP:

**Motivation**

Normally price feed should pick the last available price at or before request timestamp.

**Summary**

When price request timestamp matches close timestamp for Cryptowatch it now resolves to matching close price (previously it returned open price for the same period).

**Details**

Added check on whether price request timestamp matches end of OHLC period.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
+++ b/packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
@@ -190,7 +190,7 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
     // its close price. We prefer prices that occurred before the target time.
     else {
       // Use period close price when request time matches close time. In all other cases use open price instead.
-      const matchedPrice = time == match.closeTime ? match.closePrice : match.openPrice;
+      const matchedPrice = time === match.closeTime ? match.closePrice : match.openPrice;
       returnPrice = this.invertPrice ? this._invertPriceSafely(matchedPrice) : matchedPrice;
       if (!returnPrice) throw new Error(`${this.uuid} -- invalid price returned`);
       if (verbose) await this._printVerbose(match, returnPrice);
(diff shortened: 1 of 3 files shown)


packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
```diff
@@ -189,7 +189,9 @@ export class CryptoWatchPriceFeed extends PriceFeedInterface {
     // why we prefer `before` periods to `after` ones. Similar reasoning is why we default to a period's open price to
     // its close price. We prefer prices that occurred before the target time.
     else {
-      returnPrice = this.invertPrice ? this._invertPriceSafely(match.openPrice) : match.openPrice;
+      // Use period close price when request time matches close time. In all other cases use open price instead.
+      const matchedPrice = time == match.closeTime ? match.closePrice : match.openPrice;
```
Reinis-FRP, line 193: I acknowledge not using strict equality operator as `time` is number, but `closeTime` is string. Should I have converted `closeTime` to number instead?

nicholaspai: I would convert them to the same type and use strict equality for completeness

nicholaspai: or just do time.toString()

nicholaspai (approved):
+1 LGTM

packages/financial-templates-lib/src/price-feed/CryptoWatchPriceFeed.ts
Reinis-FRP: My assumptions were wrong and this was a problem only when running `getHistoricalPrice.js` and passing `--time` on comandline that. Next commit fixes that.

PR #3792, opened 2022-01-25
CircleCI: Add dockerhub build

evaldofelipe:
Signed-off-by: Evaldo Felipe <contato@evaldofelipe.com>

**Motivation**

This change on CI allow us to keep the docker image updated on docker hub

### DIFF
--- a/packages/common/types/node-metamask/index.d.ts
+++ b/packages/common/types/node-metamask/index.d.ts
@@ -33,3 +33,5 @@ declare module "node-metamask" {
     };
 
 }
+
+// test comment
--- a/packages/common/types/node-metamask/index.d.ts
+++ b/packages/common/types/node-metamask/index.d.ts
@@ -33,5 +33,3 @@ declare module "node-metamask" {
     };
 
 }
-
-// test comment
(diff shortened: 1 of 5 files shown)


chrismaree (commented):
so the point of this is to get builds pushing back here: https://hub.docker.com/r/umaprotocol/protocol/tags

right?

seeing docker hub no longer does this on the free accounts?

will this block CI? if the build takes longer than running the unit tests will this block?

this will only build on merge to master and not on push to the branch, right?

packages/common/types/node-metamask/index.d.ts
```diff
@@ -33,3 +33,5 @@ declare module "node-metamask" {
     };
 
 }
+
+// test comment
```
chrismaree, line 37: 👎

evaldofelipe: That was used to test changes on packages and be able to build a new image. Removed now after tests.

evaldofelipe:
> this will only build on merge to master and not on push to the branch, right?
yes!

chrismaree (approved):

mrice32 (approved):
LGTM!

PR #3794, opened 2022-01-26
fix(sdk): fix chain metadata shape when adding new chain for user

daywiss:

**Motivation**
web3 rpc is very strict when sending data to user to add a new chain, causing errors when including unexpected props. 

**Summary**

added utility to strip unneeded props from object before requesting user add the new chain

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/utils.ts
+++ b/packages/sdk/src/oracle/utils.ts
@@ -6,6 +6,7 @@ import {
   PartialChainConfig,
   ChainConfig,
   PartialConfig,
+  ChainMetadata,
   Config,
 } from "./types/state";
 import type { Provider, TransactionReceipt } from "./types/ethers";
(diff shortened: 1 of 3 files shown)


packages/sdk/src/oracle/utils.ts
```diff
@@ -189,3 +191,10 @@ export class TransactionConfirmer {
     return false;
   }
 }
+
+export function chainConfigToChainMetadata(config: ChainConfig): ChainMetadata {
+  // linters hate this, but we just need to strip out these vars for valid chain metadata
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
mrice32, line 197: You could also add this option to the eslint config so this doesn't cause issues here or in the future!
https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/utils.ts
daywiss: i generally do like this rule!

daywiss: oh thanks for correcting me, this rule is specificallly for this case, i will add that

PR #3795, opened 2022-01-26
feat: add optimistic oracle UI link to bot logs

mrice32:
**Motivation**

We would like to have a UI link in the bot logs so a user can look up a particular OO request when alerted.

**Summary**

This adds a simple markdown link to the logs pointing the user to the page for that request.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -304,6 +344,13 @@ class OptimisticOracleContractMonitor {
       }
     }
   }
+
+  _generateUILink(linkText, requester, identifier, timestamp, ancillaryData) {
+    const hexIdentifier = this.padRight(this.utf8ToHex(identifier), 64);
+    return `<${this.optimisticOracleUIBaseUrl}?chainId=${this.contractProps.chainId}&requester=${this.toChecksumAddress(
+      requester
+    )}&identifier=${hexIdentifier}&timestamp=${timestamp}&ancillaryData=${ancillaryData}|${linkText}>`;
```
daywiss, line 352: `${ancillaryData}|${linkText}`  this looks like its appending the link text after the link, is this so you can split it out of the logs later or what?

packages/monitors/test/OptimisticOracleContractMonitor.js
```diff
@@ -243,6 +244,20 @@ describe("OptimisticOracleContractMonitor.js", function () {
     assert.isTrue(lastSpyLogIncludes(spy, `https://etherscan.io/address/${optimisticRequester.options.address}`));
     assert.isTrue(lastSpyLogIncludes(spy, `https://etherscan.io/tx/${requestTxn.transactionHash}`));
 
+    assert.isTrue(
+      lastSpyLogIncludes(
```
daywiss, line 248: this isnt checking the `|${linktext}` part ( it would still pass with or without it), not sure if that matters but mentioning it in case this was an oversight

daywiss (approved):
lg just some minor comments

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: This is how you tell slack to [do this](https://example.com). See here for where we do this for etherscan links: https://github.com/UMAprotocol/protocol/blob/6b53838d5d74d6f4b4f464ceefe1facd5ba16080/packages/common/src/FormattingUtils.ts#L112.

packages/monitors/test/OptimisticOracleContractMonitor.js
mrice32: True, I was more worried about the link being in the logs than testing exactly what the link says.

I think the exact text in the log isn't something we should test, only that the log contains the information we care about.

But if you disagree, I'm happy to add tests for this.

packages/monitors/index.js
```diff
@@ -103,7 +103,11 @@ async function run({
      * Set variables common to all monitors
      *
      ***************************************/
-    const [networkId, latestBlock] = await Promise.all([web3.eth.net.getId(), web3.eth.getBlock("latest")]);
+    const [networkId, latestBlock, chainId] = await Promise.all([
```
chrismaree, line 106: what is the difference between `net.getId` and `getChainId`? I tested this and they look like they are always the same number.

packages/monitors/src/OptimisticOracleContractMonitor.js
chrismaree: ye, this is correct. for slack you need to create your links as:
```
<https://your-url.com|hyperlink>'
```

so what Matt's code is doing is building the link, then `|` then the "linkText` which will be `View this request in the UI.`.

my one main bit of feedback on this @mrice32 is why make the `linkText` a prop? you could hardcode it here as `View this request in the UI.`? Alternatively, make it the last prop in this function, default it to `View this request in the UI.` and then remove it from all the places you call the function.

packages/monitors/index.js
mrice32: They are usually the same number, but not always https://besu.hyperledger.org/en/stable/Concepts/NetworkID-And-ChainID/#:~:text=Usually%20the%20network%20ID%20is,the%20%2D%2Dnetwork%2Did%20option. For all things outside of node consensus, chainId is recommended over networkId.

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -52,6 +52,11 @@ class OptimisticOracleContractMonitor {
           return Object.values(overrides).every((param) => Object.keys(this.logger.levels).includes(param));
         },
       },
+      optimisticOracleUIBaseUrl: {
+        // Base URL for the Optimistic Oracle UI.
+        value: "https://placeholder.umaproject.org/",
```
chrismaree, line 57: do we not know the URL now? would it not be `"https://optimisticoracle.umaproject.org/"` or `"https://oo.umaproject.org/"`

packages/monitors/test/OptimisticOracleContractMonitor.js
chrismaree: agreed. this is how we do it in other places. you just care about the key info, not the exact string. exact strings make it painful if we choose the change the layout.

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: We don't know the URL. I asked @Tulun and @daywiss and they said it hasn't been decided yet. We can update this when we have a prod url.

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -52,6 +52,11 @@ class OptimisticOracleContractMonitor {
           return Object.values(overrides).every((param) => Object.keys(this.logger.levels).includes(param));
         },
       },
+      optimisticOracleUIBaseUrl: {
```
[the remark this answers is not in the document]
mrice32: Yep! I think ideally we have a good default that we don't need to override. I think overriding may come in handy occasionally if we want to use a staging UI or something temporarily.

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: Yeah, we can hardcode it. I will make that change.

packages/monitors/index.js
chrismaree: do we operate on any networks that dont have these matching?

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: Done.

chrismaree: ye that makes the code cleaner. thanks!

chrismaree (approved):
this is 🔥 🔥 🔥 . Really excited to have this UI up and running! in conjunction with the discord transport, this will be super useful for our community.

packages/monitors/index.js
mrice32: Nope, I don't think so. Although I think this can differ in local hardhat forks if I remember correctly.

mrice32: My basic push is that we should avoid expanding the use of networkId as it isn't recommended for applications.

packages/monitors/test/OptimisticOracleContractMonitor.js
daywiss: Yea sounds good

PR #3797, opened 2022-01-26
feat(transports): Add discord logger transport

chrismaree:
**Motivation**

We want to be able to send messages from our bots to discord. This PR adds a discord transport to facilitate this.

The configs are set to be the same as slack: a `defaultWebHookUrl` that is used for all messages unless a `escalationPathWebhookUrls` is provided for the associated message `notificationPath`.

Sample message

image

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -54,14 +54,15 @@ export class DiscordTransport extends Transport {
     }
 
     // For each markdown link, build the new link with the discord format. Replace the original link with the new one.
+    let modifiedMessage = msg;
     for (let i = 0; i < startIndexes.length; i++) {
       const originalUrl = msg.substring(startIndexes[i], endIndexes[i] + 1);
       const pipeIndex = originalUrl.indexOf("|");
       const markdownUrl =
         `[${originalUrl.substring(pipeIndex + 1, originalUrl.length - 1)}]` + // hyperlink
         `(${originalUrl.substring(1, pipeIndex)})`; // url
-      msg = msg.replace(originalUrl, markdownUrl);
+      modifiedMessage = modifiedMessage.replace(originalUrl, markdownUrl);
     }
-    return msg;
+    return modifiedMessage;
   }
 }
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -1,0 +1,67 @@
+import Transport from "winston-transport";
+
+import axios from "axios";
+
+type TransportOptions = ConstructorParameters<typeof Transport>[0];
+
+export class DiscordTransport extends Transport {
+  private readonly defaultWebHookUrl: string;
+  private readonly escalationPathWebhookUrls: { [key: string]: string };
+  constructor(
+    winstonOpts: TransportOptions,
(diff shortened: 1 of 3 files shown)


packages/financial-templates-lib/src/logger/Transports.ts
```diff
@@ -47,14 +50,24 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
...
     if (slackConfig) {
       transports.push(createSlackTransport(slackConfig));
     }
 
+    // If there is a discord config, create a new transport.
+    if (transportsConfig.discordConfig || process.env.DISCORD_CONFIG) {
+      transports.push(
+        new DiscordTransport(
+          { level: "info" },
+          transportsConfig.discordConfig ?? JSON.parse(process.env.DISCORD_CONFIG || "null")
```
mrice32, line 66: Won't null cause this constructor to fail? It looks like that case isn't handled, and even if there were default args, wouldn't they only work in the `undefined` case, not `null`?

mrice32 (commented):
Looks awesome!!! Minor comments.

packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -0,0 +1,68 @@
...
+  // Discord URLS are formatted differently to markdown links produced upstream in the bots. For example, slack links
+  // will look like this <https://google.com|google.com> but links for discord should look like this
+  // [google.com](https://google.com).This function takes in the one format and converts it to the other such that
+  // links sent to discord are nicely formatted and clickable.
+  formatLinks(msg: any): any {
+    // Find the start and end indexes of all the markdown links.
+    const startIndexes: any = [];
+    const endIndexes: any = [];
+    for (let i = 0; i < msg.length; i++) {
+      const startIndex = msg.indexOf("<http", i);
```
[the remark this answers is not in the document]
chrismaree: ye I can easily change this to just search for the starting `<` and it'll work with all other links.

packages/financial-templates-lib/src/logger/Transports.ts
chrismaree: it would not get to this point of it's undefined. this code block is only entered `if (transportsConfig.discordConfig || process.env.DISCORD_CONFIG) {`.

Also note that `JSON.parse("null")` returns undefined which would be fine with the constructor.

packages/financial-templates-lib/src/logger/DiscordTransport.ts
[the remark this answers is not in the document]
chrismaree: I've changed this to search from `<`.

packages/financial-templates-lib/src/logger/Transports.ts
mrice32: Haha, of course JSON.parse("null") returns undefined.

So two points here:
1. If we never get here with both of these being undefined/null, then why do we need the || "null"? Shouldn't that be _impossible_ to hit?
2. I don't think undefined works in the constructor. I think values on `opts` are used, but undefined.[anything] will throw.

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/logger/Transports.ts
chrismaree: I agree with both of your points. the net effect is we should remove the `|| "null"`. this was a carry over from the other transports. I'll remove it from all of them to be consistant.

evaldofelipe (approved):

PR #3800, opened 2022-01-27
feat(transports): Modify discord transport config

chrismaree:
**Motivation**

We dont always want to post all messages to the discord transport, like we do with other transports. For example, you might only want to post messages that include the specific escalation path. eg you send OO messages to slack but not risk/infrastructure messages.

this PR adds another setting`postOnNonEscalationPaths` that lets the user toggle this behavour.

**Details**

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -34,7 +34,7 @@ export class DiscordTransport extends Transport {
       // Select webhook from escalation path. If no escalation, use default webhook url, if postOnNonEscalationPaths is
       // true.Else, if postOnNonEscalationPaths is false and there is no escalation path, do not post. This enables the
       // transport to be configured to only send messages that have a specific escalation path.
-      if (this.postOnNonEscalationPaths && this.escalationPathWebhookUrls[info.notificationPath] != undefined) {
+      if (this.postOnNonEscalationPaths && this.escalationPathWebhookUrls[info.notificationPath]) {
         const webHook = this.escalationPathWebhookUrls[info.notificationPath] ?? this.defaultWebHookUrl;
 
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -25,11 +31,15 @@ export class DiscordTransport extends Transport {
         embeds: [{ title: `${info.at}: ${info.message}`, description: this.formatLinks(info.mrkdwn), color: 9696729 }],
       };
 
-      // Select webhook url based on escalation path. if escalation path is not found, use default webhook url.
-      const webHook = this.escalationPathWebhookUrls[info.notificationPath] ?? this.defaultWebHookUrl;
+      // Select webhook from escalation path. If no escalation, use default webhook url, if postOnNonEscalationPaths is
+      // true.Else, if postOnNonEscalationPaths is false and there is no escalation path, do not post. This enables the
+      // transport to be configured to only send messages that have a specific escalation path.
+      if (this.postOnNonEscalationPaths && this.escalationPathWebhookUrls[info.notificationPath] != undefined) {
```
mrice32, line 37: nit: did you intentionally put `!=` vs `!==` so that you check for null as well? If so, I think `this.postOnNonEscalationPaths && this.escalationPathWebhookUrls[info.notificationPath]` would be clearer (i.e. no != at all). If you only want to check for undefined, I would use `!==`.

mrice32 (approved):
LGTM!

packages/financial-templates-lib/src/logger/DiscordTransport.ts
chrismaree: ye, good point. will revert to your syntax.

PR #3803, opened 2022-01-28
improve(sdk): fix request state flags, poll for request changes

daywiss:
BREAKING CHANGE: oo flags have changed names for clarity

**Motivation**

Added some new flags for easier ui code, and also continually update active request to catch possible state changes ( and set flags accordingly)

**Summary**
Refactor flag names, add additional names to cover ui states needed based on request state. Add a command to conitinually update request if its in a state that could potentially change. Also add a block time per chain.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/README.md
+++ b/packages/sdk/src/oracle/README.md
@@ -99,14 +99,17 @@ export enum Flag {
   MissingRequest = "MissingRequest", // the client does not know the request, use client.setActiveRequest
   MissingUser = "MissingUser", // client does not have user data, use client.setUser
   WrongChain = "WrongChain", // user and request chain ids do not match, switch chains with client.switchOrAddChain
-  InProposeState = "InProposeState", // The on chain request is in a state where someone could propose, use client.proposePrice
-  InDisputeState = "InDisputeState", // The on chain request is in a stae where someone could dispute, use client.disputePrice
+  CanPropose = "CanPropose", // The on chain request is in a state where someone could propose, use client.proposePrice
+  CanDispute = "CanDispute", // The on chain request is in a state where someone could dispute, use client.disputePrice
+  CanSettle = "CanSettle", // The on chain request is in a stae where someone could settle the request.
+  InDvmVote = "InDvmVote", // Proposed answer has been disputed and passed to dvm for full vote.
+  RequestSettled = "RequestSettled", // Request is finalized, no more changes.
   InsufficientBalance = "InsufficientBalance", // The user does not have enough balance to cover bond collateral for dispute/propose
   InsufficientApproval = "InsufficientApproval", // The oracle contract does not have enough approval to cover bond for dispute/propose, use client.approve
   ChainChangeInProgress = "ChainChangeInProgress", // The user is changing his chain
-  ProposalInProgress = "ProposalInProgress", // The user is sending a proposal tx
-  ApprovalInProgress = "ApprovalInProgress", // The user is sending an approval tx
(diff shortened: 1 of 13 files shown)


packages/sdk/src/oracle/utils.ts
```diff
@@ -17,86 +17,84 @@ import { ethers } from "ethers";
...
+  const request = ignoreError(read.request);
 
-  try {
-    flags[Flag.InDisputeState] = read.request()?.state === RequestState.Proposed;
-  } catch (err) {
-    flags[Flag.InDisputeState] = false;
-  }
+  // these are a bit redundant with request state, but just an alternate way to see current request state
+  flags[Flag.CanPropose] = request?.state === RequestState.Requested;
+  flags[Flag.CanDispute] = request?.state === RequestState.Proposed;
```
daywiss, line 68: previous bug, this dispute flag could get out of date becuase it should also take into consideration expiration time. rather than add the logic here, decided rather to poll the request, updating from chain on an interval, that way this flag logic can remain simple and based on the currently known request state.  this is to prevent an edge case where we might disable dispute button even tho request may be disputable

packages/sdk/src/oracle/README.md
```diff
@@ -99,14 +99,17 @@ export enum Flag {
   MissingRequest = "MissingRequest", // the client does not know the request, use client.setActiveRequest
   MissingUser = "MissingUser", // client does not have user data, use client.setUser
   WrongChain = "WrongChain", // user and request chain ids do not match, switch chains with client.switchOrAddChain
-  InProposeState = "InProposeState", // The on chain request is in a state where someone could propose, use client.proposePrice
-  InDisputeState = "InDisputeState", // The on chain request is in a stae where someone could dispute, use client.disputePrice
+  CanPropose = "CanPropose", // The on chain request is in a state where someone could propose, use client.proposePrice
+  CanDispute = "CanDispute", // The on chain request is in a state where someone could dispute, use client.disputePrice
+  CanSettle = "CanSettle", // The on chain request is in a stae where someone could settle the request.
```
daywiss, line 104: i renamed these to make more sense from the perspective of the user, since the state is actionable

packages/sdk/src/oracle/README.md
```diff
@@ -99,14 +99,17 @@ export enum Flag {
...
+  RequestSettled = "RequestSettled", // Request is finalized, no more changes.
   InsufficientBalance = "InsufficientBalance", // The user does not have enough balance to cover bond collateral for dispute/propose
   InsufficientApproval = "InsufficientApproval", // The oracle contract does not have enough approval to cover bond for dispute/propose, use client.approve
   ChainChangeInProgress = "ChainChangeInProgress", // The user is changing his chain
-  ProposalInProgress = "ProposalInProgress", // The user is sending a proposal tx
-  ApprovalInProgress = "ApprovalInProgress", // The user is sending an approval tx
-  DisputeInProgress = "DisputeInProgress", // The user is sending a dispute tx
+  ProposalTxInProgress = "ProposalTxInProgress", // The user is sending a proposal tx
+  ApprovalTxInProgress = "ApprovalTxInProgress", // The user is sending an approval tx
+  DisputeTxInProgress = "DisputeTxInProgress", // The user is sending a dispute tx
```
daywiss, line 112: renamed these to specify that its a users transaction in progress, not the state of the request

mrice32 (approved):
LGTM!

amateima (approved):

PR #3804, opened 2022-01-28
feat(monitors): split UI link from other log messages due to length 

mrice32:
**Motivation**

UI links overflow slack messages.

**Summary**

This just splits the UI link into its own message.

Note: these UI links are never sent as errors. If the regular message is sent as error, this UI link is downgraded to info, so it won't send alerts.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/monitors/src/OptimisticOracleContractMonitor.js
+++ b/packages/monitors/src/OptimisticOracleContractMonitor.js
@@ -119,10 +119,11 @@ class OptimisticOracleContractMonitor {
           )
         )}. tx: ${createEtherscanLinkMarkdown(event.transactionHash, this.contractProps.networkId)}`;
 
+      const logLevel =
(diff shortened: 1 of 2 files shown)


chrismaree (approved):
maximum pain. I'm really not a fan of this solution...and it's also not a stop gap as there can still be messages that are too long...but for now it's ok. let's make sure we update the front end next sprint so we can revert this!

mrice32:
> maximum pain. I'm really not a fan of this solution...and it's also not a stop gap as there can still be messages that are too long...but for now it's ok. let's make sure we update the front end next sprint so we can revert this!

Agree, we should really just generally make sure all inputs to our logs have a limited length tbh. Long logs have been one of the biggest causes for bot pain.

PR #3806, opened 2022-01-31
feat(across-bots): Send funds more aggressivly over bridge when util is high

chrismaree:
If pool utilization is high we should more aggresivly send funds over the bridge. This PR adds a 75% cutoff at which the `crossDomainFinalizationThreshold` is halfed to make bridging more aggressive as we run out of funds.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
+++ b/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
@@ -67,8 +67,8 @@ export class CrossDomainFinalizer {
     for (const l2Token of bridgeableL2Tokens) {
       // For each bridgeable L2Token, check the balance in the deposit box. If it is greater than
       // crossDomainFinalizationThreshold, as a percentage, then we can bridge it. If the liquidity utilization is
-      // greater than 75% then half the crossDomainFinalizationThreshold to send funds more aggressively over the
-      // bridge at high utilizations.
+      // greater than 75% then set the crossDomainFinalizationThreshold to half its set value to send funds more
+      // aggressively over the bridge at high utilizations.
 
       try {
         const { symbol, decimals, l2PoolBalance } = await this._getL2TokenInfo(l2Token);
--- a/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
+++ b/packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
@@ -66,7 +66,7 @@ export class CrossDomainFinalizer {
     let nonce = await this.l2Client.l2Web3.eth.getTransactionCount(this.account);
     for (const l2Token of bridgeableL2Tokens) {
       // For each bridgeable L2Token, check the balance in the deposit box. If it is greater than
-      // crossDomainFinalizationThreshold, as a percentage, then we can bridge it. if the liquidity utilization is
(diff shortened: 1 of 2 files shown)


packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -65,13 +66,20 @@ export class CrossDomainFinalizer {
     let nonce = await this.l2Client.l2Web3.eth.getTransactionCount(this.account);
     for (const l2Token of bridgeableL2Tokens) {
       // For each bridgeable L2Token, check the balance in the deposit box. If it is greater than
-      // crossDomainFinalizationThreshold, as a percentage, then we can bridge it.
+      // crossDomainFinalizationThreshold, as a percentage, then we can bridge it. If the liquidity utilization is
+      // greater than 75% then half the crossDomainFinalizationThreshold to send funds more aggressively over the
```
nicholaspai, line 70: > than 75% then half

is this a typo?

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
```diff
@@ -65,13 +66,20 @@ export class CrossDomainFinalizer {
...
+      // bridge at high utilizations.
 
       try {
         const { symbol, decimals, l2PoolBalance } = await this._getL2TokenInfo(l2Token);
         const l1PoolReserves = await this._getL1PoolReserves(l2Token);
+        const l1PoolUtilization = await this._getL1PoolUtilization(l2Token);
 
-        if (l2PoolBalance.gt(toBN(this.crossDomainFinalizationThreshold).mul(l1PoolReserves).div(toBN(100)))) {
+        const scaledCrossDomainFinalizationThreshold = l1PoolUtilization.gt(toBNWei("0.75"))
```
nicholaspai, line 78: should this 0.75 value be parameterized?

nicholaspai, line 78: I'm curious why you slash the `crossDomainFinalizationThreshold` figure instead of just using the utilization threshold to determine whether to bridge. For example, could you not just set a rule: if utilization > 75% OR L2PoolBalance > (`scaledCrossDomainFinalizationThreshold * l1PoolReserves`) then bridge

nicholaspai: I'll approve but I think its somewhat likely we realize 0.75 threshold is too slow/too fast and we'll have to annoyingly wait for CI to pass to change the value so if you have time I would just parameterize this so can we modify it easier with a config change

nicholaspai (approved):

packages/insured-bridge-relayer/src/CrossDomainFinalizer.ts
chrismaree: because you don't always want to bridge if we are above 75%. this is an _and_ condition, not an _or_ condition. i.e if we are at 76% utilization and the bridge pool has 10 wei in it we don't want to bridge. briding is expensive so should not always be done but rather be done _more_ quickly.

PR #3812, opened 2022-02-02
feat(sdk): request list checker using events

daywiss:

**Motivation**
We need a way to fetch oracle events to build up a table of requests we can lookup from. 

**Summary**
This adds 2 new commands which run at the start of the app. One is responsible for fetching historical events in a robust way, the other fetches all new events.  Historical events are tricky because of chains other than mainnet restricting the range of blocks you can query. This adds logic to dynamically adjust the range to prevent errors, and also allows a configurable "start block" on each chain to prevent history from going back to block 0.  Eventually these events will be used to populate a request index page. 

**Details**
Event requests spanning too large a block range cause errors on polygon-infura. To be really resilient to this, we reduce the block range requested with 3 static utility functions. These functions start a range at the newest events, and halve the range each time an error occurs, until no more errors occur for the range. it will then query each range from newest to oldest until it reaches block 0.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -11,12 +11,7 @@ import { defaultConfig } from "./utils";
 
 export class Client {
   private intervalStarted = false;
-  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {
-    // create active request poller for all chains. Should only have one of these
-    sm.types.pollActiveRequest.create(undefined);
-    // polls user for balances/approvals on the current chain, in case it changes external to app
-    sm.types.pollActiveUser.create(undefined);
-  }
+  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {}
   setUser(params: Partial<User>): string {
     const address = params.address && ethers.utils.getAddress(params.address);
     return this.sm.types.setUser.create({ ...params, address });
@@ -112,6 +107,7 @@ export class Client {
       assert(this.intervalStarted, "Interval Stopped");
       await this.sm.tick();
     }, delayMs).catch((err) => {
+      console.error(err);
       this.intervalStarted = false;
       this.store.write((w) => w.error(err));
     });
@@ -124,8 +120,8 @@ export class Client {
 
 export function factory(config: state.PartialConfig, emit: Emit): Client {
   const store = new Store(emit);
+  const fullConfig = defaultConfig(config);
   store.write((write) => {
-    const fullConfig = defaultConfig(config);
     write.config(fullConfig);
     for (const chain of Object.values(fullConfig.chains)) {
       write.chains(chain.chainId).optimisticOracle().address(chain.optimisticOracleAddress);
@@ -135,6 +131,29 @@ export function factory(config: state.PartialConfig, emit: Emit): Client {
     }
   });
   const update = new Update(store);
+  // we can create "threads" by introducing new state machine queues, the first one is user actions
   const sm = new StateMachine(store);
+  // this one is system actions used for long running commands independent of the user
+  const poller = new StateMachine(store);
+
+  // start the request list checkers
+  for (const [chainId, config] of Object.entries(fullConfig.chains)) {
+    poller.types.fetchPastEvents.create({ chainId: Number(chainId), startBlock: config.earliestBlockNumber }, "poller");
+    // long running poller which only looks for new events
+    poller.types.pollNewEvents.create({ chainId: Number(chainId) }, "poller");
+  }
+
+  // create active request poller for all chains. Should only have one of these
+  poller.types.pollActiveRequest.create(undefined, "poller");
+  // polls user for balances/approvals on the current chain, in case it changes external to app
+  poller.types.pollActiveUser.create(undefined, "poller");
+
+  // poller statemachine runs on its own loop
+  loop(async () => {
+    return poller.tick();
+  }, 10).catch((err) => {
+    console.error("poller error", err);
+    store.write((w) => w.error(err));
+  });
   return new Client(store, update, sm);
 }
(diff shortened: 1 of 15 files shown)


packages/sdk/src/oracle/services/optimisticOracle.ts
```diff
@@ -1,16 +1,16 @@
-import { OptimisticOracleEthers, OptimisticOracleEthers__factory } from "@uma/contracts-node";
+import { optimisticOracle } from "../../clients";
```
daywiss, line 1: just switching this out for new client rather than the contracts-node interface

packages/sdk/src/oracle/store/write.ts
```diff
@@ -68,19 +69,18 @@ export class OptimisticOracle {
...
+    const id = requestId(inputRequest);
     if (!this.state.requests) this.state.requests = {};
     this.state.requests[id] = request;
   }
   defaultLiveness(defaultLiveness: ethersTypes.BigNumber): void {
     this.state.defaultLiveness = defaultLiveness;
   }
+  event(event: state.OptimisticOracleEvent): void {
+    if (!this.state.events) this.state.events = [];
+    insertOrderedAscending(this.state.events, event, eventKey);
```
daywiss, line 82: this makes sure the events maintain order in array. we basically will cache a history of all known events

packages/sdk/src/oracle/client.ts
```diff
@@ -11,12 +11,7 @@ import { defaultConfig } from "./utils";
 
 export class Client {
   private intervalStarted = false;
-  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {
-    // create active request poller for all chains. Should only have one of these
-    sm.types.pollActiveRequest.create(undefined);
-    // polls user for balances/approvals on the current chain, in case it changes external to app
-    sm.types.pollActiveUser.create(undefined);
```
daywiss, line 18: move these into their own queue to not interfere with execution of user actions

packages/sdk/src/oracle/utils.ts
```diff
@@ -194,3 +197,127 @@ export function chainConfigToChainMetadata(config: ChainConfig): ChainMetadata {
...
+  transactionIndex: BigNumberish;
+  logIndex: BigNumberish;
+}): string {
+  return [
+    // we pad these because numbers of varying lengths will not sort correctly, ie "10" will incorrectly sort before "9", but "09" will be correct.
+    event.blockNumber.toString().padStart(16, "0"),
+    event.transactionIndex.toString().padStart(16, "0"),
+    event.logIndex?.toString().padStart(16, "0"),
+    // ~ is the last printable ascii char, so it does not interfere with sorting
+  ].join("~");
```
mrice32, line 309: Why not join with ""?

packages/sdk/src/oracle/utils.ts
```diff
@@ -194,3 +197,127 @@ export function chainConfigToChainMetadata(config: ChainConfig): ChainMetadata {
...
+ * @returns {string} - the unique id
+ */
+export function eventKey(event: {
+  blockNumber: BigNumberish;
+  transactionIndex: BigNumberish;
+  logIndex: BigNumberish;
+}): string {
+  return [
+    // we pad these because numbers of varying lengths will not sort correctly, ie "10" will incorrectly sort before "9", but "09" will be correct.
+    event.blockNumber.toString().padStart(16, "0"),
```
mrice32, line 305: I was originally confused as to why you were doing this, and then I realized lodash (and vanilla js) don't have a reversed version of this function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex. :(. Kinda frustrating, since this sort of key-creation is a little nonstandard and error-prone for comparing numbers.

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
```diff
@@ -0,0 +1,57 @@
...
+      try {
+        // this just queries events between start and end
+        await update.oracleEvents(chainId, currentStart, currentEnd);
+        // we signal that the current range was a success, now move currentStart, currentEnd accordingly
+        // we set multiplier to 1 so we dont grow the range on success, this tends to create more errors and slow down querying
+        memory.state = rangeSuccessDescending({ ...rangeState, multiplier: 1 });
+      } catch (err) {
+        memory.error = err;
+        // the provider threw an error so we will reduce our range by moving startblock closer to endblock next iteration
+        memory.state = rangeFailureDescending(rangeState);
```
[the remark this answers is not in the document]
daywiss: yea, its just really hard to know the types of errors that can be thrown here, since its completely undocumented afaik.  the reason why id say this particular try catch is ok, is that the only possible error its emitted from the `update.oracleEvents` as it calls the provider to get events AND this will eventually fail if too many errors are thrown, because `rangeFailureDescending` will reduce the range until 0, at which point all polling will stop and an error will be attached to the context. At that point we can debug to figure out what errors were thrown both on `context.error` and `context.memory.error` which would be 2 distinct errors.

packages/sdk/src/oracle/utils.ts
daywiss: yea thats a good point, for this case it could just be joined by `""`, i think its just a habit of always wanting to be able to split the key into its component parts, but in this case they are always attached to the event

packages/sdk/src/oracle/utils.ts
daywiss: this wouldn't be necessary if lodash provided a `comparator(a,b)` instead of a `rank(a)` function.  And like you said this becomes a bigger annoyance when trying to maintain a descending list, its just not possible using off the shelf utils like this afaik

amateima (approved):
💯 Very nice & clean implementation for the adjustable queried block range.

PR #3813, opened 2022-02-02
fix(sdk): fix properly unqueuing commands from statemachine queue

daywiss:

**Motivation**
State machine queue was acting like a stack and not a queue, which could cause commands to starve.

**Summary**

Use shift instead of pop when dequeuing

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/tests/client.e2e.ts
+++ b/packages/sdk/src/oracle/tests/client.e2e.ts
@@ -56,12 +56,14 @@ describe("Oracle Client", function () {
   test("setRequest", async function () {
     const id = client.setActiveRequest(request);
     await client.sm.tick();
+    await client.sm.tick();
     const input = store.read().inputRequest();
     assert.ok(input);
     assert.ok(store.read().command(id));
   });
   test("setUser", async function () {
     const id = client.setUser({ address, chainId, signer, provider });
+    await client.sm.tick();
     await client.sm.tick();
     const state = store.get();
     assert.ok(state?.inputs?.user);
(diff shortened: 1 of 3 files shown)


packages/sdk/src/oracle/tests/client.e2e.ts
```diff
@@ -56,13 +56,15 @@ describe("Oracle Client", function () {
...
     await client.sm.tick();
+    await client.sm.tick();
     const input = store.read().inputRequest();
     assert.ok(input);
     assert.ok(store.read().command(id));
   });
   test("setUser", async function () {
     const id = client.setUser({ address, chainId, signer, provider });
     await client.sm.tick();
+    await client.sm.tick();
```
mrice32, line 67: Odd that you need to call this twice in a row after only one action on the client?

mrice32 (approved):

packages/sdk/src/oracle/tests/client.e2e.ts
amateima: ➕

amateima (approved):

packages/sdk/src/oracle/tests/client.e2e.ts
daywiss: yea, it is odd, but there is a new long running polling command introduced [here ](https://github.com/UMAprotocol/protocol/blob/bug/sc-4396/fix-queuing-bug-in-statemachine-oo-client/packages/sdk/src/oracle/client.ts#L16)in the client constructor which makes it hard to control which command is getting run. i dont think the e2e test was updated when thatn change was introduced, and i was getting flaky results only running the tick once.

PR #3814, opened 2022-02-02
fix(sdk): expose more across pool values and add badger/wbtc to e2e tests

daywiss:
**Motivation**
This started off as a potential bug with badger pool utilization, but no bug was found.

**Summary**

Ended up exposing utilization on pool and added some additional e2e pool tests.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -42,6 +42,7 @@ export type Pool = {
   blocksElapsed: number;
   secondsElapsed: number;
   liquidityUtilizationCurrent: string;
+  utilizedReserves: string;
   projectedApr: string;
 };
 export type User = {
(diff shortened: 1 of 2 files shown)


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -42,6 +42,7 @@ export type Pool = {
   blocksElapsed: number;
   secondsElapsed: number;
   liquidityUtilizationCurrent: string;
+  utilizedReserves: string;
```
mrice32, line 45: What is this changing? This seems like a bug fix?

mrice32 (commented):
Just one comment!

amateima (approved):

packages/sdk/src/across/clients/bridgePool.ts
daywiss: there was no bug, see resolution in [this thread](https://risklabs.slack.com/archives/C01RS73UX6K/p1643727262721569?thread_ts=1643688966.742979&cid=C01RS73UX6K).  The reason why i expose this now is to faciliate easier debugging if something like this happens again, this value was not being exposed, so it required going into sdk code, when it could easily be available as state on the FE.

mrice32 (approved):

PR #3815, opened 2022-02-02
feat(sdk): poll user balance and approval for oo client

daywiss:

**Motivation**
user balances/ approvals can change outside the app

**Summary**

adds a command to poll the active user on the active chain. its initialized on app construction. It uses the tx rate to determine how fast to poll per chain.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/services/statemachines/pollActiveUser.ts
+++ b/packages/sdk/src/oracle/services/statemachines/pollActiveUser.ts
@@ -1,38 +1,31 @@
-// this should only run as a single instance, will continue updating the active request as needed.
-// this is a single poller for requests on all chains
 import { Update } from "../update";
 import Store from "../../store";
-import { RequestState } from "../../types/state";
 import { Handlers as GenericHandlers } from "../../types/statemachine";
 import { ContextClient } from "./utils";
 import { ignoreError } from "../../utils";
 
-// required exports for state machine
 export type Params = undefined;
 export type Memory = { error?: Error; iterations: number };
+
 export function initMemory(): Memory {
   return { iterations: 0 };
 }
+
 export function Handlers(store: Store): GenericHandlers<Params, Memory> {
   const update = new Update(store);
   return {
-    async start(params: Params, memory: Memory, ctx: ContextClient): Promise<void | undefined> {
+    async start(params: Params, memory: Memory, ctx: ContextClient) {
       memory.error = undefined;
       try {
-        const request = store.read().request();
-        // requests can change externally if not already in one of these states
-        const shouldUpdate = request.state !== RequestState.Invalid && request.state !== RequestState.Settled;
-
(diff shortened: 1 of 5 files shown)


packages/sdk/src/oracle/services/statemachines/pollActiveUser.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ignoreError } from "../../utils";
+
+export type Params = undefined;
+export type Memory = { error?: Error; iterations: number };
+
+export function initMemory(): Memory {
+  return { iterations: 0 };
+}
+
+export function Handlers(store: Store): GenericHandlers<Params, Memory> {
```
mrice32, line 14: Just a high-level comment: do we rely on this poll to update after the user sends the approval transaction, for instance? Or do we force-update after that (is it possible to force-update from the FE)?

The only reason I ask is that a 30s poll can create a bad UX if the user has to wait that long to see the UI update.

daywiss: when we know that the user is interacting with our app, we will update users balance after a transaction is confirmed. This is only for the case where the balance could change external to our app.  This is mainly a problem when testing, as we might seed the account with a bunch of uma through an external script, but it will require a refresh of the app to detect it.

daywiss: [this is an example](https://github.com/UMAprotocol/protocol/blob/master/packages/sdk/src/oracle/services/statemachines/disputePrice.ts#L47) of handling the dispute transaction, and updating the users balance automatically after tx confirms

packages/sdk/src/oracle/services/statemachines/pollActiveUser.ts
```diff
@@ -0,0 +1,33 @@
...
+      memory.error = undefined;
+      try {
+        // these will fail if request is not set, but thats ok because statemachine will catch and log and re-run once set
+        await update.userCollateralBalance();
+        await update.oracleAllowance();
+      } catch (err) {
+        // its ok to ignore these errors
+        memory.error = (err as unknown) as Error;
+      }
+      memory.iterations++;
```
amateima, line 27: What's the usage of keeping the iterations number?

daywiss: its really for debugging. we can inspect global state to see if this is running, and that its not running too fast, out side of that its not strictly needed

packages/sdk/src/oracle/services/statemachines/pollActiveUser.ts
mrice32: Understood. Makes sense!

mrice32 (approved):
LGTM!

PR #3817, opened 2022-02-02
feat(Discord-transport): Only send discord logs if the message contains mrkdwn

chrismaree:
**Motivation**

We only want to send messages to Discord if and only if the message contains markdown.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -25,6 +25,7 @@ export class DiscordTransport extends Transport {
   // Note: info must be any because that's what the base class uses.
   async log(info: any, callback: () => void): Promise<void> {
     try {
+      if (!info.mrkdwn) return; // We only ever want to send messages to Discord that have Markdown in them.
       const body = {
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -25,6 +25,7 @@ export class DiscordTransport extends Transport {
   // Note: info must be any because that's what the base class uses.
   async log(info: any, callback: () => void): Promise<void> {
     try {
+      if (!info.mrkdwn) return; // We only ever want to send messages to Discord that have Markdown in them.
```
mrice32, line 28: Just OOC, why?

mrice32 (approved):

packages/financial-templates-lib/src/logger/DiscordTransport.ts
Reinis-FRP: Otherwise `description: this.formatLinks(info.mrkdwn)` passes `undefined` that throws an error in `formatLinks`

PR #3818, opened 2022-02-03
feat(sdk): add request list populated by oo events from newest to oldest

daywiss:

**Motivation**
Provide a list of requests, sorted from newest to oldest. Keep list up to date with new requests.

**Summary**
This uses the existing sorted store to maintain a unique table of all requests across all chains. Its sorted by timestamp.  The full list gets attached to global state and is kept up to date as soon as new data comes in, meaning a FE client can use this list directly. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/clients/optimisticOracle/client.ts
+++ b/packages/sdk/src/clients/optimisticOracle/client.ts
@@ -24,36 +24,49 @@ export enum RequestState {
   Settled, // Final price has been set in the contract (can get here from Expired or Resolved).
 }
 
+// all events have these values
 export type RequestKey = {
   requester: string;
   identifier: string;
   timestamp: number;
   ancillaryData: string;
 };
-export type Request = RequestKey & {
-  proposer: string;
-  disputer: string;
-  currency: string;
-  settled: boolean;
-  refundOnDispute: boolean;
-  proposedPrice: string;
-  resolvedPrice: string;
-  expirationTime: string;
-  reward: string;
-  finalFee: string;
-  bond: string;
-  customLiveness: string;
-  price: string;
-  payout: string;
-  state: RequestState;
-};
+export type Request = RequestKey &
+  // this is partial since we dont know what events we have to populate parts of this
+  Partial<{
+    proposer: string;
+    disputer: string;
+    currency: string;
+    settled: boolean;
+    refundOnDispute: boolean;
+    proposedPrice: string;
+    resolvedPrice: string;
+    expirationTime: number;
+    reward: string;
+    finalFee: string;
+    bond: string;
+    customLiveness: string;
+    price: string;
+    payout: string;
+    state: RequestState;
+    // metadata about the transaction that triggered the state changes
+    requestTx: string;
+    proposeTx: string;
+    disputeTx: string;
+    settleTx: string;
+    requestBlockNumber: number;
+    proposeBlockNumber: number;
+    disputeBlockNumber: number;
+    settleBlockNumber: number;
+  }>;
 
 export interface EventState {
   requests?: Record<string, Request>;
 }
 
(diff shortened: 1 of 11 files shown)


packages/sdk/src/stores/sorted-js-map/store.ts
```diff
@@ -39,10 +39,11 @@ export default function <Id, Data>(): SortedStore<Id, Data> {
     async set(id: Id, data: Data) {
       if (map.has(id)) {
         map.set(id, data);
+      } else {
+        const index = sortedIndex(ids, id);
+        ids.splice(index, 0, id);
+        map.set(id, data);
```
daywiss, line 45: found this bad bug, causing duplicate keys in the ids array.  this is used in api server, but have not noticed a problem. might be good to bump sdk in api once this is published.

packages/sdk/src/oracle/store/read.ts
```diff
@@ -159,8 +161,19 @@ export default class Read {
...
+  };
   oracleEvents = (chainId: number): OptimisticOracleEvent[] => {
     const chain = this.state?.chains?.[chainId];
     return chain?.optimisticOracle?.events || [];
   };
+  listChains = (): number[] => {
+    return Object.keys(this.state?.chains || {}).map(Number);
+  };
+  descendingRequests = (): RequestIndexes => {
+    return this.state.descendingRequests || [];
```
daywiss, line 177: This is the request list for the index page. frontend client can access this state directly or call the reader to get it.

packages/sdk/src/oracle/services/sortedRequests.ts
```diff
@@ -0,0 +1,16 @@
...
+import { RequestIndex, RequestIndexes } from "../types/state";
+
+export class SortedRequests {
+  private table = SortedJsMap<string, RequestIndex>();
+  async set(key: string, value: RequestIndex): Promise<void> {
+    await this.table.set(key, value);
+  }
+  async descending(): Promise<RequestIndexes> {
+    // sadly you cannot control lodash sorting descending, so reverse is necessary
+    return (await this.table.values()).reverse();
```
mrice32, line 11: Does values return a copy? I only ask because this reverse has side effects on the underlying array

mrice32 (approved):
LGTM!

amateima (approved):

packages/sdk/src/clients/optimisticOracle/client.ts
```diff
@@ -24,36 +24,49 @@ export enum RequestState {
...
-  finalFee: string;
-  bond: string;
-  customLiveness: string;
-  price: string;
-  payout: string;
-  state: RequestState;
-};
+export type Request = RequestKey &
+  // this is partial since we dont know what events we have to populate parts of this
+  Partial<{
```
Gamaranto, line 36: just my opinion, not requesting a change, but is there any field that will always be there on all requests? I'm asking because `Partial` is particularly annoying for consumers as you need to check every field for possible undefined

daywiss: the only fields guaranteed to be there are the "request key" im calling it, which are the properties required to uniquely identify the request: identifier, requester, timestamp, ancillary data. The rest have to be partial because they are dependent on the set of events we have, and unfortunately due to querying constraints on providers, we cant have the whole set of events quickly.  I agree this is really annoying to deal with, but i just dont see a way around this at the moment, so better to make this explicit in the type.

packages/sdk/src/oracle/services/sortedRequests.ts
daywiss: yea great point, i didnt even think of that. it does return a copy fortunately!

PR #3819, opened 2022-02-04
feat(sdk): allow oo client to fetch request by transaction and option…

daywiss:
…al event index

**Motivation**
Allow ui to accept transaction has to lookup a request

**Summary**

Adds a command that figures out the request based on transaction hash and optional event index.

**Details**

Originally i was going to switch all request state to the global request list fetched from events, but this had a lot of caveats, and determined it was better to implement this way, ie do the event lookup and query contract for request state.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/README.md
+++ b/packages/sdk/src/oracle/README.md
@@ -130,6 +130,7 @@ Commands on the client are currently:
 - proposePrice
 - disputePrice
 - switchOrAddChain
+- setActiveRequestByTransaction
 
 These functions are all syncronous and can throw errors, but on success will return a string which represents the id of the command issued.
 You can fetch a command through:
@@ -153,9 +154,11 @@ The oracle client is what manages state, emits state updates and lets you intera
 
 ### Contruction
 
-This uses a factory for construction.
+This uses a factory for construction.  
+See `oracle/types/state.PartialConfig` for your full configuration object.  
+See `oracle/types/state.State` for full application state object.
 
-`oracle.client.factory(config: oracle.types.state.Config, emit: oracle.services.store.Emit): Client`
+`oracle.client.factory(config: oracle.types.state.PartialConfig, emit: (state: oracle.types.state.State, prev: oracle.types.state.State) => void): Client`
 
 ### setUser
 
@@ -173,7 +176,7 @@ Clears the current user.
 
 Tell the client what request you want to interact with.
 
-`client.setActiveRequest(params: oracle.types.state.InputRequest): string`
+`client.setActiveRequest(params: {chainId:number, requester: string; identifier: string, timestamp: number, ancillaryData: string }): string`
 
 ### approveCollateral
 
@@ -198,6 +201,12 @@ If the user is on the wrong chain to interact with request, call this to get the
 This requires the `metadata` option on the config for each chain is set.
 
 `switchOrAddChain(): string`
+
+### setActiveRequestByTransaction
+
+Tell the client what request you want to interact with by specifying the transaction hash, chainId and optional eventIndex.
+
+`client.setActiveRequestByTransaction(params: { chainId: number; transactionHash: string; eventIndex?: number }): string`
 
 ## Types
 
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -4,7 +4,7 @@ import Store, { Emit } from "./store";
 import type { state } from "./types";
 import { InputRequest, User } from "./types/state";
 import { Update } from "./services/update";
-import { StateMachine } from "./services/statemachines/statemachine";
+import { StateMachine, setActiveRequestByTransaction } from "./services/statemachines";
 import { loop } from "../utils";
 import { toWei } from "../across/utils";
 import { defaultConfig } from "./utils";
@@ -22,6 +22,9 @@ export class Client {
   setActiveRequest(params: InputRequest): string {
     const requester = ethers.utils.getAddress(params.requester);
     return this.sm.types.setActiveRequest.create({ ...params, requester });
+  }
+  setActiveRequestByTransaction(params: setActiveRequestByTransaction.Params): string {
+    return this.sm.types.setActiveRequestByTransaction.create(params);
   }
   approveCollateral(): string {
     const { checkTxIntervalSec } = this.store.read().chainConfig();
--- a/packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
+++ b/packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
@@ -1,7 +1,6 @@
 import assert from "assert";
 import { Update } from "../update";
 import Store from "../../store";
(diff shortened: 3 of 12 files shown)


packages/sdk/src/oracle/client.ts
```diff
@@ -23,6 +23,9 @@ export class Client {
     const requester = ethers.utils.getAddress(params.requester);
     return this.sm.types.setActiveRequest.create({ ...params, requester });
   }
+  setActiveRequestByTransaction(params: setActiveRequestByTransaction.Params): string {
+    return this.sm.types.setActiveRequestByTransaction.create(params);
+  }
```
daywiss, line 28: this is the new function to call if you want to lookup by transaction from the UI

packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
```diff
@@ -0,0 +1,84 @@
...
+      // we can parse out the necessary params to kick off fetching the state of the request
+      const requestInput = {
+        timestamp: log.args.timestamp,
+        requester: log.args.requester,
+        ancillaryData: log.args.ancillaryData,
+        identifier: log.args.identifier,
+        chainId,
+      };
+
+      store.write((write) => write.inputs().request(requestInput));
```
daywiss, line 64: everything from here down is identical to the `setActiveRequest` command

packages/sdk/src/oracle/tests/client.e2e.ts
```diff
@@ -89,4 +89,36 @@ describe("Oracle Client", function () {
     assert.ok(result[types.state.Flag.RequestSettled]);
     assert.ok(result[types.state.Flag.InsufficientApproval]);
   });
+  test("setActiveRequestByTransaction", async function (done) {
+    const transactionHash = "0x91720719f4768e10849ebb5f41690488f7060e10534c5c4f15e69b7dc494502a";
```
daywiss, line 93: this is one of the first requests on mainnet

packages/sdk/src/oracle/README.md
```diff
@@ -199,6 +202,12 @@ This requires the `metadata` option on the config for each chain is set.
 
 `switchOrAddChain(): string`
 
+### setActiveRequestByTransaction
+
+Tell the client what request you want to interact with by specifying the transaction hash, chainId and optional eventIndex.
+
+`client.setActiveRequestByTransaction(params: { chainId: number; transactionHash: string; eventIndex?: number }): string`
```
daywiss, line 209: docs update to show how to use this

packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
```diff
@@ -0,0 +1,84 @@
...
+      const { chainId, transactionHash, eventIndex = 0 } = params;
+
+      // have to do all of this to fetch the identifier, ancData, requester and timestamp from the request
+      const provider = store.read().provider(chainId);
+      const receipt = await provider.getTransactionReceipt(transactionHash);
+      const oracleLogs = receipt.logs.reduce((result: LogDescription[], log: Log) => {
+        try {
+          result.push(optimisticOracle.contractInterface.parseLog(log));
+        } catch (err) {
+          // we can ignore this error
```
mrice32, line 27: Prefer not using try-catch for normal/expected branching.

An example of how this could cause issues. If we were to have slightly different versions of the OptimisticOracle (or the SkinnyOptimisticOracle) and we ended up using an interface where the event definitions in our abi here didn't perfectly match the contract on-chain, this would silently fail. We've seen this issue in the past when we've made minor updates to contracts on-chain.

Using try-catch this way is sometimes the _only_ option, but very often there's a way to avoid using errors for branching.

In this case, the `Log` type contains an address field that tells us what contract emitted the event, so we could pre-filter the events we parse to only those emitted by the OptimisticOracle contract that we care about. If any events left after the filtering fail to parse, we know that our abi/address is wrong or something else went wrong in the parsing process.

packages/sdk/src/oracle/types/ethers.ts
```diff
@@ -3,4 +3,8 @@ export type { Signer, BigNumber, BigNumberish, Contract } from "ethers";
 export type { Overrides } from "@ethersproject/contracts";
 export { Provider, JsonRpcSigner, JsonRpcProvider, Web3Provider, FallbackProvider } from "@ethersproject/providers";
 export { TransactionRequest, TransactionReceipt, TransactionResponse } from "@ethersproject/abstract-provider";
+export { LogDescription } from "@ethersproject/abi";
 export type { Event };
+
+// taken from ethers code https://github.com/ethers-io/ethers.js/blob/master/packages/abi/src.ts/interface.ts#L654
+export type Log = { topics: Array<string>; data: string };
```
mrice32, line 10: Is this not importable from ethers?

mrice32 (approved):

packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
```diff
@@ -0,0 +1,84 @@
...
+}
+export function Handlers(store: Store): GenericHandlers<Params, Memory> {
+  const update = new Update(store);
+  return {
+    async start(params: Params, memory: Memory) {
+      const { chainId, transactionHash, eventIndex = 0 } = params;
+
+      // have to do all of this to fetch the identifier, ancData, requester and timestamp from the request
+      const provider = store.read().provider(chainId);
+      const receipt = await provider.getTransactionReceipt(transactionHash);
```
mrice32, line 22: +1 on parsing the logs directly from the transaction rather than doing a log query. This is much more efficient.

mrice32 (approved):

packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
daywiss: Great point. Will update this

packages/sdk/src/oracle/types/ethers.ts
daywiss: If you go to that line, you see its defined in line as a function parameter, so i just assumed it wasnh exported

Gamaranto: not pushing for a change in this instance, but in general a thing I like to do to extract types from libraries is using `Parameters<typeof someFn>` or `ReturnType<typeof someFn>`

daywiss: i think i can try that

daywiss: yea that works

packages/sdk/src/oracle/services/statemachines/setActiveRequestByTransaction.ts
daywiss: updated this, first filters for event originating from oracle address, then maps them with the log decoder, its a lot cleaner this way and we dont need to ignore errors

PR #3822, opened 2022-02-07
feat(sdk): use events to populate transaction/block numbers for state

daywiss:
BREAKING CHANGE!: request type properties can now be undefined

**Motivation**
We want to be able to link to the transactions that requested/proposed/disputed requests.

**Summary**
Updates event decoding to also attach transaction hash data to the active request on an interval. Uses exisitng polling commands to update the request. 

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -21,7 +21,12 @@ export class Client {
   }
   setActiveRequest(params: InputRequest): string {
     const requester = ethers.utils.getAddress(params.requester);
-    return this.sm.types.setActiveRequest.create({ ...params, requester });
+    // these are case and number senstive
+    const ancillaryData = params.ancillaryData.toLowerCase();
+    const identifier = params.identifier.toLowerCase();
+    const chainId = Number(params.chainId);
+    const timestamp = Number(params.timestamp);
+    return this.sm.types.setActiveRequest.create({ requester, ancillaryData, identifier, chainId, timestamp });
   }
   setActiveRequestByTransaction(params: setActiveRequestByTransaction.Params): string {
     return this.sm.types.setActiveRequestByTransaction.create(params);
@@ -35,6 +40,7 @@ export class Client {
     assert(user.address, "requires a user account address");
     assert(user.signer, "requires a user signer");
     assert(user.chainId === inputRequest.chainId, "On wrong chain");
+    assert(request.currency, "Request currency is unknown");
     return this.sm.types.approve.create(
       {
         currency: request.currency,
@@ -58,6 +64,7 @@ export class Client {
     assert(user.address, "requires a user account address");
     assert(user.signer, "requires a user signer");
     assert(user.chainId === inputRequest.chainId, "On wrong chain");
+    assert(request.currency, "Request currency is unknown");
     return this.sm.types.proposePrice.create(
       {
         ...inputRequest,
@@ -79,6 +86,7 @@ export class Client {
     assert(user.address, "requires a user account address");
     assert(user.signer, "requires a user signer");
     assert(user.chainId === inputRequest.chainId, "On wrong chain");
+    assert(request.currency, "Request currency is unknown");
     return this.sm.types.disputePrice.create(
       {
         ...inputRequest,
--- a/packages/sdk/src/oracle/errors.ts
+++ b/packages/sdk/src/oracle/errors.ts
@@ -1,0 +1,26 @@
+import { exists } from "../utils";
+
+// This error should be thrown if an expected value does not exist
+export class ExistenceError extends Error {
+  constructor(message = "") {
+    super(message);
+    this.name = "ExistenceError";
+    // if this isnt included, we cannot use instanceof to check the type
+    Object.setPrototypeOf(this, ExistenceError.prototype);
+  }
+}
+
+// Special assert which checks for existence and throw existence error
+export function assertExists<T>(condition: T, message = ""): asserts condition is NonNullable<T> {
+  if (!exists(condition)) throw new ExistenceError(message);
+}
+
+// Ignore only existence errors. If thrown properly this can be used to convert a non existent value to undefined.
+export function ignoreExistenceError<X extends () => any>(call: X): ReturnType<X> | undefined {
+  try {
+    return call();
+  } catch (err) {
+    if (err instanceof ExistenceError) return undefined;
+    throw err;
+  }
+}
--- a/packages/sdk/src/oracle/services/sortedRequests.ts
+++ b/packages/sdk/src/oracle/services/sortedRequests.ts
@@ -1,16 +1,102 @@
-import { SortedJsMap } from "../../stores";
-import { RequestIndex, RequestIndexes } from "../types/state";
+import sortedIndex from "lodash/sortedIndex";
+import sortedLastIndex from "lodash/sortedLastIndex";
 
-export class SortedRequests {
-  private table = SortedJsMap<string, RequestIndex>();
-  async set(key: string, value: RequestIndex): Promise<void> {
-    await this.table.set(key, value);
+import { exists } from "../../utils";
+import { requestId } from "../utils";
+
+import { RequestIndex, RequestIndexes, InputRequest } from "../types/state";
+
+// this was copied out of store and made to be 1. a class, and 2. sync.  Could not modify old implementation since
+// other services rely on this currently and it causes cascading refactors. This has been copied without modification to logic.
+export class SortedStore<Id, Data> {
+  private ids: Id[] = [];
+  private map = new Map<Id, Data>();
+  private getStart = (id: Id, inclusive = true): number => {
+    if (inclusive) {
+      return sortedIndex(this.ids, id);
+    } else {
+      return sortedLastIndex(this.ids, id);
+    }
+  };
+  private getEnd = (id: Id, inclusive = false): number => {
+    return this.getStart(id, !inclusive);
+  };
+  private del = (id: Id): Data | undefined => {
+    const index = sortedIndex(this.ids, id);
+    this.ids.splice(index, 1);
+    const result = this.map.get(id);
+    this.map.delete(id);
+    return result;
+  };
+  // keeps typescript from complaining that return values may be undefined
+  private getSafe = (id: Id): Data => {
+    const result = this.map.get(id);
+    if (exists(result)) return result;
+    throw new Error("Missing data for index: " + id);
+  };
+  has = (id: Id): boolean => {
+    return this.map.has(id);
+  };
+  set = (id: Id, data: Data): void => {
+    if (this.map.has(id)) {
+      this.map.set(id, data);
+    } else {
+      const index = sortedIndex(this.ids, id);
+      this.ids.splice(index, 0, id);
+      this.map.set(id, data);
+    }
+  };
+  get = (id: Id): Data | undefined => {
+    return this.map.get(id);
+  };
+  values = (): Data[] => {
+    return this.ids.map(this.getSafe);
+  };
+  entries = (): [Id, Data][] => {
+    return this.ids.map((id) => [id, this.getSafe(id)]);
+  };
+  keys = (): Id[] => {
+    return [...this.ids];
+  };
+  clear = (): void => {
+    this.map.clear();
+    this.ids.length = 0;
+  };
+  size = (): number => {
+    return this.ids.length;
+  };
+  delete = (id: Id): void => {
+    if (!this.map.has(id)) return;
+    this.del(id);
+  };
+  // assume [a,b)
+  between = (a: Id, b: Id): Data[] => {
+    const start = this.getStart(a);
+    const end = this.getEnd(b);
+    return this.ids.slice(start, end).map(this.getSafe);
+  };
+  slice = (id: Id, length: number): Data[] => {
+    const start = this.getStart(id);
+    return this.ids.slice(start, start + length).map(this.getSafe);
+  };
+}
+
+export class SortedRequests extends SortedStore<string, RequestIndex> {
+  setByRequest(value: RequestIndex): void {
+    return this.set(this.id(value), value);
   }
-  async descending(): Promise<RequestIndexes> {
+  descending(): RequestIndexes {
     // sadly you cannot control lodash sorting descending, so reverse is necessary
-    return (await this.table.values()).reverse();
+    return this.values().reverse();
   }
-  async ascending(): Promise<RequestIndexes> {
-    return this.table.values();
+  ascending(): RequestIndexes {
+    return this.values();
+  }
+  getByRequest(request: InputRequest): RequestIndex {
+    // always return at least the original query data
+    return this.get(this.id(request)) || request;
+  }
+  id(request: InputRequest): string {
+    return requestId(request) + "!" + request.chainId;
   }
 }
--- a/packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
+++ b/packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
@@ -42,11 +42,10 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
         // reprocess all known events and create a table of requests from it
         await update.sortedRequests(chainId);
 
-        try {
-          // we can just try to update the current active request, we dont care if it fails, active request might not be set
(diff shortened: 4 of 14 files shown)


packages/sdk/src/oracle/client.ts
```diff
@@ -21,7 +21,12 @@ export class Client {
   }
   setActiveRequest(params: InputRequest): string {
     const requester = ethers.utils.getAddress(params.requester);
-    return this.sm.types.setActiveRequest.create({ ...params, requester });
+    // these are case and number senstive
+    const ancillaryData = params.ancillaryData.toLowerCase();
+    const identifier = params.identifier.toLowerCase();
+    const chainId = Number(params.chainId);
+    const timestamp = Number(params.timestamp);
+    return this.sm.types.setActiveRequest.create({ requester, ancillaryData, identifier, chainId, timestamp });
```
daywiss, line 29: needed to groom the inputs in order to do consistent finds and filters

packages/sdk/src/oracle/services/sortedRequests.ts
```diff
@@ -1,16 +1,22 @@
-import { SortedJsMap } from "../../stores";
-import { RequestIndex, RequestIndexes } from "../types/state";
+import { RequestIndex, RequestIndexes, InputRequest } from "../types/state";
+import { requestId, SortedStore } from "../utils";
```
daywiss, line 2: added a custom store, because i wanted sync functions

packages/sdk/src/oracle/services/sortedRequests.ts
```diff
@@ -1,16 +1,22 @@
...
 
-export class SortedRequests {
-  private table = SortedJsMap<string, RequestIndex>();
-  async set(key: string, value: RequestIndex): Promise<void> {
-    await this.table.set(key, value);
+export class SortedRequests extends SortedStore<string, RequestIndex> {
```
daywiss, line 4: custom store is a class now

packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
```diff
@@ -41,6 +41,14 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
         await update.oracleEvents(chainId, currentStart, currentEnd);
         // reprocess all known events and create a table of requests from it
         await update.sortedRequests(chainId);
+
+        try {
+          // we can just try to update the current active request, we dont care if it fails, active request might not be set
+          await update.activeRequestFromEvents();
+        } catch (err) {
+          // do nothing
+        }
```
daywiss, line 50: @mrice32 i know you comment on this a lot, sorry to keep adding more of these, but i dont see a better way since in this case its ok to ignore the error, it should only throw if there is no active request ( user is on index page)

packages/sdk/src/oracle/services/update.ts
```diff
@@ -21,20 +21,26 @@ export class Update {
...
       // create the erc20 service to handle currency
-      write.services(chainId).erc20s(fullRequest.currency);
+      write.services(chainId).erc20s(contractRequest.currency);
       write
         .chains(chainId)
         .optimisticOracle()
-        .request(request, { ...fullRequest, state });
+        // update request object with all the data we have about it. order is important,
+        // we want to prioritize latest state pulled from contract.
+        .request({ ...requestIndexData, ...request, ...contractRequest, state });
```
daywiss, line 43: this function gets called periodically and is responsible for the bulk of updating the active request

packages/sdk/src/oracle/services/update.ts
```diff
@@ -98,22 +105,59 @@ export class Update {
...
+      requestBlockNumber,
+      proposeBlockNumber,
+      disputeBlockNumber,
+      settleBlockNumber,
+    };
+
+    this.write((write) => {
+      write.chains(chainId).optimisticOracle().request(update);
+    });
+  }
```
daywiss, line 162: this new update function just grabs the data we need from events, ie transaction hashes and blockNumbers since we have them

packages/sdk/src/oracle/types/state.ts
```diff
@@ -140,10 +140,20 @@ export type RequestIndex = RequestByEvent & { chainId: number };
...
+// combine all request data into a mega request object
+export type FullRequest = InputRequest &
+  // we cant assume any of these props will exist if we get event before query contract
+  Partial<Request> &
+  // take the more specific types from the Request type by omitting overlapping properties
+  Omit<
+    RequestIndex,
+    "proposedPrice" | "resolvedPrice" | "expirationTime" | "reward" | "finalFee" | "bond" | "customLiveness"
+  >;
+
```
daywiss, line 152: this new type is a mash of several different types, because we need to join data from several sources: 1. users requested input, 2. request state from contract and 3. known events for the request.  All of these will have slightly different properties.  Because this data can be set at any time, ie events can come in before contract query, or vice versa, we cant make strong assurances about the existence of any properties.

packages/sdk/src/oracle/utils.ts
```diff
@@ -321,3 +327,78 @@ export function insertOrderedAscending<T>(array: T[], element: T, orderBy: (elem
...
+  between = (a: Id, b: Id): Data[] => {
+    const start = this.getStart(a);
+    const end = this.getEnd(b);
+    return this.ids.slice(start, end).map(this.getSafe);
+  };
+  slice = (id: Id, length: number): Data[] => {
+    const start = this.getStart(id);
+    return this.ids.slice(start, start + length).map(this.getSafe);
+  };
+}
```
daywiss, line 404: this is literally copied and translated from `/stores/sorted-js-map/store.ts`. Perhaps this can eventually replace the original, but previously ran into issues when converting that to a class, it would have cascading effects on the rest of the code in the sdk and api. So decided to move it locally to this client for now.

packages/sdk/src/oracle/services/sortedRequests.ts
```diff
@@ -1,16 +1,102 @@
...
+  between = (a: Id, b: Id): Data[] => {
+    const start = this.getStart(a);
+    const end = this.getEnd(b);
+    return this.ids.slice(start, end).map(this.getSafe);
+  };
+  slice = (id: Id, length: number): Data[] => {
+    const start = this.getStart(id);
+    return this.ids.slice(start, start + length).map(this.getSafe);
+  };
+}
```
daywiss, line 82: this is literally copied and translated from /stores/sorted-js-map/store.ts. Perhaps this can eventually replace the original, but previously ran into issues when converting that to a class, it would have cascading effects on the rest of the code in the sdk and api. So decided to move it locally to this client for now.

packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
mrice32: If you expect it in the "happy path" then it shouldn't be an error :). Why can't we return a value? that indicates no active request? If all else fails, can we throw a _particular type of error_ that we can detect and ignore, so all other errors get passed through?

I think, in general, the architecture should allow for callers to either a) inspect enough to determine whether the method is callable in the current state or b) not error when called in that state and, even bettter, return a value that indicates a no-op so the caller gets feedback.

Do you think this is happening a lot because of the architecture of side-effect-only or write-only callers who can only induce actions but have no ability to see state or get feedback from methods via return values? I think if we want to use that sort of architecture, it might require no-throw code that doesn't produce errors, otherwise, you have to continually protect against edge cases by ignoring errors, burying real, problematic bugs and errors.

mrice32 (commented):
Looks good! I think we need to avoid ignoring errors, even if it means modifying the underlying methods.

packages/sdk/src/oracle/services/statemachines/fetchPastEvents.ts
daywiss: i think you are touching on something fundamental here. you suggest a couple things, i think the one that makes the most sense to me is to check first if we can do something before trying it.  I've been using errors as a shorthand for doing that, with the assumption that if it throws an error it means i cant do it (without having to check prior to calling). As you point out theres room for truly unexpected errors which could be catastrophic that we are missing out on by doing this.  

i'll see about changing this to check for the conditions needed before calling this and see how that goes

packages/sdk/src/oracle/store/has.ts
```diff
@@ -0,0 +1,21 @@
+import Read from "./read";
+import { ignoreError } from "../utils";
```
mrice32, line 2: There's a helper function for ignoring errors?!?

packages/sdk/src/oracle/store/has.ts
```diff
@@ -0,0 +1,21 @@
...
+  private read: Read;
+  constructor(private state: State) {
+    // by design, reads do not mutate or cause side effects
+    this.read = new Read(state);
+  }
+  inputRequest = (): boolean => {
+    // the rule about reads is they will only throw errors if it cant return the value you request.
+    // If no errors are thrown the read was successful.
+    // Reads will never not return a value without throwing an error, enforced by TS.
+    return !!ignoreError(this.read.inputRequest);
```
mrice32, line 16: I think this just moves the problem.

The issue I see:
- There's an underlying bug in the code that's throwing.
- You do the existence check, that bug throws, so you never do the actual read, even though the data was there.
- You have an issue in your code, but it's invisible because the branch is being triggered by an error that wasn't the error you expected.

I think this is roughly the same as just catching and ignoring on the read because the _result_ is the same if there's an unexpected error: you don't do the read because of an error you didn't expect and the code doesn't tell you this unexpected error is being thrown.

daywiss: as per our offline convo, ive added a special type meant for throwing on if a value does not exist. Now the read class only throws these errors, and the `has` class only converts these errors into `false`, while letting the rest of the errors through. wrote unit tests to make sure this works as intended.

packages/sdk/src/oracle/services/statemachines/pollNewEvents.ts
```diff
@@ -30,21 +30,30 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
...
           // this pulls all events from current to latest block
           await update.oracleEvents(chainId, currentBlock, latestBlock);
           // update our request table list with all known events
           await update.sortedRequests(chainId);
+
+          try {
+            // we can just try to update the current active request, we dont care if it fails, active request might not be set
+            await update.activeRequestFromEvents();
+          } catch (err) {
+            // do nothing
```
mrice32, line 45: Can we use the same strategy to drop this ignored error?

packages/sdk/src/oracle/store/has.ts
```diff
@@ -0,0 +1,22 @@
+import Read from "./read";
+// we only ignore this specific error type explicitly emitted from reader class, meaning value is undefined
+import { ignoreExistenceError as ignoreError } from "../errors";
```
mrice32, line 3: nit: can we keep the ignoreExistenceError just to make it clear that we're _only_ ignoring a specific type of error? If I were reading the code and saw ignoreError, I would probably misinterpret what was being done.

packages/sdk/src/oracle/store/has.ts
```diff
@@ -0,0 +1,22 @@
...
+import type { State } from "../types/state";
+
+// This class checks for existence for values you could potentially read. It mirrors the read interface but returns bools from functions.
+export default class Has {
```
mrice32, line 7: I think this has class is totally fine, but if the purpose of this class was to respond to my previous comments, you can feel free to remove. As long as we're catching a _specific type_ of error, I have no problem with how you were originally doing catches. Either way is fine with me, just wanted to mention.

packages/sdk/src/oracle/store/has.ts
mrice32: Awesome, this is great!

mrice32 (commented):
Looks awesome! Only remaining important comment is to use this same strategy for the other ignored catch.

packages/sdk/src/oracle/store/has.ts
```diff
@@ -0,0 +1,22 @@
...
+    this.read = new Read(state);
+  }
+  inputRequest = (): boolean => {
+    // the rule about reads is they will only throw errors if it cant return the value you request.
+    // If no errors are thrown the read was successful.
+    // Reads will never not return a value without throwing an error, enforced by TS.
+    return !!ignoreError(this.read.inputRequest);
+  };
+  sortedRequestsService = (): boolean => {
+    return !!ignoreError(this.read.sortedRequestsService);
```
mrice32, line 20: nit: something to watch out for in this is that you're not just checking for undefined, you're checking for _any_ falsy value, which can capture quite a few reasonable return values for arbitrary functions:
```
> !!"" || !!null || !!undefined || !!0 || !!-0 || !!false
false
```

I might suggest either:
```
ignoreError(...) !== undefined
```

Or changing it to `didThrowExistenceError(...)` and having it explicitly return true or false.


As a secondary note: I think the ignoreExistenceError function name is a little confusing. The fact that it's returning a value is unclear from the name.

packages/sdk/src/oracle/store/has.ts
daywiss: cool, yea i will keep the class as i think it will have utility outside of this. i was just avoiding having to write it and this just gave me the motivation to do it

packages/sdk/src/oracle/store/has.ts
daywiss: yea can do that

packages/sdk/src/oracle/store/read.ts
```diff
@@ -1,24 +1,25 @@
...
+  RequestIndex,
   OptimisticOracleEvent,
+  FullRequest,
 } from "../types/state";
 import type { JsonRpcSigner, BigNumber, Provider } from "../types/ethers";
 import { TransactionConfirmer, requestId } from "../utils";
 import { OptimisticOracle } from "../services/optimisticOracle";
 import { Erc20 } from "../services/erc20";
 import { SortedRequests } from "../services/sortedRequests";
+import { assertExists as assert } from "../errors";
```
mrice32, line 22: Similar to below, I think renaming this is very confusing for the reader, as the normal assert does something different.

packages/sdk/src/oracle/errors.ts
```diff
@@ -0,0 +1,26 @@
...
+  constructor(message = "") {
+    super(message);
+    this.name = "ExistenceError";
+    // if this isnt included, we cannot use instanceof to check the type
+    Object.setPrototypeOf(this, ExistenceError.prototype);
+  }
+}
+
+// Special assert which checks for existence and throw existence error
+export function assertExists<T>(condition: T, message = ""): asserts condition is NonNullable<T> {
```
mrice32, line 14: This return type syntax is cool!

packages/sdk/src/oracle/store/has.ts
daywiss: the existence assertion ` assertExists` checks for null or undefined using ` if (!exists(condition)) `, `exists`does this check under the hood: ` return value !== null && value !== undefined; `.  

> As a secondary note: I think the ignoreExistenceError function name is a little confusing. The fact that it's returning a value is unclear from the name.

hmm yea im not sure what to call it. it converts error to undefined or passes the result of the call back

packages/sdk/src/oracle/errors.ts
daywiss: yea i had to do some research to figure this out!

packages/sdk/src/oracle/services/statemachines/pollNewEvents.ts
daywiss: im going to make a card to refactor specifically for this change, as i need to go through and change it in many places: https://app.shortcut.com/uma-project/story/4475/replace-ignored-errors-with-ignore-existence-errors-in-oo-client-sdk

packages/sdk/src/oracle/store/has.ts
mrice32: Are 0 values allowed as return values? What about booleans? Empty strings? All of these are falsy, so you may end up with false-positives (meaning you'll think there is no value when there really is one).

daywiss: sorry i see what you are saying now, yes i need to change this, thanks

mrice32 (approved):
LGTM!

PR #3823, opened 2022-02-07
feat(slack-config): Add better logging and defaults on non-profitable relay

chrismaree:
Signed-off-by: chrismaree <christopher.maree@gmail.com>

**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
+++ b/packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
@@ -100,11 +100,14 @@ export class ProfitabilityCalculator {
             ? toBNWei("1")
             : toBNWei(priceResponse.value[1]);
       } else {
-        this.l1TokenInfo[this.l1Tokens[index]].tokenEthPrice = toBNWei("0");
+        // Set the price to something very high. This means that the bot will default to continue sending transactions
+        // even if it cant find a price for the given l1Token.
+        this.l1TokenInfo[this.l1Tokens[index]].tokenEthPrice = toBNWei("100000");
         this.logger.warn({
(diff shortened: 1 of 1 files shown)


packages/insured-bridge-relayer/src/ProfitabilityCalculator.ts
```diff
@@ -100,11 +100,14 @@ export class ProfitabilityCalculator {
             ? toBNWei("1")
             : toBNWei(priceResponse.value[1]);
       } else {
-        this.l1TokenInfo[this.l1Tokens[index]].tokenEthPrice = toBNWei("0");
+        // Set the price to something very high. This means that the bot will default to continue sending transactions
+        // even if it cant find a price for the given l1Token.
+        this.l1TokenInfo[this.l1Tokens[index]].tokenEthPrice = toBNWei("100000");
```
nicholaspai, line 105: Maybe set this to MAX_UINT_VAL to be safe

nicholaspai: This makes me wonder whether we should have a special case where price can't be found, or is negative, to signal to bot to skip without throwing

nicholaspai: let's use MAX_UINT_VAL then because i could imagine for some tokens with strange unit economics (XMON one day) that 100,000 ETH is not out of reality

nicholaspai (approved):
LGTM just one nit

mrice32 (approved):
LGTM!

PR #3825, opened 2022-02-07
feat(sdk): add oo settle ability

daywiss:

**Motivation**

Want to be able to settle requests in oo UI.

**Summary**

Copies an existing command and just calls settle instead.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/services/statemachines/settle.ts
+++ b/packages/sdk/src/oracle/services/statemachines/settle.ts
@@ -1,0 +1,52 @@
+import assert from "assert";
+import { Update } from "../update";
+import Store from "../../store";
+import { Signer } from "../../types/ethers";
+import { Handlers as GenericHandlers } from "../../types/statemachine";
+import { InputRequest } from "../../types/state";
+import { ContextClient } from "./utils";
+
+export type Params = InputRequest & {
+  signer: Signer;
+  confirmations: number;
+  currency: string;
+  account: string;
+  checkTxIntervalSec: number;
+};
+
+export type Memory = { hash?: string };
+
+export function initMemory(): Memory {
+  return {};
+}
+
+export function Handlers(store: Store): GenericHandlers<Params, Memory> {
+  const update = new Update(store);
+  return {
+    async start(params: Params, memory: Memory) {
+      const { requester, identifier, timestamp, ancillaryData, chainId, signer } = params;
+      assert(chainId === (await signer.getChainId()), "Signer on wrong chainid");
+
+      const oracle = store.read().oracleService(chainId);
+      const tx = await oracle.settle(signer, requester, identifier, timestamp, ancillaryData);
+      memory.hash = tx.hash;
+      return "confirm";
+    },
+    async confirm(params: Params, memory: Memory, context: ContextClient) {
+      const { chainId, confirmations, checkTxIntervalSec } = params;
+      const { hash } = memory;
+      assert(hash, "requires hash");
+      if (await update.isConfirmed(chainId, hash, confirmations)) {
+        return "update";
+      }
+      // wait x seconds before running this state again
+      return context.sleep(checkTxIntervalSec * 1000);
+    },
+    async update(params: Params) {
+      const { chainId, currency, account } = params;
+      await update.balance(chainId, currency, account);
+      await update.request(params);
+      return "done";
+    },
+  };
+}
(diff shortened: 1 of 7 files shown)


packages/sdk/src/oracle/services/statemachines/settle.ts
```diff
@@ -0,0 +1,52 @@
...
+      const tx = await oracle.settle(signer, requester, identifier, timestamp, ancillaryData);
+      memory.hash = tx.hash;
+      return "confirm";
+    },
+    async confirm(params: Params, memory: Memory, context: ContextClient) {
+      const { chainId, confirmations, checkTxIntervalSec } = params;
+      const { hash } = memory;
+      assert(hash, "requires hash");
+      if (await update.isConfirmed(chainId, hash, confirmations)) {
+        return "update";
```
mrice32, line 40: nit: I think these should be an enum type rather than hardcoded strings to avoid typos and provide context about the range of acceptable return values.

packages/sdk/src/oracle/services/statemachines/settle.ts
```diff
@@ -0,0 +1,52 @@
...
+      }
+      // wait x seconds before running this state again
+      return context.sleep(checkTxIntervalSec * 1000);
```
mrice32, line 43: nit: magic number? Do you think we should make these configurable, have them passed down from a central index file, or have them defined as constants somewhere?

mrice32 (approved):

packages/sdk/src/oracle/services/statemachines/settle.ts
daywiss: yea that could work. i think maybe a broader refactor might be warranted on the handlers.  i will try it first with this one

packages/sdk/src/oracle/services/statemachines/settle.ts
mrice32: oh wow, I just totally misread this. This is _not_ a magic number haha. Sorry about that.

packages/sdk/src/oracle/services/statemachines/settle.ts
mrice32: Why wouldn't a single enum work that could be used for all of them?

If they're all slightly different, then maybe just an `|` type?

mrice32 (approved):

packages/sdk/src/oracle/services/statemachines/settle.ts
daywiss: > Why wouldn't a single enum work that could be used for all of them?
> If they're all slightly different, then maybe just an | type?

I'm not sure i fully understand the suggestion, but it sounds like you define a top level type which represents all possible states of all the handlers by using a union? Or maybe define and export in this file?

I think there might actually be a more automatic way to do this, but im not 100% sure it would work.  The strings that can be returned from these handlers must be of type `keyof ReturnType<typeof Handlers>` ( the strings returned must be able to be keyed into the handler object).  It would be cool if we could enforce that rule automatically without having to define extra types.

daywiss: im going to defer making changes here until we can do a more broad refactor on these handler classes

mrice32: > I think there might actually be a more automatic way to do this, but im not 100% sure it would work. The strings that can be returned from these handlers must be of type keyof ReturnType<typeof Handlers>

I actually think making explicit definitions is helpful, so a reader/user can clearly see what the different states are along with comments about what each state means. "update" and "confirm" are hard to understand with no other context, especially since most of these methods only appear to ever return one string? So there's no possibility of other statuses.

mrice32 (approved):

packages/sdk/src/oracle/services/statemachines/settle.ts
daywiss: i see, i guess i dont like to have all the redundancy, it means you have to change things in multiple places which could have been done automatically. 

I have played around with this and can use enums for states, but i dont have a way to require the developeradd them, it becomes more of a convention they need to remember to do, rather than something required to pass compilation.  i think this requires more thought beyond the scope of this pr.  ill make an issue for it.  https://app.shortcut.com/uma-project/story/4477/propose-new-design-for-state-machine-definitions

PR #3826, opened 2022-02-09
improve(sdk): refactor logic to use existence errors

daywiss:

**Motivation**
We introduced specific error type and handling around that type for unavailable state values during a read. We wanted to adopt this new pattern across the entire oracle client code.

**Summary**

Removed the old non specific ignoreError function, and replaced it with ignoreExistenceError. Also refactored and removed try/catches when possible to use ignoreExistenceError instead.  Founds a small bug with having multiple state machines unrelated to the error issue and fixed that while i was at it.

**Testing**

- [x]  New unit tests created
- [ ]  Untested


packages/sdk/src/oracle/client.ts
```diff
@@ -132,12 +137,15 @@ export class Client {
...
-  startInterval(delayMs = 10): void {
+  startInterval(delayMs = 1): void {
     assert(!this.intervalStarted, "Interval already started, try stopping first");
     this.intervalStarted = true;
     loop(async () => {
       assert(this.intervalStarted, "Interval Stopped");
+      // it turns out since these 2 state machines share the same immer state, they need to be run serially and
+      // cant be run concurrently or you get wierd state oscillations. For now keep them in the same timing loop.
       await this.sm.tick();
+      await this.poller.tick();
```
daywiss, line 148: state machines are in charge writes to global state, and there was a bug due to inconsistent state across concurrent writes, so this had to be moved in series to prevent the issue. bug unrelated to pr.

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
```diff
@@ -2,34 +2,32 @@ import { Update } from "../update";
...
+      // this will rerun when user is set.
+      await ignoreExistenceErrorAsync(update.oracle);
+
+      // get current time of chain when switching request
+      await ignoreExistenceErrorAsync(update.currentTime);
+      await ignoreExistenceErrorAsync(update.request);
+      await ignoreExistenceErrorAsync(update.collateralProps);
+      // order is important, these should be last because they depend on user being set
+      await ignoreExistenceErrorAsync(update.userCollateralBalance);
+      await ignoreExistenceErrorAsync(update.oracleAllowance);
```
daywiss, line 29: this is nice because we can actually try all the calls if they only fail due to an existence error, before the order mattered, as an early error would kick out and prevent the rest from trying

packages/sdk/src/oracle/services/update.ts
```diff
@@ -159,5 +159,5 @@ export class Update {
     this.write((write) => {
       write.chains(chainId).optimisticOracle().request(update);
     });
-  }
+  };
```
daywiss, line 162: all these calls were refactored so we could pass them safely into other functions without worrying about `this`

packages/sdk/src/oracle/services/statemachines/setActiveRequest.ts
mrice32: Cool!

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/errors.ts
```diff
@@ -24,3 +24,15 @@ export function ignoreExistenceError<X extends () => any>(call: X): ReturnType<X
     throw err;
   }
 }
+
+// same function but for async calls
+export async function ignoreExistenceErrorAsync<X extends () => any | Promise<any>>(
```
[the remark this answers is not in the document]
daywiss: yea thats a good point, ill play around with it a little more

PR #3829, opened 2022-02-11
feat!: add setting to allow OO proposer to settle anyone's proposals

mrice32:
**Motivation**

The OptimisticOracle bot should be able to optionally settle _all_ proposals.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
+++ b/packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
@@ -121,27 +121,31 @@ export class OptimisticOracleClient {
 
   // Returns an array of expired Price Proposals that can be settled and that involved
   // the caller as the proposer
-  public getSettleableProposals(caller: string): ProposePriceReturnValues[] {
+  public getSettleableProposals(callers?: string[]): ProposePriceReturnValues[] {
+    if (!callers) return this.expiredProposals;
+
     if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
       return (this.expiredProposals as SkinnyProposePrice["returnValues"][]).filter((event) => {
-        return ((event.request as unknown) as SkinnyRequest).proposer === caller;
+        return callers.includes(((event.request as unknown) as SkinnyRequest).proposer);
       });
     } else {
       return ((this.expiredProposals as unknown) as ProposePrice["returnValues"][]).filter((event) => {
-        return event.proposer === caller;
+        return callers.includes(event.proposer);
       });
     }
   }
 
   // Returns disputes that can be settled and that involved the caller as the disputer
-  public getSettleableDisputes(caller: string): DisputePriceReturnValues[] {
+  public getSettleableDisputes(callers?: string[]): DisputePriceReturnValues[] {
+    if (!callers) return this.settleableDisputes;
+
     if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
       return (this.settleableDisputes as SkinnyDisputePrice["returnValues"][]).filter((event) => {
-        return ((event.request as unknown) as SkinnyRequest).disputer === caller;
+        return callers.includes(((event.request as unknown) as SkinnyRequest).disputer);
       });
     } else {
       return (this.settleableDisputes as DisputePrice["returnValues"][]).filter((event) => {
-        return event.disputer === caller;
+        return callers.includes(event.disputer);
       });
     }
   }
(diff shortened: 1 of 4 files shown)


packages/optimistic-oracle/src/proposer.js
```diff
@@ -89,6 +89,14 @@ class OptimisticOracleProposer {
           return Array.isArray(x) && x.every((value) => typeof value === "string");
         },
       },
+      settleAllRequests: {
+        // `otherAccountsToSettle`: array of other accounts whose requests should be settled automatically.
+        value: [],
```
chrismaree, line 94: this seems a bit strange. if this is a bool then why is this default an array? does the comment on line 93 make sence here?

mrice32: Wow, nice catch. This is a mistake.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -141,14 +149,11 @@ class OptimisticOracleProposer {
...
 
-    const settleableAccounts = [this.account, ...this.otherAccountsToSettle];
-    const settleableRequests = settleableAccounts
-      .map((account) =>
-        this.optimisticOracleClient
-          .getSettleableProposals(account)
-          .concat(this.optimisticOracleClient.getSettleableDisputes(account))
-      )
-      .flat();
+    // If settleAllRequests is true, pass undefined, so it settles _any_ expired requests.
```
chrismaree, line 152: this is such a JS way of doing something by overloading the undefined case. TX would not have been happy with you here haha.

packages/financial-templates-lib/src/clients/OptimisticOracleClient.ts
```diff
@@ -121,27 +121,31 @@ export class OptimisticOracleClient {
...
   // Returns an array of expired Price Proposals that can be settled and that involved
   // the caller as the proposer
-  public getSettleableProposals(caller: string): ProposePriceReturnValues[] {
+  public getSettleableProposals(callers?: string[]): ProposePriceReturnValues[] {
+    if (!callers) return this.expiredProposals;
+
     if (this.oracleType === OptimisticOracleType.SkinnyOptimisticOracle) {
       return (this.expiredProposals as SkinnyProposePrice["returnValues"][]).filter((event) => {
-        return ((event.request as unknown) as SkinnyRequest).proposer === caller;
+        return callers.includes(((event.request as unknown) as SkinnyRequest).proposer);
```
chrismaree, line 129: I wish we did not have to do this two layered casting

mrice32: Yeah :/. I was hoping to clean this up, but I think it kinda has to be this way.

packages/optimistic-oracle/src/proposer.js
mrice32: Yeeeep. I do wish JS had true overloading for stuff like this. Would make things a lot easier.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -89,6 +89,14 @@ class OptimisticOracleProposer {
           return Array.isArray(x) && x.every((value) => typeof value === "string");
         },
       },
+      settleAllRequests: {
+        // `otherAccountsToSettle`: array of other accounts whose requests should be settled automatically.
+        value: false,
```
chrismaree, line 94: Thanks for fixing this. looks much better now.

packages/optimistic-oracle/src/proposer.js
mrice32: In compiled languages, it would just swap out the implementation at compile-time depending on the type being passed in.

packages/optimistic-oracle/src/proposer.js
```diff
@@ -89,6 +89,14 @@ class OptimisticOracleProposer {
...
+        isValid: (x) => {
+          // Must be a string array.
```
chrismaree, line 96: still wrong comment I think.

chrismaree (approved):
LGTM! thanks for fixing the nits. there is one remaining comment but other than that we good to go!

packages/optimistic-oracle/src/proposer.js
mrice32: Yeah, this is why I should actually run the tests before pushing :).

packages/optimistic-oracle/src/proposer.js
mrice32: Yep.

mrice32: Fixed.

PR #3830, opened 2022-02-14
feat(common): implement rounding utils

Reinis-FRP:
**Motivation**

Price feed implementations and proposer/dispute bots need to use correct rounding method as specified in price identifier UMIP.

**Summary**

This adds utility function that can be used by price feeds to round price values.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/common/src/RoundingUtils.ts
+++ b/packages/common/src/RoundingUtils.ts
@@ -4,16 +4,27 @@ import { BN } from "./types";
 
 const { toBN } = Web3.utils;
 
-// Takes inputAmount scaled up to inputDecimals, scales it down, rounds to roundingPrecision decimals
-// and returns the rounded amount scaled up to the same inputDecimals.
-// If roundingPrecision is negative then round to the nearest 10 to the power of absolute roundingPrecision value.
-export function roundToDecimal(inputAmount: BN, inputDecimals: number, roundingPrecision: number): BN {
+/**
+ * @title Utility function for rounding BN values.
+ * @param {BN} inputAmount that is scaled up.
+ * @param {Number} inputDecimals specifies decimal scaling for processed value.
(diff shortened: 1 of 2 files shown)


packages/common/src/RoundingUtils.ts
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * @title Utility function for rounding BN values.
+ * @param {BN} inputAmount that is scaled up.
+ * @param {Number} inputDecimals specifies decimal scaling for processed value.
+ * @param {Number} roundingPrecision specifies amount of decimals left after rounding.
+ *      Negative value would round to the nearest 10 to the power of absolute roundingPrecision value.
+ * @param {RoundingMode} roundingMode is optional rounding mode documented in
+ *      https://mikemcl.github.io/bignumber.js/#constructor-properties
+ */
+export function roundToDecimal(
```
mrice32, line 16: How are we planning to get this into the price feeds? Will we need to manually add it to each?

mrice32 (approved):
Looks great! One high level question

packages/common/src/RoundingUtils.ts
Reinis-FRP: The plan is for each price feed to add `getCurrentRoundedPrice` and `getHistoricalRoundedPrice` methods that would use results from `getCurrentPrice` and `getHistoricalPrice` respectively and round them using `roundToDecimal` with rounding parameters passed from pf config. This would also require adding rounding parameters for default config based on UMIP specs.

The reason why the rounding cannot be implemented directly in `getCurrentPrice` and `getHistoricalPrice` methods is that expression price feeds need to process unrounded results from other price feeds not to loose precision.

Once `getCurrentRoundedPrice` and `getHistoricalRoundedPrice` methods are implemented in price feeds the final step would be to use them in proposer/dispute bots

PR #3838, opened 2022-03-02
fix(sdk): take into account lp transfers into fee calc

daywiss:

**Motivation**

An issues was found with the fee calculation, an account shows a huge fee earned that was not possible.

**Summary**

This bug happens because we previously did not take into account LP token transfers, only withdraws or deposits. This is more of a proposal as to how we can take this into account.

ultimately the fee calculation is updated to this:
`const feesEarned = positionValue.sub(totalDeposited.add(totalTransferredValue));`
Where position value is your current position value from contract - (( sum of deposit events - sum of withdraw events) + (total transferred events to you - total transferred events from you)) 

previously the total transferred was not added into equation.  I'm not totally sure if this captures fees correctly but the results do look in the right ballpark, and it does not change calculations for accounts which have had no transfers.

Bear in mind also, this error is only currently seen on 3 accounts, 2 of which transferred lp tokens to the same account, all of them are related to UMA. this issue does not currently affect third parties, though it could in the future.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/across/clients/bridgePool.ts
+++ b/packages/sdk/src/across/clients/bridgePool.ts
@@ -49,6 +49,7 @@ export type User = {
   address: string;
   poolAddress: string;
   lpTokens: string;
+  lpTransfers: string;
   positionValue: string;
   totalDeposited: string;
   feesEarned: string;
@@ -146,6 +147,7 @@ export class PoolEventState {
       await Promise.all([
         ...(await this.contract.queryFilter(this.contract.filters.LiquidityAdded(), this.startBlock, endBlock)),
         ...(await this.contract.queryFilter(this.contract.filters.LiquidityRemoved(), this.startBlock, endBlock)),
+        ...(await this.contract.queryFilter(this.contract.filters.Transfer(), this.startBlock, endBlock)),
       ])
     )
       .filter(this.filterSeen)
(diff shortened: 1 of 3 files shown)


packages/sdk/src/across/clients/bridgePool.ts
```diff
@@ -234,14 +236,19 @@ function joinUserState(
     .mul(userState.balanceOf)
     .div(fixedPointAdjustment);
   const totalDeposited = BigNumber.from(eventState.tokens[userState.address] || "0");
-  const feesEarned = positionValue.sub(totalDeposited);
+  const totalTransferred = BigNumber.from(eventState.lpTransfers[userState.address] || "0");
+  const totalTransferredValue = BigNumber.from(poolState.exchangeRateCurrent)
+    .mul(totalTransferred)
+    .div(fixedPointAdjustment);
```
mrice32, line 242: I don't think this will work for more complex scenarios since you can redeem LP tokens that were transferred in and you can transfer out tokens that were deposited. This could result in totalDeposited or totalTransferredValue being negative and could likely cause fees to go negative.

I think this breaks because of the faulty underlying assumption that the transferred value is as of "today's value" but the deposit values are valued as of that point in time.

Consider what would happen if I had a deposit 6 months ago and immediately transferred the tokens out. My deposit was worth $100, but since my tokens that I transferred were valued at _today's value_, they would be valued higher, let's say $115. If you subtract these two, you end up getting a negative fee value.

daywiss: yea i think you are right. i just worked through an example. its because the redeem value implicitly uses an older exchange rate, while the transfer value uses a higher rate.

daywiss:
i talked to tom, going to put this pr on hold until we have some time for a better solution

PR #3843, opened 2022-03-07
feat: add zodiac module

pemulis:

**Motivation**

This [Zodiac module](https://gnosis.github.io/zodiac/docs/intro) allows you to control a Gnosis Safe according to a set of off-chain rules. Any address can propose transactions that follow the rules and any address can dispute a proposal to UMA's Optimistic Oracle within a challenge window.

The [UMIP](https://github.com/UMAprotocol/UMIPs/pull/492) has more details on how it works and the rationale.

This module will allow for a new era of flexible "optimistic governance," where management of DAO treasuries and other multi-signature wallets can be managed more effectively, without being limited to X-of-Y signing schemes or Snapshot votes, although the Zodiac module can enforce those things, too.

The module is intended to be used with the `ZODIAC` identifier, currently in [draft state](https://github.com/UMAprotocol/UMIPs/pull/492/files#diff-c0732d176241fe45e7f0ec0d58ec2fd991d47b4aea6db4024a9ce609d6035048).

**Summary**

This PR adds a module contract and installs the Zodiac package in `core`.

**Details**

Since this will be the first time the core engineering team sees this, the module is being submitted prior to testing and without many internal comments, just to start the discussion.

I have successfully deployed the module and connected it to a Gnosis Safe on Rinkeby.

image

Next steps (not necessarily in this order):

- Test proposals and disputes manually on Rinkeby
- Write a contract test suite
- Write the `ZODIAC` identifier UMIP describing how to resolve disputes [FIRST DRAFT DONE]
- Add comments to the contract code
- Refine the contract code and prepare for the next audit

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+    FinderInterface public finder;
+    SkinnyOptimisticOracleInterface optimisticOracle;
+
+    IERC20 public collateral;
+    uint256 public bond;
+    string public rules;
+    uint256 public liveness;
+    bytes32 public identifier = "ZODIAC";
+
+    struct Transaction {
```
auryn-macmillan, line 31: You could save some gas by storing hashes of transactions, rather than the whole transaction object. Similar to what we did for the Reality module.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+
+    event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
+
+    event ProposalDeleted(uint256 indexed proposalId);
+
+    FinderInterface public finder;
+    SkinnyOptimisticOracleInterface optimisticOracle;
+
+    IERC20 public collateral;
+    uint256 public bond;
```
auryn-macmillan, line 26: If `bond` can only be set by the DAO and the final fee changes to a value higher than bond, then could the DAO end up in a locked state where it cannot get an answer from the oracle and cannot update the required bond?

nicholaspai: Yeah great point. @pemulis i think we should save the bond per proposal and then before executing any proposal check if the bond has changed. 

If the bond has changed you can either cancel the proposal or ignore the change and just deal with that risk. But we definitely don't want to get locked up

pemulis: I think that would work. We could also change the parameter saved in the module to `additionalBond` so that the DAO can optionally require an additional bond above and beyond the final fee, and the combined bond is required to be approved and transferred by the user when making a proposal. Will tinker with this a bit.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+    event OptimisticOracleModuleDeployed(address indexed owner, address indexed avatar, address target);
+
+    event TransactionsProposed(uint256 indexed proposalId, address indexed proposer, uint256 indexed proposalTime);
+
+    event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
+
+    event ProposalDeleted(uint256 indexed proposalId);
+
+    FinderInterface public finder;
+    SkinnyOptimisticOracleInterface optimisticOracle;
```
nicholaspai, line 23: nit: why not make this variable `public`

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        // set reference to the rules for the avatar (e.g. an IPFS hash or URI)
+        rules = _rules;
+    }
+
+    function setLiveness(uint256 _liveness) public onlyOwner {
+        // set liveness for disputing proposed transactions
+        liveness = _liveness;
+    }
+
+    function proposeTransactions(Transaction[] memory transactions, bytes memory ancillaryData) public {
```
nicholaspai, line 103: you're going to want to add some reentrancy guards to this public methods that make external calls

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        int256 price = _getOracle().getPrice(identifier, proposal.requestTime, proposal.ancillaryData);
+
+        Transaction memory transaction = proposal.transactions[_transactionIndex];
+
+        require(
+            _transactionIndex == 0 || proposal.transactions[_transactionIndex - 1].to == address(0),
+            "Previous tx not yet executed"
+        );
+        require(transaction.to != address(0), "Tx already executed");
+        require(price != 0, "Proposal was rejected");
```
nicholaspai, line 166: Can we delete a proposal that was rejected to reduce storage costs and get a refund?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        require(msg.value == transaction.value, "Must send exact amount of ETH");
+
+        // Delete the transaction before execution to avoid any potential re-entrancy issues.
+        delete proposal.transactions[_transactionIndex];
+
+        require(
```
nicholaspai, line 172: Again, definitely want a reentrnacy guard here

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+            abi.decode(initializeParams, (address, address, address, uint256, string));
+        finder = FinderInterface(_finder);
+        optimisticOracle = _getOptimisticOracle();
+        // check collateral is whitelisted
+        collateral = IERC20(_collateral);
+        // check bond amount is large enough
+        bond = _bond;
+        rules = _rules;
+        setAvatar(_owner);
+        setTarget(_owner);
```
nicholaspai, line 77: What is a "target"?

nicholaspai (commented):
Just did a first pass of the code, I assume you'll add tests later so didn't comment on that. I'm just trying to understand better the logic.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
auryn-macmillan: - **Avatar:** is the Safe. The Address that will ultimately execute the transaction that is passed by the module.
- **Target:** is the address that this module will call `execTransactionFromModule()` on. In most cases this will be the Avatar, but in some cases it could be a special kind of module called a [modifier](https://github.com/gnosis/zodiac/blob/master/contracts/core/Modifier.sol), that sits between a module and an avatar and... well... modifies the transactions passed to it in some way.
- **Owner:** is the address that has permissions to call `OnlyOwner()` functions on the module.

https://gnosis.github.io/zodiac/docs/tutorial-build-a-module/create_module

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Might as well!

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Yep!

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        // note: optional ancillaryData explains the intent of the transactions to make comprehension easier
+        uint256 id = proposals.length;
+        uint256 time = block.timestamp;
+        address proposer = msg.sender;
+
+        // Add a zero-initialized element to the proposals array.
+        proposals.push();
+
+        // Initialize the new proposal.
+        Proposal storage proposal = proposals[id];
```
mrice32, line 116: Proposal storage proposal = proposals.push();
```suggestion
        Proposal storage proposal = proposals.push();
```

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        delete proposal.transactions[_transactionIndex];
+
+        require(
+            exec(transaction.to, transaction.value, transaction.data, _operation),
+            "Failed to execute the transaction"
+        );
+        emit TransactionExecuted(_proposalId, _transactionIndex);
+    }
+
+    function deleteProposal(uint256 _proposalId) public onlyOwner {
```
mrice32, line 179: Why does this need to be done by the owner? Can we do this trustlessly by grabbing the return value from the OO, seeing it's false and deleting the proposal?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        // delete a proposal that governance decided not to execute
```
mrice32, line 180: I'm a bit confused at this comment. Is this function expected to be used when the OO rejects a proposal? Or is it expected to be used when a proposal passes and _could_ be executed and governance tries to manually remove it before it can be executed?

Or is this supposed to handle proposals where the execution reverts for some reason.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
mrice32: > If the user were to pass in bond as a parameter to proposeTransactions(), wouldn't that solve the issue?

Is that a conflict of interest? Should the proposer be able to pick their own bond?

> i think we should save the bond per proposal and then before executing any proposal check if the bond has changed.

I think the OO does this for us!

To the original question, I think this actually works as it's implemented now. The `bond` value passed to the OO (I think!) is _in addition to_ the final fee so the bond will never end up being lower than the final fee. The caller will be required to approve/send the bond + finalFee to the OO.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+    event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
+
+    event ProposalDeleted(uint256 indexed proposalId);
+
+    FinderInterface public finder;
+    SkinnyOptimisticOracleInterface optimisticOracle;
+
+    IERC20 public collateral;
+    uint256 public bond;
+    string public rules;
```
mrice32, line 27: What does rules do? Is there some interface that requires it or something?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+            int256(1e18)
+        );
+
+        emit TransactionsProposed(id, proposer, time);
+    }
+
+    function executeProposal(
+        uint256 _proposalId,
+        uint256 _transactionIndex,
+        Enum.Operation _operation
```
mrice32, line 151: It seems that we're letting the caller pick whether this should be a normal call or a delegateCall? Shouldn't this be part of the proposal? I would think that that distinction could make a big difference in the outcome of a transaction.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,197 @@
...
+        address _owner,
+        address _collateral,
+        uint256 _bond,
+        string memory _rules
+    ) {
+        bytes memory initializeParams = abi.encode(_finder, _owner, _collateral, _bond, _rules);
+        setUp(initializeParams);
+    }
+
+    function setUp(bytes memory initializeParams) public override initializer {
```
mrice32, line 65: Does this work this way to allow this contract to optionally be used as a library (delegatecalled)? Or is setUp done this way for a different reason? @auryn-macmillan may be the right person to answer this question.

I just want to make sure we understand the usage pattern so we can make informed decisions about how we work with storage.

mrice32 (commented):
Looks awesome! A few comments

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: My intention with the function is to allow governance to manually remove a proposal that has passed before it can be executed but I can make this clearer in a comment.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: This is a reference to the off-chain published rules that the OO should consider when determining if a proposal is valid or not. The UMIP draft has more details. 

https://github.com/UMAprotocol/UMIPs/pull/492

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Good point, this should be part of the proposal rather than the execution.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: The `setUp` function allows you to deploy a module like this at the same time as the Gnosis Safe that will be the module's avatar.

https://gnosis.github.io/zodiac/docs/tutorial-build-a-module/create_module#factory-friendly

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Explained in the next reply but the purpose of this function is to let the owner (the Gnosis Safe) delete proposals that were previously approved but not yet executed.

pemulis: Proposals that were rejected by the OO should be automatically deleted but I'll double-check that.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: From Matt's comment, this may work as-is! I'll double-check how this function interacts with the OO's bond handling logic and add some comments to the code in the module to make it clear that proposers have to post the bond *plus* the OO's final fee for the module's collateral type, which can be found in UMA's contracts and documentation.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Out of curiosity, how much gas are we talking? The trade-off here is that UMA's dev tools (and voter community) are very familiar with these kinds of structs, since they're how the `Governor` contract works, and less familiar with dealing with transaction hashes, so there would be some extra effort to use transaction hashes, which *might* be worth it if the savings are significant.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Here's the line where the optional bond is added to the final fee. Agree with Matt's assessment that this should work as-is and I'll add comments to the module contract that explain this.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Updated this to add operation to the Transaction struct so you can set the operation per transaction.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import "../oracle/implementation/Constants.sol";
+import "../oracle/interfaces/FinderInterface.sol";
+import "../oracle/interfaces/SkinnyOptimisticOracleInterface.sol";
+import "../oracle/interfaces/OracleAncillaryInterface.sol";
+import "../common/implementation/Lockable.sol";
+import "../common/interfaces/AddressWhitelistInterface.sol";
+import "../common/implementation/AncillaryData.sol";
+
```
chrismaree, line 15: if you are going to use NatSpec you should include a `title` comment like this: https://github.com/UMAprotocol/protocol/blob/ceb5a468d6cfb9cf1996fb6ad058ac36d5f7db65/packages/core/contracts/oracle/implementation/Voting.sol#L22

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+    struct Proposal {
+        Transaction[] transactions;
+        uint256 requestTime;
+        bytes ancillaryData;
+        bool status;
+    }
+
+    Proposal[] public proposals;
+
+    /**
```
chrismaree, line 52: this natspec is done pretty inconsistently. If you are going to apply it to the constructor then you should do it on all the other functions, else get rid of it here.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+    function _isContract(address addr) private view returns (bool isContract) {
+        return addr.code.length > 0;
+    }
+
+    function _getCollateralWhitelist() internal view returns (AddressWhitelistInterface) {
+        return AddressWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.CollateralWhitelist));
+    }
+
+    function _sync() internal {
+        skinnyOptimisticOracle = _getOptimisticOracle();
```
chrismaree, line 230: I dont think this needs to be internal seeing you only do it one place. also let's be a bit more descriptive with the function name. perhaps `syncUmaEcosystemContracts`. `sync` like you have it here has implications around balance synchronisations (like in Uniswap) that is not done here.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+        int256 price = _getOracle().getPrice(identifier, proposal.requestTime, proposal.ancillaryData);
+
+        Transaction memory transaction = proposal.transactions[_transactionIndex];
+
+        require(
+            _transactionIndex == 0 || proposal.transactions[_transactionIndex - 1].to == address(0),
+            "Previous tx not yet executed"
+        );
+        require(transaction.to != address(0), "Tx already executed");
+        require(price != 0, "Proposal was rejected");
```
chrismaree, line 178: would it be better to check that `price==1e18`? is there a 0.5e18 case or something else in the UMIP that is not approved but also not rejected? put differently it might be better to check for the exact state you're looking for.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+        // execute transactions in an approved proposal using exec() function
+        Proposal storage proposal = proposals[_proposalId];
+
+        // this will revert if the price has not settled
+        int256 price = _getOracle().getPrice(identifier, proposal.requestTime, proposal.ancillaryData);
+
+        // check that proposal was rejected
+        require(price == 0, "Proposal was not rejected");
+
+        // Delete the proposal
```
chrismaree, line 207: thought this file you are inconsistent with your comments. some are upper case on the first letter, some are lower case. some have fulls stops at the end and others dont. Ideally, all comments should be full sentences, starting with an upper case and ending with a full stop.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
...
+            // If the transaction has any data with it the recipient must be a contract, not an EOA.
+            if (_transactions[i].data.length > 0) {
+                require(_isContract(_transactions[i].to), "EOA can't accept tx with data");
+            }
+            proposal.transactions.push(_transactions[i]);
+        }
+
+        // Propose a set of transactions to the OO. If not disputed, they can be executed with executeProposal().
+        // docs: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
+        _getOptimisticOracle().requestAndProposePriceFor(
```
chrismaree, line 148: where is the bonding logic? do you not need to pull the `bondAmount` from the caller?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,232 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrismaree, line 1: is the intention to write unit tests for this? even if you mock the external contracts it's properly a good idea to do so to validate that the right functions are called! We have a bunch of tooling to do this with [smock](http://smock.readthedocs.io)

pemulis: Yep, I'm working on tests right now.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: I'm going to rewrite all of the comments and make them consistent.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Good point, we can check that the proposal was approved rather than the other way around. The only approved return value is 1e18.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Comments are a WIP, will definitely improve this.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: I think so, will add that and make sure it's tested before requesting a re-review.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
mrice32: I think this is something we can estimate at the back of the envelope (and I think doing this regularly is useful!):

We pay 20k gas to write 32 bytes of data assuming that we're not reusing slots we've already used (which is true in this case).
```
    // 96 bytes + length of data (min 128 bytes)
    struct Transaction {
        address to; // 20 bytes, but will use a full 32 byte slot because not next to any smaller data types
        uint256 value; // 32 bytes
        bytes data; // 32 bytes + length of data (min 32 bytes)
        Enum.Operation operation; // 1 byte, could be moved next to the to address so it gets added to that slot
    }

   // 128+ bytes * num_txns + 128+ bytes
   struct Proposal {
        Transaction[] transactions; 128 bytes min * num_txns
        uint256 requestTime; 32 bytes
        bytes ancillaryData; 32 bytes + length of data (min 64 bytes)
        bool status; // 1 byte, but uses a full 32 byte slot
    }
```

This results in a min 256 bytes for 1 transaction, which would cost about 160k gas. This is the minimum. As ancillary data increases and the transaction data grows, this can get much larger.

If we hashed the _entire_ structure, that would reduce all of this down to a single 32 byte slot, which would cost 20k gas, which is about how much a basic ETH transfer costs.

You could also hash each transaction, but keep the outer struct intact. This would help, but wouldn't reduce gas costs nearly as much since each transaction would add about 32k gas.

So I think the gas savings are considerable.

The big disadvantage to doing this is that this would require callers who want to do anything with this data to pass the data in, so it can be verified against the hash. This makes the UX of interacting with the contract more painful, and might mean that users _have_ to rely on external tooling to make use of these contracts.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Digging around for smock specifically, can you show where we use that so I can see examples?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,271 @@
...
+    }
+
+    function deleteProposal(uint256 _proposalId) public onlyOwner {
+        // Delete a proposal that governance decided not to execute.
+        delete proposals[_proposalId];
+        emit ProposalDeleted(_proposalId);
+    }
+
+    function deleteRejectedProposal(uint256 _proposalId) public {
+        // Execute transactions in an approved proposal using exec() function.
```
Reinis-FRP, line 226: This comment is misleading here as deleting the proposal should not execute its transactions

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: I think a gas efficient version of the module contract would be great in the future but I have data availability concerns. For this initial release, I'm going to stick with putting the full transactions on-chain to ensure that DVM voters can always see them if there is a dispute.

Over time, I think we'll get more tooling and interfaces for interacting with the module and best practices that will make data availability less of a concern. For instance, people will want to ensure that explanations and discussions around proposals are always available. It would make sense to keep the full details of the transactions in the same place.

Then if you had a second version of the module that uses hashed transactions instead of full transactions, and had good reason to believe that the full transactions will almost always be available somewhere else, you might feel comfortable using that version. (This is already the case if the proposal was created and voted on through Snapshot, once this module is integrated there.)

And if the data turns out to be unavailable, the voters would need to reject the proposal. You could also imagine an avatar approving *both* modules and people mainly using the hashed transaction version while keeping the full transaction version around as a back-up.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,271 @@
...
+        // https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/implementation/Governor.sol
+        // note: Pptional explanation explains the intent of the transactions to make comprehension easier.
+        uint256 id = proposals.length;
+        uint256 time = getCurrentTime();
+        address proposer = msg.sender;
+
+        // Add a zero-initialized element to the proposals array.
+        Proposal storage proposal = proposals.push();
+        proposal.requestTime = time;
+        proposal.ancillaryData = _explanation;
```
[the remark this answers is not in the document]
pemulis: That's a good idea.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Whoops, will delete.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: However, I believe that would limit the `_explanation` to a `bytes32` length since the `AncillaryData` library has an `appendKeyValueBytes32` function but no way to append a key-value with arbitrary length bytes.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: @mrice32 pointed out in a separate thread that we might be able to solve for data availability by emitting the full transaction details in events while only storing the hash on-chain, so I'm backtracking on what I just posted and will give that a shot.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: I'll include that warning in a UMIP update.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: I made some updates to the flow. Now, when making a proposal, the contract uses all of the relevant information to encode a `proposalHash`. To execute the proposal, the same data (including transaction data) is entered by the caller and we verify that it generates the same hash that was stored in the contract and approved by the Optimistic Oracle. Then, we can iterate through the transactions in the calldata to execute them one by one.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,345 @@
...
+        bytes memory proposalData = abi.encodePacked(ancillaryData);
+
+        for (uint256 i = 0; i < _transactions.length; i++) {
+            proposalData = bytes.concat(abi.encodePacked(_transactions[i].to));
+            proposalData = bytes.concat(abi.encodePacked(_transactions[i].value));
+            proposalData = bytes.concat(abi.encodePacked(_transactions[i].data));
+            proposalData = bytes.concat(abi.encodePacked(_transactions[i].operation));
+        }
+
+        proposalHashes[id] = keccak256(abi.encodePacked(proposalData));
```
mrice32, line 203: We probably don't want to hash this way for a few reasons:
1. abi.encodePacked can be ambiguous with multiple dynamic pieces of data. This means that I could technically create different sets of transactions that result in the same hash.
2. You don't need to encode each element individually like this. It's much easier to just pass the entire struct into the abi.encode function.


I would recommend the following process instead:
```
proposalHashes[id] = keccak256(abi.encode(proposal));
```

mrice32 (commented):
A few comments, will add more a bit later today!

Looks awesome, nice work!

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,345 @@
...
+    }
+
+    function priceProposed(
+        bytes32 _identifier,
+        uint32 _timestamp,
+        bytes memory _ancillaryData,
+        SkinnyOptimisticOracleInterface.Request memory _request
+    ) external {
+        emit PriceProposed(_identifier, _timestamp, _ancillaryData, _request);
+    }
```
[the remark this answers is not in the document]
pemulis: Huh, that looks like a stub whose functionality moved elsewhere and should have been deleted. Getting rid of it.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,345 @@
...
+
+        // Create proposal in memory to emit in an event.
+        Proposal memory proposal;
+        proposal.requestTime = time;
+
+        // Construct the ancillary data.
+        bytes memory ancillaryData = _explanation;
+        proposal.explanation = _explanation;
+        AncillaryData.appendKeyValueUint(ancillaryData, "id", id);
+        AncillaryData.appendKeyValueAddress(ancillaryData, "module", address(this));
```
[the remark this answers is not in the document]
pemulis: Nice! And just to confirm, the OO is going to stamp the module contract's address and not `msg.sender`?

packages/core/contracts/zodiac/OptimisticOracleModule.sol
```diff
@@ -0,0 +1,345 @@
...
+        uint256 time = getCurrentTime();
+        address proposer = msg.sender;
+
+        // Create proposal in memory to emit in an event.
+        Proposal memory proposal;
+        proposal.requestTime = time;
+
+        // Construct the ancillary data.
+        bytes memory ancillaryData = _explanation;
+        proposal.explanation = _explanation;
```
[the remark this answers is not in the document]
pemulis: Sure, we could do that. It was added to ancillary data as a convenience to voters but they would need to go to the events to get the full transaction data, anyway.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: Cool, will make that change.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: Hmm, removing it breaks some tests. Will investigate further.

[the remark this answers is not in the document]
pemulis: Ahhh, I remember now. This is a callback function since this contract is an `OptimisticRequester`. It fires off the event with the ancillary data from the Optimistic Oracle after a price is proposed. Will add it back in.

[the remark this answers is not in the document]
mrice32: Sorry, ignore this, this is a repeat.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: Looks like this should work. Rather than encoding the whole proposal, though, I just encode the `_transactions`. This makes it easier to check during the execution step since you don't have to re-build the whole proposal in memory just to compare hashes.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        // Remove proposal hash so transactions can not be executed again.
+        proposalHashes[id] = 0;
+
+        // This will revert if the price has not settled.
+        (, int256 price) =
+            skinnyOptimisticOracle.settle(address(this), identifier, _originalTime, ancillaryData, _request);
+        require(price == 1e18, "Proposal was rejected");
+
+        for (uint256 i = 0; i < _transactions.length; i++) {
+            require(i == 0 || _transactions[i - 1].to == address(0), "Previous tx not yet executed");
```
mrice32, line 245: What's the point of this require? Don't we know that the prev transaction has already been executed?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+            Transaction memory transaction = _transactions[i];
+            require(transaction.to != address(0), "Tx already executed");
+            require(msg.value == transaction.value, "Must send exact amount of ETH");
+
+            // Delete the transaction before execution to avoid any potential re-entrancy issues.
+            delete _transactions[i];
```
mrice32, line 251: I don't think this is required since `_transactions` is in memory, not in storage and, therefore wouldn't be preserved in another call.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        // This will revert if the price has not settled.
+        (, int256 price) =
+            skinnyOptimisticOracle.settle(address(this), identifier, _originalTime, ancillaryData, _request);
+        require(price == 1e18, "Proposal was rejected");
+
+        for (uint256 i = 0; i < _transactions.length; i++) {
+            require(i == 0 || _transactions[i - 1].to == address(0), "Previous tx not yet executed");
+            Transaction memory transaction = _transactions[i];
+            require(transaction.to != address(0), "Tx already executed");
+            require(msg.value == transaction.value, "Must send exact amount of ETH");
```
mrice32, line 248: Since all transactions are being executed in the same call, I don't think this check makes sense, since there could be two calls that each use half the msg.value.

Do you envision a scenario where the proposer would need to send eth through this call?

Side note: can you explain how calls are supposed to happen? Aren't they supposed to be made _through_ the target (gnosis-safe)? This seems to allow _this contract_ to call out directly to other contracts? Is that what we want?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        // Construct the ancillary data.
+        bytes memory ancillaryData = _explanation;
+        AncillaryData.appendKeyValueUint(ancillaryData, "id", id);
+
+        // This will reject the transaction if the proposal hash generated from the inputs does not match the stored proposal hash.
+        // require(proposalHashes[id] == keccak256(abi.encodePacked(proposalData)), "proposal hash does not match");
+        require(proposalHashes[id] == keccak256(abi.encode(_transactions)), "proposal hash does not match");
+
+        // Remove proposal hash so transactions can not be executed again.
+        proposalHashes[id] = 0;
```
mrice32, line 237: nit: prefer `delete proposalHashes[id]`

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        uint256 id = prevProposalId + 1;
+        uint256 time = getCurrentTime();
+        address proposer = msg.sender;
+
+        // Create proposal in memory to emit in an event.
+        Proposal memory proposal;
+        proposal.requestTime = time;
+
+        // Construct the ancillary data.
+        bytes memory ancillaryData = _explanation;
```
mrice32, line 178: Probably better for a follow-up, but it seems a little odd that the caller gets to set most/all of the info in the ancillary data. I would worry about this potentially being misleading to voters, as ancillary data is often set by the contract/protocol and is assumed to be tamper-proof.

Might be worth a quick chat with other folks on the team.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        uint32 _originalTime,
+        bytes memory _ancillaryData,
+        SkinnyOptimisticOracleInterface.Request memory _request
+    ) public {
+        // This will revert if the price has not settled.
+        (, int256 price) =
+            skinnyOptimisticOracle.settle(address(this), identifier, _originalTime, _ancillaryData, _request);
+
+        // Check that proposal was rejected.
+        require(price == 0, "Proposal was not rejected");
```
mrice32, line 278: Should this instead check that it's `!= 1`? I would think that anything that's _not accepted_ is rejected, right?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+        // This will reject the transaction if the proposal hash generated from the inputs does not match the stored proposal hash.
+        // require(proposalHashes[id] == keccak256(abi.encodePacked(proposalData)), "proposal hash does not match");
+        require(proposalHashes[id] == keccak256(abi.encode(_transactions)), "proposal hash does not match");
+
+        // Remove proposal hash so transactions can not be executed again.
+        proposalHashes[id] = 0;
+
+        // This will revert if the price has not settled.
+        (, int256 price) =
+            skinnyOptimisticOracle.settle(address(this), identifier, _originalTime, ancillaryData, _request);
```
mrice32, line 241: This is where the skinny OO gets a little tricky.

If this is already settled, this will revert, meaning a 3rd party could block executions by pre-settling.

@nicholaspai could probably advise better than I can, but my suggestion would be:
1. Check _request.settled.
2. If true, call require(hasPrice()).
3. If false, call settle().

Same goes for the other callsite.

As a side note (that can be left for a follow-up): why did we decide to use the skinny OO here? It has considerable UX tradeoffs when asking 3rd parties to propose. Do we have any concerns around that?

mrice32 (commented):
Looks great! A few remaining comments.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    function sync() public nonReentrant {
+        // Sync the store and oracle contract addresses as well as the final fee.
+        _sync();
+    }
+
+    function proposeTransactions(Transaction[] memory _transactions, bytes memory _explanation) public nonReentrant {
+        // note: Optional explanation explains the intent of the transactions to make comprehension easier.
+        uint256 id = prevProposalId + 1;
```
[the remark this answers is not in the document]
pemulis: Good catch, adding a fix.

packages/core/contracts/zodiac/OptimisticGovernor.sol
pemulis: I think this (and some of the other things you noted) is some residual stuff from the Governor contract that is no longer relevant if the proposal itself is not stored in the contract.

See, for example: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/implementation/Governor.sol#L148

I'll strip this and some of the other checks out. I already added the missing check that is actually relevant, which is that the proposal hash should be deleted after execution.

packages/core/contracts/zodiac/OptimisticGovernor.sol
pemulis: You're right, this step has no purpose if the transactions are not in storage.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
pemulis: Done, except that I encoded the transactions array rather than the full proposal.

packages/core/contracts/zodiac/OptimisticOracleModule.sol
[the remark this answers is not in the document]
pemulis: Yeah, we could do that. This ties in with your other comment (https://github.com/UMAprotocol/protocol/pull/3843#discussion_r852422606) about possibly misleading ancillary data. I'll make the change.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -0,0 +1,313 @@
...
+        uint32 _originalTime,
+        bytes memory _ancillaryData,
+        SkinnyOptimisticOracleInterface.Request memory _request
+    ) public {
+        // This will revert if the price has not settled.
+        (, int256 price) =
+            skinnyOptimisticOracle.settle(address(this), identifier, _originalTime, _ancillaryData, _request);
+
+        // Check that proposal was rejected.
+        require(price != 1, "Proposal was not rejected");
```
mrice32, line 270: Sorry, my suggestion was wrong. This should be != 1e18 I think.

mrice32 (approved):
LGTM!

PR #3848, opened 2022-03-16
fix(discord-transport): Remove extra log messages for links and refine discord config to be more granular on where messages are sent

chrismaree:
**Motivation**

There are a number of features we'd like to add to the Discord transport:
- Use short link format on UI links to reduce log bloat. This should also remove the extra log produced for each of these.
- removing settlement logs from all discord channels
- Only having proposal logs in oo-fact-checking channel
- request and proposal logs in oo-events channel

This PR implements the above goals.

**Summary**

Short links are done by removing the duplicate logic and refining the `_generateUILink` function in the OO monitor.

The discord transport is refined to accomidate a new prop: `discordSettings` which can be used to spesific sub sets of transports.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -34,16 +34,15 @@ export class DiscordTransport extends Transport {
         embeds: [{ title: `${info.message}`, description: this.formatLinks(info.mrkdwn), color: 9696729 }],
       };
 
-      // A log message can conditionally contain additional Discord specific settings. Validate these first.
-
+      // A log message can conditionally contain additional Discord specific paths. Validate these first.
       let webHooks: string[] = [];
-      if (info.discordSettings) {
+      if (info.discordPaths) {
         // If it contains "noPost" then this is a noop message for Discord and the log should be skipped.
-        if (info.discordSettings == ["noPost"]) return;
+        if (info.discordPaths == ["noPost"]) return;
 
-        // Else, Assign a webhook for each escalationPathWebHook defined for the provided discordSettings. This lets
+        // Else, Assign a webhook for each escalationPathWebHook defined for the provided discordPaths. This lets
         // the logger define exactly which logs should go to which discord channel.
-        webHooks = info.discordSettings.map((discordSetting: string) => this.escalationPathWebhookUrls[discordSetting]);
+        webHooks = info.discordPaths.map((discordPath: string) => this.escalationPathWebhookUrls[discordPath]);
       }
 
       // Else, There are no discord specific settings. In this case, we treat this like a normal log. Either the
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -37,8 +37,8 @@ export class DiscordTransport extends Transport {
       // A log message can conditionally contain additional Discord specific paths. Validate these first.
       let webHooks: string[] = [];
       if (info.discordPaths) {
-        // If it contains "noPost" then this is a noop message for Discord and the log should be skipped.
-        if (info.discordPaths == ["noPost"]) return;
+        // If it is null then this is a noop message for Discord and the log should be skipped.
+        if (info.discordPaths === null) return;
 
         // Else, Assign a webhook for each escalationPathWebHook defined for the provided discordPaths. This lets
         // the logger define exactly which logs should go to which discord channel.
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -34,10 +34,22 @@ export class DiscordTransport extends Transport {
         embeds: [{ title: `${info.message}`, description: this.formatLinks(info.mrkdwn), color: 9696729 }],
       };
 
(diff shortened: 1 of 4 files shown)


packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -174,24 +169,19 @@ class OptimisticOracleContractMonitor {
...
         mrkdwn,
-        notificationPath: "optimistic-oracle",
-      });
-
-      // UI link is sent separately because it can break slack message limits.
-      this.logger[logLevel === "error" ? "info" : logLevel]({
-        at: "OptimisticOracleContractMonitor",
-        message: `${this.oracleType}: Price Proposal Alert 🧞‍♂️!`,
-        mrkdwn: this._generateUILink(event.requester, event.identifier, event.timestamp, event.ancillaryData),
+        discordPaths: ["oo-fact-checking", "oo-events"],
```
chrismaree, line 184: here you can see part of the new settings in action: we can define discord specific settings.

packages/monitors/src/OptimisticOracleContractMonitor.js
```diff
@@ -355,11 +333,8 @@ class OptimisticOracleContractMonitor {
     }
   }
 
-  _generateUILink(requester, identifier, timestamp, ancillaryData) {
-    const hexIdentifier = this.padRight(this.utf8ToHex(identifier), 64);
-    return `<${this.optimisticOracleUIBaseUrl}?chainId=${this.contractProps.chainId}&requester=${this.toChecksumAddress(
-      requester
-    )}&identifier=${hexIdentifier}&timestamp=${timestamp}&ancillaryData=${ancillaryData}|View this request in the UI.>`;
+  _generateUILink(transactionHash, chainId) {
+    return `<${this.optimisticOracleUIBaseUrl}/request?transactionHash=${transactionHash}&chainId=${chainId}|View in UI>`;
```
chrismaree, line 337: shortened URL format.

packages/monitors/src/OptimisticOracleContractMonitor.js
nicholaspai: These have to correspond to discord slack channels right? I would probably recommend storing these hardcoded values in a constants file or somewhere else in case channel names change. If channel names change then we only need to change the constants file and have less diff

nicholaspai (approved):
+1 on the logic but I think you should consider refactoring all the hardcoded discord channel names into a constant file

packages/monitors/src/OptimisticOracleContractMonitor.js
mrice32: +1 constants would help here

mrice32 (commented):
Looks good! A few comments on the transport.

packages/monitors/src/OptimisticOracleContractMonitor.js
chrismaree: so the logic with this is that it is _only_ for discord. it does not have to be associated with slack channels (and in fact is unrelated to slack logic!). there are only a finite set of events we care about on Discord (the total in this PR).

packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -34,10 +34,21 @@ export class DiscordTransport extends Transport {
...
       };
 
-      // Either the transport is configured to send on undefined escalation paths (will use default path if path does
-      // not exist) or the transport has a defined escalation path for the given message's notification path.
+      // A log message can conditionally contain additional Discord specific paths. Validate these first.
       let webHooks: string[] = [];
-      if (this.postOnNonEscalationPaths || this.escalationPathWebhookUrls[info.notificationPath]) {
+      if (info.discordPaths) {
+        // If it contains "noPost" then this is a noop message for Discord and the log should be skipped.
+        if (info.discordPaths == ["noPost"]) return;
```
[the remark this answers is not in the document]
chrismaree: thanks for the feedback @mrice32 . I did test this out pretty extensively locally and it worked correctly between slack/discord and the interplay therein. I have changed to using `===` and using `null` as the signal to not post to discord and fallback to default behavior when undefined.

The way to think about this transport is as follows:
- if no `notificationPath` or `discordPaths` is provided in the log and `postOnNonEscalationPaths=true` it posts to `defaultWebHookUrl`. 
- If no `notificationPath` or `discordPaths` is provided in the log and `postOnNonEscalationPaths=false` then it does not post.
- if a `notificationPath` is provided, then it acts like other transports: posts to the associated configured `escalationPathWebhookUrls` or the default, if a default is given.
- if a `discordPaths` is provided then this overrides this `notificationPath` behavior to either: 
  - do nothing if the type is `null` or
  - sent the discord message to all channels defined in the array of `discordPaths`
- if both `notificationPath` and `discordPaths` are provided then `discordPaths` takes preference.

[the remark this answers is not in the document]
mrice32: @chrismaree really? That's pretty surprising -- I tested it out locally in node before posting the comment above and got the following results:
```
Welcome to Node.js v14.17.3.
Type ".help" for more information.
> ["hey"] == ["hey"]
false
> ["hey"] === ["hey"]
false
```

Gotcha, thanks for the explanation!

PR #3850, opened 2022-03-21
fix: fix bug in relayer where speedup is attempted when impossible

mrice32:
**Motivation**

Bot considers a speed up transaction as possible when it should not, causing the bot to be confused and generate errors when the Speedup is profitable but the instant relay is not.

**Summary**

This forces speed up revenue to be 0 when the relay is not in the pending state.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
I ran this locally and it produced the correct output. will merge and keep an eye on the logs.

PR #3852, opened 2022-03-21
feat: add better provider logging and retries to redundant event calls

mrice32:
**Motivation**

We'd like the redundant provider calls to have at least the same number of retries that they normally do.

**Summary**

This updates the way that the redundant providers are created in a few ways:
1. It no longer concatenates the primary RetryProvider with the split "fallback" or redundant ones, since the primary RetryProvider is contains all of the other providers (so it amounts to calling the same provider twice).
2. If there is no Retry config, this just constructs a single provider with no retries using the NODE_URL_<chain id> env variable. This is done because if the user didn't include a retry config, they did not intend to have any retries.
3. It adds some additional logging to the retry provider. Now, when it errors, it will print out an error for each URL in the config. This means that a user can see why each provider errored and which provider urls were generating each error. This also means that any provider error that is layered on top of the retry provider will see the url that was called to create that error. Note: this does mean that provider urls are exposed in logs. I don't think that's a huge issue though.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/common/src/ProviderUtils.ts
+++ b/packages/common/src/ProviderUtils.ts
@@ -164,9 +164,14 @@ export function getWeb3ByChainId(chainId: number): Web3 {
 export function getRetryWeb3sByChainId(chainId: number): Web3[] {
   const retryConfigJson = process.env[`RETRY_CONFIG_${chainId}`] || "[]";
   const retryConfig: RetryConfig[] = JSON.parse(retryConfigJson);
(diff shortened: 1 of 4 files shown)


chrismaree (approved):
I tried running this locally and it seemed to work great! I ran one of the tests while the bots in prod were timing out and my local instance worked fine. going to merge and keen an eye on the logs.

PR #3856, opened 2022-03-23
fix(across-relayer): more accurately determine a safe last block number to query

mrice32:
**Motivation**

We're often seeing errors like this:
```
One of the blocks specified in filter (fromBlock, toBlock or blockHash) cannot be found.
```

**Summary**

I'm assuming this is happening because of block propagation delays between providers. To more accurately determine the latest block to query for each provider, this change queries all redundant providers and takes the min to ensure all providers have seen that block.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


chrismaree (approved):
this also seems reasonable. let's see if it helps with some of the painful node issues!

PR #3857, opened 2022-03-23
fix(serverless-spoke): extend stdout buffer to avoid process being killed

mrice32:
**Motivation**

Max buffer is being _sometimes_ in the bots due to the extreme amount of logging coming from the arbitrum finalizer.

We're seeing logs like this:
```
"execResponse":{"error":{"code":"ERR_CHILD_PROCESS_STDIO_MAXBUFFER"
```

**Summary**

To stop the process from being terminated due to the buffer being maxed out, this PR extends the buffer by 8x to alleviate the issue. I'm unaware of how to completely get rid of this issue, but there doesn't seem to be an obvious reason why we can't extend the buffer to 8MB from a default of 1MB.

See [docs](https://nodejs.org/docs/latest-v16.x/api/child_process.html#child_processexeccommand-options-callback) for more info on this option.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessSpoke.js
+++ b/packages/serverless-orchestration/src/ServerlessSpoke.js
@@ -21,6 +21,11 @@ let customLogger;
 
 const waitForLoggerDelay = process.env.WAIT_FOR_LOGGER_DELAY || 5;
 
+// To be used with exec to override the default input output buffer size.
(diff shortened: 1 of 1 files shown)


chrismaree (approved):
very good debugging, fixing and simple implementation.

PR #3858, opened 2022-03-23
fix: Address Discord transport rate limiting issues

chrismaree:
**Motivation**

The discord transport was having rate limiting issues when we try and send too many requests at once. This PR implements a log queue that preseves dropped logs and backs off when the discord API informs us that we are hitting it too hard.

Note: The implementation done here will work well for other transports that struggle to flush! we could re-work this same technique into the slack, pagerduty and GCP transports in the future and bring back the much loved `waitForLogger` method that was deprecated due to its issues.

**Summary**

Fixes Discord transport and introduces a new paridime of log-blocking-flushing for the UMA ecosystem 🧘‍♂️.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -6,12 +6,22 @@ import axios from "axios";
 
 type TransportOptions = ConstructorParameters<typeof Transport>[0];
 
+interface Body {
+  username: string;
+  avatar_url: string;
+  embeds: { title: string; description: string; color: number }[];
+}
+interface QueueElement {
+  body: Body;
+  webHook: string;
+}
+
 export class DiscordTransport extends Transport {
   private readonly defaultWebHookUrl: string | string[];
   private readonly escalationPathWebhookUrls: { [key: string]: string | string[] };
   private readonly postOnNonEscalationPaths: boolean;
 
-  private logQueue: any = [];
+  private logQueue: QueueElement[];
   private backOffDuration = 0;
   private isQueueBeingExecuted = false;
 
@@ -27,6 +37,8 @@ export class DiscordTransport extends Transport {
     this.defaultWebHookUrl = ops.defaultWebHookUrl;
     this.escalationPathWebhookUrls = ops.escalationPathWebhookUrls ?? {};
     this.postOnNonEscalationPaths = ops.postOnNonEscalationPaths ?? true;
+
+    this.logQueue = [];
   }
 
   // Note: info must be any because that's what the base class uses.
@@ -92,6 +104,7 @@ export class DiscordTransport extends Transport {
       let webHook, body;
       try {
         // Pop off the first element (oldest) and try send it to discord. If this errors then we are being rate limited.
+        // @ts-ignore: Object is possibly 'undefined'.
         ({ webHook, body } = this.logQueue.shift());
         await axios.post(webHook, body);
       } catch (error: any) {
--- a/packages/financial-templates-lib/src/logger/DiscordTransport.ts
+++ b/packages/financial-templates-lib/src/logger/DiscordTransport.ts
@@ -10,6 +10,11 @@ export class DiscordTransport extends Transport {
   private readonly defaultWebHookUrl: string | string[];
   private readonly escalationPathWebhookUrls: { [key: string]: string | string[] };
   private readonly postOnNonEscalationPaths: boolean;
+
+  private logQueue: any = [];
+  private backOffDuration = 0;
+  private isQueueBeingExecuted = false;
+
   constructor(
     winstonOpts: TransportOptions,
     ops: {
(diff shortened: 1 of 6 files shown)


packages/financial-templates-lib/src/logger/Logger.ts
```diff
@@ -31,15 +31,14 @@ import winston from "winston";
...
-// transports to clear. This enables slower transports like slack to still send their messages before the process yields.
-// Note: typescript infers the return type to be unknown. This is fine, as the return type should be void and unused.
-export async function waitForLogger(logger: _Logger): Promise<unknown> {
-  const loggerDone = new Promise((resolve) => logger.on("finish", resolve));
-  logger.end();
-  return await loggerDone;
+// This method will check if the AugmentedLogger's isFlushed is set to true. If not, it will block until such time
+// that it has been set to true. Note that each blocking transport should implement this isFlushed bool to prevent
+// the logger from closing before all logs have been propagated.
+export async function waitForLogger(logger: AugmentedLogger) {
```
chrismaree, line 40: goodbye broken `waitForLogger` and hello fixed version that enables transports to block! we can add this to all the other transports now.

packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -10,6 +10,11 @@ export class DiscordTransport extends Transport {
   private readonly defaultWebHookUrl: string | string[];
   private readonly escalationPathWebhookUrls: { [key: string]: string | string[] };
   private readonly postOnNonEscalationPaths: boolean;
+
+  private logQueue: any = [];
```
mrice32, line 14: It looks like the types for this queue are pretty well-defined. Might make sense to just make it explicit.

```suggestion
  interface Body {
    username: string;
    avatar_url: string;
    embeds: { title: string; description: string; color: number }[];
  }
  interface QueueElement {
    body: Body;
    webHook: string;
  };

  private logQueue: QueueElement[] = [];
```

chrismaree: I had not done this as the shift operator does not play well with this:
image

chrismaree: I'll try find a workaround.

chrismaree: I had to add a ts-ignore for this. could not find another way.

chrismaree: then I had to add `@typescript-eslint/ban-ts-comment` to get this through the linter :(

mrice32: As mentioned IRL, I think you can use the non-null assertion operator, `!` since you can't _prove_ to the compiler that it's non-null.

```ts
({ webHook, body } = this.logQueue.shift()!);
```

packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -59,18 +76,58 @@ export class DiscordTransport extends Transport {
...
+  // variable to block the bot from closing until the whole queue has been flushed.
+  async executeLogQueue(): Promise<void> {
+    if (this.isQueueBeingExecuted) return; // If the queue is currently being executed, return.
+    this.isQueueBeingExecuted = true; // Set the queue to being executed.
+    // Set the parent isFlushed to false to prevent the logger from closing while the queue is being executed. Note this
+    // is separate variable from the isQueueBeingExecuted flag as this isFlushed is global for the logger instance.
+    (this as any).parent.isFlushed = false;
+
+    // If the previous iteration set a backOffDuration then wait for this duration.
+    if (this.backOffDuration != 0) await delay(this.backOffDuration);
```
mrice32, line 100: nit: why do we store backOffDuration on the class? Why not just store it as a local variable declared at the beginning of this method? Is it relevant to anything outside this run?

mrice32 (approved):
LGTM! Only minor nits

packages/financial-templates-lib/src/logger/DiscordTransport.ts
```diff
@@ -59,18 +76,58 @@ export class DiscordTransport extends Transport {
...
+        // re-call the executeLogQuery. This will initiate the backoff delay and then continue to process the queue.
+        this.isQueueBeingExecuted = false;
+        await this.executeLogQueue();
+      }
+    }
+
+    // Only if we get to the end of the function (no errors) then we can set isQueueBeingExecuted to false and can
+    // set the parent isFlushed to true, enabling the bot to close out.
+    this.isQueueBeingExecuted = false;
+    (this as any).parent.isFlushed = true;
```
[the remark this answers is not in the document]
chrismaree: agreed. future problem. for now let's fix the discord issue seeing that's affecting users.

PR #3861, opened 2022-03-25
feat: add OptimisticDistributor

Reinis-FRP:

**Motivation**

Rewards distribution through OptimisticRewarder might not scale well for certain applications as Optimistic Oracle would need to verify each recipient payout individually. Instead it is proposed to create OptimisticDistributor contract that allows rewards distribution to multiple recipients by verifying only the merkle root of proposed rewards distribution.

**Summary**

The OptimisticDistributor (OD) contract will accept deposits from sponsors and allow anyone to propose rewards distribution through MerkleDistributor contract and verified by Optimistic Oracle.

**Details**

OptimisticDistributor would also need new price identifier(s) instructing disputers/voters on how to check proposed rewards distribution payouts and verify merkle root against previously approved rewards payout rules (passed in ancillary data). If Optimistic Oracle resolves to 1 anyone can execute transfer of rewards to the MerkleDistributor for the recipients to claim. If 1 is not resolved the rewards are kept at OptimisticDistributor until the next valid proposal.

**Testing**

- [x]  New unit tests created
- [x]  Untested


packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+     * in this contract to map to the latest version in the Finder and store the latest final fee.
+     */
+    function syncUmaEcosystemParams() public nonReentrant() {
+        store = _getStore();
+        finalFee = store.computeFinalFee(address(bondToken)).rawValue;
+        optimisticOracle = _getOptimisticOracle();
+    }
+
+    // Can be overriden for testing.
+    function getCurrentTime() public view virtual returns (uint256) {
```
Reinis-FRP, line 381: Alternative would be to inherit Testable in production code (like in LSP), but currently I used the approach as in optimistic-rewarder. Which approach shall we keep going forward?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+    {
+        return AncillaryData.appendKeyValueUint(customAncillaryData, "proposalIndex", proposalIndex);
+    }
+
+    function _ancillaryDataWithinLimits(bytes memory customAncillaryData) internal view returns (bool) {
+        // Since proposalIndex has variable length as string, it is not appended here and is assumed
+        // to be included in ANCILLARY_BYTES_RESERVE.
+        return
+            optimisticOracle.stampAncillaryData(customAncillaryData, address(this)).length + ANCILLARY_BYTES_RESERVE <=
+            optimisticOracle.ancillaryBytesLimit();
```
Reinis-FRP, line 418: Should `ancillaryBytesLimit` better be stored in this contract and updated through `syncUmaEcosystemParams`?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+
+        // Delete resolved proposal from storage. This also avoids double-spend for approved proposals.
+        delete proposals[proposalIndex];
+        emit ProposalDeleted(proposal.rewardIndex, proposalIndex);
+    }
+
+    /**
+     * @notice Allows any caller to delete distribution that was rejected by the Optimistic Oracle.
+     */
+    function deleteRejectedDistribution(uint256 proposalIndex) external nonReentrant() {}
```
Reinis-FRP, line 346: This is not implemented yet, but since rejected proposals get deleted in `executeDistribution`, not sure if separate call is required, except for allowing to reduce state bloat. But who would be motivated to call that anyways?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrismaree, line 1: comment about location: does it make sence for this to be in the `financial-templates` directory? we might want to move both the `OR` and the `OD` into their own packages along with the `Merkle-distributor`. 

dont need to do anything here for this @Reinis-FRP just wanted to get feedback on this from @nicholaspai & @mrice32

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+ * @notice Allows sponsors to distribute rewards through MerkleDistributor contract secured by UMA Optimistic Oracle.
+ */
+contract OptimisticDistributor is Lockable, MultiCaller {
+    using SafeERC20 for IERC20;
+
+    /********************************************
+     *  OPTIMISTIC DISTRIBUTOR DATA STRUCTURES  *
+     ********************************************/
+
+    // Represents reward posted by a sponsor.
```
chrismaree, line 29: good grouping by variable type.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+    struct Proposal {
+        uint256 rewardIndex;
+        uint256 timestamp;
+        bytes32 merkleRoot;
+        string ipfsHash;
+    }
+
+    /********************************************
+     *      INTERNAL VARIABLES AND STORAGE      *
+     ********************************************/
```
chrismaree, line 51: this can probs be removed. these variables are not "internal" per say. consider removing this block or changing it to more accurately describe the state variables.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+
+    event RewardCreated(
+        address indexed sponsor,
+        IERC20 rewardToken,
+        uint256 indexed rewardIndex,
+        uint256 maximumRewardAmount,
+        uint256 earliestProposalTimestamp,
+        uint256 optimisticOracleProposerBond,
+        uint256 optimisticOracleLivenessTime,
+        bytes32 priceIdentifier,
```
chrismaree, line 95: we can index up to 3. no point in not using all of them :)
```suggestion
        bytes32 indexed priceIdentifier,
```

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+        require(_getCollateralWhitelist().isOnWhitelist(address(_bondToken)), "Bond token not supported");
+        bondToken = _bondToken;
+        syncUmaEcosystemParams();
+    }
+
+    /********************************************
+     *            FUNDING FUNCTIONS             *
+     ********************************************/
+
+    /**
```
chrismaree, line 138: good commenting on props! thanks for doing this.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+        require(optimisticOracleLivenessTime >= MINIMUM_LIVENESS, "OO liveness too small");
+        require(optimisticOracleLivenessTime < MAXIMUM_LIVENESS, "OO liveness too large");
+
+        // Pull maximum rewards from the sponsor.
+        rewardToken.safeTransferFrom(msg.sender, address(this), maximumRewardAmount);
+
+        // Store funded reward and log created reward.
+        Reward memory reward =
+            Reward({
+                sponsor: msg.sender,
```
chrismaree, line 173: does the key "sponsor" make the most sence here? would "proposer" be better than "sponsor"? I bring this up when looking at the `increaseReward` method where you check the value of `sponsor` not being the 0 address and that's a bit confusing as there can then be multiple `sponsors`. WDYT?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+        bondToken = _bondToken;
+        syncUmaEcosystemParams();
+    }
+
+    /********************************************
+     *            FUNDING FUNCTIONS             *
+     ********************************************/
+
+    /**
+     * @notice Allows any caller to create a rewards object and deposit tokens that are linked to these rewards.
```
chrismaree, line 139: throughout this contract you refer to it as a "rewards object". Consider changing this to a "Reward struct" which is more descriptive of the solidity data structure.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+        bytes32 priceIdentifier;
+        bytes customAncillaryData;
+    }
+
+    // Represents proposed rewards distribution.
+    struct Proposal {
+        uint256 rewardIndex;
+        uint256 timestamp;
+        bytes32 merkleRoot;
+        string ipfsHash;
```
chrismaree, line 46: does this need to be stored on chain? is the intention for this to be run on L1 or L2s? if L2s/polygon then this is fine. if it's on L1 we should think about this a bit more.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+
+    /**
+     * @notice Allows any caller to propose distribution for funded reward starting from `earliestProposalTimestamp`.
+     * @dev The caller must approve this contract to transfer `optimisticOracleProposerBond` + final fee amount
+     * of `bondToken`.
+     * @param rewardIndex Index for identifying existing rewards object that should be proposed for distribution.
+     * @param merkleRoot Merkle root describing allocation of proposed rewards distribution.
+     * @param ipfsHash Hash of IPFS object, conveniently stored for clients to verify proposed distribution.
+     */
+    function proposeDistribution(
```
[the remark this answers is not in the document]
chrismaree: a question about the structure of this proposal function:

I am correct in saying that there is always, at maximum, one proposal per `rewardIndex`, right? i.e the proposal is showing how to distribute the reward defined in the `rewardIndex`?

If this is the case, would you the case do you think it makes sence to only allow for ONE proposal to happen at a time? consider the case where two people propose valid `merkleRoot`s right after each other and both are in liveness for the same period. As soon as the first proposal passes liveness without dispute the second proposal will become invalid as you cant double propose for the same thing. This means that, if timed correctly you'd be able to get the second proposal in and passed liveness before it can be disputed.

We have a similar issue in Across-v2 with bundle proposals and the possibility that two bundles will have overlapping information. In that context we fixed by saying that you can only ever have one proposal in liveness at a time. period. you could do the same thing here but restrict it such that you only ever have one proposal _perRewardIndex_ at a time.

The implementation for this would look like adding an enum `DistributionProposed { None, Pending, Accepted }; `within the state of  to `Reward`. This  indicates if there is a current state of the proposal for the given `rewards[rewardIndex]` which then blocks subsequent proposals until either: 1) Set to `Pending` when this method is called thereby blocking other requests (must be on `None` state to accept a proposal) 2) the proposal passes liveness and is accepted via the `executeDistribution` which would set this to `Accepted` (also blocking future proposals) or 3) is called back from the OO via the `priceDisputed` callback function that would re-set this to `None` which will enable future proposals to happen again.

I'm curious what @UMAprotocol/smart-contracts thinks about this proposed change ^^

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+                reward.rewardToken,
+                proposal.rewardIndex,
+                proposalIndex,
+                reward.maximumRewardAmount,
+                proposal.merkleRoot,
+                proposal.ipfsHash
+            );
+        } else emit ProposalRejected(proposal.rewardIndex, proposalIndex);
+
+        // Delete resolved proposal from storage. This also avoids double-spend for approved proposals.
```
chrismaree, line 338: this would change via [this](https://github.com/UMAprotocol/protocol/pull/3861/files#r849171534) proposal, if we were to implement it.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: I think the enum makes more sence but we can wait for feedback from others on this.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: both are pretty similar. IMHO doing the `testable` version is more consistant and cleaner as it means you won't need to have the `test` contract for `OptimisticDistributor` as you can simply deploy a `Timer` in your tests and then pass it in.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: it would be more consistant with the pattern you've implemented here :)

packages/core/test/financial-templates/optimistic-distributor/OptimisticDistributor.js
```diff
@@ -0,0 +1,908 @@
...
+    bondToken,
+    mockOracle,
+    optimisticDistributor,
+    optimisticOracle,
+    merkleDistributor,
+    rewardToken,
+    earliestProposalTimestamp,
+    defaultRewardParameters;
+
+  const mintAndApprove = async (token, owner, spender, amount, minter) => {
```
chrismaree, line 51: good job on creating helper functions in the test. this is great!

packages/core/test/financial-templates/optimistic-distributor/OptimisticDistributor.js
```diff
@@ -0,0 +1,908 @@
...
+    await optimisticDistributor.methods.proposeDistribution(rewardIndex, merkleRoot, ipfsHash).send({ from: proposer });
+    return [totalBond, ancillaryData, proposalTimestamp, merkleRoot];
+  };
+
+  const advanceTime = async (timeIncrease) => {
+    const currentTime = parseInt(await timer.methods.getCurrentTime().call());
+    await timer.methods.setCurrentTime(currentTime + timeIncrease).send({ from: deployer });
+  };
+
+  const didContractRevertWith = async (promise, expectedMessage) => {
```
chrismaree, line 81: I really like this. helps be more explicit with knowing what you are testing.

packages/core/test/financial-templates/optimistic-distributor/OptimisticDistributor.js
```diff
@@ -0,0 +1,908 @@
...
+    const contractBalanceAfter = toBN(
+      await rewardToken.methods.balanceOf(optimisticDistributor.options.address).call()
+    );
+
+    // Check for correct change in balances.
+    assert.equal(sponsorBalanceBefore.sub(sponsorBalanceAfter).toString(), rewardAmount);
+    assert.equal(contractBalanceAfter.sub(contractBalanceBefore).toString(), rewardAmount);
+
+    // Check that created rewards are emitted.
+    await assertEventEmitted(
```
chrismaree, line 400: very good. thank you for being detailed focused in your testing! I know some of these can be a pain to write.

chrismaree (commented):
This is excellent work @Reinis-FRP ! I left a first pass of comments and one recommendation on an architectural/structural change. your tests are excellent as well, thank you for putting in the time to get those to the level they are at. I will do a second round of review once the first batch of comments have been addressed and will spend some more time reading through your unit tests.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: fixed in a45771db21f10a72b5bfcf10529c47c873dbe312

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: fixed in 3c2941ddcef57e9e71c3497709fff5cadedca7f7

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: fixed in 1766ed802ff29f1371896ba5d54765ce2fd35e04

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: "proposer" is the party submitting merkleRoot for proposed reward payout distribution together with proposal bond + final fee, and it would be recorded as "proposer" in OO price proposal. So it would be confusing to use that naming in this contract for different role.

Here at OptimisticDistributor "sponsor" is the first party funding the reward and it is unique per Reward struct. I agree it could be bit confusing as there could be multiple sponsors funding through `increaseReward` method that currently are not recorded. Not sure if worth tracking all of them in state variables, but maybe I should emit add-on sponsors in the `RewardIncreased` events. Then the price identifier/ancillary data could have instructions on how to distribute any remaining rewards (e.g. if used in bribing as rewards per % gauge allocation) back to sponsors as part of merkle root distribution. Price identifier/ancillary data could then also decide if all remaining rewards go back to first sponsor calling `createReward` method or distributed pro-rata to all sponsors also considering all callers of `increaseReward` for the same `rewardIndex`.

Now I realize this contract currently does not strictly require storing original sponsor as part of Reward struct and one could check the validity of `rewardIndex` by non-zero `rewardToken` address instead. But there might be future upgrades that allow clawing back unclaimed rewards from MerkleDistributor if predefined time has passed (e.g. due to data availability concerns for storing IPFS payout object) - then OptimisticDistributor might need original sponsor address stored to return those unclaimed rewards.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: fixed in df7e415f38dd6c6c8269693b2c59e42b288914f1

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,420 @@
...
+    function proposeDistribution(
+        uint256 rewardIndex,
+        bytes32 merkleRoot,
+        string calldata ipfsHash
+    ) external nonReentrant() {
+        Reward memory reward = rewards[rewardIndex];
+        require(reward.sponsor != address(0), "Invalid rewardIndex");
+
+        uint256 timestamp = getCurrentTime();
+        require(timestamp >= reward.earliestProposalTimestamp, "No proposals in funding period");
```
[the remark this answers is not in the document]
Reinis-FRP: It is still funding period till `earliestProposalTimestamp`, so meaning no proposals are allowed yet.
To avoid confusion, changed to "Cannot propose in funding period" in 42be17adb2cec82c83ff2a93541d2d928e48e976

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: Primary application is planned on Polygon, but this can be considered in v2 if there is demand for L1 application.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
nicholaspai: Yeah I think a follow up PR should move both the `OR` and the `OD` into their own folder but for now keeping it here is consistent with where the `OR` is located. Do you think we should move them in this PR?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+
+    // Represents reward posted by a sponsor.
+    struct Reward {
+        address sponsor;
+        IERC20 rewardToken;
+        uint256 maximumRewardAmount;
+        uint256 earliestProposalTimestamp;
+        uint256 optimisticOracleProposerBond;
+        uint256 optimisticOracleLivenessTime;
+        bytes32 priceIdentifier;
```
nicholaspai, line 38: Will `priceIdentifier` be different per reward or will they all use the same identifier?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+
+    // Ancillary data length limit can be synced and stored in the contract.
+    uint256 public ancillaryBytesLimit;
+
+    // Index of next created reward or proposal.
+    uint256 public nextCreatedReward;
+    uint256 public nextCreatedProposal;
+
+    // Rewards and proposals are mapped to their indices.
+    mapping(uint256 => Reward) public rewards;
```
nicholaspai, line 73: Why not just use an array so we don't need to store `nextCreatedReward` and `nextCreatedProposal`? @chrismaree wdyt

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+        uint256 optimisticOracleLivenessTime,
+        bytes32 priceIdentifier,
+        IERC20 rewardToken,
+        bytes calldata customAncillaryData
+    ) external nonReentrant() {
+        require(address(merkleDistributor) != address(0), "Missing MerkleDistributor");
+        require(_getIdentifierWhitelist().isIdentifierSupported(priceIdentifier), "Identifier not registered");
+        require(_ancillaryDataWithinLimits(customAncillaryData), "Ancillary data too long");
+        require(optimisticOracleLivenessTime >= MINIMUM_LIVENESS, "OO liveness too small");
+        require(optimisticOracleLivenessTime < MAXIMUM_LIVENESS, "OO liveness too large");
```
nicholaspai, line 174: nit: this should be `<=` right? When I read `MAXIMUM_LIVENESS` I read that the value is allowable, but nothing greater

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
nicholaspai: I think having `proposer` mean different things for this contract and the OO is fine, as is the case for the across-v2 contracts

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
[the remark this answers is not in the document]
nicholaspai: Yeah I think it simplifies the contract a bit to only allow one proposal at a time. The pros are:
- gas savings: only need to store one proposal struct instead of an array
- practicality: I think these rewards will be paid out relatively infrequently, maybe once a day max, so no need to allow multiple to be in liveness at the same time

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+
+    /**
+     * @notice Sets address of MerkleDistributor contract that will be used for rewards distribution.
+     * MerkleDistributor address can only be set once.
+     * @dev It is expected that the deployer first deploys MekleDistributor contract and transfers its ownership to
+     * the OptimisticDistributor contract and then calls `setMerkleDistributor` on the OptimisticDistributor pointing
+     * on now owned MekleDistributor contract.
+     * @param _merkleDistributor Address of the owned MerkleDistributor contract.
+     */
+    function setMerkleDistributor(MerkleDistributor _merkleDistributor) external nonReentrant() {
```
nicholaspai, line 369: I'm curious why you can't pass in the `merkleDistributor` address in the constructor?

nicholaspai (commented):
This is awesome work and I really appreciate the comments it makes the contracts much easier to understand. My general feedback is that in the future, you should probably have created an interface or empty contract in a PR with fewer lines just to get feedback on the general structure of the contract before you implemented every single function and wrote all tests.

I think its OK in this case because the contract is relatively small, but I want to reach consensus on the contract before I even look at any of the tests. Sometimes I'll look at the tests to get context for understanding the intended functionality of a contract but in your case, you commented the contract so well that it was quite easy!

I left some nit comments but I think the one big change I am in agreement with Chris on is that we should consider allowing only one proposal at a time. Did you consider this?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: There might be couple of price identifiers depending on what application is using them

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: At some point I was considering addressing struct mappings by hashing some of its fields as index, thus allowing to detect duplicate proposals as raised [here](https://github.com/UMAprotocol/protocol/pull/3861/files#r848957341). In case we address this by allowing only 1 concurrent proposal per reward (with enum state element to reward struct) I indeed should use dynamic array instead.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: No, OO does not support that: https://github.com/UMAprotocol/protocol/blob/d472e177c3485d1fbb149e10a61165d41199347f/packages/core/contracts/oracle/implementation/OptimisticOracle.sol#L622

Reinis-FRP: but not sure how to rename MAXIMUM_LIVENESS properly to reflect this?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+        nextCreatedReward++;
+    }
+
+    /**
+     * @notice Allows anyone to deposit additional rewards for distribution before `earliestProposalTimestamp`.
+     * @dev The caller must approve this contract to transfer `additionalRewardAmount` amount of `rewardToken`.
+     * @param rewardIndex Index for identifying existing Reward struct that should receive additional funding.
+     * @param additionalRewardAmount Additional reward amount that the sponsor is posting for distribution.
+     */
+    function increaseReward(uint256 rewardIndex, uint256 additionalRewardAmount) external nonReentrant() {
```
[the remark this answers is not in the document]
Reinis-FRP: Potential reward recipients should be able to start their work (e.g. voting in bribing rewards) already when initial rewards are funded. In case of irreversible on-chain votes this could be a problem if promised reward is decreased after the fact.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: `merkleDistributor` should be owned by the OptimisticDistributor that is not yet deployed.

Should I create separate creator contract similarly as done in LSP taking ownership on long/short tokens?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: I would still believe "sponsor" better represents a party funding the rewards and "proposer" being a party proposing distribution of those rewards

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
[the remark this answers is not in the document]
Reinis-FRP: In order to use `priceDisputed` callback from OO to flag `Reward` as `DistributionProposed.None` the callback function would need to parse ancillary data to identify appended `proposalIndex`. Parsing strings in solidity does not sound great, so as alternative approach I now plan to:
- `proposalIndex` replaced with `rewardIndex` when appending to ancillary data
- hash `identifier`, `timestamp` and `ancillaryData`
- use above hash to address `Proposal` structs in a mapping
- in `priceDisputed` callback function hash the passed `identifier`, `timestamp` and `ancillaryData` arguments from OO that allows identifying `Proposal` struct and link it back to relevant `Reward` struct so that it can be flagged as `DistributionProposed.None` and new proposals allowed for the disputed reward.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: na let's move it separately in one PR change.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: @Reinis-FRP I think the point that matters, like you called out, is they are the one _sponsoring_ the payout so it works just fine.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+        bytes calldata customAncillaryData
+    ) external nonReentrant() {
+        require(address(merkleDistributor) != address(0), "Missing MerkleDistributor");
+        require(_getIdentifierWhitelist().isIdentifierSupported(priceIdentifier), "Identifier not registered");
+        require(_ancillaryDataWithinLimits(customAncillaryData), "Ancillary data too long");
+        require(optimisticOracleLivenessTime >= MINIMUM_LIVENESS, "OO liveness too small");
+        require(optimisticOracleLivenessTime < MAXIMUM_LIVENESS, "OO liveness too large");
+
+        // Pull maximum rewards from the sponsor.
+        rewardToken.safeTransferFrom(msg.sender, address(this), maximumRewardAmount);
```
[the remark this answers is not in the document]
Reinis-FRP: agree, will do it once I implement enum `DistributionProposed`

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,425 @@
...
+    ) external nonReentrant() {
+        Reward memory reward = rewards[rewardIndex];
+        require(reward.sponsor != address(0), "Invalid rewardIndex");
+
+        uint256 timestamp = getCurrentTime();
+        require(timestamp >= reward.earliestProposalTimestamp, "Cannot propose in funding period");
+
+        // Append proposal index to ancillary data.
+        uint256 proposalIndex = nextCreatedProposal;
+        bytes memory ancillaryData = _appendProposalIndex(proposalIndex, reward.customAncillaryData);
```
chrismaree, line 252: ^

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -0,0 +1,464 @@
...
+            emit RewardDistributed(
+                reward.sponsor,
+                reward.rewardToken,
+                proposal.rewardIndex,
+                reward.maximumRewardAmount,
+                proposalId,
+                proposal.merkleRoot,
+                proposal.ipfsHash
+            );
+        } else emit ProposalRejected(proposal.rewardIndex, proposalId);
```
Reinis-FRP, line 357: Repeated calls to `executeDistribution` would continue emitting `ProposalRejected` for rejected proposals without settlement to proposer/disputer.

Alternative would be to simplify this contract and just `require(resolvedPrice == 1e18)` without emitting `ProposalRejected`, but then the disputer would need to manually claim settlement on OO. Which approach is better?

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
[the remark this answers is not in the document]
Reinis-FRP: I switched to dynamic array for `Reward` structs and mapped `Proposal` structs by hashing `identifier`, `timestamp` and `ancillaryData` that is looked up in `priceDisputed` callback function.

Now only one undisputed distribution proposal is allowed through enum `DistributionProposed` and I am not deleting `Proposal` structs anymore.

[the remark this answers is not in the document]
chrismaree: excellent! the way you have it now is great@!

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: I think what you have here makes sense due to the reason you described. add a comment to say why you did. itthis way and then it's good.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
chrismaree: we can create a factory but leave that for a follow on PR. what you have now looks great.

packages/core/test/financial-templates/optimistic-distributor/OptimisticDistributor.js
```diff
@@ -0,0 +1,991 @@
...
+    assert.equal(oracleBalanceAfter.sub(oracleBalanceBefore).toString(), totalBond);
+
+    // Check stored proposal.
+    const storedProposal = await optimisticDistributor.methods.proposals(proposalId).call();
+    assert.equal(storedProposal.rewardIndex, rewardIndex);
+    assert.equal(storedProposal.timestamp, proposalTimestamp);
+    assert.equal(storedProposal.merkleRoot, merkleRoot);
+    assert.equal(hexToUtf8(storedProposal.ipfsHash), hexToUtf8(ipfsHash));
+
+    // Any further proposals should now be blocked till disputed.
```
chrismaree, line 630: good

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: added in a83eb842eeddb76aaa96718b74058547a424e778

chrismaree (approved):
Feel free to merge this in when you ready! LGTM

We will do another code reading session once it's merged and re-review it all before audit.

PR #3866, opened 2022-03-29
feat: added more robust bignumber formatting to winston transport

chrismaree:
**This PR also contains a release! hence the large diff!**

**Motivation**

The Winston transport does not play nicely with BigNumbers. This PR improves the situation by making it more resilient to both Ethers and Web3 big numbers by checking if the log message contains a nested big number. if it does, cast it to a string in the log.

**Summary**

Consider the following script:
```
import { Logger, delay } from "@uma/financial-templates-lib";

config();

import web3 from "web3";

const toBNEthers = (num: string | number | BigNumber) => {
  return BigNumber.from(num.toString());
};

export async function run(logger: winston.Logger): Promise<void> {

  const myEthersBN = toBNEthers(1234);
  const myBNWeb3 = web3.utils.toBN("123456");

  logger.info({
    at: "test",
    message: "This is a test🍌",
    someNest: { someKey: [myEthersBN] },
    deepNested: { key: { key: { key: { key: { value1: { myBNWeb3 }, myEthersBN } } } } },
    array: [myBNWeb3],
    myEthersBN,
  });
}
``` 

Ideally, we can log all big number types, irrespective of there level of nesting, as strings.

This PR makes this possible. The output would look like this:

```
2022-03-29 14:40:13 [info]: {
  "at": "test",
  "message": "This is a test🍌",
  "someNest": {
    "someKey": [
      "1234"
    ]
  },
  "deepNested": {
    "key": {
      "key": {
        "key": {
          "key": {
            "value1": {
              "myBNWeb3": "123456"
            },
            "myEthersBN": "1234"
          }
        }
      }
    }
  },
  "array": [
    "123456"
  ],
  "myEthersBN": "1234",
  "bot-identifier": "NO_BOT_ID"
}
```

Yay for logging!


packages/financial-templates-lib/src/logger/Logger.ts
```diff
@@ -41,29 +43,8 @@ export async function waitForLogger(logger: AugmentedLogger) {
...
-function errorStackTracerFormatter(logEntry: LogEntry) {
-  if (logEntry.error) {
-    logEntry.error = handleRecursiveErrorArray(logEntry.error);
-  }
-  return logEntry;
-}
-
-// Handle case where `error` is an array of errors and we want to display all of the error stacks recursively.
-// i.e. `error` is in the shape: [[Error, Error], [Error], [Error, Error]]
-function handleRecursiveErrorArray(error: Error | any[]): string | any[] {
```
chrismaree, line 54: pulled into a `Formatters` file.

daywiss (approved):
clever implementation. just out curiousity, was this mainly just for visibility or did this cause some kind of failure?

mrice32 (commented):
Can we split the code change from the release?

I think the process is typically:
1. Open PR to make functional change.
2. Merge PR.
3. Create the release.

Typically, releases aren't combined with other changes because the automatic updates to the changelog won't include the change that's included in the release PR.

chrismaree:
> Can we split the code change from the release?
> 
> I think the process is typically:
> 
> 1. Open PR to make functional change.
> 2. Merge PR.
> 3. Create the release.
> 
> Typically, releases aren't combined with other changes because the automatic updates to the changelog won't include the change that's included in the release PR.

my bad. will split the release to another PR.

mrice32:
> > Can we split the code change from the release?
> > I think the process is typically:
> > 
> > 1. Open PR to make functional change.
> > 2. Merge PR.
> > 3. Create the release.
> > 
> > Typically, releases aren't combined with other changes because the automatic updates to the changelog won't include the change that's included in the release PR.
> 
> my bad. will split the release to another PR.

No problem, it's a bit of a clunky process :/. Thank you!

mrice32 (approved):
LGTM!

PR #3872, opened 2022-04-04
fix: fallback to mainnet prices for L2 ERC20 tokens

amateima:

**Motivation**

https://app.shortcut.com/uma-project/story/5168/the-suint-kpi-option-is-showing-0-for-tvl-but-has-45k-uma-in-the-contract-https-polygonscan-com

**Summary**

Basically what I did was to identify the L2 ERC20 addresses for which CoinGecko didn't return a price and use the price of the corresponding mainnet address as fallback.

I also added an extra log to identify easily this issue

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/api/src/services/collateral-prices.ts
+++ b/packages/api/src/services/collateral-prices.ts
@@ -1,9 +1,9 @@
 import * as uma from "@uma/sdk";
 import bluebird from "bluebird";
 import assert from "assert";
-import { AppState, CurrencySymbol, BaseConfig, AppClients } from "../types";
+import { AppState, CurrencySymbol, BaseConfig, AppClients, ChainId, CHAIN_IDs } from "../types";
 interface Config extends BaseConfig {
-  network: number;
+  network: ChainId;
   currency?: CurrencySymbol;
 }
 import { parseUnits, msToS } from "../libs/utils";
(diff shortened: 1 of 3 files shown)


mrice32 (approved):
LGTM!

packages/api/src/services/collateral-prices.ts
```diff
@@ -13,6 +13,18 @@ type Dependencies = {
   appClients: AppClients;
 };
 
+const tokenAddressesMap = {
+  uma: {
```
daywiss, line 17: this seems unecessary, its basically ignored by the logic

packages/api/src/services/collateral-prices.ts
```diff
@@ -13,6 +13,18 @@ type Dependencies = {
...
+    [CHAIN_IDs.mainnet]: "0x04Fa0d235C4abf4BcF4787aF4CF447DE572eF828",
+    [CHAIN_IDs.polygon_matic]: "0x3066818837c5e6eD6601bd5a91B0762877A6B731",
+  },
```
daywiss, line 20: this could be simplified a lot if we basically assume we are always mapping any contract address to mainnet from any chain. could just become a KV of [sidechainaddress]: [mainnetaddress]. i guess to be really bullet proof you might need to specify the from chain, in which case id probably have a different object for each chain to mainnet with the simple {[fromaddress]:toaddress} format

packages/api/src/services/collateral-prices.ts
```diff
@@ -13,6 +13,18 @@ type Dependencies = {
...
+};
+
+const getTokenAddressFromOtherChain = (chainId: ChainId, address: string, toChainId: ChainId) => {
```
daywiss, line 23: again can be simplified if we are just assuming we want mainnet address.  i mean we dont have to assume that , and maybe its better not to, idk

daywiss (approved):
left some thoughts on how to simplify, but i think its good

packages/api/src/services/collateral-prices.ts
amateima: I liked this approach more than leaving comments about how to group the addresses

PR #3877, opened 2022-04-12
feat: add console slog to production config

chrismaree:
**Motivation**

We should be able to produce a logger that can: 1) log to GCP stack drive via the `ENVIRONMENT=production` and 2) be able to create a console tranport.

**Testing**

- [ ]  New unit tests created
- [X]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/logger/Transports.ts
+++ b/packages/financial-templates-lib/src/logger/Transports.ts
@@ -45,7 +45,7 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
     const { LoggingWinston } = require("@google-cloud/logging-winston");
     transports.push(new LoggingWinston());
     if (!require("@google-cloud/trace-agent").get().enabled) require("@google-cloud/trace-agent").start();
-  } else if (transportsConfig.createConsoleTransport || process.env.LOG_TO_CONSOLE == "true") {
+  } else if (transportsConfig?.createConsoleTransport || process.env.LOG_TO_CONSOLE == "true") {
     // Add a console transport to log to the console.
     transports.push(createConsoleTransport());
   }
--- a/packages/financial-templates-lib/src/logger/Transports.ts
+++ b/packages/financial-templates-lib/src/logger/Transports.ts
@@ -45,7 +45,11 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
     const { LoggingWinston } = require("@google-cloud/logging-winston");
     transports.push(new LoggingWinston());
     if (!require("@google-cloud/trace-agent").get().enabled) require("@google-cloud/trace-agent").start();
-  } else if (transportsConfig.createConsoleTransport != undefined ? transportsConfig.createConsoleTransport : true) {
+  } else if (
+    transportsConfig.createConsoleTransport != undefined
+      ? transportsConfig.createConsoleTransport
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/logger/Transports.ts
```diff
@@ -45,7 +45,11 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
     const { LoggingWinston } = require("@google-cloud/logging-winston");
     transports.push(new LoggingWinston());
     if (!require("@google-cloud/trace-agent").get().enabled) require("@google-cloud/trace-agent").start();
-  } else if (transportsConfig.createConsoleTransport != undefined ? transportsConfig.createConsoleTransport : true) {
+  } else if (
+    transportsConfig.createConsoleTransport != undefined
+      ? transportsConfig.createConsoleTransport
+      : true || process.env.LOG_TO_CONSOLE == "true"
```
nicholaspai, line 51: Doesn't this always resolve to `true`?

```
true || process.env.LOG_TO_CONSOLE == "true"
```

mrice32: I'm a little confused here as well.

packages/financial-templates-lib/src/logger/Transports.ts
```diff
@@ -45,7 +45,11 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
     const { LoggingWinston } = require("@google-cloud/logging-winston");
     transports.push(new LoggingWinston());
     if (!require("@google-cloud/trace-agent").get().enabled) require("@google-cloud/trace-agent").start();
-  } else if (transportsConfig.createConsoleTransport != undefined ? transportsConfig.createConsoleTransport : true) {
+  } else if (
+    transportsConfig.createConsoleTransport != undefined
```
mrice32, line 49: Is this the desired condition or am I missing some defaulting logic where undefined is assumed to be true or something?
```ts
} else if (transportsConfig.createConsoleTransport || process.env.LOG_TO_CONSOLE == "true") {
```

mrice32 (commented):
Just a minor question about the implementation!

packages/financial-templates-lib/src/logger/Transports.ts
chrismaree: Ye, you're right. this can be simplified to what you showed. I made this PR during a meeting trying to solve a bug running the relayers in prod :) 

thanks for the callout. I should have marked this PR as draft :)

packages/financial-templates-lib/src/logger/Transports.ts
chrismaree: I've refactored this. sorry it was overly verbose before!

packages/financial-templates-lib/src/logger/Transports.ts
chrismaree: it actually needs to be like this:
```
} else if (transportsConfig?.createConsoleTransport || process.env.LOG_TO_CONSOLE == "true") {
```

to re-implement the same functionality as what I had before:
1) if `createConsoleTransport` is defined then use `createConsoleTransport` value else, if `LOG_TO_CONSOLE` then true

chrismaree:
This functionality is not needed any more. closing PR.

PR #3880, opened 2022-04-19
feat: add event based request path for OO

mrice32:
**Motivation**

We would like the OO to handle event-based requests.

Note: test changes are not included yet, as I wanted to nail down the contracts first.

**Summary**

Rather than create a new flavor of the OO, it seemed somewhat simple to add this feature to the existing OO using an additional boolean in the Request struct.

For non event-based requests, this should add a few branches, but cost no real additional gas. We could potentially make an optimization to how we measure the length of the stamped ancillary data to decrease gas costs for those requests.

Event-based queries will be more expensive. They will involve an additional call to the OO, and they will require a more costly ancillary data stamping/construction when calling into the DVM.

This is how the event-based process works:
1. A contract that wants its resolution to happen after some event occurs can request a price immediately upon its construction.
2. The bool is set by calling setEventBased.
3. Now that the request is an "event based" request, it's now semantically considered a pre-request. Its timestamp no longer dictates what information can be considered. These types of proposals also disallow the "too early" value, INT_MIN. This prevents users from receiving a reward for proposing a correct "too early" value before the event has happened.
4. When disputers consider requests, they need to consider the proposal timestamp rather than the request timestamp to determine whether the request is too early.
5. When a request is sent to the DVM, an extra piece of data "proposalTimestamp" is included to indicate when the proposal happened so the voters can determine if the proposal was premature.
6. Once a request settles, if the "too early" result is returned, the contract can re-request knowing that the proposer that initiated the resolution was penalized.

Open questions: 
1. Generally the DVM works by only considering information before the _request timestamp_. This process implicitly invalidates that rule. Should we, instead, use the proposal timestamp as the request timestamp that is sent to the DVM instead of including it in ancillary data?
2. Is this simpler than having an entirely separate OO flavor? In my opinion, it seems simpler, as the additional complexity is pretty minimal and it is almost 100% backwards compatible (only exception is the getRequest return value that has an additional boolean in it).

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -347,10 +347,11 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         );
 
         // End the re-entrancy guard early to allow the caller to potentially take OO-related actions inside this callback.
-        _notEntered = true;
+        _startReentrantGuardDisabled();
         // Callback.
         if (address(requester).isContract())
             try OptimisticRequester(requester).priceProposed(identifier, timestamp, ancillaryData) {} catch {}
+        _endReentrantGuardDisabled();
     }
 
     /**
@@ -427,7 +428,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
 
         _getOracle().requestPrice(
             identifier,
-            _getDvmTimestamp(request, timestamp),
+            _getTimestampForDvmRequest(request, timestamp),
             _stampAncillaryData(ancillaryData, requester)
         );
 
@@ -450,10 +451,11 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         );
 
         // End the re-entrancy guard early to allow the caller to potentially re-request inside this callback.
-        _notEntered = true;
+        _startReentrantGuardDisabled();
         // Callback.
         if (address(requester).isContract())
             try OptimisticRequester(requester).priceDisputed(identifier, timestamp, ancillaryData, refund) {} catch {}
+        _endReentrantGuardDisabled();
     }
 
     /**
@@ -611,7 +613,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
             // In the Resolved case, pay either the disputer or the proposer the entire payout (+ bond and reward).
             request.resolvedPrice = _getOracle().getPrice(
                 identifier,
-                _getDvmTimestamp(request, timestamp),
+                _getTimestampForDvmRequest(request, timestamp),
                 _stampAncillaryData(ancillaryData, requester)
             );
             bool disputeSuccess = request.resolvedPrice != request.proposedPrice;
@@ -643,14 +645,13 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         );
 
         // Temporarily disable the re-entrancy guard early to allow the caller to take an OO-related action inside this callback.
-        _notEntered = true;
+        _startReentrantGuardDisabled();
         // Callback.
         if (address(requester).isContract())
             try
                 OptimisticRequester(requester).priceSettled(identifier, timestamp, ancillaryData, request.resolvedPrice)
             {} catch {}
-        // Re-enable this guard since the calling function may take other actions that need to be guarded after this.
-        _notEntered = false;
+        _endReentrantGuardDisabled();
     }
 
     function _getRequest(
@@ -699,7 +700,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         return
             _getOracle().hasPrice(
                 identifier,
-                _getDvmTimestamp(request, timestamp),
+                _getTimestampForDvmRequest(request, timestamp),
                 _stampAncillaryData(ancillaryData, requester)
             )
                 ? State.Resolved
@@ -722,7 +723,11 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         return IdentifierWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.IdentifierWhitelist));
     }
 
-    function _getDvmTimestamp(Request storage request, uint256 requestTimestamp) internal view returns (uint256) {
+    function _getTimestampForDvmRequest(Request storage request, uint256 requestTimestamp)
+        internal
+        view
+        returns (uint256)
+    {
         if (request.eventBased) {
             uint256 liveness = request.customLiveness != 0 ? request.customLiveness : defaultLiveness;
             return request.expirationTime.sub(liveness);
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -19,6 +19,8 @@ import "../../common/implementation/FixedPoint.sol";
 import "../../common/implementation/AncillaryData.sol";
 import "../../common/implementation/AddressWhitelist.sol";
 
+import "hardhat/console.sol";
+
 /**
  * @title Optimistic Requester.
  * @notice Optional interface that requesters can implement to receive callbacks.
@@ -697,7 +699,11 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         }
 
         return
-            _getOracle().hasPrice(identifier, timestamp, _stampAncillaryData(ancillaryData, requester))
+            _getOracle().hasPrice(
+                identifier,
+                _getDvmTimestamp(request, timestamp),
+                _stampAncillaryData(ancillaryData, requester)
+            )
                 ? State.Resolved
                 : State.Disputed;
     }
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -76,6 +76,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
     using SafeMath for uint256;
     using SafeERC20 for IERC20;
     using Address for address;
+    using AncillaryData for bytes;
 
     event RequestPrice(
         address indexed requester,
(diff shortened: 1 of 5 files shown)


nicholaspai:
> Its timestamp no longer dictates what information can be considered. This also means that proposals cannot be INT_MIN.

Why is this true?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -285,6 +305,7 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
             "proposePriceFor: Requested"
         );
         Request storage request = _getRequest(requester, identifier, timestamp, ancillaryData);
+        if (request.eventBased) require(proposedPrice != type(int256).min, "Cannot propose 'too early'");
```
nicholaspai, line 308: This now puts the magic number into the smart contract which is kind of a big deal since we can no longer change the magic number off-chain. This isn't a huge deal, but makes me wonder: why is this check that proposedPrice != "too early" even important? Can't we just let disputers/UMA voters just reject any magic number price for an event-based request?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -163,8 +164,11 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         require(_getIdentifierWhitelist().isIdentifierSupported(identifier), "Unsupported identifier");
         require(_getCollateralWhitelist().isOnWhitelist(address(currency)), "Unsupported currency");
         require(timestamp <= getCurrentTime(), "Timestamp in future");
+
+        // TODO: to make this cheaper, we should just compute the max length of the stamping and add it to the
+        // ancillary data length.
```
nicholaspai, line 169: Can you explain this change? Are you basically passing in the longest possible ancillary data length and checking that its under the limit?

nicholaspai (commented):
This is pretty cool. I agree with you that we don't need a new OO for this feature. I think we should consider removing the proposalTimestamp from the ancillary data and just sending that to the DVM and replacing the useless request timestamp for event-based price requests.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Yep, exactly. In the future, I'd like to just get a constant value for the max added length and add it to avoid the costs of constructing the ancillary data during this call.

mrice32:
> > Its timestamp no longer dictates what information can be considered. This also means that proposals cannot be INT_MIN.
> 
> Why is this true?

@nicholaspai I think my language was just unclear. The former does not imply the latter. The contract just enforces the latter. I added a sentence explaining why the contract should enforce that.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: It's a great question.

I think it's important because of the following situation.

1. Requesting contract requests the price on startup. This requests includes a reward for a correct response.
2. Instead of waiting for the event to occur, a clever proposer could just propose the "too early" value immediately. This doesn't actual provide anything valuable that's worthy of the reward. However, this is technically correct since the DVM would also indicate that this price request was submitted too early, making any disputer lose money for disputing it.
3. The proposer is, therefore, not disputed, and receives the reward. The requesting contract would need to resubmit a new request to get the actual value after the event occurs.

smb2796: Could this instead be enforced by the requesting contract using the proposePrice callback?

smb2796:
> Once a request settles, if the "too early" result is returned, the contract can re-request knowing that the proposer that initiated the resolution was penalized.

Is the expectation that requestors would also call `setRefundOnDispute` after submitting the initial request? Thoughts on this happening as part of the `setEventBased` call?

mrice32:
> > Once a request settles, if the "too early" result is returned, the contract can re-request knowing that the proposer that initiated the resolution was penalized.
> 
> Is the expectation that requestors would also call `setRefundOnDispute` after submitting the initial request? Thoughts on this happening as part of the `setEventBased` call?

@smb2796 as I was writing up the OZ scope document, I thought the same thing. I think this makes sense.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -259,6 +264,21 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         _getRequest(msg.sender, identifier, timestamp, ancillaryData).customLiveness = customLiveness;
     }
 
+    // Allows the requester to set the request to be an "event based" request.
+    // This means that "too early" values are disallowed in proposals and the
+    // proposal timestamp is appended to the ancillary data.
```
chrismaree, line 269: nit: natspec.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
chrismaree: it is already put into the contracts in other places. see: https://github.com/UMAprotocol/protocol/blob/6368e90125c3b1ae445ff0ae3c3b38b63efaa72a/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L348 where we do this in the LPS (the origin of the "magic number" was this contract in this place).

I personally think this is fine and is the standard already converged on in at least two places (including external inergrations). the reasoning for using `type(int256).min` around being a very unlikely price, in fact the most unlikely price, still holds true.

nicholaspai: Ok good point @chrismaree good to leave as is

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -19,6 +19,8 @@ import "../../common/implementation/FixedPoint.sol";
 import "../../common/implementation/AncillaryData.sol";
 import "../../common/implementation/AddressWhitelist.sol";
 
+import "hardhat/console.sol";
```
nicholaspai, line 22: leftover hardhat console

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -259,6 +266,37 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         _getRequest(msg.sender, identifier, timestamp, ancillaryData).customLiveness = customLiveness;
     }
 
+    /**
+     * @notice Sets the request to be an "event-based" request.
+     * @dev Calling this method has a few impacts on the request:
+     *
+     * 1. The timestamp at which the request is evaludated is the time of the proposal, not the timestamp associated
```
nicholaspai, line 273:
```suggestion
     * 1. The timestamp at which the request is evaluated is the time of the proposal, not the timestamp associated
```

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -309,6 +348,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
             address(request.currency)
         );
 
+        // End the re-entrancy guard early to allow the caller to potentially take OO-related actions inside this callback.
+        _notEntered = true;
```
nicholaspai, line 352: How did you catch this bug? This is a good find

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -597,11 +644,15 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
...
 
+        // Temporarily disable the re-entrancy guard early to allow the caller to take an OO-related action inside this callback.
+        _notEntered = true;
         // Callback.
         if (address(requester).isContract())
             try
                 OptimisticRequester(requester).priceSettled(identifier, timestamp, ancillaryData, request.resolvedPrice)
             {} catch {}
+        // Re-enable this guard since the calling function may take other actions that need to be guarded after this.
+        _notEntered = false;
```
nicholaspai, line 655: Why not set this to false in the other locations where you set `_notEntered = true` before a callback?

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Fixed.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -669,6 +724,15 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
         return IdentifierWhitelistInterface(finder.getImplementationAddress(OracleInterfaces.IdentifierWhitelist));
     }
 
+    function _getDvmTimestamp(Request storage request, uint256 requestTimestamp) internal view returns (uint256) {
```
nicholaspai, line 727: Might be more accurate to name this `_getTimestampForDvmRequest`

nicholaspai: Because otherwise I would think that this would return the current time of the DVM

nicholaspai (approved):
LGTM, just left some last comments! Feel free to merge after resolving

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: This method is called inside of another method, so there's other code that can run after this.

In the other cases, it's immediately followed by the end of the function. I can set it in those cases too just to be safe and consistent, though.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: It wasn't a bug, we just didn't allow you to call back into this contract.

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Done!

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Even though it's a little redundant, I tried to make this a little clearer and consistent. Let me know what you think.

nicholaspai: think better to be a bit consistent

nicholaspai: actually maybe its just worth saving the gas since its clear its near end of function. Instead of closing it you could replace it with a comment explaining why you don't reset _notEntered. WDYT?

PR #3881, opened 2022-04-19
feat: use regular oo instead of skinny oo

pemulis:

**Motivation**

Moving from the Skinny Optimistic Oracle to the regular Optimistic Oracle to have a better user experience.

**Summary**

Dealing with requests in calldata is a bad UX.

**Details**

For the most part, this PR simplifies the existing code since the regular OO makes fewer demands on the caller.

**Testing**

- [x]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -185,19 +185,6 @@ contract OptimisticGovernor is Module, Lockable {
 
         // Propose a set of transactions to the OO. If not disputed, they can be executed with executeProposal().
         // docs: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
-
-        // optimisticOracle.requestAndProposePriceFor(
-        //     identifier,
-        //     uint32(time),
-        //     ancillaryData,
-        //     collateral,
-        //     0,
-        //     bond,
-        //     uint256(liveness),
-        //     address(this),
-        //     // Canonical value representing "True"; i.e. the transactions are valid.
-        //     int256(1e18)
-        // );
         optimisticOracle.requestPrice(identifier, time, ancillaryData, collateral, 0);
         uint256 totalBond = optimisticOracle.setBond(identifier, time, ancillaryData, bond);
         optimisticOracle.setCustomLiveness(identifier, time, ancillaryData, liveness);
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -9,9 +9,8 @@ import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
 import "../oracle/implementation/Constants.sol";
 import "../oracle/interfaces/FinderInterface.sol";
-import "../oracle/interfaces/SkinnyOptimisticOracleInterface.sol";
-import "../oracle/implementation/SkinnyOptimisticOracle.sol";
-import "../oracle/interfaces/OracleAncillaryInterface.sol";
+import "../oracle/interfaces/OptimisticOracleInterface.sol";
+import "../oracle/implementation/OptimisticOracle.sol";
 import "../common/implementation/Lockable.sol";
 import "../common/interfaces/AddressWhitelistInterface.sol";
 import "../oracle/interfaces/IdentifierWhitelistInterface.sol";
@@ -35,7 +34,7 @@ contract OptimisticGovernor is Module, Lockable {
         bytes32 indexed identifier,
         uint256 indexed timestamp,
         bytes ancillaryData,
-        SkinnyOptimisticOracleInterface.Request request
+        OptimisticOracleInterface.Request request
     );
 
     event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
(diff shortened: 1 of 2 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -192,23 +189,35 @@ contract OptimisticGovernor is Module, Lockable {
...
-        uint256 totalBond = finalFee + bond;
         collateral.safeTransferFrom(msg.sender, address(this), totalBond);
-        collateral.safeIncreaseAllowance(address(skinnyOptimisticOracle), totalBond);
+        collateral.safeIncreaseAllowance(address(optimisticOracle), totalBond);
 
-        // Propose a set of transactions to the OO. If not disputed, they can be executed with executeProposal().
-        // docs: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/interfaces/SkinnyOptimisticOracleInterface.sol
-        skinnyOptimisticOracle.requestAndProposePriceFor(
+        optimisticOracle.proposePriceFor(
+            address(this),
```
mrice32, line 216: I think the proposer should be `msg.sender`, not `this`. This is because we want the returned bond and any rewards to be paid out to them, not this contract.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -218,8 +227,7 @@ contract OptimisticGovernor is Module, Lockable {
     function executeProposal(
         uint256 _proposalId,
         Transaction[] memory _transactions,
-        uint32 _originalTime,
-        SkinnyOptimisticOracleInterface.Request memory _request
+        uint32 _originalTime
```
mrice32, line 230: Something we may want to consider in a follow-up is just storing this timestamp so the user doesn't have to resubmit it, as it may be annoying for the user or the tooling to determine the block timestamp of the original proposal.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -301,7 +305,7 @@ contract OptimisticGovernor is Module, Lockable {
     }
 
     function _sync() internal {
-        skinnyOptimisticOracle = _getOptimisticOracle();
+        optimisticOracle = _getOptimisticOracle();
         finalFee = _getStore().computeFinalFee(address(collateral)).rawValue;
```
mrice32, line 309: Is finalFee used anywhere? Can we drop it?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -192,23 +189,35 @@ contract OptimisticGovernor is Module, Lockable {
...
+        //     0,
+        //     bond,
+        //     uint256(liveness),
+        //     address(this),
+        //     // Canonical value representing "True"; i.e. the transactions are valid.
+        //     int256(1e18)
+        // );
+        optimisticOracle.requestPrice(identifier, uint32(time), ancillaryData, collateral, 0);
+
+        uint256 totalBond = optimisticOracle.setBond(identifier, uint32(time), ancillaryData, bond);
```
mrice32, line 209: nit: you probably also need to make a call to set the liveness for this request.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -35,7 +34,7 @@ contract OptimisticGovernor is Module, Lockable {
         bytes32 indexed identifier,
         uint256 indexed timestamp,
         bytes ancillaryData,
-        SkinnyOptimisticOracleInterface.Request request
+        OptimisticOracleInterface.Request request
```
mrice32, line 37: Should this parameter just be dropped entirely?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -218,8 +227,7 @@ contract OptimisticGovernor is Module, Lockable {
     function executeProposal(
         uint256 _proposalId,
         Transaction[] memory _transactions,
-        uint32 _originalTime,
-        SkinnyOptimisticOracleInterface.Request memory _request
+        uint32 _originalTime
     ) public payable nonReentrant {
         // Recreate the proposal hash from the inputs and check that it matches the stored proposal hash.
         uint256 id = _proposalId;
```
mrice32, line 233: Minor note on something you didn't change:

Instead of:
```
bytes memory ancillaryData = bytes.concat(bytes("id:"), AncillaryData.toUtf8BytesUint(id));
```

You should be able to use the built-in function to do exactly this:
```
bytes memory ancillaryData = AncillaryData.appendKeyValueUint("", "id", id);
```

I think what you had probably worked as intended, but this is a little more standard across our repo.

mrice32 (commented):
Looks great! Only a few minor comments!

packages/core/contracts/zodiac/OptimisticGovernor.sol
pemulis: Looks like we can drop it now.

packages/core/contracts/zodiac/OptimisticGovernor.sol
pemulis: yep

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -192,40 +183,44 @@ contract OptimisticGovernor is Module, Lockable {
         // proposalHashes[id] = keccak256(abi.encodePacked(proposalData));
         proposalHashes[id] = keccak256(abi.encode(_transactions));
 
+        // Propose a set of transactions to the OO. If not disputed, they can be executed with executeProposal().
+        // docs: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
+
+        // optimisticOracle.requestAndProposePriceFor(
```
mrice32, line 189: nit: should we remove this comment?

mrice32 (approved):
LGTM!

packages/core/contracts/zodiac/OptimisticGovernor.sol
pemulis: yep, removing

PR #3882, opened 2022-04-20
fix(sdk): syncronize oo client index and details page

daywiss:

**Motivation**
https://app.shortcut.com/uma-project/story/4883/have-the-app-refresh-more-frequently-so-the-state-of-proposals-is-more-accurately-reflected-without-the-user-needing-to-refresh

The index page was not updating correctly when sending proposal/deposits in detail page.

**Summary**

This strips events of transaction receipt when proposign/disputing and adds them to the internal event list and updates the index state.  also fixes potential bug where polling for new events was not working.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Tested this on kovan using already pending requests. tested proposal and dispute and both worked to update index page accordingly

### DIFF
--- a/packages/sdk/src/oracle/services/statemachines/disputePrice.ts
+++ b/packages/sdk/src/oracle/services/statemachines/disputePrice.ts
@@ -1,9 +1,9 @@
 import assert from "assert";
 import { Update } from "../update";
 import Store from "../../store";
-import { Signer } from "../../types/ethers";
+import { Signer, TransactionReceipt } from "../../types/ethers";
 import { Handlers as GenericHandlers } from "../../types/statemachine";
-import { InputRequest } from "../../types/state";
+import { InputRequest, OptimisticOracleEvent } from "../../types/state";
 import { ContextClient } from "./utils";
 
 export type Params = InputRequest & {
@@ -14,7 +14,7 @@ export type Params = InputRequest & {
   checkTxIntervalSec: number;
 };
 
-export type Memory = { hash?: string };
+export type Memory = { hash?: string; receipt?: TransactionReceipt };
 
 export function initMemory(): Memory {
   return {};
@@ -36,7 +36,9 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
       const { chainId, confirmations, checkTxIntervalSec } = params;
       const { hash } = memory;
       assert(hash, "requires hash");
-      if (await update.isConfirmed(chainId, hash, confirmations)) {
+      const receipt = await update.isConfirmed(chainId, hash, confirmations);
+      if (receipt) {
+        memory.receipt = receipt as TransactionReceipt;
         return "update";
       }
       // wait x seconds before running this state again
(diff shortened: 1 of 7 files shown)


packages/sdk/src/oracle/services/statemachines/disputePrice.ts
```diff
@@ -36,23 +36,35 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
...
+        if (receipt) {
+          const events = receipt.logs.map(oracle.makeEventFromLog);
+          events.forEach((event) => {
+            w.chains(chainId)
+              .optimisticOracle()
+              .event((event as unknown) as OptimisticOracleEvent);
+          });
+        }
+      });
+      await update.sortedRequests(chainId);
```
daywiss, line 67: the problem was that when you triggered a transaction for propose or dispute from the detail page, we did not push the events from the tx receipt into the same event list that the index page was using to generate its full list of requests. So thats added here for  both types of transactions.

daywiss, line 67: the problem was that when you triggered a transaction for propose or dispute from the detail page, we did not push the events from the tx receipt into the same event list that the index page was using to generate its full list of requests. So thats added here for  both types of transactions.

packages/sdk/src/oracle/services/statemachines/proposePrice.ts
```diff
@@ -36,23 +36,35 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
...
+        if (receipt) {
+          const events = receipt.logs.map(oracle.makeEventFromLog);
+          events.forEach((event) => {
+            w.chains(chainId)
+              .optimisticOracle()
+              .event((event as unknown) as OptimisticOracleEvent);
+          });
+        }
+      });
+      await update.sortedRequests(chainId);
```
daywiss, line 67: exact same changes as the dispute logic

daywiss, line 67: exact same changes as the dispute logic

packages/sdk/src/oracle/store/write.ts
```diff
@@ -81,7 +81,10 @@ export class OptimisticOracle {
   }
   event(event: state.OptimisticOracleEvent): void {
     if (!this.state.events) this.state.events = [];
-    insertOrderedAscending(this.state.events, event, eventKey);
+    // avoid duplicates
+    if (isUnique(this.state.events, event, eventKey)) {
+      insertOrderedAscending(this.state.events, event, eventKey);
+    }
```
daywiss, line 87: since there are multiple things writing to these events now, wanted to make sure we dont push in duplicates (though it probably wouldnt matter)

daywiss, line 87: since there are multiple things writing to these events now, wanted to make sure we dont push in duplicates (though it probably wouldnt matter)

packages/sdk/src/oracle/services/statemachines/pollNewEvents.ts
```diff
@@ -46,6 +46,9 @@ export function Handlers(store: Store): GenericHandlers<Params, Memory> {
           memory.lastBlock = latestBlock;
           // just count how many successful iterations we do as a kind of sanity check
           memory.iterations++;
+        } else {
+          // if we dont have a lastblock set, set it to our current block
+          memory.lastBlock = memory.lastBlock || currentBlock;
```
daywiss, line 51: this was a bug, since memory.lastblock was never initialized, it was always "currentBlock" which mean the logic above this never ran. so now if its not set, set it to the current block and it will run next iteration

mrice32 (approved):
LGTM!

amateima (approved):

PR #3883, opened 2022-04-20
fix(sdk): improve oracle data loading

daywiss:
**Motivation**

Loading polygon requests are slow, since there are so many. 

**Summary**

This adds some levers to pull to change the performance characteristics of loading requests. You can now restrict the max block range, as well as the earliest block to ignore.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

Tested this by playing with the OO UI locally and have seen a large speed up in performance with the right parameters.

### DIFF
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -11,12 +11,7 @@ import { defaultConfig } from "./utils";
 
 export class Client {
   private intervalStarted = false;
-  constructor(
-    public readonly store: Store,
-    public readonly update: Update,
-    public readonly sm: StateMachine,
-    public readonly poller: StateMachine
-  ) {}
+  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {}
   setUser(params: Partial<User>): string {
     const address = params.address && ethers.utils.getAddress(params.address);
     return this.sm.types.setUser.create({ ...params, address });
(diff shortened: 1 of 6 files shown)


packages/sdk/src/oracle/utils.ts
```diff
@@ -212,25 +212,45 @@ export type RangeState = {
  * @param {Pick} state
  * @returns {RangeState}
  */
-export function rangeStart(state: Pick<RangeState, "startBlock" | "endBlock" | "multiplier">): RangeState {
+export function rangeStart(
```
daywiss, line 215: this initialization function is split into 2 blocks, one where a user specifies a max range, the other where no max range is specified and its deteremined by start/end blocks

packages/sdk/src/oracle/client.ts
```diff
@@ -176,20 +170,21 @@ export function factory(config: state.PartialConfig, emit: Emit): Client {
 
   // this first state machine is for user actions
   const sm = new StateMachine(store);
-  // this one is system actions used for long running commands independent of the user
-  const poller = new StateMachine(store);
 
   // start the request list checkers
   for (const [chainId, config] of Object.entries(fullConfig.chains)) {
-    poller.types.fetchPastEvents.create({ chainId: Number(chainId), startBlock: config.earliestBlockNumber }, "poller");
+    sm.types.fetchPastEvents.create(
```
daywiss, line 176: this is the command that was optimized, you can now configure the client to specify a max range to query and a block number threshold to ignore blocks before this number

amateima (approved):
LGTM

PR #3885, opened 2022-04-25
fix: fix minor issues with OG and OD

mrice32:
**Motivation**

There were a few minor issues that needed resolution in both the Optimistic Governor and Optimistic Distributor.

**Summary**

Optimistic Distributor:
- Final fee and Store were unused.

Optimistic Governor:
- The deleteRejectedProposals method had a vulnerability where ancillary data was not being validated correctly.
- Natspec was missing on almost all methods.
- The setCollateral method was calling _sync when it didn't need to.
- The setCollateral method was nonReentrant, but that would prevent it being set via executions that originated from the Optimistic Governor.
- Remove unnecessary event, since the OptimisticOracle event contains the same info.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
+++ b/packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
@@ -67,9 +67,6 @@ contract OptimisticDistributor is Lockable, MultiCaller, Testable {
     uint256 public constant MINIMUM_LIVENESS = 10 minutes;
     uint256 public constant MAXIMUM_LIVENESS = 5200 weeks;
 
-    // Final fee can be synced and stored in the contract.
-    uint256 public finalFee;
-
     // Ancillary data length limit can be synced and stored in the contract.
     uint256 public ancillaryBytesLimit;
 
@@ -88,7 +85,6 @@ contract OptimisticDistributor is Lockable, MultiCaller, Testable {
     MerkleDistributor public merkleDistributor;
 
     // Interface parameters that can be synced and stored in the contract.
-    StoreInterface public store;
     OptimisticOracleInterface public optimisticOracle;
 
     /********************************************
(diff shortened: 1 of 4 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -231,19 +260,26 @@ contract OptimisticGovernor is Module, Lockable {
...
-        bytes memory _ancillaryData
-    ) public {
+    /**
+     * @notice Method to allow anyone to delete a proposal that was rejected.
+     * @param _proposalId the id of the proposal being deleted.
+     * @param _originalTime the id of the proposal being deleted.
+     */
+    function deleteRejectedProposal(uint256 _proposalId, uint256 _originalTime) public {
+        // Construct the ancillary data.
+        bytes memory ancillaryData = AncillaryData.appendKeyValueUint("", "id", _proposalId);
```
mrice32, line 279: Without this, there's no guarantee that the ancillary data matches the proposal id, and any rejected proposal would allow anyone to delete any proposal.

nicholaspai (approved):
Nice fixes lgtm

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -67,9 +67,6 @@ contract OptimisticDistributor is Lockable, MultiCaller, Testable {
     uint256 public constant MINIMUM_LIVENESS = 10 minutes;
     uint256 public constant MAXIMUM_LIVENESS = 5200 weeks;
 
-    // Final fee can be synced and stored in the contract.
-    uint256 public finalFee;
```
Reinis-FRP, line 71: Agree, initially I did not fully realize OO methods return total bond amount that makes syncing final fee redundant.

PR #3886, opened 2022-04-26
improve: improve loading speed of request details

daywiss:

**Motivation**

Loading request details can be slow even if requests were previously visited.

**Summary**
This mainly splits up setting the active request, and updating the active request to 2 separate steps, and now only waits for the active request to be set before finishing "loading". This lets it show cached data if available. It also reduces the amount of data needed when updating the active request.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/client.ts
+++ b/packages/sdk/src/oracle/client.ts
@@ -11,7 +11,12 @@ import { defaultConfig } from "./utils";
 
 export class Client {
   private intervalStarted = false;
-  constructor(public readonly store: Store, public readonly update: Update, public readonly sm: StateMachine) {}
+  constructor(
+    public readonly store: Store,
+    public readonly update: Update,
+    public readonly sm: StateMachine,
+    public readonly poller: StateMachine
+  ) {}
   setUser(params: Partial<User>): string {
     const address = params.address && ethers.utils.getAddress(params.address);
     return this.sm.types.setUser.create({ ...params, address });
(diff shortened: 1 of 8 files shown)


packages/sdk/src/oracle/client.ts
```diff
@@ -26,10 +31,14 @@ export class Client {
     const identifier = params.identifier.toLowerCase();
     const chainId = Number(params.chainId);
     const timestamp = Number(params.timestamp);
-    return this.sm.types.setActiveRequest.create({ requester, ancillaryData, identifier, chainId, timestamp });
+    const result = this.sm.types.setActiveRequest.create({ requester, ancillaryData, identifier, chainId, timestamp });
+    this.sm.types.updateActiveRequest.create(undefined);
```
mrice32, line 35: I don't completely understand what's going on here, but this seems a little odd? Why do we create or update the active request with undefined? That seems like either a no-op or a delete?

daywiss: a couple things here. Having to pass undefined is a sideeffect of typescript and the way I'm exporting these parameter arguments to create a template.  The `updateActiveRequest` command takes no arguments because it pulls the active request from global state.  You can see here is the file that exports the "template" https://github.com/UMAprotocol/protocol/pull/3886/files#diff-20099ec61981af2c2ad5d2f6b9aecc64715656e69ca5f6552e3f4f5f6079ebd0R7 and the "parameters" are undefined.

Typescript will fail to compile if i leave this blank when calling "create", so i have to explicitly pass undefined.

PR #3890, opened 2022-04-27
fix: add v1 OO interface for backcompat for downstream users

mrice32:
**Motivation**

We need to preserve the old interface to continue interacting with the deployed OO.

**Summary**

This adds a "V1" interface that should be used when interacting with the existing deployed OO. The SDK client was also updated.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A


daywiss (approved):
thanks!

PR #3898, opened 2022-05-04
feat: Round realized LP fee % to match UMIP

nicholaspai:
**Motivation**

[A change to UMIP-136](https://github.com/UMAprotocol/UMIPs/pull/515) will require that realized LP fee %'s are accurate to 6 decimals. This PR changes the fee calculator to always return realized LP fees to this precision.

**Summary**
Reviewers, there are two questions I have:
- Currently I am changing the precision of the realized LP fee % in the `sdk`, but I wonder if this should be lower at the client level?
- Does there exist a common function we can use to implement the precision change? I don't like manipulating the string as I am doing as it seems error prone.

**Testing**

- [X]  New unit tests created
- [ ]  Untested


mrice32 (approved):
LGTM!

PR #3900, opened 2022-05-05
improve: release

daywiss:

**Motivation**

releases some fixes for oo client


nicholaspai (approved):
Thanks for doing this!

PR #3901, opened 2022-05-08
fix: split l1 block queries at specific block

mrice32:
Hotfix to split queries since there are now over 10k events.

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -374,32 +374,46 @@ export class InsuredBridgeL1Client {
     // (restarting all the time) and b) this value changes very infrequently.
     if (this.optimisticOracleLiveness == 0)
       this.optimisticOracleLiveness = Number(await this.bridgeAdmin.methods.optimisticOracleLiveness().call());
+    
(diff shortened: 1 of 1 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -374,32 +374,46 @@ export class InsuredBridgeL1Client {
     // (restarting all the time) and b) this value changes very infrequently.
     if (this.optimisticOracleLiveness == 0)
       this.optimisticOracleLiveness = Number(await this.bridgeAdmin.methods.optimisticOracleLiveness().call());
+    
+    const paginateBlockNumber = Number(process.env.L1_SPLIT_BLOCK_NUMBER);
```
nicholaspai, line 378:
```suggestion
    // This is a hacked solution to fix the problem where there are too many instances of the following events returned by a Infura, which limits return values to 10,000. So, if there are more than 10,000 "DepositRelayed" events returned by the event search, then we need to split up the search at this block. This will not be a solution once we hit 20,000 events so this is just a temporary fix.
    const paginateBlockNumber = Number(process.env.L1_SPLIT_BLOCK_NUMBER);
```

nicholaspai (approved):

PR #3902, opened 2022-05-08
fix: Paginate InsuredBridgeL1Client event search for select events

nicholaspai:
**Motivation**

Some event searches return > 10,000 values in production, we should paginate such event searches to avoid running into node-provider imposed limit.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,5 +74,6 @@
     "packages": [
       "packages/*"
     ]
-  }
+  },
+  "dependencies": {}
 }
--- a/packages/across-monitor/src/RelayEventProcessor.ts
+++ b/packages/across-monitor/src/RelayEventProcessor.ts
@@ -1,4 +1,5 @@
-import type { BridgePoolData } from "@uma/financial-templates-lib";
+import type { BridgePoolData, InsuredBridgeL1Client } from "@uma/financial-templates-lib";
+import { EventData } from "web3-eth-contract";
 
 // Note that Relay parameters are narrower in this module than in @uma/financial-templates-lib.
 interface Relay {
@@ -39,14 +40,14 @@ export class RelayEventProcessor {
     RelayCanceled: "canceled",
   };
 
-  constructor() {
+  constructor(readonly l1Client: InsuredBridgeL1Client) {
     // do nothing.
   }
 
-  // update() should be called only after having populated bridgePools.
+  // update() should be called only after having updated L1 client
   // This fetches all DepositRelayed events since lastRelayUpdate block in order to store deposit and relay hash
   // mappings that could be referenced by other events.
-  async update(endingBlock: number): Promise<void> {
+  update(endingBlock: number): void {
     // Only update if endingBlock is more recent than the lastRelayUpdate block from last update() call.
     const startingBlock = this.lastRelayUpdate + 1;
     if (startingBlock > endingBlock) {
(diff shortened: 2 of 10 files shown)


packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -13,6 +13,12 @@ export const bridgeDepositBoxDeployData = {
   288: { blockNumber: 223808 },
 };
 
+// Following settings can be overridden to optimize L1 client event search for select events. For example,
+// the "DepositRelayed" event search request could return > 10,000 return values so we need to shorten the block
+// search using these parameters because some node providers like Infura limit event search return values to 10,000.
+export const bridgePoolEarliestBlockToSearch = 13545377;
```
mrice32, line 19: How did you determine this number?

packages/insured-bridge-relayer/src/RelayerConfig.ts
```diff
@@ -13,6 +13,12 @@ export const bridgeDepositBoxDeployData = {
...
+export const bridgePoolMaxBlocksToSeach = 1_000_000;
```
mrice32, line 20: Roughly how many of these windows do you think we'll need to search through to get through the whole range?

mrice32 (commented):
This is awesome! Will take a closer look later today!

packages/insured-bridge-relayer/src/RelayerConfig.ts
nicholaspai: oldest spoke pool's deployment block number, so the weth pool's deploy block. No events for any spoke pools should be before this

packages/insured-bridge-relayer/src/RelayerConfig.ts
nicholaspai: 2, there are 1.1mil blocks since deployment block

package.json
```diff
@@ -74,5 +74,6 @@
     "packages": [
       "packages/*"
     ]
-  }
+  },
+  "dependencies": {}
```
nicholaspai, line 78: No idea why this was added but added after `yarn` and `qbuild`

packages/insured-bridge-relayer/src/RelayerConfig.ts
chrismaree: in v2 we have this on a per spoke basis which helps, rather than grouping them all together.

chrismaree (approved):

packages/across-monitor/src/index.ts
```diff
@@ -25,21 +25,29 @@ export async function run(logger: winston.Logger, l1Web3: Web3): Promise<void> {
...
+    );
 
     const acrossMonitor = new AcrossMonitor(logger, config, l1Client);
 
     for (;;) {
       await retry(
         async () => {
           // Updating acrossMonitor also updates l1Client for pool discovery.
-          await acrossMonitor.update();
+          acrossMonitor.update();
```
chrismaree, line 44: I'm slightly confused. now that `acrossMonitor.update()` no longer pulls in the event data and rather reads everything from the `l1Client` via `this.l1Client.allRelayEventData`, where is `await l1Client.update()` being called? i.e where is the monitor now actually pulling in the event data, via the l1Client?

mrice32 (approved):
LGTM outside of @chrismaree's comment!

Should we have @Reinis-FRP take a look as well?

packages/insured-bridge-relayer/src/RelayerConfig.ts
nicholaspai: I think its actually pretty safe to group them all together since the BridgePools are always deployed to L1 and we can guarantee no Relay events before this block, and the lookback is ~1.1 million blocks to this block which should be fine for quite some time

mrice32 (approved):

packages/across-monitor/src/RelayEventProcessor.ts
```diff
@@ -58,58 +59,24 @@ export class RelayEventProcessor {
...
-            slowRelayFeePct: depositRelayedEvent.returnValues.depositData.slowRelayFeePct,
-            instantRelayFeePct: depositRelayedEvent.returnValues.depositData.instantRelayFeePct,
-            realizedLpFeePct: depositRelayedEvent.returnValues.relay.realizedLpFeePct,
-            depositHash: depositRelayedEvent.returnValues.depositHash,
-          };
-          this.deposits[l1TokenAddress][depositRelayedEvent.returnValues.depositHash] = relayData;
-        }
-      })
-    );
+    for (const relayData of this.l1Client.getAllRelayedDeposits()) {
```
Reinis-FRP, line 62: `l1Client` deletes pending relays if they are disputed, hence `getRelayEventInfo` could throw as `depositHash` emitted on dispute could not be matched to Relay data.

Reinis-FRP: yes, this would indeed help. and assume it would store all deposits from the earliest block

Reinis-FRP (approved):
Looks good and `getAllRelayedDepositsSimple` on `l1Client` now makes the monitor bot much simpler!

PR #3906, opened 2022-05-13
fix(sdk/across): scope query events to user

daywiss:

**Motivation**

event queries are timing out in across

**Summary**

scope event queries to active user

**Testing**

Ran this locally and saw error go away

- [ ]  New unit tests created
- [ ]  Untested


chrismaree (approved):
this will make a really big difference to the amount of calls done and the data returned.

PR #3909, opened 2022-05-17
fix: [N09] Unexplained and unused constants

mrice32:
**Motivation**

Throughout the OptimisticGovernor contract, to check if a proposal has been approved
by the Optimistic Oracle the literal value int256 1e18 is used, where 1e18 signifies that a
proposal was not rejected by the Optimistic Oracle.

Similarly, in the update to the OptimisticOracle contract the function proposedPrice
uses a magic value type(int256).min to indicate that an event-based proposal cannot be
resolved, because the event has not yet taken place.

Lastly, in the OptimisticOracle contract the MAX_ADDED_ANCILLARY_DATA constant is
declared on line 129. On the next line the constant should be used, but instead the value of the
constant is used directly to derive another constant.

To improve the overall readability of the codebase and to facilitate refactoring, consider
defining a constant for every literal or magic value used, giving it a clear and self-explanatory
name, and then using it in place of literal values. Also consider adding an inline comment
explaining how literal values were calculated or why they were chosen.

**Summary**

Fixed these issues.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
+++ b/packages/core/contracts/oracle/implementation/OptimisticOracle.sol
@@ -127,7 +127,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
 
     // This is effectively the extra ancillary data to add ",ooRequester:0000000000000000000000000000000000000000".
     uint256 private constant MAX_ADDED_ANCILLARY_DATA = 53;
-    uint256 public constant OO_ANCILLARY_DATA_LIMIT = ancillaryBytesLimit - 53;
+    uint256 public constant OO_ANCILLARY_DATA_LIMIT = ancillaryBytesLimit - MAX_ADDED_ANCILLARY_DATA;
+    int256 public constant TOO_EARLY_RESPONSE = type(int256).min;
 
     /**
      * @notice Constructor.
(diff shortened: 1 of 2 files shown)


packages/core/contracts/oracle/implementation/OptimisticOracle.sol
```diff
@@ -127,7 +127,8 @@ contract OptimisticOracle is OptimisticOracleInterface, Testable, Lockable {
 
     // This is effectively the extra ancillary data to add ",ooRequester:0000000000000000000000000000000000000000".
     uint256 private constant MAX_ADDED_ANCILLARY_DATA = 53;
-    uint256 public constant OO_ANCILLARY_DATA_LIMIT = ancillaryBytesLimit - 53;
+    uint256 public constant OO_ANCILLARY_DATA_LIMIT = ancillaryBytesLimit - MAX_ADDED_ANCILLARY_DATA;
+    int256 public constant TOO_EARLY_RESPONSE = type(int256).min;
```
chrismaree, line 131: would be cool to refactor this in a way that we can re-use this magic number throughout some of the other contracts in this package such as here https://github.com/UMAprotocol/protocol/blob/faa54932d9a5bb4dc46bc5337def8707ff44159d/packages/core/contracts/financial-templates/long-short-pair/LongShortPair.sol#L348

chrismaree (approved):
LGTM

packages/core/contracts/oracle/implementation/OptimisticOracle.sol
mrice32: Agree! But let's do this in a follow-up just to keep these fixes focused and simple.

PR #3911, opened 2022-05-17
fix: add ability to delete disputed proposals

pemulis:

**Motivation**

This PR resolves a potential issue. Right now, a proposal that is disputed can be executed if the dispute is resolved in the proposer's favor. This can be a problem because it could allow for things like multiple withdrawals or stale smart contract updates. Here are two failure examples:

**Double Withdrawal**

1. Alice creates a valid proposal to withdraw funds.
2. Proposal is disputed.
3. Alice creates a second valid proposal to withdraw funds.
4. Second proposal is not disputed and funds are withdrawn.
5. First proposal is validated and funds are withdrawn a second time.

**Stale Smart Contract Update**

1. Alice creates a valid proposal to update an interest rate in a smart contract system controlled by an Avatar.
2. Proposal is disputed.
3. The interest rate changes several times during the 2-3 days for dispute resolution.
4. Original proposal is validated.
5. A stale interest rate update can now be executed.

The fix for this would be to add a `deleteDisputedProposal` and allow for deletion of any proposal that was disputed. It can always be re-submitted, so the contract does not need to wait for resolution of the original proposal to do whatever action. The DVM can rule that valid proposals were valid at the time they were proposed (or not, as the case may be) and settle the bonds appropriately between the proposer and disputer, but the data returned will not be used by the OptimisticGovernor contract because the proposal was deleted. This prevents the failure modes outlined above.

Contrary to initial discussion of the issue, I left the `deleteRejectedProposal` function in place. This can be used to delete rejected proposals that had not already been deleted during the dispute phase, though they should really be deleted earlier.

There were also various improvements suggested by @mrice32 that simplify the proposal and execution system.

**Summary**

- Adds a `deleteDisputedProposal` function.
- Maps proposal hashes to proposal time, rather than proposal IDs to proposal hashes.
- Removes proposal IDs since they are now superfluous.
- Simplifies parameters for proposals and execution based on the new proposal hash mapping.
- Adds a check that rejects duplicate proposals.
- Comments out tests temporarily since they will need to be re-written.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -264,7 +264,7 @@ contract OptimisticGovernor is Module, Lockable {
         // You can not execute a proposal that has been disputed at some point in the past.
         require(
             optimisticOracle.getRequest(address(this), identifier, _originalTime, ancillaryData).disputer == address(0),
-            "Must call deleteRejectedProposal instead"
+            "Must call deleteDisputedProposal instead"
         );
 
         // Remove proposal hash so transactions can not be executed again.
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -69,6 +69,7 @@ contract OptimisticGovernor is Module, Lockable {
         uint256 requestTime;
     }
 
+    // This maps proposal hashes to the proposal timestamps.
     mapping(bytes32 => uint256) public proposalHashes;
 
     /**
@@ -208,6 +209,9 @@ contract OptimisticGovernor is Module, Lockable {
         // Create the proposal hash.
         bytes32 proposalHash = keccak256(abi.encode(_transactions));
 
+        // Add the proposal hash to ancillary data.
+        bytes memory ancillaryData = AncillaryData.appendKeyValueBytes32("", "proposalHash", proposalHash);
+
         // Check that the proposal is not already mapped to a proposal time, i.e., is not a duplicate.
         require(proposalHashes[proposalHash] == 0, "Duplicate proposals are not allowed");
 
@@ -216,9 +220,9 @@ contract OptimisticGovernor is Module, Lockable {
 
         // Propose a set of transactions to the OO. If not disputed, they can be executed with executeProposal().
         // docs: https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/oracle/interfaces/OptimisticOracleInterface.sol
-        optimisticOracle.requestPrice(identifier, time, "", collateral, 0);
-        uint256 totalBond = optimisticOracle.setBond(identifier, time, "", bondAmount);
-        optimisticOracle.setCustomLiveness(identifier, time, "", liveness);
+        optimisticOracle.requestPrice(identifier, time, ancillaryData, collateral, 0);
+        uint256 totalBond = optimisticOracle.setBond(identifier, time, ancillaryData, bondAmount);
+        optimisticOracle.setCustomLiveness(identifier, time, ancillaryData, liveness);
 
         // Get the bond from the proposer and approve the bond and final fee to be used by the oracle.
         // This will fail if the proposer has not granted the OptimisticGovernor contract an allowance
@@ -226,7 +230,14 @@ contract OptimisticGovernor is Module, Lockable {
         collateral.safeTransferFrom(msg.sender, address(this), totalBond);
         collateral.safeIncreaseAllowance(address(optimisticOracle), totalBond);
 
-        optimisticOracle.proposePriceFor(msg.sender, address(this), identifier, time, "", PROPOSAL_VALID_RESPONSE);
+        optimisticOracle.proposePriceFor(
+            msg.sender,
+            address(this),
+            identifier,
+            time,
+            ancillaryData,
+            PROPOSAL_VALID_RESPONSE
+        );
 
         uint256 challengeWindowEnds = time + liveness;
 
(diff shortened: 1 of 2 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -263,16 +263,22 @@ contract OptimisticGovernor is Module, Lockable {
...
 
     /**
      * @notice Method to allow anyone to delete a proposal that was rejected.
      * @param _proposalId the id of the proposal being deleted.
-     * @param _originalTime the id of the proposal being deleted.
+     * @param _originalTime the time the proposal was made.
      */
     function deleteRejectedProposal(uint256 _proposalId, uint256 _originalTime) public {
+        // Check that proposal exists and was not already deleted.
+        require(proposalHashes[_proposalId] != "", "Proposal does not exist");
```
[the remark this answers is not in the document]
pemulis: This (and the other one) were fixes for a different issue, as were a few other changes in this PR, but I wanted to stay consistent with the audit fixes that were happening in parallel. Basically, this should work cleanly with the other fixes if we decide to merge it in, but the other fixes don't depend on the changes in this PR and could be merged separately, since we aren't sure if this is the right direction yet.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -263,16 +263,22 @@ contract OptimisticGovernor is Module, Lockable {
      * @param _proposalId the id of the proposal being deleted.
      */
     function deleteProposal(uint256 _proposalId) public onlyOwner {
+        // Check that proposal exists and was not already deleted.
+        require(proposalHashes[_proposalId] != "", "Proposal does not exist");
```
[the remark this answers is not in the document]
pemulis: This (and the other one) were fixes for a different issue, as were a few other changes in this PR, but I wanted to stay consistent with the audit fixes that were happening in parallel. Basically, this should work cleanly with the other fixes if we decide to merge it in, but the other fixes don't depend on the changes in this PR and could be merged separately, since we aren't sure if this is the right direction yet.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -70,8 +69,7 @@ contract OptimisticGovernor is Module, Lockable {
         uint256 requestTime;
     }
 
-    mapping(uint256 => bytes32) public proposalHashes;
-    uint256 public prevProposalId;
+    mapping(bytes32 => uint256) public proposalHashes;
```
mrice32, line 72: nit: I would either add a comment indicating that this maps to timestamps or change the name of the mapping to proposalTimestamps.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -190,18 +188,13 @@ contract OptimisticGovernor is Module, Lockable {
...
-        prevProposalId = id;
         uint256 time = getCurrentTime();
         address proposer = msg.sender;
 
         // Create proposal in memory to emit in an event.
         Proposal memory proposal;
         proposal.requestTime = time;
 
-        // Construct the ancillary data.
-        bytes memory ancillaryData = AncillaryData.appendKeyValueUint("", "id", id);
```
mrice32, line 203: Instead of dropping ancillary data, should we, instead, append the proposal hash? Without this, you get a weird problem: you can't make two proposals in the same block, since the requests would be identical (same timestamp and ancillary data) regardless of the transactions. Not a huge issue, but would be confusing for a user. It also just seems more comforting for a voter to have some identifying information in the ancillary data, so it's super obvious what they're voting on.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -212,59 +205,53 @@ contract OptimisticGovernor is Module, Lockable {
...
+        // Get the original proposal time.
+        uint256 _originalTime = proposalHashes[_proposalHash];
 
         // Remove proposal hash so transactions can not be executed again.
-        delete proposalHashes[id];
+        delete proposalHashes[_proposalHash];
 
         // This will revert if the price has not been settled and can not currently be settled.
-        int256 price = optimisticOracle.settleAndGetPrice(identifier, _originalTime, ancillaryData);
+        int256 price = optimisticOracle.settleAndGetPrice(identifier, _originalTime, "");
```
mrice32, line 254: If we want to ensure that this contract never allows disputed proposals to be used for settlement, we should call check that getRequest shows no disputer.
```
require(optimisticOracle.getRequest(address(this), identifier, _originalTime, ancillaryData).disputer == address(0), "Must call deleteRejectedProposal instead");
```

mrice32 (commented):
A few minor comments, but once those are resolved, I think we should be good to go!

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -237,31 +241,34 @@ contract OptimisticGovernor is Module, Lockable {
...
-        require(proposalHashes[id] == keccak256(abi.encode(_transactions)), "proposal hash does not match");
+        require(proposalHashes[_proposalHash] != 0, "proposal hash does not exist");
+
+        // Get the original proposal time.
+        uint256 _originalTime = proposalHashes[_proposalHash];
+
+        // You can not execute a proposal that has been disputed at some point in the past.
+        require(
+            optimisticOracle.getRequest(address(this), identifier, _originalTime, ancillaryData).disputer == address(0),
+            "Must call deleteRejectedProposal instead"
```
mrice32, line 267:
```suggestion
            "Must call deleteDisputedProposal instead"
```

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -308,8 +318,34 @@ contract OptimisticGovernor is Module, Lockable {
...
+
+        // Get the original proposal time.
+        uint256 _originalTime = proposalHashes[_proposalHash];
+
+        // Get the state of the proposal.
+        OptimisticOracleInterface.State state =
+            optimisticOracle.getState(address(this), identifier, _originalTime, ancillaryData);
+
+        // Check that proposal was disputed.
+        require(state == OptimisticOracleInterface.State.Disputed, "Proposal was not disputed");
```
mrice32, line 344:
```suggestion
        OptimisticOracleInterface.Request request =
            optimisticOracle.getRequest(address(this), identifier, _originalTime, ancillaryData);

        // Check that proposal was disputed.
        require(request.disputer != address(0), "Proposal was not disputed");
```

mrice32 (approved):
Two suggestions (one from a typo on my part). LGTM!

mrice32 (approved):

PR #3912, opened 2022-05-17
fix: M-01 Change of collateral could result in unintended bond value

pemulis:

**Motivation**

Updating the bond and collateral in the same function prevents the possibility of a gap between updating the collateral type and updating the bond, which could result in a proposal with an incorrect bond size.

**Summary**

Creates a `setCollateralAndBond` function that sets the collateral and bond simultaneously and remove the two separate functions.

**Details**

This PR also adds events for `SetBond` and `SetCollateral` in line with the fixes for M-02.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -113,22 +117,20 @@ contract OptimisticGovernor is Module, Lockable {
...
+     * @notice Sets the collateral and bond amount for proposals.
      * @param _collateral token that will be used for all bonds for the contract.
+     * @param _bond amount of the bond token that will need to be paid for future proposals.
      */
-    function setCollateral(IERC20 _collateral) public onlyOwner {
+    function setCollateralAndBond(IERC20 _collateral, uint256 _bond) public onlyOwner {
         // ERC20 token to be used as collateral (must be approved by UMA Store contract).
         require(_getCollateralWhitelist().isOnWhitelist(address(_collateral)), "bond token not supported");
         collateral = _collateral;
+        emit SetCollateral(_collateral);
```
chrismaree, line 128: nit: also emit `_bond`

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -113,22 +117,20 @@ contract OptimisticGovernor is Module, Lockable {
...
+
+        // Value of the bond required for proposals, in addition to the final fee. A bond of zero is
+        // acceptable, in which case the Optimistic Oracle will require the final fee as the bond.
+        bond = _bond;
+        emit SetBond(_bond);
```
chrismaree, line 133: why use two seperate events? prehaps make an event `SetBond` with `collateral` and `bond` amount

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -113,22 +117,20 @@ contract OptimisticGovernor is Module, Lockable {
...
-    }
-
-    /**
-     * @notice Sets the collateral token (and bond token) for future proposals.
+     * @notice Sets the collateral and bond amount for proposals.
      * @param _collateral token that will be used for all bonds for the contract.
+     * @param _bond amount of the bond token that will need to be paid for future proposals.
      */
-    function setCollateral(IERC20 _collateral) public onlyOwner {
+    function setCollateralAndBond(IERC20 _collateral, uint256 _bond) public onlyOwner {
```
chrismaree, line 124: a better variable name might be `_bondAmount`

packages/core/contracts/zodiac/OptimisticGovernor.sol
mrice32: +1

packages/core/contracts/zodiac/OptimisticGovernor.sol
mrice32: +1

mrice32 (commented):
LGTM once @chrismaree comments are resolved!

packages/core/contracts/zodiac/OptimisticGovernor.sol
nicholaspai, line 128: nit: should emit all events at the end at the same time, but I also think you should consoldiate these events into one. There's no advantage to emitting an event here, changing state after, then emitting another

nicholaspai (approved):
LGTM once you resolve comments

pemulis:
@mrice32 @nicholaspai running into the same issue where the automated tests are failing for reasons unrelated to this PR but I don't have the ability to re-run the tests. Could I be added to the group that can trigger a re-run of the test suite? @evaldofelipe

PR #3913, opened 2022-05-17
fix: M-02, lack of event emission after sensitive actions

pemulis:

**Motivation**

In many places, events are not emitted after executing sensitive actions. This adds events for updates to the bond, collateral, rules, liveness, and identifier.

**Summary**

Events are defined and emitted in various functions.

**Details**

No events were added to the `setUp` function in this PR, but the PR for L-02 will have `setUp` call the various set* functions internally, which will emit the events. Similarly, the PR for M-01 includes events for the combined `setCollateralAndBond` function.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -150,7 +150,7 @@ contract OptimisticGovernor is Module, Lockable {
     function setRules(string memory _rules) public onlyOwner {
         // Set reference to the rules for the avatar (e.g. an IPFS hash or URI).
         rules = _rules;
-        emit SetRules(rules);
+        emit SetRules(_rules);
     }
 
     /**
@@ -162,7 +162,7 @@ contract OptimisticGovernor is Module, Lockable {
         // Set liveness for disputing proposed transactions.
         require(_liveness > 0, "liveness can't be 0");
         liveness = _liveness;
-        emit SetLiveness(liveness);
+        emit SetLiveness(_liveness);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


chrismaree (approved):

mrice32 (approved):
LGTM! Worth noting that these events are emitted in setUp once this change is combined with #3914.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -138,6 +150,7 @@ contract OptimisticGovernor is Module, Lockable {
     function setRules(string memory _rules) public onlyOwner {
         // Set reference to the rules for the avatar (e.g. an IPFS hash or URI).
         rules = _rules;
+        emit SetRules(rules);
```
nicholaspai, line 153: nit: emit `_rules` which is an MLOAD instead of `rules` which is an `SLOAD`

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -149,6 +162,7 @@ contract OptimisticGovernor is Module, Lockable {
         // Set liveness for disputing proposed transactions.
         require(_liveness > 0, "liveness can't be 0");
         liveness = _liveness;
+        emit SetLiveness(liveness);
```
nicholaspai, line 165: Same comment as above

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -160,6 +174,7 @@ contract OptimisticGovernor is Module, Lockable {
         // Set identifier which is used along with the rules to determine if transactions are valid.
         require(_getIdentifierWhitelist().isIdentifierSupported(_identifier), "identifier not supported");
         identifier = _identifier;
+        emit SetIdentifier(identifier);
```
nicholaspai, line 177: Same comment as above

nicholaspai (approved):
Left one tiny gas optimization trick

PR #3914, opened 2022-05-17
fix: L-02, Duplicated code

pemulis:

**Motivation**

This PR uses the `set*` functions in the `setUp` function to set the bond, collateral, identifier, and so on, instead of duplicating code.

**Summary**

Deletes duplicate code and calls the `set*` functions in `setUp`.

**Details**

The `setCollateralAndBond` function from the fix for M-01 is carried into this PR since it is used in the `setUp` function.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -34,6 +34,10 @@ contract OptimisticGovernor is Module, Lockable {
 
     event ProposalDeleted(uint256 indexed proposalId);
 
+    event SetCollateral(IERC20 indexed collateral);
+
+    event SetBond(uint256 indexed bond);
+
     // Since finder is set during setUp, you will need to deploy a new Optimistic Governor module if this address need to be changed in the future.
     FinderInterface public finder;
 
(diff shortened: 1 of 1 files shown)


chrismaree (approved):

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -34,6 +34,10 @@ contract OptimisticGovernor is Module, Lockable {
 
     event ProposalDeleted(uint256 indexed proposalId);
 
+    event SetCollateral(IERC20 indexed collateral);
+
+    event SetBond(uint256 indexed bond);
```
mrice32, line 39: nit: this might cause merge conflicts... aren't these added in a different fix?

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -96,14 +100,10 @@ contract OptimisticGovernor is Module, Lockable {
             uint64 _liveness
         ) = abi.decode(initializeParams, (address, address, address, uint256, string, bytes32, uint64));
         finder = FinderInterface(_finder);
-        require(_getCollateralWhitelist().isOnWhitelist(address(_collateral)), "bond token not supported");
```
nicholaspai, line 99: In future FYI, have all require statements come before state changes. Two reasons: 
- gas savings on reverts: less gas used paid by caller by exiting earlier.
- reentrancy,  not an issue here, but if you can re-enter with the same state as the previous state, you can do evil things

nicholaspai (commented):
As matt said, think you need to rebase this PR to remove conflicts

PR #3915, opened 2022-05-17
fix: M-03, Lack of input validation

pemulis:

**Motivation**

The `OptimisticGovernor` would benefit from more rigorous input validation.

**Summary**

Checks that `_finder` is nonzero, that `_liveness` is less than 5,200 weeks, and that `_rules` is non-empty, in various setter functions.

**Details**

A bond set to zero is valid. This is clarified in inline comments in the PR for M-01.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -70,6 +70,7 @@ contract OptimisticGovernor is Module, Lockable {
      * @param _rules Reference to the rules for the Gnosis Safe (e.g., IPFS hash or URI).
      * @param _identifier The approved identifier to be used with the contract, usually "ZODIAC".
      * @param _liveness The period, in seconds, in which a proposal can be disputed.
+     * @dev if the bondAmount is zero, there will be no reward for disputers, reducing incentives to dispute invalid proposals.
      */
     constructor(
         address _finder,
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -95,6 +95,7 @@ contract OptimisticGovernor is Module, Lockable {
             bytes32 _identifier,
             uint64 _liveness
         ) = abi.decode(initializeParams, (address, address, address, uint256, string, bytes32, uint64));
+        require(_finder != address(0), "finder address can not be empty");
         finder = FinderInterface(_finder);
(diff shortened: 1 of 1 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -95,6 +95,7 @@ contract OptimisticGovernor is Module, Lockable {
             bytes32 _identifier,
             uint64 _liveness
         ) = abi.decode(initializeParams, (address, address, address, uint256, string, bytes32, uint64));
+        require(_finder != address(0), "finder address can not be empty");
         finder = FinderInterface(_finder);
         require(_getCollateralWhitelist().isOnWhitelist(address(_collateral)), "bond token not supported");
```
chrismaree, line 100: can you group the `requires` all together at the top of the function rather than doing require then assignment.

mrice32: @chrismaree I think this gets removed in a different fix, so we should avoid moving it to avoid a merge conflict.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -95,6 +95,7 @@ contract OptimisticGovernor is Module, Lockable {
             bytes32 _identifier,
             uint64 _liveness
```
mrice32, line 96: OZ commented on the 0-bond case. If we don't intend to block a 0 bond, we should probably at least add a comment explaining that a 0 bond is allowed, but will mean that there is no reward for disputes.

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -148,6 +150,7 @@ contract OptimisticGovernor is Module, Lockable {
     function setLiveness(uint64 _liveness) public onlyOwner {
         // Set liveness for disputing proposed transactions.
         require(_liveness > 0, "liveness can't be 0");
+        require(_liveness < 3144960000, "liveness can't be longer than 5200 weeks");
```
nicholaspai, line 153: you can just write `52 weeks` pretty sure

nicholaspai: https://docs.soliditylang.org/en/v0.8.10/units-and-global-variables.html#time-units

mrice32 (approved):

PR #3916, opened 2022-05-18
fix: L-01, Events lacking information

pemulis:

**Motivation**

`ProposalDeleted` and `TransactionsProposed` events would be more useful if they included more information.

**Summary**

Adds the `msg.sender` and status to the `ProposalDeleted` event and adds a `challengeWindowEnds` value to the `TransactionsProposed` event, so that disputers know by what time they must submit disputes.

**Details**

A separate PR adds a `deleteDisputedProposal` function, which has a "Disputed" status in the `ProposalDeleted` event.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -27,12 +27,13 @@ contract OptimisticGovernor is Module, Lockable {
         address indexed proposer,
         uint256 indexed proposalTime,
         Proposal proposal,
-        bytes explanation
+        bytes explanation,
+        uint256 challengeWindowEnds
     );
 
     event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
 
-    event ProposalDeleted(uint256 indexed proposalId);
+    event ProposalDeleted(uint256 indexed proposalId, address indexed sender, bytes32 indexed status);
 
     // Since finder is set during setUp, you will need to deploy a new Optimistic Governor module if this address need to be changed in the future.
     FinderInterface public finder;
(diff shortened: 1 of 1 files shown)


mrice32 (approved):

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -27,12 +27,13 @@ contract OptimisticGovernor is Module, Lockable {
...
         Proposal proposal,
-        bytes explanation
+        bytes explanation,
+        uint256 challengeWindowEnds
     );
 
     event TransactionExecuted(uint256 indexed proposalId, uint256 indexed transactionIndex);
 
-    event ProposalDeleted(uint256 indexed proposalId);
+    event ProposalDeleted(uint256 indexed proposalId, address indexed sender, bytes32 indexed status);
```
nicholaspai, line 36: Should this be `bytes` or `string`? Why bytes32

pemulis: The actual answer is "muscle memory," where it's been deeply ingrained in me to use `bytes32` instead of `string` wherever I know the length will not exceed 32 bytes. The reason for that, historically, is that `bytes32` was easier for smart contracts to handle internally than `string` and that there were gas savings in storing `bytes32` values.

I haven't looked into the matter recently, so it's possible the string handling issues have been resolved, and it's also possible that the gas savings are a moot point for event data. I'm still inclined to keep this as `bytes32` in case the event data is fed into a different smart contract somewhere else where it will actually be stored, since I know for sure that none of the values being returned here are longer than 32 bytes.

PR #3918, opened 2022-05-18
fix: L-04, Proposals can be deleted repeatedly

pemulis:

**Motivation**

Before this PR, proposal deletion functions could be called repeatedly for the same proposal. This wastes gas, so there is no incentive to do so, but it could result in multiple `ProposalDeleted` events, which could be confusing.

**Summary**

Checks that proposal still exists before deleting the proposal and emitting an event.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -277,7 +277,7 @@ contract OptimisticGovernor is Module, Lockable {
      */
     function deleteRejectedProposal(uint256 _proposalId, uint256 _originalTime) public {
         // Check that proposal exists and was not already deleted.
-        require(proposalHashes[_proposalId] != "", "Proposal does not exist");
+        require(proposalHashes[_proposalId] != bytes32(0), "Proposal does not exist");
 
         // Construct the ancillary data.
         bytes memory ancillaryData = AncillaryData.appendKeyValueUint("", "id", _proposalId);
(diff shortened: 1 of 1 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -263,6 +263,9 @@ contract OptimisticGovernor is Module, Lockable {
      * @param _proposalId the id of the proposal being deleted.
      */
     function deleteProposal(uint256 _proposalId) public onlyOwner {
+        // Check that proposal exists and was not already deleted.
+        require(proposalHashes[_proposalId] != "", "Proposal does not exist");
```
mrice32, line 267:
```suggestion
        require(proposalHashes[_proposalId] != bytes32(0), "Proposal does not exist");
```

packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -273,6 +276,9 @@ contract OptimisticGovernor is Module, Lockable {
      * @param _originalTime the id of the proposal being deleted.
      */
     function deleteRejectedProposal(uint256 _proposalId, uint256 _originalTime) public {
+        // Check that proposal exists and was not already deleted.
+        require(proposalHashes[_proposalId] != "", "Proposal does not exist");
```
mrice32, line 280:
```suggestion
        require(proposalHashes[_proposalId] != bytes32(0), "Proposal does not exist");
```

mrice32 (commented):
Just minor suggestions to make the check for an uninitialized bytes32 value more explicit.

nicholaspai (approved):

mrice32 (approved):

PR #3921, opened 2022-05-18
fix: N-04, immutable value could be used

pemulis:

**Motivation**

Since the `finder` is only set up on initialization, it could be made `immutable` and set in the constructor rather than the `setUp` function. This more clearly signals intent and reduces gas costs.

**Summary**

Set `finder` in constructor and make `immutable`.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -35,7 +35,7 @@ contract OptimisticGovernor is Module, Lockable {
     event ProposalDeleted(uint256 indexed proposalId);
 
     // Since finder is set during setUp, you will need to deploy a new Optimistic Governor module if this address need to be changed in the future.
-    FinderInterface public finder;
+    FinderInterface public immutable finder;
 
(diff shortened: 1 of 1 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -80,22 +80,22 @@ contract OptimisticGovernor is Module, Lockable {
         bytes32 _identifier,
         uint64 _liveness
     ) {
-        bytes memory initializeParams = abi.encode(_finder, _owner, _collateral, _bond, _rules, _identifier, _liveness);
+        bytes memory initializeParams = abi.encode(_owner, _collateral, _bond, _rules, _identifier, _liveness);
+        require(_finder != address(0), "finder address can not be empty");
```
mrice32, line 84: Was this check added in a different PR? Watch out for merge conflicts.

mrice32 (approved):

PR #3925, opened 2022-05-18
fix: N-08, Undocumented implicit approval requirements

pemulis:

**Motivation**

The `proposeTransactions` function will fail if the proposer has not set an appropriate allowance for the `OptimisticGovernor` contract but this is not documented in the code.

**Summary**

Adds inline documentation explaining that the function will fail if a proper allowance is not set.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/core/contracts/zodiac/OptimisticGovernor.sol
+++ b/packages/core/contracts/zodiac/OptimisticGovernor.sol
@@ -211,6 +211,8 @@ contract OptimisticGovernor is Module, Lockable {
         optimisticOracle.setCustomLiveness(identifier, time, ancillaryData, liveness);
 
         // Get the bond from the proposer and approve the bond and final fee to be used by the oracle.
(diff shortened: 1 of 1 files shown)


packages/core/contracts/zodiac/OptimisticGovernor.sol
```diff
@@ -211,6 +211,8 @@ contract OptimisticGovernor is Module, Lockable {
         optimisticOracle.setCustomLiveness(identifier, time, ancillaryData, liveness);
 
         // Get the bond from the proposer and approve the bond and final fee to be used by the oracle.
+        // This will fail if the proposer has not granted the OptimisticGovernor contract an allowance
```
mrice32, line 214: I would also suggest adding an `@dev` section in the natspec that says this too.

mrice32 (approved):

nicholaspai (approved):

PR #3931, opened 2022-05-24
feat: extend serverless to contain more configs for multi-chain

chrismaree:
**Motivation**

Across v2 monitor requires to know information about a number of chains latest blocks and last blocks seen. This PR updates the hub to add addtional storage slots to enable this.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -95,17 +95,39 @@ hub.post("/", async (req, res) => {
       // Check if bot is running on a non-default chain, and fetch last block number seen on this or the default chain.
       const [botWeb3, spokeCustomNodeUrl] = _getWeb3AndUrlForBot(configObject[botName]);
       const chainId = await _getChainId(botWeb3);
-      nodeUrlToChainIdCache[spokeCustomNodeUrl] = chainId;
-
-      // If we've seen this chain ID already we can skip it:
+      // If we've seen this chain ID already we can skip it.
       if (blockNumbersForChain[chainId]) continue;
 
-      // Fetch last seen block for this chain:
-      let lastQueriedBlockNumber = await _getLastQueriedBlockNumber(req.body.configFile, chainId, logger);
+      nodeUrlToChainIdCache[spokeCustomNodeUrl] = chainId;
 
-      // Next, get the head block for the chosen chain, which we'll use to override the last queried block number
+      // If STORE_MULTI_CHAIN_BLOCK_NUMBERS is set then this bot requires to know a number of last seen blocks across
+      // a set of chainIds. Construct a batch promise to evaluate the latest block number for each chainId.
+      if (configObject[botName]?.environmentVariables?.STORE_MULTI_CHAIN_BLOCK_NUMBERS) {
+        const multiChainIds = configObject[botName]?.environmentVariables?.STORE_MULTI_CHAIN_BLOCK_NUMBERS;
+        let promises = [];
+        for (const chainId of multiChainIds) {
+          promises.push(
+            _getLastQueriedBlockNumber(req.body.configFile, chainId, logger),
+            _getBlockNumberOnChainIdMultiChain(configObject[botName], chainId)
+          );
+        }
+        let results = await Promise.all(promises);
+        results.forEach((_, index) => {
+          if (index % 2 !== 0) return;
+          const chainId = multiChainIds[Math.floor(index / 2)];
+          blockNumbersForChain[chainId] = {
+            lastQueriedBlockNumber: results[index],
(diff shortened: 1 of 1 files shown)


packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -95,17 +95,39 @@ hub.post("/", async (req, res) => {
...
+        let promises = [];
+        for (const chainId of multiChainIds) {
+          promises.push(
+            _getLastQueriedBlockNumber(req.body.configFile, chainId, logger),
+            _getBlockNumberOnChainIdMultiChain(configObject[botName], chainId)
+          );
+        }
+        let results = await Promise.all(promises);
+        results.forEach((_, index) => {
+          if (index % 2 !== 0) return;
```
nicholaspai, line 116: Why not be more struct and just assert that the results are a multiple of two?

nicholaspai (approved):

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -408,6 +434,10 @@ function _getWeb3AndUrlForBot(botConfig) {
   }
 }
 
+async function _getBlockNumberOnChainIdMultiChain(botConfig, chainId) {
+  return await new Web3(botConfig?.environmentVariables[`NODE_URL_${chainId}`]).eth.getBlockNumber();
+}
```
mrice32, line 439: nit: any reason not to use the retry config logic as done in _getWeb3AndUrlForBot (we can always do this in a follow up)?

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -95,17 +95,39 @@ hub.post("/", async (req, res) => {
...
+        for (const chainId of multiChainIds) {
+          promises.push(
+            _getLastQueriedBlockNumber(req.body.configFile, chainId, logger),
+            _getBlockNumberOnChainIdMultiChain(configObject[botName], chainId)
+          );
+        }
+        let results = await Promise.all(promises);
+        results.forEach((_, index) => {
+          if (index % 2 !== 0) return;
+          const chainId = multiChainIds[Math.floor(index / 2)];
```
mrice32, line 117: nit: why is math floor needed here? Shouldn't these always be even?

packages/serverless-orchestration/src/ServerlessHub.js
```diff
@@ -95,17 +95,39 @@ hub.post("/", async (req, res) => {
...
+        const multiChainIds = configObject[botName]?.environmentVariables?.STORE_MULTI_CHAIN_BLOCK_NUMBERS;
+        let promises = [];
+        for (const chainId of multiChainIds) {
+          promises.push(
+            _getLastQueriedBlockNumber(req.body.configFile, chainId, logger),
+            _getBlockNumberOnChainIdMultiChain(configObject[botName], chainId)
+          );
+        }
+        let results = await Promise.all(promises);
+        results.forEach((_, index) => {
```
mrice32, line 115: nit: I think we could drop some of this even odd logic if we were to use a map. (you would lose the parallelization of the inner calls, but I don't think that's a huge deal, and could be compensated for if you did an inner Promise.all)

```ts
const blockNumbersForChain = await Promise.all(multiChainIds.map(async (chainId) => ({
  lastQueriedBlockNumber: await _getLastQueriedBlockNumber(req.body.configFile, chainId, logger),
  latestBlockNumber: await _getBlockNumberOnChainIdMultiChain(configObject[botName], chainId),
})));
```

Not necessary to do now. Could be done in follow up.

mrice32 (approved):
LGTM! Only nits

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: `_getWeb3AndUrlForBot` expects a format of provider that we dont use in across. this is janky but is done to avoid needing to modify the bot configs to have redundant information. we ONLY support `NODE_URL_n` in the relayer right now, which `_getWeb3AndUrlForBot` does not work with.

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: o ye, you right. the line above actually protects this from not being the case.

packages/serverless-orchestration/src/ServerlessHub.js
chrismaree: I originally had it exactly like this! I changed it to this structure to be more parallel. the problem is that there are 5 chains that would have some level of blocking if done the way you have it now. there will be 5 pairs of 2 Promise.all calls. my structure has 1 set of 10 calls, all at once.

PR #3933, opened 2022-05-24
fix(merkle-distributor): [H12] track claimed rewards per window

Reinis-FRP:
**Motivation**

Audit identified the following issue:
_The MerkleDistributor unnecessarily allows for separate rewards to be stolen from each other if a single merkle tree allows to withdraw more rewards than were provided to that tree._

**Summary**

This PR fixes identified issue by adding `remainingAmount` parameter to each window that is decreased with each claim. Claims that cause it to underflow are reverting.

**Testing**

- [x]  New unit tests created
- [ ]  Untested


packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
```diff
@@ -248,6 +238,8 @@ contract MerkleDistributor is Ownable {
 
     // Verify claim is valid and mark it as completed in this contract.
     function _verifyAndMarkClaimed(Claim memory _claim) private {
+        // Check claimed rewards do not exceed remaining funding for the window.
+        require(_claim.amount <= merkleWindows[_claim.windowIndex].remainingAmount, "Insufficient funding for window");
```
chrismaree, line 242: I dont think this is strictly required as the call on line 175 would revert if there was not enough but it's better to have a more descriptive revert message than just `underflow`

chrismaree (approved):
this is good, makes sence and is a clean solution. I'm not really sure why we didnt just do this in the first place... if you have a block of 14 lines explaining a known issue in the contracts it should likely be fixed •͡˘㇁•͡˘

@nicholaspai do you see any issues with what we've proposed here? I know you worked closely on this particular part of the code.

packages/core/contracts/merkle-distributor/implementation/MerkleDistributor.sol
mrice32: Agree. We should drop this as it's an unnecessary gas expenditure.

mrice32 (approved):
Looks good!

I'm not sure I 100% agree that this is needed.

Before this change, you were trusting the owner to make sure they deposit enough funds such that the sum of all leaves of all roots are funded.

After this change, you're trusting the owner to make sure they deposit enough funds such that the sum of the remainingAmount values are funded.

In the former, one root could steal from another if one had a leaf with a very large amount value. In the latter you could have remainingAmount value that is set very high.

The trust assumptions feel the same IMO. I would argue the only thing that would make this trustless would be to drop the withdrawRewards, deleteWindow methods and require that funds are pulled from the caller in setWindow.

@Reinis-FRP happy to approve as is, but it feels a bit like we're adding gas for no real change in trust assumptions.

Reinis-FRP:
> The trust assumptions feel the same IMO. I would argue the only thing that would make this trustless would be to drop the withdrawRewards, deleteWindow methods and require that funds are pulled from the caller in setWindow.

Agree to your judgement on trust assumptions when owner is EOA, but this considerably improves security when owned by [OptimisticDistributor](https://github.com/UMAprotocol/protocol/blob/master/packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol) as individual sponsors funding rewards might otherwise try to steal each others' deposits in MerkleDistributor by misconfiguring price identifier/ancilary data and making it difficult for honest verifiers to dispute malicious payout proposals (e.g. short OO liveness with high bonds).

Currently OptimisticDistributor does not implement calling `withdrawRewards` and `deleteWindow` methods on its controlled MerkleDistributor, but would be something to consider in future releases (e.g. contract controlled clawback mechanism by sponsors)

mrice32 (approved):
LGTM!

PR #3935, opened 2022-05-25
fix(optimistic-distributor): [M03] do not use priceDisputed callback

Reinis-FRP:
**Motivation**

Audit identified the following issue:
_Gas grief risk: The `priceDisputed` callback might not be guaranteed to execute which could permanently lock-in funds._

**Summary**

This PR fixes identified issue by not relying on the callbacks by the Optimistic Oracle.

**Details**

Previously Enum `distributionProposed` controlled both the acceptance of distribution payout proposals and their execution, and Optimistic Oracle callback triggered change in its status. Now without callback the contract has to actively ask Optimistic Oracle if previous proposal had been disputed on each check, hence it is more efficient also to split its functionality in:
- boolean `distributionExecuted` that blocks repeated execution,
- `blockingProposalTimestamp`: only allow new proposals if it is 0 (there are no prior proposals) or using this previous proposal timestamp to call `getRequest` on Optimistic Oracle yields non-zero disputer address.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
```diff
@@ -181,13 +174,14 @@ contract OptimisticDistributor is Lockable, MultiCaller, Testable {
...
             Reward({
-                distributionProposed: DistributionProposed.None,
+                distributionExecuted: false,
                 sponsor: msg.sender,
                 rewardToken: rewardToken,
                 maximumRewardAmount: maximumRewardAmount,
                 earliestProposalTimestamp: earliestProposalTimestamp,
                 optimisticOracleProposerBond: optimisticOracleProposerBond,
                 optimisticOracleLivenessTime: optimisticOracleLivenessTime,
+                blockingProposalTimestamp: 0,
```
chrismaree, line 184: I dont see where this variable is ever set to anything other than 0. you do `if (reward.blockingProposalTimestamp == 0) return true;` but where do you set it to a non-0 value?

chrismaree: nvm it's https://github.com/UMAprotocol/protocol/blob/99cdc5448748ef8a13f99a6f98d5f698e5b48751/packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol#L248

chrismaree, line 184: can we think about better names for this? what it really is is the dispute time, as defined on `        rewards[rewardIndex].blockingProposalTimestamp = timestamp;`. would `disputeTimestamp` be better? why is it a "blockingProposal", if its done via a dispute?

chrismaree (approved):
this makes sense and is actually cleaner than what we had before. this needs to be done in conjunction with the other OO Modification PR that changes how we deal with callbacks from within the OO.

I had one comment on naming but otherwise I think this is good.

packages/core/contracts/financial-templates/optimistic-distributor/OptimisticDistributor.sol
Reinis-FRP: This is not dispute timestamp, but the time when previous request/proposal was made through OD (or 0 if this is the first proposal). `_noBlockingProposal` then checks it and if non-0 asks OO `getRequest` with this timestamp. From OO perspective it is previous request timestamp, but from OD perspective it is previous proposal that matters (it just happens to be at the same timestamp when prior OO request was made). So maybe better alternative to use `previousProposalTimestamp` here?

PR #3937, opened 2022-05-25
fix: improvements to the uma transfer script

pemulis:

**Motivation**

- Assertion in verification was breaking if the recipient already has UMA at that address, since it was checking that the balance equaled the amount transferred. Removed this check for now in favor of manual verification.
- Added logs for manual verification of before-and-after balances for the recipient address and the Governor.

**Summary**

See motivation.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/scripts/src/admin-proposals/transferTokensFromGovernor.js
+++ b/packages/scripts/src/admin-proposals/transferTokensFromGovernor.js
@@ -41,8 +41,16 @@ async function run() {
   const gasEstimator = await setupGasEstimator();
 
   if (!verify) {
-    const originalBalance = await uma.methods.balanceOf(recipient).call();
-    const originalGovernorBalance = await uma.methods.balanceOf(governorAddress).call();
+    let originalBalance, originalGovernorBalance;
+    try {
+      [originalBalance, originalGovernorBalance] = await Promise.all([
+        uma.methods.balanceOf(recipient).call(),
+        uma.methods.balanceOf(governorAddress).call(),
+      ]);
+    } catch (error) {
+      console.log("error fetching balances:", error);
+    }
+
     console.log("Original balance of recipient:", originalBalance);
     console.log("Original balance of Governor:", originalGovernorBalance);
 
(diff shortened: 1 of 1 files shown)


packages/scripts/src/admin-proposals/transferTokensFromGovernor.js
```diff
@@ -35,11 +34,17 @@ async function run() {
...
   const web3 = web3Providers[1];
   const tokenAddress = await _getContractAddressByName("VotingToken", 1);
+  const governorAddress = await _getContractAddressByName("Governor", 1);
   console.log("tokenAddress:", tokenAddress);
   const uma = new web3.eth.Contract(ExpandedERC20.abi, tokenAddress);
 
   const gasEstimator = await setupGasEstimator();
 
+  var originalBalance = await uma.methods.balanceOf(recipient).call();
+  const originalGovernorBalance = await uma.methods.balanceOf(governorAddress).call();
```
nicholaspai, line 44: nit: could promise.all these to be faster :D

nicholaspai: also don't use `var`

nicholaspai (approved):
Some nit suggestions

packages/scripts/src/admin-proposals/transferTokensFromGovernor.js
```diff
@@ -35,11 +34,17 @@ async function run() {
   const web3Providers = { 1: getWeb3ByChainId(1) }; // netID => Web3
   const web3 = web3Providers[1];
   const tokenAddress = await _getContractAddressByName("VotingToken", 1);
+  const governorAddress = await _getContractAddressByName("Governor", 1);
   console.log("tokenAddress:", tokenAddress);
   const uma = new web3.eth.Contract(ExpandedERC20.abi, tokenAddress);
 
   const gasEstimator = await setupGasEstimator();
 
+  var originalBalance = await uma.methods.balanceOf(recipient).call();
```
smb2796, line 43: Nit: Might be better to have this within the if block, since this is misleading to be titled "original balances" when running verification step

packages/scripts/src/admin-proposals/transferTokensFromGovernor.js
pemulis: Can you give an example of how this would be re-written to use Promise.all()?

pemulis: Nevermind, added some code using Promise.all(), lemme know what you think!

PR #3948, opened 2022-05-27
fix: fix BlockFinder concurrency issue

mrice32:
**Motivation**

We're seeing bots break because of "undefined" blocks.

**Summary**

After running the debugger to get the state of the `this.blocks` array, I found that the this.blocks array was not sorted. This should be impossible since the elements are always inserted sorted and never removed.

After looking through the insertion points, I found this one, which computes the index, makes an async call, and then uses that index to insert the queried element. When this class is used concurrently, the state of this.blocks could have changed between the initiation of the async call and when it returns.

To avoid this, we need to recompute the index after the async call.

Note: it's hard to definitively say when this is fixed because the async scheduling is inherently chaotic. However, I will run local tests many times to attempt to tease out whether the error is still there.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

N/A

### DIFF
--- a/packages/sdk/src/blockFinder.ts
+++ b/packages/sdk/src/blockFinder.ts
@@ -58,9 +58,12 @@ export default class BlockFinder<T extends { number: number; timestamp: number |
 
   // Grabs the block for a particular number and caches it.
   private async getBlock(number: number) {
-    const index = sortedIndexBy(this.blocks, { number } as T, "number");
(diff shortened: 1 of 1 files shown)


daywiss (approved):
Amazing find if this works!

chrismaree (commented):
Excellent find!

PR #3953, opened 2022-06-01
add lsp demo

mrice32:


chrismaree (commented):
where are these scripts from?

PR #3956, opened 2022-06-02
chore(sdk): refactor oracle client to accept different oracle types

daywiss:

**Motivation**

We need to support the skinny OO, we should be able to instantiate the client with an arbitrary oracle client.

**Summary**

This refactors the code without changing functionaility, mainly to be able to make certain interfaces generic based on an different oracle client. Theoretically no functionality has been changed.

**Details**

This was mainly an excercise in getting typescript to work with somethhing like a plugin system, which was quite tricky and still is not perfect.  The next PR will add a skinny oo client.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested


packages/sdk/src/across/test/rateModel.e2e.ts
```diff
@@ -47,7 +47,11 @@ describe("rateModel", function () {
         );
         assert.fail("Should throw error");
       } catch (err) {
-        assert.equal(err.message.includes("method called before updating"), true);
+        if (err instanceof Error) {
```
daywiss, line 50: there was some typescript warnings here, this fixes it

packages/sdk/src/oracle/common/services/statemachine.ts
```diff
@@ -0,0 +1,266 @@
+import { Context, Memory, ContextType } from "oracle/common/types/statemachine";
```
daywiss, line 1: this file was refactored but functinality is not really changed. Only major change is that state machine handlers can be registered at run time. This is to allow different state handlers based on different oracle interfaces.

packages/sdk/src/oracle/common/services/statemachines/clearUser.ts
```diff
@@ -9,7 +9,7 @@ export function initMemory(): Memory {
   return undefined;
 }
 
-export function Handlers(store: Store): GenericHandlers<Params, Memory> {
+export function Handlers<S, O, E>(store: interfaces.Store<S, O, E>): GenericHandlers<Params, Memory> {
```
daywiss, line 12: all these handlers now have to consume a generic store interface

packages/sdk/src/oracle/common/store/store.ts
```diff
@@ -0,0 +1,49 @@
+import { RawStore } from "./rawStore";
```
daywiss, line 1: this file wasnt modified other than to implement the new interfaces

packages/sdk/src/oracle/common/types/interfaces.ts
```diff
@@ -0,0 +1,177 @@
+import { TransactionConfirmer, requestId } from "../utils";
```
daywiss, line 1: this file creates interfaces that need to be common across all oracle clients

packages/sdk/src/oracle/optimistic/client.ts
```diff
@@ -1,28 +1,31 @@
 import assert from "assert";
 import { ethers } from "ethers";
-import Store, { Emit } from "./store";
+
```
daywiss, line 3: this is the original client, it should function the same, its just been changed to use the new interfaces and plugin the necessary handlers at run time

packages/sdk/src/oracle/optimistic/types/state.ts
```diff
@@ -0,0 +1,157 @@
+import { JsonRpcSigner, BigNumber, Web3Provider, FallbackProvider } from "oracle/common/types/ethers";
```
daywiss, line 1: this file is almost identical to the previous state file specific to the current optimistic oracle contract.  some of this needs to be refactored to remove duplicate information shared between the clients.

packages/sdk/src/oracle/optimistic/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { DefaultConfig, getAddress, getMulticall2Address } from "oracle/common/utils";
```
daywiss, line 1: each optimsitic oracle will need its own address lookup function

packages/sdk/src/oracle/common/types/interfaces.ts
```diff
@@ -0,0 +1,177 @@
+import { TransactionConfirmer } from "../utils";
```
daywiss, line 1: this file defines interfaces that should be the same across clients using different oracles

packages/sdk/src/oracle/common/services/statemachine.ts
```diff
@@ -0,0 +1,266 @@
+import { Context, Memory, ContextType } from "../types/statemachine";
+import { ContextManager, shouldStep } from "../services/statemachines/utils";
```
daywiss, line 2: this file was copied from previous state machine file and modified to "register" different handler types. functionailty remains the same

packages/sdk/src/oracle/optimistic/types/state.ts
```diff
@@ -0,0 +1,158 @@
...
+import type { erc20, optimisticOracle, sortedRequests } from "../../common/services";
+import { Context, Memory } from "../../common/types/statemachine";
+
+// create partial picker: https://stackoverflow.com/questions/43159887/make-a-single-property-optional-in-typescript
+type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
+type PartialBy<T, K extends keyof T> = Omit<T, K> & Partial<Pick<T, K>>;
+
+export { Context, Memory };
+
+export type ChainServices = {
```
daywiss, line 22: this file is not new, its just copied almost exactly from previous state file. lots of duplicate data between this and common state file which i'll try to fix when i add the skinny OO client

packages/sdk/src/oracle/optimistic/utils.ts
```diff
@@ -0,0 +1,29 @@
+import { DefaultConfig, getAddress, getMulticall2Address } from "../common/utils";
+// this had to be copied in because interfaces in contracts-frontend and contracts-node are different
+// The frontend cant use contracts-node because async calls are required for addresses, when testing in node
+// we arent able to import contracts-frontend.
+export function getOptimisticOracleAddress(chainId: number): string {
```
daywiss, line 5: this isnt new, just moved out into its own file

packages/sdk/src/utils.ts
```diff
@@ -13,37 +13,47 @@ export function exists<T>(value: T | null | undefined): value is NonNullable<T>
 
 // useful for maintaining balances from events
 export type Balances = { [key: string]: string };
-export function Balances(balances: Balances = {}) {
-  function create(id: string, amount = "0") {
+export interface BalanceUtil {
```
daywiss, line 16: fixing some warnings

packages/sdk/src/oracle/common/store/store.ts
mrice32: Maybe this is obvious, but why do we have to copy these files? I would've though they could've been made more generic or functionality could have been added.

packages/sdk/src/oracle/common/types/interfaces.ts
```diff
@@ -0,0 +1,177 @@
...
+  Write: NewWrite<State, Events>;
+  Has: NewHas<State>;
+  Update: NewUpdate<Store<State, Oracle, Events>>;
+}
+export interface RawStore<S> {
+  write: (cb: RawWriteCallback<S>) => void;
+  writeAsync: (cb: RawWriteCallback<S>) => Promise<void>;
+  read: () => S;
+}
+export interface Store<S, O, E> {
```
mrice32, line 37: nit: as a reader, single-letter generics can be confusing, as it's unclear what these types are supposed to represent.

mrice32, line 37: nit: as a reader, single-letter generics can be confusing, as it's unclear what these types are supposed to represent.

packages/sdk/src/oracle/optimistic/store/write.ts
```diff
@@ -87,7 +87,7 @@ export class OptimisticOracle {
     }
   }
 }
-export class Chain {
+export class Chain implements types.interfaces.Chain<state.OptimisticOracleEvent> {
```
mrice32, line 90: I want to note a design choice here.

The way you're approaching this seems to be: keeping all the classes backwards compatible, but having them inherit from newly-generic interfaces. When we add the Skinny OO client, it will be the same, but it will need its own set of clients that inherit from these generic interfaces, but pass in different generic arguments to those interfaces. The code will mostly be the same outside of a few instances where the logic interacts with the parts of the types that are different.

The alternative design is to make the client itself generic and branch within the client where necessary. This would be less overall code, but it probably wouldn't be backwards compatible (since the generic types probably wouldn't be inferred and you might require additional user inputs), and it might require more complex logic _within_ the client to deal with the two possible contract types.

Do you agree with this analysis?

mrice32, line 90: I want to note a design choice here.

The way you're approaching this seems to be: keeping all the classes backwards compatible, but having them inherit from newly-generic interfaces. When we add the Skinny OO client, it will be the same, but it will need its own set of clients that inherit from these generic interfaces, but pass in different generic arguments to those interfaces. The code will mostly be the same outside of a few instances where the logic interacts with the parts of the types that are different.

The alternative design is to make the client itself generic and branch within the client where necessary. This would be less overall code, but it probably wouldn't be backwards compatible (since the generic types probably wouldn't be inferred and you might require additional user inputs), and it might require more complex logic _within_ the client to deal with the two possible contract types.

Do you agree with this analysis?

mrice32 (approved):
Looks great!!

amateima (approved):
💯 
Took me a while to go through all of it, but looks good to me. I don't have major comments about it

daywiss:
closing this in favor of #3989

PR #3969, opened 2022-06-11
fix: Across V1 event clients need to use paginated event search query now that DepositRelayed has over 10,000 events

nicholaspai:
**Motivation**

Some nodes like Infura cannot return more than 10,000 events, and the default search for these clients is set to look up events for ALL of history

**Summary**

Sets a default page size for event queries. Note that `InsuredBridgePriceFeed.getHistoricalPrice` is not used in `insured-bridge-relayer` and its only used in the validation script `getHistoricalPrice`.

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
+++ b/packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
@@ -103,7 +103,7 @@ export class InsuredBridgeL1Client {
     readonly rateModelStoreAddress: string | null,
     readonly startingBlockNumber = 0,
     readonly endingBlockNumber: number | null = null,
-    readonly blocksPerEventSearch: number | null = 20000
+    readonly blocksPerEventSearch: number | null = null
   ) {
     // Cast the following contracts to web3-specific type
     this.bridgeAdmin = (new l1Web3.eth.Contract(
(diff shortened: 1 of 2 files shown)


packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
```diff
@@ -103,7 +103,7 @@ export class InsuredBridgeL1Client {
     readonly rateModelStoreAddress: string | null,
     readonly startingBlockNumber = 0,
     readonly endingBlockNumber: number | null = null,
-    readonly blocksPerEventSearch: number | null = null
+    readonly blocksPerEventSearch: number | null = 20000
```
nicholaspai, line 106: This shouldn't break V1 bots in prod but worth thinking about

packages/financial-templates-lib/src/price-feed/InsuredBridgePriceFeed.ts
```diff
@@ -69,10 +69,13 @@ export class InsuredBridgePriceFeed extends PriceFeedInterface {
       depositHash: string;
     }
     let matchedRelay: MatchedRelay | undefined;
+    const latestL1Block = await this.l1Client.l1Web3.eth.getBlockNumber()
```
nicholaspai, line 72: This pricefeed is only used in `getHistorifcalPrice.js` script I believe but pls double check me

packages/financial-templates-lib/src/clients/InsuredBridgeL1Client.ts
nicholaspai: I ended up just removing and it still worked

mrice32 (approved):
LGTM!

PR #3970, opened 2022-06-13
fix(protocol): re-generate lockfile to prevent node-gyp errors on `yarn`

ryanwolhuter:
**Motivation**

When running `yarn` in the protocol root directory, I encountered several node-gyp build errors on my Apple Silicon device. These can often be caused by a multitude of factors, but in my case it seems that a stale lockfile was to blame. To prevent this issue happening to others in their setups, we can commit this newly-generated lockfile to be used from this point forward.

**Summary**

1. Deleted the existing lockfile found in the repo.
2. Ran `yarn` with no lockfile, generating a fresh new one.

**Testing**

- [ ]  New unit tests created
- [x]  Untested

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11,6 +11,14 @@
     "@openzeppelin/contracts" "4.1.0"
     "@uma/core" "^2.18.0"
 
+"@ampproject/remapping@^2.1.0":
+  version "2.2.0"
+  resolved "https://registry.yarnpkg.com/@ampproject/remapping/-/remapping-2.2.0.tgz#56c133824780de3174aed5ab6834f3026790154d"
+  integrity sha512-qRmjj8nj9qmLTQXXmaR1cck3UXSRMPrbsLJAasZpF+t3riI71BXed5ebIOYwQntykeZuhjsdweEc9BxH5Jc26w==
+  dependencies:
+    "@jridgewell/gen-mapping" "^0.1.0"
+    "@jridgewell/trace-mapping" "^0.3.9"
+
 "@antora/asciidoc-loader@2.3.4":
   version "2.3.4"
   resolved "https://registry.yarnpkg.com/@antora/asciidoc-loader/-/asciidoc-loader-2.3.4.tgz#b5c4b6447981e191cea8438226a8e76904fc1e7d"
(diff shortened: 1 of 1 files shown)


mrice32 (commented):
@ryanwolhuter it looks like this is breaking the build due to some changes in the typescript types/interfaces. Let me know if you'd like any help in resolving this!

chrismaree:
> @ryanwolhuter it looks like this is breaking the build due to some changes in the typescript types/interfaces. Let me know if you'd like any help in resolving this!

ye this is intresting. @ryanwolhuter added this specifically due to his build not working on his local machine but now it does not work in the linux env in CI. The error in CI looks similar to what @md0x was having on his local machine.

ryanwolhuter:
> @ryanwolhuter it looks like this is breaking the build due to some changes in the typescript types/interfaces. Let me know if you'd like any help in resolving this!

This is quite strange indeed. I would definitely like some help!

I tried a couple of things just to see what would happen:

* Pinned the version of typescript to exact; I figured maybe a later version of ts throws an error that the current did not.
* Pinned the version of web3 to exact; I figured maybe the types defined in there had changed.

Neither made any difference.

The error that makes the build fail is:

`@uma/common: src/MetaMaskTruffleProvider.ts(22,33): error TS2345: Argument of type 'import("/Users/ry/uma/protocol/node_modules/web3-core/node_modules/web3-core-helpers/types/index").JsonRpcPayload' is not assignable to parameter of type 'import("node-metamask").JsonRpcPayload'.
@uma/common:   Types of property 'params' are incompatible.
@uma/common:     Type 'any[] | undefined' is not assignable to type 'any[]'.
@uma/common:       Type 'undefined' is not assignable to type 'any[]'.
@uma/common: src/MetaMaskTruffleProvider.ts(29,28): error TS2345: Argument of type 'import("/Users/ry/uma/protocol/node_modules/web3-core/node_modules/web3-core-helpers/types/index").JsonRpcPayload' is not assignable to parameter of type 'import("node-metamask").JsonRpcPayload'.`

So the issue is arising from a package that UMA controls: `"node-metamask": "github:UMAprotocol/node-metamask",`

Seems like a simple fix here. It probably just needs a non-null assertion in there to deal with the potential `undefined`. 

I'm not sure how I should proceed here though, because I'm messing with things I don't yet fully understand in these repositories.

PR #3971, opened 2022-06-13
Feat: DVM2.0 Initial staking/slashing implementation

chrismaree:
**Motivation**

This is the initial PR that adds staking and slashing logic to the Voting.sol contracts. The design goals can be found [here](https://hackmd.io/@k05rNR4FS26flqCmZvAUXA/dvm-v2).

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/core/contracts/oracle/implementation/dvm2.0/Staker.sol
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    function requestUnstake(uint256 amount) public {
+        _updateTrackers(msg.sender);
+        // Staker signals that they want to unstake. After signalling, their total voting balance is decreased by the
+        // signaled amount. This amount is not vulnerable to being slashed but also does not accumulate rewards.
+
+        require(stakingBalances[msg.sender].cumulativeStaked >= amount, "Bad request amount");
+
+        stakingBalances[msg.sender].requestUnstake = amount;
```
mrice32, line 85: This method allows a user to restake as long as they haven't withdrawn, right? i.e. I call with amount = 10. Later, I call this method with amount = 0. My 10 is now staked again.

Without snapshotting the requestUnstake value at some point in the voting process, this would (I think) open a loophole for someone to unstake all their tokens during the vote and have the option to restake before rewards are assessed once they have more information about the result of the vote.

mrice32 (commented):
This is super cool!

Left a few comments.

packages/core/contracts/oracle/implementation/dvm2.0/Staker.sol
```diff
@@ -0,0 +1,228 @@
...
+
+    function _updateCumulativeSlashingTrackers() internal {
+        // Note the method below can hit a gas limit of there are a LOT of requests from the last time this was run.
+        // A future version of this should bound how many requests to look at per call to avoid gas limit issues.
+        for (uint256 i = lastRequestIndexConsidered; i < priceRequestIds.length; i++) {
+            PriceRequest storage priceRequest = priceRequests[priceRequestIds[i].requestId];
+            VoteInstance storage voteInstance = priceRequest.voteInstances[priceRequest.lastVotingRound];
+            uint256 roundId = rounds[priceRequestIds[i].roundId].snapshotId;
+
+            uint256 totalStaked = votingToken.balanceOfAt(address(this), roundId);
```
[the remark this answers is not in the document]
chrismaree: ye, I've re-worked this in a follow on commit (not yet pushed). specifically, we now need the staking contract to be snapshotable wherein we snapshot your staked amount and unrealized slashing amount on each reveal and consider this in your voting power, rather than your token balance.

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,244 @@
+// SPDX-License-Identifier: AGPL-3.0-only
```
chrismaree, line 1: this is the primary new contract implemented in this PR. it is meant to be inherited from VotingV2.sol and provides the ability to Stake and snapshot staked amounts in the voting contract. it is based off in part the [accelerating distributor](https://github.com/across-protocol/across-token/blob/master/contracts/AcceleratingDistributor.sol) from across v2 and the [ERC20Snapshot](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/token/ERC20/extensions/ERC20Snapshot.sol) from openzepplin.

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -0,0 +1,1102 @@
+// TODO: this whole /oracle/implementation directory should be restructured to seperate the DVM and the OO.
```
chrismaree, line 1: this contract is mostly a copy-paste of the main `Voting.sol` contract except for the addition of some untested slashing functions and some todos on things that will be changed in the DVM-v2 implementation. for now it can be mostly ignored in review

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -0,0 +1,1102 @@
...
+        return 1600000000000000;
+    }
+
+    function calcNoVoteSlashPerToken(
+        uint256 totalStaked,
+        uint256 totalVotes,
+        uint256 totalCorrectVotes
+    ) public pure returns (uint256) {
+        return 1600000000000000;
+    }
```
chrismaree, line 346: this is the MVP of the slashing logic. a subsequent PR will update this and add unit tests

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -0,0 +1,1102 @@
...
+
+    // Maps price request IDs to the PriceRequest struct.
+    mapping(bytes32 => PriceRequest) internal priceRequests;
+
+    struct Request {
+        bytes32 requestId;
+        uint256 roundId;
+    }
+
+    Request[] internal priceRequestIds;
```
chrismaree, line 117: additional data structure needed in computing slashing.

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -0,0 +1,1102 @@
...
+
+    uint256 public lastRequestIndexConsidered;
+
+    struct SlashingTracker {
+        uint256 wrongVoteSlashPerToken;
+        uint256 noVoteSlashPerToken;
+        uint256 totalSlashed;
+    }
+
+    SlashingTracker[] public requestSlashingTrackers;
```
chrismaree, line 171: track slashing amounts.

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,242 @@
...
+
+    // Note there is no way to cancel your unstake; you must wait until after unstakeTime and re-stake.
+
+    // If: a staker requested an unstake and time > unstakeTime then send funds to staker. Note that this method assumes
+    // that the `updateTrackers()
+    function executeUnstake() public {
+        _updateTrackers(msg.sender);
+        updateSnapshot(msg.sender);
+        VoterStake storage voterStake = voterStakes[msg.sender];
+        require(voterStake.unstakeTime != 0 && getCurrentTime() >= voterStake.unstakeTime, "Unstake time not passed");
```
mrice32, line 101: Similar comment to last time: https://github.com/UMAprotocol/protocol/pull/3971#discussion_r898374036. Is there any way to avoid this unstaking UX -- it seems like this will hurt the overall voting UX to have a delay like this.

In general, it seems sufficient to only have voters get rewards on the balance they had during the previous vote.

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,242 @@
...
+    // Owner methods
+    function setEmissionRate(uint256 _emissionRate) public onlyOwner {
+        emissionRate = _emissionRate;
+    }
+
+    function setUnstakeCoolDown(uint256 _unstakeCoolDown) public onlyOwner {
+        unstakeCoolDown = _unstakeCoolDown;
+    }
+
+    // Snapshot methods
```
mrice32, line 168: Not exactly sure how we're planning to use this snapshot functionality, but it might be overkill for what we need. For instance, I don't think we need to track voters' balances all the way back in history, right? We only need to know their balance _in the most recent vote_, no?

Ideally, though, we might be able to get away _without snapshotting at all_ (this is half baked). Since all balance updates happen via the contract, we might be able to read the current staking balance for the purposes of tallying votes or accumulating rewards before taking any balance-modifying action on the user.

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    function requestUnstake(uint256 amount) public {
+        _updateTrackers(msg.sender);
+        // Staker signals that they want to unstake. After signalling, their total voting balance is decreased by the
+        // signaled amount. This amount is not vulnerable to being slashed but also does not accumulate rewards.
+
+        require(voterStakes[msg.sender].cumulativeStaked >= amount, "Bad request amount");
+
+        voterStakes[msg.sender].requestUnstake = amount;
```
mrice32, line 89: Since the user could technically reduce this value, is there any concern about users unstaking and then restaking those tokens before executing the withdrawal?

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
chrismaree: this is not correct. you _do_ need to know the snapshots at all historic block rounds. the main reason is that when executing slashing you need to know how much the voter had staked at the round we are considering in the slashing.

you can see the two places in the slashing logic where this is required in the follow on PR.:

- https://github.com/UMAprotocol/protocol/pull/3981/files#r903470022
- https://github.com/UMAprotocol/protocol/pull/3981/files#r903469951

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
chrismaree: you cant unstake until your liveness timer has passed. you must wait until the `unstakeCoolDown` time has passed from when you staked. if they change this value (i.e re-request to unstake) then this timer is reset.

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,242 @@
...
+        votingToken.transferFrom(msg.sender, address(this), amount);
+    }
+
+    function updateSnapshot(address voterAddress) public {
+        _updateAccountSnapshot(voterAddress);
+        _updateTotalStakedSnapshot();
+    }
+
+    function requestUnstake(uint256 amount) public {
+        _updateTrackers(msg.sender);
```
md0x, line 83: A little observation: the tests still pass if you remove this line. Is it possible to omit this `_updateTrackers`?

md0x:
This implementation looks quite solid @chrismaree ! I just left a little observation.

mrice32 (approved):

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
```diff
@@ -0,0 +1,242 @@
...
+    }
+
+    function rewardPerToken() public view returns (uint256) {
+        if (cumulativeStaked == 0) return rewardPerTokenStored;
+        return rewardPerTokenStored + ((getCurrentTime() - lastUpdateTime) * emissionRate * 1e18) / cumulativeStaked;
+    }
+
+    // Owner methods
+    function setEmissionRate(uint256 _emissionRate) public onlyOwner {
+        emissionRate = _emissionRate;
```
md0x, line 161: I believe that currently, if the emissionRate is changed and there are unclaimed rewards they will be retroactively updated.  Is this a problem?

I have been exporting this idea in this branch, to better understand the mechanism. (Maybe it is not relevant at all) : https://github.com/UMAprotocol/protocol/pull/3987

chrismaree: this is an excellent catch!

packages/core/contracts/oracle/implementation/StakerSnapshot.sol
chrismaree: point of this is it gets override in the `VotingV2.sol` contract. will make sence a bit later as we go deeper into the implementation :)

PR #3975, opened 2022-06-15
feat: pagerduty v2 logger

daywiss:

**Motivation**

This adds the new v2 pagerduty event logger, which can let us use hte native pagerduty incident router.   See [api docs here](https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTgx-send-an-alert-event). 

**Summary**

Simply adds a v2 pagerduty transport which uses a client library under the hood.  This was tested with below script. This tries to maintain the structure from v1 calls

**Testing**

- [ ]  New unit tests created
- [x]  Untested

Tested this using a small script 
```js
import { createNewLogger } from "./src";

const transportsConfig = {
  pagerDutyV2Config:{   
     integrationKey:"xxx"
  }
}

async function run(): Promise<void> {    
  const logger = createNewLogger([],transportsConfig,'test-bot');    
  logger.error({ at: "tester", message: "markdown", mrkdwn: "*headding*\nsome test" });
  logger.error({ at: "tester", message: "complex type", someComplexType:{a:'b'}}); 
}

run().then(() => {
   console.log('done')
}).catch(error => {    
 console.error("Process exited with", error);  
})  
```

### DIFF
--- a/packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
+++ b/packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
@@ -1,0 +1,49 @@
+// This transport enables winston logging to send messages to pager duty v2 api.
+import Transport from "winston-transport";
+import { event } from "@pagerduty/pdjs";
+
+type TransportOptions = ConstructorParameters<typeof Transport>[0];
+export type Severity = "critical" | "error" | "warning" | "info";
+
+interface Config {
+  integrationKey: string;
+  customServices?: { [key: string]: string };
+}
+
+export class PagerDutyTransportV2 extends Transport {
+  private readonly integrationKey: string;
+  private readonly customServices: { [key: string]: string };
+  constructor(winstonOpts: TransportOptions, { integrationKey, customServices = {} }: Config) {
+    super(winstonOpts);
+    this.integrationKey = integrationKey;
+    this.customServices = customServices;
+  }
+  // pd v2 severity only supports critical, error, warning or info.
+  public static convertLevelToSeverity(level?: string): Severity {
+    if (!level) return "error";
+    if (level === "warn") return "warning";
+    if (level === "info" || level === "critical") return level;
+    return "error";
+  }
+  // Note: info must be any because that's what the base class uses.
+  async log(info: any, callback: () => void): Promise<void> {
+    try {
+      await event({
+        data: {
+          routing_key: this.integrationKey,
+          event_action: "trigger",
+          payload: {
+            summary: `${info.level}: ${info.at} ⭢ ${info.message}`,
+            severity: PagerDutyTransportV2.convertLevelToSeverity(this.level),
+            source: info["bot-identifier"] ? info["bot-identifier"] : undefined,
+            custom_details: info,
+          },
+        },
+      });
+    } catch (error) {
+      console.error("PagerDuty v2 error", error);
+    }
+
+    callback();
+  }
+}
--- a/packages/financial-templates-lib/src/logger/PagerDutyV2Transport.ts
+++ b/packages/financial-templates-lib/src/logger/PagerDutyV2Transport.ts
@@ -10,7 +10,7 @@ interface Config {
   customServices?: { [key: string]: string };
 }
 
-export class PagerDutyTransportV2 extends Transport {
+export class PagerDutyV2Transport extends Transport {
   private readonly integrationKey: string;
   private readonly customServices: { [key: string]: string };
   constructor(winstonOpts: TransportOptions, { integrationKey, customServices = {} }: Config) {
@@ -28,14 +28,17 @@ export class PagerDutyTransportV2 extends Transport {
   // Note: info must be any because that's what the base class uses.
   async log(info: any, callback: () => void): Promise<void> {
     try {
+      // we route to different pd services using the integration key (routing_key), or multiple services with the custom services object
+      const routing_key = this.customServices[info.notificationPath] ?? this.integrationKey;
       await event({
         data: {
-          routing_key: this.integrationKey,
+          routing_key,
           event_action: "trigger",
           payload: {
             summary: `${info.level}: ${info.at} ⭢ ${info.message}`,
-            severity: PagerDutyTransportV2.convertLevelToSeverity(this.level),
+            severity: PagerDutyV2Transport.convertLevelToSeverity(this.level),
             source: info["bot-identifier"] ? info["bot-identifier"] : undefined,
+            // we can put any structured data in here as long as it is can be repped as json
             custom_details: info,
           },
         },
--- a/packages/financial-templates-lib/src/logger/Transports.ts
+++ b/packages/financial-templates-lib/src/logger/Transports.ts
@@ -7,6 +7,7 @@ import { createConsoleTransport } from "./ConsoleTransport";
 import { createJsonTransport } from "./JsonTransport";
 import { createSlackTransport } from "./SlackTransport";
 import { PagerDutyTransport } from "./PagerDutyTransport";
+import { PagerDutyV2Transport } from "./PagerDutyV2Transport";
 import { DiscordTransport } from "./DiscordTransport";
 import type Transport from "winston-transport";
 import dotenv from "dotenv";
@@ -18,6 +19,7 @@ const argv = minimist(process.argv.slice(), {});
 type SlackConfig = Parameters<typeof createSlackTransport>[0];
 type DiscordConfig = ConstructorParameters<typeof DiscordTransport>[1];
 type PagerDutyConfig = ConstructorParameters<typeof PagerDutyTransport>[1];
+type PagerDutyV2Config = ConstructorParameters<typeof PagerDutyV2Transport>[1];
 
 interface TransportsConfig {
   environment?: string;
@@ -26,6 +28,7 @@ interface TransportsConfig {
   discordConfig?: DiscordConfig;
   pdApiToken?: string;
   pagerDutyConfig?: PagerDutyConfig;
+  pagerDutyV2Config?: PagerDutyV2Config;
 }
 
 export function createTransports(transportsConfig: TransportsConfig = {}): Transport[] {
@@ -74,6 +77,15 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
         new PagerDutyTransport(
           { level: "warn" },
           transportsConfig.pagerDutyConfig ?? JSON.parse(process.env.PAGER_DUTY_CONFIG || "null")
+        )
+      );
+    }
+    // If there is a Pagerduty V2 API key then add the pagerduty winston transport.
+    if (transportsConfig.pagerDutyV2Config || process.env.PAGER_DUTY_V2_CONFIG) {
+      transports.push(
+        new PagerDutyV2Transport(
+          { level: "warn" },
+          transportsConfig.pagerDutyV2Config ?? JSON.parse(process.env.PAGER_DUTY_V2_CONFIG || "null")
         )
       );
     }
--- a/packages/financial-templates-lib/package.json
+++ b/packages/financial-templates-lib/package.json
@@ -9,6 +9,7 @@
     "@uma/common": "^2.26.0",
     "@uma/contracts-node": "^0.3.14",
     "@uma/sdk": "^0.28.1",
+    "@pagerduty/pdjs": "^2.2.4",
     "@uniswap/sdk": "^2.0.5",
     "bluebird": "^3.7.2",
     "bn.js": "^4.11.9",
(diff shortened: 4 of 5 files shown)


packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
...
+  customServices?: { [key: string]: string };
+}
+
+export class PagerDutyTransportV2 extends Transport {
+  private readonly integrationKey: string;
+  private readonly customServices: { [key: string]: string };
+  constructor(winstonOpts: TransportOptions, { integrationKey, customServices = {} }: Config) {
+    super(winstonOpts);
+    this.integrationKey = integrationKey;
+    this.customServices = customServices;
```
mrice32, line 19: Is this unused? Does this mean this transport will not allow us to route to multiple services in the same bot?

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (level === "warn") return "warning";
+    if (level === "info" || level === "critical") return level;
+    return "error";
+  }
+  // Note: info must be any because that's what the base class uses.
+  async log(info: any, callback: () => void): Promise<void> {
+    try {
+      await event({
+        data: {
+          routing_key: this.integrationKey,
```
mrice32, line 33: The routing seems to be a bit different, can you explain how this works?

In the previous version, we had a pagerduty client that we would create _with_ the credentials. Then we would provide the name of the service for each event that we wanted to emit. In this version, it appears as if we're not using a client at all and just providing a single key for the route. Does this mean that customServices should be a mapping from a configured route name to an integration key that will send alerts to a particular PD service?

mrice32 (commented):
Looks great! Just a few questions!

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type TransportOptions = ConstructorParameters<typeof Transport>[0];
+export type Severity = "critical" | "error" | "warning" | "info";
+
+interface Config {
+  integrationKey: string;
+  customServices?: { [key: string]: string };
+}
+
+export class PagerDutyTransportV2 extends Transport {
```
chrismaree, line 13: would it not be better to replace the current pagerduty transport than have two?

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
+// This transport enables winston logging to send messages to pager duty v2 api.
```
chrismaree, line 1: one thing this is missing is the ability to switch on `notificationPath` to send to different `serviceIds` as done in the "v1" pagerduty transport.https://github.com/UMAprotocol/protocol/blob/9fabd6cd7d2f1d0508199f779f64957f950c5f42/packages/financial-templates-lib/src/logger/PagerDutyTransport.ts#L42 this lets us define a config like this:

```json
"PAGER_DUTY_CONFIG": {
        "pdApiToken": "x",
        "fromEmail": "x",
        "defaultServiceId": "x",
        "customServices": {
          "risk-management": "x",
          "optimistic-oracle": "x",
          "dev-x": "x",
          "outcome-expirations": "x",
          "polymarket-notifier": "x"
        }
      },
```

and then logs go to each "customService" if the `notificationPath` is set. else, it goes to the `defaultServiceId`

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
...
+import Transport from "winston-transport";
+import { event } from "@pagerduty/pdjs";
```
chrismaree, line 3: The previous implementation uses the depricated `node-pagerduty`: https://github.com/UMAprotocol/protocol/blob/9fabd6cd7d2f1d0508199f779f64957f950c5f42/packages/financial-templates-lib/src/logger/PagerDutyTransport.ts#L7

this old package should likely be removed.

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
chrismaree: ye I asked a similar question. if this is the case then we need some switching logic on `notificationPath`

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
```diff
@@ -0,0 +1,49 @@
...
+    try {
+      await event({
+        data: {
+          routing_key: this.integrationKey,
+          event_action: "trigger",
+          payload: {
+            summary: `${info.level}: ${info.at} ⭢ ${info.message}`,
+            severity: PagerDutyTransportV2.convertLevelToSeverity(this.level),
+            source: info["bot-identifier"] ? info["bot-identifier"] : undefined,
+            custom_details: info,
```
chrismaree, line 39: I think you need to do more post-processing on `info` based on it's type. if it is `mrkdwn ` then you likely want to extract just the `mrkdwn ` part. else, you should likely `mrkdwn ` the object (see [here](https://github.com/UMAprotocol/protocol/blob/9fabd6cd7d2f1d0508199f779f64957f950c5f42/packages/financial-templates-lib/src/logger/PagerDutyTransport.ts#L38).)

I would recommend you set up a sample escalation policy that only pages you and then test out your transport with sample code that looks something like this:
```ts
import { delay, Logger } from "@uma/financial-templates-lib";

export async function run(): Promise<void> {
  const logger = Logger;
  // try messages with mrkdwn, some with json objects, some with nested errors.
  logger.error({
    at: "tester",
    message: "hello world",
    mrkdwn:
      "*headding*\nsome test
  });
logger.info({
    at: "tester",
    message: "hello world",
    someComplexType:{a:b}
  });
}

if (require.main === module) {
  run()
    .then(async () => {
      await delay(5);
      process.exit(0);
    })
    .catch(async (error) => {
      console.error("Process exited with", error);
      await delay(5);
      process.exit(1);
    });
}
```

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: i dont really understand how this is used, but in the pd2 api, there is no field for `serviceId`.  In v1 we do: `const serviceId = this.customServices[info.notificationPath] ?? this.defaultServiceId;`.   In the v2 api theres maybe something similar with `payload.source=The unique location of the affected system, preferably a hostname or FQDN.` but i set this as the bot identifier.  

Take a look at the payload shape for v2 api as its quite different from v1: https://developer.pagerduty.com/docs/ZG9jOjExMDI5NTgx-send-an-alert-event

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: So from what im understanding, the v2 api is more of a logging system, not an alerting system. i think the point is to get all the data into their logs, and then make routing rules based on those which can then set up alerting.  The "routing key" in this case, theres only a single key for your pd account, and i dont see any option to create others. 

As for custom services, i dont know what that does really, but i figured if i passed though as much raw data as i could in the logs, i could create the rules in pd to do proper sorting and routing.

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: i dont think so, the apis are really different and it would force us to have to migrate everything at once, rather than maybe more controlled migration where we can pick and choose what we add, or add additional v2 calls while keeping the v1 calls in

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: im not ready to remove this, as i dont know how well the v2 api will work, theres still a lot of unknowns until i can get some real world data working

daywiss:
may not need this, will delay further work until its clear

daywiss:
closing this since it turns out we do not need to migrate to the new api to take advantage of some pd automation. we will need to  buy a new feature.

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: tested using a similar script. further parsing of info isnt really necessary, we can include all the data without extra parsing, heres what the markdown looks like:
image
image

compare this to our custom details with current alerts where the data comes in unstructured (probably because of the json stringification before sending to pd):
image

packages/financial-templates-lib/src/logger/PagerDutyTransportV2.ts
daywiss: i have a better understanding of wahts going on with this, and have added a way to route to different services, using a similar mechanism as the v1 notifier

mrice32:
@chrismaree can you take a pass on this?

packages/financial-templates-lib/src/logger/Transports.ts
```diff
@@ -77,6 +80,15 @@ export function createTransports(transportsConfig: TransportsConfig = {}): Trans
         )
       );
     }
+    // If there is a Pagerduty V2 API key then add the pagerduty winston transport.
+    if (transportsConfig.pagerDutyV2Config || process.env.PAGER_DUTY_V2_CONFIG) {
```
chrismaree, line 84: general question: if we have the v2 transport, is the intention to eventually move away all the v1 configs to the v2 config? I get that the structure is slightly different so we'll need to be sure to update all the bot configs accordingly but it seems somewhat redundant to support the old one long term. wydt?

packages/financial-templates-lib/src/logger/PagerDutyV2Transport.ts
```diff
@@ -0,0 +1,52 @@
...
+  public static convertLevelToSeverity(level?: string): Severity {
+    if (!level) return "error";
+    if (level === "warn") return "warning";
+    if (level === "info" || level === "critical") return level;
+    return "error";
+  }
+  // Note: info must be any because that's what the base class uses.
+  async log(info: any, callback: () => void): Promise<void> {
+    try {
+      // we route to different pd services using the integration key (routing_key), or multiple services with the custom services object
```
chrismaree, line 31: your comment here says "multiple services with the custom services object". just to be clear: the service key that you define here will do the multiple routing, right? as in you send it to _one_  `routing_key` and then pagerduty sends it to multiple escalation paths/servicesIds?

packages/financial-templates-lib/src/logger/PagerDutyV2Transport.ts
```diff
@@ -0,0 +1,52 @@
...
+      await event({
+        data: {
+          routing_key,
+          event_action: "trigger",
+          payload: {
+            summary: `${info.level}: ${info.at} ⭢ ${info.message}`,
+            severity: PagerDutyV2Transport.convertLevelToSeverity(this.level),
+            source: info["bot-identifier"] ? info["bot-identifier"] : undefined,
+            // we can put any structured data in here as long as it is can be repped as json
+            custom_details: info,
```
chrismaree, line 42: in the v1 bot we did some parsing on this to make it look different depending on if it has markdown in it or is a json. do you think it's worth doing the same here? https://github.com/UMAprotocol/protocol/blob/e4eb7213f1954afb2fc4d8d4e498562efedfcc1e/packages/financial-templates-lib/src/logger/PagerDutyTransport.ts#L38

chrismaree: nvm I see that it spreads this nicely for us in this screenshot you posted: https://github.com/UMAprotocol/protocol/pull/3975#discussion_r963953415

packages/financial-templates-lib/src/logger/Transports.ts
daywiss: i think it will be safe to remove v1 in the medium term, would just like to be confident that this v2 solution is working as needed before that happens

packages/financial-templates-lib/src/logger/PagerDutyV2Transport.ts
daywiss: hmm, the way i understand this ( and i could be wrong), this routing_key does double duty as your api key as well as your destination service.  As far as i know its only 1 routing key per service (not one to many), but you can essentially configure a single logger to emit to multiple services by setting the `customServices[info.notificationPath]` with each of their routing keys.

chrismaree (approved):

PR #3976, opened 2022-06-16
fix: require greater eth balance in arbitrum parent messenger

pemulis:

**Motivation**

The current check that the Arbitrum parent messenger has enough ETH only checks that it has enough ETH for a single transaction. This verifies it has enough ETH for multiple transactions, for instance, adding a collateral type and updating its final fee in the same proposal.

**Summary**

Multiplies the value required by the check.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX

### DIFF
--- a/packages/scripts/src/admin-proposals/utils.js
+++ b/packages/scripts/src/admin-proposals/utils.js
@@ -187,7 +187,7 @@ const fundArbitrumParentMessengerForOneTransaction = async (web3Provider, from, 
   console.log(
     `Arbitrum xchain messages require that the Arbitrum_ParentMessenger has at least a ${l1CallValue.toString()} ETH balance.`
   );
-  if (toBN(await web3Provider.eth.getBalance(arbitrumParentMessenger.options.address)) < toBN(l1CallValue)) {
+  if (toBN(await web3Provider.eth.getBalance(arbitrumParentMessenger.options.address)) < toBN(l1CallValue * 10)) {
     const sendEthTxn = await web3Provider.eth.sendTransaction({
       from: from,
       to: arbitrumParentMessenger.options.address,


packages/scripts/src/admin-proposals/utils.js
```diff
@@ -187,7 +187,7 @@ const fundArbitrumParentMessengerForOneTransaction = async (web3Provider, from,
   console.log(
     `Arbitrum xchain messages require that the Arbitrum_ParentMessenger has at least a ${l1CallValue.toString()} ETH balance.`
   );
-  if (toBN(await web3Provider.eth.getBalance(arbitrumParentMessenger.options.address)) < toBN(l1CallValue)) {
+  if (toBN(await web3Provider.eth.getBalance(arbitrumParentMessenger.options.address)) < toBN(l1CallValue * 10)) {
```
nicholaspai, line 190: You also need to send `l1CallValue` * 10 ETH right? 

Also, the more I think about this, perhaps we should just set to 5, or some other config/CLI arg we can pass into the script...

With a default value of 2?

nicholaspai (commented):
Had some thoughts to avoid us locking 1.5 ETH into this adapter for a while

packages/scripts/src/admin-proposals/utils.js
pemulis: If we did that, it would send 10X the amount required whenever it dropped to 9X the amount required, giving you 19X the amount required locked up. That might be a better UX experience for the proposer but would lock up more ETH. Not sure what trade-off is better there as a short-term fix.

And yeah, I think there's definitely cleaner solutions available when we do a broader refactoring of the scripts.

mrice32: I agree with @nicholaspai

I think we should send the difference between the threshold and the actual balance in the interest of correctness.

However, I think we should probably just bite the bullet and have this method take in the number of calls being made to arbitrum. This shouldn't be a major refactor, just a quick parameter change I would think.

Without doing either of these, I think modifying this check this way might make this script more difficult to use down the road rather than simpler.

mrice32 (commented):
Just one comment

chrismaree:
Hey @pemulis. are there any updates to this PR? what is the status on the changes requested?

PR #3980, opened 2022-06-20
feat: umip 162 update simulation script to include l2 relays

md0x:
Signed-off-by: Pablo Maldonado <pablo@umaproject.org>

**Motivation**

Simulate the proposal actions of the UMIP 162 https://github.com/UMAprotocol/UMIPs/pull/525

The verify script ensures that the actions were executed in l1 and that the relay events were emitted in l1 (it doesn't simulate the actions in l2-s).

**Summary**

Simulates and verifies the following actions:
* Propose and execute the registration of 00V2 as a verified contract in mainnet
* Propose and execute adding OOV2 to the finder
* For the layer 2 blockchains: 
    * Propose relay to the register OOV2 as a verified contract in l2
    *  Propose relay to add OOV2 to the finder in l2


packages/scripts/src/upgrade-tests/162/1_Propose.ts
```diff
@@ -1,85 +1,344 @@
...
+    console.log("- changeImplementationAddressData", setFinderData);
+  }
+
+  // store relayRecords
+  const relayRecordsJson = JSON.stringify(relayRecords, null, 2);
+  const relayRecordsFile = path.join(path.dirname(__filename), "response.json");
+
+  // delete file if exists
+  if (fs.existsSync(relayRecordsFile)) {
+    fs.unlinkSync(relayRecordsFile);
```
mrice32, line 290: I'm a little confused -- why do we need to save a file?

packages/scripts/src/upgrade-tests/162/1_Propose.ts
```diff
@@ -1,85 +1,344 @@
...
+): Promise<{
+  to: string;
+  value: BigNumberish;
+  data: BytesLike;
+}> => {
+  if (!tx.data) throw new Error("Transaction has no data");
+  const relayGovernanceData = await governorRootTunnel.populateTransaction.relayGovernance(targetAddress, tx.data);
+  console.log("- relayGovernanceData", relayGovernanceData);
+  const relay = await governorRootTunnel.populateTransaction.relayGovernance(targetAddress, tx.data);
+  const relayMessage = relay.data;
```
mrice32, line 81: This is much easier than using the ethers Interface classes. Didn't realize you could generate transaction data this way.

packages/scripts/src/upgrade-tests/162/1_Propose.ts
```diff
@@ -1,85 +1,344 @@
...
+  tx: PopulatedTransaction,
+  governorHub: GovernorHub,
+  chainId: BigNumberish
+): Promise<{
+  to: string;
+  value: BigNumberish;
+  data: BytesLike;
+}> => {
+  if (!tx.data) throw new Error("Transaction has no data");
+  const calls = [{ to: targetAddress, data: tx.data }];
```
mrice32, line 97: OOC, why did we decide to do all of these in separate cross chain calls rather than grouping all of them into the same txn?

Right now, we have a separate call for addMember, registerContract, removeMember, and setImplementation address in the finder.

I doubt this is a big deal, but there are two minor issues with the non-atomic approach:
1. Bridges generally process/send txns in order in the happy path, but I don't think this is a guarantee (for instance, one could have too little gas included and might need to be retried, while the one after does not, so they would be executed out of order).
2. If they are all done in the same L2 transaction, we can be assured that they either all succeed or all fail. If we do them separately, you could have the last 2 fail, leaving the contracts in an intermediate state. We could also get weird behavior if the first fails but the last 3 are still attempted. For instance, we could have the OO get added to the finder, but not be approved by the registry.

mrice32 (approved):
This is really awesome! Nice work @md0x! I have a few comments and nits, but overall, I think this works.

packages/scripts/src/upgrade-tests/162/1_Propose.ts
md0x: Good point @mrice32; I agree that these calls should be grouped by chain id to allow for atomic relays.

Because of time concerns, we finally run the script yesterday as it is now. I've left a TODO that I'll fix when we run the proposal's actions so we don't lose sight of the original script.

packages/scripts/src/upgrade-tests/162/1_Propose.ts
md0x: I was putting the call data in a JSON file so that the verification script could use it. It was not a very good idea.

I have changed the verification to use the decodeData script.

chrismaree (approved):

PR #3986, opened 2022-06-23
feat(dvm-response-publisher): add dvm response publisher bot

ryanwolhuter:
**Motivation**

We need a bot which ensures that added price requests have published prices. This bot queries past `PriceRequestAdded` events in the `Voting` contract and checks for ones that do not have a price by calling the `hasPrice` method of the `Voting` contract. When a request with a missing price is discovered, the bot calls the `publishPrice` method on the `OracleHub` contract with that price request's identifier.

**Summary**

* Add a new package with an implementation of the bot described above.

**Details**

**NB:** after discussion with @chrismaree, we have decided to pause this task until the contract interfaces for DVM v2 are available. Our reasoning is that the `hasPrice` and `publishPrice` methods require `ancillaryData` as an argument (along with `identifier` and `timestamp`) and unfortunately, the existing contract does not pass along this ancillary data when emitting the `PriceRequestAdded` event. While we could have devised a complicated solution to get this data, we concluded that this would be a waste of time, because the DVM v2 contracts have been updated to pass along the ancillary data when firing the event. We would therefore have to go back and change the solution to take this into account.

**Testing**

- [ ]  Ran end-to-end test, running the code as in production- [ ]  New unit tests created
- [x]  Untested


chrismaree:
I'm going to close this PR for now. we can come back to this later once we've deployed the V2 version of the dvm contracts.

PR #3989, opened 2022-06-24
chore(sdk)!: create oracle client interface for new oo contract integrations

daywiss:
BREAKING CHANGE: this slightly changes the request type which is used by the UI, app will need to be
adjusted to accomodate.

**Motivation**
We want to be able to support multiple oracle contracts. 

**Summary**
Originally i made  pr that altered the oracle client in a different way to support new oracle contracts. This is a different approach which isolates changes to the oracle service itself. This makes an oracle interface which needs to be implemented. This alters the design of the client somewhat, now the oracle clients are stateful now rather than stateless. Effort is taken not to break existing logic, but some types are slightly altered to simplify typing.  The existing functionality has not changed, only organization of the code. 

**Testing**

 This was tested by integrating it into the UI and looking for regressions on kovan. 

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/across/test/rateModel.e2e.ts
+++ b/packages/sdk/src/across/test/rateModel.e2e.ts
@@ -47,7 +47,11 @@ describe("rateModel", function () {
         );
         assert.fail("Should throw error");
       } catch (err) {
-        assert.equal(err.message.includes("method called before updating"), true);
+        if (err instanceof Error) {
+          assert.equal(err.message.includes("method called before updating"), true);
+        } else {
+          throw err;
+        }
       }
       rateModelDictionary.updateWithEvents(filteredEvents);
 
@@ -59,7 +63,11 @@ describe("rateModel", function () {
         );
         assert.fail("Should throw error");
       } catch (err) {
-        assert.equal(err.message.includes("before first UpdatedRateModel event"), true);
+        if (err instanceof Error) {
+          assert.equal(err.message.includes("before first UpdatedRateModel event"), true);
+        } else {
+          throw err;
+        }
       }
 
       // Cannot fetch rate model when dictionary has no entries for L1 token.
@@ -67,7 +75,11 @@ describe("rateModel", function () {
         await rateModelDictionary.getRateModelForBlockNumber(rateModelStore.address, blockSearchConfig.fromBlock - 1);
         assert.fail("Should throw error");
       } catch (err) {
-        assert.equal(err.message.includes("No updated rate model events for L1 token"), true);
+        if (err instanceof Error) {
+          assert.equal(err.message.includes("No updated rate model events for L1 token"), true);
+        } else {
+          throw err;
+        }
       }
 
       // Test for expected rate models.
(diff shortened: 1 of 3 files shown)


packages/sdk/src/clients/skinnyOptimisticOracle/client.ts
```diff
@@ -93,12 +93,12 @@ export function parseRequest(request: SolidityRequest) {
...
-    reward: request.proposedPrice.toString(),
-    finalFee: request.proposedPrice.toString(),
-    bond: request.proposedPrice.toString(),
-    customLiveness: request.proposedPrice.toString(),
+    resolvedPrice: request.resolvedPrice.toString(),
+    expirationTime: Number(request.expirationTime.toString()),
+    reward: request.reward.toString(),
+    finalFee: request.finalFee.toString(),
+    bond: request.bond.toString(),
+    customLiveness: request.customLiveness.toString(),
```
daywiss, line 101: this was a bug, fortunately not used yet

packages/sdk/src/oracle/client.ts
```diff
@@ -162,6 +164,19 @@ export class Client {
   }
 }
 
+function makeProvider(rpcUrls: string[]): FallbackProvider {
```
daywiss, line 167: this logic was pulled out of the write class unchanged for the most part. just so I can use the provider during configuration

packages/sdk/src/oracle/services/optimisticOracle.ts
```diff
@@ -1,66 +1,108 @@
...
+import { RequestPrice, ProposePrice, DisputePrice, Settle } from "../../clients/optimisticOracle";
 
-export class OptimisticOracle {
-  public readonly contract: optimisticOracle.Instance;
-  constructor(protected provider: Provider, protected address: string) {
+export type OptimisticOracleEvent = RequestPrice | ProposePrice | DisputePrice | Settle;
+
+export type { RequestPrice, ProposePrice, DisputePrice, Settle };
+
+export class OptimisticOracle implements OracleInterface {
```
daywiss, line 12: this class is the bulk of the new logic changes. it updates the oracle client to be stateful, and handle a lot of its own processing internally. It conforms to the oracle interface.

packages/sdk/src/oracle/types/interfaces.ts
```diff
@@ -0,0 +1,25 @@
+import { BigNumberish, BigNumber, Signer, TransactionResponse, TransactionReceipt } from "../types/ethers";
```
daywiss, line 1: this file defines the interfaces that are needed to generalize the oracle service

chrismaree (commented):
I dont really see unit tests here? perhaps this is part of the SDK that is not tested? did you try this out in the production OO UI to verify the correctness in prod?

daywiss:
@chrismaree I will double check that the end to end tests work, but i primarily tested this by integrating it into the frontend. I wouldnt suggest doing the integration after this upgrade though and wait until the next pr where we will add the skinny oo.

daywiss:
@chrismaree updated end to end tests that were failing.

mrice32 (approved):
LGTM!

PR #3997, opened 2022-06-29
feat: UMA2.0: Add simple Designated votingV2 to support two key contracts for new DVM design.

chrismaree:
**Motivation**

The PR adds a DesignatedVotingV2 contract to enable 2key contracts within the staking setup for DVMv2.

**Summary**

New contracts and tests added.

**Testing**

- [X]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/serverless-orchestration/src/ServerlessHub.js
+++ b/packages/serverless-orchestration/src/ServerlessHub.js
@@ -285,13 +285,10 @@ hub.post("/", async (req, res) => {
     }
 
     await delay(waitForLoggerDelay); // Wait a few seconds to be sure the the winston logs are processed upstream.
-    res
-      .status(500)
-      .send({
-        message:
-          errorOutput instanceof Error ? "A fatal error occurred in the hub" : "Some spoke calls returned errors",
-        output: errorOutput instanceof Error ? errorOutput.message : errorOutput,
-      });
+    res.status(500).send({
+      message: errorOutput instanceof Error ? "A fatal error occurred in the hub" : "Some spoke calls returned errors",
+      output: errorOutput instanceof Error ? errorOutput.message : errorOutput,
+    });
   }
 });
 


packages/core/contracts/common/implementation/Stakeable.sol
```diff
@@ -0,0 +1,50 @@
...
+pragma solidity ^0.8.0;
+
+import "../../oracle/interfaces/StakerInterface.sol";
+import "./Withdrawable.sol";
+
+/**
+ * @title Base contract that extends the Withdrawable contract enabling a specific role to stake ERC20 tokens against the
+ * Voting contract. Voting contract is fed in as a param rather than fetched from the finder to enable upgradability.
+ */
+abstract contract Stakeable is Withdrawable {
```
mrice32, line 15: Just OOC, why did you choose to make this a separate contract rather than just adding these methods to the DesignatedVotingV2 contract?

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
```diff
@@ -0,0 +1,120 @@
...
+import "../interfaces/VotingAncillaryInterface.sol";
+import "../interfaces/FinderInterface.sol";
+import "./Constants.sol";
+
+/**
+ * @title Proxy to allow voting from another address.
+ * @dev Allows a UMA token holder to designate another address to vote on their behalf.
+ * Each voter must deploy their own instance of this contract.
+ */
+contract DesignatedVotingV2 is Stakeable {
```
mrice32, line 15: If this isn't multicallable, execute unstake always has to be a separate transaction from the withdraw. Do you think that's enough reason to make it multicall?

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
```diff
@@ -0,0 +1,120 @@
...
+    /****************************************
+     *    INTERNAL VARIABLES AND STORAGE    *
+     ****************************************/
+
+    enum Roles {
```
mrice32, line 20: I thought we were trying to move off of multirole, now that OZ has a comparable library?

Note: something that the current version has that would be really desirable to keep in a new version would be for a layman to be able to easily query the owner and designated voter via etherscan. It's a feature we use all the time.

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
```diff
@@ -0,0 +1,120 @@
...
+     * @notice Forwards a reward retrieval to Voting.
+     * @dev Rewards are added to the tokens already held by this contract.
+     * @return amount of rewards that the user should receive.
+     */
+    function retrieveRewards() public onlyRoleHolder(uint256(Roles.Voter)) returns (uint256) {
+        StakerInterface voting = StakerInterface(address(_getVotingContract()));
+        uint256 rewardsMinted = voting.withdrawRewards();
+        IERC20(address(voting.votingToken())).approve(address(voting), rewardsMinted);
+        voting.stake(rewardsMinted);
+    }
```
mrice32, line 115: Is there any way for a user to pull the rewards into the contract rather than restaking them, so they don't have to always deal with a lockup?

IIRC, this was the reason we wanted the rewards to be disbursed to the user by default rather than being restaked by default.

packages/core/contracts/common/implementation/Stakeable.sol
```diff
@@ -0,0 +1,50 @@
...
+    using SafeERC20 for IERC20;
+
+    uint256 private roleId;
```
mrice32, line 18: Where is this value set? Who is it supposed to be set to? The owner or the designated voter?

packages/core/contracts/oracle/implementation/DesignatedVotingV2Factory.sol
```diff
@@ -0,0 +1,43 @@
...
+    mapping(address => DesignatedVotingV2) public designatedVotingContracts;
+
+    /**
+     * @notice Construct the DesignatedVotingFactory contract.
+     * @param finderAddress keeps track of all contracts within the system based on their interfaceName.
+     */
+    constructor(address finderAddress) {
+        finder = finderAddress;
+    }
+
```
mrice32, line 22: Might not be appropriate for this PR, but is there anything we could add to this contract to make the migration process from the old DV contracts to the new ones easier? For instance, maybe we could have a function that permissionlessly replicates the configuration of an old DV contract in a new DV contract and adds the user to the mapping all in one go 🤷. Just something to think about.

mrice32 (approved):
Looks great! A few comments

package.json
```diff
@@ -9,8 +9,8 @@
     "lint-fix": "yarn eslint --fix && yarn prettier --write",
     "eslint": "eslint './**/*.js' './**/*.ts'",
     "prettier": "prettier './**/*.js' './**/*.sol' './**/*.md' './**/*.ts'",
-    "test": "lerna run --stream --concurrency=4 test",
-    "test-concurrent": "lerna run --stream --concurrency=4 test --ignore @uma/serverless-orchestration --ignore @uma/affiliates && lerna run --stream --concurrency=1 test --scope @uma/serverless-orchestration --scope @uma/affiliates",
+    "test": "NODE_OPTIONS=--max-old-space-size=4096 lerna run --stream --concurrency=4 test",
```
chrismaree, line 12: ^

packages/core/contracts/common/implementation/Stakeable.sol
chrismaree: to follow the same pattern that the DesignatedVoting v1 had. can roll it all into one but by doing it this way it follows the same pattern.

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
chrismaree: good call! it should be marked as multicall. I will leave this for a follow on PR where we add it to the other UMA.20 contracts(`Voting.sol`).

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
chrismaree: I kept it like this as I wanted this to be a minimum viable drop-in for the current system by keeping _all_ logic the same other than adding support for the new staking logic. as a separate PR we can strip out this and move to the OZ stuff but then it should be done in the voting.sol contract at the same.

packages/core/contracts/oracle/implementation/DesignatedVotingV2.sol
chrismaree: I chose to do it like this as this now behaves the _same_ as the current Designated voting contract. claiming rewards in the current design just sends them to the designated voting contract. in this implementation by `retrieveRewards` re-staking, this behaves the same as the current design.

if we wanted voters to be able to access these this has implications for how we use this as a vesting mechanism and I don't think this decision has been made yet.

packages/core/contracts/common/implementation/Stakeable.sol
chrismaree: it's set implicitly via the `Withdrawable.sol` contract (the existing DesignatedVoting contract that lets us set this. I'm going to re-work this to make it more explicit.

packages/core/contracts/oracle/implementation/DesignatedVotingV2Factory.sol
chrismaree: ye, we'll need to think more about this and can leave it for a follow on PR.

mrice32 (approved):
LGTM!

packages/core/contracts/oracle/implementation/DesignatedVotingV2Factory.sol
```diff
@@ -0,0 +1,43 @@
...
+pragma solidity ^0.8.0;
+
+import "../../common/implementation/Withdrawable.sol";
+import "./DesignatedVotingV2.sol";
+
+/**
+ * @title Factory to deploy new instances of DesignatedVotingV2 and look up previously deployed instances.
+ * @dev Allows off-chain infrastructure to look up a hot wallet's deployed DesignatedVoting contract.
+ */
+contract DesignatedVotingV2Factory is Withdrawable {
```
md0x, line 11: Why is `DesignatedVotingV2Factory` `Withdrawable`?

md0x (approved):
Looks good to me, just left one comment/question.

packages/core/contracts/oracle/implementation/DesignatedVotingV2Factory.sol
chrismaree: it was also in v1 so I left it the same here. I dont think it really needs this inheritance path but I wanted to keep things as similar to what we had before and can remove/add new logic in subsequent PRs.

PR #4006, opened 2022-07-01
feat: simplify slash calculation

md0x:
Signed-off-by: Pablo Maldonado <pablo@umaproject.org>

**Motivation**

The previous slash calculation implied some extra floors, resulting in a loss of precision. We can simplify to do only one division.

**Summary**

Simplifying the calculation so that there is only one division

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


chrismaree (approved):
LGTM! thanks for finding and fixing this

PR #4009, opened 2022-07-04
feat(sdk)!: add skinny oo client and separate factories

daywiss:
BREAKING CHANGE: instantiating oracle client now requires passing configs keyed by optimistic or skinny oracle types

**Motivation**

We want to add skinny oo to the oracle client

**Summary**

This makes the required changes to support the skinny oo, mainly this is adding the skinny stateful client and adding a new factory function to allow user to create clients of multiple types. docs are also updated.  THis is a breaking change, and changes are tested on the OO UI frontend, there will need to be a new PR on the FE to integrate these changes.

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

### DIFF
--- a/packages/sdk/src/oracle/README.md
+++ b/packages/sdk/src/oracle/README.md
@@ -4,17 +4,20 @@ Interact with UMA's Optimistic Oracle using an eventful client designed for a fr
 
 # Quick Start
 
-Currently theres a single factory which creates the client and allows you to subscribe to state changes.
+In order to start a client, you need to know if you will be interfacing with the skinny optimistic oracle, or
+standard optimistic oracle. They have an identical interface, use the factory and specify the configuration for which type or
+types you intend to use.
 
 ```js
 import {oracle} from '@uma/sdk'
 
-// create your configuration object with known addresses
-type Config = oracle.types.state.Config
+type PartialConfig = oracle.types.state.PartialConfig
+type PartialConfigTable = oracle.types.state.PartialConfigTable
 
 // creating a config for mainnet usage add any additional chains you want following the same format
-export const config = {
+const config:PartialConfig = {
   chains: {
+    // adding a mainnet chain configuration
     1:{
       chainName: "Ethereum",
       rpcUrls: [process.env.CUSTOM_NODE_URL],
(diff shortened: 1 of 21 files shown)


packages/sdk/src/oracle/README.md
```diff
@@ -4,17 +4,20 @@ Interact with UMA's Optimistic Oracle using an eventful client designed for a fr
 
 # Quick Start
 
-Currently theres a single factory which creates the client and allows you to subscribe to state changes.
+In order to start a client, you need to know if you will be interfacing with the skinny optimistic oracle, or
+standard optimistic oracle. They have an identical interface, use the factory and specify the configuration for which type or
+types you intend to use.
```
daywiss, line 9: this is the readme, updated with examples of how to start the clients

packages/sdk/src/oracle/factory.ts
```diff
@@ -0,0 +1,28 @@
...
+import { State, OracleType, PartialConfigTable } from "./types/state";
+import { Emit } from "./store";
+import SkinnyFactory from "./skinnyFactory";
+import OptimisticFactory from "./optimisticFactory";
+
+export type PublicEmit = (oracleType: OracleType, state: State, prev: State) => void;
+const EventHandler = (oracleType: OracleType, publicEmit: PublicEmit): Emit => (state: State, prev: State) =>
+  publicEmit(oracleType, state, prev);
+
+export default (configTable: PartialConfigTable, emit: PublicEmit): ClientTable => {
```
daywiss, line 13: the factory now requires you key the config by the oracle type. you can only create 1 instance of each oracle type from the factory.  This is because the combined view of all requests ( across all chains and all oracles) need a way to locate the correct client type. This is so you can go into details and make calls, like propose, dispute, etc to the correct client.

packages/sdk/src/oracle/services/skinnyOptimisticOracle.ts
```diff
@@ -0,0 +1,159 @@
+import assert from "assert";
```
daywiss, line 1: in the future this is the only file + a factory you will really need to add for more oracles (unless the interface isnt sufficient)

packages/sdk/src/oracle/services/skinnyOptimisticOracle.ts
```diff
@@ -0,0 +1,159 @@
...
+    resolvedPrice: request.resolvedPrice,
+    expirationTime: request.expirationTime,
+    reward: request.reward,
+    finalFee: request.finalFee,
+    bond: request.bond,
+    customLiveness: request.customLiveness,
+  };
+}
+
+export class OptimisticOracle implements OracleInterface {
```
mrice32, line 61: Something I find a little confusing about this file is that the class and the getOptimisticOracleAddress are named the same as the other file, but they apply to the skinny optimistic oracle, not the optimistic oracle. Do these need to have the same names? It might improve readability if the skinny name was at least reflected in the names of the exported classes and functions.

mrice32 (approved):
LGTM!

packages/sdk/src/oracle/README.md
```diff
@@ -34,15 +37,29 @@ export const config = {
   },
 };
 
+// this starts both a skinny and optimistic oracle, you have to make sure to omit the oracle address
+// in the config to let them auto populate, or make sure you set the addresses differently for each type
+const configTable = {
+  [oracle.types.state.OracleType.Optimistic]:config,
+  [oracle.types.state.OracleType.Skinny]:config,
+}
```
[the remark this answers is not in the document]
daywiss: this is in the readme so it doenst get fixed by linter, but ive updated it manually

PR #4010, opened 2022-07-05
feat: DVM2.0 Add ability to delete spam requests from the DVM.

chrismaree:
**Motivation**

**Summary**

**Details**

**Testing**

- [ ]  New unit tests created
- [ ]  Untested

**Issue(s)**

Fixes #XXXX


packages/common/src/HardhatConfig.ts
```diff
@@ -90,6 +90,7 @@ export function getHardhatConfig(
         blockGasLimit: 15_000_000,
         timeout: 1800000,
         testBlacklist,
+        allowUnlimitedContractSize: true,
```
chrismaree, line 93: you can ignore this. the contract is below compile size.

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -351,13 +384,99 @@ contract VotingV2 is
...
+                // the deleted requests via a "jump" over the removed elements from the array.
+                deletedRequestJumpMapping[startIndex] = endIndex;
+            }
+
+            // Return the spamDeletionProposalBond.
+            votingToken.transfer(spamDeletionProposals[proposalId].proposer, spamDeletionProposalBond);
+        }
+        // Else, the spam deletion request was voted down. In this case we send the spamDeletionProposalBond to the store.
+        else {
+            votingToken.transfer(finder.getImplementationAddress(OracleInterfaces.Store), spamDeletionProposalBond);
```
mrice32, line 468: We probably need to be careful here. Is it possible for the spamDeletionProposalBond to change between the proposal and here? Maybe it's fine to just store the initial bond value in the struct since gas isn't a concern (this would be almost never called).

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -351,13 +384,99 @@ contract VotingV2 is
...
+        bytes32 identifier = SpamGuardIdentifierLib._constructIdentifier(proposalId);
+
+        (bool hasPrice, int256 resolutionPrice, ) =
+            _getPriceOrError(identifier, spamDeletionProposals[proposalId].requestTime, "");
+        require(hasPrice, "Price not yet resolved");
+
+        // If the price is 1e18 then the spam deletion request was correctly voted on to delete the requests.
+        if (resolutionPrice == 1e18) {
+            // Delete the price requests associated with the spam.
+            for (uint256 i = 0; i < spamDeletionProposals[proposalId].spamRequestIndices.length; i++) {
```
mrice32, line 442: If there's a _lot_ of spam, would this method break because it's O(n) in spam requests?

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -279,11 +304,14 @@ contract VotingV2 is
         // Traverse all requests from the last considered request. For each request see if the voter voted correctly or
         // not. Based on the outcome, attribute the associated slash to the voter.
         for (uint256 i = voterStake.lastRequestIndexConsidered; i < priceRequestIds.length; i++) {
+            if (deletedRequestJumpMapping[i] != 0) i = deletedRequestJumpMapping[i] + 1;
```
mrice32, line 307: Interesting approach -- pretty cool.

packages/core/contracts/oracle/implementation/VotingV2.sol
```diff
@@ -351,13 +384,99 @@ contract VotingV2 is
...
+
             lastRequestIndexConsidered = i + 1;
         }
     }
 
+    /****************************************
+     *       SPAM DELETION FUNCTIONS        *
+     ****************************************/
+
+    function signalRequestsAsSpamForDeletion(uint256[2][] memory spamRequestIndices) public {
```
mrice32, line 403: Could someone spam requests and then spam this with so many jumps that it breaks voters' ability to traverse the jumps? I think that's hard unless this function is less expensive than the jumping, but still might be possible since they could technically call this function multiple times.

